PR #2738, opened 2020-08-17
docs(storage): Fix the documentation link

AntonNguyen:


storage/storage.go
```diff
@@ -349,7 +349,7 @@ var (
 )
 
 // v2SanitizeHeaders applies the specifications for canonical extension headers at
-// https://cloud.google.com/storage/docs/access-control/signed-urls#about-canonical-extension-headers.
+// https://cloud.google.com/storage/docs/authentication/canonical-requests#about-headers.
```
tritone, line 352: I don't think this is correct for `v2SantizeHeaders` because the link you add is for V4 signing only.

@frankyn any thoughts on what is the appropriate link here?

AntonNguyen: Nice catch. I've updated it with this link, which looks like's it's for v2 urls: https://cloud.google.com/storage/docs/access-control/signed-urls-v2#about-canonical-extension-headers

AntonNguyen:
@tritone Is there anything else I need to do for this PR to be merged?

PR #2739, opened 2020-08-17
chore: update pkg doc on timeouts

noahdietz:
This updates the `cloud.google.com/go` package documentation regarding timeouts to include the upcoming default deadline changes.

This change should be included once regeneration with default deadlines happens.

Motivated by work on https://github.com/googleapis/gapic-generator-go/issues/411


noahdietz:
I'm going to leave this open until I've released the new generator version, updated it here, and regen has occured

doc.go
```diff
@@ -34,9 +34,16 @@ in this package for details.
...
-By default, all requests in sub-packages will run indefinitely, retrying on transient
-errors when correctness allows. To set timeouts or arrange for cancellation, use
-contexts. See the examples for details.
+By default, all Unary methods will have a default deadline applied to the
+context provided at call time, unless a deadline is already set. Streaming
+methods have no default deadline and will run indefinitely. To set timeouts or
+arrange for cancellation, use contexts. See the examples for details. Transient
+errors will be retried when correctness allows.
+
+To opt out of default deadlines, set the environment variable
```
tbpg, line 43:
```suggestion
To opt out of default deadlines, set the temporary environment variable
```

doc.go
```diff
@@ -34,9 +34,16 @@ in this package for details.
...
+GOOGLE_API_GO_EXPERIMENTAL_DISABLE_DEFAULT_DEADLINE to "true" prior to client
+creation. This affects all Google Cloud Go client libraries. This opt-out
+mechanism will be removed in a future release.
```
tbpg, line 46:
```suggestion
mechanism will be removed in a future release. File an issue at
https://github.com/googleapis/google-cloud-go if the default deadlines
cannot work for you.
```

doc.go
```diff
@@ -34,9 +34,16 @@ in this package for details.
 
 Timeouts and Cancellation
 
-By default, all requests in sub-packages will run indefinitely, retrying on transient
-errors when correctness allows. To set timeouts or arrange for cancellation, use
-contexts. See the examples for details.
+By default, all Unary methods will have a default deadline applied to the
```
tbpg, line 37: Is there another word for "Unary"? That's not used for Go code, so it's a little confusing.

doc.go
```diff
@@ -34,9 +34,16 @@ in this package for details.
...
+context provided at call time, unless a deadline is already set. Streaming
```
tbpg, line 38:
```suggestion
context provided at call time, unless a context deadline is already set. Streaming
```

doc.go
noahdietz: Hm yeah good point...how about

"By default, standard CRUD methods, like Create or Get, will have a ..."

tbpg: I think it's better if we leave out "CRUD." What if we say "non-streaming, like Create or Get, ..."?

noahdietz: Yeah, that's better!

noahdietz: Done

Issue #2750, opened 2020-08-19
computer: unable to read metadata from GKE


thecodejunkie:
**Client**

Metadata

**Environment**

Google Kubernetes Engine

**Expected behavior**

Be able to retrieve an ID token from the metadata server when running inside GKE

**Actual behavior**

When following the code example (Go) for Cloud Run service-to-service authentication at https://cloud.google.com/run/docs/authenticating/service-to-service I keep getting an error when trying to retrieve an ID token when running inside GKE. I.e my caller is living in GKE and it needs to call a Cloud Run hosted service. When I try and retrieve the ID token I get the following error

```
2020/08/13 11:21:55 metadata: GCE metadata "/instance/service-accounts/default/identity?audience=https://[REDACTED]" not defined
```

However, if I manually ssh into the Kubernetes node and curl (by directly calling http://metadata.google.internal) the metadata server then I can get the ID token

codyoss:
Thanks for the report, I will investigate and get back to you.

codyoss:
I tried a few different things, but they all seemed to work for me.
- Used the metadata client, like the link provided.
- Used the [idtoken](https://pkg.go.dev/google.golang.org/api/idtoken?tab=doc) package.
- Tried to make a request from my go program directly, with `metadata.google.internal` as the base host.
- Tried to make a request from my go program directly, with `169.254.169.254` as the base host.

Could you please provide me the following so I can better assist you here:
- What version is your cluster? For example I was using `1.15.12-gke.2`.
- What version of `cloud.google.com/go` are you pulling in?
- Are you by chance setting the following environment variable: `GCE_METADATA_HOST `?

thecodejunkie:
> What version is your cluster? For example I was using `1.15.12-gke.2`.

The master is running on `1.15.12-gke.2` and the node pool is on `1.15.11-gke.15`

> What version of `cloud.google.com/go` are you pulling in?

Right now we have `0.59.0` and I noticed that `0.64.0` is available. I will try and update the dependency and see if that resolves it

> Are you by chance setting the following environment variable: `GCE_METADATA_HOST` ?

No

thecodejunkie:
We seem to be getting the same behavior with `0.64.0`. The only config we have enabled on the cluster is workload identity, but that shouldn't be interfering. The cluster (dev environment) is using the default node pool service account. The deployment uses a non-default service account to run the pod, but that shouldn't be an issue either.

codyoss:
[...]

PR #2760, opened 2020-08-20
fix(compute/metadata): remove leading slash for Get suffix

codyoss:
Updates #2750

### DIFF
--- a/compute/metadata/metadata.go
+++ b/compute/metadata/metadata.go
@@ -296,6 +296,7 @@ func (c *Client) getETag(suffix string) (value, etag string, err error) {
 		// being stable anyway.
 		host = metadataIP
 	}
+	suffix = strings.TrimLeft(suffix, "/")
 	u := "http://" + host + "/computeMetadata/v1/" + suffix
 	req, err := http.NewRequest("GET", u, nil)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


compute/metadata/metadata.go
```diff
@@ -296,6 +296,7 @@ func (c *Client) getETag(suffix string) (value, etag string, err error) {
 		// being stable anyway.
 		host = metadataIP
 	}
+	suffix = strings.TrimLeft(suffix, "/")
 	u := "http://" + host + "/computeMetadata/v1/" + suffix
```
tbpg, line 300: We should probably use `path.Join` here instead of manually handling slashes.

codyoss: Suffix's can contain query parameters that can contain URLs, using `path.Join` will break users that are not URL encoding their parameters today. That is, if we path.Join on the whole suffix. I suppose we could split on `?`.... 

For example
```go
// suffix := "instance/service-accounts/default/identity?audience=http://example.com"
fmt.Println(path.Join(host, "/computeMetadata/v1/", suffix))
// Output: "169.254.169.254/computeMetadata/v1/instance/service-accounts/default/identity?audience=http:/example.com"
```

Maybe this is okay though as people should be encoding their query parameters... Thoughts?

PR #2770, opened 2020-08-24
test(spanner): use iter next instead of doing nothing

sukeesh:
Fixing TODO.

Replace `_ = iter` with 
iterate using Next

### DIFF
--- a/spanner/examples_test.go
+++ b/spanner/examples_test.go
@@ -442,7 +442,15 @@ func ExampleReadOnlyTransaction_ReadWithOptions() {
 			Index: "UsersByEmail",
 			Limit: 100,
 		})
-	_ = iter // TODO: iterate using Next or Do.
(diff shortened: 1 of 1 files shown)


sukeesh:
Hi @skuruppu, This is my first contribution to googleapis/google-cloud-go. So, would like to start my contributions from `*_test.go` files until I get familiar with the code/repo.

skuruppu:
Thanks very much for making this pull request @sukeesh. I will assign @hengfengli since he may be able to review this sooner than I could.

hengfengli (commented):
I am not sure if this is necessary. Because here `// TODO: iterate using Next or Do.` is similar to `// TODO: Handle error.`. `TODO` is for users not us. We suggest users to decide what they want to use with `iter` instead of giving a specific example.

PR #2773, opened 2020-08-25
fix(bigtable): change max valid Timestamp value

### LINKED ISSUE #1790: bigtable: emulator rejects puts with reversed timestamps
**Client**

* bigtable-hbase-1.x-shaded version 1.6.0
* hbase-shaded-client 1.4.5

**Environment**

OSX and Linux

**Code**

```java

// Ideally, we'd just use Long.MAX_VALUE for this. However, the bigtable client needs to translate the value into
// microseconds as that is what Bigtable uses. This means they take the value we give them and they multiply it by
// 1000 which ends up blowing the size of a Long if we don't accommodate the expansion. Therefore, we set our max
// to the max value we can use but still accommodate the multiply by 1000
public static final long REVERSE_EPOCH = Long.MAX_VALUE / 1000L;    

private long reverseTimestamp(long modTs) {
    Preconditions.checkArgument(modTs >= 0);
    return REVERSE_EPOCH - modTs;
}

private void doReversedTimestampPut(Table table) throws IOException {
    byte[] key = "key".getBytes();

    long timestamp = System.currentTimeMillis();
    Put put = new Put(key, System.currentTimeMillis());
    put.add(FAMILY, CREATION_TIMESTAMP_COL, reverseTimestamp(timestamp), Bytes.toBytes(timestamp));

    table.put(put);
}
```

```java
org.apache.hadoop.hbase.DoNotRetryIOException: Failed to perform operation. Operation='put', projectId='who_cares', tableName='not_me', rowKey='key'

	at com.google.cloud.bigtable.hbase.AbstractBigtableTable.logAndCreateIOException(AbstractBigtableTable.java:536)
	at com.google.cloud.bigtable.hbase.AbstractBigtableTable.put(AbstractBigtableTable.java:354)

<SNIP>

Caused by: com.google.bigtable.repackaged.io.grpc.StatusRuntimeException: UNKNOWN: invalid timestamp 9221789815520484000
	at com.google.bigtable.repackaged.io.grpc.Status.asRuntimeException(Status.java:517)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.async.AbstractRetryingOperation.getBlockingResult(AbstractRetryingOperation.java:476)
	at com.google.bigtable.repackaged.com.google.cloud.bigtable.grpc.BigtableDataGrpcClient.mutateRow(BigtableDataGrpcClient.java:236)
	at com.google.cloud.bigtable.hbase.AbstractBigtableTable.put(AbstractBigtableTable.java:351)
	... 28 more
```

**Expected behavior**

Reversed timestamps are allowed on mutations on the bigtable emulator.

**Actual behavior**

Timestamps after the year 10,000 are not allowed, which was introduced as a part of [this change](https://code-review.googlesource.com/c/gocloud/+/27650/).

**Additional context**

We're doing timestamp reversing as described [here](https://cloud.google.com/bigtable/docs/schema-design):

> If you usually retrieve the most recent records first, you can use a reversed timestamp in the row key by subtracting the timestamp from your programming language's maximum value for long integers (in Java, java.lang.Long.MAX_VALUE). With a reversed timestamp, the records will be ordered from most recent to least recent.

AlisskaPie:
As Timestamp could be of any int64 value, the maximum value is microseconds with [millisecond granularity ](https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#google.bigtable.admin.v2.Table.TimestampGranularity)seems should be
 maxInt64 - maxInt64%1000.

One check for passing this max value when creating a table without errors, the second check that this value is the **maximum**.

Closes #1790


tritone:
@kolea2 PTAL

bigtable/bttest/inmem_test.go
```diff
@@ -410,6 +411,53 @@ func (s *MockReadRowsServer) Send(resp *btpb.ReadRowsResponse) error {
 	return nil
 }
 
+func TestCheckTimestampMaxValue(t *testing.T) {
```
tritone, line 414: Could you add a short comment to this test to indicate what it's supposed to be verifying? You can cite the bug in question.

bigtable/bttest/inmem_test.go
```diff
@@ -410,6 +411,53 @@ func (s *MockReadRowsServer) Send(resp *btpb.ReadRowsResponse) error {
...
+			Mutation: &btpb.Mutation_SetCell_{SetCell: &btpb.Mutation_SetCell{
+				FamilyName:      "cf0",
+				ColumnQualifier: []byte("col"),
+				TimestampMicros: maxTimestamp + 1000,
+				Value:           []byte{},
+			}},
+		}},
+	}
+	if _, err := s.MutateRow(ctx, mreq2); err == nil {
+		t.Fatalf("want TimestampMicros rejection, got acception: %v", err)
```
tritone, line 458: acception --> acceptance.

bigtable/bttest/inmem_test.go
AlisskaPie: Of course, done

bigtable/bttest/inmem_test.go
AlisskaPie: Done

AlisskaPie:
Thank you for review

bigtable/bttest/inmem.go
```diff
@@ -61,7 +62,7 @@ const (
 	minValidMilliSeconds = 0
 
 	// MilliSeconds field of the max valid Timestamp.
-	maxValidMilliSeconds = int64(time.Millisecond) * 253402300800
+	maxValidMilliSeconds = math.MaxInt64 - math.MaxInt64%1000
```
kolea2, line 65: can you please add why this value (`math.MaxInt64 - math.MaxInt64%1000`) is used in a comment?

AlisskaPie: Done

PR #2779, opened 2020-08-26
feat(bigquery): add support for bignumeric

shollyman:
This PR adds basic support for the BIGNUMERIC type in BigQuery.

This library has several cases where it tries to infer the appropriate BigQuery type from a native Go type.  For `big.Rat` types, we continue the existing behavior of mapping them to NUMERIC, as the big.Rat doesn't
provide a general way of indicating desired precision or scale to determine whether BIGNUMERIC is a more
appropriate mapping.


bigquery/params.go
```diff
@@ -233,6 +234,7 @@ func paramValue(v reflect.Value) (bq.QueryParameterValue, error) {
 		return res, nil
 
 	case typeOfRat:
+		// TODO: blerg. Need an explicit destination type to disambiguate if we should send a numeric or bignumeric string.
```
tbpg, line 237: Update comment? :stuck_out_tongue:

bigquery/schema.go
```diff
@@ -346,6 +350,9 @@ func inferFieldSchema(fieldName string, rt reflect.Type, nullable bool) (*FieldS
 	case typeOfDateTime:
 		return &FieldSchema{Required: true, Type: DateTimeFieldType}, nil
 	case typeOfRat:
+		// TODO: big.Rat has arbitrary precision, so we can't discrimate without
```
tbpg, line 353: Should this be an open issue? Also, typo with discrimate.

bigquery/value.go
```diff
@@ -786,6 +800,14 @@ const (
 
 	// NumericScaleDigits is the maximum number of digits after the decimal point in a NUMERIC value.
 	NumericScaleDigits = 9
+
+	// BigNumericPrecisionDigits is the maximum number of full digits in a BIGNUMERIC value.
+	// BigNumeric does support a partial digit beyond this.
```
tbpg, line 805: What does a partial digit mean?

bigquery/value_test.go
```diff
@@ -729,20 +744,23 @@ func TestStructSaverErrors(t *testing.T) {
...
+		in             *big.Rat
+		wantNumeric    string
+		wantBigNumeric string
 	}{
-		{big.NewRat(2, 3), "0.666666667"}, // round to 9 places
-		{big.NewRat(1, 2), "0.500000000"},
-		{big.NewRat(1, 2*1e8), "0.000000005"},
-		{big.NewRat(5, 1e10), "0.000000001"},   // round up the 5 in the 10th decimal place
-		{big.NewRat(-5, 1e10), "-0.000000001"}, // round half away from zero
+		{big.NewRat(2, 3), "0.666666667", "0.66666666666666666666666666666666666667"}, // rounding last digit for both numeric/bignumeric
```
tbpg, line 753: Consider adding a name field to the test cases. Then, we don't need these comments and they make test debugging easier.

tbpg (changes requested):
Need to update the PR body? Or accidental review request?

I'm not familiar with this code. So, it seems to make sense, but this is mostly a style review.

bigquery/value_test.go
shollyman: Used the index rather than labelling them with a description.

shollyman: And now it's description-based

bigquery/params.go
shollyman: updated

bigquery/schema.go
shollyman: Updated comments

tbpg (approved):
"Partial digit" question is still open, otherwise LGTM.

bigquery/value.go
shollyman: A nuance of the internal representation.  Removed reference to this, as the bit representation isn't relevant to most users.

PR #2794, opened 2020-08-31
fix: support saving custom types that implements PropertyLoadSaver

### LINKED ISSUE #2611: datastore: PropertyLoadSaver ignored with nil map types
**Client**

Datastore

**Environment**

Firestore on Datastore mode
MacOS 10.15.5

**Go Environment**

```
$ go version
go version go1.14.6 darwin/amd64
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/yuryu/Library/Caches/go-build"
GOENV="/Users/yuryu/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/yuryu/go"
GOPRIVATE=""
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/yuryu/src/triton/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/kt/8436st1j7jb_pmp8wwcm7z38009x4l/T/go-build503312083=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

**Code**

```go
package main

import(
	"fmt"
	"cloud.google.com/go/datastore"
)

type Map map[int]int

func (_ *Map) Load(_ []datastore.Property) error {
	return nil
}

func (_ *Map) Save() ([]datastore.Property, error) {
	return []datastore.Property{}, nil
}

type Struct struct {
	Map Map
}

func main() {
	s := &Struct{}
	fmt.Println(datastore.SaveStruct(s))
}
```

**Expected behavior**

The datastore client uses Save() and Load() methods defined to load and save PropertyMap.
The [document](https://pkg.go.dev/cloud.google.com/go/datastore?tab=doc#hdr-The_PropertyLoadSaver_Interface) says 
> An entity's contents can also be represented by any type that implements the PropertyLoadSaver interface. This type may be a struct pointer, but it does not have to be.

**Actual behavior**

"[] datastore: unsupported struct field type: main.Map" error is returned if Map is nil. If I change main() to
```go
func main() {
	s := &Struct{make(Map)}
	fmt.Println(datastore.SaveStruct(s))
}
```
the error disappears.

IlyaFaer:
Fixes #2611 

It looks like no one expected custom type to be not a `struct` type. Adding conditions for such a case.


datastore/save.go
```diff
@@ -138,6 +138,17 @@ func reflectFieldSave(props *[]Property, p Property, name string, opts saveOpts,
...
+		vi := v.Addr().Interface()
+		if pSaver, ok := vi.(PropertyLoadSaver); ok {
+			val, err := pSaver.Save()
+			if err != nil {
+				return fmt.Errorf("datastore: unsupported struct field type: %v", v.Type())
+			}
+			p.Value = val
+		}
+	}
+
```
IlyaFaer, line 151: I've tried to use the existing functions (which are doing the similar things), but they doesn't seem to be correct solution for this case

IlyaFaer:
I've also noticed that one of the error checks is ommited:
https://github.com/googleapis/google-cloud-go/blob/7a8607e91a9c8ffcc936d6d74bc18c2f2d5adc2f/datastore/prop.go#L337-L338
Should be fixed a believe.

datastore/save.go
```diff
@@ -138,6 +138,17 @@ func reflectFieldSave(props *[]Property, p Property, name string, opts saveOpts,
 		}
 	}
 
+	if v.CanAddr() {
+		vi := v.Addr().Interface()
+		if pSaver, ok := vi.(PropertyLoadSaver); ok {
+			val, err := pSaver.Save()
+			if err != nil {
+				return fmt.Errorf("datastore: unsupported struct field type: %v", v.Type())
```
tritone, line 146: The error message seems misleading-- this is the case where the interface is defined but an error occurs while saving.

datastore/save_test.go
```diff
@@ -288,6 +288,20 @@ func TestSaveEmptySlice(t *testing.T) {
 	}
 }
 
+type Map map[int]int
```
tritone, line 291: These can go inside the test

IlyaFaer: My concern here is methods - nested method declarations are not allowed in Go (at least I never saw anything like this). It doesn't allow me to save the file with this types declared inside the test.

tritone: Ah yeah of course, sorry I spaced on that. Just add a comment indicating the purpose/what test it is for.

datastore/save.go
```diff
@@ -138,6 +138,17 @@ func reflectFieldSave(props *[]Property, p Property, name string, opts saveOpts,
 		}
 	}
 
+	if v.CanAddr() {
+		vi := v.Addr().Interface()
+		if pSaver, ok := vi.(PropertyLoadSaver); ok {
+			val, err := pSaver.Save()
+			if err != nil {
+				return fmt.Errorf("field save error: %v", err)
```
codyoss, line 146: Should this be prefixed with "datastore: "? I am not sure if that convention is used here or not.

IlyaFaer: As this error is going to be returned by user defined method, I think it shouldn't be prefixed with `datastore`

tritone:
LGTM pending approval by @benwhitehead

BenWhitehead:
Seems like a good fix. Merge at your leisure @tritone.

tritone (approved):

PR #2799, opened 2020-08-31
feat(workflows): start generating apiv1beta

codyoss:


internal/.repo-metadata-full.json
```diff
@@ -838,5 +838,13 @@
     "client_library_type": "generated",
     "docs_url": "https://pkg.go.dev/cloud.google.com/go/webrisk/apiv1beta1",
     "release_level": "beta"
+  },
+  "cloud.google.com/go/workflows/apiv1beta": {
+    "distribution_name": "cloud.google.com/go/workflows/apiv1beta",
```
tbpg, line 843: apiv1beta, not apiv1beta**1**? Generation works, so this seems correct, but wanted to confirm since this flows into package name.

codyoss: Yes, `apiv1beta`. Historically if the pathing in googleapis does not contain a number suffix neither does our package here.

PR #2804, opened 2020-09-01
feat(datastore): add opencensus tracing/stats support

### LINKED ISSUE #2805: datastore: add OpenCensus spans to all Datastore clients
[go/datastore-opencensus](https://docs.google.com/document/d/101LpAEjUEQL6M64XmCvNmtWZtc6uoGc3gG9fGjau2cI/edit#)

AlisskaPie:
`reserveIds`:
There wasn't any wrapping function for this operation, only generated in code, so I added one in client.go, is it okay?
P.S. I guess some higher-level functions and tests for `reserveIds` for the following files could be added in another issue:
- integration_test.go
- key.go
- example_test.go 
- datastore_ test.go

Closes #2805


codyoss:
FYI, I think we want to be careful here. We should honor [WithTelemetryDisabled](https://pkg.go.dev/google.golang.org/api/option?tab=doc#WithTelemetryDisabled).

datastore/client.go
```diff
@@ -50,6 +51,9 @@ func newDatastoreClient(conn grpc.ClientConnInterface, projectID string) pb.Data
 }
 
 func (dc *datastoreClient) Lookup(ctx context.Context, in *pb.LookupRequest, opts ...grpc.CallOption) (res *pb.LookupResponse, err error) {
+	ctx = trace.StartSpan(ctx, "cloud.google.com/go/datastore.Lookup")
```
tritone, line 54: I think the name should be `cloud.google.com/go/datastore.datastoreClient.Lookup` because it's a method on the client. Similar for other methods in this file. @codyoss do you agree?

datastore/client.go
```diff
@@ -90,13 +106,27 @@ func (dc *datastoreClient) Rollback(ctx context.Context, in *pb.RollbackRequest,
...
+	defer func() { trace.EndSpan(ctx, err) }()
+
 	err = dc.invoke(ctx, func(ctx context.Context) error {
 		res, err = dc.c.AllocateIds(ctx, in, opts...)
 		return err
 	})
 	return res, err
 }
 
+func (dc *datastoreClient) ReserveIds(ctx context.Context, in *pb.ReserveIdsRequest, opts ...grpc.CallOption) (res *pb.ReserveIdsResponse, err error) {
```
tritone, line 119: I would leave this out for now; it can go in a separate change if the datastore team wants it to be added.

datastore/transaction.go
```diff
@@ -110,17 +110,22 @@ func (c *Client) NewTransaction(ctx context.Context, opts ...TransactionOption)
...
 		req.TransactionOptions = &pb.TransactionOptions{
 			Mode: &pb.TransactionOptions_ReadOnly_{ReadOnly: &pb.TransactionOptions_ReadOnly{}},
 		}
-	} else if s.prevID != nil {
-		req.TransactionOptions = &pb.TransactionOptions{
-			Mode: &pb.TransactionOptions_ReadWrite_{ReadWrite: &pb.TransactionOptions_ReadWrite{
-				PreviousTransaction: s.prevID,
-			}},
+	} else {
+		ctx = trace.StartSpan(ctx, "cloud.google.com/go/datastore.Transaction.ReadWriteTransaction")
```
tritone, line 120: Why is the span not started for readOnly? Can you add a comment to clarify?

AlisskaPie (commented):
I've tested an integration test for `AllocateIds`, and we've passed `addOCStatsHandler` function and had false for `settings.TelemetryDisabled`.

datastore/transaction.go
tritone: I don't really understand why there wouldn't be a span for both transaction types. @benwhitehead can you follow up to confirm?

BenWhitehead: @tritone Is correct, we should cover both cases I missed adding readOnly to the list when I was grepping for stuff apologies, I'll add it to the list in the doc.

AlisskaPie: All right, I added a span for `readOnly` also.

AlisskaPie: Yes, that confused me too a bit

datastore/transaction.go
```diff
@@ -110,17 +110,25 @@ func (c *Client) NewTransaction(ctx context.Context, opts ...TransactionOption)
...
+func (c *Client) newTransaction(ctx context.Context, s *transactionSettings) (_ *Transaction, err error) {
 	req := &pb.BeginTransactionRequest{ProjectId: c.dataset}
 	if s.readOnly {
+		ctx = trace.StartSpan(ctx, "cloud.google.com/go/datastore.Transaction.ReadOnlyTransaction")
+		defer func() { trace.EndSpan(ctx, err) }()
+
 		req.TransactionOptions = &pb.TransactionOptions{
 			Mode: &pb.TransactionOptions_ReadOnly_{ReadOnly: &pb.TransactionOptions_ReadOnly{}},
 		}
-	} else if s.prevID != nil {
```
tritone, line 119: Let's not modify the `else if` to `else`, this changes the actual logic of the function. Actually @BenWhitehead , would it be okay to just have a single "cloud.google.com/go/datastore.Client.newTransaction" for this function? It would prevent the confusing nested changes that we see here.

datastore/transaction.go
```diff
@@ -110,17 +110,25 @@ func (c *Client) NewTransaction(ctx context.Context, opts ...TransactionOption)
 	return c.newTransaction(ctx, newTransactionSettings(opts))
 }
 
-func (c *Client) newTransaction(ctx context.Context, s *transactionSettings) (*Transaction, error) {
+func (c *Client) newTransaction(ctx context.Context, s *transactionSettings) (_ *Transaction, err error) {
```
tritone, line 113: Why is the return signature changed? Seems unnecessary.

datastore/transaction.go
BenWhitehead: We do need to keep all of the existing logic, so we'll need to keep the `else if`.

Due to the fact that read-write transactions can fail due to contention retries are often likely, so it would be useful to see multiple underlying RPCs attempted within the scope of a single application transaction request. (ReadOnly transactions aren't usually considered retryable because some of the parameters are time based and can easily expire across a retry attempt which could show the incorrect error to users).

tritone: Okay, got it. So we can keep separate `ReadWrite` and `ReadOnly` in the two code branches here, but @AlisskaPie please revert the change you made to the logic that changed an `else if` into an `else`.

datastore/transaction.go
AlisskaPie: I used `err` in return signature for this statement `defer func() { trace.EndSpan(ctx, err) }()` like it was done in this [CL](https://code-review.googlesource.com/c/gocloud/+/39590/1..5).
I could revert it too, if it's not okay

tritone: Ahh okay. This looks fine, then.

tritone (approved):

BenWhitehead (approved):

Issue #2751, opened 2020-08-19
pubsub: option.WithTelemetryDisabled ignored


filip-ph-johansson:
**Client**

PubSub

**Environment**

macOS Catalina 10.15.4

**Go Environment**

$ go version
go1.14.2 darwin/amd64

**Code**

```go
c, err := pubsub.NewClient(context.Background(), "GCPProjectID", option.WithTelemetryDisabled())
```

**Expected behavior**

Pubsub shouldn't push telemetry data when `option.WithTelemetryDisabled` is passed.

**Actual behavior**

The following Pubsub traces are pushed to GCP Trace:
Sent.google.pubsub.v1.Subscriber.Acknowledge
Sent.google.pubsub.v1.Subscriber.ModifyAckDeadline
Sent.google.pubsub.v1.Publisher.Publish

PR #2815, opened 2020-09-02
fix(pubsub): honor option.WithTelemetryDisabled

codyoss:
It seem that the telemetry options added here predate general
telemetry handling. There is no need to explicitly enable such
options here as it is taken care of grpc.DialPool:
https://pkg.go.dev/google.golang.org/api/transport/grpc?#DialPool.
The stats handler is only added if option.WithTelemetryDisabled is
not passed in.

Fixes: #2751


codyoss:
[Here](https://github.com/googleapis/google-api-go-client/blob/v0.31.0/transport/grpc/dial.go#L173) is where the option is added in the private dial function.

hongalex (approved):
LGTM, thanks

PR #2817, opened 2020-09-03
docs(bigtable): fixed a spelling error

AlisskaPie:


codyoss (approved):

bigtable/bttest/inmem.go
```diff
@@ -818,10 +818,10 @@ func (s *server) CheckAndMutateRow(ctx context.Context, req *btpb.CheckAndMutate
 	// Figure out which mutation to apply.
 	whichMut := false
 	if req.PredicateFilter == nil {
-		// Use true_mutations iff row contains any cells.
```
tritone, line 821: I think this is not an error-- it's short for "if and only if https://en.wikipedia.org/wiki/If_and_only_if"

PR #2818, opened 2020-09-03
test(storage): re-enable full CustomTime test

tritone:
Internal bug blocking this has been fixed and I verified
that the test now passes locally. Rollout will be at
100% tomorrow.


codyoss (approved):
LGTM

PR #2837, opened 2020-09-10
feat(spanner): add KeySetFromKeys function

### LINKED ISSUE #2798: spanner: make keysets usable
I would like to create a keyset from a slice of rows for deletion.  I have tried two different code paths:

```
func buildKeySet(input []struct{}) (out KeySet) {
  for _, i := range input {
    out = spanner.KeySets(out, spanner.Key{i.value1, i.value2})
  }
  return 
}
```
this results in the first value of the keyset being nil, causing a panic.
```
func buildKeySet(input []struct{}) (out KeySet) {
  keys []spanner.Key
  for _, i := range input {
    keys = append(keys, spanner.Key{i.value1, i.value2})
  }
  return spanner.KeySets(keys...)
}
```
this doesn't compile, as `keys` is a slice of `spanner.Key`, not a `spanner.KeySet`.  

modify the `KeySets` function to discard nil values, making it similar to the build-in `append(slice...)` function.

olavloite:
Allows the creation of a KeySet from a slice of Keys.

Fixes #2798


hengfengli (approved):
LGTM.

PR #2839, opened 2020-09-10
refactor(logging): Debranding of Stackdriver

### LINKED ISSUE #3007: logging: Debrand Cloud Logging library from Stackdriver
Requested by product team, Stackdriver Logging was renamed to Cloud Logging and we shouldn't use "Stackdriver" as a brand name anymore. Need to go through both manually generated and auto generated files to ensure we deprecate "stackdriver"

Similar effort to: https://github.com/googleapis/java-logging/pull/258 for java

simonz130:
Fixes #3007 
Stackdriver Logging is now called Cloud Logging, so we are debranding across all Cloud Logging repos.

Here is the update for the readme of logging client library in Go.


nicoleczhu:
> It seems like there are some mentions in the manual layers docs as well, should those be updated? https://pkg.go.dev/cloud.google.com/go/logging
@codyoss I updated those as well. This PR is ready for review now.

It seems that there are no autogenerated references to "Stackdriver". But please do let me know if I am accidentally hardcoding something that should be changed upstream.

codyoss (approved):
LGTM, thanks

PR #2844, opened 2020-09-11
doc(storage): clarify ObjectIterator error case

tritone:
If ObjectIterator.Next returns any error, subsequent calls
will continue to return the error. See
https://github.com/googleapis/google-api-go-client/blob/master/iterator/iterator.go#L106

This was a source of customer confusion recently, so I
added some clarity to the docs, as well as a suggestion
for how to address errors in a long-running iteration.


storage/bucket.go
```diff
@@ -1170,6 +1171,12 @@ func (it *ObjectIterator) PageInfo() *iterator.PageInfo { return it.pageInfo }
 // there are no more results. Once Next returns iterator.Done, all subsequent
 // calls will return iterator.Done.
 //
+// In addition, if Next returns an error other than iterator.Done, all
+// subsequent calls will return the same error. To continue iteration, a new
+// `ObjectIterator` must be created. Since objects are ordered lexicographically
+// by name, `Query.StartOffset` can be used to create a new iterator which will
+// start at the desired place.
```
frankyn, line 1178: Suggestion: point to an example in godocs that shows folks how to start from a specified offset.

frankyn (changes requested):

storage/bucket.go
tritone: Done; had to add example to godoc.

frankyn (approved):
LGTM! Thanks @tritone

PR #2854, opened 2020-09-14
feat(godocfx): create Go DocFX YAML generator

tbpg:
There is a some code copied from https://github.com/golang/pkgsite for getting type/function synopses. Because of those license headers, I put the code in `third_party`.

I included a script we can hopefully use from a Kokoro job. That shows how the generator is intended to be used.

Eventually, we may publish this as a standalone tool. We may end up wanting to publish docs for other modules (outside google-cloud-go). It's also possible others in the community would be interested.

cc @dmitshur @julieqiu @JustinBeckwith @spf13 @SurferJeffAtGoogle @jskeet


third_party/internal/godocfx/main.go
```diff
@@ -0,0 +1,138 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+func main() {
+	print := flag.Bool("print", false, "Print instead of save (default false)")
+	rm := flag.Bool("rm", false, "Delete out directory before generating")
+	outDir := flag.String("out", "obj/api", "Output directory (default obj/api)")
+	flag.Parse()
+	if flag.NArg() != 1 {
+		log.Fatalf("%s expected 1 arg, got %d", os.Args[0], flag.NArg())
```
noahdietz, line 55: Nit: could this error message be improved? `%s missing required argument: module path`

third_party/internal/godocfx/parse.go
```diff
@@ -0,0 +1,297 @@
...
+var onlyGo = []string{"go"}
+
+// parse parses the directory into a map of import path -> page and a TOC.
+//
+// glob is the path to parse, usually ending in `...`. glob is passed directly
+// to packages.Load as-is.
+func parse(glob string) (map[string]*page, tableOfContents, *packages.Module, error) {
+	config := &packages.Config{
+		Mode:  packages.NeedName | packages.NeedSyntax | packages.NeedTypes | packages.NeedTypesInfo | packages.NeedModule,
+		Tests: true,
```
codyoss, line 88: Why tests?

third_party/internal/godocfx/parse.go
```diff
@@ -0,0 +1,297 @@
...
+	packages.PrintErrors(pkgs) // Don't fail everything because of one package.
+
+	if len(pkgs) == 0 {
+		return nil, nil, nil, fmt.Errorf("pattern %q matched 0 packages", glob)
+	}
+
+	pages := map[string]*page{}
+	toc := tableOfContents{}
+
+	module := pkgs[0].Module
```
codyoss, line 104: Will at least one pkg always exist?

third_party/internal/godocfx/parse.go
```diff
@@ -0,0 +1,297 @@
...
+			Summary: docPkg.Doc,
+			Langs:   onlyGo,
+			Type:    "package",
+		}
+		pkgPage := &page{Items: []*item{pkgItem}}
+		pages[pkg.PkgPath] = pkgPage
+
+		for _, c := range docPkg.Consts {
+			name := strings.Join(c.Names, ", ")
+			id := strings.Join(c.Names, ",")
```
codyoss, line 158: Do we need another variable for the same values as name? Looks like name/id are the same for a few of these items.

internal/godocfx/go.mod
```diff
@@ -0,0 +1,12 @@
+module cloud.google.com/go/third_party/internal/godocfx
```
codyoss, line 1:
```suggestion
module cloud.google.com/go/internal/godocfx
```

tbpg (commented):
Thanks! Please take another look.

dmitshur:
FYI, there is a tracking issue golang/go#39883 for making that API available to for external use; mentioning it here so that this use case can be considered as part of that work. You may also want to leave a 👍  reaction and/or subscribe to it.

internal/godocfx/go.mod
tbpg: Done.

tbpg:
Thanks! I gave a :+1: to golang/go#39883.

third_party/internal/godocfx/parse.go
codyoss: Gotcha.

codyoss (commented):
LGTM

codyoss (approved):

PR #2863, opened 2020-09-15
feat(pubsub): add flush method to topic

### LINKED ISSUE #1831: pubsub: add flush feature to topic
Cannot ensure the sending of pubsub message without effort

A method in the pubsub topic object ensuring all message have been sent

Implement by myself with more or less complex mechanism (go deep in the underlying layer)

**Additional context**
I propose to add in pubsub/topic.go file the following code:
```go
// Flush ensures all remaining messages are sent
func (t *Topic) Flush() {
	t.bundler.Flush()
}
```

hongalex:
To complement the `FlushAndStop` method, add a `Flush` method that flushes all publishes topics but does not stop the topic.

Closes #1831


pubsub/topic.go
```diff
@@ -474,6 +474,14 @@ func (t *Topic) Stop() {
 	t.scheduler.FlushAndStop()
 }
 
+// Flush ensures all remaining messages are sent.
```
codyoss, line 477: Maybe this is implied, but should we explicitly say that this method is blocking?

pubsub/topic.go
```diff
@@ -474,6 +474,14 @@ func (t *Topic) Stop() {
...
+func (t *Topic) Flush() {
```
codyoss, line 478: This could also take a bool to stop perhaps? Then just recommend calling t.Flush(true) instead of t.Stop(). This is fine though as there is already a means to stop and flush.

codyoss (approved):
Just a couple of comments, but LGTM.

pubsub/topic.go
hongalex: I prefer to leave the methods separate as it's a bit more explicit and potentially less confusing. However, if there's precedent to take a bool here in other functions, I'd be happy to add it in.

jaeyeom9710:
@hongalex To clarify, will `Flush()` immediately flush and send out all messages buffered by the API, and override any batch configurations like batch interval or batch sizes?

PR #2865, opened 2020-09-15
chore: update dataproc CODEOWNERS

stephaniewang526:


.github/CODEOWNERS
```diff
@@ -8,3 +8,4 @@
 /pubsub/            @googleapis/api-pubsub
 /spanner/           @skuruppu
 /storage/           @tritone
+/dataproc/          @googleapis/yoshi-go
```
codyoss, line 11: Do you have a more targeted group we can use for this? Or a few github users? This is a large-ish group that will get tagged.

stephaniewang526: yeah I understand. We're going through an exercise to specify CODEWONERS for Data Analytics PAs so that we stop relying on individuals for code reviews in case they leave. Justin is against creating too many api-team aliases unless it is very necessary. As a result, I thought this could be an okay fall back for dataproc.

codyoss: I totally get now wanting to to tag individual people. I wonder if we should create a team for this one though. The way our nightly regen PRs work is that they will tag all of the codeowners of every client, every day. All clients receive a one line change today daily, for example https://github.com/googleapis/google-cloud-go/pull/2864 has a one line dataproc change. So this group would be tagged daily.

To be clear, I don't think we should have a api-dataproc team, but rather an analytics team.

stephaniewang526: okay could we use `yoshi-go-admins`?

codyoss: yoshi-go-admins does own it today, although not as that alias. See [line two](https://github.com/googleapis/google-cloud-go/blob/d891ce4894460c9fdc007a4f334e864055085976/.github/CODEOWNERS#L2) of the file. Any directory that does not have a have a specific owners falls back to the `*` owners.

stephaniewang526: yes but we want to rely on a team alias instead of ICs

codyoss: Instead of setting this just for dataproc, could we just switch the * instead? That way that setting can be applied to all unowned code.

stephaniewang526: Sure. Done.

codyoss (approved):

.github/CODEOWNERS
codyoss: Thanks!

PR #2624, opened 2020-07-21
fix(pubsub): use grpc conn pool option for subscriber client

hongalex:
Back in February 2020, `google.golang.org/api/transport/grpc` switched to using a single grpc connection for `Dial` starting in [v0.20.0](https://github.com/googleapis/google-api-go-client/blob/v0.20.0/transport/grpc/dial.go#L46). While the publisher client started to use `DialPool`, the subscriber client did not switch to connection pooling. This was causing an issue where number of open streams was being capped at about 100.

This change has the additional effect of separating the previously shared connection between publisher and subscriber, which should have a minimal effect on most uers.

Fixes #2593

PR #2867, opened 2020-09-15
fix(pubsub): check both client errors on Close

hongalex:
In #2624, we split the publisher and subscriber client's connection pool. Thus, we need to check both errors returned, rather than just the first one.


shollyman (approved):
Approving, but please see the comment about close semantics and address if necessary.

codyoss (approved):

pubsub/pubsub.go
```diff
@@ -78,14 +78,11 @@ func NewClient(ctx context.Context, projectID string, opts ...option.ClientOptio
...
 	pubc, err := vkit.NewPublisherClient(ctx, o...)
 	if err != nil {
-		return nil, fmt.Errorf("pubsub: %v", err)
+		return nil, fmt.Errorf("pubsub(publisher): %v", err)
 	}
 	subc, err := vkit.NewSubscriberClient(ctx, o...)
 	if err != nil {
-		// Should never happen, since we are passing in the connection.
-		// If it does, we cannot close, because the user may have passed in their
-		// own connection originally.
```
tbpg, line 87: Is this comment no longer true?

pubsub/streaming_pull_test.go
```diff
@@ -315,10 +315,12 @@ func TestStreamingPull_ClosedClient(t *testing.T) {
...
 	time.Sleep(100 * time.Millisecond)
 
-	err := client.Close()
-	if err != nil {
-		t.Fatal(err)
-	}
+	// Intentionally don't check the returned err here. In the fake,
+	// closing either the publisher/subscriber client will cause the
+	// server to clean up resources, which is different than in the
+	// live service. With the fake, client.Close() will return an
```
tbpg, line 321: Nit: s/an/a/

pubsub/pubsub.go
```diff
@@ -101,10 +98,13 @@ func NewClient(ctx context.Context, projectID string, opts ...option.ClientOptio
 // If the client is available for the lifetime of the program, then Close need not be
 // called at exit.
 func (c *Client) Close() error {
-	// Return the first error, because the first call closes the connection.
-	err := c.pubc.Close()
-	_ = c.subc.Close()
-	return err
+	if err := c.pubc.Close(); err != nil {
```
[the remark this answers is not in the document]
hongalex: Agreed. Restructured to resolve this.

pubsub/pubsub.go
hongalex: Correct. Before, the publisher connection was passed in to the subscriber connection, but this was separated in #2624.

xmlking:
@hongalex I was debugging why I am getting `"pubsub publisher closing error: rpc error: code = Canceled desc = grpc: the client connection is closing`  error when called `client.Close()` 
After upgrading to `cloud.google.com/go/pubsub v1.8.0`  I started see this error. I never got this error in previous pubsub versions. 

I accidentally discovered there is a type in the error message at:  
https://github.com/googleapis/google-cloud-go/blob/master/pubsub/pubsub.go#L100-L109

this should be:
```go
func (c *Client) Close() error {
	pubErr := c.pubc.Close()
	subErr := c.subc.Close()
	if pubErr != nil {
                 return fmt.Errorf("pubsub publisher closing error: %v", pubErr)
	}
	if subErr != nil {
		return fmt.Errorf("pubsub subscriber closing error: %v", subErr)
	}
	return nil
}
```

xmlking:
Should we detect `PubSub emulator`  environment and automatically  suppress  `the client connection is closing` error ?
otherwise this leads to confusion  to users

hongalex:
Hey @xmlking, this error does occur with the emulator and the `pstest` fake server. I'll address this in a separate PR so we can suppress "the client connection is closing" errors. Also, thanks for catching the typo.

Btw, I noticed in the linked issue, [there's a comment](https://github.com/xmlking/toolkit/blob/804d63dab872317336ef2a2f6f2cca21e2dba6cb/broker/pubsub/default.go#L108) that says 
> Live pubsub server will throw this error

Though I think the opposite is true (live pub/sub service will *not* throw this error).

xmlking:
thanks @hongalex . fixed my type

PR #2769, opened 2020-08-24
fix(storage): add content type, disposition and encoding in v4 storage post policy

pierreis:
This merge request adds content type, disposition and encoding in storage post policy. These conditions were present in the `PolicyV4Fields`, but ignored.

Fixes #2767.

PR #2873, opened 2020-09-16
test(storage): copy latest conformance test fixes

tritone:
Pick up fixes to conformance tests that cover encoding edge
cases and condition ordering for V4 signature and post policy.
This will unblock #2769.


codyoss (approved):
LGTM, but I am not too familiar with these tests and how they work.

frankyn (approved):

PR #2878, opened 2020-09-17
fix(datastore): increase deferred key iter limit

tritone:
A customer was hitting the upper bound for iteration through
deferred keys, and the logic underlying the 100 iteration limit
was flawed. Based on internal discussions and what's happening
in other clients, it's fine to increase this value.


tritone:
cc @benwhitehead

BenWhitehead:
Thanks for looking into this @tritone!

PR #2880, opened 2020-09-17
fix(godocfx): shorten function names

tbpg:
After this change, function headers show up as `func Foo` rather than
`func Foo(s string) string`, matching the behavior of pkg.go.dev.


codyoss (approved):
LGTM, it may be nice in the future to have "golden file" and a sample module to test this against to show changes like this from a testing standpoint.

PR #2881, opened 2020-09-17
test(godocfx): add golden test

tbpg:
I did some minor refactoring of the write logic to take it out of `func main` and make it testable.


internal/godocfx/main.go
```diff
@@ -76,8 +77,13 @@ func main() {
 	if *rm {
 		os.RemoveAll(*outDir)
 	}
-	if err := os.MkdirAll(*outDir, os.ModePerm); err != nil {
-		log.Fatalf("os.MkdirAll: %v", err)
+
+	write(*outDir, pages, toc, module)
```
codyoss, line 81: Error should be logged.

codyoss (approved):
One small thing, other than that LGTM.

internal/godocfx/main.go
tbpg: Done.

tbpg:
Thanks!

PR #2884, opened 2020-09-17
feat(docfx): add support for examples

tbpg:
Examples are located in _test.go files, which go/packages parses into a
separate foo_test package. So, this requires an extra step of collecting
all files from the package, including test packages, then parsing them.
That way, the go/doc package can associate types/functions/packages with
the examples for them (rather than an unassociated list).


codyoss (approved):
LGTM, nice!

PR #2885, opened 2020-09-17
docs(spanner): document NULL value

### LINKED ISSUE #2879: spanner: guarantee and document data field of NullFoo structs when Valid is false
The spanner client library has a bunch of structs (`NullBool`, `NullDate`, etc.) that have two fields, one for the data (`Bool`/`Date`/etc.) and one for whether the value is not NULL (`Valid`). However, it doesn't specify what value the data field will have if the value is NULL, which makes for some tedious code that is comfortable treating NULLs as the type's zero value.

```go
var x spanner.NullString
// ... read from DB
var str string
if x.Valid { // otherwise x.StringVal could be anything
  str = x.StringVal
}
// ... use str
```

It'd be nicer if the guaranteed and documented behaviour is that the zero value will be set for NULLs:

```go
var x spanner.NullString
// ... read from DB
// ... use x.StringVal, which will be the empty string for a NULL
```

I believe the implementation that extracts DB responses (`decodeValue` in `spanner/value.go`) explicitly sets a receiving NullFoo to its zero value (and thus its constituent fields) when the DB returns a NULL, so the implementation already has the desired behaviour. This bug asks for that to be committed to, and to be documented.

olavloite:
The contents of the value fields of the Null* structs when Valid=false was not documented, although they would always be
the zero value of the type that they contained. This behavior is now documented and also asserted by the tests in value_test.go, including when the input is initialized to a non-zero value.

Fixes #2879


dsymonds (approved):
Great!

PR #2889, opened 2020-09-18
chore(spanner/spansql): use ID type for identifiers throughout

dsymonds:
This was a long-planned cleanup, which adds some rigour to the type
definitions, and avoids a bunch of unnecessary casting in various
places.

### DIFF
--- a/spanner/spannertest/inmem.go
+++ b/spanner/spannertest/inmem.go
@@ -565,10 +565,10 @@ func (s *server) StreamingRead(req *spannerpb.ReadRequest, stream spannerpb.Span
 	var ri rowIter
 	if req.KeySet.All {
 		s.logf("Reading all from %s (cols: %v)", req.Table, req.Columns)
-		ri, err = s.db.ReadAll(req.Table, req.Columns, req.Limit)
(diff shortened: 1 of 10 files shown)


spanner/spannertest/inmem.go
```diff
@@ -593,7 +593,7 @@ func (s *server) readStream(ctx context.Context, tx *transaction, send func(*spa
 			return err
 		}
 		rsm.RowType.Fields = append(rsm.RowType.Fields, &spannerpb.StructType_Field{
-			Name: ci.Name,
+			Name: string(ci.Name), // TODO: should this be quoting via ci.Name.SQL()?
```
olavloite, line 596: I don't think this should be quoted. Row type metadata is returned from the server without quoting, regardless whether the identifier needs quoting when used in SQL or not.

olavloite (approved):
LGTM

spanner/spannertest/inmem.go
dsymonds: Yeah, you're right. This is also potentially not a plain identifier anyway (there's a TODO to do something nicer if the result column is an expression, for instance), but there's no need to quote here.

PR #2890, opened 2020-09-18
docs: publish docs during releases

tbpg:
Improvements very welcome for this process. Not sure of a better way to make
sure the right docs get published.


codyoss (approved):
Thanks for adding instructions!

RELEASING.md
```diff
@@ -85,8 +85,14 @@ the failures have been resolved.
    c. Tag the repo with the next version: `git tag $NV`.
    d. Push the tag to origin:
       `git push origin $NV`
-2. Update [the releases page](https://github.com/googleapis/google-cloud-go/releases)
+1. Update [the releases page](https://github.com/googleapis/google-cloud-go/releases)
    with the new release, copying the contents of `CHANGES.md`.
+1. (Internal only.) Wait for the new release to be available on pkg.go.dev, then
```
tritone, line 90: I think the internal part is assumed for releases. There are multiple other steps you can't accomplish without being internal as well.

Also, how long does it take for a new tag to get picked up by pkg.go.dev? Might be useful to add.

tbpg: Removed and added. :)

tbpg (commented):
Thanks!

tritone (approved):

Issue #2883, opened 2020-09-17
all: example suffixes should start with a lower case letter


tbpg:
GoDoc example suffixes should start with a lower case letter.

```
// ... The suffix may be
// absent. Otherwise, it must start with a lower-case letter and
// be preceded by '_'.
```

https://golang.org/src/go/doc/example.go

codyoss:
This is an interesting limitation... Looking around the stdlib I see a lot of contrary examples:

- [bytes](https://github.com/golang/go/blob/6796a7fb127676b61375339076ae1c982a721dde/src/bytes/example_test.go#L33)
- [database/sql](https://github.com/golang/go/blob/6796a7fb127676b61375339076ae1c982a721dde/src/database/sql/example_test.go#L21)
- [sync/atomic](https://github.com/golang/go/blob/6796a7fb127676b61375339076ae1c982a721dde/src/sync/atomic/example_test.go#L50)

To be fair I do see both patterns being used. Both seem to render correctly in godoc...

tbpg:
Better docs:

```
The naming convention to declare examples for the package, a function F, a type T and method M on type T are:

func Example() { ... }
func ExampleF() { ... }
func ExampleT() { ... }
func ExampleT_M() { ... }

Multiple example functions for a package/type/function/method may be provided by appending a distinct suffix to the name. The suffix must start with a lower-case letter.

func Example_suffix() { ... }
func ExampleF_suffix() { ... }
func ExampleT_suffix() { ... }
func ExampleT_M_suffix() { ... }
```

https://golang.org/pkg/testing/

Methods make it more difficult to find these with something like `grep`.

codyoss:
Ah, icic. Thanks for the explanation.

PR #2892, opened 2020-09-18
fix(translate): properly name examples

codyoss:
Fixes: #2883


codyoss:
These were the only offenders I found, did you have a list somewhere @tbpg?

tbpg (approved):
Thank you! I took a look through `grep -RE 'func Example[a-zA-Z]*_[A-Z]'` and didn't notice anything in particular...

PR #2895, opened 2020-09-18
fix(storage): encode + in SignedURL query params

### LINKED ISSUE #2876: storage: V4 Signed URL encoding of spaces
We updated the conformance tests in googleapis/conformance-tests#39 to take into account a reported issue with the correct encoding of spaces in query parameters (googleapis/google-cloud-ruby#7655). This issue needs to be fixed in the Go client as well. I'm going to put in a skip for this test case until the issue has been resolved.

tritone:
This must be encoded for the signature to be correct if a
query parameter contains a space. We have a conformance
test case for this which can now be unskipped.

Fixes #2876


storage/storage.go
```diff
@@ -587,8 +587,8 @@ func signedURLV4(bucket, name string, opts *SignedURLOptions, now time.Time) (st
 	for k, v := range opts.QueryParameters {
 		canonicalQueryString[k] = append(canonicalQueryString[k], v...)
 	}
-
-	fmt.Fprintf(buf, "%s\n", canonicalQueryString.Encode())
+	escapedQuery := strings.Replace(canonicalQueryString.Encode(), "+", "%20", -1)
```
tbpg, line 590: Will you add a comment explaining this? I would expect `Encode` to handle everything.

tbpg (changes requested):

storage/storage.go
codyoss: If the value is encoded, then replaced there will be no `+` left to replace right? `+` should be encoded to `%2B`.

tritone: The default behavior is that `+` is encoded to `%2B` (which is good and expected) but a space is encoded to `+` (while we want `%20` by the spec for this feature). See https://github.com/golang/go/blob/master/src/net/url/url.go#L314 . I'll add a comment here to clarify.

codyoss: Ah, gotcha!

codyoss (approved):

frankyn (approved):

tbpg (approved):

Issue #2850, opened 2020-09-12
spansql: support join


uschen:
**Is your feature request related to a problem? Please describe.**
trying to use spansql to build query in project instead of handwriting them. however, it does not support JOIN

**Describe the solution you'd like**
support JOIN and ideally complete set of sql spanner supports

**Describe alternatives you've considered**
seems there are quite some go sql builder/parser around, it shouldn't be difficult to implement(?)
* https://github.com/google/zetasql
* https://github.com/Masterminds/squirrel
* https://github.com/cockroachdb/cockroach/tree/master/pkg/sql

**Additional context**
N/A

dsymonds:
I believe the parsing for joins is (mostly) complete. Implementing evaluation of them in spannertest is tracked in #2462.

PR #2896, opened 2020-09-19
feat(spanner/spansql): parse joins

dsymonds:
This parses almost all of the JOIN clause syntax, with the exception of
HASH and hints.

It doesn't support the full range of common joins due to a lack of
support for path expressions, but that will be addressed separately.

Updates #2850.

### DIFF
--- a/spanner/spansql/parser.go
+++ b/spanner/spansql/parser.go
@@ -1779,7 +1779,31 @@ func (p *parser) parseSelectList() ([]Expr, []ID, *parseError) {
 }
 
 func (p *parser) parseSelectFrom() (SelectFrom, *parseError) {
-	// TODO: support more than a single table name.
+	debugf("parseSelectFrom: %v", p)
+
+	/*
+		from_item: {
+			table_name [ table_hint_expr ] [ [ AS ] alias ] |
+			join |
+			( query_expr ) [ table_hint_expr ] [ [ AS ] alias ] |
+			field_path |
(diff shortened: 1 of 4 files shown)


spanner/spansql/parser.go
```diff
@@ -1779,7 +1779,30 @@ func (p *parser) parseSelectList() ([]Expr, []ID, *parseError) {
...
+			from_item [ join_type ] [ join_method ] JOIN  [ join_hint_expr ] from_item
+				[ ON bool_expression | USING ( join_column [, ...] ) ]
+
+		join_type:
+			{ INNER | CROSS | FULL [OUTER] | LEFT [OUTER] | RIGHT [OUTER] }
+	*/
+
+	// A join starts with a from_item, so that can't be detected in advance.
+	// TODO: Support more than table name or join.
+
```
olavloite, line 1805:
```suggestion
	// TODO: Support more than one join.
```

olavloite (approved):
LGTM.

As far as I can see, this currently only supports one JOIN per query. If so, it might be good to add a small TODO for that as well so that anyone looking into the code knows it's a known limitation.

spanner/spansql/parser.go
dsymonds: I believe it works (it parses `SELECT * FROM A INNER JOIN B FULL JOIN C USING (x)`), though the associativity might not be correct. I'll add a TODO to clarify.

PR #2909, opened 2020-09-22
test(spanner/spansql): add fuzz target for ParseQuery

codyoss:
This target will be called from a runner managed by the oss-fuzz
project. This target is an experimental test integration with the
oss-fuzz project.


dsymonds (approved):
LGTM

Issue #2463, opened 2020-06-16
spannertest: support for aliases in SELECT statements


jrmeinride:
**Is your feature request related to a problem? Please describe.**
To make our queries shorter and easier to read we use aliases in our queries. We want to unit test code that has such queries. 

Example:

```
SELECT t1.* 
FROM table1 AS t1
	LEFT OUTER HASH JOIN table2 AS t2 
		ON t1.id = t2.table1_id
```

Granted, at this point there is no support for `JOIN` in `SELECT` statements, but we are hoping that this will be added, so that we can properly unit test our code. 

**Describe the solution you'd like**
We would like to se aliases in `SELECT` statements.

**Describe alternatives you've considered**
The only viable option would be to spin up a new table in one of our projects, but that is out of scope for unit tests.

**Additional context**
Ẃe have another feature request for JOIN in SELECT statements, that is by far a more important feature. This would be a further quality of life improvement, we can write queries without aliases, though it would be nice to have them.

dsymonds:
There's already support for aliases in SELECT expression lists and FROM clauses, I believe. Do you have an example query not using JOIN that has an alias that doesn't work/

jrmeinride:
> There's already support for aliases in SELECT expression lists and FROM clauses, I believe. Do you have an example query not using JOIN that has an alias that doesn't work?

Using the 1.6.0 version of spanner we get this error message when I run a query:

SQL for table:
```
CREATE TABLE something (
    id STRING(100) NOT NULL,
    created_time TIMESTAMP OPTIONS (allow_commit_timestamp=true)
) PRIMARY KEY(id)
```


Test code:
```
func Test_Spanner(t *testing.T) {
	fx := newFixture(t)
	defer fx.Fixture.Teardown()
	stmt := spanner.Statement{
		SQL: `SELECT s.* 
		FROM something as s
		WHERE s.id = @some_id 
		ORDER BY s.created_time ASC
		LIMIT 1`,
		Params: map[string]interface{}{
			"some_id": "1",
		},
	}
	rowIterator := fx.SpannerClient.Single().Query(context.Background(), stmt)
	defer rowIterator.Stop()
	_, err := rowIterator.Next()
	if err != nil {
		if err == iterator.Done {
			t.Log("done")
		} else {
			t.Log(err)
			t.FailNow()
		}
	}
}
```
[...]

Issue #2462, opened 2020-06-16
spannertest: Support for (HASH) JOIN in select statement


jrmeinride:
**Is your feature request related to a problem? Please describe.**
We have some functionality that uses more than one table in spanner. By using interleave the service is still fast and responsive. When adding unit tests we cannot test certain parts of our code, since there is no support for JOIN, or HASH JOIN in spannertest.

The type of query we would like to be able to run is something akin to this:

```
SELECT * 
FROM table1
LEFT OUTER HASH JOIN table2
	ON table1.id = table2.table1_id
```

**Describe the solution you'd like**
We would like spannertest to support `LEFT OUTER HASH JOIN` in `SELECT` statements.

**Describe alternatives you've considered**
The only viable option would be to set up a table for testing in one of our projects, but that can not be considered to be a unit test.

**Additional context**
-

skuruppu:
@jrmeinride another alternative could be to setup the unit tests against the [Cloud Spanner Emulator](https://cloud.google.com/spanner/docs/emulator).

In any case, I will assign this to @dsymonds to see if he's got the cycles to implement this support.

jrmeinride:
> @jrmeinride another alternative could be to setup the unit tests against the [Cloud Spanner Emulator](https://cloud.google.com/spanner/docs/emulator).
> 
> In any case, I will assign this to @dsymonds to see if he's got the cycles to implement this support.

Thank you! I will look into that.

dsymonds:
Starting to do some preliminary work on this in #2850.

PR #2924, opened 2020-09-26
feat(spanner/spansql): parse path expressions

dsymonds:
These are not properly documented, so this only attempts to support a
limited but widely used subset, which will be sufficient for more fully
supporting aliases and joins.

Updates #2462.
Updates #2463.
Updates #2850.

### DIFF
--- a/spanner/spansql/parser.go
+++ b/spanner/spansql/parser.go
@@ -197,6 +197,7 @@ const (
 	float64Token
 	stringToken
 	bytesToken
+	unquotedID
 	quotedID
 )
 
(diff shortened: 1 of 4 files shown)


spanner/spansql/parser.go
```diff
@@ -752,6 +753,22 @@ func (p *parser) skipSpace() bool {
 
 // advance moves the parser to the next token, which will be available in p.cur.
 func (p *parser) advance() {
+	// If the previous token was an identifier (quoted or unquoted),
+	// the next token being a dot means this is a path expression (not a number).
+	if p.cur.typ == quotedID || p.cur.typ == unquotedID {
+		// TODO: Is space around the dot in a path expression valid?
```
olavloite, line 759: Yes, whitespace characters (including linefeeds and tabs) are allowed both before and after the dot in a path expression.

olavloite (approved):
LGTM

spanner/spansql/parser.go
dsymonds: Okay, easy enough to implement, and made this a little simpler.

PR #2931, opened 2020-09-29
feat(spanner/spannertest): evaluate path expressions

### LINKED ISSUE #2463: spannertest: support for aliases in SELECT statements

dsymonds:
This adds support for aliases in SELECT statements, and in particular
table aliases that can then be referenced from other places in the query
evaluation.

Fixes #2463.

### DIFF
--- a/spanner/spannertest/db_eval.go
+++ b/spanner/spannertest/db_eval.go
@@ -354,6 +354,8 @@ func (ec evalContext) evalExpr(e spansql.Expr) (interface{}, error) {
 		return nil, fmt.Errorf("TODO: evalExpr(%s %T)", e.SQL(), e)
 	case coercedValue:
 		return e.val, nil
+	case spansql.PathExp:
+		return ec.evalPathExp(e)
 	case spansql.ID:
 		return ec.evalID(e)
 	case spansql.Param:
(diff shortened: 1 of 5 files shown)


spanner/spannertest/db_eval.go
```diff
@@ -449,11 +451,36 @@ func (ec evalContext) evalExpr(e spansql.Expr) (interface{}, error) {
 	}
 }
 
-func (ec evalContext) evalID(id spansql.ID) (interface{}, error) {
-	for i, col := range ec.cols {
-		if col.Name == id {
-			return ec.row.copyDataElem(i), nil
+// resolveColumn turns an ID or PathExp into a table column.
+func (ec evalContext) resolveColumn(e spansql.Expr) (int, error) {
```
olavloite, line 455: nit: wouldn't `resolveColumnIndex` be a better function name now?

olavloite (approved):

spanner/spannertest/db_eval.go
dsymonds: Yeah, fair enough. Done.

Issue #2914, opened 2020-09-24
bigtable: support partial results in Clusters() and ListAppProfiles()


IlyaFaer:
While working on #1950, I've noticed that methods `InstanceAdminClient.ListAppProfiles` and `InstanceAdminClient.Clusters` are actually supporting partial results with `failedLocations` as well (see [clusters.list](https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/projects.instances.clusters/list) and [appProfiles.list](https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/projects.instances.appProfiles/list)).

We could update these methods in the same way I've updated `InstanceAdminClient.Instances` in #1950 to support partial results.

PR #2932, opened 2020-09-29
feat(bigtable): support partial results in InstanceAdminClient.Cluste…

IlyaFaer:
…rs()

Towards #2914


ks1707 (approved):

tritone:
@kolea2 can you confirm that this makes sense to add from a product perspective? If so I will give it a review.

bigtable/admin.go
```diff
@@ -1034,6 +1036,11 @@ func (iac *InstanceAdminClient) Clusters(ctx context.Context, instanceID string)
 			StorageType: storageTypeFromProto(c.DefaultStorageType),
 		})
 	}
+	if len(res.FailedLocations) > 0 {
```
kolea2, line 1039: can we add a test?

kolea2 (dismissed):

bigtable/admin.go
tritone: +1, this would help.

IlyaFaer: Hm-m, as we're not very able to simulate "failed locations" error on real data (at least I don't know how this can be done), I think a test can be written with kind of a mock for `ListClusters()` method. But the mock is big, as there is an interface implementation restriction, and the interface describes a lot of methods. By the way, I see that emulator doesn't support most part of instance and cluster related methods. Should we implement them? Doesn't seem to be very hard.

bigtable/integration_test.go
```diff
@@ -1711,6 +1715,119 @@ func TestIntegration_AdminSnapshot(t *testing.T) {
...
+	UnavailableLocations []string
+}
+
+func (iacm *instanceAdminClientMock) ListClusters(ctx context.Context, req *btapb.ListClustersRequest, opts ...grpc.CallOption) (*btapb.ListClustersResponse, error) {
+	res := btapb.ListClustersResponse{
+		Clusters:        iacm.Clusters,
+		FailedLocations: iacm.UnavailableLocations,
+	}
+	return &res, nil
+}
```
IlyaFaer, line 1730: This method will be called by `Clusters` which I'm changing with this PR.

bigtable/integration_test.go
```diff
@@ -1711,6 +1715,119 @@ func TestIntegration_AdminSnapshot(t *testing.T) {
...
+	}
+	defer iAdminClient.Close()
+
+	cluster1 := btapb.Cluster{Name: "cluster1"}
+	failedLoc := "euro1"
+
+	iAdminClient.iClient = &instanceAdminClientMock{
+		Clusters:             []*btapb.Cluster{&cluster1},
+		UnavailableLocations: []string{failedLoc},
+	}
```
IlyaFaer, line 1810: Setting mock data for the test.

tritone:
Test looks fine to me. @crwilcox @kolea2 please take a look

crwilcox:
LGTM. Will give @kolea2 some time and merge tomorrow either way.

crwilcox (approved):

kolea2 (commented):
lgtm sorry for delay :) thanks for merging @crwilcox

PR #2936, opened 2020-09-30
feat(spanner/spansql): parse join hints

dsymonds:
This includes both the special case "HASH JOIN" syntax and the "JOIN@{...}" syntax.

Initial work towards #2462.

### DIFF
--- a/spanner/spansql/parser.go
+++ b/spanner/spansql/parser.go
@@ -776,7 +776,7 @@ func (p *parser) advance() {
 	p.cur.typ = unknownToken
 	// TODO: array, struct, date, timestamp literals
 	switch p.s[0] {
-	case ',', ';', '(', ')', '*':
+	case ',', ';', '(', ')', '{', '}', '*':
 		// Single character symbol.
 		p.cur.value, p.s = p.s[:1], p.s[1:]
 		p.offset++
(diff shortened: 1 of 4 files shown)


spanner/spansql/parser.go
```diff
@@ -1861,12 +1869,51 @@ func (p *parser) parseSelectFrom() (SelectFrom, *parseError) {
...
+			k := tok.value
+			if err := p.expect("="); err != nil {
+				return nil, err
+			}
+			tok = p.next()
+			if tok.err != nil {
+				return nil, tok.err
+			}
+			v := tok.value
+			setHint(k, v)
```
olavloite, line 1907: nit: If `HASH JOIN` is already specified in the join clause, it is illegal to also specify it as a join hint as well, so ideally we should check whether both are being set here.

Source: https://cloud.google.com/spanner/docs/query-syntax#join-hints

olavloite (approved):

spanner/spansql/parser.go
dsymonds: Yeah, there's a lot of strictness throughout this parser that isn't done. It might be worth some future work to tighten things up at some stage, but I'm sticking with just working towards supporting valid things and not putting too much effort into rejecting invalid things.

Issue #2937, opened 2020-09-30
BigQuery: upgrade to 1.11.0: cannot read from a non-query job 


pacbeckh:
**Client**

BigQuery

**Environment**

Local machine (OSx), docker container: golang:1.13

**Go Environment**

$ go version
go1.15 darwin/amd64

**Code**

```go
query := client.Query(rawQuery)
query.DisableQueryCache = true

rowIterator, err = query.Read(ctx)
```

**Expected behavior**

Run the query

**Actual behavior**

I get the error `cannot read from a non-query job `

**Cause:**
Started after upgrading from 1.9.0 to 1.11.0

Bug seems to be introduced in https://github.com/googleapis/google-cloud-go/pull/2560

The code will poll the job because the response has `JobComplete:false`

But the minimaljob created in the code does not have a config, therefore the `j.isQuery()` check below fails resulting in an error:
```
	if !j.isQuery() {
		return nil, errors.New("bigquery: cannot read from a non-query job")
	}
```
Link to code: 
https://github.com/googleapis/google-cloud-go/blob/26c4a53b2150dc3cfd2e1bb09840072346039aa5/bigquery/job.go#L284

shollyman:
Thanks for the report.  Taking a look.

shollyman:
Will work on releasing a v1.12.0 to pick this up now

shollyman:
I misspoke on the version, this has been released as [bigquery/v1.11.1](https://github.com/googleapis/google-cloud-go/releases/tag/bigquery%2Fv1.11.1) as there were no other significant changes pending.

pacbeckh:
Thanks for addressing this so quickly!

PR #2940, opened 2020-09-30
fix(bigquery): address issue with directly Read()'ing from a Query

shollyman:
Existing query test cases in integration tests weren't running long
enough to validate we were successful in falling back to the classic
query path.

PR adds a cheap-but-slow test query to exercise slower queries, and
fixes the issue reported in #2937 where we weren't passing enough of
a job object to the classic path to allow probing successfully.

Alternatives considered:
* Add a jobs.get call under the hood to get the job config (adds
  unnecessary latency)
* Remove isQueryJob() test (can consider this in the future, but
  expands the coverage area).

Fixes: #2937


hongalex (approved):
LGTM after addressing conventional commit lint. You just need to push another commit or amend the previous commit message (since the linter checks only the commit message rather than the PR title if there's only 1 commit)

PR #2942, opened 2020-09-30
feat(firestore): add opencensus tracing support 

AlisskaPie:


firestore/transaction.go
```diff
@@ -116,11 +116,13 @@ func (c *Client) RunTransaction(ctx context.Context, f func(context.Context, *Tr
 	// TODO(jba): get backoff time from gRPC trailer metadata? See
 	// extractRetryDelay in https://code.googlesource.com/gocloud/+/master/spanner/retry.go.
 	for i := 0; i < t.maxAttempts; i++ {
+		t.ctx = trace.StartSpan(t.ctx, "cloud.google.com/go/firestore.Client.BeginTransaction")
```
tritone, line 119: Is there a way to add these to the gapic methods for BeginTransaction and Commit perhaps? @benwhitehead

tritone (commented):
Generally looks good to me. @benwhitehead could you take a look?

firestore/transaction.go
BenWhitehead: I'm not sure. We'll likely have to follow up with someone familiar with the golang generator to see what they think. Do you know who that is off the top of your head?

noahdietz: tbh I'm not really sure what is being asked for here 😅  my guess is that you'd like to see the ability to enable tracing at the _GAPIC method_ scope? This seems like something that could benefit others, I'm just not familiar with how it is used/configured/toggled, so I can't give you a real yes/no.

If y'all know of examples in other languages, that would be helpful. We should probably move this offline or to a [GitHub issue](https://github.com/googleapis/gapic-generator-go/issues) to discuss further. Maybe there is something we can do in GAX instead of the generator 🤷 . I think @codyoss knows about the tracing we enable in the repo, if any.

AlisskaPie: @codyoss could you please take a look?

codyoss: I don't think this should be considered a blocker for this PR. I agree seems like a separate issue that we should track if we are interested.

firestore/transaction.go
```diff
@@ -136,11 +138,14 @@ func (c *Client) RunTransaction(ctx context.Context, f func(context.Context, *Tr
 			// Prefer f's returned error to rollback error.
 			return err
 		}
+		t.ctx = trace.StartSpan(t.ctx, "cloud.google.com/go/firestore.Client.Commit")
```
crwilcox, line 141: Is this the correct lable? Seems like this may be tracing (externally) a gapic level call?

Also, I think this exact label is also used at https://github.com/googleapis/google-cloud-go/pull/2942/files#diff-f5e901925d383c72807f3f3336f0d2681b57c12896e1c8a89051191d19c207e9R296. 

This seems non-ideal as it will overload the label?

firestore/client.go
```diff
@@ -286,7 +292,10 @@ func (c *Client) Batch() *WriteBatch {
 }
 
 // commit calls the Commit RPC outside of a transaction.
-func (c *Client) commit(ctx context.Context, ws []*pb.Write) ([]*WriteResult, error) {
+func (c *Client) commit(ctx context.Context, ws []*pb.Write) (_ []*WriteResult, err error) {
+	ctx = trace.StartSpan(ctx, "cloud.google.com/go/firestore.Client.Commit")
```
crwilcox, line 296: do we have a public Commit call or just this one? Commit is capitalized currently in the span name. Should this be `firestore.Client.commit`?

AlisskaPie: We have a [public Commit](https://github.com/googleapis/google-cloud-go/blob/5344d93872074f8d66a0c8d0458f7bcf8bfae7fd/firestore/writebatch.go#L74) for WriteBatch. It contains this private `commit` method. 

But my logic then was that this `commit` method is also contained in the [`commit1`](https://github.com/googleapis/google-cloud-go/blob/5344d93872074f8d66a0c8d0458f7bcf8bfae7fd/firestore/client.go#L312) method. Which is used in 4 other public functions ([Create](https://github.com/googleapis/google-cloud-go/blob/5344d93872074f8d66a0c8d0458f7bcf8bfae7fd/firestore/docref.go#L130), [Set](https://github.com/googleapis/google-cloud-go/blob/5344d93872074f8d66a0c8d0458f7bcf8bfae7fd/firestore/docref.go#L162), [Delete](https://github.com/googleapis/google-cloud-go/blob/5344d93872074f8d66a0c8d0458f7bcf8bfae7fd/firestore/docref.go#L242), [Update](https://github.com/googleapis/google-cloud-go/blob/5344d93872074f8d66a0c8d0458f7bcf8bfae7fd/firestore/docref.go#L667)). But for these functions we don't have a span for this `Commit` rpc call. That's why I added it here.

I used capitalized `Commit` name because of the capitalized Commit RPC call name. If you think it should have lowercase name, then I'll update.

AlisskaPie (commented):
@crwilcox Thank you!

firestore/client.go
crwilcox: I think I would prefer it to match the method name just for transparency. 👍

firestore/transaction.go
crwilcox: I think this path matches the t.c.c.Commit call. Is the other file generated so we have to do this here?

AlisskaPie: Right, the other file is generated so yes.

firestore/client.go
AlisskaPie: Done

AlisskaPie:
image

tritone:
@crwilcox could you take another look at this?

crwilcox (approved):

PR #2945, opened 2020-10-01
feat(spanner/spannertest): support LEFT OUTER JOIN with ON clause

### LINKED ISSUE #2462: spannertest: Support for (HASH) JOIN in select statement

dsymonds:
This lays the groundwork for supporting joins, and this initial subset
is a very common form.

Fixes #2462.

### DIFF
--- a/spanner/spannertest/db_query.go
+++ b/spanner/spannertest/db_query.go
@@ -31,7 +31,7 @@ pulling from a table (FROM tbl), filtering (WHERE expr), re-ordering (ORDER BY e
 or other transformations.
 
 The order of operations among those supported by Cloud Spanner is
-	FROM + JOIN + set ops [TODO: JOIN and set ops]
+	FROM + JOIN + set ops [TODO: set ops]
 	WHERE
 	GROUP BY
 	aggregation
@@ -127,6 +127,11 @@ func (raw *rawIter) Next() (row, error) {
 
 func (raw *rawIter) add(src row, colIndexes []int) {
 	raw.rows = append(raw.rows, src.copyData(colIndexes))
+}
+
+// clone makes a shallow copy.
+func (raw *rawIter) clone() *rawIter {
+	return &rawIter{cols: raw.cols, rows: raw.rows}
 }
 
 func toRawIter(ri rowIter) (*rawIter, error) {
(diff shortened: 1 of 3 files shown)


spanner/spannertest/db_query.go
```diff
@@ -587,6 +585,167 @@ func (d *database) evalSelect(sel spansql.Select, params queryParams) (ri rowIte
...
+		join predicate.
+
+		LEFT indicates that all rows from the left from_item are
+		returned; if a given row from the left from_item does not join
+		to any row in the right from_item, the row will return with
+		NULLs for all columns from the right from_item. Rows from the
+		right from_item that do not join to any row in the left
+		from_item are discarded.
+	*/
+	if ji.lhsRow == nil {
```
olavloite, line 693: nit: This causes the parser to accept anything in the ON condition if the right hand side is empty. In other words:

`SELECT * FROM JoinA AS A LEFT OUTER JOIN JoinB AS B ON whatever = other_whatever`

is accepted if `JoinB` is an empty table.

olavloite (approved):

spanner/spannertest/db_query.go
dsymonds: Isn't that intended? That's the "even if no rows in the right from_item satisfy the join predicate" implies.

Anyway, I'll do some experiments and tweak this when adding some more join types.

olavloite: No, what I mean is that the ON clause can contain any invalid / unknown field names you want (and maybe even any jibberish you want, I didn't really look into that). Which means that the query will work fine as long as the `JoinB` table is empty, and it will cause a parse error once there's any data in it.

dsymonds: Oh, right, I get you now.

Yeah, I think that's along the lines in the other CL of rejecting invalid inputs, which is a lower priority for now.

olavloite: Yeah, that sounds reasonable.

PR #2962, opened 2020-10-05
chore(all): upgrade handful of clients to GA

codyoss:


tbpg (approved):
We should consider adding comments with dates to make sure we don't move to GA too early.

codyoss:
@tbpg That is a good idea as right now I have to turn on file blame to see what should be moved. I will start adding comments with the begin generation PRs about when the client could be considered GA.

PR #2983, opened 2020-10-08
fix(spanner/spannertest): correct handling of tri-state boolean expre…

dsymonds:
…ssion evaluation

NULL BOOLs need to be preserved beyond evalBoolExpr because NULLs sort
differently to FALSE/TRUE, and need special handling when evaluating
various operators. This was the cause of a couple of divergences from
production Spanner.


olavloite (approved):
LGTM

PR #2984, opened 2020-10-08
feat(pubsublite): Basic types for the Pub/Sub Lite client library

tmdiep:
Doc containing proposed interface of the Pub/Sub Lite Go client library: http://go/pubsublite-go-clientlib-interface.

Sending this out for review, but I don't intend to merge this PR until the pubsublite submodule has been created.

If helpful, this branch contains more progress on the pubsublite client library and shows how these types would be used:
https://github.com/tmdiep/google-cloud-go/tree/working/pubsublite

### DIFF
--- a/pubsublite/doc.go
+++ b/pubsublite/doc.go
@@ -1,0 +1,24 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


tmdiep:
Cody added some comments to go/pubsublite-go-clientlib-interface about just using raw strings for some of these. I'll just do that for consistency with the other Go client libs. Will revise and move the path types to another file.

PR #2991, opened 2020-10-09
fix(spanner/spannertest): correct handling of NULL AND FALSE

dsymonds:
92bfe7f got most things correct, but I relied on the docs, which have an
error and so one case was incorrect.

(Internally filed as b/170436130)

### DIFF
--- a/spanner/spannertest/db_eval.go
+++ b/spanner/spannertest/db_eval.go
@@ -101,23 +101,35 @@ func (ec evalContext) evalBoolExpr(be spansql.BoolExpr) (*bool, error) {
 		// https://cloud.google.com/spanner/docs/operators#logical_operators
 		switch be.Op {
 		case spansql.And:
-			if lhs == nil || rhs == nil {
-				return nil, nil
-			}
-			b := *lhs && *rhs
-			return &b, nil
-		case spansql.Or:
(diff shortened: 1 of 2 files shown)


spanner/spannertest/db_eval.go
```diff
@@ -101,23 +101,35 @@ func (ec evalContext) evalBoolExpr(be spansql.BoolExpr) (*bool, error) {
...
-			return &b, nil
-		case spansql.Or:
-			if lhs == nil && rhs != nil && *rhs {
+			// NULL AND FALSE => FALSE
+			if rhs != nil && !*rhs {
 				return rhs, nil
 			}
-			if rhs == nil && lhs != nil && *lhs {
-				return lhs, nil
+			// NULL AND TRUE|FALSE => NULL
```
olavloite, line 116:
```suggestion
			// NULL AND TRUE|NULL => NULL
```

olavloite (approved):
That's interesting. I think there's as small copy-paste error in one comment, otherwise LGTM.

spanner/spannertest/db_eval.go
dsymonds: Thanks. Good catch.

PR #2997, opened 2020-10-11
fix(spanner/spansql): fix parsing of unary minus

dsymonds:
We cannot tokenise numbers with leading signs, because otherwise
`ID+100` is parsed as the identifier `ID` followed by the literal `+100`
instead of an arithmetic operation. This necessitates some rearrangement
of the parsing of int64 literals to accommodate that while also
preserving the ability to parse -INT_MAX.

### DIFF
--- a/spanner/spansql/parser.go
+++ b/spanner/spansql/parser.go
@@ -184,9 +184,13 @@ type token struct {
 	line, offset int
 
 	typ     tokenType
-	int64   int64
 	float64 float64
 	string  string // unquoted form for stringToken/bytesToken/quotedID
+
+	// int64Token is parsed as a number only when it is known to be a literal.
+	// This permits correct handling of operators preceding such a token,
+	// which cannot be identified as part of the int64 until later.
+	int64Base int
 }
 
 type tokenType int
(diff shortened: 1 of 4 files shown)


spanner/spansql/parser.go
```diff
@@ -2350,7 +2364,40 @@ func (p *parser) parseUnaryArithOp() (Expr, *parseError) {
...
+				f := ntok.float64
+				if op == "-" {
+					f = -f
+				}
+				return FloatLiteral(f), nil
+			}
+		}
+		if op == "+" {
+			// A plus followed by anything that isn't a numeric literal is invalid.
+			// Spanner does not have unary plus.
```
olavloite, line 2392: It seems to be undocumented, but the following is a valid query:

```
select +SingerId
from Singers
```

And so is also the following:

```
select ++-++SingerId
from Singers
```

So I would say it's ok to leave the implementation as is, but the comment should be changed.

spanner/spansql/parser_test.go
```diff
@@ -656,6 +659,9 @@ func TestParseFailures(t *testing.T) {
 		// Found by fuzzing.
 		// https://github.com/googleapis/google-cloud-go/issues/2196
 		{query, `/*/*/`, "invalid comment termination"},
+		// Spanner does not have unary plus.
+		// This should also not cause a panic.
+		{expr, `+(Y - Z)`, "unary plus"},
```
olavloite, line 664: See above. It is undocumented, but it is supported and will not cause a parse error on Cloud Spanner.

olavloite (approved):
See comment on unary plus, otherwise LGTM.

spanner/spansql/parser.go
dsymonds: Well, let's just fix this now. It's a good point that we should match the real behaviour; this isn't the first time the docs have been wrong.

I've filed an internal bug (yet another!) about this, but I've also updated this change to properly handle unary plus.

Issue #2855, opened 2020-09-14
BigQuery: Your client has issued a malformed or illegal request, That's all we know


deankarn:
**Client**

BigQuery

**Environment**

gcr.io/distroless/static-debian10 docker

**Go Environment**

$ go version
go version go1.15.1 

$ go env

**Code**

e.g.
```go
package main

func main() {
	client, _ := bigquery.NewClient(
		context.Background(),
		c.ProjectID,
		c.Options...,
	)
        client.Dataset(c.Dataset).Table(c.Table).Inserter().Put(...)
}
```

**Expected behavior**
To succeed or return a useful error

**Actual behavior**

A cryptic error with absolutely no details is returned:
```
bigquery: max attempts (5) reached - googleapi: got HTTP response code 400 with body: <!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
  <title>Error 400 (Bad Request)!!1</title>
  <style>
    *{margin:0;padding:0}html,code{font:15px/22px arial,sans-serif}html{background:#fff;color:#222;padding:15px}body{margin:7% auto 0;max-width:390px;min-height:180px;padding:30px 0 15px}* > body{background:url(//www.google.com/images/errors/robot.png) 100% 5px no-repeat;padding-right:205px}p{margin:11px 0 22px;overflow:hidden}ins{color:#777;text-decoration:none}a img{border:0}@media screen and (max-width:772px){body{background:none;margin-top:0;max-width:none;padding-right:0}}#logo{background:url(//www.google.com/images/branding/googlelogo/1x/googlelogo_color_150x54dp.png) no-repeat;margin-left:-5px}@media only screen and (min-resolution:192dpi){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat 0% 0%/100% 100%;-moz-border-image:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) 0}}@media only screen and (-webkit-min-device-pixel-ratio:2){#logo{background:url(//www.google.com/images/branding/googlelogo/2x/googlelogo_color_150x54dp.png) no-repeat;-webkit-background-size:100% 100%}}#logo{display:inline-block;height:54px;width:150px}
  </style>
  <a href=//www.google.com/><span id=logo aria-label=Google></span></a>
[...]

PR #3024, opened 2020-10-14
feature(bigquery): improve error guidance for streaming inserts

shollyman:
This PR attempts to clarify guidance for users of the streaming insert mechanism.

BigQuery enforces a variety of limitations on client requests (size, qps, etc).  When
those limits are more severely violated, BigQuery replies with less specific information
about the issue, and instead generates an unstructured HTTP 400 client error response.

This PR adds guidance to doc.go about this behavior, and adds an integration test that
induces these errors to ensure the error surface is behaving as documented.

Fixes: #2855


bigquery/inserter.go
```diff
@@ -19,16 +19,28 @@ import (
...
 
+// ErrRequestMalformed is surfaced for cases where the response from the backend
+// indicates a generic error, but is unstructured and ambiguous.
+//
+// For streaming insert requests, this often indicates that either the row
+// data is excessively large, or that the batch factor is too high (too many rows
+// sent as part of a single insert).  However, the error response provides no
+// further diagnostic information, and may indicate problems other than size such
+// as QPS or fundamental problems with the payload.
+var ErrRequestMalformed = fmt.Errorf("your client has issued a malformed or illegal request")
```
hongalex, line 42: prefer `errors.New` if you aren't using wrapping

hongalex (approved):

bigquery/inserter.go
codyoss: Do you expect people to type switch on this error and handle this case specially? Just trying to understand if this needs to be exported.

shollyman: I'm coming around to the viewpoint of not adding a custom error, and simply augmenting the package documentation to be more explicit.  It provides little value and doesn't reduce ambiguity sufficiently.

shollyman:
Nothing further has bubbled up in discussions, so I'm going to move forward with committing this as it  doesn't yield any changes in the library contracts; it merely attempts to improve documentation and testing.

PR #3026, opened 2020-10-15
feat(pubsublite): Types for resource paths and topic/subscription configs

tmdiep:
{Topic|Subscription}ConfigToUpdate are really just a subset of {Topic|Subscription}Config, but I adopted the *ToUpdate pattern of some of the other client libraries and for future flexibility and to provide clearer documentation.

The config fields were flattened for simplicity.

Type and field names were chosen to match Cloud Pub/Sub, where there is overlap.

See http://go/pubsublite-go-clientlib-interface.


tmdiep:
I also wanted to add manuelmenzella-google as reviewer, but the account didn't appear in the list.

Also, should we set owners for pubsublite? For the smaller future clean ups, we probably don't need yoshi-go-admins.

pubsublite/types.go
```diff
@@ -0,0 +1,140 @@
...
+	parts := strings.Split(input, "-")
+	if len(parts) != 2 {
+		return fmt.Errorf("pubsublite: invalid region %q", input)
+	}
+	return nil
+}
+
+// ZoneToRegion returns the region that the given zone is in.
+func ZoneToRegion(zone string) string {
+	return zone[0:strings.LastIndex(zone, "-")]
```
hongalex, line 139: Since this is a publicly exposed function, we might want to consider bounds checking in the case the user passes in an invalid string (perhaps by a check to `ValidateZone`)?

pubsublite/types.go
```diff
@@ -0,0 +1,140 @@
...
+func (t TopicPath) String() string {
+	return fmt.Sprintf("projects/%s/locations/%s/topics/%s", t.Project, t.Zone, t.TopicID)
+}
+
+// Location returns the path of the parent location.
+func (t TopicPath) Location() LocationPath {
+	return LocationPath{Project: t.Project, Zone: t.Zone}
+}
+
+var topicPathRE = regexp.MustCompile(`^projects/([^/]+)/locations/([^/]+)/topics/([^/]+)$`)
```
hongalex, line 63: If we are checking for structure, perhaps we should also check for validity of the topic name? Resource names are [documented for Pub/Sub](https://cloud.google.com/pubsub/docs/admin#resource_names) but not sure about Lite.

pubsublite/types.go
```diff
@@ -0,0 +1,140 @@
...
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// LocationPath stores a path consisting of a project and zone.
+type LocationPath struct {
+	// A Google Cloud project. The project ID (e.g. "my-project") or the project
+	// number (e.g. 987654321) can be provided.
```
hongalex, line 25: nit: perhaps use the quoted string for project number for explicitiness
```suggestion
	// number (e.g. "987654321") can be provided.
```

pubsublite/types.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (l LocationPath) String() string {
+	return fmt.Sprintf("projects/%s/locations/%s", l.Project, l.Zone)
+}
+
+// TopicPath stores the full path of a Google Pub/Sub Lite topic.
+// See https://cloud.google.com/pubsub/lite/docs/topics for more information.
+type TopicPath struct {
+	// A Google Cloud project. The project ID (e.g. "my-project") or the project
+	// number (e.g. 987654321) can be provided.
```
hongalex, line 42: same as above

pubsublite/types.go
```diff
@@ -0,0 +1,140 @@
...
+func (s SubscriptionPath) String() string {
+	return fmt.Sprintf("projects/%s/locations/%s/subscriptions/%s", s.Project, s.Zone, s.SubscriptionID)
+}
+
+// Location returns the path of the parent location.
+func (s SubscriptionPath) Location() LocationPath {
+	return LocationPath{Project: s.Project, Zone: s.Zone}
+}
+
+var subsPathRE = regexp.MustCompile(`^projects/([^/]+)/locations/([^/]+)/subscriptions/([^/]+)$`)
```
hongalex, line 102: same concern as the topicPathRE above

hongalex (approved):
Left some minor nits and comments, but otherwise LGTM

pubsublite/types.go
tmdiep: Done, good catch. I originally had this as a function of an already validated CloudZone, but correctness is not guaranteed now.

pubsublite/types.go
tmdiep: Done

pubsublite/types.go
tmdiep: I think we should forward this to the server for validity checking, just in case the criteria changes in the future. I double-checked the server impl and it uses the same util function as Cloud Pub/Sub.

tmdiep:
Thanks Alex, comments addressed.

pubsublite/config.go
```diff
@@ -0,0 +1,264 @@
...
+	// The full path of a topic.
+	Name TopicPath
+
+	// The number of partitions in the topic. Must be at least 1. Cannot be
+	// changed after creation.
+	PartitionCount int64
+
+	// Publish throughput capacity per partition in MiB/s.
+	// Must be >= 4 and <= 16.
+	PublishCapacityMiBPerSec int32
```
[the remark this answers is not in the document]
codyoss: Also, I would consider just making these ints for easy of use.

pubsublite/types.go
```diff
@@ -0,0 +1,143 @@
...
+	Zone string
+}
+
+func (l LocationPath) String() string {
+	return fmt.Sprintf("projects/%s/locations/%s", l.Project, l.Zone)
+}
+
+// TopicPath stores the full path of a Google Pub/Sub Lite topic.
+// See https://cloud.google.com/pubsub/lite/docs/topics for more information.
+type TopicPath struct {
```
codyoss, line 40: I wonder if this could struct could be refactored out. Maybe the client could take a zone and project in the constructor. Then topic methods could just request an ID. Then there would be no need for TopicPath, SubscriptionPath, and LocationPath.

codyoss (commented):
Sorry for not getting to this sooner. I thought I would just leave some things to consider if there is still flexibility in the API.

pubsublite/config.go
```diff
@@ -0,0 +1,264 @@
...
+// TopicConfig describes the properties of a Google Pub/Sub Lite topic.
+// See https://cloud.google.com/pubsub/lite/docs/topics for more information
+// about how topics are configured.
+type TopicConfig struct {
+	// The full path of a topic.
+	Name TopicPath
+
+	// The number of partitions in the topic. Must be at least 1. Cannot be
+	// changed after creation.
+	PartitionCount int64
```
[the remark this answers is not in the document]
tmdiep: Will do

pubsublite/config.go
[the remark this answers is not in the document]
tmdiep: In the Java client lib, we just pass it through for the server to validate. Also, I just copied a comment from the proto, but technically 16 is only the default upper limit and is bound by an API quota limit, which can be increased. I'll update this comment.

Will change int32 to int.

pubsublite/types.go
```diff
@@ -0,0 +1,143 @@
...
+func (s SubscriptionPath) Location() LocationPath {
+	return LocationPath{Project: s.Project, Zone: s.Zone}
+}
+
+var subsPathRE = regexp.MustCompile(`^projects/([^/]+)/locations/([^/]+)/subscriptions/([^/]+)$`)
+
+// ParseSubscriptionPath parses the full path of a Google Pub/Sub Lite
+// subscription, which should have the format:
+// `projects/{project}/locations/{zone}/subscriptions/{id}`.
+func ParseSubscriptionPath(input string) (SubscriptionPath, error) {
```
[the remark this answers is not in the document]
tmdiep: Hmm, not sure how often users deal with these raw absolute paths. I'll change them to internal for now, we can make them public later upon request.

pubsublite/types.go
```diff
@@ -0,0 +1,143 @@
...
+	// The ID of the Google Pub/Sub Lite topic, for example "my-topic-name".
+	TopicID string
+}
+
+func (t TopicPath) String() string {
+	return fmt.Sprintf("projects/%s/locations/%s/topics/%s", t.Project, t.Zone, t.TopicID)
+}
+
+// Location returns the path of the parent location.
+func (t TopicPath) Location() LocationPath {
```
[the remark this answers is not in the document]
tmdiep: Possibly, but I'll just change them to private for now.

pubsublite/types.go
tmdiep: We can have cross-project subscriptions, where the project number of topic and subscription resources are not necessarily owned by the same project. So having a default project set for the client could cause some confusion for, e.g. Create/Update Subscription.

I also saw that [pubsub.Client](https://godoc.org/cloud.google.com/go/pubsub#pkg-index) had evolved to support variants Topic/TopicInProject, Subscription/SubscriptionInProject, so wanted to avoid this, as it seems multi-project admin does occur.

Similar remarks for zone, as the admin server is regional. Users can use the same client to manage all resources within the region.

For these reasons, I thought it would be best to give the client more flexibility with projects and zones.

pubsublite/config.go
```diff
@@ -0,0 +1,264 @@
...
+	pb "google.golang.org/genproto/googleapis/cloud/pubsublite/v1"
+	fmpb "google.golang.org/genproto/protobuf/field_mask"
+)
+
+// TopicConfig describes the properties of a Google Pub/Sub Lite topic.
+// See https://cloud.google.com/pubsub/lite/docs/topics for more information
+// about how topics are configured.
+type TopicConfig struct {
+	// The full path of a topic.
+	Name TopicPath
```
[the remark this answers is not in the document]
tmdiep: Hmm... Unlike most other services, Pub/Sub Lite resources are zonal, hence topics are globally uniquely identified by {project, zone, id} and it's common to have to copy all components across to different objects, function args, etc. I think keeping the fields together as a full path is convenient and less error prone.

tmdiep:
Thanks Cody, I'll address most of your comments in https://github.com/googleapis/google-cloud-go/pull/3036.

PR #3030, opened 2020-10-15
docs(datastore): say to call (*Client).Close

tbpg:


codyoss (approved):
LGTM, do you think this is something we should be doing holistically across all clients? Even in the generated layers?

codyoss:
Or maybe this would be nice to put in a sort of best practices doc we put in some readme.

tbpg:
> do you think this is something we should be doing holistically across all clients?

Yes. This has caused issues for new and experienced users. So, additional nudges seem necessary.

Maybe GoDoc and a best practices doc?

tbpg:
Filed https://github.com/googleapis/google-cloud-go/issues/3031.

PR #3042, opened 2020-10-17
feat(spanner/spannertest): implement RIGHT JOIN

dsymonds:
Restructure and make a unified JOIN implementation.

FULL JOIN is still not supported, but all other variations of joins now
work.

### DIFF
--- a/spanner/spannertest/db_query.go
+++ b/spanner/spannertest/db_query.go
@@ -658,8 +658,23 @@ func newJoinIter(lhs, rhs *rawIter, lhsEC, rhsEC evalContext, sfj spansql.Select
 		jt: sfj.Type,
 		ec: lhsEC,
 
-		lhs:     lhs,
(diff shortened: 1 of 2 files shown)


spanner/spannertest/db_query.go
```diff
@@ -744,190 +765,123 @@ func (ji *joinIter) prepUsing(using []spansql.ID, lhsEC, rhsEC evalContext) erro
...
+		populate(primary, secondary)
 		return true, nil
 	}
+	ji.zero = func(primary row) {
+		populate(primary, nil)
+	}
 	return nil
 }
 
 type joinIter struct {
```
dsymonds, line 815: I suggest starting to review from the changes to the `joinIter` struct; the rest should hopefully flow directly from that.

olavloite (approved):
This looks really good.

PR #3043, opened 2020-10-18
fix(spanner/spannertest): fix ORDER BY combined with SELECT aliases

dsymonds:
This was broken due to the ORDER BY implementation that was a bit of a
hack. Replace that hack with a more explicit evaluation of the ORDER BY
sort keys, which also removed a duplicate chunk of row sorting code.

This also required a restructuring of how locking works during an
evaluation, which cleaned up some other code.

Also include a minor revision to README.md.

### DIFF
--- a/spanner/spannertest/db_query.go
+++ b/spanner/spannertest/db_query.go
@@ -191,10 +191,27 @@ type selIter struct {
 	ec   evalContext
 	cis  []colInfo
 	list []spansql.Expr
+
+	distinct bool // whether this is a SELECT DISTINCT
+	seen     []row
 }
 
-func (si selIter) Cols() []colInfo { return si.cis }
-func (si selIter) Next() (row, error) {
+func (si *selIter) Cols() []colInfo { return si.cis }
+func (si *selIter) Next() (row, error) {
+	for {
+		r, err := si.next()
+		if err != nil {
+			return nil, err
+		}
+		if si.distinct && !si.keep(r) {
+			continue
+		}
+		return r, nil
+	}
+}
+
+// next retrieves the next row for the SELECT and evaluates its expression list.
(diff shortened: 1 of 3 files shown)


spanner/spannertest/db_query.go
```diff
@@ -350,33 +380,76 @@ func (d *database) Query(q spansql.Query, params queryParams) (rowIter, error) {
...
+	for _, name := range names {
+		qc.tables = append(qc.tables, qc.tableIndex[name])
+	}
+
+	return qc, nil
+}
+
+func (d *database) evalSelect(sel spansql.Select, qc *queryContext) (si *selIter, evalErr error) {
+	var ri rowIter = &nullIter{}
+	ec := evalContext{
```
olavloite, line 438: Not directly related to this change, but as it is an improvement to the `ORDER BY` handling, consider the following:

The aliases of this evaluation context is only filled if the statement contains a `GROUP BY` (line 469). If that were to be moved out from the if statement and always filled, then ordering by an alias would automatically also be supported:


```sql
SELECT ID AS I FROM Staff WHERE Cool IS NOT NULL ORDER BY I DESC
```

Note that the aliases should only be added to the evaluation context *after* the `WHERE` clause has been evaluated, as the following is not allowed:

```sql
SELECT ID AS I FROM Staff WHERE I=2 ORDER BY I DESC
```

olavloite (approved):
LGTM, with a suggestion on adding support for `ORDER BY` on aliases.

spanner/spannertest/db_query.go
dsymonds: Yep, it's a good idea, but I'll do it in a separate change. This one is complex enough.

PR #3046, opened 2020-10-19
feat(spanner/spansql): parse CHECK constraints

dsymonds:
This results in a minor backward-incompatible change to TableConstraint
to support both Check and ForeignKey appearing as constraints.

### DIFF
--- a/spanner/spannertest/README.md
+++ b/spanner/spannertest/README.md
@@ -28,6 +28,7 @@ by ascending esotericism:
 - transaction simulation
 - expression type casting, coercion
 - subselects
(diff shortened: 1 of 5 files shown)


skuruppu (approved):
If you want to mark this as a breaking change, I think we have a release coming up where we may have to do a major version bump (late Nov -- mid Dec) so we can hold off merging this in until then.

dsymonds:
Thanks, but I think it's fine. We haven't aimed to maintain any sort of strict compatibility for spansql or spannertest, and the only folk broken should be people using constraints.

PR #3050, opened 2020-10-19
feat(spanner): add metadata to RowIterator

### LINKED ISSUE #1805: spanner: provide column types in results
The underlying proto contains ResultSetMetadata.RowType for results returned from the API but spanner.Row or spanner.RowIterator are not surfacing them to the user. I'm trying to put together a "database/sql" driver for Spanner and without the column type it is not possible. 

/cc @tbpg

olavloite:
Adds ResultSetMetaData to the RowIterator struct. The metadata are available after the first call to RowIterator.Next() as long as that call did not return any other error than iterator.Done.

Fixes #1805


hengfengli (approved):
LGTM. Thanks a lot for fixing this!

PR #3055, opened 2020-10-20
fix(pubsub): do not propagate context deadline exceeded error

hongalex:
In certain scenarios (usually on network disconnect but to be fully determined), an unknown `context deadline exceeded` error from `ModifyAckDeadline` can shutdown the iterator and cause it to fail. This is usually captured by the case prior to `default` in the switch statement, but sometimes the error is not captured properly. This PR patches the issue so users are not negatively affected.


pubsub/iterator.go
```diff
@@ -438,6 +439,14 @@ func (it *messageIterator) sendModAck(m map[string]bool, deadline time.Duration)
 				recordStat(it.ctx, ModAckTimeoutCount, 1)
 				return nil
 			default:
+				if err == nil {
+					return nil
+				}
+				// This addresses an error where `context deadline exceeded` errors
+				// not captured by the previous case do not cause fatal errors.
+				if strings.Contains(err.Error(), "context deadline exceeded") {
```
codyoss, line 447: Maybe `xerrors.Is(err, context.DeadlineExceeded)`?

pubsub/iterator.go
```diff
@@ -438,6 +439,14 @@ func (it *messageIterator) sendModAck(m map[string]bool, deadline time.Duration)
...
+					return nil
```
codyoss, line 448: Should we record the stat as a ModAckTimeoutCount?

pubsub/iterator.go
```diff
@@ -438,6 +439,14 @@ func (it *messageIterator) sendModAck(m map[string]bool, deadline time.Duration)
 				recordStat(it.ctx, ModAckTimeoutCount, 1)
 				return nil
 			default:
+				if err == nil {
+					return nil
+				}
+				// This addresses an error where `context deadline exceeded` errors
+				// not captured by the previous case do not cause fatal errors.
```
noahdietz, line 446: It might be useful to include a scenario (i.e. the one we think is producing this) where this could happen. If there is a relevant GitHub issue or public thread, that might be useful to include as well. This is a pretty odd scenario and its worth being explicit about it imo.

hongalex:
Upon further investigation, the error seems to be originating from the `gax.Sleep` call. The context is cancelled, and a fallthrough happens. However, since the err is not checked directly (rather the previously stored error code which is a `Unavailable`), the `context deadline exceeded` error is bubbled up to the client library user. Will be updating this PR to address.

noahdietz:
> the error seems to be originating from the `gax.Sleep` call

Interesting, is there anything that needs to change in `gax`? Or is there a bug in `gax`?

pubsub/iterator.go
hongalex: So the `xerrors` approach doesn't currently work due to the way the error is injected into the fake. If there's an RPC error from `ModifyAckDeadline`, then the error will show up as `rpc error: code = DeadlineExceeded desc = context deadline exceeded`, so checking for context deadline exceeded wouldn't match in this case. I believe the error people are seeing _isn't_ an RPC error, but this case let's us check both. I'm ok with removing the test though, so using `xerrors.Is` makes sense.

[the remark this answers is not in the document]
hongalex: Yeah, so I have reason to believe that the `context deadline exceeded` error might be coming from something in the library rather than from grpc-go, so there isn't an associated code here. In this case, the fake simulates the error with an `Unknown context deadline exceeded`, so it doesn't 100% mimic the error users are seeing.

pubsub/iterator.go
hongalex: Filed and linked https://github.com/googleapis/google-cloud-go/issues/3060

pubsub/iterator.go
codyoss: You could also keep the test and define a private `ServerReactorOption` that accepts an `error`? If the API for `WithErrorInjection` is still flexible, it might be worth having that take an error vs info to construct a grpc status.

hongalex: Yeah, I considered that, but it's still the fake server returning the error. Because of that, the grpc library will wrap any error with `rpc error: ...`

noahdietz (commented):
LGTM but I'll let Cody Approve the PR.

codyoss (approved):
LGTM

PR #3058, opened 2020-10-21
feat(spanner): support rw-transaction with options

hengfengli:
This is a generic api change to support passing options into transactions. This will be used to support a couple of new features.

### DIFF
--- a/spanner/client.go
+++ b/spanner/client.go
@@ -424,11 +424,29 @@ func checkNestedTxn(ctx context.Context) error {
 func (c *Client) ReadWriteTransaction(ctx context.Context, f func(context.Context, *ReadWriteTransaction) error) (commitTimestamp time.Time, err error) {
 	ctx = trace.StartSpan(ctx, "cloud.google.com/go/spanner.ReadWriteTransaction")
 	defer func() { trace.EndSpan(ctx, err) }()
+	resp, err := c.rwTransaction(ctx, f, TransactionOptions{})
+	return resp.CommitTs, err
+}
+
+// ReadWriteTransactionWithOptions executes a read-write transaction with
+// configurable options, with retries as necessary.
+//
+// ReadWriteTransactionWithOptions is a configurable ReadWriteTransaction.
+//
+// See https://godoc.org/cloud.google.com/go/spanner#ReadWriteTransaction for
+// more details.
+func (c *Client) ReadWriteTransactionWithOptions(ctx context.Context, f func(context.Context, *ReadWriteTransaction) error, options TransactionOptions) (resp CommitResponse, err error) {
+	ctx = trace.StartSpan(ctx, "cloud.google.com/go/spanner.ReadWriteTransactionWithOptions")
+	defer func() { trace.EndSpan(ctx, err) }()
+	resp, err = c.rwTransaction(ctx, f, options)
+	return resp, err
+}
+
+func (c *Client) rwTransaction(ctx context.Context, f func(context.Context, *ReadWriteTransaction) error, options TransactionOptions) (resp CommitResponse, err error) {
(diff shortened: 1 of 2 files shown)


spanner/transaction.go
```diff
@@ -1105,19 +1126,15 @@ func NewReadWriteStmtBasedTransaction(ctx context.Context, c *Client) (*ReadWrit
 // Commit tries to commit a readwrite transaction to Cloud Spanner. It also
 // returns the commit timestamp for the transactions.
 func (t *ReadWriteStmtBasedTransaction) Commit(ctx context.Context) (time.Time, error) {
```
skuruppu, line 1128: Would you also need a `CommitWithOptions`? Especially given the nature of #3056.

hengfengli: No, because we can pass `CommitOptions` to a transaction via `TransactionOptions` (calling `NewReadWriteStmtBasedTransactionWithOptions`). 

Later, for this `Commit` method, we can have two options: 

1. Update `Commit` to return `(CommitResponse, error)` (a breaking change)
2. Add a new method `CommitReturnResponse` to return `(CommitResponse, error)` (not a breaking change)

spanner/client.go
```diff
@@ -424,11 +424,29 @@ func checkNestedTxn(ctx context.Context) error {
...
+}
+
+// ReadWriteTransactionWithOptions executes a read-write transaction with
+// configurable options, with retries as necessary.
+//
+// ReadWriteTransactionWithOptions is a configurable ReadWriteTransaction.
+//
+// See https://godoc.org/cloud.google.com/go/spanner#ReadWriteTransaction for
+// more details.
+func (c *Client) ReadWriteTransactionWithOptions(ctx context.Context, f func(context.Context, *ReadWriteTransaction) error, options TransactionOptions) (resp CommitResponse, err error) {
```
olavloite, line 438: Would it maybe be better to let the `TransactionOption` be an interface, and the `options` argument variadic? That would mean that:
1. We don't need a second `ReadWriteTransactionWithOptions` method, but can just change the `ReadWriteTransaction` method. This is still a breaking change, but only for some very specific use cases. See discussion in https://stackoverflow.com/questions/55163005/could-adding-variadic-parameters-to-a-function-break-existing-code.
1. Passing a `struct` means that you have to take into account what to do with the empty value of each field in the struct. See for example this article: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

This is also the approach taken by the `NewClient` function in [client.go](https://github.com/googleapis/google-cloud-go/blob/38a56325f4eed4a00a1661895369f69acf402d84/spanner/client.go#L136).

WDYT?

spanner/transaction.go
skuruppu: Ah ok, thanks for explaining.

As for the `Commit`, it would be neat if we could have #1, but since this is a public method, to avoid the risk of inconveniencing users, I think we'll have to go with #2. We could call it `CommitWithStats` in line with `QueryWithStats` but the worry is that we might return something other than stats in the future.

spanner/client.go
hengfengli: @olavloite Yep, I have considered that as well. But the problem is the return type. We have to add a new method to return `CommitResponse`. So we can't just change the existing `ReadWriteTransaction` method.

spanner/transaction.go
hengfengli: Yes, `CommitWithStats` is very specific. We need a very generic name here.

spanner/client.go
olavloite: Ah, yes that is a good point.

olavloite (approved):

PR #3061, opened 2020-10-21
chore(storage): update storage owners

frankyn:


codyoss:
It looks like @tritone is not apart of api-storage, FYI

tritone (commented):
This means that the whole team will be emailed everything about every storage review. Are you sure you want that?

Issue #1590, opened 2019-09-20
logging: incorrect parsing of Trace Context header without sampling indicator


broady:
Sometimes the header doesn't include the sampling information.

e.g.:

```
X-Cloud-Trace-Context: 6518720b40e1efacf2299758d229210a/14866474463603563926
```

Current assumption:

```
var reCloudTraceContext = regexp.MustCompile(`([a-f\d]+)/([a-f\d]+);o=(\d)`)
```

odeke-em:
Hey @broady, thanks for this report!

Coming here from the trace-context specification for Stackdriver https://cloud.google.com/trace/docs/troubleshooting#force-trace
![Screen Shot 2019-09-20 at 1 26 05 PM](https://user-images.githubusercontent.com/4898263/65356820-3add0f80-dbaa-11e9-97e3-17fb8059f0a5.png)
and  also from the CL that I sent to enforce it in commit https://github.com/googleapis/google-cloud-go/commit/392bd16d8f72017d4999a896d28ccbfd536aca52#diff-ad8a2265fa17b7808fca5ae77cada94f are we asking that we parse a malformed `X-Cloud-Trace-Context`? That is a header value without the `;o=(\d)` part?

broady:
Hm, if this is indeed the spec then I need to file a bug against App Engine.

nicoleczhu:
Hi @broady fyi at the client library level, we now accept user input where sampleTrace field is optional (in which case, we default sampleTrace to false). 

However, this fix doesn't solve the lower lvl issue where the service is not flagging it as true when it should be. Thanks!

PR #3062, opened 2020-10-21
fix(logging): allow X-Cloud-Trace-Context fields to be optional

nicoleczhu:
Fixes: #1590 

**Changes according to spec:**
1.  `trace`, `spanID` and `traceSampled` fields in X-Cloud-Trace-Context can be optional (see context/trace_context.proto)
2. TraceSample should default to false (verified with nwang@)
3. Trace/span IDs should still be included in log contexts even when traceSampled is false, because "A non-sampled trace value is still useful as a request correlation identifier." ([LogEntry spec](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry))

**Details:**
[Intended Use case](https://cloud.google.com/trace/docs/setup#force-trace)

The following header specifications by users are now recognized: 
Screen Shot 2020-10-21 at 12 30 23 PM

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -873,7 +873,13 @@ func (l *Logger) writeLogEntries(entries []*logpb.LogEntry) {
 // (for example by calling SetFlags or SetPrefix).
 func (l *Logger) StandardLogger(s Severity) *log.Logger { return l.stdLoggers[s] }
 
-var reCloudTraceContext = regexp.MustCompile(`([a-f\d]+)?(/([a-f\d]+))?(;o=(\d))?`)
+var reCloudTraceContext = regexp.MustCompile(
+	// Matches on "TRACE_ID"
+	`([a-f\d]+)?` +
+		// Matches on "/SPAN_ID"
+		`(?:/([a-f\d]+))?` +
+		// Matches on ";0=TRACE_TRUE"
+		`(?:;o=(\d))?`)
 
 func deconstructXCloudTraceContext(s string) (traceID, spanID string, traceSampled bool) {
 	// As per the format described at https://cloud.google.com/trace/docs/setup#force-trace
@@ -887,7 +893,7 @@ func deconstructXCloudTraceContext(s string) (traceID, spanID string, traceSampl
 	//   * traceSampled (optional): 	true
 	matches := reCloudTraceContext.FindStringSubmatch(s)
 
-	traceID, spanID, traceSampled = matches[1], matches[3], (matches[5] == "1")
+	traceID, spanID, traceSampled = matches[1], matches[2], matches[3] == "1"
 
 	if spanID == "0" {
 		spanID = ""
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -873,7 +873,7 @@ func (l *Logger) writeLogEntries(entries []*logpb.LogEntry) {
 // (for example by calling SetFlags or SetPrefix).
 func (l *Logger) StandardLogger(s Severity) *log.Logger { return l.stdLoggers[s] }
 
-var reCloudTraceContext = regexp.MustCompile(`([a-f\d]+)/([a-f\d]+);o=(\d)`)
+var reCloudTraceContext = regexp.MustCompile(`([a-f\d]+)?(/([a-f\d]+))?(;o=(\d))?`)
 
 func deconstructXCloudTraceContext(s string) (traceID, spanID string, traceSampled bool) {
 	// As per the format described at https://cloud.google.com/trace/docs/setup#force-trace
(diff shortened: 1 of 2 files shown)


logging/logging.go
```diff
@@ -873,33 +873,25 @@ func (l *Logger) writeLogEntries(entries []*logpb.LogEntry) {
...
 	// We expect:
-	//   * traceID:         "105445aa7843bc8bf206b120001000"
-	//   * spanID:          ""
-	//   * traceSampled:    true
-	matches := reCloudTraceContext.FindAllStringSubmatch(s, -1)
-	if len(matches) != 1 {
-		return
-	}
+	//   * traceID (optional):         "105445aa7843bc8bf206b120001000"
+	//   * spanID (optional):          "0"
```
odeke-em, line 886: Why is '0' being used?

logging/logging.go
```diff
@@ -873,33 +873,25 @@ func (l *Logger) writeLogEntries(entries []*logpb.LogEntry) {
...
+	//   * traceSampled (optional):    1 (defaults to false)
```
odeke-em, line 887: Why 1? This comment was meant to illustrate the final value and implicitly traceSampled is false by default :) I'd revert the end of this comment and keep the `(optional):` or basically make it `true` instead of `1`.

logging/logging.go
```diff
@@ -873,33 +873,25 @@ func (l *Logger) writeLogEntries(entries []*logpb.LogEntry) {
...
+	matches := reCloudTraceContext.FindStringSubmatch(s)
 
-	sub := matches[0]
-	if len(sub) != 4 {
-		return
-	}
+	traceID, spanID, traceSampled = matches[1], matches[3], (matches[5] == "1")
```
odeke-em, line 890: Please restore (and adjust) the length checks to ensure that we have at least 5 components otherwise we'll encounter crashes later on.

odeke-em, line 890: Nit, let's use `matches[5] == "1"` but only after ensuring that we have that component.

logging/logging_unexported_test.go
```diff
@@ -288,18 +288,54 @@ func TestToLogEntryTrace(t *testing.T) {
...
 						URL:    u,
-						Header: http.Header{"X-Cloud-Trace-Context": {"105445aa7843bc8bf206b120001000/0;o=0"}},
+						Header: http.Header{"X-Cloud-Trace-Context": {"105445aa7843bc8bf206b120001000/;o=0"}},
 					},
 				},
 			},
 			logging.LogEntry{Trace: "projects/P/traces/105445aa7843bc8bf206b120001000"},
 		},
+		{
+			"X-Trace-Context header with blank TraceSampled",
```
odeke-em, line 316: `with missing traceSampled aka ?o=*`

odeke-em (commented):
Thank you for this change @nicoleczhu! I've added some suggestions, please take a look.

logging/logging.go
nicoleczhu: Good question. This is the expected value from users (in majority cases) based on our documentation instructions here: 
https://cloud.google.com/trace/docs/setup#force-trace

I can update the comments to reflect both values: default vs expected-from-users

logging/logging.go
nicoleczhu: Oh yeah I was thinking about how to best represent this. Again I just took the example expected value based on our documentation instructions here: 
https://cloud.google.com/trace/docs/setup#force-trace

I'll play around with the comments to more clearly communicate what we're trying to do.

logging/logging.go
nicoleczhu: I double checked this one, given the new regex pattern matching on FindStringSubmatch, it will always return 6 components even if string s is "".  See sample code here: https://play.golang.org/p/-adjLHwv1VD

logging/logging_unexported_test.go
nicoleczhu: Good point! I'll update the test name to "missing TraceSampled". But it also works with blank TraceSampled:)

logging/logging.go
nicoleczhu: Regx match will always return 6 components even on nil, or very wrongly formatted strings

simonz130 (approved):

logging/logging.go
odeke-em: I meant instead of `(matches[5] == "1")` do `matches[5] == "1"`

odeke-em: That regex isn't very legible, I'd say just for the sake of maintaining invariants and easily inspectable behavior, let's enforce those length counts. A single change to that regex without the appropriate checks can cause later surprises.

nicoleczhu: Ahh sure thing

nicoleczhu: That's a good point. How about I add descriptive comments to the regex itself, so folks changing the regex don't encounter surprises. 

I'd really prefer disclosing in comments, rather than enforcing a `if(true)` type of check in code. I think checking length might also confuse folks looking at the regex in the future, in that it suggests the regex pattern could return various lengths (when it never will).

odeke-em: Judgement call/preference :) but personally I'd ensure that count and even panic if not

nicoleczhu: Update:
1. I broke apart the regex, adding comments for readability. 
2. I used non-capturing groups this time, so the matched indeces are more intuitive (1,2,3) rather than (1,3,5).

Thanks for the style feedback on this!

codyoss (approved):
LGTM

PR #3075, opened 2020-10-23
feat(pubsublite): Publish settings and errors

tmdiep:
PublishSettings are almost identical to [pubsub.PublishSettings](https://godoc.org/cloud.google.com/go/pubsub#PublishSettings).

Some max thresholds were changed to be consistent with the [Pub/Sub Lite Java client library](https://github.com/googleapis/java-pubsublite/blob/master/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/Constants.java).


codyoss (commented):
LGTM, but will defer to Alex

pubsublite/settings.go
```diff
@@ -0,0 +1,104 @@
...
+	DelayThreshold: 10 * time.Millisecond,
+	CountThreshold: 100,
+	ByteThreshold:  1e6,
+	Timeout:        60 * time.Second,
+	// By default set to a high limit that is not likely to occur, but prevents
+	// OOM errors in clients.
+	BufferedByteLimit: 1 << 30, // 1 GiB
+}
+
+func validatePublishSettings(settings PublishSettings) error {
```
hongalex, line 81: Is this going to be called per publish or upon PublishClient instantiation? I don't think I've seen examples of client libraries returning grpc errors (usually they are returned by the server). @codyoss do you have thoughts on this?

pubsublite/settings_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"testing"
+	"time"
+)
+
+func TestValidatePublishSettings(t *testing.T) {
+	for _, tc := range []struct {
+		desc string
+		// settingsFunc is passed DefaultPublishSettings
```
hongalex, line 24: nit: I think this comment is a bit confusing when settingsFunc is passed non default options (or maybe I'm misunderstanding something)

tmdiep: Agree that this is confusing. I refactored these funcs to take a pointer to settings and mutate it.

pubsublite/settings.go
tmdiep: It's called upon PublishClient instantiation.

I saw them returned in the firestore library and maybe a few instances in bigtable. Agree that it can be misleading, as when printed the error message mentions "grpc".

The intention was for users to always be able to extract a code from error messages coming from the pubsublite library, but happy to just return a normal error if there are downsides to this (or it's not convention across all the libraries).

codyoss: I think a normal error would be fine here. Maybe the same for the overflow error. Hard to tell without seeing how things are used.

tmdiep: Done.
I'll use normal errors throughout the library. It should be easy to change later if there's a compelling reason to.

hongalex (approved):

Issue #1862, opened 2020-03-18
logging: library code should not crash


ghasemloo:
**Client**

Stackdriver Logging
 
**Environment**

all

**Code**

replace panics

https://github.com/googleapis/google-cloud-go/commit/0d861fa3d953cd0912ea4403a996d53aa1df2ae0

**Expected behavior**

library code should not crash (except at startup)

**Actual behavior**

library code crashes when request is nil

codyoss:
From looking at both of these they seem like actual valid reasons to panic. I could be missing something though. Did you trigger these panics while your code was running?

codyoss:
Closing due to lack of response. Please reopen if you have more information to share.

ghasemloo:
Yes, it did, the request was nil I think.

Based on Google's internal go style library code should not panic: go/gocomments#dont-panic
Filed here instead of internally since this seems to be the source of truth for the code.

ghasemloo:
@jba fyi

PR #3076, opened 2020-10-23
fix(logging): do not panic in library code

nicoleczhu:
fixes: #1862 

Changes: 
- No longer panic on User introduced errors. Instead we log the error and let program proceed
- We log.Fatalf("ptypes.TimestampProto(time.Unix(0,0)) failed: %v", err) to exit program, since it's a fatal error/likely not induced by Users


logging/logging.go
```diff
@@ -180,7 +180,7 @@ func init() {
 	var err error
 	unixZeroTimestamp, err = ptypes.TimestampProto(time.Unix(0, 0))
 	if err != nil {
-		panic(err)
+		log.Fatalf("ptypes.TimestampProto(time.Unix(0,0)) failed: %v", err)
```
codyoss, line 183: For this panic I suggest just moving this logic to `NewClient` and return the error if it occurs

logging/logging.go
```diff
@@ -803,7 +804,8 @@ func jsonValueToStructValue(v interface{}) *structpb.Value {
 		}
 		return &structpb.Value{Kind: &structpb.Value_ListValue{ListValue: &structpb.ListValue{Values: vals}}}
 	default:
-		panic(fmt.Sprintf("bad type %T for JSON value", v))
+		log.Printf("Error: bad type %T for JSON value", v)
```
codyoss, line 807: It looks like `toProtoStruct` returns an error. I think this method, and the method that calls it could be refactored to propagate an error returned here.

logging/logging.go
```diff
@@ -700,7 +700,8 @@ func fromHTTPRequest(r *HTTPRequest) *logtypepb.HttpRequest {
 		return nil
 	}
 	if r.Request == nil {
-		panic("HTTPRequest must have a non-nil Request")
+		log.Println("Error: HTTPRequest must have a non-nil Request")
```
codyoss, line 703: I think just returning nil in this case is valid, no need to log. Also, could we add a test to show this behavior is intentional.

codyoss (changes requested):

logging/logging.go
nicoleczhu: Good pt. Afaik, `init` in Go is intended for variable initialization. And state checking should happen within the context of the function that actually depends on that state

Will change

logging/logging.go
nicoleczhu: Will do. More idiomatic for sure

nicoleczhu: **Update**: On second thought, I don't think we should propagate this condition as an error. An invalid jsonValue shouldn't break `toLogEntry` entirely. Similarly, we do accept `nil` values (and just return a pb.NullValue). So in the case where `jsonValue` is of an unhandled type, we should return a `pb.NullValue` with a helpful stderr log for users.

However, if we absolutely want to error on this condition, then we should `log.Fatalf` at this step rather than propagate the error. Reasons (1) this is nested in a recursive function, error propagation will look ugly (2) we'll need to create a custom Error type, which adds complexity for a simple use case (3) the final stream downstream will exit anyway bc of this condition, so it might be better to exit program early rather than later.

Thoughts @codyoss ? I'll keep the former option for now

logging/logging.go
nicoleczhu: Sgtm! I added a test case for it.

If it's ok, I'll keep the `stderr` log for now since this is an easy user error to make, and having a log to `stderr` is helpful.

codyoss: SG, the reason I brought it up is we don't really have much logging in any of our clients. And what little we do we don't have hooks for our users to turn on/off or adjust verbosity.

logging/logging.go
codyoss: It breaks it today, so it is no worse right? It would now just fail gracefully. As for `log.Fatalf`, that is basically a `log.Print` + `panic`. But your call on if you want to log, see other comment below.

codyoss: If you do choose to log in this library could you preface logs with "logging:". That way users know where they are coming from. This is a common idiom used with errors.

nicoleczhu: Noted. Thanks for the pointer!

Maybe the right question I should ask is: should we be breaking on this error today? @codyoss 

I realize this scenario isn't strictly a user side error. It's happening when we fail to marshall a **valid JSON value** into a struct proto (somehow).

So in this case, maybe it's better to handle this similar to how we currently handle the `nil` json value? Which is to return a `structpb.Value_NullValue` and let the other logEntry fields be successfully logged (rather than error out the entire logEntry creation).

nicoleczhu: Discussed with @simonz130 , we'll take the "help users log as much as can be logged"  approach while logging to stderr the details of what failed to marshal. Using the log preface Cody suggested.

Thanks all!

logging/logging.go
nicoleczhu: Noted as a future improvement to let users adjust log verbosity ~

logging/logging.go
```diff
@@ -700,7 +693,8 @@ func fromHTTPRequest(r *HTTPRequest) *logtypepb.HttpRequest {
 		return nil
 	}
 	if r.Request == nil {
-		panic("HTTPRequest must have a non-nil Request")
+		log.Println("logging: HTTPRequest must have a non-nil Request")
```
tbpg, line 696: Should we be adding logging calls here? The existing `log` calls are configurable with `OnError`. Perhaps we should reuse that for the `default` case below? Not sure if there is a good way to plumb it through.

nicoleczhu (commented):
asdf

logging/logging.go
```diff
@@ -933,11 +926,15 @@ func (l *Logger) toLogEntry(e Entry) (*logpb.LogEntry, error) {
 			e.TraceSampled = e.TraceSampled || traceSampled
 		}
 	}
+	req, err := fromHTTPRequest(e.HTTPRequest)
+	if err != nil {
+		l.client.error(err)
```
tbpg, line 931: What is the full flow of how this would be triggered? I want to make sure I understand. Does this error come up intermittently depending on the request? Or would it be a fundamental issue in the user's code (where if it's broken at all, it will never work)?

nicoleczhu: This most often happens when a user makes a mistake, e.g.: 

```go
func indexHandler(w http.ResponseWriter, r *http.Request) {
   ...
    httpLog := logging.Entry{
	HTTPRequest : &logging.HTTPRequest {
		Request : // does not pass in r
	},
   }
   ... 
}
```

cmiiw, I don't think it'll come up intermittently depending on request, even if the request body is nil (the request itself isn't)

tbpg (approved):

nicoleczhu:
@codyoss ready for your review on previous change requested, thanks!

codyoss (approved):
LGTM

PR #3077, opened 2020-10-23
feat(pubsublite): Message type and message routers

tmdiep:
Message is similar to [pubsub.Message](https://godoc.org/cloud.google.com/go/pubsub#Message), with the following differences:

- Attributes can have multiple values for the same key.
- Pub/Sub Lite uses []byte for data, attribute values and ordering keys.

Message routers select a partition to route a published message to, based on whether it has an ordering key specified.


tmdiep:
Any opinions on the use of []byte everywhere in the Message struct are welcome.

IMO, strings are much easier to work with. However, Lite uses bytestrings for everything. The Java client libraries for CPS and Lite also have different Message types, so perhaps this deviation is acceptable for Go too.

codyoss (commented):
LGTM, but I'll defer to Alex

tmdiep:
WANT_LGTM=@hongalex, @manuelmenzella-google 

Do either of you have concerns about differing from [pubsub.Message](https://godoc.org/cloud.google.com/go/pubsub#Message)? It differs in the Java client libraries too, so I think it should be fine.

I prefer strings over []byte, but had some concerns about actual bytes being written to the message in another client library and looking strange converted to a string when read from the Go client lib.

hongalex:
I agree strings are easier to work with, but yeah considering people can use pub/sub across languages, this should be standardized across client libraries. I say stick with []bytes.

manuelmenzella-google (approved):

tmdiep:
Thanks everyone, []bytes it is! Merging...

PR #3084, opened 2020-10-26
feat(bigquery): Add support for authorized UDFs

shollyman:
Much like authorized views, BigQuery now supports authorized User
Defined Functions (UDFs). This PR augments the `Access` field of a
dataset entity to support routine references.


hongalex (approved):

jdpedrie:
This is a no-op for PHP, since we accept arbitrary request data and [link to the upstream documentation](https://github.com/googleapis/google-cloud-php/blob/master/BigQuery/src/BigQueryClient.php#L611-L613) for usage guidance.

PR #3092, opened 2020-10-27
feat(pubsublite): Mock server and utils for unit tests

tmdiep:
Introduces  MockLiteServer, which tests can use to register expected server requests and configure fake responses or errors. It handles unary RPCs, as well as unique bidi stream connections. It initially implements enough to test publishing, and will be extended for other unit tests.

There are also miscellaneous utils for comparing errors and a fake rand source for use in unit tests.


tmdiep:
WANT_LGTM=any

Example usage:
https://github.com/tmdiep/google-cloud-go/blob/working/pubsublite/publisher_internal_test.go
(I wanted to break up what would be a huge PR.)

tmdiep:
Friendly ping on this review.

Sorry for the large patch. It's just the beginnings of a mock server for Pub/Sub Lite APIs, inspired by the testing strategy I saw in some of the other libraries (e.g. https://github.com/googleapis/google-cloud-go/blob/master/firestore/mock_test.go). But I had to abstract it for Lite due to the numerous streaming RPCs that we have.

Example usage:
https://github.com/tmdiep/google-cloud-go/blob/f14e1ee0d3b586874025c72858ba3b1415ae5c40/pubsublite/internal/wire/publisher_test.go#L938

Thank you!

pubsublite/internal/test/verifier.go
```diff
@@ -0,0 +1,225 @@
...
+
+// wait until the `blockResponse` is released by the test, or a timeout occurs.
+// Returns immediately if there was no block.
+func (r *rpcMetadata) wait() error {
+	if r.blockResponse == nil {
+		return nil
+	}
+	select {
+	case <-time.After(blockWaitTimeout):
+		return status.Errorf(codes.FailedPrecondition, "mockserver: test did not unblock response within %v", blockWaitTimeout)
```
hongalex, line 48: I wonder if there's a better code here. Maybe `codes.Unavailable`?

pubsublite/internal/test/verifier.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+// Pop validates the received request with the next {request, response, error}
+// tuple.
+func (v *RPCVerifier) Pop(gotRequest interface{}) (interface{}, error) {
+	v.mu.Lock()
+	defer v.mu.Unlock()
+
+	v.numCalls++
+	elem := v.rpcs.Front()
```
hongalex, line 108: If this is a stack, shouldn't we be checking the `v.rpcs.Back()`? I think a stack would be inefficient here when implemented as a list though, so maybe it'd be better change the comments to say it's a queue instead.

pubsublite/internal/test/verifier.go
```diff
@@ -0,0 +1,225 @@
...
+
+	for elem := v.rpcs.Front(); elem != nil; elem = elem.Next() {
+		v.numCalls++
+		rpc, _ := elem.Value.(*rpcMetadata)
+		v.t.Errorf("call(%d): did not receive expected request:\n%v", v.numCalls, rpc.wantRequest)
+	}
+	v.rpcs.Init()
+}
+
+// streamVerifiers stores a stack of verifiers for unique stream connections.
```
hongalex, line 165: Same comment as earlier, is this supposed be a queue? If not, could also change the `PushBack` calls to `PushFront`

pubsublite/internal/test/verifier.go
tmdiep: You're absolutely correct, these indeed ended up being queues. I forgot to update the comments. Fixed now.
Thanks!

pubsublite/internal/test/verifier.go
tmdiep: Yeah, DeadlineExceeded fits here too. The problem was that those are retryable codes, so the client would try to reconnect another stream, which would result in another error message, and it wasn't immediately obvious what was going wrong in the test. So I had to use FailedPrecondition, which was the most fitting of all the non-retryable codes.

tmdiep:
Thanks Alex, comments addressed. PTAL.

hongalex (approved):

PR #3094, opened 2020-10-27
docs(pubsublite): add readme for pubsublite

anguillanneuf:
Set expectation for users.

### DIFF
--- a/pubsublite/README.md
+++ b/pubsublite/README.md
@@ -1,5 +1,6 @@
-## Cloud Pub/Sub Lite 
+## Cloud Pub/Sub Lite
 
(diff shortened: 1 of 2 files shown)


hongalex (approved):

hongalex (commented):
Can you add the license header here too?
Edit: actuslly license header is only needed for doc.go. could you add that file as well? Thanks!

anguillanneuf:
@hongalex What goes in doc.go for Pub/Sub Lite? Could that be in a separate PR?

hongalex:
Yeah, just the same warning message. We can flesh out the details later.

anguillanneuf:
Sounds good. Done.

Issue #3074, opened 2020-10-22
logging: add httprequest sample


nicoleczhu:
Problem: 
It's not immediately obvious to users how to log HTTPRequest log types. See #2912  (I was initially confused as well, you have to really peruse the docs for such a common use case)

Solution: 

- [x] Add godoc example for httprequest
- [ ] Add cloud.google.com example in advanced-write snippet

PR #3097, opened 2020-10-27
docs(logging): add example for logging the HTTPRequest type

nicoleczhu:
fixes: #3074 

Rubric: (go/cdpe-snippet-rubric)


logging/examples_test.go
```diff
@@ -86,6 +87,23 @@ func ExampleClient_Logger() {
 	_ = lg // TODO: use the Logger.
 }
 
+func ExampleHTTPRequest() {
```
Daniel-Sanche, line 90: should there be a region tag around this?

logging/examples_test.go
```diff
@@ -86,6 +87,23 @@ func ExampleClient_Logger() {
...
+	ctx := context.Background()
+	client, err := logging.NewClient(ctx, "my-project")
+	if err != nil {
+		// TODO: Handle error.
```
Daniel-Sanche, line 94: Are these todos meant to be merged in?

logging/examples_test.go
```diff
@@ -86,6 +87,23 @@ func ExampleClient_Logger() {
...
+	}
+	lg := client.Logger("my-log")
+	httpEntry := logging.Entry{
+		Payload: "optional message",
+		HTTPRequest: &logging.HTTPRequest{
+			// TODO: pass in request
+			Request: &http.Request{},
+		},
+	}
+	lg.Log(httpEntry)
```
Daniel-Sanche, line 104: Can you add a test for this?

logging/examples_test.go
nicoleczhu: No need for region tags. These are godocs examples so its auto appended inline with the type HTTPRequest definition

You can preview how that might look with `godoc -http=:6060`

logging/examples_test.go
nicoleczhu: They are intentional. It looks to be a godoc convention among all of the godoc examples

logging/examples_test.go
nicoleczhu: The example is the test and gets compiled. 
In this case, I dont expect outputs otherwise I will denote that with `// output: <output>`

logging/examples_test.go
Daniel-Sanche: It seems a bit strange to maintain samples in two places. I wonder if we could combine them. Maybe make synthtool pull in samples from the `golang-samples`?

That's definitely out of scope for now though!

Daniel-Sanche (approved):

logging/examples_test.go
nicoleczhu: yeah good pt. I am wondering about this as well. IT would require that all languages also include compilable / testable samples in documentation. Interesting problem to solve for sure.

PR #3114, opened 2020-10-29
test(logging): add back DeleteLog test

### LINKED ISSUE #1654: logging: fix TestLogsAndDelete
Too many runs causes us to run out of quota. We've tried and failed to dramatically increase quota (logging team gives a ridiculously low quota of, like, 20 per hour to Delete, and won't budge on it).

We should probably use a project pool.

https://sponge.corp.google.com/target?id=35806e80-802d-4ab2-8c8a-006e70d5b9aa&target=cloud-go-libraries/google-cloud-go/continuous/go111&searchFor=&show=ALL&sortBy=STATUS

```
=== RUN   TestLogsAndDelete
--- FAIL: TestLogsAndDelete (0.08s)
    logging_test.go:497: deleting "GO-LOGGING-CLIENT/TEST-LOG-20191021-67194905680597-0001": rpc error: code = ResourceExhausted desc = Quota exceeded for quota metric 'logging.googleapis.com/control_requests' and limit 'ControlRequestsPerMinutePerProject' of service 'logging.googleapis.com' for consumer 'project_number:36639933145'.
```

nicoleczhu:
fixes #1654 

To de-flake:
- Instead of calling Logs() which is flakey, we write a new log entry then do a single deletion
- Also verified with BE that the quota for all logadmin api calls is 600/min and 1k/hour

It might still flake for a new reason since DeleteLogs is getting deprecated and may be inherently flakey. I will monitor it.


logging/logging_test.go
```diff
@@ -473,38 +473,27 @@ func TestPing(t *testing.T) {
...
-				}
-			} else {
-				nDeleted++
-			}
+	initLogs()
+	c, a := newClients(ctx, testProjectID)
+	defer c.Close()
+	defer a.Close()
+	lg := c.Logger(testLogID)
+	err := lg.LogSync(ctx, logging.Entry{Payload: "hello"})
```
codyoss, line 483: nit: combine with if check below `if err := lg.LogSync(ctx, logging.Entry{Payload: "hello"}); err != nil { ...}`

codyoss (approved):
LGTM, with a small nit. Thanks.

logging/logging_test.go
nicoleczhu: noted for future , thanks!

PR #3116, opened 2020-10-30
feat(bigtable): run E2E test over DirectPath

mohanli-ml:
In the relevant GAX library PR https://github.com/googleapis/google-api-go-client/pull/732, we removed environment variable GOOGLE_CLOUD_ENABLE_DIRECT_PATH, and expose an internal option to Yoshi library, so that Yoshi library can decide if it want to attempt DirectPath or not. In this PR, 
we make use of this option to run E2E test over DirectPath. Relevant bug is b/172700854.


codyoss:
Let's tackle the other PR first. This won't work until that is in, released and bumped the dependency here.

tritone (commented):
+1 to Cody's comment. Also FYI @kolea2

codyoss:
@ mohanli-ml I did release a new version of `google.golang.org/api` if you want to bump the dependency here the code should now compile.

bigtable/bigtable.go
```diff
@@ -81,6 +83,13 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
 		// TODO(grpc/grpc-go#1388) using connection pool without WithBlock
 		// can cause RPCs to fail randomly. We can delete this after the issue is fixed.
 		option.WithGRPCDialOption(grpc.WithBlock()))
+
+	// Expose an experimental flag to make client attempt DirectPath.
+	// Once e2e tests are finished, this flag will be removed, and client
```
tbpg, line 88: What does "Once e2e tests are finished" mean? Seems like we should be able to add this option from the test directly, rather than through an environment variable?

As-is, this seems to be doing more than adding a test, given this is a non-test file. In general, we should avoid adding to the public API when we know we're going to remove it in the future.

bigtable/integration_test.go
```diff
@@ -2212,3 +2235,28 @@ func deleteTable(ctx context.Context, t *testing.T, ac *AdminClient, name string
 		t.Logf("DeleteTable: %v", err)
 	}
 }
+
+func verifyDirectPathRemoteAddress(testEnv IntegrationEnv, t *testing.T) {
+	if !testEnv.Config().AttemptDirectPath {
```
tbpg, line 2240: Call `t.Helper()` to get better line numbers when this fails.

bigtable/integration_test.go
```diff
@@ -2212,3 +2235,28 @@ func deleteTable(ctx context.Context, t *testing.T, ac *AdminClient, name string
...
+		return
+	}
+	if remoteIp, res := isDirectPathRemoteAddress(testEnv); !res {
```
tbpg, line 2243: s/Ip/IP/g

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tbpg (changes requested):

bigtable/bigtable.go
mohanli-ml: We are going to release directpath for bigtable, and we designed some e2e test scenarios that needs to be covered before the release (See the background and what we have done for java-bigtable in go/directpath-e2e-test-scenarios). I replace the environment with a global variable, but the idea should be the same.

We need this option now because we do not want to rashly set `internaloption.EnableDirectPath` to true, which is what we will do after all e2e tests. A similar PR for previous work in java-bigtable can be found https://github.com/googleapis/java-bigtable/pull/467

bigtable/integration_test.go
mohanli-ml: Done. This is a good point. Thanks!

bigtable/integration_test.go
mohanli-ml: Done. Thanks for pointing out!

bigtable/bigtable.go
tbpg: Sorry, I'm still confused. Can we avoid adding anything to the exported API? Can the test pass the `internaloption.EnableDirectPath` option when appropriate?

mohanli-ml: Hey Tyler, sorry I did not make myself clear. So DirectPath is a new feature for bigtable, and ultimately we will need to add `internaloption.EnableDirectPath(true)` to the library code to support this feature. For now, we are using a flag that is passed from tests in `internaloption.EnableDirectPath()` because we needs to test this feature first. But since ultimately we will change the library code, I think it is reasonable to add the internal option now. Does this make sense to you? Please let me know if you have more questions. Thanks!

mohanli-ml: I moved the internal option to test files, and the flag is removed.

bigtable/export_test.go
```diff
@@ -71,6 +77,7 @@ type IntegrationEnv interface {
 	NewInstanceAdminClient() (*InstanceAdminClient, error)
 	NewClient() (*Client, error)
 	Close()
+	GetPeer() *peer.Peer
```
tbpg, line 80: This can just be called `Peer`. See https://golang.org/doc/effective_go.html#Getters.

tbpg (approved):

tbpg:
Also, no need to force push. We'll squash into a single commit with the right message when merging.

bigtable/export_test.go
mohanli-ml: Fixed. Thanks!

kolea2 (approved):

tritone (approved):

PR #3118, opened 2020-10-30
docs(logging): Require HTTPRequest.Status when grouping request logs

Hexcles:
This field must be set for Cloud Log Viewer to render request logs correctly.


nicoleczhu:
@Hexcles Thanks for the PR! 

The Status field doesn't have to be set for Logs Viewer to correctly render HTTPRequest type logs. Or did you mean more from a UX perspective?

nicoleczhu (approved):
LGTM. Thanks Robert!

I'll update the Devsite samples shortly as well

Issue #3088, opened 2020-10-26
logging: logAdmin listLogEntries to filter by last 24 hours by default


nicoleczhu:
**Proposal:** 
Similar to Python client lib, we should consider defaulting ListLogEntry filter to search only past 24 hours. Otherwise, the call takes ~5 minutes by default. 
 
Rationale: 
- **24H is the default timestamp filter for gcloud CLI.** 
- It [confuses users](https://github.com/googleapis/python-logging/issues/68)
- This leads to 15-37 minutes delays in tests

**Counter proposal:**
Be minimalistic in Go library. Thoroughly document the fact that this call takes 5 minutes in all user docs, not just here:
https://cloud.google.com/logging/docs/view/advanced-queries#finding-quickly

PR #3120, opened 2020-10-31
fix(logging): Entries has a 24H default filter

nicoleczhu:
Would like to push this through as a minor feat, and a nonbreaking change. User's existing code will actually not be broken by the PR. Users will simply retrieve less logs upon running listEntries. Previously, users would have hit API quota anyway. This is more a bug fix than a new feature

fixes: #3088


codyoss (changes requested):
I have a couple of questions about these changes, raised internally.

nicoleczhu:
After discussion: we'll batch this change with future improvements

logging/logadmin/logadmin.go
```diff
@@ -243,9 +243,24 @@ func listLogEntriesRequest(parent string, opts []EntriesOption) *logpb.ListLogEn
 	for _, opt := range opts {
 		opt.set(req)
 	}
+	req.Filter = defaultTimestamp(req.Filter)
 	return req
 }
 
+func defaultTimestamp(filter string) string {
```
simonz130, line 250: document this method

logging/logadmin/logadmin.go
```diff
@@ -243,9 +243,24 @@ func listLogEntriesRequest(parent string, opts []EntriesOption) *logpb.ListLogEn
...
+	dayAgo := time.Now().Add(-24 * time.Hour).UTC()
```
simonz130, line 251: Confirming - is the semantic in gcloud to look at the last 24 hours in UTC time zone or current time zone?

nicoleczhu: It should be current timezone to avoid unforeseen edge cases.  I'll leave this logic as is

logging/logadmin/logadmin.go
nicoleczhu: Sounds good. documenting

codyoss (approved):
Seems reasonable and a more desirable behavior. It looks like this was also considered a non-breaking change in the node libraries.

PR #3123, opened 2020-11-02
feat(pubsublite): Move internal implementation details to internal/wire subpackage

tmdiep:
This makes it possible for the bulk of the Lite library internal implementation to be used from different packages. Some cleanups so that the wire package compiles.

The deletion of pubsublite.Message was deliberate and the file was renamed to message_router.go.


hongalex (approved):

tmdiep:
Thanks Alex. Merging..

PR #3128, opened 2020-11-02
fix(spanner): do not end span with iterator.Done

### LINKED ISSUE #3110: spanner: tracing error "no more items in iterator" on iter.Do
**Client**

Spanner (v1.10.0)

**Code**

The following code is a code snippet from [spanner#read.go](https://github.com/googleapis/google-cloud-go/blob/master/spanner/read.go#L191), and the explanation of the problem comes after the code.

```go
func (r *RowIterator) Do(f func(r *Row) error) error {
	defer r.Stop()
	for {
		row, err := r.Next()
		switch err {
		case iterator.Done:
			return nil
		case nil:
			if err = f(row); err != nil {
				return err
			}
		default:
			return err
		}
	}
}

func (r *RowIterator) Stop() {
	if r.streamd != nil {
		defer trace.EndSpan(r.streamd.ctx, r.err)
	}
	if r.cancel != nil {
		r.cancel()
	}
	if r.release != nil {
		r.release(r.err)
		if r.err == nil {
			r.err = spannerErrorf(codes.FailedPrecondition, "Next called after Stop")
		}
		r.release = nil
	}
}
```

When using [spanner#RowIterator.Do](https://godoc.org/cloud.google.com/go/spanner#RowIterator.Do), when `err` equals `iterator.Done`, then `r.Stop ()` is called. In the `Stop()` function, `err` is being traced with "no more items in iterator". Is this really correct? This trace is hindering the visualization of possible real errors that may be happening.

**Expected behavior**

I think the error "no more items in iterator" should be ignored in `trace.EndSpan`.

**Actual behavior**

The "no more items in iterator" error is overflowing the opencensus trace.

**Screenshots**

Look the error rate of at 99.8%. All of this errors is `UNKNOWN: no more items in iterator`.
image

olavloite:
Spans should not be ended with iterator.Done as an error. This is not an actual error, but an indication that the iterator reached its end. Setting this as an error on a span pollutes the traces and makes it harder to find actual errors that might occur.

Fixes #3110

### DIFF
--- a/spanner/client_test.go
+++ b/spanner/client_test.go
@@ -2183,3 +2183,59 @@ func queryOptionsTestCases() []QueryOptionsTestCase {
 		},
 	}
 }
+
+func TestClient_DoForEachRow_ShouldNotEndSpanWithIteratorDoneError(t *testing.T) {
+	// This test cannot be parallel, as the TestExporter does not support that.
+	te := itestutil.NewTestExporter()
+	defer te.Unregister()
+	_, client, teardown := setupMockedTestServer(t)
+	defer teardown()
+
+	iter := client.Single().Query(context.Background(), NewStatement(SelectSingerIDAlbumIDAlbumTitleFromAlbums))
+	iter.Do(func(r *Row) error {
+		return nil
+	})
+	select {
+	case <-te.Stats:
+	case <-time.After(1 * time.Second):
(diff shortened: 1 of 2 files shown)


hengfengli (approved):
LGTM. 👍

PR #3132, opened 2020-11-03
feat(pubsublite): Flow controller and offset tracker for the subscriber

tmdiep:
Ports the flow control token counter, batcher and offset tracker from the Pub/Sub Lite Java client library.


tmdiep:
WANT_LGTM=@manuelmenzella-google 
Other reviewers optional/FYI. Feel free to leave readability comments.

These types are almost identical to their Java counterparts and similarly named for easier cross-language code changes in the future:

* [TokenCounter](https://github.com/googleapis/java-pubsublite/blob/master/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/TokenCounter.java)
* [FlowControlBatcher](https://github.com/googleapis/java-pubsublite/blob/master/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/FlowControlBatcher.java)
* [NextOffsetTracker](https://github.com/googleapis/java-pubsublite/blob/master/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/internal/wire/NextOffsetTracker.java)

pubsublite/internal/wire/flow_control.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// A TokenCounter stores the amount of outstanding byte and message flow control
+// tokens that the client believes exists for the stream.
+type tokenCounter struct {
+	Bytes    int64
+	Messages int64
+}
+
+func saturatedAdd(sum, delta int64) int64 {
```
manuelmenzella-google, line 42: Is it worth using big.Int instead?

manuelmenzella-google (approved):

pubsublite/internal/wire/flow_control.go
tmdiep: Hmm, I think here we just want to cap overflow, just in case. But the scenario should never occur because the server would be sending back more bytes/messages than the user configured in flow control settings.

PR #3133, opened 2020-11-03
feat(spanner): export ToSpannerError

### LINKED ISSUE #3122: spanner: new method for declaring spanner.Error
In spanner, Code field in Error struct is deprecated and spanner recommends using ErrCode method for extracting the err code.
But we don't have a method for declaring it(ex. NewError()). we can't declare spanner.Error struct without Error.Code field in unit tests. 

I want a new method for declaration of spanner.Error struct.
ex. cloud.google.com/go/spanner/errors.go
```
package spanner

import ...

type Error struct {
. . .
}

// something new below
func NewError(err error) *Error {
 . . .
}
```

spanner user can declare spanner.Error struct in unit tests without deprecated spanner.Error.Code field.

olavloite:
The `ToSpannerError` method should be exported to allow users to create Spanner errors to use with unit tests. Currently, that was only possible by creating a struct directly and filling the deprecated `Code` field. That is however no longer recommended, as the `Code` field might be removed in a future release.

Fixes #3122


spanner/errors.go
```diff
@@ -115,8 +115,15 @@ func spannerErrorf(code codes.Code, format string, args ...interface{}) error {
...
-// toSpannerError converts general Go error to *spanner.Error.
-func toSpannerError(err error) error {
+// ToSpannerError converts a general Go error to *spanner.Error. If the given
+// error is already a *spanner.Error, the original error will be returned.
+//
+// Spanner Errors are normally created by the Spanner client library from the
+// returned status of a RPC. This method can also be used to create Spanner
+// errors for use in tests. The recommended way to create test errors is
+// calling this method with a status error, e.g.
+// ToSpannerError(status.Newf(codes.NotFound, "Table not found").Err())
```
hengfengli, line 125: nits: `status.Newf` -> `status.New`? It seems that formatting is not used here.

hengfengli (approved):
LGTM.

spanner/errors.go
olavloite: Good point, done.

PR #3135, opened 2020-11-03
feat(datastore): accept pb.Value in interfaceToProto

### LINKED ISSUE #3078: datastore: allow users to pass in a googleapis.datastore.v1.Value to filters
We have several different entities which are indexed on the deprecated [UserProperty](https://cloud.google.com/appengine/docs/standard/python/datastore/typesandpropertyclasses#UserProperty). The Datastore V1 API allows us to perform a query on this property by setting the [meaning](https://cloud.google.com/datastore/docs/reference/data/rpc/google.datastore.v1#value) of the Value in the PropertyFilter to 20 (`MEANING_PREDEFINED_ENTITY_USER`), indicating that the entity value isn't some arbitrary entity but is one that the API knows how to handle.

Presently the Datastore client library doesn't have any way to specify the `meaning` of the filter value, so we cannot express the queries we need for these entities.

I believe the simplest solution would be to allow library users to pass in a [Value](https://pkg.go.dev/google.golang.org/genproto/googleapis/datastore/v1#Value) that they have constructed themselves when calling [Filter](https://github.com/googleapis/google-cloud-go/blob/846989f5b66cd1dfafc6348894029c0cbc5415cd/datastore/query.go#L181) on a `Query`. This could be achieved by checking if the input to the [interfaceToProto](https://github.com/googleapis/google-cloud-go/blob/846989f5b66cd1dfafc6348894029c0cbc5415cd/datastore/save.go#L374) function is already a `Value`, and if so passing it through unchanged. In this way, users who need to specify the `meaning` can do so, but the API is otherwise unaltered.

We could backfill these entities to instead be indexed on a different property, but in our case the number of entities is 2e10 so this would likely be prohibitively expensive. Our other option would be to fork the Datastore library and add in the proposed change.

corajr:
Modifies the `interfaceToProto` function to allow end users to pass in an existing protobuf `Value`. This is to support querying for entities that need the `meaning` field to be set.

Resolves #3078.

### DIFF
--- a/datastore/save.go
+++ b/datastore/save.go
@@ -427,6 +427,12 @@ func interfaceToProto(iv interface{}, noIndex bool) (*pb.Value, error) {
 			return nil, err
 		}
 		val.ValueType = &pb.Value_EntityValue{EntityValue: e}
+	case *pb.Value:
+		if v != nil {
+			return v, nil
+		}
+		val.ValueType = &pb.Value_NullValue{}
+		return val, nil
 	case []interface{}:
 		arr := make([]*pb.Value, 0, len(v))
 		for i, v := range v {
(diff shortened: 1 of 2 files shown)


corajr:
@googlebot I signed it!

PR #3141, opened 2020-11-04
docs: remove redundant step in submodule RELEASING.md guidance

shollyman:
The cited `go generate` is responsible for updating the repo reference
in internal/version.go, so listing it both explicitly as a manual edit
and a generate command seems redundant.


codyoss (approved):
Thanks!

PR #3143, opened 2020-11-05
feat(pubsublite): Abstraction for leaf and composite services

tmdiep:
Pub/Sub Lite publisher and subscriber clients will consist of a hierarchy of "services".

* abstractService will be embedded directly by the gRPC stream wrappers.
* compositeService manages multiple child services. For example, the publisher client publishes messages to multiple topic partitions (1 stream per partition).
* A compositeService can itself be a child of another compositeService. For example, the subscriber client is a tree 3 levels deep. A subscriber for a single partition consists of a commit and subscribe stream. The client may be receiving messages for multiple partitions.
* A compositeService can dynamically add and remove child services. For example, the subscriber client (in assignment mode) receives partition assignments from the server and needs to dynamically update the partitions it is receiving messages for.


tmdiep:
Please see the description and [subscriber prototype](https://github.com/tmdiep/google-cloud-go/blob/working/pubsublite/internal/wire/subscriber.go) for why this is necessarily complex :/

pubsublite/internal/wire/service_test.go
```diff
@@ -0,0 +1,519 @@
...
+	t.Run("Starting", func(t *testing.T) {
+		parent.Start()
+
+		child1.receiver.VerifyStatus(t, serviceStarting)
+		child2.receiver.VerifyStatus(t, serviceStarting)
+		parent.receiver.VerifyStatus(t, serviceStarting)
+
+		// child3 is added after Start() and should be automatically started.
+		parent.AddServices(child3)
+		child3.receiver.VerifyStatus(t, serviceStarting)
```
hongalex, line 245: Perhaps check that the status of child3 is `serviceUnitialized` prior to adding to the parent since I noticed that case hasn't been covered in other tests

pubsublite/internal/wire/service.go
```diff
@@ -0,0 +1,344 @@
...
+		return nil
+	}
+}
+
+// unsafeUpdateStatus assumes the service is already holding a mutex when
+// called, as it often needs to be atomic with other operations.
+func (as *abstractService) unsafeUpdateStatus(targetStatus serviceStatus, err error) bool {
+	if as.status >= targetStatus {
+		// Already at the same or later stage of the service lifecycle.
+		return false
```
hongalex, line 140: Should `as.err` not be updated in this situation?

pubsublite/internal/wire/service.go
```diff
@@ -0,0 +1,344 @@
...
+	RemoveStatusChangeReceiver(serviceHandle)
+	Handle() serviceHandle
+	Error() error
+}
+
+// abstractService can be embedded into other structs to provide common
+// functionality for managing service status and status change receivers.
+type abstractService struct {
+	mu sync.Mutex
+
```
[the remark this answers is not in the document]
tmdiep: Done

pubsublite/internal/wire/service_test.go
tmdiep: Good point. Done.

pubsublite/internal/wire/service.go
tmdiep: Yes, the errors for the obsolete statuses are ignored (only `as.status == targetStatus` actually occurs). Errors are really only set for serviceTerminating or serviceTerminated and we try not to clobber the first error (below).

hongalex (approved):

PR #3152, opened 2020-11-06
feat(pubsublite): Periodic background task

tmdiep:
Introduces periodicTask to manage a recurring background task that can be paused and resumed.

### DIFF
--- a/pubsublite/internal/wire/periodic_task.go
+++ b/pubsublite/internal/wire/periodic_task.go
@@ -19,54 +19,51 @@ import (
 
 // periodicTask manages a recurring background task.
 type periodicTask struct {
-	period  time.Duration
-	ticker  *time.Ticker
-	stop    chan struct{}
-	stopped bool
-	task    func()
+	period time.Duration
+	task   func()
+	ticker *time.Ticker
+	stop   chan struct{}
 }
 
 func newPeriodicTask(period time.Duration, task func()) *periodicTask {
 	return &periodicTask{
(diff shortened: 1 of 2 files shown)


pubsublite/internal/wire/periodic_task.go
```diff
@@ -0,0 +1,72 @@
...
+		ticker: time.NewTicker(period),
+		stop:   make(chan struct{}),
+		period: period,
+		task:   task,
+	}
+}
+
+// Start the polling goroutine.
+func (pt *periodicTask) Start() {
+	go pt.poll()
```
hongalex, line 40: Does it matter that `Start` can be called multiple times?

tmdiep: Yeah, I considered handling this, but didn't think it was strictly necessary. The callers of this need to prevent multiple Starts anyway (along with other startup operations).

Added a comment in case future usage doesn't guard against duplicate starts.

tmdiep: Obsolete

tmdiep:
Ugh, sorry. I failed to notice that Ticker.Reset is not supported before Go 1.15, so had to reduce the interface to just Start/Stop. This also ended up resolving your comment about duplicate starts.

PTAL, thanks!

hongalex (approved):

hongalex:
Ah yeah good catch. I thought it was just tests being flaky

PR #3154, opened 2020-11-06
feat(pubsub): Enable server side flow control by default with the option to turn it off

fayssalmartanigcp:
This change enables sending flow control settings automatically to the server.
If ReceiveSettings.MaxOutstandingMessages > 0 or ReceiveSettings.MaxOutstandingBytes > 0,
flow control will be enforced at the server side (in addition to the client side).

This behavior is enabled by default and ReceiveSettings.UseLegacyFlowControl can be set for users
who would like to opt-out of this feature in case they encounter issues with server side flow
control.


codyoss:
@fayssalmartanigcp Thank you for the contribution. Is there an issue associated with this PR, if not would you mind opening one?

pubsub/subscription.go
```diff
@@ -489,6 +489,12 @@ type ReceiveSettings struct {
 	// for unprocessed messages.
 	MaxOutstandingBytes int
 
+	// if UseLegacyFlowControl is true, then enforcing flow control settings at
```
hongalex, line 492: Exported fields in Go should start with the field name, e.g. "UseLegacyFlowControl disables flow control settings on the Cloud Pub/Sub server and ...."

fayssalmartanigcp: done.

hongalex (approved):

pubsub/iterator.go
```diff
@@ -76,7 +76,13 @@ type messageIterator struct {
 func newMessageIterator(subc *vkit.SubscriberClient, subName string, maxExtensionPeriod *time.Duration, po *pullOptions) *messageIterator {
 	var ps *pullStream
 	if !po.synchronous {
-		ps = newPullStream(context.Background(), subc.StreamingPull, subName, po.maxOutstandingMessages, po.maxOutstandingBytes)
+		maxMessages = po.maxOutstandingMessages
+		maxBytes = po.maxOutstandingBytes
+		if po.UseLegacyFlowControl {
```
hongalex, line 81: CI is currently failing because this needs to be lower case

```suggestion
		if po.useLegacyFlowControl {
```

pubsub/iterator.go
```diff
@@ -76,7 +76,13 @@ type messageIterator struct {
 func newMessageIterator(subc *vkit.SubscriberClient, subName string, maxExtensionPeriod *time.Duration, po *pullOptions) *messageIterator {
 	var ps *pullStream
 	if !po.synchronous {
-		ps = newPullStream(context.Background(), subc.StreamingPull, subName, po.maxOutstandingMessages, po.maxOutstandingBytes)
+		maxMessages = po.maxOutstandingMessages
+		maxBytes = po.maxOutstandingBytes
```
hongalex, line 80:
```suggestion
		maxMessages := po.maxOutstandingMessages
		maxBytes := po.maxOutstandingBytes
```

Issue #3158, opened 2020-11-07
profiler: pollProfilerService consumes 100% CPU when TLS validation fails


bitglue:
**Client**

profiler

**Environment**

Debian buster, in Docker, in AWS

**Go Environment**

```
root@82812248ba78:/go/src/github.com/remind101/r101-experiments# go version
go version go1.14.9 linux/amd64
root@82812248ba78:/go/src/github.com/remind101/r101-experiments# go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/root/.cache/go-build"
GOENV="/root/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/go/src/github.com/remind101/r101-experiments/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build695409103=/tmp/go-build -gno-record-gcc-switches"
```

**Code**

e.g.
```go
	cfg := profiler.Config{
￼		Service:        os.Getenv("EMPIRE_APPNAME") + "." + os.Getenv("EMPIRE_PROCESS"),
￼		ServiceVersion: os.Getenv("EMPIRE_RELEASE"),
￼		ProjectID:      projectID,
￼	}
￼
￼	return profiler.Start(cfg)
```

**Expected behavior**

profiling has negligible impact on performance

**Actual behavior**

application is 1300% slower

**Screenshots**

I was able to manually grab a cpu profile from a pprof http endpoint, which reveals the program is looping with great vigor in profiler.pollProfilerService():

```
(pprof) top
[...]

PR #3159, opened 2020-11-07
fix(profiler): avoid 100% CPU usage when missing certs

bitglue:
The previous implementation relied on .createProfile() sleeping a bit,
which in turn relies on gax.Invoke's retry behavior and the initial
backoff in the retryer used. This works most times, except gax.Invoke
has logic to ignore the retryer and immediately fail when the error is
certificate validation. Consequently, .createProfile() returns
immediately, and pollProfilerService() will consume 100% CPU.

To avoid any possibility of the endless loop in pollProfilerService()
from consuming 100% CPU, a ticker is added. This addresses this
particular instance, and also any unanticipated or future cases that
might arise with changes in gax, etc.

It also seemed like a good idea to exit pollProfilerService() when the
context is cancelled. In the current implementation the context is
always ctx.Background() so I don't see how this could happen, but you
never know how things might change in the future.

Fixes https://github.com/googleapis/google-cloud-go/issues/3158


profiler/profiler.go
```diff
@@ -603,9 +605,20 @@ func initializeConfig(cfg Config) error {
 func pollProfilerService(ctx context.Context, a *agent) {
 	debugLog("Cloud Profiler Go Agent version: %s", version.Repo)
 	debugLog("profiler has started")
+	defer debugLog("profiler has stopped")
+
+	ticker := time.NewTicker(time.Minute)
```
kalyanac, line 610: The agent can sometimes be asked by the backend to backoff for a long duration. The ticker will keep generating ticks even the agent is in backoff. This is a waste of resources.

Additionally, in case where a single task is being profileed, the same agent will be asked to collect multiple profile types in a duration of one minute. This approach will force the agent to wait one minute before calling createProfile again, which is not expected.

bitglue: > This is a waste of resources

The overhead of generating a tick once per minute is, as far as I can tell, miniscule. Admittedly I am not an expert on the implementation of the ticker or go's internals, so if you think some other mechanism for guaranteeing a maximum iteration rate is more appropriate I'm open to suggestions.

I do think it's critical, however, to have _something_ in `pollProfilerService()` which very simply (without relying on behavior of dependencies or remote APIs) can be shown to have a reasonable iteration rate under any circumstances.

> Additionally, in case where a single task is being profileed, the same agent will be asked to collect multiple profile types in a duration of one minute.

So let's reduce `minimumIteration` to a smaller value? What is the maximum reasonable rate of profile requests? Surely, it is a finite number, with great latitude between "a reasonable rate" and "so fast the CPU usage is unacceptable".

PR #3160, opened 2020-11-08
feat(pubsublite): Refactoring and unit tests for retryableStream

tmdiep:
Removed an unnecessary goroutine in retryableStream for simplification. Accommodated the partition assignment stream, which does not return an initial response.

Added dedicated unit tests for retryableStream.

Refactored the mock lite server to eliminate test flakiness:
- Server streams can run briefly after the test has already ended and results in a panic when testing.T is accessed in the verifiers. Moved all RPC verifiers to a Verifiers object and ensure it is flushed at the end of the test.
- When the client terminates a stream too quickly, the server never receives a request, which makes tests non-deterministic. Introduced a Barrier to synchronize client and server.


hongalex (approved):
Apologies review for this took so long. LGTM

tmdiep:
No problem at all. Thanks for reviewing!

Issue #2476, opened 2020-06-17
storage: when using an emulator, it is not possible to use the same Client object for both uploading and other operations


justinruggles:
**Client**

Storage

**Environment**

MacOS 10.14.6

**Go Environment**

```sh
$ go version
go version go1.14.1 darwin/amd64
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/REDACTED/Library/Caches/go-build"
GOENV="/Users/REDACTED/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GONOPROXY="REDACTED"
GONOSUMDB="REDACTED"
GOOS="darwin"
GOPATH="/Users/REDACTED/go"
GOPRIVATE="REDACTED"
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="REDACTED/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/03/wp97sc8538b75zs4q619t5x00000gn/T/go-build425541068=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

**Code**

```go
package main

import (
	"context"
	"encoding/json"
	"flag"
	"io"
	"log"
	"os"
	"path/filepath"

	"cloud.google.com/go/storage"
	"google.golang.org/api/option"
)

const TestBucket = "test-bucket"
const TestProject = "test-project"

func uploadFile(ctx context.Context, filename string, project string, bucket string, object string) error {
	os.Setenv("STORAGE_EMULATOR_HOST", "localhost:8080")
	defer os.Unsetenv("STORAGE_EMULATOR_HOST")

[...]

PR #3172, opened 2020-11-09
fix(storage): fix endpoint selection logic

andyrzhao:
The hand-written storage.go client need to be updated to work with DCA (https://google.aip.dev/auth/4114).
In particular, we need to update endpoint selection logic taking into account of the custom http client used by reader.go, the rawService, as well as the STORAGE_EMULATOR_HOST override.


andyrzhao:
To help understand the non-trivial endpoint selection logic, refer to the doc at: https://docs.google.com/document/d/1FmjwOW0g0y3QPUY5PQlQ4H2A0_ShPm3YjmdgOAj40zM/edit

storage/storage.go
```diff
@@ -105,41 +106,48 @@ type Client struct {
...
 	if err != nil {
 		return nil, fmt.Errorf("dialing: %v", err)
 	}
-	rawService, err := raw.NewService(ctx, option.WithHTTPClient(hc))
+	// RawService should be created with the chosen endpoint to take account of user override.
+	rawService, err := raw.NewService(ctx, option.WithEndpoint(ep), option.WithHTTPClient(hc))
 	if err != nil {
 		return nil, fmt.Errorf("storage client: %v", err)
 	}
-	if ep == "" {
```
tritone, line 129: So if I'm understanding this correctly, we no longer need the `ep == ""` case since we can always expect a valid endpoint to be returned given that we are passing in WithEndpoint to NewService. Is that correct?

storage/storage.go
```diff
@@ -105,41 +106,48 @@ type Client struct {
 func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error) {
 	var host, readHost, scheme string
 
+	// In general, it is recommended to use NewService instead of NewClient, since NewService
```
tritone, line 109: I guess the recommended path is to call NewService without WithHTTPClient, which avoids the need for NewClient?

tritone (commented):
Generally this looks good. Spoke with @andyrzhao offline and he will add tests. I _think_ this should fix #2476 .

storage/storage.go
```diff
@@ -105,41 +106,48 @@ type Client struct {
...
+	// configures the correct default endpoints when initializing the internal http client.
+	// However, in our case, reader.go needs to access the client directly, so the client
```
tbpg, line 111: Anything more specific than reader.go? Type/function name?

storage/storage.go
tbpg, line 109: Make this `raw.NewService` etc. to make it clear what's being referred to here?

storage/storage.go
```diff
@@ -105,41 +106,48 @@ type Client struct {
...
+	// the default endpoints.
 	if host = os.Getenv("STORAGE_EMULATOR_HOST"); host == "" {
 		scheme = "https"
 		readHost = "storage.googleapis.com"
 
 		// Prepend default options to avoid overriding options passed by the user.
 		opts = append([]option.ClientOption{option.WithScopes(ScopeFullControl), option.WithUserAgent(userAgent)}, opts...)
+
+		opts = append(opts, internaloption.WithDefaultEndpoint("https://storage.googleapis.com/storage/v1/"))
+		opts = append(opts, internaloption.WithDefaultMTLSEndpoint("https://storage.mtls.googleapis.com/storage/v1/"))
```
tbpg, line 124: `internaloption` API question: why do we have `WithDefault...` rather than prepending the defaults like we do above (see "Prepend default options" comment)? Should have asked this when updating `internaloption`, but I didn't think of it.

storage/storage.go
```diff
@@ -105,41 +106,48 @@ type Client struct {
...
-		rawService.BasePath = ep
-		u, err := url.Parse(ep)
-		if err != nil {
-			return nil, fmt.Errorf("supplied endpoint %v is not valid: %v", ep, err)
-		}
-		readHost = u.Host
+	// Update readHost with the chosen endpoint.
+	u, err := url.Parse(ep)
+	if err != nil {
+		return nil, fmt.Errorf("supplied endpoint %v is not valid: %v", ep, err)
```
tbpg, line 148:
```suggestion
		return nil, fmt.Errorf("supplied endpoint %q is not valid: %v", ep, err)
```

storage/storage.go
tbpg, line 109: Any tests to update? Should there be?

tbpg (changes requested):

storage/storage.go
andyrzhao: Ah yea, the reason why "WithDefaultEndpoint" was introduced is that we needed a way to distinguish between user specified override and auto-generated default endpoint. In other words, with the "Prepend" technique, downstream code has no way to tell whether someone wanted to override the endpoint explicitly, since it's all coming from a "WithEndpoint". This distinction is needed to support DCA, where we automatically upgrade the default endpoint to defaultMTLSendpoint, unless explicitly overridden by the user. Therefore, base clients should almost always use WithDefaultEndpoint. (Note that this PR sets WithEndpoint for NewService because the endpoint has already been computed - the alternative would be to pass in the full set of ClientOptions.)

storage/storage.go
andyrzhao: Yes, I'm looking to add unit test coverage for the matrix in the doc I linked.

storage/storage.go
andyrzhao: the endpoint is computed by htransport, which will now always return a non-empty endpoint after this change, since we are setting the defaults. Passing "WithEndpoint" to NewService is just a short-cut since the endpoint has already been computed (the alternative would be to pass the full set of user-specified ClientOptions to NewService, which would end up recomputing the endpoint.)

storage/storage.go
andyrzhao: Yes, that's right. Maybe we can consider exporting the internal http client from NewService in the future so it can support the direct-access, client reuse scenario found in reader.go.

storage/storage.go
tbpg: It seems like we should do that directly, rather than indirectly through "default" vs not. So, we could add another setting that's something like `attemptDCA` with a default of true. If the user specifies an option that means we shouldn't attempt DCA, set `attemptDCA=false`.

Right now, the logic you laid out (which makes sense) is all implicit, making this more difficult to understand. I could totally be missing something, though.

andyrzhao: So DCA aside, I'd argue that having the client options "WithDefaultEndpoint" and "WithDefaultMTLSEndpoint" actually makes the client intentions more explicit and less error prone, and we no longer have to rely on endpoint options being propagated in the right order, which is made even more troublesome by hand-written clients. Having the defaults as separate options also allows us to support other use-cases, such as "endpoint merging". For example, cbro added the following logic in https://github.com/googleapis/google-api-go-client/blob/master/transport/internal/dca/dca.go a while back:

// If the endpoint override is an address (host:port) rather than full base
// URL (ex. https://...), then the user-provided address will be merged into
// the default endpoint. For example, WithEndpoint("myhost:8000") and
// WithDefaultEndpoint("https://foo.com/bar/baz") will return "https://myhost:8080/bar/baz"

And circling back to your original comment about DCA, I agree the logic feels a bit like a black box today, but it is by design, as we eventually want to make it the default behavior that happens without any additional user configuration. You can read more about the DCA spec at https://google.aip.dev/auth/4114, including the env variables we already support today to control the behavior. Thanks!

storage/storage.go
andyrzhao: done.

storage/storage.go
andyrzhao: done.

andyrzhao: Added tests.

storage/storage.go
andyrzhao: done.

tbpg (approved):

tritone (approved):

Issue #3162, opened 2020-11-09
spanner/spansql: support for Update in parseDMLStmt


lazeratops:
**Is your feature request related to a problem? Please describe.**

Currently `spannertest` seems to be unusable with `UPDATE` SQL queries. `parseDMLStmt()` in `spansql/parser.go` seems to only support `DELETE`. There is a `TODO: Insert, Update` comment, but I was unable to find an associated feature request issue for this and not sure if it is planned, hence this request to put it on the roadmap. Relevant line: https://github.com/googleapis/google-cloud-go/blob/master/spanner/spansql/parser.go#L1284

**Describe the solution you'd like**

I would like for `UPDATE` DML to be parsed via `spansql` , which would enable us to properly test these statements via `spannertest`. Ideally `INSERT` would be parsed as well, but as `UPDATE` is more relevant for our immediate purpose I am focusing on that one here.

**Describe alternatives you've considered**

An alternative would be to update via an Update mutation instead. We will likely go with this solution for the time being, but being forced to change the implementation to a mutation solely to account for missing test functionality does not seem ideal.

Thanks!

dsymonds:
Syntax types and parsing for `UPDATE` DML is now done in the spansql package. I'll be working on adding its implementation in spannertest next.

dsymonds:
Implementation is out for review.

You could test it locally real quick by running commands like these in your repo:
```
go mod edit -replace cloud.google.com/go/spanner=github.com/dsymonds/google-cloud-go/spanner@spannertest-update-dml
go mod download
go mod tidy
go test ./...
```

and chime in on #3201 if there's something significantly wrong for your use case.

PR #3192, opened 2020-11-10
feat(spanner/spansql): define structures and parse UPDATE DML statements

dsymonds:
Updates #3162.

### DIFF
--- a/spanner/spansql/parser.go
+++ b/spanner/spansql/parser.go
@@ -1281,7 +1281,13 @@ func (p *parser) parseDMLStmt() (DMLStmt, *parseError) {
 		DELETE [FROM] target_name [[AS] alias]
 		WHERE condition
 
-		TODO: Insert, Update.
+		UPDATE target_name [[AS] alias]
+		SET update_item [, ...]
+		WHERE condition
+
+		update_item: path_expression = expression | path_expression = DEFAULT
+
+		TODO: Insert.
 	*/
 
 	if p.eat("DELETE") {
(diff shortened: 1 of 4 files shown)


spanner/spansql/parser.go
```diff
@@ -1304,9 +1310,64 @@ func (p *parser) parseDMLStmt() (DMLStmt, *parseError) {
...
+		return UpdateItem{}, err
+	}
+	ui := UpdateItem{
+		Column: col,
+	}
+	if err := p.expect("="); err != nil {
+		return UpdateItem{}, err
+	}
+	if p.eat("DEFAULT") {
+		return ui, nil
```
olavloite, line 1362: I think you need a special value for `DEFAULT`, as this implementation will make it impossible to set something to `NULL`.

spanner/spansql/sql_test.go
```diff
@@ -232,6 +232,20 @@ func TestSQL(t *testing.T) {
...
 			reparseDML,
 		},
+		{
+			&Update{
+				Table: "Ta",
+				Items: []UpdateItem{
+					{Column: "Cb", Value: IntegerLiteral(4)},
+					{Column: "Ce", Value: StringLiteral("wow")},
+					{Column: "Cf", Value: ID("Cg")},
+					{Column: "Ch", Value: nil},
```
olavloite, line 242: I would have expected this to set the value of `Ch` to `NULL`, regardless of the default value of `Ch`.

olavloite (approved):
The current implementation does not seem to support setting a column to `NULL`. I'm happy with regarding that as a TODO, as this would probably already help people a lot.

spanner/spansql/parser.go
dsymonds: The Spanner docs say that `DEFAULT` will set the column values to NULL (I guess because Spanner doesn't have non-NULL defaults?), but here ui.Value==nil [from `X = DEFAULT`] is distinct from ui.Value==NullLiteral(0) [from `X = NULL`], so both are possible.

spanner/spansql/sql_test.go
dsymonds: I'll add a `NULL` case explicitly to make this clearer.

PR #3201, opened 2020-11-12
feat(spanner/spannertest): support UPDATE DML

### LINKED ISSUE #3162: spanner/spansql: support for Update in parseDMLStmt

dsymonds:
Fixes #3162.

### DIFF
--- a/spanner/spannertest/README.md
+++ b/spanner/spannertest/README.md
@@ -19,16 +19,16 @@ by ascending esotericism:
 
 - expression functions
 - more aggregation functions
-- INSERT/UPDATE DML statements
 - SELECT HAVING
 - case insensitivity
-- FULL JOIN
+- FULL JOIN, multiple joins
 - alternate literal types (esp. strings)
 - STRUCT types
 - transaction simulation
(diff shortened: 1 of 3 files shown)


olavloite (approved):
LGTM

Issue #916, opened 2018-03-07
datastore: ignoring time zones


kemics:
Hello,

I am using datastore in my project and emulator to test it locally (https://cloud.google.com/datastore/docs/tools/datastore-emulator). And when I put some structure with UTC field and after that get it from datastore its zone becomes my local.

I checked the source code of `datastore` package and founds some test cases with time zone, for example:
https://github.com/GoogleCloudPlatform/google-cloud-go/blob/42f89745cc30769fe3ebaa7633ec13ab05bda051/datastore/datastore_test.go#L1290-L1296
And if I print `got` value before `got` and `want` comparison, for example here:
https://github.com/GoogleCloudPlatform/google-cloud-go/blob/42f89745cc30769fe3ebaa7633ec13ab05bda051/datastore/datastore_test.go#L1986
They will have different locations (because I am not in UTC zone), but the test will pass.

It happens because time fields are compared via this function:
https://github.com/GoogleCloudPlatform/google-cloud-go/blob/42f89745cc30769fe3ebaa7633ec13ab05bda051/internal/testutil/cmp.go#L40-L45
And I found no words about time comparison and that it ignores time zone, but it does, such simple example proves it:
```
package testutil

import (
	"time"
	"fmt"
)

func ExampleEqualTime() {
	date := time.Date(2017, time.April, 12, 0, 0, 0, 0, time.FixedZone("CUSTOM", -3720))
	fmt.Println(Equal(date, date.UTC()))

	// Output: true
}
```

So I am wondering if it is a bug or not really well-documented feature.
It matters because it is now a challenge to create a location agnostic tests.

It is really easy to reproduce such behaviour, f.e this example fails because my time zone is not UTC:
```
func ExamplePutAndGetTime() {
	client, err := datastore.NewClient(context.Background(), "sample1213")
	if err != nil {
		panic(err)
	}
	defer client.Close()

	type Event struct {
		Time time.Time
	}
	key := datastore.IDKey("Event", 1, nil)

	date := time.Date(2017, time.April, 12, 0, 0, 0, 0, time.UTC)

	_, err = client.Put(context.Background(), key, &Event{date})
	if err != nil {
		panic(err)
	}

	var got Event
	if err := client.Get(context.Background(), key, &got); err != nil {
		panic(err)
	}

	fmt.Println(got.Time)

	// Output: 2017-04-12 00:00:00 +0000 UTC
}
```

jba:
[...]

PR #3202, opened 2020-11-12
feat(datastore): support civil package types save

### LINKED ISSUE #3089: datastore: support civil.Date for properties
We sometimes store dates-without-location (in the sense of the `civil` package) in datastore.  Sadly, that means we have to handle them, temporarily at least, as `time.Time`.  This is sort of a foot-gun for all the usual reasons (described, for example, in e.g. golang/go#19700).

We could have a struct-field of type `civil.Date`; that way we never have to have a `time.Time`-representing-a-civil-date floating around.  Ideally this would be compatible with how the libraries in other languages, some of which have a date-property, do things.

We could write a PropertyLoadSaver for each such model, and specially handle the field.  That's definitely possible but it's some extra work for each such model.

IlyaFaer:
Support saving properties of `civil` package types: `Date`, `Time` and `DateTime`.

Closes #3089


datastore/save.go
```diff
@@ -53,6 +54,32 @@ func reflectFieldSave(props *[]Property, p Property, name string, opts saveOpts,
 	switch x := v.Interface().(type) {
 	case *Key, time.Time, GeoPoint:
 		p.Value = x
+	case civil.Date, civil.DateTime, civil.Time:
```
tritone, line 57: I would just separate each of these cases given that you are effectively handling them completely separately below.

datastore/save.go
```diff
@@ -53,6 +54,32 @@ func reflectFieldSave(props *[]Property, p Property, name string, opts saveOpts,
...
+			}
+			val, err := time.Parse(format, timeVal.String())
+			if err != nil {
+				return fmt.Errorf("datastore: error while parsing time: %v", err)
+			}
+			p.Value = val
+			*props = append(*props, p)
+			return nil
+		}
+		if dateTimeVal, ok := v.Interface().(civil.DateTime); ok {
```
tritone, line 78: I don't see a test for this case, could you add one?

datastore/save.go
```diff
@@ -53,6 +54,32 @@ func reflectFieldSave(props *[]Property, p Property, name string, opts saveOpts,
...
+		if timeVal, ok := v.Interface().(civil.Time); ok {
+			var format string
+			if timeVal.Nanosecond == 0 {
+				format = "15:04:05"
+			} else {
+				format = "15:04:05.000000000"
+			}
+			val, err := time.Parse(format, timeVal.String())
+			if err != nil {
+				return fmt.Errorf("datastore: error while parsing time: %v", err)
```
tritone, line 72: I would say "while parsing civil.Time" to be totally clear.

tritone (changes requested):
Do we need to make changes to load as well?

tritone:
FYI @benwhitehead @crwilcox

datastore/save.go
```diff
@@ -53,6 +54,32 @@ func reflectFieldSave(props *[]Property, p Property, name string, opts saveOpts,
 	switch x := v.Interface().(type) {
 	case *Key, time.Time, GeoPoint:
 		p.Value = x
+	case civil.Date, civil.DateTime, civil.Time:
+		if dateVal, ok := v.Interface().(civil.Date); ok {
+			p.Value = dateVal.In(time.Now().Location())
```
tbpg, line 59: Seems like this should always use UTC?

datastore/save.go
```diff
@@ -53,6 +54,32 @@ func reflectFieldSave(props *[]Property, p Property, name string, opts saveOpts,
...
+			*props = append(*props, p)
+			return nil
+		}
+		if timeVal, ok := v.Interface().(civil.Time); ok {
+			var format string
```
tbpg, line 64: Is there a particular reason to use `time.Time` and try to figure out formatting, rather than using `civil.Date.MarshalText` and `civil.Date.UnmarshalText`?

tbpg (changes requested):

datastore/save.go
tbpg: +1.

datastore/save.go
tbpg: Whoops -- duplicated comment. +1.

datastore/save.go
IlyaFaer: If we'll use Marshal*, we'll get just time "hh:mm:ss". With current implementation we're getting "yyyy-mm-dd hh:mm:ss +tz". I wasn't very sure which variant is more correct. I'm okay to use Marshal* methods, if it's okay.

datastore/save_test.go
```diff
@@ -371,6 +395,65 @@ func TestSaveFieldsWithInterface(t *testing.T) {
...
+					Time: civTimeVal,
+				},
+			},
+			want: []Property{
+				{
+					Name:  "CivDateTime",
+					Value: dateTimeVal,
+				},
+			},
+		},
```
IlyaFaer, line 456: Here is a test case for `DateTime`

datastore/save.go
IlyaFaer: There is a case: https://github.com/googleapis/google-cloud-go/pull/3202/files#r522815770
Still, there was a copy-past field names error. I've fixed it.

datastore/save.go
IlyaFaer: After overthinking this once again, yeah, I think it should.

tbpg (approved):
LGTM. Let's make sure to have a user test with a pseudo-version before tagging a release. Please wait for @tritone.

tritone (approved):
Looks good to me. @crwilcox @benwhitehead could you please take a look before we merge?

crwilcox:
What is the experience for someone who doesn't have civil?

tbpg:
> What is the experience for someone who doesn't have civil?

No changes. This should only change behavior for people who are using `civil`.

crwilcox:
@tbpg so there is no chance of panics or anything from a missing import?

tbpg:
@crwilcox, no. I don't think there is a way to cause a panic from a missing import in Go. If something is used, it is imported, otherwise there is a compile-time error.

crwilcox:
Thank you for clearing things up @tbpg. I had missed that the package already depended on civil which is why I was a bit confused how this worked. lgtm.

crwilcox (approved):

datastore/load.go
```diff
@@ -379,6 +383,15 @@ func setVal(v reflect.Value, p Property) (s string) {
...
 			v.Set(reflect.ValueOf(x))
+		case typeOfCivilDate:
+			date := civil.DateOf(pValue.(time.Time))
+			v.Set(reflect.ValueOf(date))
+		case typeOfCivilDateTime:
+			dateTime := civil.DateTimeOf(pValue.(time.Time))
+			v.Set(reflect.ValueOf(dateTime))
+		case typeOfCivilTime:
+			timeVal := civil.TimeOf(pValue.(time.Time))
+			v.Set(reflect.ValueOf(timeVal))
```
dnerdy, line 394: Thank you @IlyaFaer for working on this!

I'm following up via #3089.

There's an issue loading civil dates and times on a machine when the timezone is set to anything other than UTC. Time properties are loaded in a machine's local location (see #916) and therefore need to be converted to UTC before using them to create civil values.

I think this is the adjustment that's needed:
```
case typeOfCivilDate:
	date := civil.DateOf(pValue.(time.Time).In(time.UTC))
	v.Set(reflect.ValueOf(date))
case typeOfCivilDateTime:
	dateTime := civil.DateTimeOf(pValue.(time.Time).In(time.UTC))
	v.Set(reflect.ValueOf(dateTime))
case typeOfCivilTime:
	timeVal := civil.TimeOf(pValue.(time.Time).In(time.UTC))
	v.Set(reflect.ValueOf(timeVal))
```

codyoss:
@dnerdy Would you please open an issue for this, thanks for finding this!

PR #3208, opened 2020-11-13
docs: expand on developer environment setup

crwilcox:
I just walked through setting up the development environment. I fixed a few small typos, but also tried to add my notes for what I saved in `.zshrc`. I figure most contributors are doing a similar thing?


CONTRIBUTING.md
```diff
@@ -167,7 +170,37 @@ $ gcloud kms keys create key2 --keyring $MY_KEYRING --location $MY_LOCATION --pu
...
+export GCLOUD_TESTS_GOLANG_FIRESTORE_PROJECT_ID=your-firestore-project
+
+# The path to the JSON key file of the Firestore project's service account.
+export GCLOUD_TESTS_GOLANG_FIRESTORE_KEY=~/directory/your-firestore-project-abcd1234.json
+
+# The full name of the keyring for the tests, in the form "projects/P/locations/L/keyRings/R".
+# The creation of this is described below.
+export MY_KEYRING=my-golang-sdk-test
+export MY_LOCATION=global
+export GCLOUD_TESTS_GOLANG_KEYRING=projects/$GCLOUD_TESTS_GOLANG_PROJECT_ID/locations/$MY_LOCATION/keyRings/$MY_KEYRING
```
tritone, line 197: maybe not worth adding a full path here and in some of the paths above, as this will vary by user (and it should definitely be an absolute path in any case).

CONTRIBUTING.md
```diff
@@ -179,6 +212,11 @@ running:
 $ go test -v cloud.google.com/go/...
 ```
 
+You can also run a specific test suite, for instance datastore be running:
```
tritone, line 216: "...in order to run datastore:"?

go.mod
```diff
@@ -5,24 +5,25 @@ go 1.11
 require (
 	cloud.google.com/go/bigquery v1.8.0
 	cloud.google.com/go/datastore v1.1.0
+	cloud.google.com/go/firestore v1.3.0 // indirect
```
tritone: Remove these changes from the PR.

CONTRIBUTING.md
tritone: I also am not sure if this works for directories which are not separate modules, not sure though...

CONTRIBUTING.md
crwilcox: By full path hat do you mean? The included path is a constructed form from location/name/project id.

CONTRIBUTING.md
```diff
@@ -190,6 +224,11 @@ running:
 $ go test -v cloud.google.com/go/...
 ```
 
+You can also run a specific test suite, for instance in order to run datastore:
+``` sh
+$ go test -v cloud.google.com/go/datastore
```
tbpg, line 229: `datastore` is a different module. So, if you run this from the root of the repo, it won't use your local version of the `datastore` module. It will use the latest tagged version (you can try it by making a test fail every time locally then running this command from the repo root).

The solution is to change into a directory inside the module you want to test, then run the tests. I usually use `go test -v` or `go test -v ./...`, but you can also use the full path if you prefer.

crwilcox: Just verified this. Good to know! Will update guide.

CONTRIBUTING.md
```diff
@@ -190,6 +224,12 @@ running:
 $ go test -v cloud.google.com/go/...
```
crwilcox, line 224: @tbpg should this also be changed to explicitly target the current development directory?

tbpg: Yes. It might also be good to note that this doesn't run the tests in other modules. So, this won't run the tests for the `datastore` module.

tbpg (approved):
Thanks!

CONTRIBUTING.md
tritone: I was confused, I was thinking this was a local file path for some reason.

tritone (approved):

crwilcox:
Thanks @tbpg and @tritone for the feedback. Glad I could incorporate some small changes I think may help with user friction.

PR #3214, opened 2020-11-16
chore(internal): stop generating Ads packages in genproto

tbpg:
These packages don't have a corresponding client library (in
cloud.google.com/go or another module). They aren't ready for production
use. So, we're going to stop generating them for now and will revisit
in the future when ready.

A go-genproto PR will remove the existing Ads packages.

If needed, these packages can be generated locally or as part of a build
process. However, we're not ready to generate and publish them
automatically.

@anashoommen @aohren

### DIFF
--- a/internal/gapicgen/generator/genproto.go
+++ b/internal/gapicgen/generator/genproto.go
@@ -42,6 +42,19 @@ var denylist = map[string]bool{
 	"google.golang.org/genproto/googleapis/devtools/containeranalysis/v1": true,
 }
 
+var skipPrefixes = []string{
+	"google.golang.org/genproto/googleapis/ads",
+}
+
+func hasPrefix(s string, prefixes []string) bool {
+	for _, prefix := range prefixes {
+		if strings.HasPrefix(s, prefix) {
+			return true
+		}
+	}
+	return false
+}
+
 // regenGenproto regenerates the genproto repository.
 //
 // regenGenproto recursively walks through each directory named by given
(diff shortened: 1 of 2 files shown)


aohren:
LGTM

internal/gapicgen/generator/genproto.go
```diff
@@ -110,7 +123,7 @@ func regenGenproto(ctx context.Context, genprotoDir, googleapisDir, protoDir str
 	// Run protoc on all protos of all packages.
 	grp, _ := errgroup.WithContext(ctx)
 	for pkg, fnames := range pkgFiles {
-		if !strings.HasPrefix(pkg, "google.golang.org/genproto") || denylist[pkg] {
+		if !strings.HasPrefix(pkg, "google.golang.org/genproto") || denylist[pkg] || hasPrefix(pkg, skipPrefixes) {
```
hongalex, line 126: If `skipPrefixes` gets larger, it would be more efficient to preemptively remove strings that match the prefix from `pkgFiles` rather than checking it everytime in this for loop. Considering the size of `skipPrefixes` for now, I don't think it's a blocking optimization.

hongalex (approved):

aohren (approved):

internal/gapicgen/generator/genproto.go
tbpg: I agree on both counts. Going for simplicity for now.

tbpg (commented):
Thanks!

PR #3216, opened 2020-11-16
feat(pubsub): Make AckHandler a public interface

tmdiep:


tbpg (commented):
I think this is good for the API of `NewMessage`, but I don't think `NewMessage` should be exported.

PR #3220, opened 2020-11-17
fix(pubsublite): fixed return value of AdminClient.TopicSubscriptions

tmdiep:
It should only return an iterator.
Also added unit tests.


pubsublite/main_test.go
```diff
@@ -0,0 +1,50 @@
+// Copyright 2020 Google LLC
```
hongalex, line 1: This seems to be a duplicate of internal/wire/main_test. Is there a reason why it's being added here too?

hongalex (approved):

pubsublite/main_test.go
tmdiep: Ah yeah, this was unfortunate. `TestMain` only applies to the package where it resides. I will have to copy to a third package, so might move the bulk of it to a common util (probably in a separate PR, since this one is already approved) shared by all the main_test.go clones.

tmdiep:
Thanks Alex. Will do the refactoring for main_test.go in a separate PR. Also wanted to point out that the change to `TopicSubscriptions` is technically a breaking change, but since this module is still in early development, I didn't mark it as such. Wasn't sure if the auto release tool would bump up the major revision number.

PR #3221, opened 2020-11-17
feat(pubsublite): Single and multi partition subscribers

tmdiep:
Implements:
* `subscriberStream`: directly wraps the Subscribe bidi stream.
* `singlePartitionSubscriber`: wraps the committer and `subscriberStream`.
* `multiPartitionSubscriber`: receives messages from multiple fixed partitions.


tmdiep:
WANT_LGTM=@manuelmenzella-google 
Other reviewers optional. Feel free to leave comments.

pubsublite/internal/wire/subscriber.go
```diff
@@ -0,0 +1,321 @@
...
+		ack := newAckConsumer(msg.GetCursor().GetOffset(), msg.GetSizeBytes(), s.onAck)
+		if err := s.acks.Push(ack); err != nil {
+			return err
+		}
+
+		// Release the mutex before delivering the message to the receiver. This
+		// allows the user to stop the subscriber (which would discard remaining
+		// messages) and other operations (e.g. ack processing, batch flow control)
+		// to occur.
+		s.mu.Unlock()
```
manuelmenzella-google, line 218: This means that we are locking/unlocking this mutex once per message. Are we worried about the overhead? Would it be better to gather all messages and send them to the client in a batch?

pubsublite/internal/wire/subscriber.go
```diff
@@ -0,0 +1,321 @@
...
+	defer s.mu.Unlock()
+
+	switch status {
+	case streamConnected:
+		s.unsafeUpdateStatus(serviceActive, nil)
+
+		// Reinitialize the offset and flow control tokens when a new subscribe
+		// stream instance is connected.
+		if seekReq := s.offsetTracker.RequestForRestart(); seekReq != nil {
+			if s.stream.Send(&pb.SubscribeRequest{
```
manuelmenzella-google, line 146: Nit: do you mind adding a comment explaining why it's safe to ignore the return value of Send()?

manuelmenzella-google (approved):

pubsublite/internal/wire/subscriber.go
tmdiep: Done. Thanks for reviewing!

pubsublite/internal/wire/subscriber.go
tmdiep: Actually, I've thought of a way to deliver a batch of messages to the shim layer (which delivers single messages to users) and avoid the per-message locking/unlocking here. If it's a relatively minimal change, I'll update this PR. Otherwise will submit as-is and ament in a subsequent PR.

Thanks for pointing this out!

tmdiep: The change was minimal, so I just amended this PR.

PR #3222, opened 2020-11-17
feat(bigtable): Backup Level IAM

AlisskaPie:
Implementing Backup Level IAM methods.


AlisskaPie:
[Bigtable Backup Level IAM](https://docs.google.com/document/d/1ScYmO98FZN-gyGes_Vb6g79GVO8YURQwC48jG3u2c0g/edit#heading=h.s7j00pnblo7n)
Java implementation: https://github.com/googleapis/java-bigtable/pull/450/files
Python: https://github.com/googleapis/python-bigtable/pull/160/files

bigtable/admin.go
```diff
@@ -1663,3 +1664,53 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
 	_, err = ac.tClient.UpdateBackup(ctx, req)
 	return err
 }
+
+// GetIamPolicy gets the IAM access control policy for the specified backup.
+func (ac *AdminClient) GetIamPolicy(ctx context.Context, cluster, backup string) (*iampb.Policy, error) {
```
tbpg, line 1669: s/Iam/IAM/g

https://github.com/golang/go/wiki/CodeReviewComments#initialisms

tbpg, line 1669: Looks like this should be named `BackupIAM` to match with `TableIAM`: https://pkg.go.dev/cloud.google.com/go/bigtable#AdminClient.TableIAM

bigtable/admin.go
```diff
@@ -1663,3 +1664,53 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+	ctx = mergeOutgoingMetadata(ctx, ac.md)
+	prefix := ac.instancePrefix()
+	backupPath := prefix + "/clusters/" + cluster + "/backups/" + backup
+
+	req := iampb.GetIamPolicyRequest{Resource: backupPath}
```
tbpg, line 1674: Looks like this function should return an `*iam.Handle` and use https://pkg.go.dev/cloud.google.com/go/iam#InternalNewHandleClient to do it, rather than making the request directly.

bigtable/admin.go
```diff
@@ -1663,3 +1664,53 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+	policy, err := ac.tClient.GetIamPolicy(ctx, &req)
+	if err != nil {
+		return nil, err
+	}
+	return policy, nil
+}
+
+// SetIamPolicy sets the IAM access control policy for the specified backup.
+// Replaces any existing policy.
+func (ac *AdminClient) SetIamPolicy(ctx context.Context, cluster, backup string, policy *iampb.Policy) (*iampb.Policy, error) {
```
tbpg, line 1684: I think `SetIamPolicy` and `TestIamPolicy` can be deleted once `BackupIAM` above returns `*iam.Handle`.

tbpg (changes requested):

bigtable/admin.go
AlisskaPie: Thanks for quick response! 
Hmm, I thought we should have three methods, like in other languages, but your suggestion makes more sense.
Not a problem, I'll update.

bigtable/admin.go
```diff
@@ -604,6 +603,13 @@ func (ac *AdminClient) TableIAM(tableID string) *iam.Handle {
 		"projects/"+ac.project+"/instances/"+ac.instance+"/tables/"+tableID)
 }
 
+// BackupIAM creates an IAM client specific to a given Cluster and Backup.
```
tbpg, line 606:
```suggestion
// BackupIAM creates an IAM Handle specific to a given Cluster and Backup.
```

bigtable/admin.go
```diff
@@ -604,6 +603,13 @@ func (ac *AdminClient) TableIAM(tableID string) *iam.Handle {
...
+func (ac *AdminClient) BackupIAM(ctx context.Context, cluster, backup string) *iam.Handle {
+	prefix := ac.instancePrefix()
+	backupPath := prefix + "/clusters/" + cluster + "/backups/" + backup
```
tbpg, line 609:
```suggestion
	backupPath := ac.instancePrefix() + "/clusters/" + cluster + "/backups/" + backup
```

bigtable/integration_test.go
```diff
@@ -1243,6 +1248,88 @@ func TestIntegration_TableIam(t *testing.T) {
 	}
 }
 
+func TestIntegration_BackupIAM(t *testing.T) {
```
tbpg, line 1251: Given this is an integration test file, it feels like we shouldn't be mocking out the interactions. It's probably enough to follow the pattern of `TestIntegration_TableIam` (which should be `IAM`, but that's another story :upside_down_face:).

tbpg (changes requested):
Thanks!

bigtable/integration_test.go
tritone: +1. @kolea2 do you know how other languages are handling this testing, and if we will face any quota-related issues?

bigtable/admin.go
tritone: Should make this correction in L600 above as well.

bigtable/integration_test.go
AlisskaPie: I added a mock setup because I got an error while comparing policies. I had an `Etag` mismatch, even in the read-modify-write cycle.
From [documentation](https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/Policy): 
> It is strongly suggested that systems make use of the etag in the read-modify-write cycle to perform policy updates in order to avoid race conditions: An etag is returned in the response to getIamPolicy, and systems are expected to put that etag in the request to setIamPolicy to ensure that their change will be applied to the same version of the policy.

But I had and still have this error without mocking.
image

Will wait for response!

P.S. Anyway, I can also rename `TestIntegration_TableIam` to `TestIntegration_TableIAM` here :)
It's related to this PR.

tbpg: The Etag should only change when there is a modification. So, something else must have modified the value between the initial read and the subsequent attempted write. Mind uploading the updated test without the mock, even if it isn't passing yet?

tbpg (changes requested):
Thanks!

bigtable/integration_test.go
AlisskaPie: Yes, sure!
Uploaded.

kolea2: @tritone for Java we take the existing IAM policy and reset it. Probably could be improved: https://github.com/googleapis/java-bigtable/blob/master/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableTableAdminClientIT.java#L194-L217. I haven't seen any quota related issues so far.

bigtable/admin.go
```diff
@@ -597,12 +597,18 @@ func (ac *AdminClient) WaitForReplication(ctx context.Context, table string) err
...
-// TableIAM creates an IAM client specific to a given Instance and Table within the configured project.
+// TableIAM creates an IAM Handle specific to a given Instance and Table within the configured project.
 func (ac *AdminClient) TableIAM(tableID string) *iam.Handle {
 	return iam.InternalNewHandleGRPCClient(ac.tClient,
 		"projects/"+ac.project+"/instances/"+ac.instance+"/tables/"+tableID)
 }
 
+// BackupIAM creates an IAM Handle specific to a given Cluster and Backup.
+func (ac *AdminClient) BackupIAM(ctx context.Context, cluster, backup string) *iam.Handle {
+	backupPath := ac.instancePrefix() + "/clusters/" + cluster + "/backups/" + backup
```
kolea2, line 608: is this anywhere else in the code? If not, it seems useful as a common utility.

AlisskaPie: I've met this code in few places. In [UpdateBackup](https://github.com/googleapis/google-cloud-go/blob/2a980a2cbae52faffd778c75767c3f96f121a2d1/bigtable/admin.go#L1642-L1646), [BackupInfo](https://github.com/googleapis/google-cloud-go/blob/2a980a2cbae52faffd778c75767c3f96f121a2d1/bigtable/admin.go#L1604-L1608), [RestoreTable](https://github.com/googleapis/google-cloud-go/blob/2a980a2cbae52faffd778c75767c3f96f121a2d1/bigtable/admin.go#L1443-L1446), [DeleteBackup](https://github.com/googleapis/google-cloud-go/blob/2a980a2cbae52faffd778c75767c3f96f121a2d1/bigtable/admin.go#L1628-L1632)

kolea2: cool - I think it would be worth it to refactor this as a utility helper then in this PR

AlisskaPie: Sorry for the delay here, I had an emergency.

Done, thanks for advice!

kolea2: No worries, lgtm thank you!

bigtable/integration_test.go
kolea2: @tbpg @tritone any additional thoughts here?

tbpg (approved):

kolea2 (approved):

tritone (approved):

PR #3231, opened 2020-11-19
refactor(pubsub): use Message and PublishResult in internal/pubsub

tmdiep:
Moves `Message` and `PublishResult` (which will be shared by the `pubsub` and `pubsublite` modules) to `internal/pubsub`. Adds aliases for these types in the `pubsub` module and refactors all usage.

Follow up from discussion in https://github.com/googleapis/google-cloud-go/pull/3200.


hongalex:
After some discussion with @tmdiep, we found that while the implementation is fine, using a type alias seems to make the docs and IDE experience poor.

As an example, none of the methods are generated properly for `Message` and `PublishResult`
image
image

IDE support seems to be fine, so this is an extra redirect that users will need to make only when visiting the pkg.go.dev site.
image

Lastly, I'm still in favor of exposing the `AckHandler` as a public interface. This is the equivalent of Java's [`AckReplyConsumer`](https://github.com/googleapis/java-pubsub/blob/621d2c8fcdd18eca764a005a7cdbc58b535c19e9/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/AckReplyConsumer.java) which is public, and I think having feature parity (and the benefit of simpler testing) justifies this addition.

dsymonds:
Yes, I was just about to point out that a type alias to move the entire thing into an internal package isn't great for that reason (e.g. https://pkg.go.dev/google.golang.org/grpc/status#Status is awkward to deal with).

I'm still unclear what the goal here is. The discussion on #3200 seemed to jump around different goals. If the goal is for there to be some sort of drop-in replacement package with a comparable API (but no need for types to be actually identical), why isn't it just `git cp pubsub/* pubsublite/*` followed by changing the implementation?

Exposing `AckHandler` remains a poor idea, despite Java's precedence. It does not feel like Go to inject those sorts of hooks into a visible interface in that way, and, as I mentioned on #3200, there are better approaches to testing that don't need any changes to the API of the `pubsub` package (e.g. unlike in Java, where you need a declared interface to be able to substitute, Go code can define ad hoc interfaces that are implicitly satisfied).

tmdiep:
> if there's an internal/pubsub package, it can have a little registration mechanism (and probably the ackHandler type definition) where the pubsub package can register a func that can be invoked by the pubsublite package to twiddle that field. 

@dsymonds - I was going to also prototype your suggestion above to see how it compares to this approach.

> If the goal is for there to be some sort of drop-in replacement package with a comparable API (but no need for types to be actually identical), why isn't it just `git cp pubsub/* pubsublite/*` followed by changing the implementation?

A few of us did advocate for this, but it wasn't good enough, as others wanted the shim to implement, e.g. `func Publish(context.Context, *pubsub.Message) *pubsub.PublishResult` interface so that users could substitute it for `pubsub.Topic.Publish()` without having to copy/paste their existing pubsub code & change the package, or write their own translation layer.

The other reason is that we may have message, publisher and subscriber clients in the `pubsublite` package that aren't identical to `pubsub`'s API. So we wanted to reserve the `pubsublite` namespace for a future Lite-specific client, and have a separate shim package for existing pubsub users.

tmdiep:
> It does not feel like Go to inject those sorts of hooks into a visible interface in that way

[heap](https://golang.org/pkg/container/heap/#Interface) operating on a user-provided interface seemed similar in principle. But agree that this is uncommon.

dsymonds:
> > It does not feel like Go to inject those sorts of hooks into a visible interface in that way
> 
> [heap](https://golang.org/pkg/container/heap/#Interface) operating on a user-provided interface seemed similar in principle. But agree that this is uncommon.

`container/heap`'s entire reason for existence is to interact with a user-provided data structure, so accepting an interface for that is expected (`sort.Interface` is the simpler case). Here, only weird testing edge cases and another semi-related package would ever use `pubsub.NewMessage` and provide a custom `AckHandler`, and 99.99% of the other code has no use for it, so it is a very different situation.

I'm looking forward to seeing your other prototype!

tmdiep:
> I'm looking forward to seeing your other prototype!

I've been trying, but can't find a way to make this work. Every solution I've come up with runs into cyclic dependencies or exposing/adding more to the `pubsub` API.

We have numerous Message instances, created by one of many subscriber client instances (which processes the message's ack/nack), so there cannot be one global ack consumer to register for `pubsublite`. When the `pubsub` package tries to set an ackHandler for a message, it doesn't know which one of these ack consumers to associate with the message without being passed some sort of ID (requires a new public function in the `pubsub` package, which defeats the purpose). If I try to store a global map of `pubsub.Message` to ack handler in `internal/pubsub` (`pubsublite` performs the registration), so that the `pubsub` package can perform the look up and forward the ack/nack event, I run into cyclic dependency issues.

A workable solution will not be nice, e.g. a global queue of messages to be created.

I think this PR is the more viable option. Perhaps we could have verbose comments for the type aliases (copying most of what is in internalpubsub) as temporary mitigation and lodge a FR for a godoc to inline documentation for type aliases.

Personally, I think exposing `AckHandler` would be more convenient for users than asking them to write wrappers for their unit tests. The extra public functions don't cause any harm to the rest of the pubsub API. As for users being confused about their utility, we could document the fact that they are only expected to be called by user-written unit tests.

tmdiep:
@codyoss @hongalex @tbpg @dsymonds 

Are folks ok with proceeding with this solution? If so, I'll need to move the changes in `internal/pubsub` to a separate PR and wait for a root module release before updating this PR.

As Alex pointed out, it does affect docs. But we can mitigate by copying over some comments to the type aliases.

hongalex:
I'm ok with this solution as we've tried out the other suggestions without success. With that said, I'd like to wait for @dsymonds to weigh in before we make any changes to avoid the partial rollbacks we made last time.

hongalex (approved):

pubsub/message.go
```diff
@@ -15,55 +15,50 @@
...
-	// size is the approximate size of the message's data and attributes.
-	size int
-
-	// OrderingKey identifies related messages for which publish order should
-	// be respected. If empty string is used, message will be sent unordered.
-	OrderingKey string
-
-	// ackh handles Ack() or Nack().
-	ackh ackHandler
+type Message = internalpubsub.Message
```
codyoss, line 27: Only downside of this is that the godoc will now be a little less clear(indirect). Takes an extra click to see all of the exported fields. Seems like an okay tradeoff though. Example: [Api](https://pkg.go.dev/google.golang.org/genproto@v0.0.0-20201203001206-6486ece9c497/protobuf/api#Api)

dsymonds:
I still think some of the original goals (e.g. having a drop-in replacement package, and sharing types; it feels like impositions from other languages) are mistakes, but you're right that we didn't find a better solution that met those goals, so I'm not going to try to hold this up.

pubsub/message.go
tmdiep: Agreed. In the latest commit for this PR, I elaborated on the comments for the aliases to save users having to click through in some cases.

I also changed the import to `ipubsub` (more for aesthetic reasons), but can change it back to `internalpubsub` if it's better to be clear about where the actual types are defined.

PR #3232, opened 2020-11-19
fix(pubsub): Remove NewMessage and NewPublishResult

tmdiep:
Rolls back the public functions added in https://github.com/googleapis/google-cloud-go/pull/3200. We have decided not to expose these to users.


hongalex (approved):

tmdiep:
Submitting in case we need to do a pubsub release before this issue is resolved.

PR #3233, opened 2020-11-19
feat(spanner): support request and transaction tags

olavloite:
Adds support for request and transaction tags. This change depends on proto changes that have not yet been merged.


spanner/client_test.go
```diff
@@ -2239,3 +2239,169 @@ func TestClient_DoForEachRow_ShouldEndSpanWithQueryError(t *testing.T) {
 		t.Errorf("Span status mismatch\nGot: %v\nWant: %v", s.Code, codes.InvalidArgument)
 	}
 }
+
+func TestClient_ReadOnlyTransaction_Tag(t *testing.T) {
```
syeduguri, line 2243: Better to have separate tests for SingleUse read/query, PartitionedDML and BatchDML requests.

spanner/client_test.go
```diff
@@ -2239,3 +2239,169 @@ func TestClient_DoForEachRow_ShouldEndSpanWithQueryError(t *testing.T) {
...
+			tx.Commit(context.Background())
+			checkCommitRequestForTags(t, server.TestSpanner, QueryOptions{
+				RequestTag:     "", // Should never be set
+				transactionTag: to.TransactionTag,
+			})
+		}
+	}
+}
+
+func checkRequestsForTags(t *testing.T, server InMemSpannerServer, reqCount int, qo QueryOptions) {
```
syeduguri, line 2356: Better to rename it as "expectedQueryOptions".

spanner/transaction.go
```diff
@@ -76,6 +76,10 @@ type txReadOnly struct {
 
 // TransactionOptions provides options for a transaction.
 type TransactionOptions struct {
+	// The transaction tag to use for a read/write transaction.
+	// This tag is automatically included with each statement and the commit
+	// request of a read/write transaction.
+	TransactionTag string
```
syeduguri, line 82: Just wondering why there was empty TransactionOptions struct before this change.

spanner/transaction.go
```diff
@@ -151,6 +160,10 @@ func (t *txReadOnly) ReadWithOptions(ctx context.Context, table string, keys Key
 					KeySet:      kset,
 					ResumeToken: resumeToken,
 					Limit:       int64(limit),
+					RequestOptions: &sppb.RequestOptions{
+						RequestTag:     requestTag,
+						TransactionTag: t.qo.transactionTag,
```
syeduguri, line 165: We don't need transactionTag for readonly transactions.

syeduguri (commented):
Looks like changes covered all APIs and scenarios. Just want to know where we handled PartitionedDML requests.
Left few minor comments.

spanner/client_test.go
olavloite: I don't think QueryOptions is the right name in this case, as `QueryOptions` also contain other options than tags, and those are not checked here. I think checkFor...RequestOptions would fit better here.

spanner/transaction.go
olavloite: We had a major version bump recently, and added a number of breaking changes already back then that we knew would be needed for (among others) tagging support. Adding a struct is not breaking, but we also added a number of new methods that use this struct as a parameter.

spanner/transaction.go
olavloite: That is true, but this method is used for both read-only and read/write transactions. The `transactionTag` field can however only be set for read/write transactions, so it will always be empty when this is used in combination with a read-only transaction.

olavloite:
> Looks like changes covered all APIs and scenarios. Just want to know where we handled PartitionedDML requests.
> Left few minor comments.

@syeduguri Thanks for the review.
Good point regarding PDML. That was missing but has been added now. PTAL.

syeduguri (approved):

syeduguri (approved):

syeduguri:
Changes looks good to me.

spanner/client_test.go
```diff
@@ -2239,3 +2239,209 @@ func TestClient_DoForEachRow_ShouldEndSpanWithQueryError(t *testing.T) {
...
+	for _, qo := range []QueryOptions{
+		{},
+		{RequestTag: "request-tag-1"},
+	} {
+		client.PartitionedUpdateWithOptions(context.Background(), NewStatement(UpdateBarSetFoo), qo)
+		checkRequestsForExpectedRequestOptions(t, server.TestSpanner, 1, qo)
+	}
+}
+
+func TestClient_Apply_Priority(t *testing.T) {
```
hengfengli, line 2371: `Apply_Tagging`?

spanner/transaction.go
```diff
@@ -223,27 +236,46 @@ func (t *txReadOnly) ReadRowUsingIndex(ctx context.Context, table string, index
...
+// QueryOptions provides options for executing a sql query or update statement.
 type QueryOptions struct {
-	Mode    *sppb.ExecuteSqlRequest_QueryMode
-	Options *sppb.ExecuteSqlRequest_QueryOptions
+	Mode       *sppb.ExecuteSqlRequest_QueryMode
+	Options    *sppb.ExecuteSqlRequest_QueryOptions
+	RequestTag string
+	// transactionTag is unexported as applications should set this on the transaction and
+	// the transaction will then set this field for each request during the transaction.
+	transactionTag string
```
hengfengli, line 246: Why don't we put this to `TransactionOptions`? `transactionTag` seems to be more tight with a transaction instead of a query/update?

spanner/transaction.go
```diff
@@ -874,6 +922,10 @@ func (t *ReadWriteTransaction) BatchUpdate(ctx context.Context, stmts []Statemen
 		Transaction: ts,
 		Statements:  sppbStmts,
 		Seqno:       atomic.AddInt64(&t.sequenceNumber, 1),
+		RequestOptions: &sppb.RequestOptions{
+			RequestTag:     opts.RequestTag,
+			TransactionTag: t.qo.transactionTag,
```
hengfengli, line 927: What I mean is that we extract `TransactionTag` from `t.to` (transaction options) instead of `t.qo` (query options)? Just a thought.

spanner/client_test.go
olavloite: Done

spanner/transaction.go
olavloite: That's a reasonable question. I'll look into that to see whether that would be more reasonable once the proto changes have been merged. I tried to re-generate locally, but that seems to fail at the moment.

olavloite: I've removed the `transactionTag` from the `QueryOptions` as it was not a logical place to keep it.

spanner/transaction.go
olavloite: Good point, done (see also above).

hengfengli (approved):
LGTM

skuruppu (approved):

skuruppu:
@olavloite unfortunately we have to roll this back since we are not able to release it yet as the backend is not ready for this. Hope you don't mind.

olavloite:
> @olavloite unfortunately we have to roll this back since we are not able to release it yet as the backend is not ready for this. Hope you don't mind.
> 
> FYI @hengfengli

I'll prepare a revert. We probably also need to revert (or tweak) #3905 as well.

PR #3238, opened 2020-11-19
feat(pubsublite): assigning subscriber implementation

tmdiep:
`assigningSubscriber` uses the Pub/Sub Lite partition assignment service to listen to its assigned partition numbers and dynamically add/remove `singlePartitionSubscribers`.


tmdiep:
WANT_LGTM=@dpcollins-google for checking assigner and assigningSubscriber logic.
Other reviewers optional. Feel free to leave comments.

pubsublite/internal/wire/assigner.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (ps partitionSet) Ints() (partitions []int) {
+	for p := range ps {
+		partitions = append(partitions, p)
+	}
+	return
+}
+
+func (ps partitionSet) Contains(partition int) bool {
```
dpcollins-google, line 48: nit. consider a typedef

pubsublite/internal/wire/assigner.go
```diff
@@ -0,0 +1,169 @@
...
+	_, exists := ps[partition]
+	return exists
+}
+
+// A function that generates a 16-byte UUID.
+type generateUUIDFunc func() (uuid.UUID, error)
```
dpcollins-google, line 54: How can this fail?

pubsublite/internal/wire/assigner.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (a *assigner) newStream(ctx context.Context) (grpc.ClientStream, error) {
+	return a.assignmentClient.AssignPartitions(ctx)
+}
+
+func (a *assigner) initialRequest() (interface{}, bool) {
+	return a.initialReq, false // No initial response expected
+}
+
+func (a *assigner) validateInitialResponse(_ interface{}) error {
```
dpcollins-google, line 119: hmm. How did you configure the retryable stream to not have an initial response?

pubsublite/internal/wire/subscriber.go
```diff
@@ -399,3 +401,107 @@ func newMultiPartitionSubscriber(subFactory *singlePartitionSubscriberFactory) *
...
+		ctx:              ctx,
+		subClient:        subClient,
+		cursorClient:     cursorClient,
+		settings:         settings,
+		subscriptionPath: subscriptionPath,
+		receiver:         receiver,
+	}
+
+	if len(settings.Partitions) > 0 {
+		return newMultiPartitionSubscriber(subFactory), nil
```
dpcollins-google, line 500: I would actually delete this class entirely. assigningSubscriber is a generalization of this, where the "assigner" only ever produces a single set of values.

pubsublite/internal/wire/assigner.go
dpcollins-google: Oh wow the uuid library can fail XD weird....

dpcollins-google (changes requested):
LGTM after changes from a pub/sub lite perspective, I'd prefer if you get someone familiar with go to review as well.

pubsublite/internal/wire/subscriber.go
tmdiep: True.  Although we can avoid extra complexity in the `assigner` to handle this mode, avoid creating the PartitionAssignmentClient, etc. So I might leave `newMultiPartitionSubscriber`, as it's only a few lines of extra code (+tests) anyway.

pubsublite/internal/wire/assigner.go
tmdiep: Ah yeah... I used raw types in the API due to other reviewers' comments and just kept it consistent in the wire package. Will think about defining strong types (in a separate PR, since it requires unrelated changes). Since this is called `partitionSet`, it might be ok.

pubsublite/internal/wire/assigner.go
tmdiep: The second return value of `initialRequest` is a bool indicating whether an initial response is expected. I will refactor it to a strong type in a separate PR (pulls in other unrelated changes) to make it clearer.

As for why - in long running integration tests, I found that when the partition assignment stream reconnects and the client UUID hasn't changed, the server doesn't bother to return the initial set of partition assignments (so my test would just hang waiting for it, then timeout and fail).

tmdiep:
WANT_LGTM=@hongalex for Go, and also Dan doesn't have write access and can't approve anyway.

manuelmenzella-google (approved):

PR #3243, opened 2020-11-20
fix: remove pubsublite from internal reporting

anguillanneuf:
The `release_level` for the pubsublite Go client is incorrectly thought to be `GA` by a Drift table because of the presence of `cloud.google.com/go/pubsublite/apiv1`.

This PR proposes that we remove `cloud.google.com/go/pubsub/apiv1` entirely for internal reporting (see b/173748364 for context), until the manual library is released.

The alternative is to add the following to `.repo-metadata-full.json`:

```json
  "cloud.google.com/go/pubsublite": {
    "distribution_name": "",
    "description": "Cloud Pub/Sub Lite",
    "language": "Go",
    "client_library_type": "manual",
    "docs_url": "",
    "release_level": "unspecified"
  },
```

@codyoss @hongalex


tbpg (changes requested):
This file is autogenerated. Would you mind updating the config file?

https://github.com/googleapis/google-cloud-go/blob/cf308308a3da134163a19ab78d6c605d6a2c2ef8/internal/gapicgen/generator/config.go#L683-L698

That config controls what libraries gets generated. So, it's probably better to change the pubsublite version to something < ga. But, I'm not familiar with what's intended.

tbpg (approved):

anguillanneuf:
Alex, Cody, correct me if wrong, the release level for the generated client is indeed considered "GA". 
I'm downgrading it to "unspecified" for reporting purposes.

jmdobry:
Shouldn't the release level value be modified, instead of commenting out the whole config? The library is "released" in the sense that it can be used, I think it needs to have a release level. We've already got tooling aware of and tracking this library, I feel like commenting out the whole config could have adverse effects.

hongalex:
So Cody marked the GAPIC Lite client as GA last month in #2962. However, as Jason mentioned, commenting out the config will prevent automatic updates to the Lite GAPIC client in this repository, which I'm not in favor of.

We haven't done any 1.0 releases of the pubsublite/apiv1 library, nor have we marked it as GA anywhere else, so changing it to `beta` should be fine here (and is more accurate). We've also been dissuading users from using the GAPIC client directly since it doesn't provide a great out of the box experience anyway.

hongalex (approved):

anguillanneuf:
I'm not sure what the test failure is about. I will leave it for the Go reviewers. Feel free to merge if it's a non-issue.

tbpg (approved):

PR #3253, opened 2020-11-23
feat(pubsublite): allow committer to terminate when acks are canceled

tmdiep:
Allows an `AckConsumer` to be canceled from outside the `wire` package. For example, the CPS subscriber shim will receive a batch of messages from the wire subscriber (but deliver single messages to users), and will cancel remaining ack consumers for the batch if the user stops the subscriber midway.

This scenario is now handled by the `committer`, which terminates once all outstanding ack consumers have either been acked or canceled.

A couple of renames for clarity:

* `ackConsumer.Clear()` to `ackConsumer.Cancel()`
* `commitCursorTracker.Done()` to `commitCursorTracker.UpToDate()`

### DIFF
--- a/pubsublite/internal/wire/acks.go
+++ b/pubsublite/internal/wire/acks.go
@@ -21,7 +21,10 @@ import (
 
 // AckConsumer is the interface exported from this package for acking messages.
 type AckConsumer interface {
+	// Ack the message.
 	Ack()
+	// Cancel ack processing for this message.
+	Cancel()
 }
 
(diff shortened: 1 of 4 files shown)


manuelmenzella-google (commented):
This is a bit scary because it allows users of wire.Subscriber to call Cancel() on a message's AckConsumer and leave the Subscriber in a stuck state, without it ever noticing. Perhaps we should wrap the AckConsumer provided by wire.Subscriber in another struct which knows to permanently fail the Subscriber if any ack is cancelled? We can sync offline, too.

PR #3254, opened 2020-11-23
test(pubsublite): integration test utils

tmdiep:
Utilities for use by integration tests:

* `MsgTracker` verifies that all published messages are received.
* `OrderingSender` / `OrderingReceiver` verifies that messages in a partition are ordered.
* `DuplicateMsgDetector` detects duplicate messages.


pubsublite/internal/test/msg_tracker.go
```diff
@@ -0,0 +1,83 @@
...
+// NewMsgTracker creates a new message tracker.
+func NewMsgTracker() *MsgTracker {
+	return &MsgTracker{
+		msgMap: make(map[string]bool),
+		done:   make(chan struct{}, 1),
+	}
+}
+
+// Add a set of tracked messages.
+func (mt *MsgTracker) Add(msgs ...string) {
```
manuelmenzella-google, line 39: IIUC, this type is designed to be used only once; that is, the user is not supposed to interleave calls to Add() and Remove(). Can we pass the full set of messages in NewMsgTracker() instead of having an Add() method?

pubsublite/internal/test/msg_validators.go
```diff
@@ -0,0 +1,172 @@
...
+	defer dm.mu.Unlock()
+
+	if (dm.duplicateReceiveCount + dm.duplicatePublishCount) == 0 {
+		return ""
+	}
+	return fmt.Sprintf("duplicate publish count = %d, receive count = %d", dm.duplicatePublishCount, dm.duplicateReceiveCount)
+}
+
+// HasDuplicates returns true if duplicate messages were detected.
+func (dm *DuplicateMsgDetector) HasDuplicates() bool {
```
manuelmenzella-google, line 168: Should we split this into HasReceiveDuplicates and HasPublishDuplicates? It sounds to me that publish duplicates are expected in normal operation, but receive duplicates are indicative of a serious bug.

manuelmenzella-google (approved):

pubsublite/internal/test/msg_validators.go
tmdiep: Good point. Done.

pubsublite/internal/test/msg_tracker.go
tmdiep: My longrunning integration tests loop and can reuse the same MsgTracker instance (to save having to add a mutex to its parent in order to replace it). Added a comment about correct usage of MsgTracker and pointed out the fact that Add and Remove are not supposed to be interleaved.

Issue #3104, opened 2020-10-28
logging: unflake tests


nicoleczhu:
Look into why tests were flaking (if it's due to slow log-list call because of missing timestamp/index to speed up the query) or some other issue
- [ ] --- SKIP: TestLogSync (0.00s)
- [ ] --- SKIP: TestLogAndEntries (0.00s)
- [ ] --- SKIP: TestStandardLogger (0.00s)

PR #3263, opened 2020-11-24
ci(logging): renabled un-flaked log tests

nicoleczhu:
fixes: #3104 

getEntries was optimized with the timestamp filter, from 10m per query to ~10secs. However, this PR still does add <30 seconds to the CI execution time, as we've renabled 3 tests.

### DIFF
--- a/logging/logging_test.go
+++ b/logging/logging_test.go
@@ -155,9 +155,6 @@ timestamp >= "%s"`,
 }
 
 func TestLogSync(t *testing.T) {
-	// TODO(deklerk): Un-flake and re-enable
-	t.Skip("Inherently flaky")
-
 	initLogs() // Generate new testLogID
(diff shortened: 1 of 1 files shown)


logging/logging_test.go
```diff
@@ -306,6 +300,7 @@ func entryForTesting(payload interface{}) *logging.Entry {
 	}
 }
 
+// Call this function sparingly. It takes ~10s to get logs, even with indexed filters.
```
tbpg, line 303: Totally minor since this isn't exported, but I generally try to follow the GoDoc pattern and start the comments with the thing being described. Something like:
```suggestion
// allTestLogEntries should be called sparingly. It takes ~10s to get logs, even with indexed filters.
```

tbpg (approved):
Thanks!

PR #3277, opened 2020-11-25
feat(pubsublite): routing publisher implementation

tmdiep:
`partitionCountWatcher` polls for updates to the number of topic partitions.

`routingPublisher` routes messages to multiple `singlePartitionPublishers` and handles the topic partition count increasing. Decreases will terminate the routingPublisher.

The initial set of partition publishers must successfully start before the routingPublisher accepts published messages. However, when new partition publishers are added later, messages may be routed to them while they are starting up. The messages are queued in the `publishMessageBatcher` and sent to the server when the Publish stream connects.

The entire `routingPublisher` tree terminates if the `partitionCountWatcher` or any of the `singlePartitionPublishers` terminate.

### DIFF
--- a/pubsublite/internal/wire/partition_count.go
+++ b/pubsublite/internal/wire/partition_count.go
@@ -78,7 +78,7 @@ func (p *partitionCountWatcher) Start() {
 func (p *partitionCountWatcher) Stop() {
 	p.mu.Lock()
 	defer p.mu.Unlock()
-	p.unsafeInitiateShutdown(serviceTerminated, nil)
+	p.unsafeInitiateShutdown(nil)
 }
 
 // updatePartitionCount is called in a goroutine.
@@ -101,12 +101,12 @@ func (p *partitionCountWatcher) updatePartitionCount() {
 		}
 		if err != nil {
 			err = fmt.Errorf("pubsublite: failed to update topic partition count: %v", err)
-			p.unsafeInitiateShutdown(serviceTerminated, err)
+			p.unsafeInitiateShutdown(err)
 			return 0, err
 		}
 		if resp.GetPartitionCount() <= 0 {
 			err := fmt.Errorf("pubsublite: topic has invalid number of partitions %d", resp.GetPartitionCount())
-			p.unsafeInitiateShutdown(serviceTerminated, err)
+			p.unsafeInitiateShutdown(err)
 			return 0, err
 		}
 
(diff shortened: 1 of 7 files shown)


tmdiep:
WANT_LGTM=@manuelmenzella-google, @palmere-google 
Other reviewers optional, but comments welcome.

If anyone feels strongly about waiting for the new partition publishers to successfully start before routing messages to them, feel free to let me know. It can be done, but introduces some complexity, so I went with this. If any partition publisher fails to start up, the entire routing publisher will be terminated anyway.

pubsublite/internal/wire/settings.go
```diff
@@ -77,6 +81,7 @@ var DefaultPublishSettings = PublishSettings{
 	// By default set to a high limit that is not likely to occur, but prevents
 	// OOM errors in clients.
 	BufferedByteLimit: 1 << 30, // 1 GiB
+	ConfigPollPeriod:  10 * time.Minute,
```
manuelmenzella-google, line 84: This feels a little long; did you chat with Evan to figure out what the right period is? I'm worried that a user would expect a faster reaction from their clients.

manuelmenzella-google (approved):

pubsublite/internal/wire/partition_count.go
```diff
@@ -0,0 +1,138 @@
...
+	p.mu.Unlock()
+
+	newPartitionCount, err := func() (int, error) {
+		req := &pb.GetTopicPartitionsRequest{Name: p.topicPath}
+		resp, err := p.adminClient.GetTopicPartitions(p.ctx, req, p.callOption)
+
+		p.mu.Lock()
+		defer p.mu.Unlock()
+
+		if p.status >= serviceTerminating {
```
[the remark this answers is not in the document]
tmdiep: Yes, that's exactly right, it's to prevent the receiver callback. Added comment.

pubsublite/internal/wire/settings.go
tmdiep: @palmere-google 

Yes, Evan said he'll set it to 10 min and I see that it has been for the Java client lib: https://github.com/googleapis/java-pubsublite/pull/387.

I remember in the design doc that the duration was somewhat influenced by the read-only admin operations quota limit. I implemented retry with backoff for the ResourceExhausted code in Go to mitigate this.

palmere-google: Sorry, I missed this

Yeah, I think with our current admin read limits, ten minutes is a good place to start. We can make the period smaller, but if we do, we should probably consider increasing the default admin read quota

palmere-google (approved):

pubsublite/internal/wire/settings.go
tmdiep: Ok thanks, I'll leave the default to 10 min for now.

PR #3281, opened 2020-11-26
feat(pubsublite): settings and message transforms for Cloud Pub/Sub shim

tmdiep:
Adds settings and message transformers for the Cloud Pub/Sub shim package.

The shim implements various features in the Lite Java client library for feature parity, e.g. nack handler, custom message transformers, key extractor, etc. It also implements some features from the Go pubsub library, e.g. publish buffer byte limit.


tmdiep:
WANT_LGTM= @hongalex, @manuelmenzella-google 

Suggestions for this package are welcome. I just called it `ps` to mirror `pstest` and don't have any better ideas, but think it may be too short.

pubsublite/ps/doc.go
```diff
@@ -0,0 +1,33 @@
...
+/*
+Package ps contains a publisher and subscriber client for the Cloud Pub/Sub Lite
+service that emulates the Cloud Pub/Sub API.
+
+If interfaces are defined for pubsub.Topic.Publish() and
+pubsub.Subscription.Receive(), the clients in this package can be used as
+drop-in replacements. As noted in comments, the two services have some
+differences:
+
+- Pub/Sub Lite does not support nack for messages.
```
manuelmenzella-google, line 23: I would also mention that Lite has no concept of "ack expiration;" you **must** call Ack on every message you receive, or shut down the subscriber client.

pubsublite/ps/message.go
```diff
@@ -0,0 +1,134 @@
...
+				msgpb.Attributes[key] = &pb.AttributeValues{Values: [][]byte{[]byte(value)}}
+			}
+		}
+	}
+	return msgpb, nil
+}
+
+// transformReceivedMessage is the default ReceiveMessageTransformerFunc
+// implementation.
+func transformReceivedMessage(from *pb.SequencedMessage, to *pubsub.Message) error {
```
manuelmenzella-google, line 94: Is there a reason why transformPublishedMessage returns a tuple of (message, error) and transformReceivedMessage takes the output message as a pointer, and just returns an error? Can we normalize this?

pubsublite/ps/settings.go
```diff
@@ -0,0 +1,187 @@
...
+type PublishMessageTransformerFunc func(*pubsub.Message) (*pb.PubSubMessage, error)
+
+// PublishSettings control the batching of published messages. These settings
+// apply per partition.
+//
+// Use DefaultPublishSettings for defaults, as an empty PublishSettings will
+// fail validation.
+type PublishSettings struct {
+	// Publish a non-empty batch after this delay has passed. Must be > 0.
+	DelayThreshold time.Duration
```
manuelmenzella-google, line 55: I see that these names match what's in the "wire" package, but I wonder if they should include "batching" in their names.

pubsublite/ps/settings.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// The maximum number of bytes that the publisher will keep in memory before
+	// returning ErrOverflow. Must be > 0.
+	//
+	// Note that Pub/Sub Lite topics are provisioned a publishing throughput
+	// capacity, per partition, shared by all publisher clients. Setting a large
+	// buffer size can mitigate transient publish spikes. However, consistently
+	// attempting to publish messages at a much higher rate than the publishing
+	// throughput capacity can cause the buffers to overflow. For more
+	// information, see https://cloud.google.com/pubsub/lite/docs/topics.
```
manuelmenzella-google, line 81: Is this something that Java/Python do, too?

manuelmenzella-google (approved):

pubsublite/ps/settings.go
tmdiep: I used identical names to CPS [pubsub.PublishSettings](https://godoc.org/cloud.google.com/go/pubsub#PublishSettings) (the `wire` settings were previously in the `pubsublite` package, and thus also inherited these names). Agree that having "batching" in the name makes it clearer and might rename the fields for the `wire` package in a separate PR - although I might err on the side of consistency, have yet to decide :)

pubsublite/ps/settings.go
tmdiep: Not Java. I haven't looked at Python closely, but I don't remember seeing it implemented there either.

It's for feature parity with the CPS Go client's [pubsub.PublishSettings.BufferedByteLimit](https://godoc.org/cloud.google.com/go/pubsub#PublishSettings).

pubsublite/ps/doc.go
tmdiep: Done

pubsublite/ps/message.go
tmdiep: Good point, normalized to output pointers.

transformReceivedMessage must be provided an output message because users will not be able to create a new `pubsub.Message` from scratch (we will need to create a private ack handler implementation).

tmdiep:
Sorry for the churn - I added a non-trivial amendment, as it's probably better to replicate CPS behavior and treat zero Publish/Receive settings as default settings. When we add more settings fields in the future, it's better to treat zero as defaults for back-compatibility.

pubsublite/ps/settings.go
hongalex: We added Publisher flow control in Pub/Sub for [Java](https://github.com/googleapis/java-pubsub/pull/119/files) and I think Python too. I'm currently adding this for Go, which will add `MaxOutstandingMessages` and `MaxOutstandingBytes`. We should have a discussion on how we're going to keep feature parity between the two libraries.

hongalex (approved):

pubsublite/ps/settings.go
tmdiep: Definitely; I've made a note to start a discussion once everyone is back from holidays.

P.S. Thanks for all the reviews!

PR #3283, opened 2020-11-26
fix(pubsublite): deliver messages sequentially

tmdiep:
Motivation: we want to match the Java Lite implementation, that delivers messages for a partition sequentially and block progress if the user's message receiver func blocks.

Messages were previously delivered to the user receiver func from the goroutine that processes the underlying Subscribe gRPC stream. If the Subscribe stream breaks, a new goroutine is started to handle a new stream instance, so the user's receiver func may be invoked while concurrently processing a previous invocation.

`messageDeliveryQueue` uses a channel to ensure that messages are delivered sequentially for a partition.

### DIFF
--- a/pubsublite/internal/wire/periodic_task.go
+++ b/pubsublite/internal/wire/periodic_task.go
@@ -59,10 +59,16 @@ func (pt *periodicTask) Stop() {
 
 func (pt *periodicTask) poll(ticker *time.Ticker, stop chan struct{}) {
 	for {
+		// stop has higher priority.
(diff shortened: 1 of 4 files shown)


manuelmenzella-google (approved):

tmdiep:
Thanks for reviewing!

I just realized that I forgot to add tests for the subscriber, particularly the case where the client-provided message receive func blocks. Added a commit to this PR, since it's just 2 extra unit tests.

PR #3286, opened 2020-11-27
feat(spanner/spannertest): support AVG aggregation function

### LINKED ISSUE #3285: spannertest: Support AVG
It's better if `spannertest` supports [`AVG`](https://cloud.google.com/spanner/docs/functions-and-operators#avg), one of the aggregated functions provided by Spanner. 

I feel it's not so difficult because the processes should be similar with "SUM" and "COUNT", which are already supported.

sryoya:
Fixes #3285 

Sorry for creating PR without waiting for the response to the above issue 😅

### DIFF
--- a/spanner/spannertest/funcs.go
+++ b/spanner/spannertest/funcs.go
@@ -128,9 +128,6 @@ var aggregateFuncs = map[string]aggregateFunc{
 					// "Returns NULL if the input contains only NULLs".
 					return nil, typ, nil
 				}
-				// floating point handling is non-deterministic
-				// type of the result from INT64 input is Float64
-				// https://cloud.google.com/spanner/docs/functions-and-operators#avg
 				return (float64(sum) / n), float64Type, nil
 			}
 			var sum float64
@@ -146,8 +143,6 @@ var aggregateFuncs = map[string]aggregateFunc{
 				// "Returns NULL if the input contains only NULLs".
 				return nil, typ, nil
 			}
-			// floating point handling is non-deterministic
-			// https://cloud.google.com/spanner/docs/functions-and-operators#avg
 			return (sum / n), typ, nil
 		},
 	},
--- a/spanner/spannertest/funcs.go
+++ b/spanner/spannertest/funcs.go
@@ -109,6 +109,48 @@ var aggregateFuncs = map[string]aggregateFunc{
 			return sum, typ, nil
 		},
 	},
+	"AVG": {
+		Eval: func(values []interface{}, typ spansql.Type) (interface{}, spansql.Type, error) {
+			if typ.Array || !(typ.Base == spansql.Int64 || typ.Base == spansql.Float64) {
+				return nil, spansql.Type{}, fmt.Errorf("AVG only supports arguments of INT64 or FLOAT64 type, not %s", typ.SQL())
+			}
+			if typ.Base == spansql.Int64 {
+				var sum int64
+				var n float64
+				for _, v := range values {
+					if v == nil {
+						continue
+					}
+					sum += v.(int64)
+					n++
+				}
+				if n == 0 {
+					// "Returns NULL if the input contains only NULLs".
+					return nil, typ, nil
+				}
+				// floating point handling is non-deterministic
+				// type of the result from INT64 input is Float64
+				// https://cloud.google.com/spanner/docs/functions-and-operators#avg
+				return (float64(sum) / n), float64Type, nil
+			}
+			var sum float64
+			var n float64
+			for _, v := range values {
+				if v == nil {
+					continue
+				}
+				sum += v.(float64)
+				n++
+			}
+			if n == 0 {
+				// "Returns NULL if the input contains only NULLs".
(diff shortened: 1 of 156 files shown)


spanner/spannertest/integration_test.go
```diff
@@ -889,6 +889,42 @@ func TestIntegration_ReadsAndQueries(t *testing.T) {
 				{int64(1), int64(25)}, // Jack(ID=1, Tenure=10), Sam(ID=3, Tenure=9), George(ID=5, Tenure=6)
 			},
 		},
+		{
```
dsymonds, line 892: I think there's too many test cases here. This doesn't need to exhaustively test, just give it a quick exercise. Maybe just grab one of them from https://cloud.google.com/spanner/docs/aggregate_functions#avg.

spanner/spannertest/funcs.go
```diff
@@ -109,6 +109,48 @@ var aggregateFuncs = map[string]aggregateFunc{
...
+					}
+					sum += v.(int64)
+					n++
+				}
+				if n == 0 {
+					// "Returns NULL if the input contains only NULLs".
+					return nil, typ, nil
+				}
+				// floating point handling is non-deterministic
+				// type of the result from INT64 input is Float64
```
dsymonds, line 132: I don't think these comments are useful here.

spanner/spansql/keywords.go
```diff
@@ -135,6 +135,7 @@ var funcs = map[string]bool{
 	"MAX":       true,
 	"MIN":       true,
 	"SUM":       true,
+	"AVG":       true,
```
dsymonds, line 138: keep these in alphabetical order.

spanner/spansql/parser_test.go
```diff
@@ -111,6 +111,20 @@ func TestParseQuery(t *testing.T) {
 				},
 			},
 		},
+		{`SELECT AVG(PointsScored) AS total_points, FirstName, LastName AS surname FROM PlayerStats GROUP BY FirstName, LastName`,
```
dsymonds, line 114: this test case is really not needed.

dsymonds (commented):
Looks good overall, but just needs to be pared back.

spanner/spansql/keywords.go
sryoya: I didn't notice the order rule, thanks.
https://github.com/googleapis/google-cloud-go/pull/3286/commits/64453991b15679ec55ac385d6a943f3bf999b2b2

spanner/spannertest/funcs.go
sryoya: Ok. As I just brought them from the official doc, we may not need to keep it in the code.
https://github.com/googleapis/google-cloud-go/pull/3286/commits/d9b8efb9a03562c1638653fe74fca69275fb9ce3

spanner/spannertest/integration_test.go
sryoya: Sorry, I didn't understand the test policy about how exhaustively tests should be implemented. I kept only one simple test and removed the others.

Btw, I tried to use either of the sample SQLs written in the official doc, but they don't seem to work correctly. 
I raised that as another issue here.
https://github.com/googleapis/google-cloud-go/issues/3296

So as I said in the above, I kept one original test case using the prepared data in this test func for the time being.

spanner/spansql/parser_test.go
sryoya: Removed.
https://github.com/googleapis/google-cloud-go/pull/3286/commits/e514c20b4199458ab35272eebc11cf0bd0f05efb

sryoya:
@dsymonds Thanks for your review. I updated the points you commented. Can you check them again?

spanner/spannertest/integration_test.go
dsymonds: That's fine, but you can still use that example, just slightly modified. There's other test cases here like that. We don't want to tie too many test cases to the built-up test data, as if that ever changes then there's lots of tests that need adjusting; for something like `AVG` it should suffice to see it exercised in an array provided as a param.

sryoya: Though I added the related comment in #3296 , I cannot make the SQL work as it is in the official doc.

But,
> We don't want to tie too many test cases to the built-up test data, as if that ever changes then there's lots of tests that need adjusting; for something like AVG it should suffice to see it exercised in an array provided as a param.

I completely agree with that. Actually, I feel that when I was implementing. So, I want to somehow create the test that doesn't depend on the built-up test data.

sryoya: I tried to work out the case that can test "AVG" without the built-up data, but, unfortunately, I could not get that from the currently supported SQLs. 
Let me know if you have any.

dsymonds: Sure, alright. Let's just use the existing data for this.

dsymonds (dismissed):
LGTM

dsymonds (approved):

Issue #3327, opened 2020-12-02
spanner: support big.Rat as NUMERIC in mutation


apstndb:
**Is your feature request related to a problem? Please describe.**

Query parameters accept `big.Rat` and their variants as `NUMERIC` typed value
but mutations doesn't accept `big.Rat` and their variants and emit an error.
```
spanner: code = "InvalidArgument", desc = "client doesn't support type *big.Rat"
```
This behavior is consistent with the documentation, but it seems strange that mutations only don't support it.
https://github.com/googleapis/google-cloud-go/blob/spanner/v1.12.0/spanner/mutation.go#L105-L121

**Describe the solution you'd like**

Mutations can support `NUMERIC` and `ARRAY<NUMERIC>` columns.
https://github.com/googleapis/google-cloud-go/blob/spanner/v1.12.0/spanner/value.go#L3061-L3071

**Describe alternatives you've considered**

Some mutations can be rewritten by DMLs with/without query parameters but they are not exactly same behavior.

skuruppu:
Thanks for the report and for the fix. I will let @olavloite review it since he is more familiar with this code.

skuruppu:
Thanks for your work @apstndb and @olavloite. I think this issue is now fixed so marking it as closed but please reopen if I missed something.

PR #3328, opened 2020-12-02
feat(spanner): support NUMERIC in mutations

apstndb:
This PR adds support of NUMERIC type in mutations.
ref #3327


olavloite (approved):
Good catch, thanks for fixing this.

It would have been good to have a test case that would note this in case there are new data types added in the future, but apparently that is missing. We can add that through a separate PR.

olavloite:
@tbpg Do you have any idea why the kokoro run seems to be hanging indefinitely? I've tried to trigger the execution a number of times, but it never seems to finish (or report a status).

tbpg:
@olavloite, sorry, just saw this. I added the label to start Kokoro. Must have missed the initial webhook or something.

olavloite:
> @olavloite, sorry, just saw this. I added the label to start Kokoro. Must have missed the initial webhook or something.

No problem, thanks for looking into it :-)

PR #3333, opened 2020-12-02
refactor(pubsublite): routing publisher ignores decreasing partition count

tmdiep:
There is a non-zero chance that the partition count decreases due to stale db reads.

Updated tests. Also fixed a flaky test: https://github.com/googleapis/google-cloud-go/issues/3332.


tmdiep:
Adding @manuelmenzella-google, as Evan won't be able to approve this PR without write access.

palmere-google (approved):

codyoss (approved):
Might be good to create an issue to track the TODO if there is not one already.

PR #3341, opened 2020-12-03
feat(spanner): Add support for RPC priority

olavloite:
Adds support for setting a priority per RPC. This can be any read/query/update statement and it can be set for a commit RPC.


skuruppu (approved):

olavloite:
@skuruppu and @larkee Please take a final look, this should now also be ready for merging.

spanner/client.go
```diff
@@ -509,6 +511,14 @@ func ApplyAtLeastOnce() ApplyOption {
 	}
 }
 
+// Priority returns an ApplyOptions that sets the RPC priority to use for the
+// commit operation.
+func Priority(priority sppb.RequestOptions_Priority) ApplyOption {
```
codyoss, line 516: Maybe this is not true for spanner, but I don't think we directly expose the raw protos in other manual layers. Is there a Priority struct in this package that can be used and mapped to instead?

spanner/transaction.go
```diff
@@ -135,26 +142,31 @@ func (t *txReadOnly) ReadWithOptions(ctx context.Context, table string, keys Key
 	}
 	index := ""
 	limit := 0
+	var ro *sppb.RequestOptions
 	if opts != nil {
 		index = opts.Index
 		if opts.Limit > 0 {
 			limit = opts.Limit
 		}
+		if opts.Priority != sppb.RequestOptions_PRIORITY_UNSPECIFIED {
```
hengfengli, line 151: nits: can this be replaced by `createRequestOptions()`? The logic looks a bit similar.

spanner/client.go
olavloite: Specifically for proto `enum` (option) values we do expose the proto values do avoid having to manually add and maintain a copy of the enum in the manual client. See for example these options: https://github.com/googleapis/google-cloud-go/blob/5fa2f7516d0c46a4c5fe8660e8138edc11afad17/spanner/transaction.go#L231-L234

spanner/transaction.go
olavloite: Good point, and that also applies to a number of other places as well. I've introduced a generic `requestPrioritizer` interface that can be used to generate the request options. This might some tweaking when the tagging feature can be merged as well, as both will generate `RequestOptions`, but as it is all internal interfaces it should not really be a problem.

hengfengli (approved):
LGTM.

Issue #3348, opened 2020-12-03
bigtable: Simulate latency/errors in Bigtable emulator for performance / DiRT testing


AdamBSteele:
**Is your feature request related to a problem? Please describe.**
I want to use the Cloud Bigtable emulator to test my application's behavior when my Bigtable instance is returning high latency or error rates.

**Describe the solution you'd like**
Add new settings to the Bigtable emulator for inject latency / errors into grpc calls.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

telpirion:
Performance tests will be addressed by the [Cloud Bigtable clients test](https://github.com/googleapis/cloud-bigtable-clients-test) project.

bhshkh:
[Cloud Bigtable clients test](https://github.com/googleapis/cloud-bigtable-clients-test) project addressed conformance testing and not performance tests.

PR #3349, opened 2020-12-03
feat(bigtable): Add emulator flags to simulate gRPC latency

AdamBSteele:
Add `introduce_grpc_latency` flags to the Cloud Bigtable emulator so users can test their application's behavior when Bigtable response times are high.


AdamBSteele:
Issue for linking: https://github.com/googleapis/google-cloud-go/issues/3348

bigtable/cmd/emulator/cbtemulator.go
```diff
@@ -21,30 +21,71 @@ import (
...
 var (
-	host = flag.String("host", "localhost", "the address to bind to on the local machine")
-	port = flag.Int("port", 9000, "the port number to bind to on the local machine")
+	host                   = flag.String("host", "localhost", "the address to bind to on the local machine")
+	port                   = flag.Int("port", 9000, "the port number to bind to on the local machine")
+	introduce_grpc_latency = flag.Bool("introduce_grpc_latency", false, "Introduce gRPC latency for testing")
+	grpc_read_p50          = flag.Duration("grpc_p50_read", time.Duration(0), "Target gRPC latency for ReadRows method at p50. Valid if introduce_grpc_latency is set.")
+	grpc_read_p99          = flag.Duration("grpc_p99_read", time.Duration(0), "Target gRPC latency for ReadRows method at p99. Valid if introduce_grpc_latency is set.")
+	grpc_write_p50         = flag.Duration("grpc_p50_write", time.Duration(0), "Target gRPC latency for MutateRows method at p50. Valid if introduce_grpc_latency is set.")
+	grpc_write_p99         = flag.Duration("grpc_p99_write", time.Duration(0), "Target gRPC latency for MutateRows method at p99. Valid if introduce_grpc_latency is set.")
```
igorbernstein2, line 39: To be complete there should be knobs for singular MutateRow, CheckAndMutate and ReadModifyWrite. In terms of ui, I think it would be nicer to do something a bit more generic:

`--inject-latency="MutateRow:p50:10ms" --inject-latency="MutateRow:p99:100ms" `

bigtable/cmd/emulator/cbtemulator.go
```diff
@@ -21,30 +21,71 @@ import (
...
 		grpc.MaxSendMsgSize(maxMsgSize),
 	}
+	if *introduce_grpc_latency {
+		opts = append(opts, grpc.StreamInterceptor(introduceStreamLatency))
+	}
+
 	srv, err := bttest.NewServer(fmt.Sprintf("%s:%d", *host, *port), opts...)
 	if err != nil {
-		log.Fatalf("failed to start emulator: %v", err)
+		log.Fatalf("Failed to start emulator: %v", err)
```
igorbernstein2, line 88: go style: `Error strings should not be capitalized (unless beginning with an exported name, a proper noun or an acronym) and should not end with punctuation.`

igorbernstein2 (commented):
Also, please add a test :)

igorbernstein2:
Also, consider moving this to bttest/ so that it can be used programmatically

bigtable/cmd/emulator/cbtemulator.go
AdamBSteele: Thanks for feedback - updated code to follow suggestions.  Please lmk if this approach looks reasonable. Will add tests if so

bigtable/cmd/emulator/cbtemulator.go
```diff
@@ -37,11 +38,18 @@ const (
 
 func main() {
 	grpc.EnableTracing = false
+	flag.Var(&emulatorInterceptorBuilder.LatencyTargets, "inject-latency", "Introduce gRPC latency for testing like \"ReadRows:p50:50ms\"")
+	flag.Var(&emulatorInterceptorBuilder.GrpcErrorCodeTargets, "inject-errors", "Introduce gRPC error codes for testing like \"ReadRows:10%:14\"")
```
igorbernstein2, line 42: Can we keep all of the flags grouped together? Currently some flags are defined as globals while others are local vars in main

bigtable/bttest/grpc_interceptor.go
```diff
@@ -0,0 +1,294 @@
+/*
```
igorbernstein2, line 1: consider renaming this file so that its name is scoped to error injection

bigtable/bttest/grpc_interceptor.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func (esib *EmulatorInterceptorBuilder) BuildStreamInterceptor() grpc.ServerOption {
+	return grpc.StreamInterceptor(esib.interceptorFunc())
+}
+
+func (esib *EmulatorInterceptorBuilder) interceptorFunc() grpc.StreamServerInterceptor {
+	fmt.Println("Building Stream Server Interceptor:")
+	if len(esib.LatencyTargets) > 0 {
+		fmt.Printf(" - Latency Targets: %s\n", esib.LatencyTargets.String())
```
igorbernstein2, line 55: Should these be logs?

bigtable/bttest/grpc_interceptor.go
```diff
@@ -0,0 +1,294 @@
...
+
+		startTime := time.Now()
+
+		// Latency injection
+		pVal := rand.Int31n(100)
+		for _, lt := range esib.LatencyTargets {
+			if strings.HasSuffix(info.FullMethod, lt.methodSuffix) && pVal >= lt.percentile {
+				addedLatency := time.Until(startTime.Add(lt.expectedDuration))
+				if addedLatency > time.Duration(0) {
+					log.Printf("Sleeping %v for Latency Target [%s]", addedLatency, lt.String())
```
igorbernstein2, line 74: This might get very verbose.

bigtable/bttest/grpc_interceptor.go
```diff
@@ -0,0 +1,294 @@
...
+		err := handler(srv, ss)
+		if err != nil {
+			return err
+		}
+
+		// If no actual errors, run error injection
+		eRand := rand.Int31n(100)
+		for _, gt := range esib.GrpcErrorCodeTargets {
+			if strings.HasSuffix(info.FullMethod, gt.methodSuffix) && eRand <= gt.errorRate {
+				log.Printf("Injecting Emulator Error for Error Target %s", gt.String())
```
igorbernstein2, line 90: Same as above

bigtable/bttest/grpc_interceptor.go
```diff
@@ -0,0 +1,294 @@
...
+
+/*
+	grpcErrorCodeTargets define how often each method should throw a given GRPC Error Code
+	e.g. "ReadRows:10%:14" will have a 10% chance to throw Error Code 14 (Unavailable) on each ReadRows request
+*/
+
+type grpcErrorCodeTargets []grpcErrorCodeTarget
+
+// Set() used by Flags lib to create new grpc error code target
+func (ets *grpcErrorCodeTargets) Set(s string) error {
```
igorbernstein2, line 202: Add?

bigtable/bttest/grpc_interceptor.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+type grpcErrorCodeTarget struct {
+	methodSuffix  string
+	errorRate     int32
+	grpcErrorCode codes.Code
+}
+
+// If method suffix matches and errorRandValue falls in errorRate, inject error
+func (gt *grpcErrorCodeTarget) getError(errorRandValue int32, fullMethod string) error {
```
igorbernstein2, line 226: This looks unused

bigtable/bttest/grpc_interceptor.go
```diff
@@ -0,0 +1,294 @@
...
+	if err != nil || (i < 0 || i > 100) {
+		return fmt.Errorf("Invalid error rate: %s. Expected integer between 0 and 100", s)
+	}
+	gt.errorRate = int32(i)
+	return nil
+}
+
+func (gt *grpcErrorCodeTarget) setGrpcErrorCode(s string) error {
+	var c codes.Code
+	err := c.UnmarshalJSON([]byte(s))
```
igorbernstein2, line 284: Please leave a commen there. I'm a bit surprised to see mention mentioned

bigtable/bttest/grpc_interceptor.go
```diff
@@ -0,0 +1,294 @@
...
+		ss grpc.ServerStream,
+		info *grpc.StreamServerInfo,
+		handler grpc.StreamHandler) error {
+
+		startTime := time.Now()
+
+		// Latency injection
+		pVal := rand.Int31n(100)
+		for _, lt := range esib.LatencyTargets {
+			if strings.HasSuffix(info.FullMethod, lt.methodSuffix) && pVal >= lt.percentile {
```
igorbernstein2, line 71: I think the list needs to be sorted for this work. Otherwise something like `ReadRows:p50:10ms`, `ReadRows:p99:100ms` will never experience p99 latency

bigtable/bttest/grpc_interceptor.go
```diff
@@ -0,0 +1,294 @@
...
+		// Return any actual errors from handler
+		err := handler(srv, ss)
+		if err != nil {
+			return err
+		}
+
+		// If no actual errors, run error injection
+		eRand := rand.Int31n(100)
+		for _, gt := range esib.GrpcErrorCodeTargets {
+			if strings.HasSuffix(info.FullMethod, gt.methodSuffix) && eRand <= gt.errorRate {
```
igorbernstein2, line 89: Similar sort issue as above

bigtable/bttest/grpc_interceptor.go
AdamBSteele: I don't think we need to sort latency. The code iterates through all LatencyTargets regardless of whether a match has happened.  With `ReadRows:p50:10ms`, `ReadRows:p99:100ms`, we'll sleep for ~10ms then sleep for ~90ms.

Errors definitely should be sorted, since the code will return when it finds a matching target.  Added new function/test to correctly handle this case

bigtable/bttest/grpc_interceptor.go
AdamBSteele: 👍

bigtable/bttest/grpc_interceptor.go
AdamBSteele: +1 that Add() is a better name.
Used `Set()` to satisfy the ["Value" interface](https://golang.org/src/flag/flag.go?s=7440:7500#L281) from the flags package

bigtable/cmd/emulator/cbtemulator.go
AdamBSteele: We're defining the `emulatorInterceptorBuilder` in vars, but I think we have to wait until main to bind those vars using `flag.Var`

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,315 @@
...
+var (
+	validStreamMethodSuffixes = []string{"ReadRows", "MutateRows"}
+)
+
+/*
+   Creates interceptors inject latency or errors into cbtemulator
+*/
+type EmulatorInterceptorBuilder struct {
+	LatencyTargets       latencyTargets
+	GrpcErrorCodeTargets grpcErrorCodeTargets
```
igorbernstein2, line 46: not a golang expert, but this feels weird: the variable name is public, but the type is private

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,315 @@
...
+/*
+   Stack error targets are so that arguments like:
+     - [ReadRows:20%:12, ReadRows:10%:14]
+   transform to:
+     - [ReadRows:20%:12, ReadRows:30%:14]
+   Thus, when we check against randint(100), we'll
+     throw the first error 20% of the time and the second error 10% of the time
+*/
+func (esib *EmulatorInterceptorBuilder) stackGrpcErrorCodeTargets() {
+	sort.SliceStable(esib.GrpcErrorCodeTargets, func(i, j int) bool {
```
igorbernstein2, line 62: The side effect of mutating the underlying data feels a bit hacky here and will have really bad consequences if someone calls interceptorFunc() multiple times (and I wont blame them because it looks like a getter :) )

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,315 @@
...
+	log.Println("Building Stream Server Interceptor:")
+	if len(esib.LatencyTargets) > 0 {
+		log.Printf(" - Latency Targets: %s\n", esib.LatencyTargets.String())
+	}
+	if len(esib.GrpcErrorCodeTargets) > 0 {
+		log.Printf(" - Error Targets: %s\n", esib.GrpcErrorCodeTargets.String())
+		esib.stackGrpcErrorCodeTargets()
+	}
+
+	return func(srv interface{},
```
igorbernstein2, line 82: This feels like a bit too much logic for an anonymous lambda.

bigtable/bttest/grpc_error_injection.go
igorbernstein2: Also I dont quite understand the need for stacking them

bigtable/bttest/grpc_error_injection.go
AdamBSteele: Goal here is to ensure that LatencyTargets and GrpcErrorCodeTargets can be used by `flags.Var` in `cbtemulator.go`, while keeping the underlying types private.

It feels a bit weird to me as well - open to suggestions

bigtable/bttest/grpc_error_injection.go
AdamBSteele: Good call - moved the stacking logic into the `Set()` function.

bigtable/bttest/grpc_error_injection.go
AdamBSteele: Removed all the logic that was in front of `return func(...`

I feel like there should be a way to express this without using that anonymous function, but I am totally blanking on the how it would be done.

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,307 @@
...
+	return strings.Join(s, ", ")
+}
+
+type latencyTarget struct {
+	methodSuffix     string
+	percentile       float64
+	expectedDuration time.Duration
+}
+
+// Create new latencyTarget from string like "ReadRows:p50:100ms"
```
tritone, line 128: This seems not great to me as a way to read in and pass around data.

Why not have the latency targets specified in a json file? Then you can easily define unmarshaling into language types using json annotations and skip all the parsing logic.

bigtable/cmd/emulator/cbtemulator.go
```diff
@@ -37,11 +38,18 @@ const (
 
 func main() {
 	grpc.EnableTracing = false
+	flag.Var(&emulatorInterceptor.LatencyTargets, "inject-latency", "Introduce gRPC latency for testing like \"ReadRows:p99.9:100ms\"")
+	flag.Var(&emulatorInterceptor.GrpcErrorCodeTargets, "inject-errors", "Introduce gRPC error codes for testing like \"ReadRows:12%:14\"")
```
tritone, line 42: Is the idea that you would need to separately spin up an emulator for each latency/error target that you want to test? I'm not super familiar with how the bigtable emulator is used in practice but just wondering if this will be acceptable to test a variety of latency/error behaviors.

I work mostly on the storage client (which is HTTP based) and we've done some similar work to inject errors in an emulator, however in what we do there is pass in the instructions in the request header so they can be executed on a per-request basis.

bigtable/cmd/emulator/cbtemulator.go
```diff
@@ -37,11 +38,18 @@ const (
 
 func main() {
 	grpc.EnableTracing = false
+	flag.Var(&emulatorInterceptor.LatencyTargets, "inject-latency", "Introduce gRPC latency for testing like \"ReadRows:p99.9:100ms\"")
```
tritone, line 41: I think the docs here need a bit more detail, it's not obvious to me looking at this what the example you specify will actually do in terms of emulator behavior.

bigtable/bttest/grpc_error_injection.go
tritone: You can also name a func and pass it around in Go, but it doesn't seem atypical to have an anonymous one as here either.

bigtable/bttest/grpc_error_injection.go
tritone: Agreed that it doesn't seem super well-encapsulated as is. Maybe you could use a constructor function, or a setter method for the private field? Or if you take my suggestion about using JSON then you can just have a constructor that takes the filename and then the fields can be private.

tritone (commented):
A couple questions, let me know what you think...

bigtable/cmd/emulator/cbtemulator.go
AdamBSteele: You would need to spin up an emulator for any set of latency/error targets (args can be supplied multiple times like `--inject-latency=ReadRows:p50:100ms --inject-latency=ReadRows:p99:1s`).

I do like the idea of passing instructions on a per-request basis. I'm not sure how we might do that Bigtable Client Libraries / gRPC. (Maybe UserAgent?) @igorbernstein2 wdyt

igorbernstein2: I can think of 2 alternatives:
1. pass the params as a header metadata. This would require the client to be able to inject headers, which I dont think is possible across all clients
2. expose a grpc service from the emulator that allows you to configure this via rpcs. Which I think it cleaner, but also more noisy. 

I personally like the simplicity of configuring via cli args and dont think that spinning up multiple instances is terrible

tritone: Sure, that's fine with me as long as you think it's reasonable-- I just raised it because I'd heard complaints in other clients about emulator spin up time and I wasn't sure how big a typical config matrix for tests using this feature might be.

bigtable/bttest/grpc_error_injection.go
enocom: +1 to using a constructor function. This way clients (not just the emulator) have a clear means to instantiating an `EmulatorInterceptor`, as opposed to just relying on an invocation of the emulator from the command line.

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,307 @@
...
+
+/*
+   Creates interceptors inject latency or errors into cbtemulator
+*/
+type EmulatorInterceptor struct {
+	LatencyTargets       latencyTargets
+	GrpcErrorCodeTargets grpcErrorCodeTargets
+}
+
+func (esib *EmulatorInterceptor) StreamInterceptor() grpc.ServerOption {
```
enocom, line 48: Small nit: What is `esib`? Perhaps `ei` or just `e` would be adequate? Cf. https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,307 @@
...
+	log.Println("Building Stream Server Interceptor:")
+	if len(esib.LatencyTargets) > 0 {
+		log.Printf(" - Latency Targets: %s\n", esib.LatencyTargets.String())
```
enocom, line 51: Because you've defined a String method on the `latencyTarget` type below, this call to `String()` is unnecessary and will be handled for you by the `%s` verb.

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	if len(esib.GrpcErrorCodeTargets) > 0 {
+		log.Printf(" - Error Targets: %s\n", esib.GrpcErrorCodeTargets.String())
```
enocom, line 54: Ditto above.

bigtable/bttest/grpc_error_injection.go
enocom, line 48: Also, since this is an exported identifier, please add a brief GoDoc to describe what it does.

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,307 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var (
+	validStreamMethodSuffixes = []string{"ReadRows", "MutateRows"}
+)
+
+/*
+   Creates interceptors inject latency or errors into cbtemulator
```
enocom, line 41: GoDoc comments should start with the exported identifier that they're documenting, e.g., "EmulatorInterceptor injects latency..."

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,307 @@
...
+func (esib *EmulatorInterceptor) interceptorFunc() grpc.StreamServerInterceptor {
+	return func(srv interface{},
+		ss grpc.ServerStream,
+		info *grpc.StreamServerInfo,
+		handler grpc.StreamHandler) error {
+
+		startTime := time.Now()
+
+		// Latency injection
+		pVal := rand.Float64() * 100
```
enocom, line 69: Probably should seed the random number generator above. See https://pkg.go.dev/math/rand#pkg-overview.

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,307 @@
...
+		for _, lt := range esib.LatencyTargets {
+			if strings.HasSuffix(info.FullMethod, lt.methodSuffix) && pVal >= lt.percentile {
+				addedLatency := time.Until(startTime.Add(lt.expectedDuration))
+				if addedLatency > time.Duration(0) {
```
enocom, line 73: This can be just `addedLatency > 0`

bigtable/bttest/grpc_error_injection.go
enocom: Alos, it's common to use `//` for GoDoc comments instead of multi-line comments.

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,307 @@
...
+			if strings.HasSuffix(info.FullMethod, gt.methodSuffix) && eRand <= gt.stackedErrorRate {
+				return status.Error(gt.grpcErrorCode, "Injected Emulator Error")
+			}
+		}
+
+		return nil
+	}
+}
+
+/*
```
enocom, line 97: Ditto above about single line comment-style docstrings.

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,307 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+This file provides interceptors to hook into grpc streaming calls
```
enocom, line 18: Since there's already a package comment for bttest, this could move down to the EmulatorInterceptor docs.

bigtable/bttest/grpc_error_injection.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+/*
+   latencyTargets define how long it will take for a method
+   to return at different percentiles. e.g. ReadRows:p50:100ms
+*/
+
+type latencyTargets []latencyTarget
+
+// Set() and String() used by Flags lib
```
enocom, line 104: This should be in the form: "Set configures a latency target..." or whatever.

bigtable/bttest/grpc_error_injection.go
enocom: See https://blog.golang.org/godoc for details. There are a number of comments below that don't follow GoDoc. Not a huge problem, but worth fixing.

enocom (commented):
Left a few comments -- mostly about GoDoc style.

I vote for introducing a constructor to create an EmulatorIntereceptor, if only to make the code less reliant on the emulator as the only entry point.

telpirion:
@codyoss Let me check with our Bigtable stakeholders before I close this.

bhshkh:
@AdamBSteele , please resolve review comments from @enocom

bhshkh:
> @googleapis/api-bigtable If this PR is still wanted we should work to get it merged as it is quite old now.

Yes, it is still wanted

AdamBSteele:
Sorry - I have changed jobs a few times since this PR was created. Will not
be able to delicate time to resolving

Adam Steele

bigtable/bttest/grpc_error_injection.go
bhshkh: On removing String(), IDE throws warning:

log.Printf format %s has arg ei.LatencyTargets of wrong type cloud.google.com/go/bigtable/bttest.latencyTargets

enocom: I might be wrong then ;-)

codyoss:
Closing based on author response.

PR #3367, opened 2020-12-03
fix(pubsub): respect streamAckDeadlineSeconds with MaxExtensionPeriod

### LINKED ISSUE #3278: pubsub: StreamAckDeadlineSeconds should respect MaxExtensionPeriod
**Client**

PubSub

**Additional context**

[StreamAckDeadlineSeconds is being set to 60 seconds](https://github.com/googleapis/google-cloud-go/blob/f37f118c87d4d0a77a554515a430ae06e5852294/pubsub/pullstream.go#L53) which would prompt an undelivered message (that's sent to the client from pubsub but doesn't reach there) to be redelivered in 60 seconds. This deadline should respect the [MaxExtensionPeriod setting](https://github.com/googleapis/google-cloud-go/blob/f37f118c87d4d0a77a554515a430ae06e5852294/pubsub/subscription.go#L476) and should be set depending on MaxExtensionPeriod. The client is expecting redeliveries in a shorter amount of time than MaxExtensionPeriod

hongalex:
`StreamAckDeadlineSeconds` comes into effect when a stream shuts down prior to message modack (which the client library does automatically when a message is received). Users may want to lower this bound to decrease message redelivery times, which is specified by `MaxExtensionPeriod`.

This PR also refactors the field into the `pullOptions` struct, where it lies with the `maxExtension` field.

Fixes #3278


hongalex:
cc @beltran, who opened the issue for Java

shollyman (approved):
Extracting values from function args to struct fields looks straightforward.

nit: pullOptions may deserve some inline comments similar to messageIterator; as a casual reader I don't understand what maxExtension vs maxExtensionPeriod convey.

PR #3384, opened 2020-12-04
feat(bigtable): Add a DirectPath fallback integration test

mohanli-ml:
Add a DirectPath fallback test for bigtable in Golang. The logic of the test is:

1. First wait until DirectPath traffic is observed;
2. Blackhole DirectPath net, and wait until the client fallback to use CFE;
3. Unblackhole DirectPath net, and the client should upgrade back to use DirectPath again.
The DirectPath and CFE traffic are distinguished by peer IP.


bigtable/integration_test.go
```diff
@@ -2142,6 +2144,63 @@ func TestIntegration_AdminBackup(t *testing.T) {
 	}
 }
 
+// Blackhole directpath address to test fallback.
+func TestIntegration_DirectPathFallback(t *testing.T) {
```
tbpg, line 2148:
```suggestion
// TestIntegration_DirectPathFallback tests the fallback when the directpath address is unavailable.
func TestIntegration_DirectPathFallback(t *testing.T) {
```

bigtable/integration_test.go
```diff
@@ -2142,6 +2144,63 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+		minCompleteRPC = 40
+	)
+
+	countEnough := false
+	start := time.Now()
+	for !countEnough && time.Since(start) < 2*time.Minute {
+		for i := 0; i < numRPCsToSend; i++ {
+			_, _ = table.ReadRow(ctx, "jadams")
+			if _, useDp := isDirectPathRemoteAddress(testEnv); useDp != isBlackhole {
+				atomic.AddUint64(&numCount, 1)
```
tbpg, line 2195: I think I'm missing something -- is this run in parallel?

bigtable/integration_test.go
```diff
@@ -2259,3 +2318,38 @@ func isDirectPathRemoteAddress(testEnv IntegrationEnv) (_ string, _ bool) {
 	// DirectPath ipv6 can use either ipv4 or ipv6 traffic.
 	return remoteIP, strings.HasPrefix(remoteIP, directPathIPV4Prefix) || strings.HasPrefix(remoteIP, directPathIPV6Prefix)
 }
+
+func blackholeOrAllowDirectPath(testEnv IntegrationEnv, t *testing.T, blackholeDP bool) {
+	blackholeDpv6Cmd := "sudo ip6tables -I INPUT -s 2001:4860:8040::/42 -j DROP && sleep 5 && echo blackholeDpv6"
+	blackholeDpv4Cmd := "sudo iptables -I INPUT -s 34.126.0.0/18 -j DROP && sleep 5 && echo blackholeDpv4"
+	allowDpv6Cmd := "sudo ip6tables -I INPUT -s 2001:4860:8040::/42 -j ACCEPT && sleep 5 && echo allowDpv6"
+	allowDpv4Cmd := "sudo iptables -I INPUT -s 34.126.0.0/18 -j ACCEPT && sleep 5 && echo allowDpv4"
```
tbpg, line 2326: This makes me really nervous. I don't expect a test to need `sudo` or alter `iptables`. Is there another way to do this?

bigtable/integration_test.go
```diff
@@ -2142,6 +2144,63 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer cleanup()
+
+	if err := populatePresidentsGraph(table); err != nil {
+		t.Fatal(err)
+	}
+
+	// Precondition: wait for DirectPath to connect.
+	countEnough := exerciseDirectPath(ctx, testEnv, table /*blackholeDP = */, false)
```
tbpg, line 2161: Go style does not have inline comments like this -- it's usually a sign the function being called should be split into more than one function. Perhaps one for the `true` case and one for the `false` case. That would also hopefully make it easier to understand & debug.

bigtable/integration_test.go
```diff
@@ -2142,6 +2144,63 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+	// Make sure that the client will start reading from IPv6 again by sending new requests and
+	// checking the injected IPv6 counter has been updated.
+	blackholeOrAllowDirectPath(testEnv, t /*blackholeDP = */, false)
+	countEnough = exerciseDirectPath(ctx, testEnv, table /*blackholeDP = */, false)
+	if !countEnough {
+		t.Fatalf("Failed to fallback to CFE after blackhole DirectPath")
+	}
+}
+
+func exerciseDirectPath(ctx context.Context, testEnv IntegrationEnv, table *Table, isBlackhole bool) bool {
```
tbpg, line 2182: It's unclear to me what the return value is. Perhaps a different function name?

tbpg (changes requested):

bigtable/integration_test.go
mohanli-ml: Fixed. I remember I should let the first word in the comment the same as the function name. Thanks for this!

bigtable/integration_test.go
mohanli-ml: The bigtable client will open 4 channels by default (https://github.com/googleapis/google-cloud-go/blob/master/bigtable/bigtable.go#L78), so they could be run in parallel.

bigtable/integration_test.go
mohanli-ml: The test itself is a bit tricky because we want to create a bad situation to make the client think it can not receive dp response (i.e., blackholed), so that we can test fallback. Previously in java, we modified the grpc netty library to achieve the blackhole (code here: https://github.com/googleapis/java-bigtable/blob/master/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/DirectPathFallbackIT.java#L184), but this is very hacky and sorry I have no idea how to do a similar thing in Go.

bigtable/integration_test.go
mohanli-ml: Sorry I do not understand this comment. I found this test https://github.com/googleapis/google-cloud-go/blob/master/bigtable/integration_test.go#L2013 also have similar inline comments. So could you give me an example about how should I write comments? Thanks for help!

bigtable/integration_test.go
mohanli-ml: I changed the function name to `examineTraffic`. The function checks and returns if enough DirectPath or CFE traffic is seen within 2 minutes.

bigtable/integration_test.go
tritone: Agreed that this seems really problematic as-is.

I'm more familiar with the HTTP clients but if I were testing something similar there, I would look into either setting up a small mock server to hit for my test, or using a custom roundtripper at the transport layer to produce the behavior I was looking for. Is it possible to do something similar in grpc?

tritone (commented):
cc @kolea2

bigtable/integration_test.go
mohanli-ml: In this test, we want to create a low level failure on TCP/IP, and grpc does not have any API to do that. Besides, a mock server is more like what we will use in unit tests, but this is an integration test and we want to make it as real as possible, so we may not want to use a mock server. 

When we are running the tests, we have the ownership of the VM that runs the test (create the VM, use `sudo` to setup the VM, run the test, destroy the VM..), so it will fine to use `sudo` without worrying about environment consistency.

Also notice that users are not supposed to run this test, since users are not supposed to know if the traffic is DirectPath or CFE. I added a check for `AttemptDirectPath` flag at the beginning of the test. Does this looks good?

kolea2: From @igorbernstein2, this go test might help us out: https://github.com/grpc/grpc-go/blob/ff1fc890e43ac77e922f53a2cef396b3c6a8f2a1/interop/grpclb_fallback/client.go#L78

igorbernstein2: I agree with the feedback, tests should avoid mutating the machine state.
Ideally it should follow a pattern similar to the java impl:
https://github.com/googleapis/java-bigtable/blob/master/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/DirectPathFallbackIT.java

Where the network failures are inject in process. It appears that grpc-go has some hooks to do something similar via the DialContext. the grpclb_fallback test that @kolea2 linked to, seems to use those hooks to test this exact scenario

mohanli-ml: @kolea2 @igorbernstein2 Thanks for the comments! I read the grpclb_fallback test and still have some questions.

1. The Dialer is from Golang instead of grpc-go, see https://golang.org/pkg/net/#Dialer. I made a custom dialer to print the dial address, as in https://paste.googleplex.com/6601422297628672. But when I run the test, the test can pass, but no custom log was printed. Any idea why it is not working?

2. If we can make the custom dialer work, then how can I block the directpath ip address in the custom dialer? Any example or API that I can learn from?

3. 1. The grpclb_fallback test is in fact using shell command to blackhole ip, see https://github.com/grpc/grpc-go/blob/ff1fc890e43ac77e922f53a2cef396b3c6a8f2a1/interop/grpclb_fallback/client.go#L202-L207, which seems the same as what I want to do? And sorry I am still a bit confused why we should not use shell command in tests?

Thanks!

mohanli-ml:
I have an idea about the `iptables`. I still think we need to use `iptables` for blackhole directpath traffic (which is the same as Igor's example https://github.com/grpc/grpc-go/blob/ff1fc890e43ac77e922f53a2cef396b3c6a8f2a1/interop/grpclb_fallback/client.go#L46-L47). However, inspired by this example, I made the shell command flags, so that we can avoid them accidentally mess up machines. Please note that this test is supposed to only run internally on google3, instead of kokoro. What do you think? @tbpg @tritone @kolea2 @igorbernstein2

bigtable/integration_test.go
tbpg: This particular code will not run in parallel. The code you linked to is for pooling the underlying network connections.

bigtable/integration_test.go
```diff
@@ -81,6 +87,10 @@ func init() {
 		"The zone in which to create the new test instance.")
 	flag.StringVar(&instanceToCreateZone2, "it.instance-to-create-zone2", "us-east1-c",
 		"The zone in which to create a second cluster in the test instance.")
+	flag.StringVar(&blackholeDpv6Cmd, "it.blackhole-dpv6-cmd", "", "Command to make LB and backend addresses blackholed over dpv6")
+	flag.StringVar(&blackholeDpv4Cmd, "it.blackhole-dpv4-cmd", "", "Command to make LB and backend addresses blackholed over dpv4")
+	flag.StringVar(&allowDpv6Cmd, "it.allow-dpv6-cmd", "", "Command to make LB and backend addresses allowed over dpv6")
+	flag.StringVar(&allowDpv4Cmd, "it.allow-dpv4-cmd", "", "Command to make LB and backend addresses allowed over dpv4")
```
tbpg, line 93: Add a comment with the expected values for these flags?

bigtable/integration_test.go
```diff
@@ -2142,6 +2152,80 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer cleanup()
+
+	if !testEnv.Config().AttemptDirectPath {
+		return
+	}
+
+	if len(blackholeDpv6Cmd) == 0 {
+		t.Fatal("-it.blackhole-dpv6-cmd unset")
```
tbpg, line 2169: Looks like this will fail the test if the flags aren't passed in. Perhaps we should skip instead?

bigtable/integration_test.go
```diff
@@ -2142,6 +2152,80 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+	if len(allowDpv4Cmd) == 0 {
+		t.Fatal("-it.allowdpv4-cmd unset")
+	}
+
+	if err := populatePresidentsGraph(table); err != nil {
+		t.Fatal(err)
+	}
+
+	// Precondition: wait for DirectPath to connect.
+	countEnough := examineTraffic(ctx, testEnv, table /*blackholeDP = */, false)
```
tbpg, line 2186: I'm not sure what this `blackholdDP = ` comment means.

bigtable/integration_test.go
```diff
@@ -2259,3 +2343,33 @@ func isDirectPathRemoteAddress(testEnv IntegrationEnv) (_ string, _ bool) {
...
+			cmdRes := exec.Command("bash", "-c", blackholeDpv4Cmd)
+			out, _ := cmdRes.CombinedOutput()
+			t.Logf(string(out))
+		} else {
+			cmdRes := exec.Command("bash", "-c", allowDpv4Cmd)
+			out, _ := cmdRes.CombinedOutput()
+			t.Logf(string(out))
+		}
+	} else {
+		if blackholeDP {
```
tbpg, line 2359:
```suggestion
		}
		return
	}
	if blackholeDP {
```

https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow is related.

tbpg (changes requested):
I like the flag setup much more. Thanks!

bigtable/integration_test.go
mohanli-ml: Thanks for pointing out. I removed the atomic operation.

bigtable/integration_test.go
mohanli-ml: Comment added.

bigtable/integration_test.go
mohanli-ml: We have that 3 lines above:
```
if !testEnv.Config().AttemptDirectPath {
  return
}
```
If the test want to test DirectPath, then we can fail the test if no command is given.

bigtable/integration_test.go
mohanli-ml: Sorry I should make `/*blackholeDP=*/` after the comma, like `countEnough := examineTraffic(ctx, testEnv, table, /*blackholeDP = */false)`. I fixed this. Here I want to explain the meaning of the parameter, i.e., DirectPath is blackholed or not, because it will be confusing if I just put `true` and `false` there.

bigtable/integration_test.go
mohanli-ml: Fixed. Thanks for catching the format!

bigtable/integration_test.go
```diff
@@ -2142,6 +2152,80 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+	)
+
+	countEnough := false
+	start := time.Now()
+	for !countEnough && time.Since(start) < 2*time.Minute {
+		for i := 0; i < numRPCsToSend; i++ {
+			_, _ = table.ReadRow(ctx, "jadams")
+			if _, useDP := isDirectPathRemoteAddress(testEnv); useDP != expectDP {
+				numCount++
+			}
```
tbpg, line 2221: Does `useDP` ever change in this loop? It appears that it will always be the same value?

Also, why send any RPCs if this will never be true? Could we just return before doing anything?

Perhaps the caller should check `isDirectPathRemoteAddress` then this function doesn't need the `expectDP` argument?

bigtable/integration_test.go
```diff
@@ -2259,3 +2343,34 @@ func isDirectPathRemoteAddress(testEnv IntegrationEnv) (_ string, _ bool) {
 	// DirectPath ipv6 can use either ipv4 or ipv6 traffic.
 	return remoteIP, strings.HasPrefix(remoteIP, directPathIPV4Prefix) || strings.HasPrefix(remoteIP, directPathIPV6Prefix)
 }
+
+func blackholeOrAllowDirectPath(testEnv IntegrationEnv, t *testing.T, blackholeDP bool) {
```
tbpg, line 2347: Can we split this into two functions, one corresponding to `blackholeDP=true` and one for `false`?

That would get rid of the need for the `blackholeDP` argument at all, making calls more clear and easier to debug.

bigtable/integration_test.go
tbpg: Those types of comments are not ordinarily done in Go style, which is why `gofmt` doesn't handle them well -- they're uncommon and not designed for. I gave a few suggestions below of ways we could simplify.

tbpg (changes requested):
Mostly a Go style review. I'm not 100% sure I understand the goal of all of the code, so some questions are very open.

epope18 (approved):

bigtable/integration_test.go
```diff
@@ -2142,6 +2152,80 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+
+// examineTraffic counts RPCs use DirectPath or CFE traffic.
+func examineTraffic(ctx context.Context, testEnv IntegrationEnv, table *Table, expectDP bool) bool {
+	var numCount uint64
+	const (
+		numRPCsToSend  = 20
+		minCompleteRPC = 40
+	)
+
+	countEnough := false
```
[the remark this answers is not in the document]
mohanli-ml: Yeah, I have removed the variable.

bigtable/integration_test.go
```diff
@@ -2142,6 +2152,80 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+	blackholeOrAllowDirectPath(testEnv, t /*blackholeDP = */, false)
+	countEnough = examineTraffic(ctx, testEnv, table /*blackholeDP = */, false)
+	if !countEnough {
+		t.Fatalf("Failed to fallback to CFE after blackhole DirectPath")
+	}
+}
+
+// examineTraffic counts RPCs use DirectPath or CFE traffic.
+func examineTraffic(ctx context.Context, testEnv IntegrationEnv, table *Table, expectDP bool) bool {
+	var numCount uint64
```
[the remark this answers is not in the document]
mohanli-ml: Done.

bigtable/integration_test.go
```diff
@@ -2142,6 +2152,80 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+	countEnough := false
+	start := time.Now()
+	for !countEnough && time.Since(start) < 2*time.Minute {
+		for i := 0; i < numRPCsToSend; i++ {
+			_, _ = table.ReadRow(ctx, "jadams")
+			if _, useDP := isDirectPathRemoteAddress(testEnv); useDP != expectDP {
+				numCount++
+			}
+			time.Sleep(100 * time.Millisecond)
+			countEnough = numCount >= minCompleteRPC
```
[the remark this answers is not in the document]
mohanli-ml: Done.

bigtable/integration_test.go
mohanli-ml: `useDP` does get changed. `useDP` is returned by `isDirectPathRemoteAddress()`, which will return `true` if the traffic is DirectPath, and `false` if the traffic is CFE. So when DP is allowed, we expect `useDP` to be `true`, but when DP is blackholed, we expect `useDP` to be false. 

Sorry the name of the variable `expectDP` is misleading, I have rename it as `blackholeDP`. Thus, when `useDP != blackholeDP`, we get the traffic as expected. Specifically as described in the PR description, this test has three stages, as follows:

1. First wait until DirectPath traffic is observed;
In this stage, `useDP` should be `true`, and `blackholeDP` should be false.

2. Blackhole DirectPath net, and wait until the client fallback to use CFE;
In this stage, `useDP` should be `false`, and `blackholeDP` should be true.

3. Unblackhole DirectPath net, and the client should upgrade back to use DirectPath again.
The DirectPath and CFE traffic are distinguished by peer IP.
In this stage, `useDP` should be `true`, and `blackholeDP` should be false.

bigtable/integration_test.go
mohanli-ml: Done

bigtable/integration_test.go
```diff
@@ -2142,6 +2152,81 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+func TestIntegration_DirectPathFallback(t *testing.T) {
+	ctx := context.Background()
+	testEnv, _, _, table, _, cleanup, err := setupIntegration(ctx, t)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cleanup()
+
+	if !testEnv.Config().AttemptDirectPath {
+		return
```
tbpg, line 2165: Should this `return` or `t.Skip`? Return means the test "passes." So, skipping seems more correct.

bigtable/integration_test.go
```diff
@@ -2142,6 +2152,81 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+		t.Fatal("-it.allowdpv4-cmd unset")
+	}
+
+	if err := populatePresidentsGraph(table); err != nil {
+		t.Fatal(err)
+	}
+
+	// Precondition: wait for DirectPath to connect.
+	countEnough := examineTraffic(ctx, testEnv, table, false)
+	if !countEnough {
```
tbpg, line 2187: Perhaps this name could be more clear?

```suggestion
	dpEnabled := examineTraffic(ctx, testEnv, table, false)
	if !dpEnabled {
```

bigtable/integration_test.go
```diff
@@ -2259,3 +2344,33 @@ func isDirectPathRemoteAddress(testEnv IntegrationEnv) (_ string, _ bool) {
...
+		out, _ := cmdRes.CombinedOutput()
+		t.Logf(string(out))
+	} else {
+		cmdRes := exec.Command("bash", "-c", blackholeDpv4Cmd)
+		out, _ := cmdRes.CombinedOutput()
+		t.Logf(string(out))
+		cmdRes = exec.Command("bash", "-c", blackholeDpv6Cmd)
+		out, _ = cmdRes.CombinedOutput()
+		t.Logf(string(out))
+	}
```
tbpg, line 2360:
```suggestion
	cmdRes := exec.Command("bash", "-c", blackholeDpv4Cmd)
	out, _ := cmdRes.CombinedOutput()
	t.Logf(string(out))
	
	if testEnv.Config().DirectPathIPV4Only {
		return
	}
	cmdRes = exec.Command("bash", "-c", blackholeDpv6Cmd)
	out, _ = cmdRes.CombinedOutput()
	t.Logf(string(out))
```

bigtable/integration_test.go
```diff
@@ -2142,6 +2152,81 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+
+	// Disable the blackhole, and client should use DirectPath again.
+	allowDirectPath(testEnv, t)
+	countEnough = examineTraffic(ctx, testEnv, table, false)
+	if !countEnough {
+		t.Fatalf("Failed to fallback to CFE after blackhole DirectPath")
+	}
+}
+
+// examineTraffic counts RPCs use DirectPath or CFE traffic.
```
tbpg, line 2206:
```suggestion
// examineTraffic returns whether RPCs use DirectPath (blackholeDP = false) or CFE (blackholeDP = true).
```

Is that right? I'm not sure if there is a more clear way to say it.

bigtable/integration_test.go
mohanli-ml: Done

bigtable/integration_test.go
mohanli-ml: Done.

bigtable/integration_test.go
mohanli-ml: Done.

bigtable/integration_test.go
mohanli-ml: Yes this is correct.

PR #3388, opened 2020-12-04
test(pubsublite): fix flaky subscriber stream tests

tmdiep:
Fixes https://github.com/googleapis/google-cloud-go/issues/3329 and other similar instances.

The subscribeStream transitions to the active state before it terminates, however the state change notifications are sent asynchronously, so they can be received by observers out of order. In these test cases, we expect the subscribeStream to terminate due to fatal error, so it is only necessary to check FinalError().


tmdiep:
@codyoss - I just noticed that a few tests with the same root cause are flaky and causing CI to fail. Sorry!

codyoss (approved):
Thanks!

Issue #3330, opened 2020-12-02
bigtable: Make user agent the same as library version


mohanli-ml:
Currently the user agent of Bigtable in Go is standalone, as in https://github.com/googleapis/google-cloud-go/blob/master/bigtable/doc.go#L121m. Ideally, the user agent should be the same as the library release, which is v1.6.0 currently, see https://pkg.go.dev/cloud.google.com/go/bigtable.

jimfulton:
The fix in #3429 has rotted.  IDK enough about the release process to know how to automate this and prevent rot, _however_

Why should the user agent and server versions be in sync?

Issue #2749, opened 2020-08-19
internal: can we have a correct library version instead of a date? 


hengfengli:
Currently, the version of the entire client lib is pinned to a date: https://github.com/googleapis/google-cloud-go/blob/e455ed5db86b21a22e9a5b5018a90b74dfc4a392/internal/version/version.go#L27-L29. This can be confusing when comparing to other language client libs. 

What is the story/background behind this? Are we able to have a proper library version, e.g.,  `0.64.0`, instead of a date?

codyoss:
I don't have the context to why this is a date today, but I agree I think we can do better here. I will have to do a little research into what other languages are populating into this value. I assume a semver like `v0.64.0`.

crwilcox:
https://github.com/googleapis/google-cloud-go/issues/3330 related

codyoss:
I think this issue is for the most part complete. I would like to also get this hooked into the GAPIC layer more, but that will happen outside of this repo. Closing this issue for now as the original request of having better data in the manual is now true.

PR #3429, opened 2020-12-08
bigtable: Update user-agent to the latest version

mohanli-ml:
Currently the user agent of Bigtable in Go is standalone, as in https://github.com/googleapis/google-cloud-go/blob/master/bigtable/doc.go#L121m. Update it to the latest version https://pkg.go.dev/cloud.google.com/go/bigtable. Related issue: https://github.com/googleapis/google-cloud-go/issues/3330.

Background: DirectPath will use user agent to deny a specific library version, so we want to update the user agent with the library release date.


tritone (approved):
LGTM

@igorbernstein2 @tbpg is there any reason we can't either 1. Use https://github.com/googleapis/google-cloud-go/blob/40f29fe4298f1cb22cee0fe509ed95ac2ef7c531/internal/version/version.go#L29 for this, which is used by at least some other packages, or 2. Switch to using the release tag itself?

It's easy to forget to update this value as is (as evidenced by the fact that it hasn't been updated in 2.5 years).

codyoss:
@tritone I will be changing this soon to be the actual tag as a part of #2749.

codyoss:
@mohanli-ml Won't direct path be denied in older version automatically since the new option is not being sent, why is extra header validation needed? Also, like you mention in #3330 ideally this would be semver. If we switch this header to be semver you would no longer be able to parse a date for this feature.

mohanli-ml:
@codyoss Yes, older versions before directpath release will not support directpath. However, we need this user agent for future. For example, we might have 3 versions that can support DP, and we realize one version has some problem, so we want to deny that one, but leave the other two still use DP. And for semver, both semver and date can work, as long as the user agent is compatible with the library version. Let me know if you have more questions. Thanks!

codyoss:
@tritone Is this still waiting on anything? My comments were just that and don't want to hold anything up here.

tritone:
I'm fine with the switch to use the semver here. @igorbernstein2 @kolea2 can you confirm that this won't cause any backend or data collection issues that you know of?

kolea2 (approved):

PR #3438, opened 2020-12-09
feat(spanner/spannertest): support array literals

dsymonds:
These are relatively easy, and often appear with UNNEST.

### DIFF
--- a/spanner/spannertest/db_eval.go
+++ b/spanner/spannertest/db_eval.go
@@ -428,6 +428,17 @@ func (ec evalContext) evalExpr(e spansql.Expr) (interface{}, error) {
 		return bool(e), nil
 	case spansql.Paren:
 		return ec.evalExpr(e.Expr)
+	case spansql.Array:
+		var arr []interface{}
+		for _, elt := range e {
+			v, err := ec.evalExpr(elt)
+			if err != nil {
+				return nil, err
+			}
+			arr = append(arr, v)
+		}
+		// TODO: enforce or coerce to consistent types.
+		return arr, nil
 	case spansql.ArithOp:
 		return ec.evalArithOp(e)
 	case spansql.LogicalOp:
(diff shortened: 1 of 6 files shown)


skuruppu (approved):

spanner/spannertest/db_eval.go
```diff
@@ -774,14 +785,29 @@ func (ec evalContext) colInfo(e spansql.Expr) (colInfo, error) {
 		return colInfo{Type: qp.Type}, nil
 	case spansql.Paren:
 		return ec.colInfo(e.Expr)
+	case spansql.Array:
+		// Assume all element of an array literal have the same type.
+		if len(e) == 0 {
+			// TODO: What does the real Spanner do here?
```
olavloite, line 791: From https://cloud.google.com/spanner/docs/arrays#using_array_literals

You can write an empty array of a specific type using ARRAY<type>[]. You can also write an untyped empty array using [], in which case Cloud Spanner SQL attempts to infer the array type from the surrounding context. If Cloud Spanner SQL cannot infer a type, the default type ARRAY<INT64> is used.

So this seems to be correct.

spanner/spansql/parser.go
```diff
@@ -2707,12 +2731,12 @@ func (p *parser) parseOnDelete() (OnDelete, *parseError) {
 
 // parseCommaList parses a parenthesized comma-separated list,
```
olavloite, line 2732:
```suggestion
// parseCommaList parses a comma-separated list enclosed between bra and ket,
```

olavloite (approved):

spanner/spansql/parser.go
dsymonds: done.

spanner/spannertest/db_eval.go
dsymonds: I wasn't sure about the INT64 part of the type though. The real spanner uses INT64 in some situations (e.g. `SELECT NULL`), but I haven't tested out what it does with, say, `SELECT []`. That's all this covers.

PR #3439, opened 2020-12-09
test(pubsublite): enable integration tests for admin resource operations

tmdiep:
The server now accepts project ids in resource paths. However, it always returns the project number in returned configs, so we have to exclude the project field in comparisons.


hongalex (approved):
LGTM pending CI

PR #3442, opened 2020-12-10
feat(pubsublite): subscriber client

tmdiep:
`SubscriberClient` wraps Pub/Sub Lite's `wire.Subscriber` and emulates the [`pubsub.Subscription.Receive`](https://godoc.org/cloud.google.com/go/pubsub#Subscription.Receive) API.

Not mergeable until dependent PRs have been committed to master.


tmdiep:
WANT_LGTM=@dpcollins-google for Pub/Sub Lite logic, @hongalex for consistency with CPS. I've tried to replicate most of the documented behavior, e.g. waiting for all receiver funcs to finish before returning from `Receive()`.

pubsublite/ps/subscriber.go
```diff
@@ -0,0 +1,317 @@
...
+		return si.err
+	}
+	return err
+}
+
+// MessageReceiverFunc handles messages sent by the Cloud Pub/Sub Lite service.
+//
+// The implementation must arrange for pubsub.Message.Ack() or
+// pubsub.Message.Nack() to be called after processing the message. Since Cloud
+// Pub/Sub Lite has no concept of ACK deadlines, it is encouraged to process
```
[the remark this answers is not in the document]
tmdiep: Comments fixed, thanks! Also updated some comments for the settings to reiterate that they apply per-partition.

pubsublite/ps/subscriber.go
```diff
@@ -0,0 +1,313 @@
...
+			return nil, errDuplicateReceive
+		}
+		subInstance, err := newSubscriberInstance(ctx, s.wireSubFactory, s.settings, f)
+		if err != nil {
+			return nil, err
+		}
+
+		s.receiveActive = true
+		return subInstance, nil
+	}()
```
hongalex, line 299: nit: I'm not 100% I see the value in wrapping this inside an anonymous function. In each of the return statements here, we should just return the error directly to `Receive` correct?

pubsublite/ps/example_test.go
```diff
@@ -72,3 +80,61 @@ func ExamplePublisherClient_Error() {
...
 }
+
+func ExampleSubscriberClient_Receive() {
+	ctx := context.Background()
+	subscription := pubsublite.SubscriptionPath{
+		Project:        "project-id",
+		Zone:           "zone",
+		SubscriptionID: "subscription-id",
+	}
+	subscriber, err := ps.NewSubscriberClient(ctx, ps.DefaultReceiveSettings, subscription)
```
hongalex, line 91: Perhaps we should clarify that we don't need to pass in `ps.DefaultReceiveSettings` to achieve default functionality

hongalex (approved):
some minor nits

pubsublite/ps/subscriber.go
tmdiep: Agree this looks a little strange. I wanted to ensure subInstance.Wait() isn't holding the mutex - mostly just so that a duplicate Receive() call returns an error immediately rather than blocking.

Refactored this to be more readable.

pubsublite/ps/example_test.go
tmdiep: Good point. Added a note "DefaultReceiveSettings and empty ReceiveSettings{} are equivalent" and same for the PublishSettings example above.

PR #3443, opened 2020-12-10
test(bigtable): delete table retry fix

AlisskaPie:


chelleyyy:
Nicht erlaubter Zugriff

AlisskaPie:
While working on `BackupIAM` and `TableIam` tests, I've noticed that the table was actually deleted with `deleteTable` function but I had `Not Found` log error:
image

I've made an update using the logic in `retry` function:
> It returns when one of the following occurs:
> // When f's first return value is true, Retry immediately returns with f's second
> // return value.
> // When the provided context is done, Retry returns with an error that
> // includes both ctx.Error() and the last error returned by f.

tritone (approved):
I think this seems correct now.

kolea2 (approved):

PR #3444, opened 2020-12-10
feat(spanner): support CommitStats

olavloite:
Adds support for `CommitStats` to the Spanner client.

Based on https://github.com/googleapis/google-cloud-go/pull/3056.


skuruppu (approved):

olavloite:
@skuruppu The current build failure seems to be caused by a test project that does not (yet) support `CommitStats`. Running the integration tests against the Go cloud test project does not return any `CommitStats`, running the same tests against another dev project does return `CommitStats`.

skuruppu:
> @skuruppu The current build failure seems to be caused by a test project that does not (yet) support `CommitStats`. Running the integration tests against the Go cloud test project does not return any `CommitStats`, running the same tests against another dev project does return `CommitStats`.

Oh interesting. I think it's because it's not fully released in the backend. I'll give you a go ahead to rerun the tests when the release is complete.

Thanks for rebasing the PR :)

skuruppu:
Hmmm still failing the test but I don't understand why.

```
--- FAIL: TestIntegration_ReadWriteTransaction_StatementBasedWithOptions (0.13s)
    integration_test.go:1263: Missing commit stats in commit response
--- FAIL: TestIntegration_ReadWriteTransactionWithOptions (0.16s)
    integration_test.go:1086: Missing commit stats in commit response
```

skuruppu:
@olavloite I can confirm that equivalent tests pass for all other languages so would you be able to check if it's an issue with the Go implementation?

olavloite:
> @olavloite I can confirm that equivalent tests pass for all other languages so would you be able to check if it's an issue with the Go implementation?

Hmmm... This is strange. If I run the above mentioned integration tests from my local environment against a normal test instance, the integration tests succeed.

olavloite:
> > @olavloite I can confirm that equivalent tests pass for all other languages so would you be able to check if it's an issue with the Go implementation?
> 
> Hmmm... This is strange. If I run the above mentioned integration tests from my local environment against a normal test instance, the integration tests succeed.

Never mind, the problem is that the build is failing when it is running the integration tests against the emulator, and that does not support commit stats yet.

PR #3449, opened 2020-12-11
test(pubsublite): integration tests for publish and subscribe

tmdiep:
Integration tests for the publisher and subscriber clients of the CPS shim package.

Also checks for strict ordering in `OrderingReceiver` (disallows duplicates).


pubsublite/ps/integration_test.go
```diff
@@ -0,0 +1,673 @@
...
+	gibi               = 1 << 30
+	defaultTestTimeout = 5 * time.Minute
+)
+
+var (
+	resourceIDs = uid.NewSpace("go-ps-test", nil)
+	rng         *rand.Rand
+
+	// A random zone is selected for each integration test run.
+	supportedZones = []string{
```
manuelmenzella-google, line 48: Since we just turned up a few new zones, I wonder if you could include some of those here to get more coverage. I don't think we'll keep this in sync, but I'd like to have more confidence that the new zones are working well.

manuelmenzella-google (approved):
This is super thorough, thank you Thanh-Mai!

pubsublite/ps/integration_test.go
tmdiep: I ran the integration tests on each of the new zone and they all worked perfectly!

Good idea, added the new zones and also moved the list to test utils to share with the admin integration tests. Having more zones will reduce the likelihood of quota issues, if many integration tests happen to be running simultaneously.

PR #3456, opened 2020-12-14
test(spanner): configurable host for it tests

thiagotnunes:
Makes it possible to set a different spanner api host to be used for the integration tests.


mohanli-ml (approved):
LGTM. Thanks!

spanner/integration_test.go
```diff
@@ -165,6 +169,10 @@ func parseInstanceName(inst string) (project, instance string, err error) {
 	return matches[1], matches[2], nil
 }
 
+func getSpannerHost() (string) {
```
olavloite, line 172: This needs to be the following for the format check to stop complaining:

```suggestion
func getSpannerHost() string {
```

You can check the formatting by executing `gofmt -s .` in the `cloud.google.com/spanner` directory.

olavloite (approved):
LGTM, but needs a small change for the formatter to stop complaining.

thiagotnunes:
@olavloite thanks for spotting that!

PR #3458, opened 2020-12-14
feat(pubsublite): add client library metadata to headers

tmdiep:
Adds client library version and other properties to gRPC stream metadata for debugging purposes.


tmdiep:
@codyoss: We'd like to send the pubsublite library version to the server for debugging purposes. `debug.BuildInfo` works well for binaries built with modules, but do you know what options we have if GOPATH is used?

hongalex (approved):

pubsublite/internal/wire/rpc.go
```diff
@@ -164,18 +164,64 @@ func newPartitionAssignmentClient(ctx context.Context, region string, opts ...op
...
-	md = md.Copy()
-	val := fmt.Sprintf("partition=%d&topic=%s", topic.Partition, url.QueryEscape(topic.Path))
-	md[routingMetadataHeader] = append(md[routingMetadataHeader], val)
-	return metadata.NewOutgoingContext(ctx, md)
+const (
+	routingMetadataHeader    = "x-goog-request-params"
+	clientInfoMetadataHeader = "x-goog-pubsub-context"
+
+	languageKey     = "language"
+	languageValue   = "GOLANG"
```
codyoss, line 172: This should probably just be Go? Also you can tell a call came from a Go library already by looking at metadata in the `x-goog-api-client` header. If it is Go there will be a gl-go value present with the version of Go that is being used.

codyoss (commented):
Just one small comment or else seems good to me. Only other thing is that I think we need to document these additional headers somewhere(internal). That way things can be consistent across languages and people can tell what the values mean. It is very possible this is already done and I am oblivious though.

pubsublite/internal/wire/rpc.go
tmdiep: Thanks Cody. These values are for a pub/sub specific header x-goog-pubsub-context. I'll check if it's documented anywhere. "GOLANG" was already defined for Go and recognized on the server. I agree that it should have just been Go.

The servers haven't been inspecting x-goog-api-client, but I think it would be a good idea to use it as backup, for when the pubsub header is missing.

codyoss: Ack

PR #3461, opened 2020-12-14
fix(spanner): fix session leak

### LINKED ISSUE #3460: spanner: If panic occurs in ReadWriteTransaction, the session will leak
**Client**

Spanner

**Environment**

macOS Catalina

**Go Environment**

$ go version

```
go version go1.15.6 darwin/amd64
```

$ go env

**Code**

e.g.
```go
package main

import (
	"context"
	"fmt"
	"log"

	"cloud.google.com/go/spanner"
)

const db = "projects/your-project/instances/your-instance/databases/your-db"

func main() {
	ctx := context.Background()
	client, _ := spanner.NewClientWithConfig(ctx, db, spanner.ClientConfig{
		SessionPoolConfig: spanner.SessionPoolConfig{MaxOpened: 5},
	})

	for i := 0; i < 5; i++ {
		func() {
			defer func() { recover() }()

			_, err := client.ReadWriteTransaction(ctx, func(ctx context.Context, tx *spanner.ReadWriteTransaction) error {
				panic("test")
				return nil
			})
			if err != nil {
				log.Println(err)
			}
		}()
		fmt.Println(i)
	}

	log.Println("start")
	_, err := client.ReadWriteTransaction(ctx, func(ctx context.Context, tx *spanner.ReadWriteTransaction) error {
		return nil
	})
	log.Println("end")
	if err != nil {
		log.Println(err)
	}
}
```

**Expected behavior**

When a panic occurs in `ReadWriteTransaction`, the session will leak.
I want to solve this problem.

I think that I need to rebuild the client to recover from the leaked session, but this is difficult to take care of.

**Actual behavior**

This is the output result. The last result is not returned.

```
0
1
2
3
4
2020/12/14 21:54:28 start
```

If you set the loop count to 4, it will return.

```
0
1
2
3
2020/12/14 21:55:46 start
2020/12/14 21:55:46 end
```

**Additional context**

I have a proposed fix for the problem and will create a pull request.

go.mod

```
module spanner-session-leak

go 1.15

require cloud.google.com/go/spanner v1.12.0
```

8398a7:
Fixed a problem where the session was not reused when a panic occurred.

Fixes #3460


skuruppu:
Thanks for the fix @8398a7. I would like @olavloite to approve this one.

spanner/client_test.go
```diff
@@ -828,6 +828,37 @@ func TestClient_ReadWriteTransaction_Update_QueryOptions(t *testing.T) {
...
+			t.Fatalf("Unexpected error during transaction: %v", err)
+		}
+	}()
+
+	_, err := client.ReadWriteTransaction(ctx, func(ctx context.Context, tx *ReadWriteTransaction) error {
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("Unexpected error during transaction: %v", err)
+	}
```
olavloite, line 859: The current test case will hang indefinitely if a session leak would be re-introduced by accident in the future. The below suggestion would cause the test to fail fast instead.

```suggestion
	if g, w := client.idleSessions.idleList.Len(), 1; g != w {
		t.Fatalf("idle session count mismatch.\nGot: %v\nWant: %v", g, w)
	}
```

olavloite (changes requested):
Thanks, that's a good catch!

Would you mind changing the test case so that it does not hang indefinitely if the problem were to reappear? I've added a suggestion, but feel free to do it in another way if you have a better idea.

spanner/client_test.go
8398a7: It's definitely a problem when it hangs.

Fixed.
df60414

olavloite (approved):

olavloite:
@tbpg This one also seems to be stuck waiting for kokoro to report the build status. Could it be that this is related to the fact that the PR is created from a fork? Other PRs do not seem to have this problem, AFAICT.

tbpg:
Kokoro only starts automatically for a "trusted" group.

olavloite:
> Kokoro only starts automatically for a "trusted" group.

Ah. Thanks for finding that out. I guess that's reasonable.

PR #3462, opened 2020-12-14
feat(asset)!: remove v1beta1

codyoss:
Protos are no longer present and the api has been turned down.


tbpg (approved):

tbpg:
I wonder if this should be `feat!`?

PR #3465, opened 2020-12-15
feat(spanner): include User agent

mohanli-ml:
Configure the UserAgent in addition to x-goog-api-client. This will be needed for DirectPath.


spanner/doc.go
```diff
@@ -357,3 +357,7 @@ at https://godoc.org/go.opencensus.io/trace. OpenCensus tracing requires Go 1.8
 or higher.
 */
 package spanner // import "cloud.google.com/go/spanner"
+
+// clientUserAgent identifies the version of this package.
+// It should be the same as https://pkg.go.dev/cloud.google.com/go/spanner.
+const clientUserAgent = "spanner-go/v1.12.0"
```
thiagotnunes, line 363: Could this be dynamic? I am afraid we will forget to update this when a new release comes out.

thiagotnunes: For instance, see https://github.com/googleapis/google-cloud-go/commit/4b1db5fe8a5217bf17941627ec16ae8830aa794f.

It would be nice to have a review from @olavloite here too.

mohanli-ml: It seems we can not do it dynamically in Go. For example, see https://github.com/googleapis/google-cloud-go/blob/master/bigtable/doc.go#L121, and https://github.com/googleapis/google-cloud-go/blob/master/internal/version/version.go#L29. And there is an issue about using semver instead of date, see https://github.com/googleapis/google-cloud-go/issues/2749.

olavloite: We should really automate this in some way, otherwise it will certainly be forgotten for future releases. The consts that are mentioned above are also automatically generated as part of the release process. The https://github.com/googleapis/google-cloud-go/blob/master/internal/version/version.go#L29 version is generated by this file: https://github.com/googleapis/google-cloud-go/blob/master/internal/version/update_version.sh

Could we for now at least add a `TODO` here that this needs to be automated with a reference to the bash script mentioned above? This is probably something that we would want to implement for all submodules and integrate in the automated release processes.

spanner/client.go
```diff
@@ -177,6 +177,7 @@ func NewClientWithConfig(ctx context.Context, database string, config ClientConf
 			),
 		),
 		option.WithGRPCConnectionPool(config.NumChannels),
+		option.WithUserAgent(clientUserAgent),
```
olavloite, line 180: I think this will be overridden by any `user-agent` that a user has specified in the supplied options here: https://github.com/googleapis/google-cloud-go/blob/be88d0793e6c0825050037140e8c3d95e0cd0fed/spanner/client.go#L182-L184

Is that a problem?

mohanli-ml:
I have added a TODO for automating the user agent. BTW, there is a plan to use the library version instead of a date, see https://github.com/googleapis/google-cloud-go/issues/2749.

spanner/doc.go
codyoss: I agree. Bigtable is the only other client that does this today, and recent PRs have shown that the value has been forgotten about over time. If we were going to start doing something like this I think we should have a consistent header key and value that all of our clients use.

spanner/client.go
mohanli-ml: This should be good for us. Thanks for catching this!

spanner/doc.go
mohanli-ml: I have added a TODO for automating the user agent. There is a plan to use the library version instead of a date, see #2749, so I guess the best solution is to figure out a way to update the

olavloite (approved):

thiagotnunes (approved):

PR #3466, opened 2020-12-15
feat(spanner): run E2E test over DirectPath

mohanli-ml:
Run spanner-go integration tests over directpath traffic.


spanner/integration_test.go
```diff
@@ -39,14 +39,21 @@ import (
...
 	instancepb "google.golang.org/genproto/googleapis/spanner/admin/instance/v1"
 	sppb "google.golang.org/genproto/googleapis/spanner/v1"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/peer"
 	"google.golang.org/grpc/status"
 )
 
+const (
+	directPathIPV6Prefix = "[2001:4860:8040"
```
olavloite, line 53: Should we consider placing these in an environment variable instead of hard coded in the test?

olavloite (approved):
LGTM, with a non-blocking question on using env vars instead of hard coding the addresses.

spanner/integration_test.go
mohanli-ml: Previously we also do hard code, for example, see https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/test/java/com/google/cloud/spanner/GceTestEnvConfig.java#L52. So can you briefly explain the benefits of using environment variables than hard code here? Thanks!

olavloite: The benefit would depend on whether there's any chance this might change in the future, or whether there might be multiple different endpoints that we would want to test against in the future. If the chances of both those are (practically) zero, then the there's not very much benefit. The only would be that it seems slightly counter-intuitive to have an [environment variable](https://github.com/googleapis/google-cloud-go/blob/299de04ec22e3cc31d76ded257a86008960ea4bb/spanner/integration_test.go#L173) for setting the endpoint to use for the tests, and a constant IP prefix that is used to verify the host that is used.

mohanli-ml: Thanks for the comments! We specifically required that DirectPath backend IP must be start with these fixed IP prefix in the design, so the prefix will not be changed in the future (as far as I know). For the endpoint, the same endpoint can be accessed by CFE and DirectPath at the same time. Because of the DirectPath design, we can verify the traffic by checking the backend IP.

olavloite: Thanks for the explanation. That sounds reasonable to me.

PR #3471, opened 2020-12-16
fix(internal/gapicgen): don't regen files that have been deleted

codyoss:
Currently this method will find all the files that haven been
 updated since the last hash. This is then used to find out
 which packages will need to be regenered. We should not try to
 regenerate files that have been deleted, this causes a file does
 not exist error when running the generator.

 The provided flag restricts to files that:
   - (A) Added
   - (C) Copied
   - (M) Modified
   - (R) Renamed


codyoss:
@tbpg When you have a chance, could you review this PR. This is currently blocking regens.

internal/gapicgen/generator/git.go
```diff
@@ -134,7 +134,7 @@ func CommitsSinceHash(gitDir, hash string, inclusive bool) ([]string, error) {
 // hash for the given gitDir.
 func UpdateFilesSinceHash(gitDir, hash string) ([]string, error) {
 	out := bytes.NewBuffer(nil)
-	c := command("git", "diff-tree", "--no-commit-id", "--name-only", "-r", fmt.Sprintf("%s..HEAD", hash))
+	c := command("git", "diff-tree", "--no-commit-id", "--name-only", "--diff-filter=ACMR", "-r", fmt.Sprintf("%s..HEAD", hash))
```
tbpg, line 137: Mind adding a comment explaining `ACMR`? I found the PR description helpful. :smile:

tbpg (approved):

internal/gapicgen/generator/git.go
codyoss: Sure thing, inlining something useful.

PR #3473, opened 2020-12-17
bigtable: Only test data operation for DirectPath integration tests

mohanli-ml:
Because bigtable resources on TEST environment is very limited, we have to use bigtable instances on project "autonomous-mote-782". At the same time, VMs used for directpath auto tests must belong to project "directpath-pilot-testing". As a result, currently we are only able to test bigtable data client operations. This PR updates the integration test to run DirectPath tests without creating Table or ColumnFamily.


mohanli-ml:
I am trying to see if I can use prod endpoint for this tests. If so then we will not need this PR, please hold until I have more updates.

PR #3477, opened 2020-12-19
test(profiler): avoid escaping and double newlines in the rendered output

aalexand:
…tput.

Change-Id: I0689356803fd9530d34fcb693fc9c2c0fa14c641


profiler/proftest/proftest.go
```diff
@@ -472,6 +472,10 @@ func parseBackoffDuration(line string) (time.Duration, error) {
 func (tr *GCETestRunner) PollForAndReturnSerialOutput(ctx context.Context, inst *InstanceConfig, finishString, errorString string) (string, error) {
 	var output string
 	defer func() {
+		// Avoid escaping and double newlines in the rendered output (b/175999077).
+		// TODO: Use strings.ReplaceAll once no need to support Go 1.11.
```
kalyanac, line 476: once support for Go 1.11 is dropped 
or
once there is no need to support Go 1.11

kalyanac:
Failures do not look to be related to the PR.

kalyanac (approved):

profiler/proftest/proftest.go
aalexand: Fixed.

PR #3487, opened 2020-12-24
feat(spanner): Add a DirectPath fallback integration test

mohanli-ml:
Add a DirectPath fallback test for spanner in Golang. The logic of the test is:

1. First wait until DirectPath traffic is observed;
2. Blackhole DirectPath net, and wait until the client fallback to use CFE;
3. Unblackhole DirectPath net, and the client should upgrade back to use DirectPath again.

The DirectPath and CFE traffic are distinguished by peer IP.


spanner/integration_test.go
```diff
@@ -2957,6 +2968,52 @@ func TestIntegration_StartBackupOperation(t *testing.T) {
...
+	}
+	if len(allowDpv6Cmd) == 0 {
+		t.Fatal("-it.allowdpv6-cmd unset")
+	}
+	if len(allowDpv4Cmd) == 0 {
+		t.Fatal("-it.allowdpv4-cmd unset")
+	}
+
+	// Precondition: wait for DirectPath to connect.
+	countEnough := examineTraffic(ctx, client /*blackholeDP = */, false)
```
thiagotnunes, line 2997: nit: `/*blackholeDP = */` should be added after the `,`

thiagotnunes (approved):

spanner/integration_test.go
mohanli-ml: The problem is that gofmt will automatically move `/*blackholeDP = */` before the `,`, do you want me to remove the comment?

thiagotnunes: No, nevermind, maybe that is the standard.

josephalvear:
Not interested.   
  
  
  
  
  
    
  

  
  
>

PR #3489, opened 2020-12-24
feat(bigtable): allow restore backup to different instance

AlisskaPie:


kolea2 (approved):

bigtable/admin.go
```diff
@@ -1441,10 +1445,17 @@ func UpdateInstanceAndSyncClusters(ctx context.Context, iac *InstanceAdminClient
...
-	prefix := ac.instancePrefix()
-	backupPath := prefix + "/clusters/" + cluster + "/backups/" + backup
+	return ac.RestoreTableTo(ctx, ac.instance, table, cluster, backup)
+}
 
+// RestoreTableTo creates a new table by restoring from this completed backup.
+//
+// diffInstance is an instance in which the new table will be restored to.
+// Instance must be in the same project as the project containing backup.
+func (ac *AdminClient) RestoreTableTo(ctx context.Context, diffInstance, table, cluster, backup string) error {
```
tritone, line 1455: instead of diffInstance I would do something like dstInstance. It also might be worth indicating that the other parameters are for the source. @kolea2 you can advise on the best names here.

bigtable/admin.go
```diff
@@ -1441,10 +1445,17 @@ func UpdateInstanceAndSyncClusters(ctx context.Context, iac *InstanceAdminClient
 
 // RestoreTable creates a table from a backup. The table will be created in the same cluster as the backup.
 func (ac *AdminClient) RestoreTable(ctx context.Context, table, cluster, backup string) error {
-	ctx = mergeOutgoingMetadata(ctx, ac.md)
-	prefix := ac.instancePrefix()
-	backupPath := prefix + "/clusters/" + cluster + "/backups/" + backup
+	return ac.RestoreTableTo(ctx, ac.instance, table, cluster, backup)
+}
 
+// RestoreTableTo creates a new table by restoring from this completed backup.
```
tritone, line 1451: I would try to make the doc here as parallel as possible to RestoreTable above, but clarify what is different. As it is I'm a bit confused about "completed" and whether that is also required for RestoreTable (but I'm not that familiar with how bigtable backups work.

bigtable/integration_test.go
```diff
@@ -2033,6 +2033,26 @@ func TestIntegration_AdminBackup(t *testing.T) {
 	table := testEnv.Config().Table
 	cluster := testEnv.Config().Cluster
```
tritone, line 2035: Are there unit tests that cover RestoreTable currently or only integration tests? I don't see any unit tests that call RestoreTable directly but just want to make sure I'm not missing something. @kolea2

tritone (commented):
Looks fine to me aside from naming and docs issues. @AlisskaPie if we don't hear from you we'll just merge this into the branch as-is and make a separate PR to fix these issues before merging to master.

kolea2:
Merging as WIP to `backup-different-instance` branch

PR #3491, opened 2020-12-25
feat(spanner/spannertest): support Not Null constraint

### LINKED ISSUE #3490: spannertest: Support Not Null constraint
`Not Null` constraint seems not to be supported yet and writing null value to a not-null column is accepted.

I think it can make the library users miss bugs that could be caught in unit tests.
Actually, I missed a bug that my application wrote null to a not-null column in the unit test and finally found this in the integration test with the real spanner.

Here is the error I actually confirmed (the table and column names were replaced)
```
Code: FailedPrecondition
Error message: "{Column Name} must not be NULL in table {Table Name}"
```
1. add `NotNull` Field to `colInfo`
2. add a validation that checks if null is about to be written to a Not Null column.

sryoya:
# About
Fixes #3490 

# Note
- This PR doesn't include a test for the newly-added logics because it seems there are no test for unhappy paths in the related functions. Let me know if I need to cover them by test.
- The "alterColumn" function also should have some checks using `Not Null`, but, I've not done that in this PR. Assuming how the library users use this library, I suppose the check in writing a record is more prioritized.


spanner/spannertest/db.go
```diff
@@ -68,6 +68,7 @@ type table struct {
 type colInfo struct {
 	Name     spansql.ID
 	Type     spansql.Type
+	NotNull  bool
```
dsymonds, line 71: given this structure is generally populated in a few different places for different purposes, and `NotNull` is only going to be set for table columns, please add a line column to that effect.

```go
  NotNull bool  // only set for table columns
```

spanner/spannertest/db.go
```diff
@@ -632,8 +635,9 @@ func (t *table) addColumn(cd spansql.ColumnDef, newTable bool) *status.Status {
 	}
 
 	t.cols = append(t.cols, colInfo{
-		Name: cd.Name,
-		Type: cd.Type,
+		Name:    cd.Name,
+		Type:    cd.Type,
+		NotNull: cd.NotNull,
```
dsymonds, line 640: down on line 718 there's a TODO that could be at least partially done, namely setting this value in an `ALTER TABLE`. No need to do any validation right now (leave that as a TODO), but may as well maintain the colInfo.NotNull field.

spanner/spannertest/db.go
dsymonds: also, update the code near line 222 to return this.

dsymonds (commented):
Looks good for the most part. A couple of extra bits that should be done in this same change.

No need to worry about a test; I assume you've exercised this with your own code already?

spanner/spannertest/db.go
sryoya: I see. And, thanks for letting me know the missing place to update.

https://github.com/googleapis/google-cloud-go/pull/3491/commits/931c0baa70312dd195320deb77114ae3ce7532db

spanner/spannertest/db.go
sryoya: Certainly, that was worth doing now.

Done, apart from the validation based on the requirement to be not null.
https://github.com/googleapis/google-cloud-go/pull/3491/commits/840de7cdb3e5954dd3d1d4f3845014281de1e9f0

Btw, I updated the below related codes in this func as well because the previous code didn't seem to accept a simple statement that just adds/removes Not Null without doing any alternation.

spanner/spannertest/db.go
```diff
@@ -705,38 +710,42 @@ func (t *table) alterColumn(alt spansql.AlterColumn) *status.Status {
...
-		} else if oldT.Base == spansql.String && newT.Base == spansql.Bytes {
-			conv = func(x interface{}) interface{} { return []byte(x.(string)) }
-		}
-		if conv != nil {
-			for _, row := range t.rows {
-				if row[ci] != nil { // NULL stays as NULL.
-					row[ci] = conv(row[ci])
+	// TODO: We don't track whether commit timestamps are permitted on a per-column basis, so that's ignored.
+
+	if oldT != newT {
```
sryoya, line 722: It should be fine now as these are not pointers of a struct and have just primitive types. But, if there is a possibility that another field with non-primitive types would be added, we may want to do another comparing way like making a custom comparing func or using `reflect.DeepEqual`.

sryoya:
Thank you for the review!

>  I assume you've exercised this with your own code already?

Sure, I confirmed that inserting null into Not Null column was declined.

spanner/spannertest/db.go
dsymonds: yep, direct comparison of spansql.Type values is fine; it is done in dozens of places throughout this code.

dsymonds: however, please invert this condition and early-out, and leave the remainder outdented like it was before.

sryoya: Ok, reverted the changes.
https://github.com/googleapis/google-cloud-go/pull/3491/commits/4df4f5db3958312d002897301242112326bb8239

I assume you intend to update here in another chance?

dsymonds: Not sure what you mean by your last question, but all I was asking for was that when one writes something like

```go
if cond {
  ... // long section
  return x
}
return y
```

then it is often better to invert the condition to reduce the indentation of that long section:

```go
if !cond {
  return y
}
... // long section
return x
```

It makes it easier to follow what's going on.

Anyway, the current state of this PR looks fine now.

dsymonds (approved):

spanner/spannertest/db.go
sryoya: Ok, I got what you mean. Thank you for your explanation and I agree with the basic principle you said.

Certainly, I could make the code like this.

```.go
if needsTypeRelatedTransformations { // means type transformation or length changes is demanded in the Alter Column statement
  if !stringOrBytes(oldT.Base) || !stringOrBytes(newT.Base) || oldT.Array && newT.Array{
   return y // error or kind of error
 }
  ... // long section
  return x // or go to another alternation process without noting
}
... // another alternation logics
return z // probably nil in this function.
```

And, I just wanted to confirm you are going to add the first IF statement above (and invert the condition as you say) next time because I just worried about a simple Statement adding/removing Not Null can't be accepted in the current code as long as it also has a type transformation.

But, as you say, it should be out of the scope of this PR and can fix anytime.

Thank you for your careful reviews 😄

dsymonds: Ah, well if there's multiple transformations that need doing, the code needs more refactoring. Particularly if it needs to be able to fail (e.g. adding `NOT NULL` to a column with existing `NULL`s) then it likely can't handle one transformation completely at a time anyway, and would need to figure out the full change it'll make, check the data, then make the change all at once.

sryoya: Ok, I'm glad to know what you mean and let you know what I meant.

As you say, this function seems to need more refactoring. 
And, I guess there are several ways.
As you say, the code can figure out, check, and execute all demanding alternations at once.
Or, we can also do them per by a kind of alternation like Not Null alternation, type transformation and name change.

The previous change I tried to make is following the latter way, but I think either way fine.

PR #3493, opened 2020-12-30
feat(bigtable): Incrementable values

AlisskaPie:


billyjacobson:
Looks good from a Bigtable perspective

billyjacobson:
@tbpg This looks good on my end, could you give it a language review? Thanks

bigtable/bigtable.go
```diff
@@ -619,6 +621,14 @@ func (m *Mutation) Set(family, column string, ts Timestamp, value []byte) {
 	}}})
 }
 
+// SetInt64 sets a value in a specified column, with the given timestamp.
+// It performs the 64-bit big-endian signed integer encoding.
+func (m *Mutation) SetInt64(family, column string, ts Timestamp, value int) {
```
tbpg, line 626: Seems like this should be an `int64` argument if we always treat it as an `int64`, especially given the name of the method?

bigtable/bigtable.go
```diff
@@ -619,6 +621,14 @@ func (m *Mutation) Set(family, column string, ts Timestamp, value []byte) {
...
+	buf := new(bytes.Buffer)
```
tbpg, line 627:
```suggestion
	var buf bytes.Buffer
```

bigtable/bigtable.go
```diff
@@ -619,6 +621,14 @@ func (m *Mutation) Set(family, column string, ts Timestamp, value []byte) {
...
+	binary.Write(buf, binary.BigEndian, int64(value))
```
tbpg, line 628: `Write` returns an error -- we should check it.

bigtable/filter.go
```diff
@@ -125,6 +127,15 @@ func (cf columnFilter) proto() *btpb.RowFilter {
 // See https://github.com/google/re2/wiki/Syntax for the accepted syntax.
 func ValueFilter(pattern string) Filter { return valueFilter(pattern) }
 
+// ValueFilterInt64 returns a filter that matches cells whose value
+// matches the provided RE2 pattern.
+// It performs the 64-bit big-endian signed integer encoding.
+func ValueFilterInt64(pattern int) Filter {
```
tbpg, line 133: Same `int64` argument and `binary.Write` comments here and for `ValueRangeFilterInt64`.

bigtable/bigtable.go
tbpg, line 626: Looks like we need tests for this.

tbpg (changes requested):

PR #3512, opened 2021-01-08
fix(pubsublite): close clients after publisher and subscriber have terminated

tmdiep:
The gapic clients aren't closed once the `PublisherClient` and `SubscriberClient` have terminated. This leaves client connections open. We could add an explicit Close() (like the `AdminClient`) and require users to call it, but we know when the publisher and subscriber have permanently terminated (i.e. when `PublisherClient.Stop` is called, and when `SubscriberClient.Receive` returns), so we'll can just close all the gapic clients internally.

This caused all unit tests to fail because a single connection to the mock lite server was being reused. The first test would close it, causing all subsequent tests to fail. So the unit test clients were refactored to create new connections for every test.


pubsublite/internal/wire/subscriber.go
```diff
@@ -379,13 +379,17 @@ func (f *singlePartitionSubscriberFactory) New(partition int) *singlePartitionSu
 // multiPartitionSubscriber receives messages from a fixed set of topic
 // partitions.
 type multiPartitionSubscriber struct {
+	// Immutable after creation.
+	clients     apiClients
```
hongalex, line 383: I don't see where `clients.Close()` is called. Is this intentional?

tmdiep: Ah my bad, I somehow failed to patch the clients.Close() call for the subscriber clients. Fixed.
Thanks for noticing!

pubsublite/admin_test.go
```diff
@@ -362,6 +366,7 @@ func TestAdminListSubscriptions(t *testing.T) {
 	defer mockServer.OnTestEnd()
 
 	admin := newTestAdminClient(t)
+	//defer admin.Close()
```
hongalex, line 369: Perhaps this should be uncommented?

hongalex (approved):
LGTM pending one comment

pubsublite/admin_test.go
tmdiep: Done. Thanks!

PR #3516, opened 2021-01-08
feat(spanner): attempt DirectPath by default

mohanli-ml:
Update client to attempt DirectPath by default.

Note that it doesn't mean that after this change client will just use DirectPath, but will call the DirectPath codepath by default.

The actually enablement of DirectPath is controlled by service owner via ACL config. For now, after this change, although all users will attempt DirectPath, but they will all just fallback to the original CFE path.


tbpg (commented):
Would this have latency implications for users who don't have access to DirecPath?

mohanli-ml:
Not sure I 100% understand the question: 
1. If the user's client library version is old and client does not attempt DirectPath, then the user will directly use CFE, which is what it currently is.
2. If the user's client attempt DirectPath (after this PR gets merged), the request will be checked by ACLs set at grpclb. For now there is no Allow ALCs, so traffic will fallback to CFE, and the latency will be sending a RPC request to the grpclb, and waiting for the response. The latency can be affected by many factors, such as if the client and grpclb are in the same cell, but normally it should be at the level of sub milliseconds. In the worst case, however, the client may wait up to 10 seconds for the response.

tbpg:
I guess my question is if this will disrupt users without DirectPath access. Would their requests make an extra round trip? Is that something we should call out prominently in the changelog? I bring it up because users tend to care about network changes (we get issues pretty frequently about various network things).

I don't know -- this may have already been talked about. Up to @skuruppu.

mohanli-ml:
Users without DirectPath access should not be disrupted. The traffic will fallback to CFE as what is currently is, and users are not supposed to know this.

tbpg (approved):
Got it. Thanks! Please wait for @skuruppu.

skuruppu (approved):

skuruppu:
FYI @thiagotnunes

PR #3521, opened 2021-01-11
fix(datastore): Ensure the datastore time is returned as UTC

crwilcox:


datastore/load.go
```diff
@@ -536,7 +536,7 @@ func propToValue(v *pb.Value) (interface{}, error) {
 	case *pb.Value_DoubleValue:
 		return v.DoubleValue, nil
 	case *pb.Value_TimestampValue:
-		return time.Unix(v.TimestampValue.Seconds, int64(v.TimestampValue.Nanos)), nil
+		return time.Unix(v.TimestampValue.Seconds, int64(v.TimestampValue.Nanos)).In(time.UTC) nil
```
tbpg, line 539: Missing a comma here. :smile:

datastore/load.go
```diff
@@ -369,7 +369,7 @@ func setVal(v reflect.Value, p Property) (s string) {
 			if ok {
 				s := micros / 1e6
 				ns := micros % 1e6
-				v.Set(reflect.ValueOf(time.Unix(s, ns)))
+				v.Set(reflect.ValueOf(time.Unix(s, ns).In(time.UTC)))
```
tbpg, line 372: Mind adding a test that ensures the times are returned in UTC?

tbpg (changes requested):

datastore/load.go
tritone: +1, might be easier said than done though :)

datastore/load.go
crwilcox: Thanks @tbpg If not obvious, I need to author tests/review existing for this yet. I am trying to wrap another task, but also was hoping to have CI let me know how much is left here :) Should get back to this later today.

datastore/load.go
crwilcox: test added.

datastore/load_test.go
```diff
@@ -486,6 +486,21 @@ func TestLoadToInterface(t *testing.T) {
 			dst:  &withUntypedInterface{Field: 1e9},
 			want: &withUntypedInterface{Field: "Newly set"},
 		},
+		{
+			name: "struct with timestamp",
```
tritone, line 490: Just to confirm, does this test case fail currently without the fix?

tritone (commented):
Looks good.

datastore/load_test.go
```diff
@@ -584,6 +569,36 @@ func TestLoadToInterface(t *testing.T) {
 	}
 }
 
+func TestTimezone(t *testing.T) {
+	// Expect Local times to be represented in UTC
```
tritone, line 573: I'd put this comment above L572 since it describes the whole test.

datastore/load_test.go
```diff
@@ -584,6 +569,36 @@ func TestLoadToInterface(t *testing.T) {
...
+		Time: time.Unix(1605504600, 0).In(time.UTC),
+	}
+
+	err := loadEntityProto(dst, src)
+
+	if err != nil {
+		t.Fatalf("loadEntityProto: %v", err)
+	}
+
+	gotZone, _ := dst.Time.Zone()
```
tritone, line 594: Maybe actually compare dst to want as well, and note why you need to compare the time zones separately.

tbpg (approved):
Thanks!

Issue #3519, opened 2021-01-09
bigquery: handleInsertErrors: equality check should be >=, not >


muscovite:
**Client**

BigQuery Go API

**Environment**

All

**Go Environment**

All

**Code**

https://github.com/googleapis/google-cloud-go/blob/bigquery/v1.14.0/bigquery/inserter.go#L226

The `>` check

```go
 if int(e.Index) > len(rows)
```

should be `>=`

```go
 if int(e.Index) >= len(rows)
```

Otherwise, the next lines can attempt to access an out of range index:
```go
rie := RowInsertionError{
	InsertID: rows[e.Index].InsertId,
	RowIndex: int(e.Index),
}
```

**Expected behavior**

No out of bounds errors

**Actual behavior**

```
panic: runtime error: index out of range [1] with length 1

goroutine 1432095 [running]:
extrahop/vendor/cloud.google.com/go/bigquery.handleInsertErrors(0xc0002c8500, 0x3e8, 0x42a, 0xc000012708, 0x1, 0x1, 0x0, 0xc00f377ce8)
        /depot/extrahop/go/src/extrahop/vendor/cloud.google.com/go/bigquery/inserter.go:236 +0x56a
extrahop/vendor/cloud.google.com/go/bigquery.(*Inserter).putMulti(0xc00029a7a0, 0xdbd580, 0xc04648e8d0, 0xc04649c200, 0x1, 0x1, 0x0, 0xdbd580)
        /depot/extrahop/go/src/extrahop/vendor/cloud.google.com/go/bigquery/inserter.go:191 +0x3b1
extrahop/vendor/cloud.google.com/go/bigquery.(*Inserter).Put(0xc00029a7a0, 0xdbd580, 0xc04648e870, 0xb6edc0, 0xc00023c980, 0x0, 0x0)
... // the rest of the stack trace is no longer in library code
```

shollyman:
Thanks for the report.  Just to verify, were you seeing this behavior against the live service, or was this purely a byproduct of testing?

muscovite:
Thanks for the fix! This was seen against the live service. For full context, though, the API caller does something fairly hacky where the row data is passed through http.RoundTripper as gzip-compressed bytes, and the `src` arg to `Put` is a single dummy row. So I think it's unlikely that "normal" API usage would have run into this issue.

shollyman:
[...]

PR #3524, opened 2021-01-11
fix(bigquery): address possible panic due to offset checking in handleInsertErrors 

shollyman:
External reporter identified case where improper bounds checking can
trigger panic when comparing the index value from structured error
response.

Fixes: #3519


bigquery/inserter_test.go
```diff
@@ -138,10 +144,29 @@ func TestHandleInsertErrors(t *testing.T) {
...
+				{Errors: []*bq.ErrorProto{{Message: "m0"}}, Index: 2},
+			},
+			want: fmt.Errorf("internal error: unexpected row index: 2"),
+		},
 	} {
 		got := handleInsertErrors(test.in, rows)
-		if !testutil.Equal(got, test.want) {
-			t.Errorf("%#v:\ngot\n%#v\nwant\n%#v", test.in, got, test.want)
+		_, ok := got.(PutMultiError)
+		if ok {
```
tbpg, line 157:
```suggestion
		if _, ok := got.(PutMultiError); ok {
```

bigquery/inserter_test.go
```diff
@@ -138,10 +144,29 @@ func TestHandleInsertErrors(t *testing.T) {
...
+			// compare structure of the PutMultiError
+			if !testutil.Equal(got, test.want) {
+				t.Errorf("(case: %s)\nin %#v\ngot\n%#v\nwant\n%#v", test.description, test.in, got, test.want)
+			}
+		} else {
```
tbpg, line 162:
```suggestion
			}
			continue
		}
```

shollyman: Unsure I'm following.  We still want to probe errors when not a PutMultiError, not simply move to the next case.

shollyman: Revisited.  Continue inside the type assertion case, and then do the secondary check.

codyoss (approved):
LGTM

PR #3528, opened 2021-01-12
fix(pubsublite)!: add separate publisher and subscriber client constructors with settings (api review)

tmdiep:
We expect most users to just use default publish/receive settings. The NewPublisherClient and NewSubscriberClient constructors will omit the settings arg (using default settings). The additional *WithSettings constructors will allow users to specify settings. This is consistent with NewClient/NewClientWithConfig in the bigtable and spanner libraries.

AI from API review.


tmdiep:
Converted this PR to draft as I'm considering adding NewPublisherClient() and NewPublisherClientWithSettings() for consistency with bigtable and spanner. Pending objections via email.

tmdiep:
WANT_LGTM=all

After seeing [NewClientWithConfig](https://pkg.go.dev/cloud.google.com/go/spanner#NewClientWithConfig) in the spanner and bigtable libraries, I thought it would be better to be consistent. And possibly less confusing without the nil arg.

I considered renaming *Settings to *Config, but these are meant to mirror pubsub.PublishSettings and pubsub.ReceiveSettings.

pubsublite/ps/publisher.go
```diff
@@ -70,11 +70,20 @@ type PublisherClient struct {
...
+func NewPublisherClient(ctx context.Context, topic pubsublite.TopicPath, opts ...option.ClientOption) (*PublisherClient, error) {
+	return NewPublisherClientWithSettings(ctx, topic, DefaultPublishSettings, opts...)
+}
+
+// NewPublisherClientWithSettings creates a new Cloud Pub/Sub Lite client to
+// publish messages to a given topic, using the specified PublishSettings.
+//
+// See https://cloud.google.com/pubsub/lite/docs/publishing for more information
+// about publishing.
+func NewPublisherClientWithSettings(ctx context.Context, topic pubsublite.TopicPath, settings PublishSettings, opts ...option.ClientOption) (*PublisherClient, error) {
```
tbpg, line 86: Which settings are needed at client creation time? Looking at `pubsub.PublishSettings` and `ps.PublishSettings`, it appears to be `Timeout`?

Could timeout be specified as a `ClientOption` instead? That's where we have other connection configuration options.

manuelmenzella-google: I'm not sure I fully understand this proposal -- where would the remaining settings be passed? We want these to be immutable and chosen when the client is created (e.g. message batching settings). I don't mind having custom New*WithOptions() methods but we should have a way to set client-wide settings when building the client.

tmdiep: Yes, Timeout is required during client initialization. Maybe others in the near future, e.g. I'm considering adding a setting for users to enable logging some debug info.

Also as Manuel mentioned, it's to make it clear that the settings are immutable and must be specified before initialization. I prefer this over `pubsub`, which puts the settings on the client (`Topic`) struct, documents that they have to be set before the first Publish call, and defers initialization until the first call.

tbpg: Gotcha. In that case, it makes sense to me to have the separate function that accepts settings.

What about gRPC connection settings? Should those follow the pattern of other gRPC settings and be set via an option? Or is that something unique to `pubsublite`?

tmdiep: Yes, the low-level gRPC connection settings will be set via common options. The Timeout field in the pubsublite settings structs bound reconnection that the library automatically performs when the streams break due to retryable errors. I found that passing a ctx with timeout to NewClient tears down the entire connection permanently after the timeout expires, which is not the behavior that we want.

codyoss: I think what Tyler may have been talking about here is making use of [option.WithGRPCDialOption](https://pkg.go.dev/google.golang.org/api/option#WithGRPCDialOption). Here you can pass things like [grpc.WithTimeout](https://pkg.go.dev/google.golang.org/grpc#WithTimeout).

tmdiep: Ah got it, thanks!

The docs for [grpc.WithTimeout](https://pkg.go.dev/google.golang.org/grpc#WithTimeout) say it's deprecated and to use [grpc.DialContext](https://pkg.go.dev/google.golang.org/grpc#DialContext) + context.WithTimeout instead. I noticed that it takes an endpoint addr, which we don't want users to set because it needs to be transformed to a regional endpoint, done internally [here](https://github.com/googleapis/google-cloud-go/blob/4f7c6d8299fdcd4fa1ad94490c5b70057902afb6/pubsublite/internal/wire/rpc.go#L137). So I think it would be better to create grpc.DialContext internally (will fix in a separate PR).

Additionally, the `Timeout` fields also apply after we already have a ClientConn, when creating new gRPC streams, which involves a handshaking protocol (initialization request + response) and retry logic that is unique to Pub/Sub Lite ([code](https://github.com/googleapis/google-cloud-go/blob/bbc61ed368453b28aaf5bed627ca2499a3591f63/pubsublite/internal/wire/streams.go#L257)). The server can deliberately break streams with retryable error codes, which prompts the client to recreate the stream (all handled internally by the pubsublite library).

TL;DR - I think we'll need `Timeout` as a fields in the settings structs, as it affects high-level logic that is unique to pubsublite.

Hopefully I'm making sense. LMK if I'm missing something in regards to the DialContext.

tbpg (approved):

pubsublite/ps/publisher.go
tbpg: Makes sense to me if the configuration is very pubsublite-specific that we put it into the settings struct. Thank you for the explanation.

I'll let Cody comment on if there is another option/setting that could accomplish the same thing generally. But, I can't think of anything.

codyoss: I think you made a good argument for the setting. I can't think of an alternative either.

codyoss (approved):

pubsublite/ps/publisher.go
tmdiep: Thanks everyone for the review and pointers.

PR #3529, opened 2021-01-12
refactor(pubsublite): use iota for DeliveryRequirement enum values

tmdiep:
No-op change. AI from API review.


codyoss (approved):
LGTM, may be good to add a static translation map so we are not dependent on adding in the same order.

tmdiep:
Thanks Cody. We don't expect this particular enum to change, so I might just keep it simple.

Issue #3533, opened 2021-01-12
bigquery: add Determinism to RoutineMetadata


lightyear15:
**Client**
BigQuery
package cloud.google.com/go/bigquery version v1.14.0 

**Expected behavior**
struct ``RoutineMetadata`` containing a field to express function determinism, like in package 
google.golang.org/api/bigquery/v2
where struct ``Routine`` has field ``DeterminismLevel``

**Actual behavior**
No way to express level of determinism

shollyman:
Will add this, but I don't understand the comment around message arrival.  Could you elaborate?

lightyear15:
> Will add this, but I don't understand the comment around message arrival. Could you elaborate?

No sorry,
that was a left over from the issue template :sweat_smile:

PR #3534, opened 2021-01-12
feat(bigquery): add support for allowing Javascript UDFs to indicate determinism

shollyman:
BigQuery supports javascript UDFs, which allow users to define function
bodies to be executed via javascrtipt.  Allowing users to indicate
determinism level of the UDF helps the query engine make more efficient
choices around aspects like query caching.

Fixes: #3533


codyoss (approved):
LGTM

bigquery/routine.go
```diff
@@ -299,6 +315,21 @@ func (rm *RoutineMetadataToUpdate) toBQ() (*bq.Routine, error) {
...
+		if x, ok := rm.DeterminismLevel.(RoutineDeterminism); ok {
+			r.DeterminismLevel = string(x)
+			processed = true
+		}
+		if x, ok := rm.DeterminismLevel.(string); ok {
+			r.DeterminismLevel = x
+			processed = true
+		}
+		if !processed {
+			panic(fmt.Sprintf("DeterminismLevel should be either type string or RoutineDetermism in update, got %T", rm.DeterminismLevel))
```
tswast, line 330: Why panic instead of returning `error`?

shollyman: This mimics the behavior of the converters in https://pkg.go.dev/cloud.google.com/go/internal/optional.  All the other fields panic if you use incorrect type, so yielding a custom error for the one field seems a mismatch.

tswast (approved):

PR #3540, opened 2021-01-13
docs(pubsublite): more descriptive comments for the publish package

tmdiep:
Added comments and example usage.


pubsublite/publish/metadata.go
```diff
@@ -19,28 +20,43 @@ import (
...
+//   result := publisher.Publish(ctx, &pubsub.Message{Data: []byte("payload")})
+//   id, err := result.Get(ctx)
+//   if err != nil {
+//     // TODO: Handle error.
+//   }
+//   metadata, err := publish.ParseMetadata(id)
+//   if err != nil {
+//     // TODO: Handle error.
+//   }
+func ParseMetadata(serverID string) (*Metadata, error) {
```
manuelmenzella-google, line 50: Nit; is this really a serverID? Maybe messageID, but that's possibly overloaded. messageDestination, even? I am not sure, I don't have a great alternative.

manuelmenzella-google (approved):

pubsublite/publish/metadata.go
tmdiep: As discussed, renamed to just "id", for lack of better names.

PR #3549, opened 2021-01-13
feat(storage): add projection parameter for BucketHandle.Objects()

### LINKED ISSUE #3445: storage: add Projection to Query
GCS allows users to pass in a projection when listing objects: https://cloud.google.com/storage/docs/json_api/v1/objects/list#parameters

Currently we have this hardcoded as "full" which is not ideal for some use cases. https://github.com/googleapis/google-cloud-go/blob/2fa1b727f8a7b20aa62fe0990530744f6c109be0/storage/bucket.go#L1207

We should allow users to pass in the desired Projection via a [Query](https://github.com/googleapis/google-cloud-go/blob/2fa1b727f8a7b20aa62fe0990530744f6c109be0/storage/storage.go#L1310)

ava12:
Closes #3445.


storage/storage.go
```diff
@@ -1411,6 +1423,16 @@ func (q *Query) SetAttrSelection(attrs []string) error {
 	return nil
 }
 
+// GetProjection returns either value of Projection field or
+// default projection when this field is empty.
+func (q *Query) GetProjection() string {
```
tritone, line 1428: No need for a getter function here; you can just put this logic in ObjectIterator.fetch directly.

storage/storage.go
```diff
@@ -1306,6 +1306,14 @@ func encodeUint32(u uint32) string {
 	return base64.StdEncoding.EncodeToString(b)
 }
 
+const (
+	// ProjectionFull returns all fields of object(s).
+	ProjectionFull = "full"
+
+	// ProjectionNoAcl returns all fields of object(s) except for owner and acl.
```
tritone, line 1313: Capitalize Owner and ACL.

storage/storage.go
```diff
@@ -1341,6 +1349,10 @@ type Query struct {
 	// lexicographically before endOffset. If startOffset is also set, the objects
 	// listed will have names between startOffset (inclusive) and endOffset (exclusive).
 	EndOffset string
+
+	// Projection defines set of properties to return. This may be either
```
tritone, line 1353: I'd say this:
```
// Projection defines the set of properties to return. It will default to ProjectionFull,
// which returns all properties. Passing ProjectionNoACL will omit Owner and ACL,
// which may improve performance when listing many objects.
```

storage/bucket_test.go
```diff
@@ -663,3 +663,31 @@ func TestNewBucket(t *testing.T) {
 		t.Errorf("got=-, want=+:\n%s", diff)
 	}
 }
+
+func TestObjectsWithProjection(t *testing.T) {
```
tritone, line 667: Instead of this, can we do an integration test like what we have for SetAttrSelection? https://github.com/googleapis/google-cloud-go/blob/master/storage/integration_test.go#L1174

tritone (commented):
Thanks! A few comments...

tritone (commented):
Looks good! Running gofmt should resolve the presubmit failure.

storage/storage.go
```diff
@@ -1341,6 +1349,11 @@ type Query struct {
 	// lexicographically before endOffset. If startOffset is also set, the objects
 	// listed will have names between startOffset (inclusive) and endOffset (exclusive).
 	EndOffset string
+
+	// Projection defines the set of properties to return. It will default to ProjectionFull,
+	// which returns all properties. Passing ProjectionNoACL will omit Owner and ACL,
+	// which may improve performance when listing many objects.
+	Projection string
```
codyoss, line 1356: Could this ever have more then two possible values @tritone ? Maybe this should be a named iota?

tritone: Discussed with Cody, this seems like a good idea. Let's do the following:

```
type Projection int

const (
  ProjectionDefault  Projection = iota
  ProjectionFull
  ProjectionNoACL
)

func (p Projection) string() {
  // return "", "full", "noAcl" respectively
}
```

ava12: Why do we need a separate `ProjectionDefault`? We can simply use `ProjectionFull` as default.

tritone: @codyoss suggested this to follow https://google.aip.dev/126 . I do think it seems good to disambiguate between the user not specifying a projection and the user specifically asking for "full" in case API values or defaults change in the future.

ava12: I see.
Do we really need to return an empty string for default projection? I guess we need a string only when we pass projection to generated `ObjectsListCall`, so there should be only "full" and "noAcl".

tritone: Yes, I think it's best to have it for completeness. @codyoss feel free to chime in if you have thoughts.

tritone (approved):
LGTM. @codyoss @frankyn feel free to chime in if you have thoughts.

storage/integration_test.go
```diff
@@ -1236,6 +1237,42 @@ func testObjectsIterateAllSelectedAttrs(t *testing.T, bkt *BucketHandle, objects
...
+		if err == iterator.Done {
+			t.Fatalf("no objects")
+		}
+		if err != nil {
+			t.Fatalf("iterator.Next: %v", err)
+		}
+
+		if expectACL {
+			if attrs.Owner == "" {
+				t.Errorf("projection: %q, empty Owner", projection)
```
codyoss, line 1260: nit: errors should be in the format of "got something, want something else": https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

frankyn (approved):

PR #3558, opened 2021-01-14
feat(bigtable): attempt DirectPath by default

mohanli-ml:
Update client to attempt DirectPath by default.

Note that it doesn't mean that after this change client will just use DirectPath, but will call the DirectPath codepath by default.

The actually enablement of DirectPath is controlled by service owner via ACL config. For now, after this change, although all users will attempt DirectPath, but they will all just fallback to the original CFE path.


tritone (commented):
Seems fine to me, @kolea2 or @igorbernstein2 can you confirm?

kolea2 (approved):

igorbernstein2 (approved):
lgtm

tritone (approved):

PR #3559, opened 2021-01-14
fix(pubsublite)!: use raw strings for resource paths (api review)

tmdiep:
* Moved resource path structs from `types.go` to `internal/wire/resources.go`. Not deleted entirely, as we need to parse the components of the path for some operations.
* The public API now only accepts raw strings containing full resource paths. Comments for every resource path string specify the required format.
* The resource path strings are validated before sending requests to the server.
* Updated docs and examples. Corrected references to the service to "Pub/Sub Lite" (removed Cloud prefix).

AI from API review.


pubsublite/admin.go
```diff
@@ -144,28 +183,41 @@ func (ac *AdminClient) UpdateSubscription(ctx context.Context, config Subscripti
...
-	return protoToSubscriptionConfig(subspb)
+	return protoToSubscriptionConfig(subspb), nil
 }
 
 // Subscriptions retrieves the list of subscription configs for a given project
-// and zone.
-func (ac *AdminClient) Subscriptions(ctx context.Context, location LocationPath) *SubscriptionIterator {
+// and zone. A valid parent path has the format:
+// "projects/PROJECT_ID/locations/ZONE".
+func (ac *AdminClient) Subscriptions(ctx context.Context, parent string) *SubscriptionIterator {
```
manuelmenzella-google, line 215: Is there a reason we can't return (*SubscriptionIterator, err) here instead? That seems better than embedding the error in the iterator itself.

pubsublite/internal/wire/resources.go
```diff
@@ -40,6 +41,109 @@ func ValidateRegion(input string) error {
...
+
+func (l LocationPath) String() string {
+	return fmt.Sprintf("projects/%s/locations/%s", l.Project, l.Zone)
+}
+
+var locPathRE = regexp.MustCompile(`^projects/([^/]+)/locations/([^/]+)$`)
+
+// ParseLocationPath parses a project/location path.
+func ParseLocationPath(input string) (LocationPath, error) {
+	parts := locPathRE.FindStringSubmatch(input)
```
manuelmenzella-google, line 70: I think we need to check for "nil" here. Does len(nil) return zero or something like that?

manuelmenzella-google (approved):

pubsublite/internal/wire/resources.go
tmdiep: Yes, len(nil) returns zero.

pubsublite/admin.go
tmdiep: It seems to be convention for the list methods to return just the iterator and then fail Next() if the RPC fails. So I just tried to be consistent with the other Go libs and gapic generated libs.

PR #3567, opened 2021-01-15
feat(networkconnectivity): initial generation of apiv1alpha1

bcoe:
Initial generation of network connectivity.


internal/gapicgen/generator/config.go
```diff
@@ -770,6 +770,14 @@ var microgenGapicConfigs = []*microgenConfig{
 		apiServiceConfigPath:  "google/cloud/functions/v1/functions_gapic.yaml",
 		releaseLevel:          "ga",
 	},
+	{
+		inputDirectoryPath:    "google/cloud/networkconnectivity/v1alpha1",
+		pkg:                   "networkconnectivity",
+		importPath:            "cloud.google.com/go/networkconnectivity/v1alpha1",
```
codyoss, line 776:
```suggestion
		importPath:            "cloud.google.com/go/networkconnectivity/apiv1alpha1",
```

codyoss: Looks like we might need to upgrade our instructions, but at least this failed a unit test! The version should start with `api`. This was done as to not clash with Go's module version system where a folder named something like `v1` has meaning to Go tooling. After updating the code will need to be regenerated.

codyoss (changes requested):
It looks like genlocal still need to be run again. Doing so will move the generated artifacts to a new folder. You might need to manually delete out the file that were generated in the wrong folder.

codyoss (approved):
LGTM, thanks!

PR #3568, opened 2021-01-15
docs(storage): fix CustomTime docs

### LINKED ISSUE #3474: storage: Cannot set GCS Object CustomTime Attribute to zero value in Update()
**Client**

Storage

**Environment**

From go.mod:
>cloud.google.com/go v0.74.0

**Go Environment**

$ go version
> go version go1.11.13 darwin/amd64

$ go env
> GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/forest/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/forest/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/Cellar/go@1.11/1.11.13/libexec"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go@1.11/1.11.13/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/cd/9d8qff7j3dg8rz5qqrdl9gtw0000gn/T/go-build823705322=/tmp/go-build -gno-record-gcc-switches -fno-common"

**Code**

```go
func main() {
	//...
	ctx := context.Background()
	if err != nil {
		panic(err)
	}
	attrs, err := client.Bucket(bucket).Object(name).Attrs(ctx)
        // Some non-zero Custom-Time timestamp. e.g. 2020-12-17 17:23:35 +0000 UTC
	fmt.Printf("[Before] id:\"%s\" customTime:\"%s\"\n", attrs.Name, attrs.CustomTime)
	attrs, err = client.Bucket(bucket).Object(name).Update(ctx, storage.ObjectAttrsToUpdate{CustomTime: time.Time{}})
	if err != nil {
		panic(err)
	}
        // e.g. 2020-12-17 17:23:35 +0000 UTC (should be zero-value of time.Time)
	fmt.Printf("[After] id:\"%s\" customTime:\"%s\"\n", attrs.Name, attrs.CustomTime)
        //...
}
```

Console output showing the before and after states of the above code using live data:
> [Before] id:"85e5c77c1bb215ad327b372a0863f214" customTime:"2020-12-17 17:06:21 +0000 UTC"
> [After] id:"85e5c77c1bb215ad327b372a0863f214" customTime:"2020-12-17 17:06:21 +0000 UTC"
>
> [Before] id:"85e5c77c1bb215ad327b372a0863f214.info" customTime:"0001-01-01 00:00:00 +0000 UTC"
> [After] id:"85e5c77c1bb215ad327b372a0863f214.info" customTime:"0001-01-01 00:00:00 +0000 UTC"

The `85e5c77c1bb215ad327b372a0863f214` file already existed with a set `Custom-Time`. I need it unset.

The `85e5c77c1bb215ad327b372a0863f214.info` file was just created so it does not have a `Custom-Time`. This file is fine as is.

**Expected behavior**

I should be able to set the `Custom-Time` Attribute on the desired GCS Object to its zero value. 

**Actual behavior**

The documentation for this method indicates that in order to remove an Attribute from an Object, the field must be set to its zero value. Although the [GCS documentation](https://cloud.google.com/storage/docs/metadata#custom-time) indicates that the `Custom-Time` Attribute may not be updated on an Object, the `Update` method that is responsible for the logic in question ([lines](https://github.com/googleapis/google-cloud-go/blob/299de04ec22e3cc31d76ded257a86008960ea4bb/storage/storage.go#L885-L888)) appears to overwrite the existing Object entirely and create a new copy with the same name. In its current form, I am allowed to use this `Update` method in order to change the `Custom-Time` to any non-zero time.Time value. So, I know that it is certainly possible to circumvent the restrictions mentioned in the documentation for the `Custom-Time` Attribute.

The current behavior, show below, prevents the `CustomTime` field from being set to its zero value, unlike the other fields which are supported under the `optional` package ([lines](https://github.com/googleapis/google-cloud-go/blob/299de04ec22e3cc31d76ded257a86008960ea4bb/internal/optional/optional.go#L26-L44)).

https://github.com/googleapis/google-cloud-go/blob/299de04ec22e3cc31d76ded257a86008960ea4bb/storage/storage.go#L885-L888

My current workaround is to set the CustomTime to a sufficiently distant time in the future. This is not ideal, but it works for now.

**Recommendation**

The same logic used for the other fields in the `storage.ObjectAttrsToUpdate` struct, supported by the `optional` package, should be applied to the `CustomTime` field.

The following Type could be added to the `optional` package:
```go
// Time is either a time.Time or nil.
Time interface{}
```

Followed by the addition of a support function for this new Type:
```go
// ToTime returns its argument as a time.Time.
// It panics if its argument is nil or not a time.Time.
func ToTime(v Time) time.Time {
	x, ok := v.(time.Time)
	if !ok {
		doPanic("Time", v)
	}
	return x
}
```

Then the corresponding `storage.ObjectAttrsToUpdate.CustomTime` field must be updated accordingly:
```go
        //...
	//CacheControl       optional.String
	CustomTime         optional.Time
	//Metadata           map[string]string // set to map[string]string{} to delete
        //...
```

And finally, the `if` check in `storage.ObjectHandle.Update` needs to change to support the new `CustomTime` type:
```go
if uattrs.CustomTime != nil {
	attrs.CustomTime = optional.ToTime(uattrs.CustomTime)
	forceSendFields = append(forceSendFields, "CustomTime")
}
```

A fix like this would be backwards compatible for any clients using the `CustomTime` field from the version that it was introduced and upward.

**Additional context**

I upgraded from `v0.40.0` to `v0.74.0`, specifically to use the latter's support of the `storage.ObjectAttrsToUpdate.CustomTime` field.

tritone:
Update docs for ObjectAttrsToUpdate to account for behavior
of CustomTime field.

Closes #3474


storage/storage.go
```diff
@@ -941,7 +941,8 @@ func (o *ObjectHandle) ObjectName() string {
 
 // ObjectAttrsToUpdate is used to update the attributes of an object.
 // Only fields set to non-nil values will be updated.
-// Set a field to its zero value to delete it.
+// For all fields except CustomTime, set the field to its zero value to delete
+// it. CustomTime cannot be deleted or changed to an earlier time once set.
```
frankyn, line 945: For more information read: https://cloud.google.com/storage/docs/metadata#custom-time

frankyn (changes requested):
Suggestion

storage/storage.go
tritone: This is already present in the docs for ObjectAttrs, I don't think it needs to be repeated here (we don't go into detail for the other fields of this struct).

tritone: Actually I guess it isn't there, I will add it in that location instead.

frankyn (approved):

Issue #3570, opened 2021-01-16
spanner/spansql: ON clause is missing in JOIN query


ww24:
**Client**

Cloud Spanner

**Environment**

e.g. Alpine Docker on GKE

**Go Environment**

```
$ go version
go version go1.15.6 darwin/amd64

$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/ww24/Library/Caches/go-build"
GOENV="/Users/ww24/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/ww24/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/ww24/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.15.6/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.15.6/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/6s/c4848x1s28d6rf1wsnbb07qrv5hj0j/T/go-build364615131=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

**Code**

e.g.
```go
package main

import (
	"fmt"

	"cloud.google.com/go/spanner/spansql"
)

func main() {
	q := spansql.Query{
		Select: spansql.Select{
			List: []spansql.Expr{
				spansql.ID("A"), spansql.ID("B"),
			},
			From: []spansql.SelectFrom{
				spansql.SelectFromJoin{
					Type: spansql.InnerJoin,
					LHS:  spansql.SelectFromTable{Table: "Table1"},
[...]

PR #3571, opened 2021-01-16
fix(spanner/spansql): fix SelectFromJoin behavior

ww24:
Fix https://github.com/googleapis/google-cloud-go/issues/3570


dsymonds (approved):
LGTM

PR #3572, opened 2021-01-18
chore(spanner): source default opts from gapic client

### LINKED ISSUE #3532: spanner: source defaultClientOptions for grpc.Pool
Recently a new feature was rolled to use self-signed JWTs by default when using a service account with a client that has not been manually configured with options like scopes and endpoints. It was pointed out that spanner does not currently get this feature as it dials its own pool in the manual layer. There are two options I see as viable to get this functionality working.

1. Initialize the internal client options with [defaultClientOptions](https://github.com/googleapis/google-cloud-go/blob/6b432db005b77b22486fa5ab09f40a45d6499390/spanner/apiv1/spanner_client.go#L59). This way if more default options get rolled out in the future they will get auto-applied. Note if this option was chosen we would want to audit the grpcDialOptions in particualar to make sure we don't have a regression. It looks like a couple overlap in the generated client and manual layer.
2. Add two more options to the list of [allOpts](https://github.com/googleapis/google-cloud-go/blob/6b432db005b77b22486fa5ab09f40a45d6499390/spanner/client.go#L169). Notably replace option.WithScopes with internaloption.WithDefaultScopes and add `internaloption.WithDefaultAudience("https://spanner.googleapis.com/")`.

I am glad to contribute the required changes, but will defer to you on which approach makes the most sense for the client.

olavloite:
Sources the default client options from the generated client instead of the hand-written client. This makes sure that new default options that are introduced in the future will automatically be picked up by the Spanner client.

Fixes #3532

### DIFF
--- a/spanner/apiv1/spanner_client_options.go
+++ b/spanner/apiv1/spanner_client_options.go
@@ -1,0 +1,25 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spanner
+
+import "google.golang.org/api/option"
+
+// Returns the default client options used by the generated Spanner client.
+//
+// This function is only intended for use by the client library, and may be
+// removed at any time without any warning.
+func DefaultClientOptions() []option.ClientOption {
+	return defaultClientOptions()
+}
--- a/spanner/client.go
+++ b/spanner/client.go
@@ -205,8 +205,6 @@ func allClientOpts(numChannels int, userOpts ...option.ClientOption) []option.Cl
 		internaloption.EnableDirectPath(true),
 	}
 	allDefaultOpts := append(generatedDefaultOpts, clientDefaultOpts...)
-	// userOpts will take precedence above allDefaultOpts, as the values in
-	// userOpts will be applied after the values in allDefaultOpts.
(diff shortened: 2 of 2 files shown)


skuruppu (approved):

codyoss (approved):

olavloite:
@codyoss @skuruppu While testing something else today, I ran into a slight difference in behavior that was introduced by this change, and I want to check with you whether you consider that a problem:
1. Previously, the hand-written Spanner client would manage the default scopes itself, and set these when a new client was created. Now we source that from the default options of the generated client. The values were equal in both the hand-written and generated client. **BUT**: The hand-written client would set these by calling `WithScopes`, while the generated client calls `WithDefaultScopes`.
2. That means that when a connection is set up the client will now [use a self-signed JWT token](https://github.com/googleapis/google-api-go-client/blob/e9ea6ce8cd1d70391301fec53e453d41b2d560aa/internal/creds.go#L74) if the credentials that are used is a service account. Previous to this change, the client would use an OAuth2 flow. This works fine in most cases.
3. However if I use a different endpoint than the default endpoint, and my service account was generated for the default endpoint, I will get an `Unauthenticated` error. I can circumvent this by also calling `WithScopes(...)`.

So this means that the following code snippet would work prior to this change, but will not work any more when using a service account that has been created for the default endpoint, but that is also authorized for the custom endpoint:

```go
client, err := spanner.NewClient(
    context.Background(),
    "projects/my-project/instances/my-instance/databases/my-database",
    option.WithEndpoint("custom-endpoint.googleapis.com:443"),
)
```

The following is now needed to make it work:

```go
client, err := spanner.NewClient(
    context.Background(),
    "projects/my-project/instances/my-instance/databases/my-database",
    option.WithEndpoint("custom-endpoint.googleapis.com:443"),
    option.WithScopes("https://www.googleapis.com/auth/spanner.data"),
)
```

Is this something that we need to do something about?

codyoss:
@olavloite This sound like it might be a bug. I will investigate and get back to you. Thanks for bringing this to my attention.

codyoss:
@olavloite This should no longer be an issue on HEAD. Thanks for the report!

olavloite:
Great, thanks for the update.

PR #3585, opened 2021-01-20
chore: switch regen PRs to chores

codyoss:
The important bits will now be parsed footers. There is no need
for the regen PR itself to show up in the changelog.


codyoss:
cc @bcoe

tbpg (approved):

PR #3596, opened 2021-01-21
feat(spannertest): support alter not null with non-string, non-bytes types

petercgrant:
Adds support in `spannertest` for `alter` statements that change only `not null` constraints. This previously worked only for `string` or `bytes` columns.


dsymonds:
Thanks, but I think this part of the code needs a bigger overhaul. I'll have a go at that this week, and that should subsume this.

dsymonds:
Superseded by #3616.

PR #3597, opened 2021-01-22
fix(pubsublite): ignore outstanding acks for removed partition subscribers

tmdiep:
When the `assigningSubscriber` removes a `singlePartitionSubscriber`, it should ignore unacked messages when shutting it down (we still attempt to flush the commit cursor for messages that were already acked). This avoids conflicting with the commits from the new subscriber that will be assigned the partition.

Alternatively, we can wait for user code to ack all received messages before acking the new assignment. But this will delay the reassignment operation if user code is taking too long to ack the message.


tmdiep:
Discussed offline with @dpcollins-google to check that this is the behavior that we want. The Java and Python Lite libraries ignore outstanding acks, so we'll do the same in Go (this PR).

It would be better UX to wait for the outstanding unacked messages before acking the assignment, but if user code takes too long to ack the messages, the client would be removed from the assignment set on the server. The partition would be reassigned to another client, and there could potentially be conflicting commits.

manuelmenzella-google (approved):

PR #3600, opened 2021-01-22
feat(logging): add localIP and Cache fields to HTTPRequest conversion from proto

### LINKED ISSUE #3588: logging: missing assignment in function "toHTTPRequest"
In this [file](https://github.com/googleapis/google-cloud-go/blob/master/logging/logadmin/logadmin.go#L164) of line 164, we can see the assignment of logging.HTTPRequest's field "RemoteIP", and maybe the field "LocalIP" should be assigned also?

here shows the fixed:
image

nicoleczhu:
Fix #3588 

Changes: 
- Turned integration test back on for `logadmin`. According to previous comments, this may be flakey so I will closely monitor and debug. 
- Added fields LocalIP, CacheLookup, and CacheFillBytes back to httprequest conversion from proto to logging.httprequest


codyoss (approved):
LGTM

PR #3601, opened 2021-01-23
test(storage): Unflake TestIntegration_RequesterPays

### LINKED ISSUE #1753: storage: unflake TestIntegration_RequesterPays
Multiple flakes in the past week:
https://sponge.corp.google.com/target?id=3c0d95d4-29b3-4d0d-8953-0423aa0d7536&target=cloud-devrel/client-libraries/go/google-cloud-go/continuous/go113&searchFor=&show=ALL&sortBy=STATUS

https://sponge.corp.google.com/target?id=a1194c56-fd86-427a-8121-6393e31895f4&target=cloud-devrel/client-libraries/go/google-cloud-go/continuous/go112&searchFor=&show=ALL&sortBy=STATUS

```--- FAIL: TestIntegration_RequesterPays (18.19s)
    integration_test.go:3166: integration_test.go:2139: close write: googleapi: Error 429: The rate of change requests to the object go-integration-test-20200123-82263014289332-0007/foo exceeds the rate limit. Please reduce the rate of create, update, and delete requests., rateLimitExceeded
```

Looks like we need to introduce a short delay or retry here due to rate limits.

tritone:
Use an additional object for the ACL section of the test to avoid
rate limit flakes.

Fixes #1753


storage/integration_test.go
```diff
@@ -2212,6 +2211,10 @@ func TestIntegration_RequesterPays(t *testing.T) {
 	})
 
 	// ACL operations.
+	// Create another object for these to avoid object rate limits.
+	call("write object", func(b *BucketHandle) error {
+		return writeObject(ctx, b.Object("bar"), "text/plain", []byte("hello"))
```
frankyn, line 2216: Do you think tests should start using a UUID as a suffix?

tritone: In practice it hasn't been an issue except occasionally with this one test (and that's because this test needs to create an object and modify it several times quickly). I can add a UUID for each of foo and bar in this test though.

tritone: Honestly this won't make any difference to the result since the test still needs to modify the same object multiple times; however using a UID still seems nice/good so I did it.

storage/integration_test.go
```diff
@@ -2195,23 +2194,29 @@ func TestIntegration_RequesterPays(t *testing.T) {
 		}
 	}
 	// Object operations.
+	obj1 := uidSpace.New()
```
frankyn, line 2197: Nit, prefix with a term like `acl-go-test` so it's traceable back to this code section.

tritone: Done

codyoss (approved):

PR #3615, opened 2021-01-26
feat(internal/godocfx): xref function declarations

tbpg:
Type declarations are handled differently. So, I'll handle them in a
future PR. There are some TODOs at the top of parse.go with additional
ways we can improve.

I chose to define linkify in parse.go rather than pkgsite because (1) we
need the package/import name logic and (2) I want to keep changes to a
minimum in the third_party code.


codyoss (approved):
LGTM

Issue #2601, opened 2020-07-14
firestore: `QuerySnapshotIterator.Next()` can return `iterator.Done` even if `Stop()` hasn't been called


cpick:
**Client**

Firestore

**Environment**

`alpine:3.11` Docker image running on GKE

**Go Environment**

Built within docker image: `golang:1.14-alpine3.11`
$ go version
go version go1.14.4 linux/amd64
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/root/.cache/go-build"
GOENV="/root/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build686252092=/tmp/go-build -gno-record-gcc-switches"

**Code**

$ grep firestore go.mod
	cloud.google.com/go/firestore v1.2.0

```go
func (db *DB) watchSnapshots(ctx context.Context, id string, t time.Time, cb ForEachCallback) error {
	qsnaps := db.collection().
		Query.
		Where(idField, "==", id).
		Where(timeField, ">", t).
		OrderBy(timeField, firestore.Asc).
		Snapshots(ctx)
	defer qsnaps.Stop()

	for {
		qsnap, err := qsnaps.Next()
		if err == iterator.Done {
			return errors.New("stop called") // never expected since we only call `Stop()` via defer
		}
		if err != nil {
			return fmt.Errorf("next: %w", err)
		}

[...]

PR #3620, opened 2021-01-27
docs: add docstring to docref watch

crwilcox:
https://github.com/googleapis/google-cloud-go/issues/2601

Upon inspection, it seems we added a disclaimer to one of the two watch docstrs.


firestore/docref.go
```diff
@@ -795,7 +795,8 @@ type DocumentSnapshotIterator struct {
 // the current state of the document. If the document has been deleted, Next
 // returns a DocumentSnapshot whose Exists method returns false.
 //
-// Next never returns iterator.Done unless it is called after Stop.
+// Next is not expected to return iterator.Done unless it is called after Stop.
```
tritone, line 798: I am reluctant to add this because the reported issue was only for `QuerySnapshotIterator`. I have no idea whether this can also happen for this iterator or if the project is specific to QuerySnapshots.

crwilcox: If one path is having this, the other one is likely having it as well. Watch is implemented to setup a WatchStream with a target, which is a bit generic. The underlying stream they iterate on is the 'same' and they both are permuting io.EOF to iterator.Done

tritone: Ah yeah, I took another look and that seems correct. Makes sense.

tritone (approved):

PR #3640, opened 2021-02-01
fix(spanner/spannertest): support queries in ExecuteSql

### LINKED ISSUE #3639: spanner/spannertest: ping causes nil pointer dereference error
The ping method in the session pool is executed without a transaction selector. This will cause a [nil pointer dereference](https://github.com/googleapis/google-cloud-go/blob/c7ecf0f3f454606b124e52d20af2545b2c68646f/spanner/spannertest/inmem.go#L491) error in the `ExecuteSql` method of inmem.go.

https://stackoverflow.com/questions/65987211/golang-spanner-test-library-crashes-after-a-while

olavloite:
Normal queries from the Spanner client use the ExecuteStreamingSql method,
while DML statements use ExecuteSql. This distinction was also built into
spannertest where ExecuteSql would only support DML statements and required
a transaction to be specified. The session pool however uses ExecuteSql to
execute a simple `SELECT 1` query without specifying any transaction. This
would cause a nil pointer dereference.

This PR introduces support for queries in the ExecuteSql method. The current
logic assumes that the statement is a query if the transaction is a single-
use read-only transaction.

Fixes #3639


spanner/spannertest/inmem.go
```diff
@@ -485,9 +485,21 @@ func (s *server) readTx(ctx context.Context, session string, tsel *spannerpb.Tra
 }
 
 func (s *server) ExecuteSql(ctx context.Context, req *spannerpb.ExecuteSqlRequest) (*spannerpb.ResultSet, error) {
-	// Assume this is probably a DML statement. Queries tend to use ExecuteStreamingSql.
+	// Assume this is probably a DML statement or a ping from the session pool.
+	// Queries normally use ExecuteStreamingSql.
 	// TODO: Expand this to support more things.
 
+	// If it is a single-use transaction we assume it is a query.
+	if req.Transaction == nil ||
```
dsymonds, line 493: the proto getters work fine with nil receivers by design, so you can drop the first and third part of the clause.

```go
if req.Transaction.GetSelector() == nil || req.Transaction.GetSingleUse().GetReadOnly() != nil {
```

spanner/spannertest/inmem.go
```diff
@@ -640,6 +674,25 @@ func (s *server) readStream(ctx context.Context, tx *transaction, send func(*spa
 	return nil
 }
 
+func (s *server) buildResultSetMetaData(ri rowIter) (*spannerpb.ResultSetMetadata, error) {
```
dsymonds, line 677: "Metadata"

everything else here treats that as a single word.

spanner/spannertest/integration_test.go
```diff
@@ -137,6 +151,28 @@ func TestIntegration_SpannerBasics(t *testing.T) {
 	}
 	it.Stop()
 
+	// Try to execute the equivalent of a session pool ping.
```
dsymonds, line 154: Link to the bug and mention that this used to trigger a panic.

spanner/spannertest/integration_test.go
```diff
@@ -137,6 +151,28 @@ func TestIntegration_SpannerBasics(t *testing.T) {
...
+	s, err := generatedClient.CreateSession(ctx, &spannerpb.CreateSessionRequest{Database: dbName()})
+	if err != nil {
+		t.Fatalf("Creating session: %v", err)
+	}
+	rs, err := generatedClient.ExecuteSql(ctx, &spannerpb.ExecuteSqlRequest{
+		Session: s.Name,
+		Sql:     "SELECT 1",
+	})
+	if len(rs.Rows) != 1 {
```
dsymonds, line 163: check for an error first.

dsymonds (approved):

spanner/spannertest/inmem.go
olavloite: Done.

spanner/spannertest/inmem.go
olavloite: Done.

spanner/spannertest/integration_test.go
olavloite: Done.

spanner/spannertest/integration_test.go
olavloite: Done.

olavloite:
@skuruppu Apparently your review is also required in this case to allow merging.

skuruppu (approved):

PR #3643, opened 2021-02-01
fix(firestore): address a missing branch in watch.stop() error remapping

crwilcox:
While debugging another issue, I came across what appears to be a missing branch


tritone (approved):

firestore/watch.go
```diff
@@ -446,9 +446,12 @@ func (s *watchStream) stop() {
 		return
 	}
 	if err != nil {
+		// if an error occurs while closing the stream
 		s.err = err
+	} else {
```
codyoss, line 451: In general, prefer not including else blocks: https://golang.org/doc/effective_go.html#if

crwilcox: Happy to make this a return. I am going to wait for @jba to maybe weigh in. Since this is either code that should have been removed, or this branch was missed :)

jba:
Wow, that is definitely a bug. Thanks for catching. LGTM.

PR #3647, opened 2021-02-02
feat(pubsublite): allow increasing the number of topic partitions

tmdiep:
Allows increasing the number of partitions when updating a topic config. Updates documentation, examples and integration tests.

Will not be merged until server changes have been fully rolled out.


palmere-google (approved):

pubsublite/integration_test.go
```diff
@@ -189,6 +189,7 @@ func TestIntegration_ResourceAdminOperations(t *testing.T) {
 
 	topicUpdate1 := TopicConfigToUpdate{
 		Name:                       topicPath,
+		PartitionCount:             2,
```
hongalex, line 192: What happens when a `PartitionCount` doesn't change? Will the server throw an error here?

tmdiep: Good question. I just tried it and it's a no-op.

PR #3650, opened 2021-02-02
feat(pubsublite): optional logging settings for publisher and subscriber clients

tmdiep:
The pubsublite library handles a number of events automatically and most of this is opaque to users. We would like to log some key information/warning messages to help users troubleshoot issues. The logs are not intended to be verbose.

The first usages are:
* The [set of partitions](https://github.com/googleapis/google-cloud-go/blob/89ad55d72f79995a68f9c2ed1cd9b5ba50009d6d/pubsublite/internal/wire/assigner.go#L153) that a subscriber client was assigned by the server. We considered adding hooks to notify user code of assignment updates, but ultimately decided to just log this information. Users need to know the partition numbers to filter metrics in order to troubleshoot issues.
* Warning when a [partition count decrease](https://github.com/googleapis/google-cloud-go/blob/89ad55d72f79995a68f9c2ed1cd9b5ba50009d6d/pubsublite/internal/wire/publisher.go#L317) was ignored.

This initial PR just introduces the setting.

There are not many precedents for logging in other Go client libraries. This setting is similar to the `OnError` func in [logging.Client](https://pkg.go.dev/cloud.google.com/go/logging#Client). The only other non-test example of logging was the warning messages in [datastore.NewClient](https://github.com/googleapis/google-cloud-go/blob/7f8e434b6a614de1a0dc2b66890db1888cd079af/datastore/datastore.go#L94).

### DIFF
--- a/pubsublite/pscompat/settings.go
+++ b/pubsublite/pscompat/settings.go
@@ -41,7 +41,8 @@ type KeyExtractorFunc func(*pubsub.Message) []byte
 // terminate.
 type PublishMessageTransformerFunc func(*pubsub.Message, *pb.PubSubMessage) error
 
-// LogFunc is a function to log informational messages.
+// LogFunc is a function to log informational messages. It may be called
+// concurrently from multiple goroutines and is expected to return quickly.
 type LogFunc func(v ...interface{})
 
 // PublishSettings configure the PublisherClient. Batching settings
--- a/pubsublite/pscompat/settings.go
+++ b/pubsublite/pscompat/settings.go
@@ -41,6 +41,9 @@ type KeyExtractorFunc func(*pubsub.Message) []byte
 // terminate.
 type PublishMessageTransformerFunc func(*pubsub.Message, *pb.PubSubMessage) error
 
+// LogFunc is a function to log informational messages.
+type LogFunc func(v ...interface{})
+
 // PublishSettings configure the PublisherClient. Batching settings
 // (DelayThreshold, CountThreshold, ByteThreshold, BufferedByteLimit) apply per
 // partition.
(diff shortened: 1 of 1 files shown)


pubsublite/pscompat/settings.go
```diff
@@ -41,6 +41,10 @@ type KeyExtractorFunc func(*pubsub.Message) []byte
 // terminate.
 type PublishMessageTransformerFunc func(*pubsub.Message, *pb.PubSubMessage) error
 
+// LogFunc is a function to log informational messages. It may be called
+// concurrently from multiple goroutines and is expected to return quickly.
+type LogFunc func(v ...interface{})
```
dsymonds, line 46: I don't think this warrants a named type.

pubsublite/pscompat/settings.go
```diff
@@ -95,6 +99,11 @@ type PublishSettings struct {
 	// Optional custom function that transforms a pubsub.Message to a
 	// PubSubMessage API proto.
 	MessageTransformer PublishMessageTransformerFunc
+
+	// Optional function to log informational messages that may be useful for
+	// debugging. For example, this can be set to log.Println. If not set,
+	// messages will not be logged.
+	OnLog LogFunc
```
dsymonds, line 106: OnFoo funcs are typically invoked when "foo" happens. when does "log" happen? here, this should either be something like "OnError" (invoked when something goes wrong), or named more directly like "Logger".

dsymonds (commented):
This feels potentially useful, but also feels like it should be designed for google-cloud-go as a whole, not just for a subpackage of a single module. e.g. some sort of hook that can be added at a client construction time.

tmdiep:
@tbpg @codyoss

I like the idea of a common mechanism in google-cloud-go. This is not particularly urgent for pubsublite, so happy to shelve this PR in favor of designing that instead.

codyoss:
Currently I don't believe any of our client do any logging, there are a couple exceptions but we have generally tried to avoid it. Instead some publish some telemetry and make use of [traces](cloud.google.com/go/internal/trace). What sort of information were you planning on exposing?

tmdiep:
> Lets bring this up at the next libraries meeting for disscussion.

@codyoss Thanks! Unfortunately I won't be able to join due to timezones. Will check meeting notes for the outcome.

PR #3656, opened 2021-02-02
feat(internal/docfx): full cross reference linking

tbpg:
The changes to `pkgsite` are mostly copied and pasted. I modified it a little to make the API easier to use. I also added the explanatory comment at the top of `PrintType` because it took a while for me to wrap my head around it.


codyoss (approved):
LGTM

PR #3662, opened 2021-02-03
fix(pubsublite): change pubsub.Message.ID to an encoded publish.Metadata

tmdiep:
This mirrors the Java implementation: https://github.com/googleapis/java-pubsublite/pull/482. If the user's custom message transformer sets the `pubsub.Message.ID`, the `SubscriberClient` will terminate with error. The internal wire subscriber needs to pass the partition number to the CPS subscriber shim via `ReceivedMessage`.

Updated tests and documentation. Changed the comments for `publish.ParseMetadata` to inline documentation examples.

### DIFF
--- a/pubsublite/internal/wire/subscriber.go
+++ b/pubsublite/internal/wire/subscriber.go
@@ -37,8 +37,9 @@ var (
 // ReceivedMessage stores a received Pub/Sub message and AckConsumer for
 // acknowledging the message.
 type ReceivedMessage struct {
-	Msg *pb.SequencedMessage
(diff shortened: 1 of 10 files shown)


tmdiep:
Note: as was done for the Java lib in https://github.com/googleapis/java-pubsublite/pull/482, we will rename the publish metadata to MessageMetadata. This will be done in a subsequent PR to isolate unrelated changes and add other reviewers for the public API change.

manuelmenzella-google (approved):

PR #3665, opened 2021-02-03
fix(internal/godocfx): use exact list of top-level decls

tbpg:
Here is a resulting diff from the `firestore` package:
```
-    content: var MergeAll <a href="#setoption">SetOption</a> = <a href="#merge">merge</a>{/*
-      contains filtered or unexported fields */}
+    content: var MergeAll <a href="#setoption">SetOption</a> = merge{/* contains filtered
+      or unexported fields */}
```

Note that `merge` should not be linked. This PR fixes that.


codyoss (approved):
LGTM

PR #3671, opened 2021-02-04
test(pubsublite): fix flaky TestAssignerHandlePartitionFailure

tmdiep:
Checks only the final error to ensure the test is deterministic.

Fixes https://github.com/googleapis/google-cloud-go/issues/3667.


codyoss (approved):
LGTM

PR #3692, opened 2021-02-10
chore(all): rename file to not conflict with licensee

codyoss:
Related: https://github.com/licensee/licensee/issues/479


tbpg (approved):
Let's hope this works!

PR #3703, opened 2021-02-12
docs(all): add details on how to cast errors to grpc.Status

codyoss:


doc.go
```diff
@@ -85,6 +85,20 @@ https://godoc.org/google.golang.org/grpc/grpclog for more information.
 For HTTP logging, set the GODEBUG environment variable to "http2debug=1" or "http2debug=2".
 
 
+Inspecting errors
+
+All of the errors returned by the generated clients can be converted into a
```
tbpg, line 90: Will this always be "all"? Are we doing REST GAPICs? Also, users probably don't know which clients are generated and which aren't.

doc.go
```diff
@@ -85,6 +85,20 @@ https://godoc.org/google.golang.org/grpc/grpclog for more information.
...
+`grpc.Status`. Converting your errors to this type can be a useful to get
+more information about what went wrong while debugging.
+ if err != {
+    if s, ok := status.FromError(err); ok {
+	   log.Println(s.Message())
+		for _, d := range s.Proto().Details {
+		   log.Println(d)
+		}
```
tbpg, line 98: Extra indentation here?

tbpg (changes requested):

doc.go
codyoss: That is a very good call out. It is true when we start to have to REST GAPICs that this will no longer be true. I will switch the language to "Most". I think the example is still good as I made sure to do the type check to be safe.

doc.go
codyoss: done

tbpg (approved):

PR #3707, opened 2021-02-17
chore(internal/gapicgen): enable metadata gen

noahdietz:
* add metadata flag to configs
* update gapic-generator-go to v0.18.2
* gen code changes include:
  * generation of gapic_metadata.json
  * removal of stuttering import aliases for `*pb` packages like `structpb`

### DIFF
--- a/internal/gapicgen/generator/config.go
+++ b/internal/gapicgen/generator/config.go
@@ -42,6 +42,10 @@ type microgenConfig struct {
 	// useful if a client needs to be deprecated, but retained in the repo
 	// metadata.
 	stopGeneration bool
+
+	// disableMetadata is used to toggle generation of the gapic_metadata.json
+	// file for the client library.
+	disableMetadata bool
 }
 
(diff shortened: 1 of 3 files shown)


internal/gapicgen/generator/config.go
```diff
@@ -42,6 +42,10 @@ type microgenConfig struct {
 	// useful if a client needs to be deprecated, but retained in the repo
 	// metadata.
 	stopGeneration bool
+
+	// metadata is used to toggle generation of the gapic_metadata.json file
+	// for the client library.
+	metadata bool
```
codyoss, line 48: When is this not set to true? Should we reverse the var so by default it does not need to be set for every bit of config?

noahdietz: Good idea. idk of a case where it shouldn't be enabled.

noahdietz: Done!

codyoss (approved):
LGTM

PR #3716, opened 2021-02-19
feat(spansql/spannertest): support NOT IN clause

### LINKED ISSUE #3715: spannertest/spansql: Support "NOT IN" clause
I noticed "NOT IN" is not supported, while simple "IN" is supported.
Actually I got the following error.

```
spanner: code = "InvalidArgument", desc = "bad query: got \"IN\", want LIKE or BETWEEN"
```

- Cause
The function to parse Comparison Operators seems to try to parse "NOT IN" and treat it as unexpected tokens.

- Solution
Make this function to skip parsing when "NOT IN" is provided.

sryoya:
# About
Fixes #3715

# Note
I made the function return early when it finds "IN" as well as "NOT IN", though "IN" can be already skipped in the following codes. That's because I suppose it's more readable to notice that "IN" including "NOT IN" is out of the parsing scope of this function.
For the same point, I put this validation before `if tok.value == "NOT"`, not in this "if" statement.


dsymonds:
See #3724 instead.

sryoya:
Ok, thanks! Sounds better.

PR #3724, opened 2021-02-22
fix(spanner/spansql): fix parsing of NOT IN operator

### LINKED ISSUE #3715: spannertest/spansql: Support "NOT IN" clause

dsymonds:
The parsing of comparison operators was erroneously capturing any NOT
token assuming it would be followed by a LIKE or BETWEEN. Instead,
change that code to explicitly look ahead for the limited combinations
allowed, and do the same for parsing IN and NOT IN.

Fixes #3715.

### DIFF
--- a/spanner/spansql/parser.go
+++ b/spanner/spansql/parser.go
@@ -2389,15 +2389,12 @@ func (p *parser) parseInOp() (Expr, *parseError) {
 		return nil, err
 	}
 
-	// TODO: do we need to do lookahead?
-
 	inOp := InOp{LHS: expr}
-	if p.eat("NOT") {
+	if p.eat("NOT", "IN") {
 		inOp.Neg = true
-	}
-
-	if !p.eat("IN") {
-		// TODO: push back the "NOT"?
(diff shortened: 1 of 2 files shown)


skuruppu (approved):

dsymonds:
Thanks, can you submit @skuruppu? I no longer have access to do that.

skuruppu:
@dsymonds for some reason, I'm not able to click a button to update the branch to pick up latest changes from main.

dsymonds:
Okay, manually rebased. Try again please.

skuruppu:
Seems good. Now waiting for Kokoro to run.

PR #3738, opened 2021-02-25
fix(internal/godocfx): use correct anchor links

### LINKED ISSUE #3737: godocfx: page-local anchor links broken
`godocfx` generates anchor links for the current page based on the unqualified name. But, the IDs are based on the IDs of that element. We should update the link generator to use the correct anchor when linking to an item on the current page.

Right now, the links are broken.

tbpg:
This works by pre-processing the package and generating all of the correct anchor links. Then, when converting an ID to a link, we check if the ID is listed in the map of anchors.

It's possible there will be some duplication between IDs. But, that is already an issue because there is no way to know which one is correct to link to. An alternative would be to set the anchor for any duplicated links to `""` so they don't link at all. In practice, I don't think this is an issue. So, I left it as-is and we can revisit later if needed.

Fixes #3737.

### DIFF
--- a/internal/godocfx/parse.go
+++ b/internal/godocfx/parse.go
@@ -34,6 +34,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


codyoss (approved):
LGTM!

PR #3745, opened 2021-02-26
test(pubsublite): print proto type in verifier error messages

tmdiep:
Helps debugging failed unit tests.


codyoss (approved):
There is also %T for formatted strings.

tmdiep:
Thanks Cody!
%T is much nicer.

PR #3749, opened 2021-02-26
test(bigquery): update and re-enable a streaming payload limit test

shollyman:
BigQuery updated the streaming limits, breaking this test.  Those
limits are now consistent once again, so we re-enable the test in
accordance with new simplified limits.

Fixes: https://github.com/googleapis/google-cloud-go/issues/3612


codyoss (approved):
LGTM

PR #3752, opened 2021-02-28
fix(pubsub): hide context.Cancelled error in sync pull

### LINKED ISSUE #3746: pubsub: Receive fails to hide context.Canceled error when using synchronous pull
**Client**

PubSub

**Environment**

Various

**Go Environment**

$ go version
go version go1.16 linux/amd64
<details>
<summary>$ go env</summary>

```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/mgl/.cache/go-build"
GOENV="/home/mgl/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/mgl/src/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/mgl/src/go"
GOPRIVATE=""
GOROOT="/usr/lib/go-1.16"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go-1.16/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.16"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/dev/null"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build4089208195=/tmp/go-build -gno-record-gcc-switches"
```
</details>

**Code**

```go
func foo(client *pubsub.Client) {
	sub := client.Subscription("foo")
	// enabling synchronous receive causes Receive to return a context cancelled error instead of nil
	sub.ReceiveSettings.Synchronous = true
	ctx, cancel := context.WithCancel(context.Background())
	err := sub.Receive(ctx, func(c context.Context, m *pubsub.Message) {
		m.Ack()
		cancel()
	})
	if err != nil {
		panic(err)
	}
}
```

**Expected behavior**

The call to `sub.Receive` should return a `nil` error when it is terminated through context cancellation.

**Actual behavior**

The call to `sub.Receive` returns a version of `context.Canceled`, wrapped/converted to a gRPC status error, if and only if `Synchronous` pull is enabled.

**Additional context**

I've traced this so far as this error handling logic in `messageIterator.pullMessages`: https://github.com/googleapis/google-cloud-go/blob/39517ae11355340ce44ed192a0e096580ef5ee4e/pubsub/iterator.go#L247-L254

In this case, it's trying to replace the context cancellation error with nil, but that cancellation error has already been wrapped into a gRPC one at this point ... which isn't surprising, because it came from a gRPC client call :)

I believe this can be fixed by updating the `switch` thus:
```go
	switch {
	case err == context.Canceled:
		return nil, nil
	case status.Code(err) == codes.Canceled: // ⇐ add this case
		return nil, nil
	case err != nil:
		return nil, err
	default:
		return res.ReceivedMessages, nil
	}
```

hongalex:
As described, in the linked bug below, `sub.Receive` should not propagate errors from user cancellations. This is true for streaming pull but not synchronous pull.
This PR adds a simple switch case as suggested by @mgabeler-lee-6rs and includes tests to guard against this.

Fixes #3746


pubsub/iterator.go
```diff
@@ -247,6 +247,8 @@ func (it *messageIterator) pullMessages(maxToPull int32) ([]*pb.ReceivedMessage,
 	switch {
 	case err == context.Canceled:
 		return nil, nil
+	case status.Code(err) == codes.Canceled:
```
codyoss, line 250: This will also now mask any rpcs that are canceled server side too, right? Is this okay?

hongalex: Just confirmed that server doesn't cancel any RPCs intentionally (they should always be initiated from client).

codyoss: Ack

codyoss (approved):
LGTM

PR #3759, opened 2021-03-02
fix(tables): use area120tables_v1alpha1.yaml as api-service-config

noahdietz:
Use proper API Service Config file for area120/tables GAPIC.


codyoss (approved):
LGTM

PR #3760, opened 2021-03-02
feat(bigquery): support type alias names for numeric/bignumeric schemas

shollyman:
BigQuery documents aliases DECIMAL for NUMERIC, and BIGDECIMAL for BIGNUMERIC.  Add them to the alias list.


codyoss (approved):
LGTM

PR #3767, opened 2021-03-03
feat(internal/godocfx): keep some cross links on same domain

### LINKED ISSUE #3739: godocfx: link cross references to same domain when needed
If we're generating links to other pages, we should be sure to stay on the preferred domain for the package we're linking to.

tbpg:
The magic of this PR (that took me way too long to figure out) comes from adding `| packages.NeedDeps` to the `packages.Load` `Config`. That lets us get the `Fset` and `Syntax` of all packages the current package imports. We can then use that information to pull out the anchors for every element we link to.

This also fixes an issue where we were handling some packages that didn't have the Module field set -- we can ignore them. It appears to be a new package gets returned with the ID of the glob you asked for. For example, `cloud.google.com/go/...`. We don't seem to need that.

Fixes #3739.

### DIFF
--- a/internal/godocfx/godocfx_test.go
+++ b/internal/godocfx/godocfx_test.go
@@ -168,3 +168,33 @@ func TestGoldens(t *testing.T) {
 		}
 	}
 }
+
+func TestHasPrefix(t *testing.T) {
+	tests := []struct {
+		s        string
+		prefixes []string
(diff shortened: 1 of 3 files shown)


codyoss (approved):
Nice, LGTM!

PR #3769, opened 2021-03-03
chore(internal/kokoro): port check_incompat_changes

### LINKED ISSUE #3764: internal/kokoro: collect packages to apidiff with a regexp, not a list
`check_incompat_changes.sh` only diffs those packages explicitly listed [here](https://github.com/googleapis/google-cloud-go/blob/master/internal/kokoro/check_incompat_changes.sh#L37-L38). This is not scalable and it is currently out of date. We should use a regexp of some sort to list out all apiv1 packages for diffing.

noahdietz:
Ports the `apidiff` logic of `check_incompat_changes.sh` into a Go program `internal/apidiff`. 

This new version reads from the `internal/.repo-metadata-full.json` file and runs `apidiff` against all packages labeled as `ga`. 

When it encounters a manual client or a generated subclient of a submodule, it attempts to find the submodule, change into that directory, and run `apidiff` from there. 

It will collect all errors/diffs and report them at the end. The script will only fail if there are diffs. If there is an error during apidiff of one of the packages, it will be reported, but it will not fail the script. 

Fixes #3764 by parsing the repo-metadata, not a regexp in a bash script.


noahdietz:
Currently failing because `apidiff` is complaining about a constant that doesn't actually change in `pubsub`
```
cloud.google.com/go/pubsub:
- MaxPublishRequestBytes: value changed from 0.000582077 to 10000000
```

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,255 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+)
+
+const rootMod = "cloud.google.com/go"
+
+var repoMetadataPath string
```
codyoss, line 29: nit: Could you move these into main to avoid global state?

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,255 @@
...
+var verbose bool
+
+var root string
```
codyoss, line 32: nit: Make these fields a struct and have methods that operate on them?

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,255 @@
...
+var manifest map[string]manifestEntry
+var diffs map[string]string
+var errs map[string]error
+
+func init() {
+	flag.StringVar(&repoMetadataPath, "repo-metadata", "", "path to a repo-metadata-full JSON blob [required]")
+	flag.BoolVar(&verbose, "verbose", false, "enable verbose command logging")
+
+	root, _ = os.Getwd()
```
codyoss, line 41: If this was done outside of init, see above, I think we could check the error.

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,255 @@
...
+	head := mustExec("git", "log", "-1")
+	if strings.Contains(head, "BREAKING_CHANGE_ACCEPTABLE") {
+		os.Exit(0)
+	}
+}
+
+func loadManifest() {
+	f, err := os.Open(repoMetadataPath)
+	if err != nil {
+		log.Fatalln(err)
```
codyoss, line 195: nit: for this and other error could we return the error and only os.Exit from the main function?

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,255 @@
...
+// limitations under the License.
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
```
codyoss, line 22: We should either add a build constraint to linux,darwin or use golang.org/x/sys/execabs due to a security issue. See https://blog.golang.org/path-security for more details.

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+	}
+
+	return out, err
+}
+
+func shouldRun() {
+	head := mustExec("git", "log", "-1")
+	if strings.Contains(head, "BREAKING_CHANGE_ACCEPTABLE") {
+		os.Exit(0)
```
codyoss, line 188: nit: Return a bool and exit from main. Maybe log something?

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+}
+
+func loadManifest() {
+	f, err := os.Open(repoMetadataPath)
```
codyoss, line 193: Pleas defer a Close.

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+	err = json.NewDecoder(f).Decode(&manifest)
```
codyoss, line 197: nit: Can be combined with the line below.

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+}
+
+func cloneGoCloud() string {
+	temp, err := os.MkdirTemp("/tmp", "google-cloud-go-*")
```
codyoss, line 204: The main module is 1.11 compatible. So we have to keep using `ioutil.TempDir` for now.

codyoss (changes requested):
Mostly nits, overall LGTM.

internal/apidiff/apidiff.go
noahdietz: Adding a build constraint. Thanks! TIL.

internal/apidiff/apidiff.go
noahdietz: Done!

internal/apidiff/apidiff.go
noahdietz: Done.

internal/apidiff/apidiff.go
noahdietz: Done.

internal/apidiff/apidiff.go
noahdietz: Done.

internal/apidiff/apidiff.go
noahdietz: Done.

internal/apidiff/apidiff.go
noahdietz: Yep, that sounds good. Done.

internal/apidiff/apidiff.go
noahdietz: Can I keep these as they are flag value containers? `versbose` is used in many other functions, it would be annoying to pass around a local variable from main.

internal/apidiff/apidiff.go
noahdietz: Done.

internal/kokoro/check_incompat_changes.sh
```diff
@@ -21,33 +21,4 @@ if [[ `go version` != *"go1.16"* ]]; then
     exit 0
```
noahdietz, line 21: @codyoss I could move this check into `shouldRun()` if you want to get rid of this script altogether. WDYT?

noahdietz:
Blocked by https://github.com/golang/go/issues/44796

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,284 @@
...
+
+	return pkg
+}
+
+func shouldRun() (bool, error) {
+	head, err := execE("git", "log", "-1")
+	if err != nil {
+		return false, err
+	}
+	return !strings.Contains(head, "BREAKING_CHANGE_ACCEPTABLE"), nil
```
noahdietz, line 255: @codyoss can I change this to check that the commit either contains `BREAKING_CHANGE`  _or_ starts with `!`? That way the conventional commit practices used in today's commits will _just work_, especially on go-genproto.

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,284 @@
...
+		out, err = execE("apidiff", "-w", base, imp)
+		if err != nil {
+			return "", fmt.Errorf("%s: %s", err, out)
+		}
+	}
+
+	return out, err
+}
+
+func (d *differ) loadManifest() error {
```
noahdietz, line 225: @codyoss just realizing that we cannot use this script on go-genproto because there is no repo-metdata file there... :/

internal/apidiff/apidiff.go
tbpg: Breaking changes should be rare and difficult to make. I'm OK with requiring `BREAKING_CHANGE` in the body.

noahdietz: It's actually required by those publishing proto changes to googleapis today. They must either provide a `BREAKING_CHANGE` tag in the body or start their commit with `!` e.g. `!fix: remove accidentally published field`. So we'd be piggy-backing on the existing practice.

noahdietz:
Oops, I messed up my git history some how and ive got waaaay too many commits that don't belong. Ignore this please!

noahdietz:
@tbpg @codyoss can I get another pass on this? I am manually ignoring the incorrect pubsub error (if it is the only error there is  for that pkg)

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+func cd(dir string) error {
+	if verbose {
+		log.Printf("+ cd %s\n", dir)
+	}
+	return os.Chdir(dir)
+}
+
+func execE(cmd string, args ...string) (string, error) {
```
tbpg, line 281: What does the E stand for?

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,287 @@
...
+	flag.BoolVar(&verbose, "verbose", false, "enable verbose command logging")
+}
+
+func main() {
+	flag.Parse()
+	if repoMetadataPath == "" {
+		log.Fatalln("Missing required flag: -repo-metadata")
+	}
+
+	run, err := shouldRun()
```
tbpg, line 50: I think `shouldRun` can be inlined -- about the same number of lines with error handling.

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+	if !run {
+		log.Println("Not running apidiff because description contained tag BREAKING_CHANGE_ACCEPTABLE.")
+		return
+	}
+
+	d, err := newDiffer()
```
tbpg, line 59: Seems like `newDiffer` can be inlined, too.

Overall, there is a lot of indirection in `func main`, but I don't understand the names, so I don't know what is happening.

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	err = d.loadManifest()
```
tbpg, line 64: A couple functions get called in `func main` and modify state in `d` (this, `diffModules`). Why? Reading this, I don't know what is being modified, or what the result of `loadManifest` is, or why it is a method at all.

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	_, err = execE("go", "install", "golang.org/x/exp/cmd/apidiff@latest")
```
tbpg, line 69: Can we have a `tools/go.mod` or something so we can just `go get` the version of the dep we need?

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	err = d.diffModules()
```
tbpg, line 74: Same question -- where is the result of this? I'd expect a `diff` result from `diffModules`. Instead, it gets stored in `d.results`, and I don't know that without reading the whole thing.

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	hadErrors := len(d.errs) > 0
```
tbpg, line 79: Why the indirection with `errs`?

tbpg (changes requested):
Sounds good to move this to a Go program, but I'm having trouble following the flow of execution. I left some comments about the APIs. Overall, it feels more... complicated (?) than I expected. Hopefully the suggestions I gave help simplify the flow.

internal/apidiff/apidiff.go
noahdietz: I didn't want any single error to stop the diffing of the remaining packages. So the errors are collected and reported at the end.

internal/apidiff/apidiff.go
noahdietz: This is kind of a hold over from an earlier version - the E suffix means that it returns the error rather than panics (previously had a `mustExec` that panicked)

noahdietz:
> Overall, it feels more... complicated (?) than I expected

Yeah that's reasonable - it is more complicated. This now:
- supports diffing submodules, which the previous script didn't
- reads a config file instead of iterating over a hard coded list
- collects all diffs/errors before reporting instead of exiting at first issue

So it's not a direct port, it's a "port+" :) Your suggestions are totally valid though.

internal/apidiff/apidiff.go
noahdietz: I had started with a lot of global state, it was then suggested I wrap it in a struct to capture the global state. I should've just made `main` slightly more verbose for readability. Inlined this bit.

internal/apidiff/apidiff.go
noahdietz: Done. I removed most indirection.

internal/apidiff/apidiff.go
noahdietz: Done.

internal/apidiff/apidiff.go
noahdietz: Done. I made it return the diffs and diffing errors.

internal/apidiff/apidiff.go
noahdietz: I changed it to `exec` and alias `os/exec` to `osexec`

internal/apidiff/apidiff.go
noahdietz: Could that be done separately? Also, since `x/exp` isn't versioned and it hardly changes, at least `apidiff` doesn't, could we just use latest?

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,259 @@
...
+
+type manifest map[string]manifestEntry
+
+func diffModules(root string, m manifest) (map[string]string, map[string]error, error) {
+	diffs := map[string]string{}
+	issues := map[string]error{}
+
+	goCloud, err := cloneGoCloud()
+	if err != nil {
+		return diffs, issues, err
```
tbpg, line 120: Probably just `nil, nil, err`?

internal/apidiff/apidiff.go
```diff
@@ -0,0 +1,259 @@
...
+	ReleaseLevel      string `json:"release_level"`
+}
+
+type manifest map[string]manifestEntry
+
+func diffModules(root string, m manifest) (map[string]string, map[string]error, error) {
+	diffs := map[string]string{}
+	issues := map[string]error{}
+
+	goCloud, err := cloneGoCloud()
```
tbpg, line 118: Could get rid of the `error` return if we clone GoCloud before calling this function. Doesn't really matter.

tbpg (approved):
Much more understandable -- thank you!

internal/apidiff/apidiff.go
noahdietz: Moved this into main

internal/apidiff/apidiff.go
noahdietz: Moved this into main and removed `cloneGoCloud`

noahdietz:
@codyoss I think you need to approve since you requested changes.

codyoss (approved):

codyoss:
@noahdietz Sorry about that, did not realize this was waiting on me.

noahdietz:
Only waiting like 1 min :)

PR #3778, opened 2021-03-08
test(storage): add retry conformance tests [WIP]

tritone:


storage/conformance_test.go
```diff
@@ -29,8 +34,191 @@ import (
...
+// 2. A list of non-idempotent calls (should give expectSuccess = False for
+//    everything).
+var idempotentOps = []retryFunc{
+	// storage.objects.get
+	func(ctx context.Context, c *Client) error {
+		_, err := c.Bucket(bucketName).Object(objName).Attrs(ctx)
+		return err
+	},
+	// GET object (XML API)
+	func(ctx context.Context, c *Client) error {
```
frankyn, line 69: How do you reset the retry counter for this subsequent call?

storage/conformance_test.go
```diff
@@ -29,8 +34,191 @@ import (
...
+		r, err := c.Bucket(bucketName).Object(objName).NewReader(ctx)
+		if err != nil {
+			return err
+		}
+		_, err = io.Copy(ioutil.Discard, r)
+		return err
+	},
+	// storage.objects.update (preconditions)
```
frankyn, line 77: Ah i see, you're including conditionally idempotent operations here as well

storage/conformance_test.go
```diff
@@ -16,21 +16,315 @@ package storage
...
+	//}
+	return resp, err
+}
+
+// Create custom client that sends instruction
+func wrappedClient(host, testID string) (*Client, error) {
+	ctx := context.Background()
+	base := http.DefaultTransport
+	trans, err := htransport.NewTransport(ctx, base, option.WithScopes(raw.DevstorageFullControlScope),
+		option.WithUserAgent("custom-user-agent"))
```
frankyn, line 313: You don't  need to supply `option`'s do you?

storage/conformance_test.go
```diff
@@ -16,21 +16,315 @@ package storage
...
+		return nil, fmt.Errorf("failed to create http client: %v", err)
+	}
+	c := http.Client{Transport: trans}
+
+	// Add RoundTripper to the created HTTP client.
+	wrappedTrans := &withTestID{rt: c.Transport, testID: testID}
+	c.Transport = wrappedTrans
+
+	// Supply this client to storage.NewClient
+	client, err := NewClient(ctx, option.WithHTTPClient(&c), option.WithEndpoint(host + "/storage/v1/"))
```
frankyn, line 324: Does with `WithEndpoint` support host only or does it need `/storage/v1/`*?

frankyn (changes requested):

storage/conformance_test.go
tritone: Yeah that's true, it's only needed for the real service I believe. Will remove.

storage/conformance_test.go
tritone: It needs the full endpoint unfortunately.

PR #3780, opened 2021-03-08
fix(pubsub): update quiescenceDur in failing e2e test

### LINKED ISSUE #3773: pubsub/internal/longtest: TestEndToEnd_LongProcessingTime failed
This test failed!

I will stop commenting.

---

commit: f88bdc85072e5ad511a907d98207ebf7d22e9df7
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/fdc7e9bd-aa37-4d98-9045-c37046d91aef), [Sponge](http://sponge2/fdc7e9bd-aa37-4d98-9045-c37046d91aef)
status: failed
<details><summary>Test output</summary><br><pre>    endtoend_test.go:302: 0: start receive
2021/03/04 08:12:41 quiesced
    endtoend_test.go:253: timed out waiting for consumer to finish</pre></details>

hongalex:
`TestEndToEnd_LongProcessingTime` used to depend on a grpc `cancelled` error to finish before the 15 second test timer expires. After #3752, we hide these cancelled errors in synchronous pull to be consistent with streaming pull. 

Since no fatal error was being thrown, the `Receive` call gets stuck as we stop processing after the default 60 seconds. Increase this to 120 seconds (or 12 * ackDeadline), which is the upper bound of processing times so all messages received are eventually acked and doesn't block the `Receive` call. Also, decrease the number of messages published so we only have one batch received. This could have been left as-is, but we would increase the total time the test takes, and is unnecessary to what the test was trying to accomplish.

Fixes #3773


codyoss (approved):
LGTM

PR #3790, opened 2021-03-10
feat(pubsublite): adding ability to create subscriptions at head

hannahrogers-google:


tmdiep (commented):
Thanks Hannah!

pubsublite/admin.go
```diff
@@ -147,9 +147,28 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
...
+	End StartingOffset = iota
+
+	// Beginning represents the offset of the oldest retained message.
+	Beginning
+)
+
 // CreateSubscription creates a new subscription from the given config. If the
 // subscription already exists an error will be returned.
 func (ac *AdminClient) CreateSubscription(ctx context.Context, config SubscriptionConfig) (*SubscriptionConfig, error) {
+	return ac.CreateSubscriptionAtOffset(ctx, config, End)
```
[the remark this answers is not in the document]
hannahrogers-google: After talking with Kir/Manu, we decided to make the default End for consistency with CPS

pubsublite/admin.go
```diff
@@ -147,9 +147,28 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
 	}
 }
 
+// StartingOffset is the offset at which a newly created subscription will start
+// receiving messages.
+type StartingOffset int
+
+const (
+	// End represents the current HEAD offset.
+	End StartingOffset = iota
```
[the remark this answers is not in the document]
hannahrogers-google: Done.

pubsublite/admin.go
```diff
@@ -147,9 +147,28 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
 	}
 }
 
+// StartingOffset is the offset at which a newly created subscription will start
+// receiving messages.
+type StartingOffset int
+
+const (
+	// End represents the current HEAD offset.
```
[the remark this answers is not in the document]
hannahrogers-google: Good point! Done.

pubsublite/admin.go
```diff
@@ -147,9 +147,28 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
 	}
 }
 
+// StartingOffset is the offset at which a newly created subscription will start
+// receiving messages.
+type StartingOffset int
```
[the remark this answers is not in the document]
hannahrogers-google: Yes, good point, changed to OffsetLocation instead.

pubsublite/admin.go
```diff
@@ -147,9 +147,28 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
...
 // CreateSubscription creates a new subscription from the given config. If the
 // subscription already exists an error will be returned.
 func (ac *AdminClient) CreateSubscription(ctx context.Context, config SubscriptionConfig) (*SubscriptionConfig, error) {
+	return ac.CreateSubscriptionAtOffset(ctx, config, End)
+}
+
+// CreateSubscriptionAtOffset creates a new subscription from the given config
+// at the provided starting offset. If the subscription already exists an error
+// will be returned.
+func (ac *AdminClient) CreateSubscriptionAtOffset(ctx context.Context, config SubscriptionConfig, startingOffset StartingOffset) (*SubscriptionConfig, error) {
```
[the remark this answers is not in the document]
hannahrogers-google: Yes very true.  I like the idea of having a CreateSubscriptionOptions struct.  This way, we can have a generic OffsetLocation enum, while still using 'StartingOffset' in the options struct, which will be the same terminology used in the UI and in gCloud.  WDYT about the changes?

avawes (changes requested):
Hacked

pubsublite/admin.go
```diff
@@ -147,9 +147,41 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
...
+	// skips the entire message backlog.
+	End
+
+	// Beginning represents the offset of the oldest retained message.
+	Beginning
+)
+
+// CreateSubscriptionOpts specify the options to use when creating a
+// subscription.
+type CreateSubscriptionOpts struct {
```
tmdiep, line 169: Nit: Perhaps CreateSubscriptionOptions? Just to match CreateSubscriptionWithOptions.

pubsublite/admin.go
```diff
@@ -147,9 +147,41 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
...
 
+// OffsetLocation refers to the location of an offset with respect to the
+// message backlog.
+type OffsetLocation int
+
+const (
+	// UnspecifiedOffsetLocation represents an OffsetLocation that has not
+	// been set.
+	UnspecifiedOffsetLocation OffsetLocation = iota
+
```
tmdiep, line 158: Let's omit this for now to disallow its usage. We can add it later if there's a need for an "unspecified" flag somewhere. End can be initialized with `iota + 1`.

tmdiep (approved):

pubsublite/admin.go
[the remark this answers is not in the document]
tmdiep: Looks great. Please wait for an LGTM also from @tbpg or @codyoss, since this is an API surface change.

pubsublite/admin.go
hannahrogers-google: Done.

pubsublite/admin.go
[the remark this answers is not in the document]
hannahrogers-google: After further discussion offline, we decided instead to go with option 2 (variadic 'opts').  This gives us more flexibility in the future, but is more straightforward at the moment (as we only have 1 additional CreateSubscriptionOption)

pubsublite/admin.go
hannahrogers-google: Done.  Now using a CreateSubscriptionOption.

tmdiep:
Thanks Hannah! Still LGTM.

hannahrogers-google:
Friendly ping @tbpg or @codyoss, can you PTAL at the API surface changes?

pubsublite/admin.go
```diff
@@ -147,9 +159,36 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
...
+	return startingOffset{location}
+}
+
 // CreateSubscription creates a new subscription from the given config. If the
 // subscription already exists an error will be returned.
-func (ac *AdminClient) CreateSubscription(ctx context.Context, config SubscriptionConfig) (*SubscriptionConfig, error) {
+//
+// By default, a new subscription will only receive messages published after
+// the subscription was created. Use StartingOffset to override.
+func (ac *AdminClient) CreateSubscription(ctx context.Context, config SubscriptionConfig, opts ...CreateSubscriptionOption) (*SubscriptionConfig, error) {
```
tbpg, line 184: This is, unfortunately, a breaking change. Users may have defined an interface with the `CreateSubscription(context.Context, pubsublite.SubscriptionConfig)` method. If we change this signature, `AdminClient` will no longer fulfill that interface.

Is that OK?

Alternatively, could we add it to the `SubscriptionConfig`?

pubsublite/admin.go
```diff
@@ -29,6 +29,18 @@ var (
 	errNoSubscriptionFieldsUpdated = errors.New("pubsublite: no fields updated for subscription")
 )
 
+// BacklogLocation refers to a location with respect to the message backlog.
```
tbpg, line 32: This is used as a bool (`!= Beginning`). Should we expose it as a bool? Or will there possibly be other locations in the future?

tbpg (commented):
Sorry for the slow review! I'm not familiar with the feature in and of itself, but left a few comments.

pubsublite/admin.go
```diff
@@ -29,6 +29,18 @@ var (
...
+type BacklogLocation int
+
+const (
+	// End refers to the location past all currently published messages. End
+	// skips the entire message backlog.
+	End BacklogLocation = iota + 1
```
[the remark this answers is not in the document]
hannahrogers-google: Yes, this should be clear to users based on context.  When calling the create subscription method, users will have to specify StartingOffset(Beginning/End) - The StartingOffset should indicate that the option between beginning/end is a location in backlog.

pubsublite/admin.go
hannahrogers-google: This should be fine, I don't believe this library has been released yet.  Right @tmdiep?

pubsublite/admin.go
hannahrogers-google: We want to expose this feature to users as an option between Beginning/End for consistency with the admin 'seek' operation that will soon be available.  For the seek operation, users will also have an option between BacklogLocation beginning/end.

pubsublite/admin.go
codyoss, line 38: Should there be an undefined BacklogLocation. If there was one you could default to a specified value if not set a by a user.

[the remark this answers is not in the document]
tmdiep: If this is any clearer, perhaps we can consider StartingAt(BeginningOfBacklog|EndOfBacklog).

pubsublite/admin.go
tmdiep: The library is released as Beta, but since that was only recent, the likelihood of a user creating an interface for AdminClient is very low. This low risk is fine to me.

pubsublite/admin.go
tmdiep: The zero value was reserved for a future undefined/unspecified value, if required. For the create subscription option, we want users to specify either Beginning or End.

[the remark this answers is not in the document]
tbpg: I'm good with any of the above. One other idea would be basing it on "oldest" and "newest" (is the end new or old?)

pubsublite/admin.go
tbpg: For my own understanding, why don't we want to put it into the config?

(The two common patterns for functions like this are config arguments and options -- I'm surprised to see both.)

pubsublite/admin.go
```diff
@@ -147,9 +159,36 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
...
-func (ac *AdminClient) CreateSubscription(ctx context.Context, config SubscriptionConfig) (*SubscriptionConfig, error) {
+//
+// By default, a new subscription will only receive messages published after
+// the subscription was created. Use StartingOffset to override.
+func (ac *AdminClient) CreateSubscription(ctx context.Context, config SubscriptionConfig, opts ...CreateSubscriptionOption) (*SubscriptionConfig, error) {
+	skipBacklog := true
+	for _, opt := range opts {
+		if _, ok := opt.(startingOffset); ok {
+			skipBacklog = opt.(startingOffset).backlogLocation != Beginning
+		}
```
tbpg, line 189: Can avoid the type assertion with Cody's idea above. But, for future type assertions, you can do this:

```suggestion
		if s, ok := opt.(startingOffset); ok {
			skipBacklog = s.backlogLocation != Beginning
		}
```

pubsublite/admin.go
```diff
@@ -147,9 +159,36 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
 	}
 }
 
+// CreateSubscriptionOption is an option for AdminClient.CreateSubscription.
+type CreateSubscriptionOption interface {
```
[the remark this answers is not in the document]
hannahrogers-google: Good point - thanks! Done.

pubsublite/admin.go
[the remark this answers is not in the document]
hannahrogers-google: End represents the newest message, Beginning represents the oldest retained.  My preference would be to keep this as End/Beginning, as this is what we use for gCloud, the UI, and other client libraries.

pubsublite/admin.go
hannahrogers-google: All of the fields in the SubscriptionConfig map directly to a field in the Subscription resource proto - these fields are defining features of the subscription itself.  On the other hand, the StartingOffset option only makes sense on subscription creation - it is not part of the subscription resource and does not map to any field in the subscription proto (it instead maps to a field in the create subscription request).

The SubscriptionConfig is used elsewhere in the library (ex. it is returned from the UpdateSubscription request).  For this reason, I think that it is misleading to have the StartingOffset in the subscription config when it only makes sense on subscription creation.

tbpg (approved):

PR #3799, opened 2021-03-12
test(pubsublite): fix flaky TestAssignerHandlePartitionFailure

tmdiep:
The fake error response needs to be set before the assigner is started to prevent races.

Fixes https://github.com/googleapis/google-cloud-go/issues/3789.


codyoss (approved):
LGTM

PR #3803, opened 2021-03-12
docs(spanner): fix the result handling after BufferWrite

### LINKED ISSUE #3802: spanner: (Doc) Missing error handling after BufferWrite
The below example code in the doc doesn't catch and handle the error from `txn.BufferWrite` . I guess it disables the retry to be triggered by this error.

https://github.com/googleapis/google-cloud-go/blob/f321666c750649d5d177e44c6c73033f54ae5d23/spanner/doc.go#L303

Just make it catch and return the error.

sryoya:
Fixes #3802


spanner/doc.go
```diff
@@ -300,12 +300,10 @@ of the transaction:
         }
         balance -= 10
         m := spanner.Update("Accounts", []string{"user", "balance"}, []interface{}{"alice", balance})
-        txn.BufferWrite([]*spanner.Mutation{m})
-
+        return txn.BufferWrite([]*spanner.Mutation{m})
```
olavloite, line 303: nit: could we move this line to after the comment. It does not feel right to have a comment after the return statement of a function.

olavloite (commented):
Good catch.

`Transaction.BufferWrite` will however normally not throw any errors (unless you have manually closed the transaction) as it only buffers the mutation in an in-memory data structure. So this would not affect transaction retries if you were to implement your own transaction functions using this sample. Any `Aborted` error would be returned by the `Commit` that is executed after the transaction function has returned.

spanner/doc.go
sryoya: Yup, I actually I wondered if it's right. 
Then, fixed this with the similar code I referred to.

https://github.com/googleapis/google-cloud-go/pull/3803/commits/64546a29dd025fb6a75b2991c7cc38151e38bfd2

olavloite (approved):

olavloite:
> Btw, how can we do this? I'm just curisous.

You cannot do it with the normal transaction runner, but you could do it if you use [statement-based transactions](https://github.com/googleapis/google-cloud-go/blob/4a0a22e44ac8e852e640d07fcc825bc589b906c8/spanner/transaction.go#L1103), by calling `Rollback` on the returned transaction. Using statement-based transactions is however only recommended for advanced usage (like writing a SQL driver), as it requires you to handle all retries of aborted transactions and other possible failures manually.

> But, One thing I've experienced is that since I added handling for error from the BufferWrite in my company's application , the failure rate of writes to Spanner has clearly decreased.

I'm pretty sure that that is either coincidence or caused by a different change that was applied at the same time, as the `BufferWrite` function will never return a retryable error. It would therefore never affect the number of read/write transactions that need to be retried.

olavloite:
@tbpg Would you be able to merge this one? The integration tests are not running as the PR comes from a fork. They are green however (and not really relevant, as it is only a documentation change).

tbpg (approved):

PR #3812, opened 2021-03-13
test(logging): add kokoro build script for environment tests

nicoleczhu:
This environment test script runs logging & error reporting tests from [this submodule](https://github.com/googleapis/env-tests-logging). It deploys various GCP services like Cloud Run, Functions, GKE, etc, write logs, and tests whether the log entries are as expected. 

Questions: 
1. I'm using `go_integration_service_account` from keystore. I may need to add IAM permissions to deploy CloudRun and CloudFunctions (required by the tests). Is that ok or should I create a new SA for this?
2. I'm open to putting this script somewhere else (should I nest it under internal/logging or somewhere else?) 
Please don't hestitate to let me know if this can be more cleanly introduced into this repo. Thanks for any pointers.


internal/kokoro/environment.sh
```diff
@@ -0,0 +1,73 @@
...
+cd "env-tests-logging/"
+
+# Disable buffering, so that the logs stream through.
+export PYTHONUNBUFFERED=1
+
+# Debug: show build environment
+env | grep KOKORO
+
+# Set up service account credentials
+export GOOGLE_APPLICATION_CREDENTIALS=$KOKORO_KEYSTORE_DIR/72523_go_integration_service_account
```
codyoss, line 44: nit: Nothing that get logged here is secret today, but I would still just just suggest moving this to the top before set -x is enabled in case something more sensitive shows up in the future.

nicoleczhu: Sounds good. Tis moved. 

@codyoss Are you ok with my using the following?

1. I'm using go_integration_service_account from keystore. I may need to add IAM permissions to deploy CloudRun and CloudFunctions (required by the tests). Is that ok or should I create a new SA for this?

2. I'm putting this environment script under /kokoro directory. Let me know if you want me to nest it somewhere else!

codyoss: 1. Yes!
2. Yes!

codyoss (approved):

internal/kokoro/environment.sh
nicoleczhu: Thanks for the PR!

PR #3816, opened 2021-03-15
feat(internal/godocfx): handle Markdown content

tbpg:
The first commit copied the go/doc package into `third_party`.

Then, I added ToMarkdown to the doc package. I got the idea from @dmitshur,
who pointed me to http://golang.org/issues/34875.

Once the comment is converted to Markdown, it gets converted to HTML by
`goldmark`.

Future changes will:
* Add syntax highlighting for code blocks.
* Maybe add \` tags around code elements in the comment.
* Maybe support Markdown in other doc comments.


codyoss (commented):
LGTM

third_party/go/doc/comment.go
```diff
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// +build go1.16
```
codyoss, line 5: Maybe just change the version in `go.mod` instead?

codyoss (approved):

PR #3819, opened 2021-03-15
feat(internal/godocfx): add prettyprint class to code blocks

### LINKED ISSUE #3804: internal/godocfx: better styling of package doc
Package docs have a few issues:
- [x] No syntax highlighting for code blocks. We need `class="prettyprint"`.
- [x] Markdown style links don't work.
- [x] Markdown style lists don't work.

These are all issues with the underlying `doc.ToHTML` that we use. We can't switch to a Markdown parser directly for the package doc because there are the magic GoDoc headers that don't start with `#`.

https://github.com/googleapis/google-cloud-go/blob/f321666c750649d5d177e44c6c73033f54ae5d23/internal/godocfx/parse.go#L576-L580

tbpg:
Fixes #3804.

### DIFF
--- a/internal/godocfx/goldmark-codeblock/codeblock.go
+++ b/internal/godocfx/goldmark-codeblock/codeblock.go
@@ -1,0 +1,71 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


internal/godocfx/goldmark-codeblock/codeblock.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	for _, opt := range opts {
+		opt.SetHTMLOption(&r.Config)
+	}
+	return r
+}
+
+func (r *codeBlockHTMLRenderer) renderCodeBlock(w util.BufWriter, source []byte, n ast.Node, entering bool) (ast.WalkStatus, error) {
+	if entering {
+		_, _ = w.WriteString(`<pre><code class="prettyprint">`)
```
codyoss, line 44: Should the errors for this write method be returned?

codyoss (approved):
LGTM, just one question.

internal/godocfx/goldmark-codeblock/codeblock.go
tbpg: Done. Thanks!

PR #3820, opened 2021-03-15
feat(pubsub): add schema support

hongalex:
This PR adds schema support to the Pub/Sub library. This includes a new `SchemaClient` that is used to manage the schema admin operations (Create, Get, List, Delete). Updating schemas is currently not supported. In addition, schemas can be added to a topic upon topic creation, but cannot be edited later.

Other notes:
- Schemas are supported by the pstest fake, but the validation RPCs don't perform any check other than an empty string check.
- The `ValidateMessage` RPC (defined in the proto) is split into two RPCs (`ValidateMessageWithConfig`, which validates against an in-memory schema config, and `ValidateMessageWithPath`, which validates against an existing Schema resource in Pub/Sub


pubsub/pstest/fake.go
```diff
@@ -1180,3 +1184,105 @@ func WithErrorInjection(funcName string, code codes.Code, msg string) ServerReac
...
+		return ret.(*pb.ValidateSchemaResponse), err
+	}
+
+	if req.Schema.Definition == "" {
+		return nil, status.Error(codes.InvalidArgument, "schema definition cannot be empty")
+	}
+	return &pb.ValidateSchemaResponse{}, nil
+}
+
+func (s *GServer) ValidateMessage(_ context.Context, req *pb.ValidateMessageRequest) (*pb.ValidateMessageResponse, error) {
```
kamalaboulhosn, line 1271: Does this also have caveats about what validation it actually does?

pubsub/schema.go
```diff
@@ -0,0 +1,292 @@
...
+
+	"google.golang.org/api/option"
+
+	vkit "cloud.google.com/go/pubsub/apiv1"
+	pb "google.golang.org/genproto/googleapis/pubsub/v1"
+)
+
+// SchemaClient is a Pub/Sub schema client scoped to a single project.
+//
+// It is EXPERIMENTAL and a part of a closed alpha that may not be
```
kamalaboulhosn, line 29: We can remove the EXPERIMENTAL notices.

pubsub/schema.go
```diff
@@ -0,0 +1,292 @@
...
+	_, err := s.sc.ValidateMessage(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	return &ValidateMessageResult{}, nil
+}
+
+// ValidateMessage validates a message against an schema specified
+// by a schema path pointing to an existing schema.
+func (s *SchemaClient) ValidateMessageWithPath(ctx context.Context, msg []byte, encoding SchemaEncoding, schemaPath string) (*ValidateMessageResult, error) {
```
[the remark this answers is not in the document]
hongalex: Yeah I have mixed feelings about this. We might want to allow users to access schemas outside of their project. This is sort of true for Pub/Sub as well, so we provide additional methods like `TopicInProject` and `SubscriptionInProject`, but I don't particularly like this overloaded concept. I think if we clarify the format for schemaPath when we use this (in this method as well as in the get `Schema`, this should be ok as a path.

pubsub/pstest/fake.go
hongalex: Added caveats and the actual checks.

pubsub/schema.go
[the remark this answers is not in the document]
hongalex: @kamalaboulhosn proposed a solution which is to default to `schemaIDs` rather than path, but not provide a `SchemaInProject`. Instead, users can instantiate another `schemaClient` if they want to access schemas outside of their existing project. I like this too, so I'll implement this for now.

codyoss (approved):
LGTM for style, but I'll defer to @kamalaboulhosn for what this feature is trying to accomplish.

pubsub/topic.go
```diff
@@ -195,6 +196,14 @@ type TopicConfig struct {
 	// published to this topic, in the format
 	// "projects/P/locations/L/keyRings/R/cryptoKeys/K".
 	KMSKeyName string
+
+	// Schema defines the schema settings upon topic creation. This cannot
+	// be modified after a topic has been created.
+	//
+	// It is EXPERIMENTAL and a part of a closed alpha that may not be
```
kamalaboulhosn, line 203: We can remove the "EXPERIMENTAL" notices. Server support fully works.

kamalaboulhosn (approved):

PR #3831, opened 2021-03-17
chore(internal/apidiff): diff beta pkgs

noahdietz:
Expand `apidiff` presubmit check to `beta` client packages as well.


tbpg (commented):
Sent a question over chat.

PR #3840, opened 2021-03-19
chore(internal/apidiff): check for bang breaking change indicator

noahdietz:
Allow breaking changes when either of the conventional commit breaking change indicators are available: `!` or `BREAKING CHANGE:`.

An example of the former is: `fix!(dialogflow): remove rpc or fields that are unintended to release`

### DIFF
--- a/internal/apidiff/apidiff.go
+++ b/internal/apidiff/apidiff.go
@@ -51,8 +51,7 @@ func main() {
 	if err != nil {
 		log.Fatalln(err)
 	}
-	if strings.Contains(head, "BREAKING_CHANGE") {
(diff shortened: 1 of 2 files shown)


internal/apidiff/apidiff.go
```diff
@@ -215,6 +214,23 @@ func diff(m manifest, modDir, imp, pkg, base string) (string, error) {
 	return out, err
 }
 
+func checkAllowBreakingChange(commit string) bool {
```
tbpg, line 217: Annoying comment: mind adding a unit test?

noahdietz: Not annoying at all, done!

codyoss (approved):
LGTM

PR #3841, opened 2021-03-19
feat(spanner): add the Mutex lock in the beginning func of ReadWriteTransaction

sryoya:
I suppose there is no reason not to add the Mutex lock here even though the other functions mutating the same field with this func    having the Mutex lock.
If it's intentional, please close this PR.


olavloite:
It is intentional, in the sense that it is not necessary. The `begin` function is only called internally before the transaction is returned to the client application. That means that the client library knows that it will never be called simultaneously from different goroutines, and also never simultaneously with any other function that would alter the state of the transaction.

sryoya:
I see, good to know that! Thanks for the explanation.

PR #3843, opened 2021-03-20
fix(pubsublite): remove publish error translation

tmdiep:
The original intention of translating pubsublite internal errors to equivalent pubsub errors, was so that mostly the same code can be used to publish. I'd like to backtrack on this implementation decision - it wasn't explicitly discussed at any point.

The errors returned by pubsublite are wrapped to provide more useful context (e.g. the `ErrOversizedMessage` error includes the proto size of the user's message), so we should just propagate errors to users as-is.

The publisher clients of pubsublite and pubsub have sufficient differences in behavior, such that users will have to handle publish errors differently anyway. For example, pubsublite's PublisherClient can terminate permanently (while pubsub's Topic doesn't) and users have to create a new instance to republish failed messages.

pubsublite has a superset of pubsub errors. We will soon have an additional `ErrServiceUnavailable` error to indicate prolonged backend service unavailability.


tmdiep:
@hongalex, @manuelmenzella-google - interested in your thoughts about this. It wasn't explicitly discussed at any point, it was just an implementation decision that I made, which I no longer think is a good idea.

manuelmenzella-google:
I think this is a good change, and I like that we return more helpful messages. I do wonder, though: what do other languages (Java, Python) do here?

tmdiep:
@manuelmenzella-google Java and Python currently don't have a max publish buffer size (but will as a result of b/182864157). For oversized messages, the publisher clients will receive a fatal InvalidArgument error once the message is received by the server (which checks the accounted byte size and includes this in the error message).

manuelmenzella-google:
Sounds good, I like this change. Thanks.

manuelmenzella-google (approved):

PR #3846, opened 2021-03-22
fix(pubsublite): improve handling of backend unavailability

tmdiep:
- Define `ErrBackendUnavailable` to allow users to detect and handle backend unavailability during publish and subscribe.
- Increase settings to very high values:
  - `PublishSettings.Timeout` and `ReceiveSettings.Timeout` to 1 week.
  - `PublishSettings.BufferedByteLimit` to 10 GB.
- Add examples for handling publish and subscribe errors.


pubsublite/internal/wire/rpc.go
```diff
@@ -135,6 +135,9 @@ const pubsubLiteDefaultEndpoint = "-pubsublite.googleapis.com:443"
 func defaultClientOptions(region string) []option.ClientOption {
 	return []option.ClientOption{
 		internaloption.WithDefaultEndpoint(region + pubsubLiteDefaultEndpoint),
+		option.WithGRPCDialOption(grpc.WithKeepaliveParams(keepalive.ClientParameters{
```
manuelmenzella-google, line 138: Can you add a comment on why we need to override this?

pubsublite/internal/wire/streams.go
```diff
@@ -210,13 +220,10 @@ func (rs *retryableStream) connectStream() {
 		rs.mu.Lock()
 		defer rs.mu.Unlock()
 
-		if rs.status == streamReconnecting {
+		if rs.status == streamReconnecting || rs.status == streamTerminated {
```
manuelmenzella-google, line 223: Nit: the comment probably needs updating if you merge the two conditionals.

pubsublite/internal/wire/streams.go
```diff
@@ -88,10 +96,11 @@ type streamHandler interface {
...
 	// Immutable after creation.
-	ctx          context.Context
-	handler      streamHandler
-	responseType reflect.Type
-	timeout      time.Duration
+	ctx               context.Context
+	handler           streamHandler
+	responseType      reflect.Type
+	timeout           time.Duration
+	streamInitTimeout time.Duration
```
manuelmenzella-google, line 103: I'm really sorry to slow you down, but would you mind splitting out the changes for error handling from those for the stream initialization retry? I think the former are a must have, but I am slightly concerned about the latter (I'm sorry I didn't raise this before, we talked about this already).

I'm a little worried about the complexity, and the lack of consistency across clients. Can we discuss it during our dev sync today?

manuelmenzella-google (changes requested):

pubsublite/internal/wire/rpc.go
tmdiep: Done

pubsublite/internal/wire/streams.go
tmdiep: Done. I've removed stream init retries from this PR.

pubsublite/internal/wire/streams.go
tmdiep: I just reverted this to keep the conditionals separate.

pubsublite/internal/wire/settings.go
```diff
@@ -82,7 +82,7 @@ var DefaultPublishSettings = PublishSettings{
 	DelayThreshold: 10 * time.Millisecond,
 	CountThreshold: 100,
 	ByteThreshold:  1e6,
-	Timeout:        60 * time.Minute,
+	Timeout:        72 * time.Hour,
```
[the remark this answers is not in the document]
tmdiep: There isn't a defined infinite duration. It's an int64, so we can compute the max value, however I wanted to avoid depending on implementation details.

I changed this to 1 week, as it would be concerning if a client wasn't able to connect within that time. The original 3 days was to allow clients to cover a weekend.

manuelmenzella-google (approved):

PR #3848, opened 2021-03-22
chore(internal/gapicgen): fix commit message processing

codyoss:
This had no affect on the generation itself, but this sometimes
caused duplication in changelogs.


codyoss:
Here is an example of the duplication mentioned: https://github.com/googleapis/google-cloud-go/releases/tag/spanner%2Fv1.15.0

tbpg (approved):
Had to look up the signature of `CommitsSinceHash`. LGTM.

Issue #3634, opened 2021-01-29
dialogflow: support apiv3


ffeldhaus:
With the [GA release of Dialogflow CX on December 15th](https://cloud.google.com/dialogflow/docs/release-notes#December_15_2020), the API endpoint changed to v3. Currently the Client Libraries for Go only [support the v3beta1 endpoint](https://github.com/googleapis/google-cloud-go/tree/master/dialogflow/cx/apiv3beta1).

Please implement the v3 endpoint as documented here:
https://cloud.google.com/dialogflow/cx/docs/reference/rest/v3-overview

codyoss:
Thanks for your interest in the api. This will be generated in the near future. 😄

ffeldhaus:
@codyoss do you have an update on when the apiv3 will be available?

codyoss:
This should get released sometime later today.

PR #3850, opened 2021-03-23
feat(dialogflow/cx): start generating apiv3

codyoss:
Fixes: #3634


tbpg (changes requested):
Looks like Kokoro is failing, but it seems unrelated?

codyoss:
@tbpg Should be fixed now, I must have had a dirty workspace from some other testing I was doing.

internal/gapicgen/generator/config.go
```diff
@@ -570,6 +570,15 @@ var microgenGapicConfigs = []*microgenConfig{
 		apiServiceConfigPath:  "google/cloud/dialogflow/cx/v3beta1/dialogflow_v3beta1.yaml",
 		releaseLevel:          "beta",
 	},
+	{
+		inputDirectoryPath:    "google/cloud/dialogflow/cx/v3",
+		pkg:                   "cx",
```
tbpg, line 575: This is a not-great package name. But, too late for that. :)

tbpg (approved):

PR #3871, opened 2021-03-29
chore(internal/gapicgen): use API Service Config not gapic config

noahdietz:
This fixes some generation configs that I noticed were using the GAPIC yaml for the `apiServiceConfig` param. These are actually different yaml files and the generator silently accepted them (gonna fix this https://github.com/googleapis/gapic-generator-go/issues/573). The Go microgenerator doesn't use the GAPIC yamls at all.


codyoss (approved):
Thanks!

PR #3882, opened 2021-04-02
docs: update a few URLs in example documentation

veleek:
Updated a few URLs to point to current locations (as opposed to redirect URLs).
In the Service Account URLs case, it was pointing to a non-existant cloud console page.


veleek:
@googlebot I signed it!

codyoss (approved):

codyoss:
@veleek Thanks for the updated URLs, looks good to me.

PR #3886, opened 2021-04-03
fix(pubsub): respect subscription message ordering field in scheduler

### LINKED ISSUE #3884: pubsub: messages with ordering keys are processed in order even in subscriptions without ordering enabled
**Client**

Pub/Sub

**Expected behavior**

When a subscription has enable ordering disabled, messages published with ordering keys should not be processed in order.

**Actual behavior**

Messages are always processed in order by internal/scheduler/receive_scheduler.go based on the message's ordering key

**Additional context**

This will be fixed by a `GetSubscription` call at the start of `Receive` to check for the subscription's server-side "EnableMessageOrdering" field.

hongalex:
Fixes #3884


pubsub/subscription.go
```diff
@@ -901,9 +911,16 @@ func (s *Subscription) Receive(ctx context.Context, f func(context.Context, *Mes
 						old(ackID, ack, receiveTime)
 					}
 					wg.Add(1)
+					// Make sure the subscription has ordering enabled before adding to scheduler.
+					var key string
+					if s.enableOrdering {
+						key = msg.OrderingKey
+					} else {
+						key = ""
```
codyoss, line 919: nit: can be removed

codyoss (approved):
Just one small thing, LGTM

pubsub/integration_test.go
```diff
@@ -1344,6 +1355,76 @@ func TestIntegration_OrderedKeys_ResumePublish(t *testing.T) {
 	}
 }
 
+// TestIntegration_OrderedKeys_SubscriptionOrdering tests that messages
+// with ordering keys are not processed as such if the subscription
+// does not message ordering enabled.
```
indera-shsp, line 1360: nit: `does not have` ?

indera-shsp:
@hongalex 
Thank you for this fix.

pubsub/subscription.go
```diff
@@ -772,6 +774,14 @@ func (s *Subscription) Receive(ctx context.Context, f func(context.Context, *Mes
 	s.mu.Unlock()
 	defer func() { s.mu.Lock(); s.receiveActive = false; s.mu.Unlock() }()
 
+	// Check config to check EnableMessageOrdering field.
+	// See: https://github.com/googleapis/google-cloud-go/issues/3884
+	cfg, err := s.Config(ctx)
+	if err != nil {
+		return fmt.Errorf("sub.Config err: %v", err)
+	}
+	s.enableOrdering = cfg.EnableMessageOrdering
```
indera-shsp, line 783: Related doc

https://cloud.google.com/pubsub/docs/ordering

hongalex: This page does not explicitly call out why we're setting `EnableMessageOrdering`, aside from the code sample, so I'm inclined to leave this out.

indera-shsp: 👍  

My comment was more like - look there is an example where this property is mentioned.

This was a good read too:
https://medium.com/google-cloud/google-cloud-pub-sub-ordered-delivery-1e4181f60bc8

`These two properties allow publishers and subscribers to decide independently if messages are ordered. If the publisher does not specify ordering keys with messages or the subscriber does not enable ordered delivery, then message delivery is not in order and behaves just like Cloud Pub/Sub without the ordered delivery feature. Not all subscriptions on a topic need to have the same setting for enable_message_ordering. Therefore, different use cases that receive the same messages can determine if they need ordered delivery without impacting each other.
`

PR #3895, opened 2021-04-06
test(bigquery): add integration test for Column ACLs

shollyman:
Now that we have a v1 PolicyTagManager client, we can test the policy tag integration
with column ACLs.

This also plumbed in a new PTM client as part of the integration testing
setup, much like we make a cloud storage client available.

### DIFF
--- a/bigquery/integration_test.go
+++ b/bigquery/integration_test.go
@@ -899,7 +899,7 @@ func TestIntegration_Tables(t *testing.T) {
 	}
 }
 
-// helper function that establishes a basic policy tag setup.
+// setupPolicyTag is a helper for setting up policy tags in the datacatalog service.
 //
 // It returns a string for a policy tag identifier and a cleanup function, or an error.
 func setupPolicyTag(ctx context.Context) (string, func(), error) {
--- a/bigquery/integration_test.go
+++ b/bigquery/integration_test.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	"cloud.google.com/go/civil"
+	datacatalog "cloud.google.com/go/datacatalog/apiv1"
 	"cloud.google.com/go/httpreplay"
 	"cloud.google.com/go/iam"
 	"cloud.google.com/go/internal"
@@ -43,6 +44,7 @@ import (
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
+	datacatalogpb "google.golang.org/genproto/googleapis/cloud/datacatalog/v1"
 )
 
 const replayFilename = "bigquery.replay"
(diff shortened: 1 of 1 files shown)


bigquery/integration_test.go
```diff
@@ -889,6 +899,88 @@ func TestIntegration_Tables(t *testing.T) {
...
 }
 
+// setupPolicyTag is a helper for setting up policy tags in the datacatalog service.
+//
+// It returns a string for a policy tag identifier and a cleanup function, or an error.
+func setupPolicyTag(ctx context.Context) (string, func(), error) {
+	location := "us"
+	req := &datacatalogpb.CreateTaxonomyRequest{
+		Parent: fmt.Sprintf("projects/%s/locations/%s", testutil.ProjID(), location),
+		Taxonomy: &datacatalogpb.Taxonomy{
```
tswast, line 909: What's the resource identifier for a taxonomy? Will it assign a random ID? Do we have a way of cleaning these up if cleanupFunc fails to run?

bigquery/integration_test.go
```diff
@@ -889,6 +899,88 @@ func TestIntegration_Tables(t *testing.T) {
...
+	taxonomyID := resp.GetName()
+	cleanupFunc := func() {
+		policyTagManagerClient.DeleteTaxonomy(ctx, &datacatalogpb.DeleteTaxonomyRequest{
+			Name: taxonomyID,
+		})
+	}
+
+	tagReq := &datacatalogpb.CreatePolicyTagRequest{
+		Parent: resp.GetName(),
+		PolicyTag: &datacatalogpb.PolicyTag{
```
tswast, line 930: If we delete the taxonomy, will it delete policy tags? Or do we need to clean these up first? If so, same question re: identifiers.

(It's on my mind because I noticed we have a lot of stale resources in Python https://github.com/googleapis/python-bigquery/issues/590)

tswast (commented):
Looks good! I assume we aren't trying to test the actual Column ACLs query logic because that's not a client responsibility?

bigquery/integration_test.go
shollyman: Yes, explicitly documented as deleting all descendants.

bigquery/integration_test.go
shollyman: Yeah, identifier uses an integer value for the last token replacement in `projects/{project}/locations/{location}/taxonomies/{taxonomies}`.

Part of the reason I populate displayname and description is so these are easier to filter for if we need to do bulk cleanups.

shollyman:
> Looks good! I assume we aren't trying to test the actual Column ACLs query logic because that's not a client responsibility?

Yup.  We're not the right place for this level of e2e enforcement validation.

tswast (approved):

codyoss (approved):
LGTM

PR #3896, opened 2021-04-07
feat(spanner): support RPC priority for partition query

yfuruyama:
RPC priority was added by #3341, but it looks like the partition (batch) query was not supported in that PR.

This PR is for supporting RPC priority for the partition query.

This graph shows the CPU usage for low priority when I used this fix for running a partition query.
image

### DIFF
--- a/spanner/batch.go
+++ b/spanner/batch.go
@@ -183,6 +183,7 @@ func (t *BatchReadOnlyTransaction) partitionQuery(ctx context.Context, statement
 		Params:       params,
 		ParamTypes:   paramTypes,
 		QueryOptions: qOpts.Options,
+		RequestOptions: createRequestOptions(&qOpts),
 	}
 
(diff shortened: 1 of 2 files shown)


olavloite:
@yfuruyama Thanks for reporting this. That's indeed an omission. There's however a couple of other requests there as well that need to be updated, as this PR only covers queries and not read operations. I'll add an additional PR that covers that as well and reference this.

yfuruyama:
@olavloite Oops, sorry I should have asked you before creating this. Sure, I'll close this PR for now and look forward to seeing your PR :)

olavloite:
> @olavloite Oops, sorry I should have asked you before creating this. Sure, I'll close this PR for now and look forward to seeing your PR :)

Absolutely nothing to be sorry for :-) This PR made it clear right away what is missing.

PR #3899, opened 2021-04-07
feat(bigtable): Customer Managed Encryption (CMEK)

crwilcox:


bigtable/admin.go
```diff
@@ -119,6 +120,76 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
...
+	info.KMSKeyVersion = pbInfo.KmsKeyVersion
+
+	return info
+}
+
+type EncryptionType int32
+
+const (
+	// Encryption type was not specified, though data at rest remains encrypted.
+	ENCRYPTION_TYPE_UNSPECIFIED EncryptionType = 0
```
codyoss, line 145: nit: Go styling for const is the same as exported fields. i.e EncryptionTypeUnspecified.

bigtable/admin.go
```diff
@@ -119,6 +120,76 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
...
+	// TODO: backups also needs this.
+	// TODO: why not just return the clusterstates, as it has a map of encryption info.
+	encryptionInfo := map[string][]EncryptionInfo{}
+	for key, cs := range res.ClusterStates {
+		// TODO: if we didn't wrap EncryptionInfo, this could reduce to
+		// returning map[string][]*btapb.EncryptionInfo directly
+		// encryptionInfo[key] = cs.EncryptionInfo
+		encInfo := cs.GetEncryptionInfo()
+		if encInfo == nil {
+			return nil, nil
```
codyoss, line 178: Should this return an error or value?

bigtable/admin.go
```diff
@@ -119,6 +120,76 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
...
+	// CMEK-protected tables.
+	// CMEK-protected backups are pinned to the key version that was in use at
+	// the time the backup was taken. This key version is populated but its
+	// status is not tracked and is reported as `UNKNOWN`.
+	CUSTOMER_MANAGED_ENCRYPTION EncryptionType = 2
+)
+
+// Gets the current encryption info for the table across all of the clusters.
+// The returned map will be keyed by cluster id and contain a status for all of the keys in use.
+func (ac *AdminClient) EncryptionInfo(ctx context.Context, table string) (map[string][]EncryptionInfo, error) {
```
codyoss, line 162: I think this should return a struct that wraps a map. If the endpoint starts returning more info later we would want to introduce that as a non-breaking change.

bigtable/admin.go
```diff
@@ -960,13 +1043,26 @@ type ClusterConfig struct {
 	InstanceID, ClusterID, Zone string
 	NumNodes                    int32
 	StorageType                 StorageType
+
+	// Describes the Cloud KMS encryption key that will be used to protect the
```
codyoss, line 1047: nit: // KMSKeyName describes...

bigtable/admin.go
crwilcox: Looking now it probably should be deleted :)

bigtable/admin.go
```diff
@@ -119,6 +120,76 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
 	return fmt.Sprintf("projects/%s/instances/%s/clusters/%s/backups/%s", ac.project, ac.instance, cluster, backup)
 }
 
+// EncryptionInfo represents the encryption info of a table.
+type EncryptionInfo struct {
+	// TODO: compare to https://github.com/googleapis/java-bigtable/pull/656/files#diff-10655a12f2a8430533dca76435f6829a1a0bc978ada1295529b59a46afbcd95fR28
+	EncryptionStatus *status.Status // TODO: should this be wrapped as EncryptionType has been?
```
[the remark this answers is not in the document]
crwilcox: @kolea2 as it seems there may be contention cross-language on how much to wrap/not-wrap?

bigtable/integration_test.go
```diff
@@ -1450,6 +1477,186 @@ func TestIntegration_AdminCreateInstance(t *testing.T) {
...
+	}
+	defer adminClient.Close()
+
+	table := testEnv.Config().Table
+	clusterID := testEnv.Config().Cluster
+
+	// TODO: It may be good to automate, or allow this to be passed in ;)
+	project := "firestore-harvard-library"
+	loc := "us-central1" // NOTE: cannot use "global"
+	ring := "crwilcox-cmek-test-key2"
```
crwilcox, line 1522: note: today this is hardcoded to a key. I can make this a configuration variable at least and have CI use it. Would be ideal to automate creation, or use an available cmek key on the project.

bigtable/admin.go
```diff
@@ -119,6 +120,71 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
...
+	// CMEK-protected tables.
+	// CMEK-protected backups are pinned to the key version that was in use at
+	// the time the backup was taken. This key version is populated but its
+	// status is not tracked and is reported as `UNKNOWN`.
+	CustomerManagedEncryption EncryptionType = 2
+)
+
+// Gets the current encryption info for the table across all of the clusters.
+// The returned map will be keyed by cluster id and contain a status for all of the keys in use.
+func (ac *AdminClient) EncryptionInfo(ctx context.Context, table string) (map[string][]EncryptionInfo, error) {
```
tritone, line 162: The map return value could also be a custom type maybe?

bigtable/admin.go
```diff
@@ -247,12 +313,13 @@ type FamilyInfo struct {
...
 
-// TableInfo retrieves information about a table.
-func (ac *AdminClient) TableInfo(ctx context.Context, table string) (*TableInfo, error) {
+func (ac *AdminClient) getTable(ctx context.Context, table string) (*btapb.Table, error) {
 	ctx = mergeOutgoingMetadata(ctx, ac.md)
 	prefix := ac.instancePrefix()
 	req := &btapb.GetTableRequest{
 		Name: prefix + "/tables/" + table,
+		View: btapb.Table_FULL,
+		// View: btapb.Table_ENCRYPTION_VIEW,
```
tritone, line 322: What's the difference between these? Should you provide some kind of switch for callers?

bigtable/admin.go
```diff
@@ -960,13 +1038,27 @@ type ClusterConfig struct {
 	InstanceID, ClusterID, Zone string
 	NumNodes                    int32
 	StorageType                 StorageType
+
+	// KMSKeyName describes the Cloud KMS encryption key that will be used to
+	// protect the destination Bigtable cluster. The requirements for this key
+	// are:
+	//  1) The Cloud Bigtable service account associated with the project that
+	//  contains this cluster must be granted the
+	//  `cloudkms.cryptoKeyEncrypterDecrypter` role on the CMEK key.
```
tritone, line 1047: It seems strange to have full docs only for this field? Also, "CMEK key" seems redundant 😄

bigtable/admin.go
```diff
@@ -1607,13 +1711,14 @@ func (it *BackupIterator) Next() (*BackupInfo, error) {
...
-	ExpireTime  time.Time
-	State       string
+	Name           string
+	SourceTable    string
+	SizeBytes      int64
+	StartTime      time.Time
+	EndTime        time.Time
+	ExpireTime     time.Time
+	State          string
+	EncryptionInfo EncryptionInfo
```
tritone, line 1721: If this is nullable I would make the field a pointer so that nil can be distinguished from an empty struct. Then, you can make newEncryptionInfo also return a pointer (and make that pointer nil if the value passed in is nil), removing the need for the logic on L1680 and following.

bigtable/integration_test.go
tritone: Not sure how comparable this is to storage CMEK, but I think for that we require the person running the tests to create a keyring themself and set and env var to it.

bigtable/integration_test.go
```diff
@@ -1450,6 +1477,186 @@ func TestIntegration_AdminCreateInstance(t *testing.T) {
...
+	defer deleteTable(ctx, t, adminClient, table)
+	if err := adminClient.CreateTable(ctx, table); err != nil {
+		t.Fatalf("Creating table: %v", err)
+	}
+
+	encryptionKeyVersion := kmsKeyName + "/cryptoKeyVersions/1"
+
+	// The encryption info can take 30-300s (currently about 120-190s) to
+	// become ready.
+	for i := 0; i < 30; i++ {
```
tritone, line 1556: There are probably nicer ways to write this-- I can dig up some examples.

tritone (commented):
Some initial thoughts, happy to discuss further!

bigtable/integration_test.go
crwilcox: I think this is going to be the case here also. That said, if we already have that set for storage, I can likely use the same key.

bigtable/admin.go
```diff
@@ -119,6 +120,71 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
...
+	// TODO: compare to https://github.com/googleapis/java-bigtable/pull/656/files#diff-10655a12f2a8430533dca76435f6829a1a0bc978ada1295529b59a46afbcd95fR28
+	EncryptionStatus *status.Status // TODO: should this be wrapped as EncryptionType has been?
+	EncryptionType   EncryptionType
+	KMSKeyVersion    string
+}
+
+func newEncryptionInfo(pbInfo *btapb.EncryptionInfo) EncryptionInfo {
+	info := EncryptionInfo{}
+	info.EncryptionStatus = pbInfo.EncryptionStatus
+	// TODO: could also just return this as a string, but wrapped
```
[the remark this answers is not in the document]
crwilcox: So, as it is now, I took it verbatim to match the enum in the pb. Would it be better to still use iota, or to maintain consistency across the two?  (iota will be the same if the order is correct, perhaps this is more style?)

bigtable/admin.go
```diff
@@ -119,6 +120,71 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
...
+	info.EncryptionType = EncryptionType(pbInfo.EncryptionType.Number())
+	info.KMSKeyVersion = pbInfo.KmsKeyVersion
+
+	return info
+}
+
+type EncryptionType int32
+
+const (
+	// Encryption type was not specified, though data at rest remains encrypted.
```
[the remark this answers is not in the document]
crwilcox: @tritone I made this iota as you recommended.

Curious though, since these need to match the pb, would it be better to
```
EncryptionTypeUnspecified EncryptionType = EncryptionType(btapb.EncryptionInfo_ENCRYPTION_TYPE_UNSPECIFIED)
```

these should be functionally equivalent.

bigtable/integration_test.go
crwilcox: used key we have for other cmek stuffs.

CONTRIBUTING.md
```diff
@@ -182,6 +182,13 @@ $ gcloud kms keys create key2 --keyring $MY_KEYRING --location $MY_LOCATION --pu
 $ export GCLOUD_TESTS_GOLANG_KEYRING=projects/$GCLOUD_TESTS_GOLANG_PROJECT_ID/locations/$MY_LOCATION/keyRings/$MY_KEYRING
 # Authorizes Google Cloud Storage to encrypt and decrypt using key1.
 $ gsutil kms authorize -p $GCLOUD_TESTS_GOLANG_PROJECT_ID -k $GCLOUD_TESTS_GOLANG_KEYRING/cryptoKeys/key1
+# Authorizes Google Cloud Bigtable to encrypt and decrypt using key1
+gcloud kms keys add-iam-policy-binding key1 \
+    --keyring $MY_KEYRING \
+    --location $MY_LOCATION \
+    --role roles/cloudkms.cryptoKeyEncrypterDecrypter \
+    --member allAuthenticatedUsers \
```
kolea2, line 190: is this true? Wouldn't you want this to point to a service account?

bigtable/admin.go
```diff
@@ -119,6 +120,74 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
...
+	info.EncryptionStatus = pbInfo.EncryptionStatus
+	info.EncryptionType = EncryptionType(pbInfo.EncryptionType.Number())
+	info.KMSKeyVersion = pbInfo.KmsKeyVersion
+
+	return &info
+}
+
+// Status references google.golang.org/grpc/status.
+// It represents an RPC status code, message, and details of EncryptionInfo.
+// https://godoc.org/google.golang.org/grpc/internal/status
```
codyoss, line 144: nit: Make this a pkg.go.dev link

bigtable/admin.go
```diff
@@ -119,6 +120,74 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
...
+	// managed by the customer.
+	// The in-use version of the key and its status are populated for
+	// CMEK-protected tables.
+	// CMEK-protected backups are pinned to the key version that was in use at
+	// the time the backup was taken. This key version is populated but its
+	// status is not tracked and is reported as `UNKNOWN`.
+	CustomerManagedEncryption
+)
+
+type EncryptionInfoByCluster map[string][]EncryptionInfo
```
codyoss, line 166: nit: I think this should be map[string][]*EncryptionInfo

bigtable/admin.go
```diff
@@ -119,6 +120,74 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
...
+	EncryptionStatus *Status
+	EncryptionType   EncryptionType
+	KMSKeyVersion    string
+}
+
+func newEncryptionInfo(pbInfo *btapb.EncryptionInfo) *EncryptionInfo {
+	if pbInfo == nil {
+		return nil
+	}
+	info := EncryptionInfo{}
```
codyoss, line 134: nit: this could just be

```
return &EncryptionInfo{
  EncryptionStatus: pbInfo.EncryptionStatus,
  EncryptionType: EncryptionType(pbInfo.EncryptionType.Number()),
  KMSKeyVersion: pbInfo.KmsKeyVersion,
}
```

bigtable/admin.go
```diff
@@ -1058,7 +1184,9 @@ func (iac *InstanceAdminClient) Clusters(ctx context.Context, instanceID string)
 // GetCluster fetches a cluster in an instance
 func (iac *InstanceAdminClient) GetCluster(ctx context.Context, instanceID, clusterID string) (*ClusterInfo, error) {
 	ctx = mergeOutgoingMetadata(ctx, iac.md)
-	req := &btapb.GetClusterRequest{Name: "projects/" + iac.project + "/instances/" + instanceID + "/clusters/" + clusterID}
+	req := &btapb.GetClusterRequest{
+		Name: "projects/" + iac.project + "/instances/" + instanceID + "/clusters/" + clusterID,
```
codyoss, line 1188: nit: This might read a little better with `fmt.Sprintf`

bigtable/admin.go
```diff
@@ -1569,16 +1698,19 @@ func newBackupInfo(backup *btapb.Backup) (*BackupInfo, error) {
...
-	return &BackupInfo{
-		Name:        name,
-		SourceTable: tableID,
-		SizeBytes:   backup.SizeBytes,
-		StartTime:   startTime,
-		EndTime:     endTime,
-		ExpireTime:  expireTime,
-		State:       backup.State.String(),
-	}, nil
+	return &bi, nil
```
codyoss, line 1713: nit: I think returning bi directly is fine. I believe this is consistent with how the libraries are currently written.

CONTRIBUTING.md
crwilcox: I figured that, as this encryption key isn't really a secret, opening it up to all authn'd users was fine. 

@tritone do you know what `gsutil kms` is doing behind the scenes?

codyoss (commented):
I think there are still 1 or 2 open comments but this LGTM, I'll let someone with bigtable knowledge sign off though.

bigtable/integration_test.go
```diff
@@ -1450,6 +1459,176 @@ func TestIntegration_AdminCreateInstance(t *testing.T) {
...
+
+	keyRingName := os.Getenv("GCLOUD_TESTS_GOLANG_KEYRING")
+	if keyRingName == "" {
+		t.Fatal("GCLOUD_TESTS_GOLANG_KEYRING must be set. See CONTRIBUTING.md for details")
+	}
+	kmsKeyName := keyRingName + "/cryptoKeys/key1"
+
+	conf := &InstanceWithClustersConfig{
+		InstanceID: instanceToCreate,
+		// InstanceType: PRODUCTION,
```
kolea2, line 1510: nit - is this needed?

bigtable/integration_test.go
```diff
@@ -1450,6 +1459,176 @@ func TestIntegration_AdminCreateInstance(t *testing.T) {
...
+	clusterState := table2.ClusterStates[clusterID]
+	if got, want := len(clusterState.EncryptionInfo), 1; !cmp.Equal(got, want) {
+		t.Fatalf("Table Encryption Info Length: %v, want: %v", got, want)
+	}
+	tableEncInfo := clusterState.EncryptionInfo[0]
+	if got, want := int(tableEncInfo.EncryptionStatus.Code), 0; !cmp.Equal(got, want) {
+		t.Fatalf("EncryptionStatus: %v, want: %v", got, want)
+	}
+	// NOTE: this EncryptionType is btapb.EncryptionInfo_EncryptionType
+	if got, want := tableEncInfo.EncryptionType, btapb.EncryptionInfo_CUSTOMER_MANAGED_ENCRYPTION; !cmp.Equal(got, want) {
```
kolea2, line 1568: it would be nice to have an assertion for non cmek clusters on the table encryption info

kolea2 (approved):
LGTM with some nits (and my one comment about key creation is still outstanding)

CONTRIBUTING.md
tritone: Hmm not exactly sure what the question is here but here's the code: https://github.com/GoogleCloudPlatform/gsutil/blob/7845859b924004d16db0408335fd434424f6c8d7/gslib/commands/kms.py

Looks like it requests a service account and creates one if it doesn't exist.

bigtable/admin.go
```diff
@@ -119,6 +120,72 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
 	return fmt.Sprintf("projects/%s/instances/%s/clusters/%s/backups/%s", ac.project, ac.instance, cluster, backup)
 }
 
+// EncryptionInfo represents the encryption info of a table.
+type EncryptionInfo struct {
+	EncryptionStatus *Status
```
tritone, line 125: Minor nit: there is a lot of repeating of "Encryption" here, can you simplify the naming?

bigtable/admin.go
```diff
@@ -119,6 +120,72 @@ func (ac *AdminClient) backupPath(cluster, backup string) string {
...
+
+// EncryptionInfoByCluster is a map of cluster name to EncryptionInfo
+type EncryptionInfoByCluster map[string][]*EncryptionInfo
+
+// EncryptionInfo gets the current encryption info for the table across all of the clusters.
+// The returned map will be keyed by cluster id and contain a status for all of the keys in use.
+func (ac *AdminClient) EncryptionInfo(ctx context.Context, table string) (EncryptionInfoByCluster, error) {
+	ctx = mergeOutgoingMetadata(ctx, ac.md)
+
+	res, err := ac.getTable(ctx, table)
```
tritone, line 171: Are partial results possible? If so I guess it just returns an error and doesn't attempt to return what is present?

bigtable/admin.go
```diff
@@ -247,12 +314,12 @@ type FamilyInfo struct {
...
 }
 
-// TableInfo retrieves information about a table.
-func (ac *AdminClient) TableInfo(ctx context.Context, table string) (*TableInfo, error) {
+func (ac *AdminClient) getTable(ctx context.Context, table string) (*btapb.Table, error) {
 	ctx = mergeOutgoingMetadata(ctx, ac.md)
 	prefix := ac.instancePrefix()
 	req := &btapb.GetTableRequest{
 		Name: prefix + "/tables/" + table,
+		View: btapb.Table_FULL,
```
tritone, line 322: Is there a performance hit from getting all fields? I guess it probably doesn't matter for an admin feature.

tritone (commented):
Couple little questions, generally looking good!

bigtable/admin.go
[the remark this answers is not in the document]
tritone: I think what you have now is fine. Maybe add a comment that you expect these to match the proto values exactly.

bigtable/admin.go
crwilcox: I understand this point of view. I mirrored the proto messages, but could consider removing some of the repetition. 

https://github.com/googleapis/googleapis/blob/master/google/bigtable/admin/v2/table.proto#L219

bigtable/admin.go
crwilcox: It is possible that some cluster states could be returned while others return error. All will be returned with the 'error' cluster states being returned as well indicating their status. I say 'error' since most non-zero are simply stating an encryption being provisioned, or that it is revoked.

bigtable/admin.go
crwilcox: There may be a small hit here, but we don't expose the ability to change view levels at the moment. I could refactor this to use different view levels for different methods, but not sure if that is worth it. @igorbernstein2 any thoughts here on how expensive using Table_FULL is? Before this was using SCHEMA VIEW (default) but think I need more to get encryption information (may be able to go lower than FULL though)

crwilcox: altered to narrow scope. This likely wasn't too expensive, but I liked the idea and ran with it.

bigtable/admin.go
crwilcox: ended up removing repetition.

CONTRIBUTING.md
tritone: Could you re-use the service account used elsewhere in the setup? `GCLOUD_TESTS_GOLANG_KEY` should already point to a service account json; is there a gcloud command to grab the email from that? I see https://cloud.google.com/sdk/gcloud/reference/iam/service-accounts/list . I just don't know enough about IAM to be confident that adding this binding isn't some kind of security risk.

tritone (approved):
LGTM pending resolution for @kolea2 's question re: contributing setup.

codyoss (approved):

CONTRIBUTING.md
crwilcox: So, the complication here is we store the path to the key, but the email you need (the member id) for the service account is within the json.

```
  "client_email": "{PROJECT}@{PROJECT}.iam.gserviceaccount.com",
```

I could infer this as it generally follows that form, but it may not.  I was avoiding adding another EnvVar.

Issue #3486, opened 2020-12-23
logging: resource autodetection for CloudRun


nicoleczhu:
`Logs` generated in CloudRun containers should have the following fields by default: 

```json
resource: {
type: "cloud_run_revision"
labels: {
revision_name: ...
configuration_name: ...
service_name: ...
location: e.g. "us-west1-1"
project_id: ...
}
}
```

pontuspalmenas:
We experienced this issue a couple of weeks ago, and welcome the suggested solution in the PR. I have patched our source adding the `MonitoredResource` to the `Entry` with `detectGCRResource()` from the PR and verified that it works for us.

PR #3909, opened 2021-04-09
feat(logging): cloud run and functions resource autodetection

nicoleczhu:
Fixes: #3486 

Changes: 

- [x] cloud run has correct resource labels and type
- [x] functions has correct resource labels and type
- [x] e2e environment tests covered [here](https://github.com/googleapis/env-tests-logging/)


logging/logging.go
```diff
@@ -298,13 +298,75 @@ func detectGAEResource() *mrpb.MonitoredResource {
 	}
 }
 
+func isCloudRun() bool {
+	_, config := os.LookupEnv("K_CONFIGURATION")
```
simonz130, line 302: is this returning false when the variable is null or is empty?

daniel-sanche, line 302: Note that `K_CONFIGURATION` is part of the [knative](https://github.com/knative/serving/issues/2102) contract (which Cloud Run is based on), not just cloud run itself. This may misfire on other knative environments

This is something I'm not accounting for in Python currently.

logging/logging.go
```diff
@@ -298,13 +298,75 @@ func detectGAEResource() *mrpb.MonitoredResource {
...
+}
+
 func detectResource() *mrpb.MonitoredResource {
 	detectedResource.once.Do(func() {
 		switch {
-		// GAE needs to come first, as metadata.OnGCE() is actually true on GAE
-		// Second Gen runtimes.
+		// GAE, CloudRun need to come first, as metadata.OnGCE() returns true on these runtimes.
 		case os.Getenv("GAE_ENV") == "standard":
 			detectedResource.pb = detectGAEResource()
```
simonz130, line 365: what happens in AppEngine Flex?

simonz130 (commented):
This is awesome! Thank you!

logging/logging.go
```diff
@@ -298,13 +298,75 @@ func detectGAEResource() *mrpb.MonitoredResource {
...
+			"service_name":       os.Getenv("K_SERVICE"),
+			"revision_name":      os.Getenv("K_REVISION"),
+			"configuration_name": os.Getenv("K_CONFIGURATION"),
+		},
+	}
+}
+
+func isCloudFunction() bool {
+	_, name := os.LookupEnv("FUNCTION_NAME")
+	_, target := os.LookupEnv("FUNCTION_TARGET")
```
daniel-sanche, line 329: It might be worth adding a comment that these are for two different versions of Cloud Functions

logging/logging.go
nicoleczhu: Correct, it returns false when the envvar doesn't exist

logging/logging.go
```diff
@@ -298,13 +298,75 @@ func detectGAEResource() *mrpb.MonitoredResource {
...
+			"location":           regionFromZone(zone),
+			"service_name":       os.Getenv("K_SERVICE"),
+			"revision_name":      os.Getenv("K_REVISION"),
+			"configuration_name": os.Getenv("K_CONFIGURATION"),
+		},
+	}
+}
+
+func isCloudFunction() bool {
+	_, name := os.LookupEnv("FUNCTION_NAME")
```
daniel-sanche, line 328: These variables seem pretty generic. Users on other platforms might have their own `FUNCTION_NAME` set by coincidence, and have their code interpreted as running on Cloud Functions.

In Python, I do this:
```
_CLOUD_RUN_SERVICE_ID = "K_SERVICE"
_FUNCTION_TARGET = "FUNCTION_TARGET"
_FUNCTION_SIGNATURE = "FUNCTION_SIGNATURE_TYPE"
_FUNCTION_NAME = "FUNCTION_NAME"
_FUNCTION_REGION = "FUNCTION_REGION"
_FUNCTION_ENTRY = "ENTRY_POINT"
_FUNCTION_ENV_VARS = [_FUNCTION_TARGET, _FUNCTION_SIGNATURE, _CLOUD_RUN_SERVICE_ID]
_LEGACY_FUNCTION_ENV_VARS = [_FUNCTION_NAME, _FUNCTION_REGION, _FUNCTION_ENTRY]

if all([env in os.environ for env in _LEGACY_FUNCTION_ENV_VARS]) or all(
        [env in os.environ for env in _FUNCTION_ENV_VARS]):
   return _create_functions_resource()
```

logging/logging.go
```diff
@@ -298,13 +298,75 @@ func detectGAEResource() *mrpb.MonitoredResource {
...
+		},
+	}
+}
+
 func detectResource() *mrpb.MonitoredResource {
 	detectedResource.once.Do(func() {
 		switch {
-		// GAE needs to come first, as metadata.OnGCE() is actually true on GAE
-		// Second Gen runtimes.
+		// GAE, CloudRun need to come first, as metadata.OnGCE() returns true on these runtimes.
```
daniel-sanche, line 363: similarily, GCF needs to come before cloud run

logging/logging.go
```diff
@@ -334,6 +396,10 @@ func monitoredResource(parent string) *mrpb.MonitoredResource {
 	}
 }
 
+func regionFromZone(zone string) string {
+	return zone[:strings.LastIndex(zone, "-")]
```
daniel-sanche, line 400: Is there ever a case where this may not work as expected? (Maybe Anthos/GKE on-prem?) Will it handle the error properly?

logging/logging.go
nicoleczhu: This was already implemented in place before this PR. I believe it doesn't detect flex, and it incorrectly labels flex logs as GCE. =[

I plan to fix this in the future, once I have environment tests configured for app engine.

simonz130: Let's open a tracking bug to fix AppEngine Flex detection so we don't forget it

logging/logging.go
nicoleczhu: I think only GCF and GCR use region in the `location` label. The other services use zone. This is based on matching stdout logs ([see here](https://docs.google.com/document/d/1yL0YXM5grqmqMuab-lKHvzP_MvnFfmK8ys24GDFIWwQ/edit#heading=h.qmdgbrm3rdg3
)). pls cmiiw though!

logging/logging.go
nicoleczhu: oo yes , will update the comment. In the implementation, it does come before.

logging/logging.go
nicoleczhu: added comments for both

logging/logging.go
nicoleczhu: hmm good point, at the least I should look for other cloud run envvars, not just k_config

logging/logging.go
nicoleczhu: Great point. I'll follow suite

nicoleczhu: suit*

logging/logging.go
nicoleczhu: Added here: https://github.com/googleapis/google-cloud-go/issues/3919

nicoleczhu:
Ready for another look @simonz130 or @daniel-sanche  when you have time
@daniel-sanche I realize the library memoizes already. `detectResource` is run once upon logger initialization

codyoss (approved):
LGTM, thank you!

simonz130 (approved):

logging/logging.go
daniel-sanche: I think the other k_* variables will be the same. It might be a good idea to check the metadata server too?

But this probably doesn't have to be addressed right now

daniel-sanche (approved):

Issue #2810, opened 2020-09-02
metadata: add Region function for Cloud Run


icco:
**Is your feature request related to a problem? Please describe.**

https://cloud.google.com/run/docs/reference/container-contract#metadata-server says there is a region field, but there is no region function on the metadata  pacakge

**Describe the solution you'd like**
A clear and concise description of what you want to happen. What should the API be? How would you use it?

I'd like a region() function like zone()

**Describe alternatives you've considered**

Right now I'm just querying the metadata server by hand

**Additional context**

codyoss:
Hey, thanks for the request. As far as I can tell region is not a standardized endpoint across the compute platform. In my testing it appears to not be supported by compute VMs. Until it is more broadly supported I don't think that it should have a dedicated function. In the meantime you can always use [metadata.Get](https://pkg.go.dev/cloud.google.com/go/compute/metadata?tab=doc#Get) to retrieve the information. Although it sounds like you might already be doing this.

```go
region, err := metadata.Get("instance/region")
if err != nil {
	return err
} 
fmt.Println(region)
// Output: projects/PROJECT-NUMBER/regions/REGION
```
I do understand that having a convince wrapper, like Zone, would be nice here to trim up the string to the desired output.

icco:
I didn't know about `.Get`, that's a nice fix for now. Thank you!

codyoss:
Closing this issue as this is still not adopted by all compute platforms. Will reconsider adding this function in the future if region becomes standardized.

PR #3921, opened 2021-04-13
feat(compute/metadata): add instance/region

atotto:
add instance/region metadata for Google Cloud Run and Google Cloud Functions etc.

https://cloud.google.com/run/docs/reference/container-contract#metadata-server


codyoss:
Hey @atotto thanks for the PR. Can you please create an issue for discussion. This seems like a duplicate of #2810

atotto:
@codyoss Thank you for your reply. I didn't notice about that issue. 
Hmm, It seems that a conclusion has already been reached. I will close this PR.

PR #3926, opened 2021-04-14
fix(spanner): invalid numeric should throw an error

### LINKED ISSUE #3920: spanner: invalid numeric should throw an error
**Client**

Spanner

**Version**

spanner/v1.17.0

**Expected behavior**

Writing a numeric number "1.1234567891" should throw an error. 

**Actual behavior**

"1.123456789" is written successfully and the number is rounded.

hengfengli:
Fix #3920


olavloite (commented):
I'm not sure this is something that we can change now, at least not in this way. There might be users who are already using `NUMERIC` and that rely (implicitly or explicitly) on the client library to automatically round any `big.Rat` to a value that is acceptable for Spanner. Changing the behavior into returning an error in those cases is a breaking change, as it would suddenly cause those applications to receive an error from the client library where they previously got an OK.

One option could be to add this as a feature that users can use if they want to, for example by setting an option when creating the client, or maybe on a finer level (for example as an option on a statement?)

hengfengli:
@olavloite I saw you changes (https://github.com/googleapis/java-spanner/commit/9f2678568be77e82c14632b1c7ffcaafb71e7679) in Java which is also an improvement, right? How did you deal with the breaking change over there?

olavloite:
> @olavloite I saw you changes ([googleapis/java-spanner@9f26785](https://github.com/googleapis/java-spanner/commit/9f2678568be77e82c14632b1c7ffcaafb71e7679)) in Java which is also an improvement, right? How did you deal with the breaking change over there?

That was [before it was released](https://cloud.google.com/spanner/docs/release-notes#September_28_2020), so we didn't have to worry about it.

olavloite:
FYI: In the dotnet client the user is [given the option](https://github.com/googleapis/google-cloud-dotnet/blob/f4e1ad2ae0fe6611c020b28a61af6b22c15ca32e/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerNumeric.cs#L297) whether to truncate or to throw an error if a `decimal` is used that does not fit in a `SpannerNumeric`. I think that in this case that something similar would be the most logical option (although in this case it would probably be return error or round). To make it non-breaking, the default behavior would be to round a value that does not fit.

In a later major version bump it might also be a possibility to change the default behavior.

hengfengli:
If I go with adding a flag `DisableLossOfPrecisionHandling` into the ClientConfig, is there any way to pass it into `encodeValue` without changing the function definition? I'd like to make minimum changes. I don't see an easy way to do that. 

```go
	// DisableLossOfPrecisionHandling is the configuration about how to handle
	// values with significant digits that would be lost by converting from a
	// number to a numeric. By default, loss of precision handling will be
	// enabled, e.g., 1.1234567891 will be rounded to 1.123456789. If this is
	// set to true, then an error will be raised.
	DisableLossOfPrecisionHandling bool
```

olavloite:
> If I go with adding a flag `DisableLossOfPrecisionHandling` into the ClientConfig, is there any way to pass it into `encodeValue` without changing the function definition? I'd like to make minimum changes. I don't see an easy way to do that.
> 
> ```go
> 	// DisableLossOfPrecisionHandling is the configuration about how to handle
> 	// values with significant digits that would be lost by converting from a
> 	// number to a numeric. By default, loss of precision handling will be
> 	// enabled, e.g., 1.1234567891 will be rounded to 1.123456789. If this is
> 	// set to true, then an error will be raised.
> 	DisableLossOfPrecisionHandling bool
> ```

Hmmm.... Yup, that is going to be very difficult. `encodeValue` is just a package-level exported function. There is no `Encoder` struct or anything like that where you could set the config from a specific client. And even if you were to change the function definition it would still be very difficult to pass in the correct parameter if we were to add this option to the `ClientConfig`. `encodeValue` is called from multiple places that are not related to a specific client. See for example `Statement`. A `Statement` is not related to a specific client, but does call `encodeValue`.

As far as I can see now, the best option might be to add a global configuration option to `value.go` that is not client-dependent. That would mean that an application can only change the behavior for **all** clients that it creates, and not on a per-client basis.

Another option, albeit one that would require a lot more work, would be to add an additional type. It would be similar to `NullNumeric`. So you could add something like `SpannerNumeric` that contains a `big.Rat` and an option that determines what to do if precision would be lost. This would mean adding support for `SpannerNumeric` everywhere where currently `NullNumeric` is used, which is quite a lot of places.

My personal opinion is that the first option (adding a global configuration) would be sufficient for now, and that adding an extra type is something that we should only do if we know that there is a need for it.

Regarding the option: I would recommend that it is defined as an `enum` instead of a `boolean`. That way, it will be easier to add additional options for what to do when losing precision later without it having to be a breaking change. The current behavior is `Round` (which rounding mode?). The one that you want to add at the moment is `Fail`, but there might be a need in the future to add the option of other rounding modes and/or `Truncate`.

hengfengli:
@olavloite Thanks for the ideas. I had same thoughts with changing `encode_value` function and adding a global option, both of which I am not very satisfied with. 

I didn't think about adding a new type `SpannerNumeric` but I agree that this is not necessary at the moment. 

I will try to add a global configuration option.

hengfengli:
@olavloite Please take a look. I am still not sure if this is a good idea to add a global variable like this. It would be hard to remove in the future. I wonder if it is okay to not add this global variable, but let users to check the error by themselves, e.g., just make the `validateNumeric` function public for users. 

I was considering to add a field to ClientConfig, which updates this global variable to avoid users to directly modify this variable. However, I realize that it would be confusing if a user creates multiple clients, with or without this field passing in.

olavloite:
> @olavloite Please take a look. I am still not sure if this is a good idea to add a global variable like this. It would be hard to remove in the future. I wonder if it is okay to not add this global variable, but let users to check the error by themselves, e.g., just make the `validateNumeric` function public for users.

Yeah, this is not ideal, but I'm not sure that making the `validateNumeric` exported is a better solution than this global variable. Making `validateNumeric` exported would mean that we don't break anything for any users, and those who want to validate numeric input before sending it to Spanner can do so by calling this method. The disadvantage is that they would have to do it for all input, and missing one will cause it to be silently rounded.
Having the global variable has the disadvantage that it is not something that we can easily remove later, and a global variable for something like this also feels a little bit awkward. It would have been better if it would have been possible to configure it on the client, but I cannot see how that would be doable without a major rewrite of the entire type encoding and decoding. The advantage with a global variable is that users do not need to change anything to keep the current working, and they can easily switch it on in a test environment to check whether it is actually a problem for them or not at the moment.

Another possibility might be to combine the two: So both having the global variable __and__ make the `validateNumeric` exported. That would allow users to validate their input at the source, which would make it easier for them to turn on the option at a later moment if they do not want any silent rounding.

> I was considering to add a field to ClientConfig, which updates this global variable to avoid users to directly modify this variable. However, I realize that it would be confusing if a user creates multiple clients, with or without this field passing in.

Yes, I agree that it is a good idea not to include it in the `ClientConfig`. That would be confusing if you create multiple clients.

spanner/value.go
```diff
@@ -49,13 +50,55 @@ const (
 	NumericScaleDigits = 9
 )
 
+// NumericLossHandlingMode describes the mode of how to deal with loss of
+// precision on numeric values.
+type NumericLossHandlingMode int
```
olavloite, line 55: I think it would be good to use the same name as in .NET: `LossOfPrecisionHandling`

spanner/value.go
```diff
@@ -49,13 +50,55 @@ const (
...
+
+const (
+	// NumericRound rounds an invalid numeric value, e.g., 0.1234567895 rounds
+	// to 0.123456790.
```
olavloite, line 59:
```suggestion
	// NumericRound automatically rounds a numeric value that has a higher precision
	// than what is supported by Spanner, e.g., 0.1234567895 rounds to 0.123456790.
```

spanner/value.go
```diff
@@ -49,13 +50,55 @@ const (
...
+	NumericRound NumericLossHandlingMode = iota
+	// NumericError throws an error when meeting invalid numeric value.
```
olavloite, line 61:
```suggestion
	// NumericError returns an error for numeric values that have a higher precision
	// than what is supported by Spanner. E.g. the client returns an error if the
	// application tries to insert the value 0.1234567895.
```

spanner/value.go
```diff
@@ -49,13 +50,55 @@ const (
 	NumericScaleDigits = 9
 )
 
+// NumericLossHandlingMode describes the mode of how to deal with loss of
+// precision on numeric values.
```
olavloite, line 54:
```suggestion
// LossOfPrecisionHandling configures how to deal with loss of precision on numeric values.
// The value of this configuration is global and will be used for all Spanner clients.
```

spanner/value.go
```diff
@@ -49,13 +50,55 @@ const (
...
+	if r == nil {
+		return nil
+	}
+	// Add one more digit to the scale component to find out if there are more
+	// digits than required.
+	strRep := r.FloatString(NumericScaleDigits + 1)
+	strRep = strings.TrimRight(strRep, "0")
+	strRep = strings.TrimLeft(strRep, "-")
+	s := strings.Split(strRep, ".")
+	if len(s) != 2 {
```
olavloite, line 88: Will this ever be true for any `big.Rat` value? If so, what happens now if such a value is inserted into Spanner?

spanner/value.go
```diff
@@ -49,13 +50,55 @@ const (
...
+		return fmt.Errorf("invalid numeric float string: %s", strRep)
+	}
+	whole := s[0]
+	scale := s[1]
+	if len(scale) > NumericScaleDigits {
+		return fmt.Errorf("max scale for a numeric is %d. The requested numeric has more", NumericScaleDigits)
```
olavloite, line 94: nit:
```suggestion
		return fmt.Errorf("max scale for a numeric is %d. The requested numeric has %d", NumericScaleDigits, len(scale))
```

spanner/value.go
hengfengli: From the function definition, https://golang.org/pkg/math/big/#Rat.FloatString, it will always return a float string with the radix point. What do you mean by `such a value`? Do you mean an overflowed value, e.g., `0.1234567895`? Currently, we call `FloatString` so it will be rounded to `0.123456790` and then in spanner, it will be `0.123456790`.

olavloite: The if statement checks if the length of the slice is 2, which means it checks whether the string representation contains a point (`.`). So what I meant is:
- Is it at all possible that `strRep` will ever __not__ contain a `.`? If that is never possible, then this if statement is not necessary and can be removed.
- If it is possible, would it always be a problem if that value was inserted into Spanner? If for example the value is `100`, it should as far as I can tell be possible to insert the value into a `NUMERIC` column in Spanner. In that case the if check should probably check whether the number itself is too big to fit in a `NUMERIC` column, but it should not reject it as invalid.

spanner/value.go
hengfengli: This is not quite right. If the number has more than 10 digits after the radix point, `scale` will also be 10 as well because I limit it to 10 when I convert it to a float string: 

```go
strRep := r.FloatString(NumericScaleDigits + 1)
```

spanner/value.go
hengfengli: I changed to `LossOfPrecisionHandlingOption`, which is the type. The actual global variable is renamed to `LossOfPrecisionHandling`.

spanner/value.go
hengfengli: I check the length here to avoid the invalid index access (a runtime error) to the array later: 

```go
	whole := s[0]
	scale := s[1]
```

It's a just protection on our end. If the returned string is `100`, it means the `FloatString` does not work properly, which should not be ignored.

spanner/value.go
olavloite: Ah, good point.

spanner/value.go
```diff
@@ -49,13 +50,59 @@ const (
 	NumericScaleDigits = 9
 )
 
+// LossOfPrecisionHandlingOption describes the option to deal with loss of
+// precision on numeric values.
+type LossOfPrecisionHandlingOption int
+
+const (
+	// NumericRound automatically rounds a numeric value that has a higher
+	// than what is supported by Spanner, e.g., 0.1234567895 rounds to
```
olavloite, line 59: nit: the word `precision` is missing after `higher`.

spanner/value.go
olavloite: Ok, that's reasonable I guess. My preference would be to remove the `if` statement in this case, as the contract of that function is to never return a string without a point. But if you feel it is better to keep it there, then I'm fine with that as well.

olavloite (approved):
LGTM (with a small nit on the documentation for the rounding option)

spanner/value.go
hengfengli: You are right. We can actually remove the `if` because `FloatString` always returns a radix point as we specify to have a certain number of digits after the radix point and the missing numbers will be filled with zeros.

hengfengli:
@skuruppu Can I get an approval from you? Otherwise, I can't merge this.

skuruppu (approved):

PR #3930, opened 2021-04-14
fix(bigtable/bttest): make table gc release memory

philpearl:
**How does the emulator store rows?**
Each table in emulator contains a Btree which stores the rows:
```
type table struct {
	rows        *btree.BTree             // indexed by row key
        ....
}
```

Each object in this btree is of type row. A row has a families map. A family has a cells map keyed by column name

A sample row would be:
```
{
	families: {
		"col-family-1": {
			cells: {
				"col-1": [{
					"value": "val-1"
				}],
				"col-2": [{
					"value": "val-2"
				}],
				"col-3": [{
					"value": "val-3"
				}]
			},
		},
		"col-family-2": ....
	}
}
```
Many fields have not been shown in above sample for brevity. 

Now, when cells col-1 and col-2 are deleted, the emulator [just sets "col-1" and "col-2" to empty array](https://github.com/googleapis/google-cloud-go/blob/8c76415d27f6ab0a5a4698303bf05a26d469d183/bigtable/bttest/inmem.go#L1550) i.e. the row becomes:

```
{
	families: {
		"col-family-1": {
			cells: {
				"col-1": [],
				"col-2": [],
				"col-3": [{
					"value": "val-3"
				}]
			},
		},
		"col-family-2": ....
	}
}
```

In this PR, this behaviour is being changed to [remove the key](https://github.com/googleapis/google-cloud-go/pull/3930/files#diff-15ce35feffa4a739c0646793b7b7338622ae25fc3dba739acfcbde6293251b7bR1579-R1580) itself i.e. the row will become:

```
{
	families: {
		"col-family-1": {
			cells: {
				"col-3": [{
					"value": "val-3"
				}]
			},
		},
		"col-family-2": ....
	}
}
```
Similarly, when all the families are deleted, the empty row will be deleted as well.

Closes: https://github.com/googleapis/google-cloud-go/issues/6102

**Description by PR and [issue](https://github.com/googleapis/google-cloud-go/issues/6102) original author:** 
We had a problem where we were unable to run the BigTable emulator for extended periods in dev environments without giving it huge amounts of RAM, even if we had quite aggressive GC policies applied to the tables. This fixes that by making GC in the emulator actually delete empty rows and columns.


bigtable/bttest/inmem.go
```diff
@@ -1242,13 +1242,23 @@ func (t *table) gc() {
...
 		r.mu.Lock()
 		r.gc(rules)
+		if len(r.families) == 0 {
+			toDelete = append(toDelete, i)
+		}
 		r.mu.Unlock()
 		return true
 	})
+
+	for _, i := range toDelete {
```
tritone, line 1259: Since the emulator is supposed to be able to handle concurrent operations, it looks like this triggers a race condition (which was flagged by the CI):

```
==================
WARNING: DATA RACE
Write at 0x00c00004c090 by goroutine 18:
  github.com/google/btree.(*BTree).deleteItem()
      /go/pkg/mod/github.com/google/btree@v1.0.1/btree.go:731 +0x144
  github.com/google/btree.(*BTree).Delete()
      /go/pkg/mod/github.com/google/btree@v1.0.1/btree.go:712 +0x43b
  cloud.google.com/go/bigtable/bttest.(*table).gc()
      /tmpfs/src/google-cloud-go/bigtable/bttest/inmem.go:1260 +0x3f6
  cloud.google.com/go/bigtable/bttest.TestConcurrentMutationsReadModifyAndGC.func5()
      /tmpfs/src/google-cloud-go/bigtable/bttest/inmem_test.go:125 +0x6c

Previous read at 0x00c00004c090 by goroutine 40:
  github.com/google/btree.(*BTree).Get()
      /go/pkg/mod/github.com/google/btree@v1.0.1/btree.go:819 +0x144
  cloud.google.com/go/bigtable/bttest.(*table).mutableRow()
      /tmpfs/src/google-cloud-go/bigtable/bttest/inmem.go:1211 +0xfd
  cloud.google.com/go/bigtable/bttest.(*server).MutateRow()
      /tmpfs/src/google-cloud-go/bigtable/bttest/inmem.go:793 +0x1bd
  cloud.google.com/go/bigtable/bttest.TestConcurrentMutationsReadModifyAndGC.func3()
      /tmpfs/src/google-cloud-go/bigtable/bttest/inmem_test.go:109 +0x249

Goroutine 18 (running) created at:
  cloud.google.com/go/bigtable/bttest.TestConcurrentMutationsReadModifyAndGC()
      /tmpfs/src/google-cloud-go/bigtable/bttest/inmem_test.go:123 +0xc8a
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1193 +0x202

Goroutine 40 (running) created at:
  cloud.google.com/go/bigtable/bttest.TestConcurrentMutationsReadModifyAndGC()
      /tmpfs/src/google-cloud-go/bigtable/bttest/inmem_test.go:101 +0xbcc
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1193 +0x202
==================
```

@crwilcox @kolea2 @igorbernstein2  any thoughts on how this could be accomplished without creating a race condition? Or thoughts on other workarounds that would make sense here?

philpearl: Whoops! Need to hold the table lock I think. Sorry, forgot to run with -race. Will fix.

philpearl:
Ah, actually, scratch that. It's still not safe. I'm not convinced it was safe before as GC was changing the cell slices, but this has certainly made it more obviously unsafe.

The easy option would be to hold the write lock over the GC cycle, whether it changes or not. A harder option would be to do the whole thing in two phases

philpearl:
Ah, no, wrong again. I think this version is good. The row lock protects us.

philpearl:
Wrong a third time! We need to re-check the rows as we delete them with a lock held. It could be that cells have been added to the row in the meantime. Fix incoming

tritone (commented):
This looks fine to me now. @kolea2 @crwilcox could you take a look?

crwilcox:
This looks good to me also. I have asked some other folks that are more familiar with the emulator to take a look also, in case there is a hidden gotcha.

bigtable/bttest/inmem.go
```diff
@@ -1334,7 +1362,15 @@ func (r *row) gc(rules map[string]*btapb.GcRule) {
 			continue
 		}
 		for col, cs := range fam.cells {
-			r.families[fam.name].cells[col] = applyGC(cs, rule)
+			cs = applyGC(cs, rule)
+			if len(cs) == 0 {
+				delete(fam.cells, col)
```
igorbernstein2, line 1367: above there is a note about iterating & deleting being a possible issue for rows. Why is it ok for columns and families?

philpearl: It's OK to manipulate the row itself as we're holding a lock on the row (grabbed further up the call stack).

philpearl:
Any chance of getting this merged?

bigtable/bttest/inmem.go
jimfulton: Why not move the lock/unlock calls to this method? (I know it makes the diff bigger, but it makes the code easier to follow.)

jimfulton (commented):
LGTM, but needs at least a basic test.

philpearl:
@jimfulton do you have any suggestions about what the test would look like and what it would test? Presumably there are already tests that the GC operates correctly in terms of what's visible in the database. There are already tests that catch issues with --race (I know because I broke them!). Are you suggesting a test that counts memory allocations?

philpearl:
Well past the first anniversary of waiting for a review...

telpirion:
Hi @philpearl , it looks like @jimfulton gave you some feedback (specifically about tests). I'm new to this PR, so it will take me some time to get up to speed.

One question: does this PR have a matching [issue](https://github.com/googleapis/google-cloud-go/blob/main/CONTRIBUTING.md#contributing)? If not, would you please file one?

philpearl:
I've raised an issue. https://github.com/googleapis/google-cloud-go/issues/6102. I'm afraid I didn't understand what type of test jim fulton was expecting and he didn't answer my query.

philpearl:
Hi @telpirion - any further thoughts? Are you up-to-speed yet?

bhshkh:
Reviewing

bhshkh:
The changes LGTM

> I've raised an issue. #6102. I'm afraid I didn't understand what type of test jim fulton was expecting and he didn't answer my query.

You can add a test with simple gc rule such as maxNumVersions and verify that empty rows, cells, cell families did get deleted.

bhshkh:
Added the tests

kolea2 (approved):

Issue #3916, opened 2021-04-12
all: generate standalone examples


tbpg:
cc @codyoss @noahdietz

tbpg:
https://github.com/googleapis/google-cloud-go/pull/3922 and https://github.com/googleapis/google-cloud-go/pull/3924 were also for this, but didn't tag this issue.

tbpg:
Closing, as this is fixed. :smile:

PR #3931, opened 2021-04-14
fix(internal/gensnippets): run goimports

tbpg:
Updates #3916.

### DIFF
--- a/internal/gensnippets/main.go
+++ b/internal/gensnippets/main.go
@@ -29,7 +29,6 @@ import (
 	"io/fs"
 	"log"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


internal/gensnippets/main.go
```diff
@@ -77,6 +78,14 @@ func main() {
 			}
 		}
 	}
+
+	if len(dirs) > 0 {
+		cmd := exec.Command("goimports", "-w", ".")
```
codyoss, line 83: Do you think we should use `golang.org/x/sys/execabs` instead?

codyoss (approved):
LGTM, just one question.

internal/gensnippets/main.go
tbpg: Yes. Done. Thanks!

PR #3935, opened 2021-04-14
feat(bigquery): support job deletion

shollyman:
Supports job.delete call.  Leaving in draft as this isn't pending backend release.


codyoss (approved):
LGTM

PR #3938, opened 2021-04-14
chore(all): un-gitignore file not in repo

nicoleczhu:
I accidentally merged this file, but only needed it locally and it breaks our CI if merged in master :[


nicoleczhu:
@codyoss would you mind giving this a quick lgtm? This file wasn't supposed to be pushed.

codyoss (approved):

PR #3941, opened 2021-04-14
chore: release 0.82.0

nicoleczhu:
Release-As: 0.82.0


codyoss:
Let's just do .82 as I think there are feats. I will just edit the message and squash appropriately.

codyoss (approved):

PR #3949, opened 2021-04-15
feat(internal): auto-generate snippets

tbpg:
I created a new package to make it easier to run than calling the binary. I included a binary for local development.

I needed to install `build-base` in the Docker image to get `gcc`, which was needed to `Load` all packages.

I modified `gapics.go` to replace `genproto` in every `go.mod` because the changes need to be visible everywhere to `Load` every module.

I ran the generator locally and it seems to work. I only included one directory in this PR because there are a lot of files. I can upload the rest in this PR after initial review.

Updates #3916.

PR #3951, opened 2021-04-15
refactor(internal/gapicgen): move cmdWrapper to new package

codyoss:
Refactored so that the code could be used by the bot. This will
help have better debugging in logs.


codyoss:
Wait for https://github.com/googleapis/google-cloud-go/pull/3949, then will refactor those changes.

internal/gapicgen/cmd/genbot/generate.go
```diff
@@ -157,12 +157,9 @@ func hasChanges(dir string) (bool, error) {
 	inmem := &bytes.Buffer{}
 	w := io.MultiWriter(os.Stderr, inmem)
 
-	c := exec.Command("bash", "-c", "git status --short")
+	c := command.Create("bash", "-c", "git status --short")
```
tbpg, line 160: What do you think of naming the package `vexec` or `execv`? I like `execv` more because it doesn't sound like "very exec. much shell. :dog: "

I like that more than naming it `command` because then we can rename `Create` to `Command` and it becomes a drop-in replacement for `os/exec`, just like `execabs`.

internal/gapicgen/command/command.go
```diff
@@ -0,0 +1,42 @@
...
+func Create(name string, arg ...string) *CmdWrapper {
+	c := &CmdWrapper{exec.Command(name, arg...)}
+	c.Stdout = os.Stdout
+	c.Stderr = os.Stderr
+	c.Stdin = os.Stdin
+	return &CmdWrapper{exec.Command(name, arg...)}
+}
+
+func (c *CmdWrapper) Run() error {
+	log.Printf(">>>> %v <<<<", strings.Join(c.Args, " ")) // NOTE: we have some multi-line commands, make it clear where the command starts and ends
```
tbpg, line 40: I wonder if this should log `c.Dir`?

tbpg (changes requested):
Looks good. A couple comments (naming and log content).

internal/gapicgen/cmd/genbot/generate.go
codyoss: good idea, execv it is!

internal/gapicgen/command/command.go
codyoss: I like that idea!

tbpg (approved):

PR #3954, opened 2021-04-16
refactor(internal): move gensnippets into gapicgen

codyoss:
Updates #3916.


tbpg (approved):
Thanks! I'm a fan of fewer modules.

PR #3956, opened 2021-04-16
feat(bigquery): DRAFT for RAP

shollyman:
Contains:

* policy iterator, accessed via Table reference
* refactored internal iam client to support another resource type for IAM get/set/test methods
* new query stats related to RLS
* unit test for new iterator


codyoss:
Can this be closed or still needed?

shollyman:
I'll go ahead and close this.  It was pending more work from the service team. which has not yet materialized.

PR #3957, opened 2021-04-17
fix(storage): retry io.ErrUnexpectedEOF

tritone:
This error can be caused by network flakiness. We already retry
it in the apiary client logic for uploads, so we should retry
here as well.

Apiary logic is here: https://github.com/googleapis/google-api-go-client/blob/master/internal/gensupport/resumable.go#L240


tbpg (approved):

storage/go110.go
```diff
@@ -17,13 +17,17 @@
...
 import (
+	"io"
 	"net/url"
 	"strings"
 
 	"google.golang.org/api/googleapi"
 )
 
 func shouldRetry(err error) bool {
+	if err == io.ErrUnexpectedEOF {
```
frankyn, line 28: I'd like to get this repro'd using the emulator and moved our sync to dig in.

frankyn (changes requested):

frankyn (approved):

PR #3958, opened 2021-04-17
Update storage.go

I-Cat:
@storage-dpe Storage DPE
@yoshi-go-admins


tritone:
I don't think this is necessary, and we should not change the licenses on individual files separately in any case.

PR #3962, opened 2021-04-19
feat(internal): generate region tags for snippets

tbpg:
I found a couple "manual" clients that needed repo-metadata entries, so
I added them and included the updated JSON.

There are many packages where it doesn't make sense to generate snippets
with any sort of correct region tag. So, I just skipped them. We can
come back if needed.

The region tags end up slightly repetitive (e.g. 
`accessapproval_generated_accessapproval_apiv1_Client_UpdateAccessApprovalSettings`),
but I think they're more predictable as-is because the package path may not
always line up with the API shortname.

More in depth configuration of the region tag would probably make things
considerably more complex (need to associate every example with the underlying
RPC?)

cc @amanda-tarafa


codyoss (approved):
LGTM, thanks for filing in the extra configs!

PR #3965, opened 2021-04-20
feat(spanner/spannertest): support multiple aggregations

petercgrant:
Support multiple aggregations in one query.
Split from #3937

### DIFF
--- a/spanner/spannertest/db_query.go
+++ b/spanner/spannertest/db_query.go
@@ -519,7 +519,7 @@ func (d *database) evalSelect(sel spansql.Select, qc *queryContext) (si *selIter
 
 	// Handle aggregation.
 	// TODO: Support more than one aggregation function; does Spanner support that?
-	aggI := -1
+	var aggI []int
 	for i, e := range sel.List {
 		// Supported aggregate funcs have exactly one arg.
 		f, ok := e.(spansql.Func)
@@ -530,12 +530,9 @@ func (d *database) evalSelect(sel spansql.Select, qc *queryContext) (si *selIter
 		if !ok {
 			continue
 		}
-		if aggI > -1 {
-			return nil, fmt.Errorf("only one aggregate function is supported")
-		}
-		aggI = i
+		aggI = append(aggI, i)
 	}
-	if aggI > -1 {
+	if len(aggI) > 0 {
 		raw, err := toRawIter(ri)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


spanner/spannertest/db_query.go
```diff
@@ -519,7 +519,7 @@ func (d *database) evalSelect(sel spansql.Select, qc *queryContext) (si *selIter
 
 	// Handle aggregation.
 	// TODO: Support more than one aggregation function; does Spanner support that?
```
olavloite, line 521: nit: remove TODO

spanner/spannertest/db_query.go
```diff
@@ -609,27 +571,68 @@ func (d *database) evalSelect(sel spansql.Select, qc *queryContext) (si *selIter
...
+							return nil, err
+						}
+						values = append(values, x)
+					}
+				}
+				x, typ, err := fn.Eval(values, argType)
+				if err != nil {
+					return nil, err
+				}
+				aggType = typ
```
olavloite, line 610: This makes the type of all the aggregation expressions in the list equal to the type of the last expression. So the following statement will fail because it will try to interpret all the columns as INT64:

```sql
SELECT Cool, MIN(Name), MAX(Name), COUNT(*) FROM Staff GROUP BY Cool
```

spanner/spannertest/integration_test.go
```diff
@@ -1145,6 +1145,13 @@ func TestIntegration_ReadsAndQueries(t *testing.T) {
 				{"Daniel"},
 			},
 		},
+		{
```
olavloite, line 1148: This could use some additional test cases to cover:
- Combination of aggregated values of different types
- Combination of agg_func(*) and agg_func(col_name) combinations
- Several aggregation expressions in combination with a GROUP BY clause

olavloite (changes requested):

olavloite (approved):
Thanks, LGTM!

olavloite:
@hengfengli I've verified that the tests succeed. The reason that kokoro is not running them, is that the PR is coming from a fork. I don't have permission to merge this PR, so is that something that you could do?

hengfengli:
@olavloite Thanks for reviewing this PR. I'll do it. 

@petercgrant Can you please rebase this PR with the main branch? Thanks.

petercgrant:
@olavloite Thank you so much for all of the constructive feedback.

@hengfengli It is now rebased. Thanks for getting it merged.

hengfengli (approved):

hengfengli:
@petercgrant Thanks for contributing it.

PR #3966, opened 2021-04-20
feat(spannertest): support unnest and multi-join

petercgrant:
Adds support to spannertest for various complex joins:
- `UNNEST` in a `FROM` clause: `SELECT * FROM a, UNNEST(a.ArrayColumn)`.
- Multi-table cartesian product: `SELECT * FROM a, b`.
- Multiple `JOIN` clauses: `SELECT * FROM a JOIN b USING(x) JOIN c USING(x)`.

Split from #3937

### DIFF
--- a/spanner/spannertest/integration_test.go
+++ b/spanner/spannertest/integration_test.go
@@ -453,7 +453,8 @@ func TestIntegration_ReadsAndQueries(t *testing.T) {
 	allTables := []string{
 		"Staff",
 		"PlayerStats",
-		"JoinA", "JoinB", "JoinC", "JoinD", "JoinE", "JoinF",
+		"Episodes",
+		"JoinA", "JoinB", "JoinC", "JoinD", "JoinE", "JoinF", "JoinG",
 		"SomeStrings", "Updateable",
 	}
 	errc := make(chan error)
@@ -658,6 +659,7 @@ func TestIntegration_ReadsAndQueries(t *testing.T) {
 		`CREATE TABLE JoinD ( x INT64, z STRING(MAX) ) PRIMARY KEY (x, z)`,
 		`CREATE TABLE JoinE ( w INT64, x STRING(MAX) ) PRIMARY KEY (w, x)`,
 		`CREATE TABLE JoinF ( y INT64, z STRING(MAX) ) PRIMARY KEY (y, z)`,
+		`CREATE TABLE JoinG ( x INT64, a STRING(MAX), b STRING(MAX) ) PRIMARY KEY (x, a, b)`,
 		// Some other test tables.
 		`CREATE TABLE SomeStrings ( i INT64, str STRING(MAX) ) PRIMARY KEY (i)`,
 		`CREATE TABLE Updateable (
@@ -665,6 +667,11 @@ func TestIntegration_ReadsAndQueries(t *testing.T) {
 			first STRING(MAX),
 			last STRING(MAX),
 		) PRIMARY KEY (id)`,
+		`CREATE TABLE Episodes (
+			EpisodeID INT64,
+			StaffNames ARRAY<STRING(MAX)>,
+			StaffIDs ARRAY<INT64>,
+		) PRIMARY KEY (EpisodeID)`,
 	)
 	if err != nil {
 		t.Fatalf("Creating sample tables: %v", err)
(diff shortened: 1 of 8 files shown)


spanner/spannertest/integration_test.go
```diff
@@ -1145,6 +1160,31 @@ func TestIntegration_ReadsAndQueries(t *testing.T) {
...
+			},
+		},
+		{
+			`SELECT EpisodeID, offset, Cool FROM Episodes AS e, UNNEST(e.StaffIDs) as id WITH OFFSET as offset JOIN Staff AS s ON id = s.ID LEFT JOIN Staff as taller ON s.Height < taller.Height WHERE taller.ID IS NULL ORDER BY offset`,
+			nil,
+			[][]interface{}{
+				{int64(3), int64(0), true},
+				{int64(4), int64(3), true},
+			},
+		},
```
olavloite, line 1187: This (or somewhere else, but I think this is the best place) could use some more test cases, especially for the base cases of what this PR introduces. So:
1. Simple `UNNEST` clauses, with and without `WITH OFFSET` (and with and without alias for the offset).
2. Simple Cartesian products using both 2 and 3 tables, and using both all tables filled with data and scenarios where at least one table does not contain data.

spanner/spansql/parser.go
```diff
@@ -908,7 +908,7 @@ func (p *parser) eat(want ...string) bool {
 
 	for _, w := range want {
 		tok := p.next()
-		if tok.err != nil || tok.value != w {
+		if tok.err != nil || !strings.EqualFold(tok.value, w) {
```
olavloite, line 911: The reason that I noticed this one in the original PR is that everything in `spannertest` is case sensitive. Adding case insensitivity specifically only here will probably cause more confusion than be of any help.

spanner/spansql/parser.go
```diff
@@ -1832,6 +1832,23 @@ func (p *parser) parseSelect() (Select, *parseError) {
 		}
 
 		for {
+			if len(sel.From) > 0 {
+				var lhs SelectFrom = SelectFromList(sel.From)
+				if len(sel.From) == 1 {
+					lhs = sel.From[0]
+				}
+				sfj, err := p.parseJoin(lhs)
```
olavloite, line 1840: Does this work if the query contains from table after a join? So something like this:
```sql
SELECT *
FROM Table1
INNER JOIN Table2 ON Table1.Id=Table2.Id, SomeOtherTable
```

spanner/spansql/parser.go
```diff
@@ -2084,6 +2037,71 @@ func (p *parser) parseSelectFrom() (SelectFrom, *parseError) {
...
+			alias := ID("offset")
+			if p.eat("AS") { // TODO: The "AS" keyword is optional.
+				alias, err = p.parseAlias()
+				if err != nil {
+					return nil, err
+				}
+			}
+			sfu.OffsetAlias = alias
+		}
+		// TODO: hint, offset
```
olavloite, line 2072: nit: remove `offset` as that seems to be handled

spanner/spansql/parser.go
```diff
@@ -2084,6 +2037,71 @@ func (p *parser) parseSelectFrom() (SelectFrom, *parseError) {
...
+		return sfu, nil
+	}
+
+	// A join starts with a from_item, so that can't be detected in advance.
+	// TODO: Support subquery, field_path, array_path, WITH.
+	// TODO: Verify associativity of multile joins.
```
olavloite, line 2078: nit 
```suggestion
	// TODO: Verify associativity of multiple joins.
```

spanner/spansql/parser_test.go
```diff
@@ -208,6 +208,39 @@ func TestParseQuery(t *testing.T) {
...
+						"Alias",
+					}},
+					From: []SelectFrom{SelectFromTable{
+						Table: "Characters",
+						Alias: "c",
+					}},
+				},
+			},
+		},
+		// Joins with hints.
```
olavloite, line 225: The join(s) below do(es) not use hints.

psytale:
@petercgrant wondering if you are going to follow up on the comments or if we should try and push through another PR with multiple joins, I made a similar feature the same day as your PR but your solution feels a lot more comprehensive :) Just feeling blocked by not having it upstreamed.

Cheers

petercgrant:
@psytale It's unlikely I will be able to resume this work. Since the last commit, I've accepted a new position and no longer have any personal IP rights. I'm happy to answer any questions about what I did previously. Feel free to do with this as you will.

psytale:
> @psytale It's unlikely I will be able to resume this work. Since the last commit, I've accepted a new position and no longer have any personal IP rights. I'm happy to answer any questions about what I did previously. Feel free to do with this as you will.

Cheers, thanks for the swift reply, will see if I can pick it up!

PR #3967, opened 2021-04-20
feat(spannertest): support index reads

petercgrant:
Adds support in spannertest for secondary index reads.
Split from #3937.

### DIFF
--- a/spanner/spannertest/db.go
+++ b/spanner/spannertest/db.go
@@ -274,11 +274,11 @@ func (r *indexRow) lessThan(than *indexRow, inclusive bool) bool {
 	for j := range r.i.cols {
 		left, ok := r.item(j)
 		if !ok {
-			return !r.i.desc[j]
+			return !r.i.desc[j] || inclusive
 		}
 		right, ok := than.item(j)
 		if !ok {
-			return r.i.desc[j]
+			return r.i.desc[j] || inclusive
 		}
 		n := compareVals(left, right)
 		if n < 0 {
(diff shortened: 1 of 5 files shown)


spanner/spannertest/db.go
```diff
@@ -617,6 +814,79 @@ func (d *database) ReadAll(tbl spansql.ID, cols []spansql.ID, limit int64) (*raw
...
+		// Specific keys.
+		for _, key := range keys {
+			indexKey, err := i.key(key.Values)
+			if err != nil {
+				return err
+			}
+			// Not an error if the key does not exist.
+			row := i.rowForKey(indexKey)
+			if !add(row) {
+				return nil
```
olavloite, line 847: This implementation means that specific keys will take precedence over key ranges. That is not consistent with the specification of the Read method of Cloud Spanner. The simplest example of when this would go wrong is if you execute a read operation with a limit of 1 and you specify both a single key and a key range. The single key is the row that always will be returned (assuming it exists), even if the key range starts at a lower value than the single key. This in other words does not respect the ordering of the entire result set when both single keys and key ranges have been specified.

spanner/spannertest/db.go
```diff
@@ -765,6 +1054,69 @@ func (t *table) insertRow(rowNum int, r row) {
...
+	for _, idx := range t.indexes {
+		idx.insertRow(r)
+	}
+}
+
+func (t *table) updateRow(rowNum int, r row, colIndexes []int) {
+	var updatedIndexes []int
+	for j, idx := range t.indexes {
+		for _, c := range colIndexes {
+			if idx.indexedCols[c] {
```
olavloite, line 1066: Shouldn't this also check if the column is in the list of `STORING` columns?

spanner/spannertest/integration_test.go
```diff
@@ -439,6 +441,358 @@ func TestIntegration_SpannerBasics(t *testing.T) {
...
+				return nil
+			}); err != nil {
+				t.Fatalf("Getting ages using index: %v", err)
+			}
+			if !reflect.DeepEqual(ages, tc.want) {
+				t.Fatalf("Read index %s gave wrong ages: got %+v, want %+v", tc.indexName, ages, tc.want)
+			}
+		}
+	})
+	t.Run("Read all from index desc", func(t *testing.T) {
```
olavloite, line 480: Isn't this just a repetition of the test case that is defined on line 454?

spanner/spannertest/integration_test.go
```diff
@@ -439,6 +441,358 @@ func TestIntegration_SpannerBasics(t *testing.T) {
...
+			return nil
+		}); err != nil {
+			t.Fatalf("Getting ages using index: %v", err)
+		}
+		wantAges = []int64{102, 49, 35, 0, 0}
+		if !reflect.DeepEqual(ages, wantAges) {
+			t.Fatalf("Read index gave wrong ages: got %+v, want %+v", ages, wantAges)
+		}
+	})
+	t.Run("Read ranges from index", func(t *testing.T) {
```
olavloite, line 497: These test cases (or in a separate test function) should be extended with cases for:
* Reading single keys (one or more)
* The combination of reading single keys + ranges
* Limit in combination with the above.

spanner/spannertest/db.go
```diff
@@ -73,6 +75,20 @@ type colInfo struct {
 	Alias    spansql.PathExp // an alternate name for this column (result sets only)
 }
 
+type index struct {
```
dsymonds, line 78: noting that an important aspect of spannertest is that it is _not_ intended to be performant, the only real aspect of correctness with regard to indexes (aside from tracking which ones exist, etc.) is that partial indexes (e.g. NULL_FILTERED) should return a subset of values when the index is used for reads. Other than that, there doesn't need to be any substantial use of indexes, nor any efficiency-oriented data structures like the btree.

Perhaps most of this should be dropped, and only track the index construction, then modify the read code so it can simulate the row skipping when using a partial index?

dsymonds: I mean, spannertest is explicitly not for performance, so if there isn't anything like null filtering (where you get different response values if you use such an index for reading) then there isn't much point in actually maintaining an index.

codyoss:
This PR has had no activity from either end in a long time. Closing. Feel free to reopen if there is a plan to make further requested changes.

PR #3974, opened 2021-04-20
chore(internal): add LibraryType to repo-metadata

codyoss:
This new data will be parsed and processed from our metadata
pipeline.

### DIFF
--- a/internal/gapicgen/generator/gapics.go
+++ b/internal/gapicgen/generator/gapics.go
@@ -275,31 +275,14 @@ type manifestEntry struct {
 	LibraryType       LibraryType `json:"library_type"`
 }
 
-type LibraryType int
-
-func (l LibraryType) MarshalJSON() ([]byte, error) {
-	switch l {
-	case GapicAutoLibraryType:
-		return []byte(`"GAPIC_AUTO"`), nil
-	case GapicManualLibraryType:
-		return []byte(`"GAPIC_MANUAL"`), nil
-	case CoreLibraryType:
-		return []byte(`"CORE"`), nil
-	case AgentLibraryType:
-		return []byte(`"AGENT"`), nil
-	case OtherLibraryType:
-		return []byte(`"OTHER"`), nil
-	default:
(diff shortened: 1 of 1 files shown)


internal/gapicgen/generator/gapics.go
```diff
@@ -266,14 +266,42 @@ func (g *GapicGenerator) microgen(conf *microgenConfig) error {
...
+		return []byte(`"AGENT"`), nil
+	case OtherLibraryType:
+		return []byte(`"OTHER"`), nil
+	default:
+		return nil, fmt.Errorf("unable to marshal value: %v", l)
+	}
+}
+
+const (
+	GapicAutoLibraryType LibraryType = iota
```
tbpg, line 298: Could we just make these strings to avoid needing `MarshalJSON` at all?

internal/gapicgen/generator/gapics.go
```diff
@@ -266,14 +266,42 @@ func (g *GapicGenerator) microgen(conf *microgenConfig) error {
...
-	DistributionName  string `json:"distribution_name"`
-	Description       string `json:"description"`
-	Language          string `json:"language"`
-	ClientLibraryType string `json:"client_library_type"`
-	DocsURL           string `json:"docs_url"`
-	ReleaseLevel      string `json:"release_level"`
+	DistributionName  string      `json:"distribution_name"`
+	Description       string      `json:"description"`
+	Language          string      `json:"language"`
+	ClientLibraryType string      `json:"client_library_type"`
```
tbpg, line 272: What's the difference between `library_type` and `client_library_type`? Fine if there is a difference, just want to make sure we shouldn't use this field.

tbpg (changes requested):

internal/gapicgen/generator/gapics.go
codyoss: That is a great point. Done

tbpg (approved):

PR #3985, opened 2021-04-21
fix(pubsub): make config call permission error in Receive transparent

### LINKED ISSUE #3947: pubsub: simple subscribers require permission pubsub.subscriptions.get since v1.10.2
**Client**

PubSub

**Environment**

- Alpine Docker on GKE
- Service using a service account that has `roles/pubsub.subscriber` IAM membership for my subscription

**Go Environment**

```
go version go1.16.2
```

**Code**

```go
package main

import (
	"context"
	"log"
	"os/signal"

	"cloud.google.com/go/pubsub"
)

func main() {
	ctx := context.Background()
	client, err := pubsub.NewClient(ctx, "my-project")	
	if err != nil {
		log.Fatal("failed to create client:", err)
	}

	rxCtx, stop := signal.NotifyContext(ctx, os.Interrupt)
	defer stop()

	sub := client.Subscription("my-subscription")
	err = sub.Receive(rxCtx, func(_ context.Context, msg *pubsub.Message) {
		log.Println("received a message:", msg.ID)
		msg.Ack()
	})

	if err != nil {
		log.Fatal("failed to receive:", err)
	}
}
```

**Expected behavior**

I can receive messages

**Actual behavior**

`failed to receive: sub.Config err: rpc error: code = PermissionDenied desc = User not authorized to perform this action.`

**Additional context**

Started after upgrading to v1.10.2.

Looks to be introduced by the [Subscription.Config(context.Context) call](https://github.com/googleapis/google-cloud-go/pull/3886/files#diff-820c00bfb025d27cd84a23cb38671b023d41f92431bea486d7d66a3c6eb9ccebR779) in [fix](https://github.com/googleapis/google-cloud-go/pull/3886) for #3884.

[Subscription.Config()](https://github.com/googleapis/google-cloud-go/blob/d864346067f8bff9dd7c9b094628548f648cbee5/pubsub/subscription.go#L549) requires `pubsub.subscriptions.get` permission for

```go
// Config fetches the current configuration for the subscription.
func (s *Subscription) Config(ctx context.Context) (SubscriptionConfig, error) {
	pbSub, err := s.c.subc.GetSubscription(ctx, &pb.GetSubscriptionRequest{Subscription: s.name})
...
```

hongalex:
In #3886, a `Subscription.Config()` call was added to check for `EnableMessageOrdering` on the subscription before delivering messages with ordering keys in order. This negatively affected users who didn't have the roles/viewer or roles/pubsub.viewer roles on their service account. 

This PR incorporates two changes:
1. The call to `Config()` is only done (once) when we detect an ordering key in a received message.
2. If the config call fails, we will deliver messages with ordering keys in order by default.

Fixes #3947


codyoss (approved):
LGTM

PR #3905, opened 2021-04-08
feat(spanner): enable request options for batch read

olavloite:
Enables RequestOptions (priority and tagging) for batch read transactions. This problem was first reported in #3896.

cc @yfuruyama

Fixes #3904

PR #3987, opened 2021-04-22
Revert "feat(spanner): enable request options for batch read"

olavloite:
BREAKING CHANGE: Removes `BatchReadOnlyTransaction.PartitionReadUsingIndexWithOptions` and `BatchReadOnlyTransaction.PartitionReadWithOptions`

Reverts googleapis/google-cloud-go#3905


skuruppu (approved):

olavloite:
@skuruppu The build is failing because it (correctly) detects that this is a breaking change. We are removing two exported methods that were added by the original change. I've updated the commit message so the api checker should accept this.

PR #3995, opened 2021-04-22
refactor(internal/gapicgen): moved things around for future reuse

codyoss:
- move git operations to its own package
- fix compile error in gensnippets
- move ForEachMod into execv
- fix a couple of lint errors

### DIFF
--- a/internal/gapicgen/cmd/gensnippets/main.go
+++ b/internal/gapicgen/cmd/gensnippets/main.go
@@ -35,8 +35,8 @@ func main() {
 	if flag.NArg() > 0 {
 		rootDir = flag.Arg(0)
 	}
-
-	if err := gensnippets.Generate(rootDir, *outDir); err != nil {
+	// TODO(tbp): route proper api short names
+	if err := gensnippets.Generate(rootDir, *outDir, map[string]string{}); err != nil {
 		log.Fatal(err)
 	}
 }
--- a/internal/gapicgen/execv/command.go
+++ b/internal/gapicgen/execv/command.go
@@ -16,9 +16,11 @@
 package execv
 
 import (
+	"io/fs"
 	"log"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 )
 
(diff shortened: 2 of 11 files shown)


codyoss:
Trying to split up all these refactoring I am doing into smaller consumable PRs. Changes should be easier to review as individual commits. Lumped these refactorings together though.

internal/gapicgen/cmd/gensnippets/main.go
```diff
@@ -35,8 +35,8 @@ func main() {
 	if flag.NArg() > 0 {
 		rootDir = flag.Arg(0)
 	}
-
-	if err := gensnippets.Generate(rootDir, *outDir); err != nil {
+	// TODO(tbp): route proper api short names
+	if err := gensnippets.Generate(rootDir, *outDir, map[string]string{}); err != nil {
```
tbpg, line 39: Whoops. Thank you.

internal/gapicgen/git/github.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2021 Google LLC
```
tbpg, line 1: Pretty sure we keep this the same.

internal/gapicgen/execv/command.go
```diff
@@ -42,3 +44,22 @@ func (c *CmdWrapper) Run() error {
 	log.Printf("[%s] >>>> %v <<<<", c.Dir, strings.Join(c.Args, " ")) // NOTE: we have some multi-line commands, make it clear where the command starts and ends
 	return c.Cmd.Run()
 }
+
+// ForEachMod runs the given function with the directory of
+// every non-internal module.
+func ForEachMod(rootDir string, fn func(dir string) error) error {
```
tbpg, line 50: Does this need to be moved? Doesn't really fit with the `execv` package given it doesn't run commands. Doesn't appear to be used anywhere other than the file it was defined in?

tbpg (approved):

internal/gapicgen/execv/command.go
codyoss: It does not have to for this PR, but there is some logic in genbot that could be refactored to use it in a future PR around updating dependencies. It was not a super great fit I agree, but it is also always used in the context of running commands so is seemed like an okay fit as a helper method for commands? I can move it back and just have some copy/paste if you think that would be better.

internal/gapicgen/git/github.go
codyoss: Ack

internal/gapicgen/execv/command.go
tbpg: SGTM.

PR #4003, opened 2021-04-23
fix(pubsub): fix failing message storage policy tests

### LINKED ISSUE #3065: pubsub: check message storage policy reasons in integration test
**Client**

PubSub

**Environment**

macOS

**Go Environment**

go 1.15

**Additional context**

A recent move in our regression test runner project's organization is causing message storage policy (geofencing) related tests to fail. Refactor tests to make them not dependent on any initial region set.
Tracks #3033 and #3034, which I will be closing in the meantime.

hongalex:
Due to new internal org level policies, we don't have a consistent set of initial message storage policy regions. This PR refactors tests to make them not dependent on any initial region set.

In addition, cleans up the update labels test by removing the check for the unrelated message storage policy field.

Fixes #3065


codyoss (approved):
LGTM

PR #4004, opened 2021-04-23
chore(pubsub): fix small lingering TODO comments

hongalex:
Remove test endpoint now that the filtering feature is GA 

Not included:
1. Replace `strings.Replace` with `strings.ReplaceAll`. This should be added once we drop support for go 1.11


pubsub/integration_test.go
```diff
@@ -1217,8 +1217,7 @@ func TestIntegration_OrderedKeys_JSON(t *testing.T) {
 	scanner := bufio.NewScanner(inFile)
 	for scanner.Scan() {
 		line := scanner.Text()
-		// TODO: use strings.ReplaceAll once we only support 1.11+.
-		line = strings.Replace(line, "\"", "", -1)
+		line = strings.ReplaceAll(line, "\"", "")
```
codyoss, line 1220: This was adding in 1.12. Need to keep using Replace for now to stay compatible.

codyoss (changes requested):

pubsub/integration_test.go
hongalex: You're right, should've checked beforehand instead of relying on the comment. I'll change this PR to only remove the experimental notice.

codyoss (approved):
LGTM

PR #4014, opened 2021-04-27
feat(bigtable): allow restore backup to different instance (#3489)

kolea2:
* feat: restore table to a different instance

* add check that table restored

* review fixes

* change instance config zone


kolea2:
@liubonan fyi

bigtable/admin.go
```diff
@@ -1457,10 +1457,17 @@ func UpdateInstanceAndSyncClusters(ctx context.Context, iac *InstanceAdminClient
...
 // RestoreTable creates a table from a backup. The table will be created in the same cluster as the backup.
 func (ac *AdminClient) RestoreTable(ctx context.Context, table, cluster, backup string) error {
+	return ac.RestoreTableTo(ctx, ac.instance, table, cluster, backup)
+}
+
+// RestoreTableTo creates a new table by restoring from this completed backup.
+//
+// diffInstance is an instance in which the new table will be restored to.
+// Instance must be in the same project as the project containing backup.
+func (ac *AdminClient) RestoreTableTo(ctx context.Context, diffInstance, table, cluster, backup string) error {
```
tritone, line 1467: Can you rename params to something like `dst` for the destination? What do other languages use? Also I think the destination should probably be the last param rather than the first.

Docs can also be clarified/cleaned up a bit to highlight difference from `RestoreTable`.

bigtable/integration_test.go
```diff
@@ -2353,6 +2373,31 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("NewAdminClient: %v", err)
+	}
+	defer dAdminClient.Close()
+
+	defer deleteTable(ctx, t, dAdminClient, diffTable)
+	if err = adminClient.RestoreTableTo(ctx, diffInstance, diffTable, cluster, "mybackup"); err != nil {
+		t.Fatalf("RestoreTableTo: %v", err)
+	}
+	_, err = dAdminClient.TableInfo(ctx, diffTable)
```
tritone, line 2397: Confirm that returned table info is correct as well?

bigtable/integration_test.go
```diff
@@ -2353,6 +2373,31 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+	}
+	dAdminClient, err := env.NewAdminClient()
+	if err != nil {
+		t.Fatalf("NewAdminClient: %v", err)
+	}
+	defer dAdminClient.Close()
+
+	defer deleteTable(ctx, t, dAdminClient, diffTable)
+	if err = adminClient.RestoreTableTo(ctx, diffInstance, diffTable, cluster, "mybackup"); err != nil {
+		t.Fatalf("RestoreTableTo: %v", err)
```
tritone, line 2395: Use `Errorf`

bigtable/integration_test.go
```diff
@@ -2353,6 +2373,31 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+	}
+	_, err = dAdminClient.TableInfo(ctx, diffTable)
+	if err != nil {
+		t.Fatalf("Restored to different instance TableInfo: %v", err)
```
tritone, line 2399: Use `Errorf` and improve message.

bigtable/admin.go
```diff
@@ -1457,10 +1457,17 @@ func UpdateInstanceAndSyncClusters(ctx context.Context, iac *InstanceAdminClient
...
 	ctx = mergeOutgoingMetadata(ctx, ac.md)
-	prefix := ac.instancePrefix()
+	prefix := "projects/" + ac.project + "/instances/" + diffInstance
```
tritone, line 1469: Could this go in a function or method as well?

kolea2: yes, this could definitely be refactored in a bunch of places in this file. Mind if I create a tracking issue to refactor after?

tritone: Sure, totally up to you.

bigtable/integration_test.go
```diff
@@ -2278,92 +2309,118 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+	if err = adminClient.CreateBackup(ctx, tblConf.TableID, sourceCluster, backupName, time.Now().Add(8*time.Hour)); err != nil {
+		t.Errorf("Creating backup: %v", err)
 	}
 
 	// List backup
-	backups, err = list(cluster)
+	backups, err = list(sourceCluster)
 	if err != nil {
-		t.Fatalf("Listing backups: %v", err)
+		t.Errorf("Listing backups: %v", err)
```
tritone, line 2331: At least some of these should stay as `Fatalf`-- for example, if there is a non-nil error here, or if there aren't at least 1 backup returned, then continuing with the test will cause a panic. Only use `Errorf` if it's useful for the test to continue execution.

bigtable/admin.go
```diff
@@ -1457,14 +1457,21 @@ func UpdateInstanceAndSyncClusters(ctx context.Context, iac *InstanceAdminClient
 
 // RestoreTable creates a table from a backup. The table will be created in the same cluster as the backup.
 func (ac *AdminClient) RestoreTable(ctx context.Context, table, cluster, backup string) error {
-	ctx = mergeOutgoingMetadata(ctx, ac.md)
-	prefix := ac.instancePrefix()
-	backupPath := ac.backupPath(cluster, backup)
+	return ac.RestoreTableFrom(ctx, ac.instance, table, cluster, backup)
+}
 
+// RestoreTableFrom creates a new table to a different instance by restoring from this completed backup.
```
tritone, line 1463: I think the docs could still be a little clearer. What is "this" here?

My understanding is: the admin client is tied to a specific instance `foo`. In the latest version of this change, we take a table _from_ the source instance/cluster and restore it to a new table in `foo`. Is that correct?

bigtable/integration_test.go
```diff
@@ -2250,16 +2250,47 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
-		t.Fatalf("NewAdminClient: %v", err)
+		t.Errorf("NewAdminClient: %v", err)
 	}
 	defer adminClient.Close()
 
-	table := testEnv.Config().Table
-	cluster := testEnv.Config().Cluster
-
+	tblConf := TableConf{
+		TableID: "tableconf-backup",
```
tritone, line 2258: Should this still be taken from testEnv?

bigtable/integration_test.go
```diff
@@ -2237,7 +2237,7 @@ func TestIntegration_InstanceUpdate(t *testing.T) {
 func TestIntegration_AdminBackup(t *testing.T) {
 	testEnv, err := NewIntegrationEnv()
 	if err != nil {
-		t.Fatalf("IntegrationEnv: %v", err)
+		t.Errorf("IntegrationEnv: %v", err)
```
tbpg, line 2240: If `NewIntegrationEnv` fails, does it make sense to continue the test? `Fatalf` makes sense to me.

bigtable/integration_test.go
```diff
@@ -2250,16 +2250,47 @@ func TestIntegration_AdminBackup(t *testing.T) {
 
 	adminClient, err := testEnv.NewAdminClient()
 	if err != nil {
-		t.Fatalf("NewAdminClient: %v", err)
+		t.Errorf("NewAdminClient: %v", err)
```
tbpg, line 2253: Same comment. If there was an error creating `adminClient`, it doesn't make sense to continue the test as `adminClient.CreateTableFromConf` will always fail.

bigtable/integration_test.go
```diff
@@ -2250,16 +2250,47 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+	}
+	defer iAdminClient.Close()
+	diffInstance := testEnv.Config().Instance + "-diff"
+	diffCluster := sourceCluster + "-diff"
+	conf := &InstanceConf{
+		InstanceId:   diffInstance,
+		ClusterId:    diffCluster,
+		DisplayName:  "different test sourceInstance",
+		Zone:         instanceToCreateZone2,
+		InstanceType: DEVELOPMENT,
```
tbpg, line 2286: Outside this PR, but this should just be `development`.

https://golang.org/doc/effective_go#mixed-caps

bigtable/integration_test.go
```diff
@@ -2278,92 +2309,118 @@ func TestIntegration_AdminBackup(t *testing.T) {
 		return infos, err
 	}
 
-	if err := adminClient.CreateTable(ctx, table); err != nil {
-		t.Fatalf("Creating table: %v", err)
-	}
-
 	// Precondition: no backups
```
tbpg, line 2312: Precondition implies we should use `Fatal`.

tbpg:
https://github.com/golang/go/wiki/TestComments#keep-going is relevant to using `t.Fatalf` vs `t.Errorf`.

bigtable/admin.go
kolea2: yes sorry this was leftover docs that I have now fixed. You are correct

bigtable/integration_test.go
kolea2: Do you mean the table id? Just curious, what's the benefit there?

tritone: Yep-- it seems best to have these configs live in a single place for all tests so that they can be easily modified/swapped.

kolea2: got it thanks!

bigtable/admin.go
kolea2: Realized this already existed and refactored.

kolea2:
@tbpg @tritone I believe all comments have been addressed. Thanks for the review!

bigtable/admin.go
```diff
@@ -1584,14 +1584,21 @@ func UpdateInstanceAndSyncClusters(ctx context.Context, iac *InstanceAdminClient
...
-	prefix := ac.instancePrefix()
-	backupPath := ac.backupPath(cluster, backup)
+	return ac.RestoreTableFrom(ctx, ac.instance, cluster, table, backup)
+}
 
+// RestoreTableFrom creates a new table in the admin's instance by restoring from the specified backup in a different instance.
+// sourceInstance (ex. "my-instance") and sourceCluster (ex. "my-cluster") are the instance and cluster in which the new table will be restored from.
+// tableName (ex. "my-restored-table") will be the name of the newly created table
+// backupName (ex. "my-backup") is the name of the backup to restore
+func (ac *AdminClient) RestoreTableFrom(ctx context.Context, sourceInstance, sourceCluster, tableName, backupName string) error {
```
tritone, line 1594: Docs are much clearer, thanks!

Nit: could you switch the args to `table` and `backup` to match `RestoreTable` above? It looks weird that the names are different.

bigtable/admin.go
```diff
@@ -1584,14 +1584,21 @@ func UpdateInstanceAndSyncClusters(ctx context.Context, iac *InstanceAdminClient
 
 // RestoreTable creates a table from a backup. The table will be created in the same cluster as the backup.
 func (ac *AdminClient) RestoreTable(ctx context.Context, table, cluster, backup string) error {
-	ctx = mergeOutgoingMetadata(ctx, ac.md)
-	prefix := ac.instancePrefix()
-	backupPath := ac.backupPath(cluster, backup)
+	return ac.RestoreTableFrom(ctx, ac.instance, cluster, table, backup)
+}
 
+// RestoreTableFrom creates a new table in the admin's instance by restoring from the specified backup in a different instance.
```
tbpg, line 1590: Does it have to be a different instance? Could it be the same instance? Should people prefer this or `RestoreTable`? Should we update this comment to refer to `RestoreTable` or `RestoreTable`'s comment to refer to this?

bigtable/integration_test.go
```diff
@@ -2441,12 +2441,43 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
 	// in case the client fails
-	defer deleteTable(ctx, t, adminClient, table)
+	defer deleteTable(ctx, t, adminClient, tblConf.TableID)
+
+	sourceInstance := testEnv.Config().Instance
+	sourceCluster := testEnv.Config().Cluster
+
+	iAdminClient, err := testEnv.NewInstanceAdminClient()
+	if err != nil {
+		t.Errorf("NewInstanceAdminClient: %v", err)
```
tritone, line 2463: This also should be a `Fatalf`.

bigtable/integration_test.go
```diff
@@ -2480,77 +2507,107 @@ func TestIntegration_AdminBackup(t *testing.T) {
 
 	// Create backup
 	backupName := "mybackup"
-	defer adminClient.DeleteBackup(ctx, cluster, "mybackup")
 
-	if err = adminClient.CreateBackup(ctx, table, cluster, backupName, time.Now().Add(8*time.Hour)); err != nil {
-		t.Fatalf("Creating backup: %v", err)
+	if err = adminClient.CreateBackup(ctx, tblConf.TableID, sourceCluster, backupName, time.Now().Add(8*time.Hour)); err != nil {
+		t.Errorf("Creating backup: %v", err)
```
tritone, line 2512: I still think that this and some of the following should be changed to `Fatalf`. For example, if the backup is not successfully created, will the checks below be useful?

bigtable/integration_test.go
```diff
@@ -2480,77 +2507,107 @@ func TestIntegration_AdminBackup(t *testing.T) {
 
 	// Create backup
 	backupName := "mybackup"
-	defer adminClient.DeleteBackup(ctx, cluster, "mybackup")
```
tritone, line 2483: Is there a reason for removing this?

tritone (commented):
Generally looks good, a few more testing Q's

bigtable/integration_test.go
tbpg: Looks like this is still an issue (e.g. listing backups doesn't make sense if `CreateBackup` fails). Might need a push?

tbpg (changes requested):
A couple minor comments. Overall, looks good.

bigtable/integration_test.go
kolea2: Yes, this was unnecessary. We test `deleteBackup` later, so this was actually throwing an error in GCP trying to delete a backup that was not present.

bigtable/admin.go
kolea2: Done, thanks for the suggestion

bigtable/integration_test.go
kolea2: added back in per discussion offline

tritone (approved):

bigtable/admin.go
```diff
@@ -1583,15 +1583,24 @@ func UpdateInstanceAndSyncClusters(ctx context.Context, iac *InstanceAdminClient
...
 // RestoreTable creates a table from a backup. The table will be created in the same cluster as the backup.
+// To restore a table to a different instance, see RestoreTableFrom.
 func (ac *AdminClient) RestoreTable(ctx context.Context, table, cluster, backup string) error {
-	ctx = mergeOutgoingMetadata(ctx, ac.md)
-	prefix := ac.instancePrefix()
-	backupPath := ac.backupPath(cluster, backup)
+	return ac.RestoreTableFrom(ctx, ac.instance, table, cluster, backup)
+}
 
+// RestoreTableFrom creates a new table in the admin's instance by restoring from the specified backup in a different instance.
```
tbpg, line 1591:
```suggestion
// RestoreTableFrom creates a new table in the admin's instance by restoring from the given backup and instance.
```

bigtable/admin.go
```diff
@@ -1583,15 +1583,24 @@ func UpdateInstanceAndSyncClusters(ctx context.Context, iac *InstanceAdminClient
...
+// To restore within the same instance, see RestoreTable.
+// sourceInstance (ex. "my-instance") and sourceCluster (ex. "my-cluster") are the instance and cluster in which the new table will be restored from.
+// tableName (ex. "my-restored-table") will be the name of the newly created table
+// backupName (ex. "my-backup") is the name of the backup to restore
```
tbpg, line 1595:
```suggestion
// tableName (ex. "my-restored-table") will be the name of the newly created table.
// backupName (ex. "my-backup") is the name of the backup to restore.
```

tbpg (approved):
A couple minor comment comments. Looks good.

kolea2:
Thank you @tbpg @tritone !

PR #4015, opened 2021-04-27
chore(internal/gapicgen): add regen-only flag

noahdietz:
Adds a `regen-only` flag to local mode that disables generation version updates, manifest updates, and vetting/compilation. This is useful for iterating on generator changes or just fiddling. 

This also makes `gensnippets` respect the `only-gapics` flag, disabling snippet regen when that flag is enabled.

Finally, it optionally supplies the `release-level` option to the generator depending on if it is empty or not in the `microgenConfig`, and uses the flag name `api-service-config` instead of `gapic-service-config`, because that is the more accurate name.


codyoss (approved):
LGTM

PR #4018, opened 2021-04-27
chore(storage): generate GAPIC under internal subdir

noahdietz:
Add a gRPC Go GAPIC for the Storage API under an internal subdirectory of the submodule to ensure no one can import it. 

This GAPIC is being added for proof of concept development and should not be used in production or by any other client library.

Generated with the following:

```bash
$ cd google-cloud-go
$ export GOOGLE_CLOUD_GO=$(pwd)
$ cd internal/gapicgen
$ go run cloud.google.com/go/internal/gapicgen/cmd/genbot \
  -local \
  -regen-only \
  -only-gapics \
  -gapic=cloud.google.com/go/storage/internal/apiv1 \
  -gocloud-dir=$GOOGLE_CLOUD_GO
```


storage/internal/apiv1/storage_client.go
```diff
@@ -0,0 +1,2388 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+)
+
+var newClientHook clientHook
+
+// CallOptions contains the retry settings for each method of Client.
+type CallOptions struct {
+	DeleteBucketAccessControl        []gax.CallOption
```
tritone, line 39: I assume it's fine to include everything in the surface (not just media ops) because we are keeping this under internal?

storage/internal/apiv1/storage_client.go
```diff
@@ -0,0 +1,2388 @@
...
+	}
+}
+
+func defaultCallOptions() *CallOptions {
+	return &CallOptions{
+		DeleteBucketAccessControl: []gax.CallOption{
+			gax.WithRetry(func() gax.Retryer {
+				return gax.OnCodes([]codes.Code{
+					codes.DeadlineExceeded,
+					codes.Unavailable,
```
tritone, line 116: Are these retryers all just entirely defaults, or are they defined in service proto?

storage/internal/apiv1/storage_client.go
```diff
@@ -0,0 +1,2388 @@
...
+func (c *Client) GetObject(ctx context.Context, req *storagepb.GetObjectRequest, opts ...gax.CallOption) (*storagepb.Object, error) {
+	return c.internalClient.GetObject(ctx, req, opts...)
+}
+
+// GetObjectMedia reads an object’s data.
+func (c *Client) GetObjectMedia(ctx context.Context, req *storagepb.GetObjectMediaRequest, opts ...gax.CallOption) (storagepb.Storage_GetObjectMediaClient, error) {
+	return c.internalClient.GetObjectMedia(ctx, req, opts...)
+}
+
+// InsertObject stores a new object and metadata.
```
tritone, line 1107: Nice detailed docs here, has this been updated at all based on your investigations into resumable uploads @noahdietz ?

storage/internal/apiv1/storage_client_example_test.go
```diff
@@ -0,0 +1,1046 @@
...
+	if err != nil {
+		// TODO: Handle error.
+	}
+	defer c.Close()
+
+	// TODO: Use client.
+	_ = c
+}
+
+func ExampleClient_DeleteBucketAccessControl() {
```
tritone, line 38: Out of curiosity, have these examples actually proven useful and/or been published in docs for any product?

tritone (approved):

storage/internal/apiv1/storage_client.go
noahdietz: > it's fine to include everything in the surface...because we are keeping this under internal

Yep! Also, the generator doesn't support generating bits of a proto :)

storage/internal/apiv1/storage_client.go
noahdietz: They are defined in the grpc_service_config.json in [googleapis](https://github.com/googleapis/googleapis/blob/master/google/storage/v1/storage_grpc_service_config.json) by the AP team.

They are pretty basic. And retry doesn't apply to streaming methods _after_ the stream has been created (i.e. resumption).

storage/internal/apiv1/storage_client_example_test.go
noahdietz: They GoDoc examples so they are included in the package [examples](https://pkg.go.dev/cloud.google.com/go/language/apiv1#pkg-examples) and with each [method doc](https://pkg.go.dev/cloud.google.com/go/language/apiv1#example-Client.AnalyzeEntities).

The snippetgen project generates stuff for the cloud site and tbp@ would be your PoC there.

storage/internal/apiv1/storage_client.go
noahdietz: > has this been updated at all based on your investigations into resumable uploads

Nope, this is actually just the documentation from the [proto itself](https://github.com/googleapis/googleapis/blob/master/google/storage/v1/storage.proto#L226-L250)! I agree, its quite solid.

codyoss (approved):

Issue #3963, opened 2021-04-19
storage: retry is not consistent between `storage.Reader` and `storage.Writer`


maros7:
**Client**

Storage

**Environment**

MacOSX/Linux

```bash
$ go version; go env                                                                                                                                                                       ✔  18:06:46 
go version go1.16.3 darwin/amd64
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/XXX/Library/Caches/go-build"
GOENV="/Users/XXX/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/XXX/go/pkg/mod"
GONOPROXY=""
GONOSUMDB="XXX"
GOOS="darwin"
GOPATH="/Users/XXX/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.16.3/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.16.3/libexec/pkg/tool/darwin_amd64"
GOVCS=""
GOVERSION="go1.16.3"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/XXX/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/03/XXX/=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

**Code**

Regarding this code:
https://github.com/googleapis/google-cloud-go/blob/34bbe35bbc31c3d337cacff86a3372bcae1f1f06/storage/not_go110.go#L26-L42

The `http.Client` is wrapping an error in `url.Error`:
https://github.com/golang/go/blob/9baddd3f21230c55f0ad2a10f5f20579dcf0a0bb/src/net/http/client.go#L717-L728

[...]

PR #4019, opened 2021-04-29
storage: fix retry strategy to unwrap errors

frankyn:
I updated retry strategy in google-cloud-go to follow what [google-api-go-client](https://github.com/googleapis/google-api-go-client/blob/e8cf7f70a1e0160405dd0f789a006e6b6074fbdb/internal/gensupport/resumable.go#L233-L258) by unwrapping errors which can be unwrapped. 

For testing, I followed wrapping guidance in [godocs](https://golang.org/pkg/errors/#pkg-overview) to verify the change works as expected.

Thank you in advance for your review.

Fixes: #3963


storage/go110.go
```diff
@@ -46,6 +46,8 @@ func shouldRetry(err error) bool {
 		return false
 	case interface{ Temporary() bool }:
```
tritone, line 47: We also need to modify this case to allow a wrapped temporary error to be retried. Should match https://github.com/googleapis/google-api-go-client/blob/master/internal/gensupport/resumable.go#L248 .

storage/go110_test.go
```diff
@@ -45,6 +46,8 @@ func TestInvoke(t *testing.T) {
 		{2, &googleapi.Error{Code: 599}, &googleapi.Error{Code: 428}},
 		{1, &url.Error{Op: "blah", URL: "blah", Err: errors.New("connection refused")}, nil},
 		{1, io.ErrUnexpectedEOF, nil},
+		{1, fmt.Errorf("Test unwrapping of a temporary error: %w", &googleapi.Error{Code: 500}), nil},
```
tritone, line 49: This is the correct way to test but it's not available in Go < 1.13, and we need to support back to 1.11. I believe you can use https://pkg.go.dev/golang.org/x/xerrors as a workaround.

tritone (commented):
Thanks for doing this!

storage/go110.go
frankyn: thanks @tritone, would you like me to break the switch case into if conditions instead?

storage/go110_test.go
```diff
@@ -19,6 +19,7 @@ package storage
 import (
 	"context"
 	"errors"
+	"golang.org/x/xerrors"
```
tritone, line 22: The linting is failing because this import should come after the standard library ones. It should go before L27. You can run `goimports` to correct.

tritone (approved):
Looks good to me aside from lint issue!

tritone:
@shollyman we tagged you because this is relevant to #4021 as well. Happy to discuss if you have any questions.

maros7:
I don't think that this will work as intended and please correct me if I'm off here. 

But e.g. `http.tlsHandshakeTimeoutError` will be wrapped in an `url.Error` (https://github.com/golang/go/blob/9baddd3f21230c55f0ad2a10f5f20579dcf0a0bb/src/net/http/client.go#L717-L728).
So, you will hit https://github.com/googleapis/google-cloud-go/blob/d4189e07bc773305293270902ab6c470610f4793/storage/go110.go#L36 and unwrapping will never happen due to the `return false` on L46.

tritone:
> I don't think that this will work as intended and please correct me if I'm off here.
> 
> But e.g. `http.tlsHandshakeTimeoutError` will be wrapped in an `url.Error` (https://github.com/golang/go/blob/9baddd3f21230c55f0ad2a10f5f20579dcf0a0bb/src/net/http/client.go#L717-L728).
> So, you will hit
> 
> https://github.com/googleapis/google-cloud-go/blob/d4189e07bc773305293270902ab6c470610f4793/storage/go110.go#L36
> 
> and unwrapping will never happen due to the `return false` on L46.
> Edit:
> See [maros7@0fe720d](https://github.com/maros7/google-cloud-go/commit/0fe720d79c0c5b51ee6278a3c34ee48dfc58fbc8)


Good catch, I think you're correct about this! Thanks for adding the relevant test case as well.

@frankyn let's use the proposed fix from @maros7 above.

tritone (approved):

frankyn:
Chatted with @tritone because url.Error Unwrap() is not supported prior to go1.13.x and is causing the tests to fail.
We will leave the logic to unwrap url.Error  and added a note above the line for future us.

shollyman (approved):
Thanks.  I'll see about porting this to BQ when I get a free moment as well.

PR #4022, opened 2021-04-29
fix(internal/gapicgen): use correct region tags for gensnippets

tbpg:
A side effect of this is that only methods that correspond to RPCs get
region tags.

I changed the error handling so any snippet issues are only warnings and
don't `log.Fatal` (cc @noahdietz).

I deleted and regenerated all of the snippets, which led to non-Method
snippets (e.g. `NewClient`-type functions) being deleted.

The first commit has the actual generator changes. The second commit is
the regeneration.

cc @amanda-tarafa.


codyoss (approved):
LGTM

Issue #3889, opened 2021-04-05
firestore: Delete sentinel doesn't seem to work against map field


apinet:
**Client**
cloud.google.com/go/firestore

**Environment**
google cloud functions (http)
runtime go113

**Go Environment**
go1.16.2 darwin/amd64

**Code**
```go
package main

/*
 type Doc {
 AMap map[string] string `firestore:"aMap"`
}
*/

func main() {
_,err := client.Collection("docs").Doc("id").Update(ctx, []firestore.Update{Path: "aMap.key1", Value: firestore.Delete}) 
}
```

**Expected behavior**
aMap.keys1 field must be removed from document.

**Actual behavior**
The field aMap.keys1 is set to 0 instead (Delete sentinel seems to be uninterpreted)

crwilcox:
I can't reproduce this given the information. I have added an integration test that shows this functioning as expected.

https://github.com/googleapis/google-cloud-go/pull/4026

PR #4026, opened 2021-04-29
test(firestore): add additional test verifying deleting a map value via update

crwilcox:
Demonstrate no reproduction for #3889


tritone (approved):
Looks good, thanks for adding coverage for this case.

PR #4029, opened 2021-04-29
build: switch release-please type to go-yoshi

chingor13:
`go` was an alias to the `GoYoshi` release strategy (https://github.com/googleapis/release-please/blob/c80f8a096178fcf355579a3e70a255aac82c725a/src/releasers/index.ts#L54-L55). `GoYoshi` is also available via the `go-yoshi` release type config.

We will be replacing `go` with a generic golang releaser that only updates the `CHANGES.md` and handles tagging.


codyoss (approved):
LGTM, thanks!

PR #4041, opened 2021-05-03
test(bigtable): switch to cmp for comparison

kolea2:


crwilcox:
For context my addition of EncryptionInfo, #3899, added a pointer to EncryptionInfo, which caused tests that compare values returned from listbackups and getbackup directly. As the pointers are different, that failed. This method will allow our tests to compare values, not pointer addr.

crwilcox (approved):

tritone (approved):

bigtable/admin.go
```diff
@@ -1769,6 +1769,35 @@ func (ac *AdminClient) BackupInfo(ctx context.Context, cluster, backup string) (
 	return newBackupInfo(resp)
 }
 
+func (info *BackupInfo) Equal(info2 *BackupInfo) bool {
```
codyoss, line 1772: This seems very Java-like to me. We don't typically have Equals methods in Go. Can we make use of `cmp` or add a testing helper instead?

codyoss (changes requested):

bigtable/admin.go
kolea2: that worked thanks! Updated, PTAL @codyoss

codyoss (approved):

crwilcox (approved):

codyoss (approved):
LGTM thanks!

Issue #4021, opened 2021-04-29
bigquery: client does not retry transient, low-level errors


maros7:
**Client**

BigQuery

I've raised a similar issue for the Storage client https://github.com/googleapis/google-cloud-go/issues/3963.

**Environment**

MacOSX / `gcr.io/distroless/base-debian10` on Cloud Run

**Go Environment**

```bash
$ go version; go env                                                                                                                                                                       ✔  18:06:46 
go version go1.16.3 darwin/amd64
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/XXX/Library/Caches/go-build"
GOENV="/Users/XXX/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/XXX/go/pkg/mod"
GONOPROXY=""
GONOSUMDB="XXX"
GOOS="darwin"
GOPATH="/Users/XXX/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.16.3/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.16.3/libexec/pkg/tool/darwin_amd64"
GOVCS=""
GOVERSION="go1.16.3"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/XXX/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/03/XXX/=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

**Code**

https://github.com/googleapis/google-cloud-go/blob/bc0549d588ee26b6eb3774625f04d601b64a330a/bigquery/bigquery.go#L168-L189

**Expected behavior**

The BigQuery client should retry transient, low-level errors (TLS handshake timeout, et al.).

[...]

PR #4046, opened 2021-05-03
feat(bigquery): augment retry predicate to support additional errors

shollyman:
Based on similar refactor in storage:
https://github.com/googleapis/google-cloud-go/pull/4019

Fixes: https://github.com/googleapis/google-cloud-go/issues/4021

### DIFF
--- a/bigquery/bigquery.go
+++ b/bigquery/bigquery.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
+	"strings"
 	"time"
 
 	"cloud.google.com/go/internal"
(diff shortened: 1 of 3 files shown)


bigquery/bigquery.go
```diff
@@ -177,13 +179,32 @@ func retryableError(err error) bool {
...
+	case interface{ Temporary() bool }:
+		if e.Temporary() {
+			return true
+		}
 	}
-	var reason string
-	if len(e.Errors) > 0 {
-		reason = e.Errors[0].Reason
+	// Unwrap is only supported in go1.13.x+
+	if e, ok := err.(interface{ Unwrap() error }); ok {
```
codyoss, line 206: nit: could this just be another case and false be the default?

codyoss (approved):
LGTM, just on small comment.

bigquery/bigquery.go
tritone: Unfortunately this doesn't work-- *url.Error can also wrap errors that end up being retryable. So, we need to execute both cases, which doesn't work in a switch.

tritone (commented):
Looks good! We also added io.ErrUnexpectedEOF for storage; not sure whether or not that makes sense for you.

tritone (approved):

bigquery/bigquery.go
shollyman: The difference is subtle; if we have a wrapped error that matched one of the previous cases but didn't qualify for retry, it allows another chance to unwrap and re-evaluate the inner error.

shollyman:
> Looks good! We also added io.ErrUnexpectedEOF for storage; not sure whether or not that makes sense for you.

Haven't seen this in the wild, but I see there's a few common imports that traffic it (http and grpc).  I'll add it.

PR #4058, opened 2021-05-04
profiler: fix integration test

nolanmar511:
This change switches to using go modules consistently and switches to pulling profiler agent from https://github.com/googleapis/google-cloud-go.git rather than https://code.googlesource.com/gocloud.

TESTED= When running e2e test with go1.12, there were failures before this change and no failure with this change.


codyoss (approved):
LGTM

wyk9787 (approved):

PR #4063, opened 2021-05-06
test(logging): install app-engine-go component in env test

nicoleczhu:
Fixes broken GAE env test. 

One has to manually add `gcloud components install app-engine-go` now


codyoss (approved):
LGTM

PR #4069, opened 2021-05-07
test(pubsublite): fix flaky TestPeriodicTask

tmdiep:
Two executions of the task can run before the first call to `Stop`. So we stop the task immediately after receiving the value. This doesn't guarantee no flakes in the future. If it occurs again, we can increase `pollInterval`.

Fixes: https://github.com/googleapis/google-cloud-go/issues/4064

### DIFF
--- a/pubsublite/internal/wire/periodic_task_test.go
+++ b/pubsublite/internal/wire/periodic_task_test.go
@@ -20,26 +20,37 @@ import (
 )
 
 func TestPeriodicTask(t *testing.T) {
+	const pollInterval = 10 * time.Millisecond
 	var callCount int32
 	values := make(chan int32)
 	task := func() {
 		values <- atomic.AddInt32(&callCount, 1)
 	}
-	ptask := newPeriodicTask(10*time.Millisecond, task)
+	ptask := newPeriodicTask(pollInterval, task)
+	var lastValue int32
 
-	t.Run("Start1", func(t *testing.T) {
+	t.Run("Start", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pubsublite/internal/wire/periodic_task_test.go
```diff
@@ -20,48 +20,49 @@ import (
...
+		lastValue = <-values
+		if got, want := lastValue, int32(1); got != want {
 			t.Errorf("got %d, want %d", got, want)
 		}
 	})
 
-	t.Run("Stop1", func(t *testing.T) {
+	t.Run("Stop", func(t *testing.T) {
+		// Flushes last value immediately before stopping.
+		// Note: if this test is still flaky, pollInterval can be increased.
```
tbpg, line 44: What if we move the value checks inside the task? We could use an atomic to track how many times the task has run then take specific actions based on the value and if we've restarted the task. I'd hope if the second run of the task called `Stop`, there wouldn't be a third run until the task has been restarted.

That might help avoid flake?

Feel free to leave as is if you prefer. It took me a good think to understand the flow here.

tbpg (approved):

pubsublite/internal/wire/periodic_task_test.go
tmdiep: Thanks Tyler. Your suggestion indeed sounds much better, but when I tried to call `ptask.Stop` from within the task func, data races were detected.

I've just simplified the test and added more comments to hopefully make it clearer. The approach hasn't changed though.

PR #4071, opened 2021-05-07
profiler: modify integration test to pass with go1.11

nolanmar511:
TESTED= Started integration test locally using go 1.11.13. With this change, the test passed; without it the test failed.

### DIFF
--- a/profiler/integration_test.go
+++ b/profiler/integration_test.go
@@ -64,8 +64,9 @@ retry apt-get -y -q install git >/dev/null
 mkdir -p /tmp/gocache
 export GOCACHE=/tmp/gocache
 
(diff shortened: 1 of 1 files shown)


wyk9787:
> @wyk9787
> 
> > Why are we still testing against Go 1.11? I thought we only support officially maintained Go releases.
> 
> https://github.com/googleapis/google-cloud-go#go-versions-supported
> We currently maintain compatibility to support some older environments.
> 
> > I couldn't seem to find where the tested Golang versions are specified.
> 
> This is specified internally. We test against min and max supported versions. So right now that would 1.11 and 1.16.

Got it. Thanks!

PR #4075, opened 2021-05-08
fix(bigtable/bttest): remove duplicate response cells from RowFilter_Interleave

dragonsinth:
As implemented, bttest's RowFilter_Interleave will return the same cell multiple times if it passes multiple filters.  This fixes the implementation and test expectations, and adds more clarity to the test code.

### DIFF
--- a/bigtable/bttest/inmem.go
+++ b/bigtable/bttest/inmem.go
@@ -518,14 +518,27 @@ func filterRow(f *btpb.RowFilter, r *row) (bool, error) {
 				}
 			}
 		}
-		var count int
+		var found bool
 		for _, fam := range r.families {
-			for _, cs := range fam.cells {
+			for colName, cs := range fam.cells {
+				if len(cs) == 0 {
+					continue
+				}
+				found = true
 				sort.Sort(byDescTS(cs))
-				count += len(cs)
+
+				// Remove dups
+				wIdx := 1
+				for i := 1; i < len(cs); i++ {
+					if cs[i].ts != cs[wIdx-1].ts {
+						cs[wIdx] = cs[i]
+						wIdx++
+					}
+				}
+				fam.cells[colName] = cs[:wIdx]
 			}
 		}
-		return count > 0, nil
+		return found, nil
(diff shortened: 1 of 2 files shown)


dragonsinth:
Nevermind... I tested against real big table and found that it does indeed send duplicate cells over the wire!

dragonsinth:
FWIW... I spent a bit of time working on a copy of `inmem_test.go` that hits real bigtable.  Turns out there's a few tests that don't pass vs. real bigtable (although interleave does):

```
=== RUN   TestCheckAndMutateRowWithoutPredicate
    real_test.go:947: CheckAndMutateRow error: rpc error: code = InvalidArgument desc = No mutations provided
    real_test.go:957: CheckAndMutateRow error: rpc error: code = InvalidArgument desc = No mutations provided

=== RUN   TestCheckAndMutateRowWithPredicate/no_match
    real_test.go:1158: CheckAndMutateRow error: rpc error: code = InvalidArgument desc = Row keys must be non-empty

=== RUN   TestCheckAndMutateRowWithPredicate/rowkey_regex
    real_test.go:1158: CheckAndMutateRow error: rpc error: code = InvalidArgument desc = No mutations provided

=== RUN   TestCheckAndMutateRowWithPredicate/pass_all
    real_test.go:1158: CheckAndMutateRow error: rpc error: code = InvalidArgument desc = No mutations provided

=== RUN   TestFilters
    real_test.go:1391: ReadRows error: rpc error: code = InvalidArgument desc = block_all_filter must be true if set
    real_test.go:1391: ReadRows error: rpc error: code = InvalidArgument desc = pass_all_filter must be true if set

=== RUN   TestFilterRowWithSingleColumnQualifier
    real_test.go:1732: Failed to read rows: rpc error: code = InvalidArgument desc = Error in chained row filter #0 : Error in interleaved row filter #0 : Error in field 'condition filter predicate' : Chain must contain at least two RowFilters
```

Would there be any interest in...
a) changes to bttest to replicate this behavior?
b) changes to the existing test code so that it can pass vs. real bigtable?
c) the changes I made to be able to run the tests vs. real bigtable?

PR #4076, opened 2021-05-09
feat(bigquery): expose ProjectID from Client & Job

derekperkins:
As more and more BigQuery features like BI Engine, Reservations, etc. are opt-in at a project level only, we are now using more dependency injection for multiple clients. As currently written, we have to create a throwaway dataset from the client to expose the project associated with the client and keep track of that separately to map to any jobs. The project is trivially available from both the client and job, so this PR exposes both.


derekperkins:
@shollyman @codyoss Any chance this could make it into the 1.19 release?

shollyman (approved):
Thanks.  This seems reasonable to expose.  I've also created https://github.com/googleapis/google-cloud-go/issues/4228 for myself to come back through and improve job reference construction.

shollyman:
Feedback from others was we should consider whether we should simply change to exported variables.  Primary issue is still the retained context; need to take a closer look at this to see if we want to open it up rather than simply exposing functions as laid out here.

derekperkins:
For reference, ProjectID is exported in Dataset. I don't particularly care whether ProjectID is exported or whether it is exposed with a function to protect the integrity of the object. I just worry that taking a closer look will cause this to end up in the backlog.

https://github.com/googleapis/google-cloud-go/blob/22344f085efd0573a1f9d5904e4f430e2a4d718d/bigquery/dataset.go#L30-L31

shollyman:
Closing this PR, functionality was added in https://github.com/googleapis/google-cloud-go/pull/4312.  Thanks for bringing it to my attention, ended up highlighting the autodetection gap.

derekperkins:
Thanks for making that happen, much appreciated

PR #4082, opened 2021-05-10
feat(gsuiteaddons): start generating apiv1

codyoss:


codyoss:
cc @tbpg

tbpg (changes requested):
Would you mind updating this filter, as this does not appear to be a Cloud API?

https://github.com/googleapis/google-cloud-go/blob/258980a039db26c5a1761efdb65109a52dfdd83e/internal/godocfx/main.go#L151-L154

internal/godocfx/main.go
```diff
@@ -151,6 +151,7 @@ func process(mod indexEntry, tempDir, outDir string, print bool) error {
 	filter := []string{
 		"cloud.google.com/go/analytics",
 		"cloud.google.com/go/area120",
+		"cloud.google.com/go/gsuiteaddons/apiv1",
```
tbpg, line 154:
```suggestion
		"cloud.google.com/go/gsuiteaddons",
```

So we don't have to update this for new versions in the future. The filter is prefix based.

tbpg (approved):

PR #4085, opened 2021-05-11
fix(internal/gapicgen): filter out internal directory changes

codyoss:
- Filter out internal dir changes since they do not need a module
  update.
- Remove unneeded param from method
- Log extra details when a excuting a command yields an error


codyoss:
@noahdietz Can I get you to take a peek at this PR. It fixes a bug that is currently blocking regen.

noahdietz (approved):

Issue #4079, opened 2021-05-10
logging: fix GKE resource autodetection


nicoleczhu:
GKE was: 
```json
resource: {
  type: "gce_instance"
  labels: {
    project_id: "log-bench"
    zone: "us-west1-a"
    instance_id: "4837119632664073093"
  }
}
```

Should be:
```json 
resource: {
  type: "k8s_container"
  labels: {
    pod_name: "go-gke-6969bf5b9d-rwbw8"
    container_name: "hello-app"
    cluster_name: "go-gke"
    namespace_name: "default"
    project_id: "log-bench"
    location: "us-west1-a"
  }
}
```

PR #4092, opened 2021-05-12
fix(logging): correctly detect GKE resource

nicoleczhu:
Fixes: #4079 

Changes: 
- [x] Refactored resource autodetection logic to new file `resource.go`
- [x] Fixes previous incorrect detection where GKE logs were labeled as GCE. Labels still missing `container_name` because there's no unhacky way to get this info.
- [x] End to end GKE environment tests are now available for [this here](https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:cloud-devrel%2Fclient-libraries%2Fgo%2Fgoogle-cloud-go%2Fenvironment%2Flogging%2Fkubernetes) 

Decisions: 
- `namespace_name`: `Finally, the default namespace to be used for namespaced API operations is placed in a file at /var/run/secrets/kubernetes.io/serviceaccount/namespace in each container.`
https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod
- `podname_name`: HOSTNAME envvar
- `container_name`: CONTAINER_NAME optional user provided envvar


logging/resource.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	return true
+}
+
+func detectKubernetesResource() *mrpb.MonitoredResource {
+	projectID, err := metadata.ProjectID()
+	if err != nil {
+		return nil
+	}
+	zone, err := metadata.Zone()
```
daniel-sanche, line 156: You should check to make sure this is in "us-central1-a" format, rather than "projects/xxxx/regions/us-central1-a". When  you click the "Logs" button in the GKE UI, it will assume the first format.

I had this issue recently for Cloud Functions here: https://github.com/googleapis/python-logging/pull/291

logging/resource.go
```diff
@@ -0,0 +1,266 @@
...
+	if err != nil {
+		return nil
+	}
+	clusterName, err := metadata.InstanceAttributeValue("cluster-name")
+	if err != nil {
+		return nil
+	}
+	namespaceBytes, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
daniel-sanche, line 164: I don't think we can guarantee this file will always be accessible. It looks like you're already doing error checking so it should be fine, just wanted to double check

logging/resource.go
```diff
@@ -0,0 +1,266 @@
...
+	zone, err := metadata.Zone()
+	if err != nil {
+		return nil
+	}
+	return &mrpb.MonitoredResource{
+		Type: "cloud_run_revision",
+		Labels: map[string]string{
+			"project_id":         projectID,
+			"location":           regionFromZone(zone),
+			"service_name":       os.Getenv("K_SERVICE"),
```
daniel-sanche, line 135: There are a lot of repeated magic strings in this file. Maybe turn these into constants?

logging/resource.go
```diff
@@ -0,0 +1,266 @@
...
+			"instance_name": name,
+			"zone":          zone,
+		},
+	}
+}
+
+func detectResource() *mrpb.MonitoredResource {
+	detectedResource.once.Do(func() {
+		switch {
+		// AppEngine, Functions, CloudRun are detected first, as metadata.OnGCE()
```
daniel-sanche, line 212: it looks like this should be updated to include GKE

logging/resource.go
```diff
@@ -0,0 +1,266 @@
...
+			detectedResource.pb = detectCloudFunction()
+		case isCloudRun():
+			detectedResource.pb = detectCloudRunResource()
+		case isKubernetesEngine():
+			detectedResource.pb = detectKubernetesResource()
+		case metadata.OnGCE():
+			detectedResource.pb = detectGCEResource()
+		}
+	})
+	return detectedResource.pb
```
daniel-sanche, line 226: Shouldn't there be a default global resource here?

daniel-sanche (commented):
> Labels still missing container_name because there's no unhacky way to get this info.

One (imperfect) option would be to try to detect `CONTAINER_NAME` (or something similar) and instruct users to [populate it using the downward API](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/)

This could be useful as a second option for `pod_name` and `namespace_name` too, in case there are edge cases where the current heuristics don't work

logging/resource.go
nicoleczhu: Good pt, I double checked this. Go library formats it for us already :P

logging/resource.go
nicoleczhu: correct, it will just be "" if readfile fails.  We can rely on this to exist majority of the time though. 

**its a part of kubernetes downward api contract:** 
https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod
Finally, the default namespace to be used for namespaced API operations is placed in a file at /var/run/secrets/kubernetes.io/serviceaccount/namespace in each container.

logging/resource.go
nicoleczhu: updating comment

logging/resource.go
nicoleczhu: this is handled by a separate function already see `monitoredResource`

logging/resource.go
nicoleczhu: Agree magic numbers are not great. I actually wrote one iteration with constants (camel case). But, since resource.labels have to be in snake case, I liked the explicit nature of having the snake-case keys inline with the variable setting. 

Also go is statically typed, I like the idea of the compiler not having to know all the constants that the switch case might not end up accessing.

nicoleczhu:
> > Labels still missing container_name because there's no unhacky way to get this info.
> 
> One (imperfect) option would be to try to detect `CONTAINER_NAME` (or something similar) and instruct users to [populate it using the downward API](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/)
> 
> This could be useful as a second option for `pod_name` and `namespace_name` too, in case there are edge cases where the current heuristics don't work

Adopted! @daniel-sanche, mind taking another look and stamping? Thank you.

logging/resource.go
daniel-sanche: Ok cool, yeah I just wanted to make sure all the necessary error-checking was done since I'm not super familiar with go. But it looks like it should be!

logging/resource.go
daniel-sanche: I was also thinking of things like K_SERVICE, which are used in multiple places. For Python I put these at the top, but I think it's fine as-is too

daniel-sanche (approved):

PR #4094, opened 2021-05-12
chore(internal/apidiff): fix incompat check retry

noahdietz:
Fixes a typo (using the wrong flag) on the incompatibility check retry attempt.

Adds a `-gapic` flag to enable running this differ against just one package.


codyoss (approved):
LGTM

Issue #4098, opened 2021-05-13
bigquery: Cleanup bq/storage/v1alpha2


codyoss:
Code should be removed from both this repo and genproto. Also config and denylist entries removed.

codyoss:
I believe this is done.

PR #4099, opened 2021-05-13
chore(internal/gapicgen): stop bq storage v1alpha2 proto gen

codyoss:
Updates: #4094
Updates: #4098


noahdietz (approved):
Sorry for missing this when I disabled GAPIC gen

PR #4103, opened 2021-05-13
feat(bigquery): add support for parameterized types

shollyman:
Exposes type parameterization constraints for
STRING, BYTES, NUMERIC, BIGNUMERIC types.


stephaniewang526 (approved):

bigquery/schema.go
```diff
@@ -70,6 +70,41 @@ type FieldSchema struct {
 
 	// Describes the nested schema if Type is set to Record.
 	Schema Schema
+
+	// Maximum length of the field for STRING or BYTES type.
+	//
+	// It is invalid to set value for types other than STRING or BYTES.
+	//
+	// For STRING type, this represents the maximum UTF-8 length of strings
+	// allowed in the field.  For BYTES type, this represents the maximum
```
codyoss, line 79:
```suggestion
	// allowed in the field. For BYTES type, this represents the maximum
```

bigquery/schema.go
```diff
@@ -70,6 +70,41 @@ type FieldSchema struct {
...
+	// number of bytes in the field.
+	MaxLength int64
+
+	// Precision can be used to constrain the maximum number of
+	// total digits allowed for NUMERIC or BIGNUMERIC types.
+	//
+	// It is invalid to set values for Precision for types other than
```
codyoss, line 86: I wonder if it would be beneficial to provide some of these validations in our codebase instead of having them enforced by the API. This is a bigger overall question for schema validations in general.

codyoss (approved):
LGTM, just a small nit

bigquery/schema.go
shollyman: Absolutely not, at least for the manual clients.  We've learned that as the service evolves those constraints change over time, and keeping validation in sync is a nightmare, particularly for things like preview functionality.

PR #4116, opened 2021-05-18
fix(spanner): add missing NUMERIC type to the doc for Row

yfuruyama:
I found `NUMERIC` data type is missing in the doc for [Row](https://pkg.go.dev/cloud.google.com/go/spanner#Row), so added to the doc.


hengfengli (approved):
LGTM.

Issue #4127, opened 2021-05-19
spanner: Layout of godoc for Encoder/Decoder is broken


yfuruyama:
**Client**

Cloud Spanner

**Environment**

N/A

**Go Environment**

N/A

**Code**

N/A

**Expected behavior**

Godoc for Encoder/Decoder is correctly rendered.

**Actual behavior**

Layout of godoc for [Encoder](https://pkg.go.dev/cloud.google.com/go/spanner#Encoder)/[Decoder](https://pkg.go.dev/cloud.google.com/go/spanner#Decoder) is broken. 

**Screenshots**

![image](https://user-images.githubusercontent.com/1595215/118771205-3c019b00-b8bd-11eb-9c54-3082a4002433.png)

**Additional context**

None.

PR #4128, opened 2021-05-19
fix(spanner): fix layout of godoc for Encoder and Decoder

yfuruyama:
Fix: https://github.com/googleapis/google-cloud-go/issues/4127

I added an additional indent to code sample to fix the layout issue for [Encoder](https://pkg.go.dev/cloud.google.com/go/spanner#Encoder)/[Decoder](https://pkg.go.dev/cloud.google.com/go/spanner#Decoder).

This is a screenshot for the godoc for the fixed code (I used the `godoc` command to check the fix, so the style of godoc is an old one):

image


hengfengli (approved):
LGTM. Thanks for the fix!

spanner/value.go
```diff
@@ -71,44 +71,44 @@ var (
 // Encoder is the interface implemented by a custom type that can be encoded to
 // a supported type by Spanner. A code example:
 //
-// type customField struct {
-//     Prefix string
-//     Suffix string
-// }
+//   type customField struct {
```
codyoss, line 74: nit: It would be nice to turn these into a first class godoc example. Here is an example from the storage library:
- Godoc: https://pkg.go.dev/cloud.google.com/go/storage#example-ACLHandle.Delete
- Code: https://github.com/googleapis/google-cloud-go/blob/a5bbdb584c82f628069a5c0975e273ea45273782/storage/example_test.go#L527-L537

hengfengli: I am not sure if this is a good case for that. There are two reasons to put in this way: 

* The first class godoc example seems to be a main function to guide users how to use the api. 
* The example can be close to the code. Probably better in the comment instead of jumping to another file.

yfuruyama: Thanks for suggestion! This example contains a type definition and method definition, so if I understand correctly, we need a separate file to turn this into first class godoc example.

For example, sort package has a similar example for [Example (SortWrapper)](https://golang.org/pkg/sort/#pkg-overview) by [this code](https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/sort/example_wrapper_test.go). I'm wondering if I should do similar thing for this example.

hengfengli:
I'll merge this in first. We can improve the doc later if necessary.

PR #4143, opened 2021-05-21
feat(pubsublite): flush and reset committer

tmdiep:
Preparation for admin seek.

* Adds `ackTracker.Reset()` and `commitCursorTracker.Reset()` to reset these trackers to their initial state.
* Adds `committer.BlockingReset()`, which flushes pending commits to the server and then resets the state of the committer.


pubsublite/internal/wire/committer.go
```diff
@@ -228,9 +251,16 @@ func (c *committer) unsafeInitiateShutdown(targetStatus serviceStatus, err error
 }
 
 func (c *committer) unsafeCheckDone() {
```
manuelmenzella-google, line 253: Nit; do you mind adding a short doc explaining what this function does? It took me a while to figure it out.

manuelmenzella-google (approved):

pubsublite/internal/wire/committer.go
tmdiep: Done

PR #4144, opened 2021-05-21
feat(pubsublite): detect stream reset signal

tmdiep:
Preparation for admin seek.

Modifies `retryableStream` to detect the Pub/Sub Lite `RESET` signal and forward the notification to the `streamHandler` implementation. The `RESET` signal is just a `Status` that contains particular error details. Upon receipt, the Subscribe stream will reset its state and handle an out-of-band seek operation.


pubsublite/internal/wire/rpc.go
```diff
@@ -130,7 +131,29 @@ func retryableReadOnlyCallOption() gax.CallOption {
 	})
 }
 
-const pubsubLiteDefaultEndpoint = "-pubsublite.googleapis.com:443"
+const (
+	pubsubLiteDefaultEndpoint = "-pubsublite.googleapis.com:443"
+	pubsubLiteDomain          = "pubsublite.googleapis.com"
```
manuelmenzella-google, line 136: Nit; can you call this pubsubLiteErrorDomain? I was confused and thought this was an actual URL domain

pubsublite/internal/wire/streams.go
```diff
@@ -63,11 +65,15 @@ type streamHandler interface {
...
 	// onStreamStatusChange is used to notify stream handlers when the stream has
-	// changed state. A `streamReconnecting` status change is fired before
-	// attempting to connect a new stream. A `streamConnected` status change is
-	// fired when the stream is successfully connected. These are followed by
-	// onResponse() calls when responses are received from the server. These
-	// events are guaranteed to occur in this order.
+	// changed state.
+	// - A `streamReconnecting` status change is fired before attempting to
+	//   connect a new stream.
+	// - A `streamResetState` status change may be fired if the server broke a
```
manuelmenzella-google, line 71: It feels a bit strange to have streamResetState fire after streamReconnecting. Shouldn't we have the opposite order?

manuelmenzella-google (approved):

pubsublite/internal/wire/rpc.go
tmdiep: Done

pubsublite/internal/wire/streams.go
tmdiep: Ah yes, that's the order I initially had, and then ran into a couple of issues:

* The receivers stop/disable things upon `streamReconnecting` and also need to do the same actions for `streamResetState`, so it was more practical to send the former first to avoid the duplication.
* When the stream is reinitialized, the server can send the `reset` signal straight after the initial request. So we should fire these notifications consistently in this order.

Amended the comments to hopefully reduce the confusion.

PR #4145, opened 2021-05-21
refactor(pubsublite): allow subscribe stream trackers to be reset

tmdiep:
Preparation for admin seek.

* Enables the `flowControlBatcher` and `subscriberOffsetTracker` to be reset to their initial state.
* Enables `messageDeliveryQueue` to the stopped, discarding any undelivered messages, and restarted from an empty state. Also added `Wait()` to wait for the goroutine to terminate, to ensure the user's message receiver func has exited before reacting to the seek.


pubsublite/internal/wire/subscriber.go
```diff
@@ -46,57 +47,70 @@ type ReceivedMessage struct {
...
-	if mq.status == serviceUninitialized {
-		go mq.deliverMessages()
-		mq.status = serviceActive
+	if mq.stopC != nil {
+		return
 	}
+
+	mq.stopC = make(chan struct{})
+	mq.messagesC = make(chan *ReceivedMessage, mq.bufferSize)
+	mq.active.Add(1)
```
manuelmenzella-google, line 75: Does this mean that Start() can be called multiple times before calling Stop()? Wo't stopC and messageC be overwritten in that case? I'm not sure I fully understand the semantics for this change. It would be nice to have a few additional coments.

tmdiep: Ah, sorry for the confusion. I added some comments.

Stop() will set stopC and messageC to nil. Start will check whether stopC is nil before recreating. Essentially, the goroutine "owns" these 2 channels and they are destroyed once it stops.

manuelmenzella-google (approved):

PR #4148, opened 2021-05-21
test(bigtable): refactor bigtable integration tests

crwilcox:
The integration test suite for bigtable has drifted a bit. This hardens some tests, corrects issues in others. Also some refactoring bringing the flags together and moving to automatically test creation tests and generate instance names.


bigtable/export_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"flag"
 	"fmt"
 	"strings"
+
+	//"testing"
```
kolea2, line 26: still needed?

bigtable/CONTRIBUTING.md
```diff
@@ -0,0 +1,24 @@
...
+The Bigtable integration tests will target the emulator by default. Some of the
+tests can only be run against production however. In order to do this you will
+need to specify a few command line flags.
+
+```
+go test -test.run="TestIntegration_*" -v \
+    -it.use-prod \
+    -it.project="your-project-id" \
+    -it.cluster="your-test-cluster" \
+    -it.instance="your-test-instance"
```
kolea2, line 16: I would mention that other cmd line flags can be found in `export_test.go` if needed?

bigtable/integration_test.go
```diff
@@ -1263,14 +1249,25 @@ func TestIntegration_Admin(t *testing.T) {
...
 
-	// Validate Encyrption Info not configured
-	encryptionInfo, err := adminClient.EncryptionInfo(ctx, "mytable")
-	if err != nil {
-		t.Fatalf("EncryptionInfo: %v", err)
-	}
-	if got, want := len(encryptionInfo), 0; !cmp.Equal(got, want) {
-		t.Fatalf("Number of Clusters with Encryption Info: %v, want: %v", got, want)
+	// Validate Encryption Info configured to default. (not supported by emulator)
+	if testEnv.Config().UseProd {
```
kolea2, line 1253: is there something we can log that the tests were skipped?

bigtable/integration_test.go
```diff
@@ -1263,14 +1265,25 @@ func TestIntegration_Admin(t *testing.T) {
...
```
tritone, line 1269: Is it possible to set up the tests so that we can separate out ones that need prod and ones that don't? I feel like this  setup where we skip part of a test is not great.

bigtable/CONTRIBUTING.md
```diff
@@ -0,0 +1,24 @@
+# Bigtable client contribution guide
+
+For a complete guide to contributing to Google Cloud Go client libraries, view
+the [top-level contributing guide](../CONTRIBUTING.md).
+
+## Running Integration Tests
+The Bigtable integration tests will target the emulator by default. Some of the
+tests can only be run against production however. In order to do this you will
+need to specify a few command line flags.
```
tritone, line 9: Does our continuous build run against prod or against the emulator? Might be good to add some indication about what the CI is doing.

bigtable/export_test.go
```diff
@@ -45,13 +55,25 @@ func init() {
...
 	flag.StringVar(&c.Cluster, "it.cluster", "", "Bigtable cluster to use")
-	flag.StringVar(&c.Table, "it.table", "", "Bigtable table to create")
+	flag.StringVar(&c.Table, "it.table", "it-table", "Bigtable table to create")
 	flag.BoolVar(&c.AttemptDirectPath, "it.attempt-directpath", false, "Attempt DirectPath")
 	flag.BoolVar(&c.DirectPathIPV4Only, "it.directpath-ipv4-only", false, "Run DirectPath on a ipv4-only VM")
 
 	// Backwards compat
 	flag.StringVar(&legacyUseProd, "use_prod", "", `DEPRECATED: if set to "proj,instance,table", run integration test against production`)
 
+	// Don't test instance creation by default, as quota is necessary and aborted tests could strand resources.
```
tritone, line 65: Do we have some way of cleaning up stranded resources after the fact?

bigtable/export_test.go
```diff
@@ -45,13 +55,25 @@ func init() {
...
+	flag.BoolVar(&runCreateInstanceTests, "it.run-create-instance-tests", true,
+		"Run tests that create instances as part of executing. Requires sufficient Cloud Bigtable quota. Requires that it.use-prod is true.")
+	flag.StringVar(&instanceToCreateZone, "it.instance-to-create-zone", "us-central1-b",
+		"The zone in which to create the new test instance.")
+	flag.StringVar(&instanceToCreateZone2, "it.instance-to-create-zone2", "us-east1-c",
+		"The zone in which to create a second cluster in the test instance.")
+	// Use sysctl or iptables to blackhole DirectPath IP for fallback tests.
+	flag.StringVar(&blackholeDpv6Cmd, "it.blackhole-dpv6-cmd", "", "Command to make LB and backend addresses blackholed over dpv6")
```
tritone, line 73: Are these flags actually used by the CI at all, or are they just for manual testing?

bigtable/integration_test.go
crwilcox: Skipped tests are logged, at least with `-v`

bigtable/integration_test.go
crwilcox: I agree. I am going to follow this with a PR that also allows using env vars, so we can add a prod targeted IT run to continuous


It isn't really an either or, but there are some tests the emulator doesn't support.

bigtable/CONTRIBUTING.md
crwilcox: emulator right now. I am working to fix this though. Today our emulator tests run on presubmit as well which may/may not be the correct thing (they are pretty fast against emulator)

bigtable/export_test.go
crwilcox: All tests that create instances are also deferring deletion. that said, they could strand things. There isn't a script to do that in the repo. I have a sh script locally for this since this can be a problem, specifically when debugging (less so in a test run)

bigtable/export_test.go
crwilcox: manual right now, which is a problem of sorts since CI should be running these on continuous runs I think. Especially since some behavior can only be validated against prod.

bigtable/export_test.go
tritone: For storage, we give test buckets a standard prefix, and then when running integration tests there is a method that cleans up any extra buckets in your project with that prefix. Would something like that help here? I think we used the TestMain to accomplish this. https://golang.org/pkg/testing/#hdr-Main

kolea2 (approved):

bigtable/export_test.go
crwilcox: Yep. I think this is something we could do. We do want one instance to hang around. The other ones use a standard prefix as well (create-{unix_timestamp}) so we could likely do cleanup. Practically speaking this hasn't been an issue. Though it may be a sane addition.

crwilcox: https://github.com/googleapis/google-cloud-go/issues/4157 to track work here.

bigtable/CONTRIBUTING.md
crwilcox: https://github.com/googleapis/google-cloud-go/issues/4158

Issue #4157, opened 2021-05-24
bigtable: introduce resource cleanup to integration tests


crwilcox:
Today, some of the bigtable integration tests create instances. It would be a good add to, as part of TestMain, to clean these up if they are stale. This may require changing naming a bit and including some form of a timestamp so we can detect stale instances easily.

PR #4159, opened 2021-05-24
test(bigtable): run against live endpoint for continuous tests

### LINKED ISSUE #4158: bigtable: run tests against live bigtable for continuous test runs
today the bigtable tests are running under emulation for the continuous runs. As some tests require live bigtable, we should use the live instance for continuous.

crwilcox:
Fixes #4158 (target live service for continuous tests) and #4157  (cleanup)


crwilcox:
Manual continuous run is green 🥳

kolea2 (approved):
this is awesome!

bigtable/integration_test.go
```diff
@@ -96,7 +100,48 @@ func TestMain(m *testing.M) {
...
 	}
-	os.Exit(m.Run())
+	exit := m.Run()
+	if err := cleanup(c); err != nil {
+		log.Printf("Post-test cleanup failed: %v", err)
+	}
+	os.Exit(exit)
+}
+
+func cleanup(c IntegrationTestConfig) error {
```
tritone, line 110: Nice, thanks for adding this.

bigtable/integration_test.go
```diff
@@ -80,14 +83,15 @@ var instanceToCreate string
...
-		instanceToCreate = fmt.Sprintf("create-%d", time.Now().Unix())
+		instanceToCreate = fmt.Sprintf("bt-it-%d", time.Now().Unix())
 	}
 }
 
 func TestMain(m *testing.M) {
 	flag.Parse()
 
-	c := integrationConfig
+	env, _ := NewIntegrationEnv()
```
tritone, line 93: You should receive the error from `NewIntegrationEnv` and fail on it if it's not nil.

bigtable/integration_test.go
```diff
@@ -96,7 +100,48 @@ func TestMain(m *testing.M) {
...
+			timestamp := info.Name[len(instancePrefix):]
+			t, err := strconv.ParseInt(timestamp, 10, 64)
+			if err != nil {
+				return err
+			}
+			uT := time.Unix(t, 0)
+			sevenDays := time.Hour * 24 * 7
+			if time.Now().After(uT.Add(sevenDays)) {
+				// delete an instance older than 7 days
+				iac.DeleteInstance(ctx, info.Name)
```
tritone, line 139: I think even a shorter timeframe than 7 days is probably good given how expensive bigtable instances are, no? Also, is it just instances that need to be deleted, or any other resources as well?

tritone (commented):
Good improvements, thanks!

bigtable/integration_test.go
crwilcox: There are instances and clusters, but they are coupled, so 99% sure I don't need to also call delete cluster. @kolea2 can confirm likely.

kolea2: yep! https://cloud.google.com/bigtable/docs/deleting-instance

bigtable/integration_test.go
crwilcox: I have altered this to panic if this happens with an error.

bigtable/integration_test.go
```diff
@@ -96,7 +103,46 @@ func TestMain(m *testing.M) {
...
+	if err := cleanup(c); err != nil {
+		log.Printf("Post-test cleanup failed: %v", err)
+	}
+	os.Exit(exit)
+}
+
+func cleanup(c IntegrationTestConfig) error {
+	// Cleanup resources marked with bt-it- after a time delay
+	timeToWait := time.Hour * 48 // two days
+	instancePrefix := "bt-it-"
```
codyoss, line 116: nit: could this be a const so the cleanup and create can stay in sync should one change over time.

PR #4170, opened 2021-05-25
feat(logging): add new jsonlog package for logging to stderr

codyoss:
This new package is meant to be used in environments like Cloud
Run, Cloud Function, GKE, etc. In these environments the Cloud
Logging agent can parse the output of this package and transform
it to the same format used by the Cloud Logging API. Fields like
the monitored resource are auto-detected by the agent and do not
need to be filled out by log producer.

For better code reuse and to not expose any additional unneeded
APIs I refactored some common code into internal.


logging/internal/common.go
```diff
@@ -39,3 +45,125 @@ func LogIDFromPath(parent, path string) string {
...
+//   * spanID (optional):       	"1"
+//   * traceSampled (optional): 	true
+func DeconstructXCloudTraceContext(s string) (traceID, spanID string, traceSampled bool) {
+	// As per the format described at https://cloud.google.com/trace/docs/setup#force-trace
+	//    "X-Cloud-Trace-Context: TRACE_ID/SPAN_ID;o=TRACE_TRUE"
+	// for example:
+	//    "X-Cloud-Trace-Context: 105445aa7843bc8bf206b120001000/1;o=1"
+	//
+	// We expect:
+	//   * traceID (optional): 			"105445aa7843bc8bf206b120001000"
```
nicoleczhu, line 76: Tiny nit: the formatting/indents are off. Is it just how Github UI is rendering it?

logging/jsonlog/jsonlog.go
```diff
@@ -0,0 +1,327 @@
...
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+// NewLogger creates a Logger that logs structured JSON to Stderr. The value of
+// parent must be in the format of:
+//    projects/PROJECT_ID
+//    folders/FOLDER_ID
+//    billingAccounts/ACCOUNT_ID
+//    organizations/ORG_ID
+func NewLogger(parent string, opts ...LoggerOption) (*Logger, error) {
```
nicoleczhu, line 44: Can it just be called `Logger` for consistency? Or does the name need to be different from the `Logger` fn in logging.go?

logging/jsonlog/example_test.go
```diff
@@ -0,0 +1,66 @@
...
+		// TODO: handle error.
+	}
+	l.Infof("Hello World!")
+}
+
+func ExampleNewLoggerFromRequest() {
+	// Create a Logger with additional information pulled from the current
+	// request context.
+	var req *http.Request
+	l, err := jsonlog.NewLoggerFromRequest("projects/PROJECT_ID", req)
```
nicoleczhu, line 37: Neat to have common requests on a logger level! Can we also continue to allow users to just log a Httprequest obj without having to create a special request-logger?

Otherwise, for consistency with common labels, can we impl this as: 
`jsonlog.NewLogger()`
`l.WithRequest`

logging/jsonlog/jsonlog.go
```diff
@@ -0,0 +1,327 @@
...
+	sampled bool
+	spanID  string
+}
+
+// Debugf is a convenience method for writing an Entry with a Debug Severity
+// and the provided formatted message.
+func (l *Logger) Debugf(format string, a ...interface{}) {
+	e := entry{
+		Message:  fmt.Sprintf(format, a...),
+		Severity: "DEBUG",
```
nicoleczhu, line 114: Does it make sense for the severity strs to be declared as constants? Not a go expert, asking more from readability  standpoint.

logging/jsonlog/jsonlog.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+	}
+	l := &Logger{
+		w:   os.Stderr,
+		req: req,
+	}
+
+	var traceHeader string
+	if r != nil && r.Header != nil {
+		traceHeader = r.Header.Get("X-Cloud-Trace-Context")
```
nicoleczhu, line 81: maybe declare "X-Cloud-Trace-Context" as a const in the common file?

logging/jsonlog/jsonlog_test.go
```diff
@@ -0,0 +1,369 @@
...
+	if err != nil {
+		t.Fatalf("unable to marshal: %v", err)
+	}
+	l.Infof("Hello, %s!", "World")
+	if diff := cmp.Diff(string(b), strings.TrimSpace(buf.String())); diff != "" {
+		t.Fatalf("Logger.Infof() mismatch (-want +got):\n%s", diff)
+	}
+	buf.Reset()
+
+	e.Severity = "NOTICE"
```
nicoleczhu, line 69: Since these unit tests are repetitive, maybe we can save some lines and put test cases in an struct array? e.g. https://github.com/googleapis/google-cloud-go/blob/812ef1ffdce2e87570660b58f0e725ad51f68546/logging/logging_unexported_test.go#L298

nicoleczhu: oh i see you did this already in TestLoggerWithLabels

logging/jsonlog/option.go
```diff
@@ -0,0 +1,63 @@
...
+
+type onErrorHook struct{ hook func(error) }
+
+func (h onErrorHook) set(l *Logger) {
+	l.errhook = h.hook
+}
+
+// WithWriter changes where the JSON payloads of a Logger
+// are written to. By default they are written to Stderr.
+func WithWriter(w io.Writer) LoggerOption { return withWriter{w: w} }
```
nicoleczhu, line 38: Neat! thank you for factoring this request in.

nicoleczhu (commented):
Thanks for this AMAZING pr @codyoss ! I left some clarifying & UX questions

logging/jsonlog/jsonlog.go
codyoss: I think it should be different. Logger() implies just a logger is returned which is true for cloud.google.com/go/logging.Logger. That is not true here as we preform some light validation which can result in an error being returned. For the other Logger this same type of validation happens at the same time as connections dailing in NewClient I believe.

Because an error is also returned I think the `New` semantic works well here.

logging/internal/common.go
codyoss: Whoops forgot to remove the inline comment when I added it to the function docs, good catch!

nicoleczhu (approved):

codyoss:
@nicoleczhu I was thinking of switching the branch to merging this into so that we could release this as a preview and get some user feedback. How does that sound to you?

codyoss:
Release: https://github.com/googleapis/google-cloud-go/releases/tag/logging%2Fv1.5.0-jsonlog-preview

AidanWenzel:
@codyoss Will this ever be merged into the core `logging` release?

PR #4171, opened 2021-05-26
pubsub: Add Publisher and Subscriber interfaces which are shared by Pub/Sub and Pub/Sub Lite

### LINKED ISSUE #4141: pubsub: add interface types for shared dataplane operations with Pub/Sub Lite
You should be able to define an interface as:

```
type Publisher {
  Stop()
  Publish(ctx context.Context, msg *pubsub.Message) *pubsub.PublishResult
}
```

Which is implemented by both Pub/Sub's Topic and Pub/Sub Lite's pscompat.PublisherClient, so you can be agnostic to which you are using. Similarly, on the subscriber side you can define an interface:

```
type Subscriber {
  Receive(ctx context.Context, f MessageReceiverFunc) error
}
```

Which is implemented by both the Pub/Sub Subscription and Pub/Sub Lite pscompat.SubscriberClient. Also, pubsub.MessageReceiverFunc should be defined as an alias for the type of function that needs to be passed to the Receive method, which is currently an anonymous function type.

I think that these interfaces should be defined and added to the pubsub package, as it would assist users in being able to write code that is agnostic to which system it is talking to.

Thoughts?

dpcollins-google:
Fixes #4141


dpcollins-google:
@hongalex

dpcollins-google:
manually set the CLA: The coauthor is me, just poorly configured in git.

hongalex:
What do you think we should do should we ever want to add new methods to this interface (which would be a breaking change)? Are we confident in the stability of publisher/subscriber interfaces that we won't ever add new methods?

@codyoss @shollyman Do you know of examples of exposing interfaces like this in other libraries?

codyoss:
We purposefully do not export interfaces for our clients the reasons @hongalex mentioned above -- breaking changes. If you really wanted to asset a common interface I would make it private in internal.

codyoss (changes requested):
See comment above about interfaces in Go.

pubsublite/pscompat/subscriber.go
```diff
@@ -230,7 +230,7 @@ func (si *subscriberInstance) Wait(ctx context.Context) error {
 // is connected to multiple partitions. Only one call from any connected
 // partition will be outstanding at a time, and blocking in this receiver
 // callback will block the delivery of subsequent messages for the partition.
-type MessageReceiverFunc func(context.Context, *pubsub.Message)
+type MessageReceiverFunc = pubsub.MessageReceiverFunc
```
tmdiep, line 233: If there are no objections to defining MessageReceiverFunc in the pubsub package, we could just remove this and change the parameter of SubscriberClient.Receive() to pubsub.MessageReceiverFunc. Any extra comments here can just be moved to SubscriberClient.Receive().

tmdiep:
I just looked at the test failures for this PR and realized that they caught a breaking change. We need to declare `type MessageReceiverFunc = func(context.Context, *Message)` in `pubsub` (note: add = to define as an alias). If declared as `type MessageReceiverFunc func(context.Context, *Message)`, this breaks any users' existing interfaces (I confirmed with a test).

I now think that we should just leave `pubsub` as-is and make `pubsublite`'s interfaces identical. I made this change in  https://github.com/googleapis/google-cloud-go/pull/4281, and also added examples for users to copy, which also have the benefit of triggering compilation errors.

pubsublite/pscompat/subscriber.go
tmdiep: Retracting, see comment below.

PR #4194, opened 2021-06-02
feat(spanner): add custom instance config to tests

mohanli-ml:
In spanner integration tests, an instance will be created before the test. Currently, the instance is created with the first instance config. However, for staging environment the instance config is shared among GCP projects and can run out of resources. So this PR tries to solve this issue by allowing users to custom the instance config by adding an environment variable GCLOUD_TESTS_GOLANG_SPANNER_INSTANCE_CONFIG.

### DIFF
--- a/spanner/integration_test.go
+++ b/spanner/integration_test.go
@@ -63,6 +63,11 @@ var (
 	// by setting the environment variable GCLOUD_TESTS_GOLANG_SPANNER_HOST
 	spannerHost = getSpannerHost()
 
+	// instanceConfig specifies the instance config used to create an instance for testing.
+	// It can be changed by setting the environment variable
+	// GCLOUD_TESTS_GOLANG_SPANNER_INSTANCE_CONFIG.
+	instanceConfig = getInstanceConfig()
+
 	dbNameSpace       = uid.NewSpace("gotest", &uid.Options{Sep: '_', Short: true})
 	instanceNameSpace = uid.NewSpace("gotest", &uid.Options{Sep: '-', Short: true})
 	backupIDSpace     = uid.NewSpace("gotest", &uid.Options{Sep: '_', Short: true})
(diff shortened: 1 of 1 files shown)


spanner/integration_test.go
```diff
@@ -63,6 +63,11 @@ var (
 	// by setting the environment variable GCLOUD_TESTS_GOLANG_SPANNER_HOST
 	spannerHost = getSpannerHost()
 
+	// instanceConfig specifies the instance config used to create a instance for testing.
```
hengfengli, line 66: a -> an

spanner/integration_test.go
```diff
@@ -262,6 +271,10 @@ func initIntegrationTests() (cleanup func()) {
 	if err != nil {
 		log.Fatalf("Cannot get any instance configurations.\nPlease make sure the Cloud Spanner API is enabled for the test project.\nGet error: %v", err)
 	}
+	configName := config.Name
+	if instanceConfig != "" {
```
hengfengli, line 275: If `instanceConfig` is not empty, there is no need to call `ListInstanceConfigs`. 

```go
        var configName string
	if instanceConfig != "" {
		configName = fmt.Sprintf("projects/%s/instanceConfigs/%s", testProjectID, instanceConfig)
	} else {
	        configIterator := instanceAdmin.ListInstanceConfigs(ctx, &instancepb.ListInstanceConfigsRequest{
		        Parent: fmt.Sprintf("projects/%s", testProjectID),
	        })
	        config, err := configIterator.Next()
	        if err != nil {
		        log.Fatalf("Cannot get any instance configurations.\nPlease make sure the Cloud Spanner API is enabled for the test project.\nGet error: %v", err)
	        }
	        configName := config.Name
        }
```

spanner/integration_test.go
mohanli-ml: Done

spanner/integration_test.go
mohanli-ml: Added an if-else.

hengfengli (approved):
LGTM.

Issue #2600, opened 2020-07-14
bigtable: emulator crashes with concurrent operations


coryan:
We never filed a bug for this, though we had plenty of communication with the emulator developers in googleapis/google-cloud-cpp#441.

The emulator crashes when performing concurrent operations. I think maybe SampleRowKeys combined with DropRowRange.

The first uses a lock to prevent problems:

https://github.com/googleapis/google-cloud-go/blob/fb3791807bf818c5c9d78d2c7e84303192d62de6/bigtable/bttest/inmem.go#L1064

The second does not seem to use the same lock:

https://github.com/googleapis/google-cloud-go/blob/fb3791807bf818c5c9d78d2c7e84303192d62de6/bigtable/bttest/inmem.go#L252-L287

That seems racy.

dragonsinth:
Confirmed
https://github.com/googleapis/google-cloud-go/pull/4200

coryan:
I think we need to reopen this:

```
Cloud Bigtable emulator running on 127.0.0.1:8480
fatal error: concurrent map iteration and map write

goroutine 13076 [running]:
runtime.throw(0x9ab765, 0x26)
        /usr/lib/google-golang/src/runtime/panic.go:623 +0x72 fp=0xc000925768 sp=0xc000925738 pc=0x42c3a2
runtime.mapiternext(0xc000925888)
        /usr/lib/google-golang/src/runtime/map.go:860 +0x597 fp=0xc0009257f0 sp=0xc000925768 pc=0x40ed87
cloud.google.com/go/bigtable/bttest.(*row).size(0xc00d15d900, 0x64)
        /usr/local/google/home/igorbernstein/go/pkg/mod/cloud.google.com/go/bigtable@v1.0.0/bttest/inmem.go:1312 +0x101 fp=0xc000925958 sp=0xc0009257f0 pc=0x865461
cloud.google.com/go/bigtable/bttest.(*server).SampleRowKeys.func1(0xa3ce80, 0xc00d15d900, 0xc008ebf701)
        /usr/local/google/home/igorbernstein/go/pkg/mod/cloud.google.com/go/bigtable@v1.0.0/bttest/inmem.go:1080 +0x99 fp=0xc0009259d8 sp=0xc000925958 pc=0x8677c9
github.com/google/btree.(*node).iterate(0xc009421b00, 0x1, 0x0, 0x0, 0x0, 0x0, 0x100, 0xc000925c20, 0x101)
[...]

PR #4200, opened 2021-06-02
fix(bigtable/bttest): fix locking bug in DropRowRange

dragonsinth:
DropRowRange's locking is inconsistent with the rest of bttest
https://github.com/googleapis/google-cloud-go/issues/2600

ModifyColumnFamilies demonstrates the correct pattern:
- Use the server global mutex to reference the table by name
- Use the table's mutex to guard data reads and writes.

The existing implementation can data race if another concurrently running method (such as SampleRowKeys)
has already dropped the server mutex and acquired the table mutex.  Nothing then prevents DropRowRange
from mutating table data during a Sample or Scan.


tritone:
Thanks very much for the fix here! I'm going to close in favor of #4207 which includes a test that reproduces the issue as well.

dragonsinth:
Sweet!  Thank you.

PR #4206, opened 2021-06-02
feat(firestore): Add support for PartitionQuery

crwilcox:
Other implementations of this:
Java 
Merged https://github.com/googleapis/java-firestore/commit/39965489cbc836af573e500d57007c88241d7eb6
https://github.com/googleapis/java-firestore/pull/202 (“add support for QueryPartition API”)
https://github.com/googleapis/java-firestore/pull/418 (“verify page count before invoking QueryPartition RPC”)
https://github.com/googleapis/java-firestore/pull/417 (“add implicit ordering for startAt(DocumentReference) calls”)
Released in java-firestore v2.0.0

Node.js 
https://github.com/googleapis/nodejs-firestore/pull/1320 (“add support for QueryPartition API”)
https://github.com/googleapis/nodejs-firestore/pull/1328  (“add implicit ordering for startAt(DocumentReference) calls”)
Merged. https://github.com/googleapis/nodejs-firestore/pull/1110/files#diff-487a284b3e5f6cec3fe8dc89ea74120d


firestore/collgroupref.go
```diff
@@ -36,3 +45,107 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
 		},
 	}
 }
+
+// GetPartitions returns a slice of QueryPartition objects. The number must be
```
tritone, line 49: This is kind of unclear to me. What is a QueryPartition? Why would I want to get it? The type docs don't really illuminate further. Adding more color and a link to the public docs for the feature would help.

Also, "The number" == `partitionCount`?

firestore/collgroupref.go
```diff
@@ -36,3 +45,107 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
...
+	if partitionCount <= 0 {
+		return nil, errors.New("a positive partitionCount must be provided")
+	} else if partitionCount == 1 {
+		return []QueryPartition{{CollectionGroupQuery: cgr.Query, StartAt: "", EndBefore: ""}}, nil
+	}
+
+	db := cgr.c.path()
+	ctx = withResourceHeader(ctx, db)
+
+	// CollectiongGroup Queries need to be ordered by __name__ ASC
```
tritone, line 61: Typo in CollectionGroup, and comments should have a period.

firestore/integration_test.go
```diff
@@ -1601,3 +1601,35 @@ func TestDetectProjectID(t *testing.T) {
...
+
+	for _, tc := range []struct {
+		collectionID           string
+		expectedPartitionCount int
+	}{
+		// Verify no failures if a collection doesn't exist
+		{collectionID: "does-not-exist", expectedPartitionCount: 1},
+		// Verify a collectionID with a small number of results returns a partition
+		{collectionID: coll.collectionID, expectedPartitionCount: 1},
+		// TODO: could implement a scenario where a collectionID had a good
```
tritone, line 1622: Do other features or other firestore libraries use test fixtures like this? Is there any way around requiring a DB that large?

firestore/collgroupref.go
```diff
@@ -36,3 +45,107 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
...
+		lastPart.StartAt = cursorReferences[len(cursorReferences)-1].GetReferenceValue()
+	}
+	partitionQueries = append(partitionQueries, lastPart)
+
+	return partitionQueries, nil
+}
+
+// byReferenceValue implements sort.Interface for []*firestorepb.Value
+
+type byReferenceValue []*firestorepb.Value
```
tritone, line 126: This could probably take a less generic name (and could def use actual generics, which we will get eventually 😁 )

firestore/collgroupref.go
```diff
@@ -36,3 +45,107 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
...
+
+func (a byReferenceValue) Len() int           { return len(a) }
+func (a byReferenceValue) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a byReferenceValue) Less(i, j int) bool { return compareValues(a[i], a[j]) < 0 }
+
+// QueryPartition provides a Collection Group Reference and start and end split
```
tritone, line 132: Again, it would be helpful to clarify what the user is actually supposed to use this for.

tritone (commented):
Code generally looks fine, would like some more docs to understand better what this is for!

tritone:
Also, just realized there was #3476 previously (integration test there is interesting).

firestore/collgroupref.go
crwilcox: moved to order.go, renamed to be byFirestoreValue also as that is more accurate. It happens to be used for ref vals here.

firestore/integration_test.go
crwilcox: No fixtures to my knowledge, though java does generate a collection with 2 * 128 + 127 docs. I could add this but had concerns that it would slow CI down for folks. There are techniques I could use to try and reuse existing docs so this wouldn't happen on our CI as long as the same project was used (fixed collection name).

crwilcox: Added test but as a separate case

firestore/integration_test.go
```diff
@@ -1601,3 +1601,93 @@ func TestDetectProjectID(t *testing.T) {
...
+	docs, err := colGroup.Documents(ctx).GetAll()
+	if err != nil {
+		t.Fatalf("Did not expect error: %v", err)
+	}
+	if got, want := len(docs), documentCount; got != want {
+		t.Errorf("Unexpected number of documents in collection group: got %d, want %d", got, want)
+	}
+
+	// Get partitions, allow up to 10 to come back, expect less will be returned.
+	partitions, err := colGroup.GetPartitions(ctx, 10)
```
crwilcox, line 1670: Our other tests don't seem to validate the actual partition count, though I'd expect more than one partition. I could soften this to that. With 383 docs, here is what the partitions are:
```
Start:
  End: proj/.../(default)/documents/largeCollection/doc121

Start: proj/.../(default)/documents/largeCollection/doc121
  End: proj/.../(default)/documents/largeCollection/doc232

Start: proj/.../(default)/documents/largeCollection/doc232
  End: proj/.../(default)/documents/largeCollection/doc34

Start: proj/.../(default)/documents/largeCollection/doc34
  End:
```

firestore/collgroupref.go
```diff
@@ -36,3 +46,146 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
...
+		return nil, err
+	}
+	queries := make([]Query, len(qp))
+	for _, part := range qp {
+		queries = append(queries, part.ToQuery())
+	}
+	return queries, nil
+}
+
+// getPartitions returns a slice of QueryPartition objects, describing a start
```
tritone, line 69: `queryPartition` lower-cased.

firestore/collgroupref.go
```diff
@@ -36,3 +46,146 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
...
+
+	// From Proto documentation:
+	// To obtain a complete result set ordered with respect to the results of the
+	// query supplied to PartitionQuery, the results sets should be merged:
+	// cursor A, cursor B, cursor M, cursor Q, cursor U, cursor W
+	// Once we have exhausted the pages, the cursor values need to be sorted in
+	// lexicographical order by segment (areas between '/').
+	sort.Sort(byFirestoreValue(cursorReferences))
+
+	partitionQueries := make([]queryPartition, 0, len(cursorReferences))
```
tritone, line 123: Naming is a bit strange-- these are partitions, not queries right?

firestore/collgroupref.go
```diff
@@ -36,3 +46,146 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
...
+	StartAt string
+
+	// EndBefore is a document reference value, relative to the collection, not
+	// a complete parent path.
+	// Example: "documents/collectionName/documentName"
+	EndBefore string
+}
+
+// ToQuery converts a QueryPartition object to a Query object
+func (qp queryPartition) ToQuery() Query {
```
tritone, line 180: Should this be exported?

firestore/collgroupref.go
```diff
@@ -36,3 +46,146 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
 		},
 	}
 }
+
+// TODO: add a fromProto to compliment toProto for query. With that serialization is
+// very possible. user could also do json.Marshal(toProto())
+
+// GetPartitionedQueries returns a slice of Query objects, each containing a
```
tritone, line 53: Since all the other types/methods are now non-exported, this is the only thing that will go into the public package docs. Is there more to add here? Is there a link to the docs on cloud site for this feature?

firestore/integration_test.go
```diff
@@ -1601,3 +1601,88 @@ func TestDetectProjectID(t *testing.T) {
...
+		deleteBatch.Delete(doc)
+	}
+	batch.Commit(ctx)
+	defer deleteBatch.Commit(ctx)
+
+	// Verify that we retrieve 383 documents for the colGroup (128*2 + 127)
+	colGroup := iClient.CollectionGroup(collectionID)
+	docs, err := colGroup.Documents(ctx).GetAll()
+	if err != nil {
+		t.Fatalf("Did not expect error: %v", err)
```
tritone, line 1659: More useful to put the method called here than "did not expect error".

Also, nice test!

tritone (commented):
Looks good, a couple more small comments. I agree that this surface seems better.

firestore/collgroupref.go
```diff
@@ -36,3 +46,146 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
...
+// must be a positive value and the number of returned partitions may be less
+// than the requested number if providing the desired number would result in
+// partitions with very few documents.
+func (cgr CollectionGroupRef) getPartitions(ctx context.Context, partitionCount int) ([]queryPartition, error) {
+	orderedQuery := cgr.query().OrderBy(DocumentID, Asc)
+
+	if partitionCount <= 0 {
+		return nil, errors.New("a positive partitionCount must be provided")
+	} else if partitionCount == 1 {
+		return []queryPartition{{CollectionGroupQuery: orderedQuery, StartAt: "", EndBefore: ""}}, nil
```
noahdietz, line 80: Nit: I guess this is done for the one-liner syntax, but specifying the empty string values seems unnecessary :)

firestore/collgroupref.go
```diff
@@ -36,3 +46,146 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
...
+	}
+	cursorReferences := make([]*firestorepb.Value, 0, partitionCount)
+	iter := cgr.c.c.PartitionQuery(ctx, pbr)
+	for {
+		cursor, err := iter.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+
```
noahdietz, line 109: Nit: extraneous line break

firestore/collgroupref.go
```diff
@@ -36,3 +46,146 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
...
+			return nil, fmt.Errorf("GetPartitions: %v", err)
+		}
+		cursorReferences = append(cursorReferences, cursor.Values...)
```
noahdietz, line 112: Nit: Is `cursor` a proto message? If so, prefer using the accessor `cursor.GetValues()`. If not, ignore this.

crwilcox: It is. Thank you!

firestore/collgroupref.go
```diff
@@ -36,3 +46,146 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
...
+		partitionQueries = append(partitionQueries, qp)
+		previousCursor = cursorRef
+	}
+
+	// In the case there were no partitions, we still add a single partition to
+	// the result, that covers the complete range.
+	lastPart := queryPartition{
+		CollectionGroupQuery: orderedQuery,
+		StartAt:              "",
+		EndBefore:            "",
```
noahdietz, line 147: Nit: Again, I think these are unnecessary, but maybe y'all want to be explicit with these offsets.

firestore/collgroupref.go
crwilcox: @noahdietz  we have a few others, I won't change them as part of this pr, but if this is generally true I can make alterations
image

noahdietz: Using the accessor protects from accidental referencing of `nil` when a non-primitive field is unset ([code for this accessor](https://github.com/googleapis/go-genproto/blob/b9e0b3197ced/googleapis/firestore/v1/query.pb.go#L525-L530)). In this case,  expanding a `nil` slice is handled safely by the language (like `len` on a `nil` slice), so it's not a huge deal. But with messages (which are represented as struct pointers) and `optional` primitives (which are represented as primitive pointers), it can be a foot gun. 

Unless you are checking for presence or setting a value, its best practice to use the accessor to retrieve the value.

firestore/collgroupref.go
crwilcox: At the current point we don't have any docs really, other than rest docs, which I am not sure will bring clarity: https://firebase.google.com/docs/firestore/reference/rest/v1beta1/projects.databases.documents/partitionQuery

noahdietz (approved):

tritone (approved):

PR #4209, opened 2021-06-03
chore: skip protos with broken imports

codyoss:


noahdietz (approved):
I'm guessing the issue is related to its dependency on grafeas and possibly another odd import `cloud/containers/workflow/vulnerability/proto/package.proto`?

codyoss:
@noahdietz yup!

PR #4218, opened 2021-06-04
feat(datastore): save enums as string repr

### LINKED ISSUE #4196: datastore: load and write string representation of int enum
I have issues loading enums from/saving enums to datastore.

I would be great to use the `String` method if it exists on enums ? Or give a property like `datastore:",string"` to specify we want the string repr.
Maybe there is a solution I didn't found ?

I make it working by changing my enum to string but loose good (automatic) enum checks on my API.

**Additional context**
For example:
```
//go:generate enum Role -text -nosql -transform=kebab

type Role int

const (
	Admin Role = iota
	Standard
)

type User struct {
	Name       string
	Role        Role
}
```
This will save `0` for `admin` and `1` for `standard`.

scorsi:
Add the hability to save string representation of enums when Stringer interface is implemented.

There is just one issue for now: actually if a struct (not an enum) implements the Stringer interface, the save method will also save its string representation instead of a Datastore Entity (as expected).

The solution would be to add a `stringer` or `string` tag to allow or not using the Stringer method ? Like:
```go
type MyStruct struct {
  MyNestedStruct                   MyNestedStruct
  AFieldInStringerRepresentation   AFieldInStringerRepresentation   `datastore:",stringer"`
}
```
I would like to have opinions of contributors about this if possible :) 

Should fixes #4196

### DIFF
--- a/datastore/save.go
+++ b/datastore/save.go
@@ -76,6 +76,8 @@ func reflectFieldSave(props *[]Property, p Property, name string, opts saveOpts,
 		p.Value = x.In(time.UTC)
 		*props = append(*props, p)
 		return nil
+	case fmt.Stringer:
+		p.Value = x.String()
 	default:
 		switch v.Kind() {
 		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
--- a/datastore/save_test.go
+++ b/datastore/save_test.go
@@ -34,6 +34,7 @@ func TestInterfaceToProtoNil(t *testing.T) {
 		(*float64)(nil),
 		(*GeoPoint)(nil),
 		(*time.Time)(nil),
+		(*fmt.Stringer)(nil),
 	} {
 		got, err := interfaceToProto(in, false)
 		if err != nil {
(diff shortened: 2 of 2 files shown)


scorsi:
@googlebot I signed it!

crwilcox:
@scorsi one concern I have: If a current user has an enum with a stringer, it is saving the enum as the integer value. I may misunderstand but it seems this may affect how their data is stored?

scorsi:
> @scorsi one concern I have: If a current user has an enum with a stringer, it is saving the enum as the integer value. I may misunderstand but it seems this may affect how their data is stored?

This is why I'm proposing to add a optin field "stringer" or similar in the tag to activate or not the string representation using stringer. It won't change any actual behavior as we must optin.

zamicol:
@crwilcox, why can't I just save a specific type to the database using the PropertyLoadSaver interface?  All I need is for Save() to support individual fields and I'm golden.  

I'll also give you this real world example:  

```
type Hex []byte

// String implements fmt.Stringer.
func (t Hex) String() string {
	return fmt.Sprintf("%X", []byte(t))
}
```

Hex is stored as []byte (which is correct).

---

I want the "enum" `Use` to be stored in the database as string.  This is how I expected it to work:

```
type Use int64

func (u *Use) Save() (ps []datastore.Property, err error) {
	return nil, nil
}

func (u *Use) Load(p []datastore.Property) error {
	s, ok := p[0].Value.(string)
	if ok {
		*u = Use.Parse(s)
	}
	return nil
}
```

Because that doesn't work on types and only structs, I'm using a shim struct to save to datastore, so `DatastoreMyStruct := MyStruct.ToDatastore()" and then saving that.  I have to do that on each struct that uses my enum.

PR #4219, opened 2021-06-04
chore(bigtable/bttest): remove half-baked instances support

dragonsinth:
A couple of years ago, support for DeleteInstance was added. However, this never really went anywhere; there's no way to actually create emulated instances or manage them.  Can we remove this cruft, and simplify unimplemented methods via modern protoc generated stubs?


crwilcox:
@dragonsinth could you explain the problem this is solving? How does this change the user experience?

And thanks for your recent contributions to bttest. They are appreciated, but I want to make sure I understand your goal here :)

dragonsinth:
This is just code hygiene, pure and simple!  It solves no particular problem, just removes some dead code and test to make maintenance easier going forward.

This is assuming no one has any plans to implement sophisticated instance management on the emulator.  Since this was added in 2019 figured it was probably a dead effort.

crwilcox:
@igorbernstein2 any idea if this was something on the backlog?

bigtable/bttest/instance_server.go
```diff
@@ -16,118 +16,26 @@ package bttest
...
-}
-
-func (s *server) SetIamPolicy(ctx context.Context, req *iampb.SetIamPolicyRequest) (*iampb.Policy, error) {
-	return nil, errUnimplemented
-}
-
-func (s *server) TestIamPermissions(ctx context.Context, req *iampb.TestIamPermissionsRequest) (*iampb.TestIamPermissionsResponse, error) {
-	return nil, errUnimplemented
+func (s *server) TestIamPermissions(_ context.Context, _ *iampb.TestIamPermissionsRequest) (*iampb.TestIamPermissionsResponse, error) {
+	return nil, status.Errorf(codes.Unimplemented, "method TestIamPermissions not implemented")
```
jimfulton, line 40: Why not move these placeholders to inmem.go and do away with this file entirely.  These are really just completing (for some definition of completing) BigtableInstanceAdminServer.

jimfulton:
FWIW, I endorse removing dead code. Every line of code is a liability that needs to earn it's keep.  Realzy code can be added later if someone decides to implement instances.

bigtable/bttest/instance_server.go
dragonsinth: 🤷  that file seems pretty cluttered already

crwilcox:
While going over the backlog of PRs I reviewed this again. Were these by chance here to satisfy an interface?

dragonsinth:
Perhaps-- but the interface isn't exposed over grpc so I'm not sure what the utility of it would be.  At least in OSS.  There might be Google internal refs

codyoss:
@googleapis/api-bigtable Let's decide if we want to move forward with this or close out the issue.

PR #4221, opened 2021-06-04
test: display environment variables when encountering an error

florinpapa:


codyoss (changes requested):
Logs are publicly accessible.

florinpapa:
Thanks @codyoss for the review!

Issue #4146, opened 2021-05-21
bigquery: Can't send NULL as bigquery.QueryParameter


prochac:
I have no clue how to send NULL as an bigquery.QueryParameter when using bigquery.Client#Query method.
All I consider intuitive solution doesn't work.
I believe the best how to express my problem is the code. Please, see below.

**Code**

e.g.
```go
package main

import (
	"context"
	"fmt"
	"strings"
	"testing"
	"time"

	"cloud.google.com/go/bigquery"
	"github.com/stretchr/testify/require"
	"google.golang.org/api/option"
)

func TestFoo(t *testing.T) {
	ctx := context.Background()

	var (
		project   string 
		dataset   string 
		tableName string 
		filename  string 
	)

	c, err := bigquery.NewClient(ctx, project, option.WithCredentialsFile(filename))
	require.NoError(t, err)
	table := c.Dataset(dataset).Table(tableName)
	_ = table.Delete(ctx)

	// Create Table with one NULLABLE and one NOT NULL column, in SQL language.
	// Or one required and one not required, in GBQ language.
	err = table.Create(ctx, &bigquery.TableMetadata{
		Schema: bigquery.Schema{
			{Name: "date", Type: bigquery.TimestampFieldType, Required: true},
			{Name: "null_date", Type: bigquery.TimestampFieldType, Required: false},
		},
		Description: "Foo",
	})
	require.NoError(t, err)

	// Use data loader with JSON as a data source.
	rs := bigquery.NewReaderSource(strings.NewReader(
		`{"date":"` + time.Now().Format(time.RFC3339Nano) + `","null_date":null}`,
	))
	rs.FileConfig.SourceFormat = bigquery.JSON
	job, err := table.LoaderFrom(rs).Run(ctx)
	require.NoError(t, err)
	status, err := job.Wait(ctx)
	require.NoError(t, err)
	require.NoError(t, status.Err())

	// Try to use SQL insert.
	fullTableName := fmt.Sprintf("%s.%s.%s", project, dataset, tableName)
	q := c.Query(`INSERT INTO ` + fullTableName + ` VALUES(?,?)`)
	q.Parameters = []bigquery.QueryParameter{
		{Value: time.Now()},
		// leads to following error:
		//	bigquery: nil parameter
		//{Value: nil},

		// leads to following error:
		//	googleapi: Error 400: Value has type STRUCT<> which cannot be inserted into column null_date, which has type TIMESTAMP at [1:57], invalidQuery
[...]

PR #4225, opened 2021-06-07
feat(bigquery): support nullable params and geography params

shollyman:
Details:

* Support the passing of Null<type> values for query parameters.

* Add support for GEOGGRAPHY param type, which was added after the initial
      GEOGRAPHY type support launched.

Fixes: https://github.com/googleapis/google-cloud-go/issues/4146


hongalex (approved):

shollyman:
Revamping this currently, value propagation had some issues

tbpg (approved):
Looks good to me, but would prefer two smaller PRs in the future. :smile:

shollyman:
> Looks good to me, but would prefer two smaller PRs in the future. 😄

Reasonable request.  :) 

My rationale was that the delta from geography was small, and touched all the same areas so I lumped them together this time.

PR #4235, opened 2021-06-08
feat(gkeconnect/gateway): start generating apiv1beta1

codyoss:


noahdietz (approved):

codyoss:
@noahdietz For some reason the import for [gatewaypb](https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/gkeconnect/gateway/v1beta1) is not being generated into the imports. Any ideas why? Looks like the RPC don't deal with types from that package but just [httpbody](https://pkg.go.dev/google.golang.org/genproto@v0.0.0-20210608205507-b6d2f5bf0d7d/googleapis/api/httpbody). Maybe that is why?

noahdietz:
Oop yes this was fixed in v0.20.2 of the generator. Turns out the proto has no request/response messages of its own (they are are `google.api.HttpBody`) and the generator apparently was only adding the API's pb import when it resolved request/response messages. Updating the generator in the pipeline should fix it. Sorry about that.

codyoss:
No worries, thanks for the info!

Issue #3823, opened 2021-03-16
logging: default bundler limits doesn't match official API


loburm:
**Client**

Logging

**Code**

That's a limit that we see inside of package:
```go
	// DefaultEntryCountThreshold is the default value for the EntryCountThreshold LoggerOption.
	DefaultEntryCountThreshold = 1000

	// DefaultEntryByteThreshold is the default value for the EntryByteThreshold LoggerOption.
	DefaultEntryByteThreshold = 1 << 20 // 1MiB

	// DefaultBufferedByteLimit is the default value for the BufferedByteLimit LoggerOption.
	DefaultBufferedByteLimit = 1 << 30 // 1GiB
```

I'm not 100% sure what is the difference between DefaultEntryByteThreshold and DefaultBufferedByteLimit, but official [API](https://cloud.google.com/logging/quotas) has a different limits:

> Size of a log entry 256 KB
> Size of an entries.write request 10 MB

nicoleczhu:
good pt, i think we are just using bundler defaults [see here](https://github.com/googleapis/google-api-go-client/blob/e00d87e52f2e2d0189045089be07de3dc043b865/support/bundler/bundler.go#L26)

ill get in a pr to raise these defaults to something that makes more sense given actual api limits

PR #4247, opened 2021-06-11
feat(logging)!: increase DefaultEntryByteThreshold to 8Mb

nicoleczhu:
Fixes: #3823 

DefaultEntryByteThreshold = `1 << 20 // 1MiB` ===> `1 << 23 // 8MiB` since size of an entries.write request	can be up to 10 MB. 

`DefaultBufferedByteLimit` is unchanged as that's more a cap on memory utilization.


nicoleczhu:
@loburm mind sanity checking this? I think changing default thresholds may produce unforeseen results, so we might want to release this with preview

simonz130 (approved):

loburm:
In overall it looks good to me.

nicoleczhu:
@codyoss can I get your pointers, this PR is marked as a breaking change in go116 because: 
```
The following breaking changes were found:
cloud.google.com/go/logging:
- DefaultEntryByteThreshold: value changed from 1048576 to 8388608
```

It's technically not since we're increasing the threshold, not decreasing it. How do I override this, or fix this test? Thank you!

codyoss:
It is a breaking change. I silly example is if someone wrote a test validate its value, or grabbed an element of an array at this index, which could panic if the array was not large enough. That being said I with proper usage I think this is probably fine ... @tbpg would you agree?

To get around this check the commit message needs to be formatted in a special manner: https://github.com/googleapis/google-cloud-go/blob/00140bde9fc18d3ac21ea717df4aff8ed8db2262/internal/apidiff/apidiff.go#L223-L238

Also, the subsequent release would need to be manual overridden as it will propose a v2.

PR #4250, opened 2021-06-11
testing(bigquery): skip error test, error responses in flux

shollyman:
Related: https://github.com/googleapis/google-cloud-go/issues/4248


shollyman:
Seems this was transient, closing instead of proceeding.

PR #4251, opened 2021-06-12
feat(pubsub/pstest): add channel to support user-defined publish responses

hongalex:
Taken from the [Java fake publisher service](https://github.com/googleapis/java-pubsub/blob/master/google-cloud-pubsub/src/test/java/com/google/cloud/pubsub/v1/FakePublisherServiceImpl.java#L39), these changes improves testing by allowing users to specify `PublishResponses` that the fake server will return for `Publish` calls.

This is implemented with a buffered publishResponses channel, that users can add to by disabling automatic publish responses (the current behavior) and calling `AddPublishResponse`.

```go
srv.SetAutoPublishResponse(false)
srv.AddPublishResponse(&pb.PublishResponse{
  MessageIds: []string{"1"}
}, nil)
srv.Publish("projects/p/topics/t", []byte("some-message"), nil)
```

Lastly, the default buffered channel size is 100 (an arbitrary number suitable for testing). If the size needs to be increased, or you want to reset the publish queue, call `ResetPublishResponses`.

The changes in this PR will be used to help test publisher flow control.


pubsub/pstest/fake.go
```diff
@@ -613,6 +658,15 @@ func (s *GServer) Publish(_ context.Context, req *pb.PublishRequest) (*pb.Publis
 	if top == nil {
 		return nil, status.Errorf(codes.NotFound, "topic %q", req.Topic)
 	}
+
+	if !s.autoPublishResponse {
+		r := <-s.publishResponses
```
shollyman, line 663: Unclear if it matters, but do you want blocking behavior here, or should it emit an error in the case of some kind of timeout via a select loop or similar?

shollyman (approved):
LGTM, with a minor caveat which may or may not be relevant given my lack of experience with the p/s fake.

pubsub/pstest/fake.go
hongalex: I want blocking behavior, but could be convinced to add a timeout/error in the future.

kamalaboulhosn (approved):

PR #4253, opened 2021-06-13
fix(pubsub): off-by-one error leading to bundlers never being freed

mzanibelli:
### Context

When adding an item to the publish scheduler with message ordering enabled, a new bundler is added for this particular key. 

The publisher keeps an internal state of the current bundlers and is supposed to remove a bundler when all of its items have been processed. 

To know if the bundler is ready for cleanup, an internal counter is incremented on Add() and decremented on handle(). 

### Problem

This counter is erroneously initialized at 2, leading to never reaching 0 and the bundler never beeing freed. 

### Consequence

This causes a memory leak when a high number of different ordering keys are used: the internal bundlers list becomes huge and can keep growing - reaching millions of entries.

### Remediation

Don't execute the increment statement upon bundle initialization since the internal counter is already at value 1.


mzanibelli:
@googlebot I signed it!

mzanibelli:
@googlebot I signed it!

mzanibelli:
@googlebot I signed it!

hongalex (approved):
Thanks for submitting this PR, this is a good catch!

hongalex:
@mzanibelli So this fix actually uncovered an issue with our other methods (specifically `FlushAndStop`), which tries to [iterate over bundlers](https://github.com/googleapis/google-cloud-go/blob/59420d965d2ab6e87fe07f8cb832c2ee6acf6d7c/pubsub/internal/scheduler/publish_scheduler.go#L156) and Go detects a data race with this and the [deletion of the bundlers](https://github.com/googleapis/google-cloud-go/blob/59420d965d2ab6e87fe07f8cb832c2ee6acf6d7c/pubsub/internal/scheduler/publish_scheduler.go#L122). I thought locking the mutex in FlushAndStop would fix this, but this causes other tests in `publish_scheduler_test` to fail. I don't have time to take a look at this now, but just wanted to give you a heads up as to why the CI is failing.

mzanibelli:
Hello! Indeed I did not run the tests with the race detector enabled.
The problem seems deeper. I tried implementing `sync.Map` usage (for both bundlers and key counters) but it leads to other race problems let alone performance problems. I think some work is needed at the architecture level where a way to avoid updating the scheduler properties inside the bundler's handler would streamline the whole thing.
Let me know if I can help! Thanks for looking into this.

mzanibelli:
Hey! I have attempted [something](https://gist.github.com/mzanibelli/59b019bb3c1bca4ac2ad29648ba65831) which is not very clean but the general idea is: the race being on `s.bundlers` iteration / deletion, when we flush (and stop) we don't need to perform the deletion at all and we can just cleanup the whole state just before returning from `Flush*`.

Now here is the catch: not sure what happens if we `Add` during a simple `Flush` because the guard provided by `s.done` during `FlushAndStop` is not effective. I think it should be ok since either the bundler is still here, either everything is clean so we don't have a corrupted state but be careful.

It's not very clean yet and I must admit I don't like this solution, but tests pass with the race detector so I might aswell mention it to you guys.

hongalex:
I have a very rough draft of replacing everything with sync.Map and it seems to be fine (as in the tests pass). From some initial benchmarking, replacing map with sync.Map hasn't affected unordered publish performance. I'll send out a PR of this later today after benchmarking with ordered keys as well.

mzanibelli:
I retried the `sync.Map` implementation just out of curiosity and you are right! I must have been sleeping the other day but I had `TestPublishScheduler_DoesntRaceWithPublisher` failing if I remember correctly. I still look forward to checking the way you handled a specific corner case that leaves me doubtful, but anyways good job! It seems one of the rare occasions `sync.Map` would be appropriate.

mzanibelli:
Better solution [here](https://github.com/googleapis/google-cloud-go/pull/4282).

Issue #4167, opened 2021-05-24
storage: nil pointer crash in writer goroutine after CloseWithError


dt:
**Client**

Storage

**Code**

Extracted/adapted from a more complex reproduction, but more or less:
```go
   for {
	c := storage.NewClient(...)
        w := c.NewWriter(...)
        w.CloseWithError(io.EOF)
        c.Close(...)
   }
```

**Expected behavior**

The write is aborted.

**Actual behavior**

The process crashes:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x375c91a]
...
cloud.google.com/go/storage.(*Writer).open.func1(0xc03023bb00, 0xc00dea2380, 0xc0381d70e0, 0xc00cac1840, 0x1, 0x1)
        /go/src/.../vendor/cloud.google.com/go/storage/writer.go:131 +0xfa fp=0xc0040f0fb0 sp=0xc0040f0e30 pc=0x375c91a
runtime.goexit()
        /usr/local/go/src/runtime/asm_amd64.s:1374 +0x1 fp=0xc0040f0fb8 sp=0xc0040f0fb0 pc=0x4c4ae1
created by cloud.google.com/go/storage.(*Writer).open
        /go/src/.../vendor/cloud.google.com/go/storage/writer.go:118 +0x412
```

**Additional Information**

It looks like `Writer` spawns [this goroutine](https://github.com/googleapis/google-cloud-go/blob/storage/v1.15.0/storage/writer.go#L118) to handle the upload side of the pipe. Inside this goroutine it is reading fields of `storage.Client` such as `Client.raw` on the line mentioned in the above crash stack trace:
https://github.com/googleapis/google-cloud-go/blob/825ddd692363eb2dd8cd253cc5976867e432f547/storage/writer.go#L131

Looking at `Client.Close()`, it nils out `raw` when the client is closed: https://github.com/googleapis/google-cloud-go/blob/825ddd692363eb2dd8cd253cc5976867e432f547/storage/storage.go#L167

So it looks like the caller has called `Client.Close()` while while the writer goroutine was still running. However reviewing the calling the code in my application, the the `Writer` is _always_ closed -- via either `Close` or a `CloseWithError` in a defer --  before the function that created it returns to the function which created and would close the `Client`. 

So it appears that the writer was somehow still running after it was closed, which meant it was then poised to crash when the client was closed out from under it?

[...]

PR #4255, opened 2021-06-14
fix(storage): fix Writer.ChunkSize validation

tritone:
Since w.opened is set to true before this validation currently,
subsequent calls to Writer.Close will deadlock. There is no
reason not to do this check first before opening the pipe.

Thanks to @dt for pointing this out.

Updates #4167


codyoss (approved):
LGTM

Issue #2492, opened 2020-06-18
bigtable: ability to set deadlines for cbt commands


yananchang:
**Is your feature request related to a problem? Please describe.**
There is no way to set deadlines for cbt commands such as 'cbt deleteallrows'.

**Describe the solution you'd like**
To provide ways such as params to set deadlines.

**Describe alternatives you've considered**
Several partial delete to make sure falling within the default deadline; or delete and recreate the table.

**Additional context**
Please update if other workarounds are available.

PR #4276, opened 2021-06-17
feat(bigtable/cmd/cbt): Add a timeout option

jimfulton:
Fixes: #2492


jimfulton:
This now has a test! And includes some refactoring that makes some kinds of tests easier.

bigtable/cmd/cbt/testing.go
```diff
@@ -0,0 +1,45 @@
...
+
+package main
+
+import (
+	"bytes"
+	"io"
+	"os"
+)
+
+func captureStdout(f func()) string {
```
garye, line 25: Document this? Wasn't immediately clear to me what f was for.

garye (approved):

bigtable/cmd/cbt/testing.go
jimfulton: Done.

Note that here, we're mainly preventing spurious print output.  In other branches/PR, we're making assertions about captured stdout.

bigtable/cmd/cbt/cbt.go
```diff
@@ -581,7 +607,7 @@ func doCount(ctx context.Context, args ...string) {
 	if len(args) != 1 {
 		log.Fatal("usage: cbt count <table>")
 	}
-	tbl := getClient(bigtable.ClientConfig{}).Open(args[0])
+	tbl := getTable(bigtable.ClientConfig{}, args[0])
```
crwilcox, line 610: I am having some difficulty fully understanding the need of this refactoring and its relation to the timeout change. Can you describe it a bit?

jimfulton: I wanted to test that a proper context (with  a timeout) was passed to `ReadRows`.  

`getTable` returns a `tableLike` object, which simply has `ReadRows`. (I think I expand it in another PR, IIRC :) )

Normally, `getTable` calls `getClient(clientConf).Open(tableName)`, but if the global `table` is set, it returns that.

`timeout_test.go` sets `table` to a mock that captures the context passed to `ReadRows` so we can assert that it was set correctly.

crwilcox (approved):

PR #4281, opened 2021-06-17
fix(pubsublite): make SubscriberClient.Receive identical to pubsub

tmdiep:
* Ensures that pscompat.SubscriberClient.Receive and pubsub.Subscription.Receive have identical interfaces.
* Adds examples for how to declare common interfaces for pubsublite and pubsub.
* Amended docs to point to these examples.
* Other misc doc cleanups.


codyoss (approved):
LGTM, thanks!

tmdiep:
Thanks for reviewing!

PR #4282, opened 2021-06-17
fix(pubsub): fix memory leak issue in publish scheduler

### LINKED ISSUE #4288: pubsub: Memory growth observed while sending messages with ordering keys
**Client**

PubSub v1.10.1

**Environment**

Ubuntu 20.04

**Go Environment**

go 1.16

**Expected behavior**

memory growth should not be observed while sending messages using transient `ordering keys` (ordering keys which have a limited life span) 

**Actual behavior**

Memory growth is observed while sending messages using `Topic.Publish` method with ordering keys. Attaching the proof report image. It looks like the maps `s.bundlers` and `s.outstanding` of PublishScheduler are growing. 

Should the increment of outstanding `s.outstanding[key]++` be done when the `key` is found in the map only as it is already initialised to `1` when it is not found in the map. 
https://github.com/googleapis/google-cloud-go/blob/71c046962a49468d8759113638a314326f4256a7/pubsub/internal/scheduler/publish_scheduler.go#L147 

image

hongalex:
This PR builds off the work completed by @mzanibelli in #4253. A detailed description was written there, but essentially we were instantiating outstanding messages to be 2 instead of 1, leading bundles to never be freed, which leads to memory leaks when using large number of ordering keys. This switches the underlying implementation to a `sync.Map` to properly synchronize access to the `bundlers` and `outstanding` maps, allowing the user to safely call `Flush` and `FlushAndStop` without race conditions.

Running benchmarks against both the unordered and ordered publish benchmarks reveals no performance degradation. Below are results from the ordered case, publishing 100k 10kb messages, using different ordering keys.

prev(using built-in maps)
```
BenchmarkPublishThroughput-12    	       1	68024964685 ns/op	  14.70 MB/s	5010368568 B/op	35265687 allocs/o
BenchmarkPublishThroughput-12    	       1	68054813422 ns/op	  14.69 MB/s	5002350416 B/op	35229700 allocs/op
```

now(using `sync.Map`)
```
BenchmarkPublishThroughput-12    	       1	68124099897 ns/op	  14.68 MB/s	5023081976 B/op	36219292 allocs/op
BenchmarkPublishThroughput-12    	       1	68060299607 ns/op	  14.69 MB/s	5027381288 B/op	36268891 allocs/op
```

In addition, a test was added to `topic_test` to verify the correct behavior of calling `PublishScheduler.Add` while a `Flush` is in flight.

This also fixes a small issue when using the pstest fake's `AddPublishResponse` method which allows you to call `Publish` before calling `AddPublishResponse` by removing the lock acquisition.

Fixes #4288


pubsub/internal/scheduler/publish_scheduler.go
```diff
@@ -106,20 +104,23 @@ func (s *PublishScheduler) Add(key string, item interface{}, size int) error {
...
 	if !ok {
-		s.outstanding[key] = 1
+		s.outstanding.Store(key, 1)
 		b = bundler.NewBundler(item, func(bundle interface{}) {
 			s.workers <- struct{}{}
 			s.handle(bundle)
 			<-s.workers
 
 			nlen := reflect.ValueOf(bundle).Len()
 			s.mu.Lock()
```
mzanibelli, line 118: So that's it! You left the lock between Add() and handle() so you don't load non-existing keys from the map inside the else clause. This gave me a hard time and I thought I was doing something wrong when I tried this.

kamalaboulhosn (approved):

Issue #1717, opened 2020-01-04
bigtable: add keys-only argument to cbt read to view keys without large column values


jkleckner:
When examining tables using `cbt`, it would be very handy to be able to list keys only without potentially very large cell values.

This could be done by adding an argument `keys-only` for the `read` command to suppress printing of cell values.

Thus, something like:
```
cbt --project test_project --instance test_instance \
    read \
    test_table \
    count=10 \
    keys-only=true \
    start=$'start_prefix\x1Fsub_prefix'
```

Pega88:
+1 on this. The only way around it now is to set the column filter with a very minimal column family/column qualifier, which is still reading more data than needed. passing an empty column filter does also not give the desired result (gives no result)

DeaconDesperado:
:+1:

limjix:
+1 on this too.

jimfulton:
I wonder if it would be better to be able to set a maximum output width for values, with output greater than the max width elided.

That way, you'd still see the smaller values and a sample of the larger ones.

scordata:
+1

telpirion:
This feature has been implemented in `cbt` version 0.12.0. (The `cbt` tool has also moved to another [repo](https://github.com/googleapis/cloud-bigtable-cbt-cli).)

PR #4286, opened 2021-06-18
feat(bigtable/cmd/cbt): Add keys-only option to lookup and read to avoid fetching and displaying cell values

jimfulton:
Fixes: #1717


jimfulton:
The sample failure seems to be unrelated to the code. It cites code that doesn't exist in the repository. :)

```
# contrib.go.opencensus.io/exporter/stackdriver
/go/pkg/mod/contrib.go.opencensus.io/exporter/stackdriver@v0.13.8/stats.go:459:19: end.Sub(start).Milliseconds undefined (type time.Duration has no field or method Milliseconds)
```

Does anyone have any idea how to address it?

bigtable/cmd/cbt/cbt.go
```diff
@@ -988,49 +994,91 @@ func doListClusters(ctx context.Context, args ...string) {
...
+	}
+
+	opt, keysOnly, err := getDataFilter(parsed)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var opts []bigtable.ReadOption
+	if opt != nil {
+		opts = append(opts, opt)
```
crwilcox, line 1067: Should this be `opts = []bigtable.ReadOption{opt}`?

bigtable/cmd/cbt/cbt.go
```diff
@@ -988,49 +994,91 @@ func doListClusters(ctx context.Context, args ...string) {
...
 	}
 
 	table, row := args[0], args[1]
-	tbl := getClient(bigtable.ClientConfig{AppProfile: parsed["app-profile"]}).Open(table)
+	tbl := getTable(
+		bigtable.ClientConfig{AppProfile: parsed["app-profile"]},
+		table)
 	r, err := tbl.ReadRow(ctx, row, opts...)
```
crwilcox, line 1074: It seems that opts can only have a single entry in it, if this is the case, what is the reason for unpacking it? Is this to avoid having to nil-check here?

bigtable/cmd/cbt/cbt.go
```diff
@@ -962,49 +985,91 @@ func doListClusters(ctx context.Context, args ...string) {
...
 		filters = append(filters, columnFilters)
 	}
 
-	if len(filters) > 1 {
-		opts = append(opts, bigtable.RowFilter(bigtable.ChainFilters(filters...)))
-	} else if len(filters) == 1 {
-		opts = append(opts, bigtable.RowFilter(filters[0]))
+	var filter bigtable.Filter = nil
+	var option bigtable.ReadOption
+	if len(filters) > 0 {
```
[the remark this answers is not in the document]
jimfulton: Sure, at the cost of more checks, or duplicating the  `bigtable.RowFilter` call. :shrug: 
The original code duplicated the `bigtable.RowFilter` call in the size checks.

bigtable/cmd/cbt/cbt.go
jimfulton: No, because `opt` may be nil.

bigtable/cmd/cbt/cbt.go
jimfulton: `opts` can have 0 or 1 entries in it.  So the unpacking handles both cases.  Of course, this is the original code, which I tried not to disturb (aside from factoring out the filtering logic, which I now kind of regret :smiley: ).

bigtable/cmd/cbt/cbt.go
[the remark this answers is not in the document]
jimfulton: I got rid of the nested if by reverting to the logic in the original code.

bigtable/cmd/cbt/cbt.go
jimfulton: > factoring out the filtering logic, which I now kind of regret

Well, no. Factoring out the filtering code was a great idea :smiley:, because it was repeated for read and lookup.

bigtable/cmd/cbt/cbt.go
crwilcox: Can it though? (line 1063)

jimfulton: Oh, I misunderstood your question.  I thought you were referring to line 1062.  Yes, `opts = []bigtable.ReadOption{opt}` is better than using `append`. I'll change that.

jimfulton: done

jimfulton:
Kokoro failed with the message "Tool Failed" and no log.  IDK how to force a build other than by pushing a change.

I guess I'll wait for the branch to get out of date so I can merge the changes. :)

bigtable/cmd/cbt/testing.go
```diff
@@ -48,3 +54,28 @@ func captureStdout(f func()) string {
 	w.Close()
 	return <-outC
 }
+
+func assertEqual(t *testing.T, got, want interface{}, opts ...cmp.Option) {
```
codyoss, line 58: nit: Please avoid adding asset functions and validate the function inline. There are a couple different options: just use the stdlib and t.Error, make use of testutil, or cmp.Diff/Equal may be worth using.

https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

bigtable/cmd/cbt/testing.go
```diff
@@ -48,3 +54,28 @@ func captureStdout(f func()) string {
...
+	if !testutil.Equal(got, want, opts...) {
+		_, fpath, lno, ok := runtime.Caller(1)
+		if ok {
+			_, fname := filepath.Split(fpath)
+			t.Errorf("%s:%d: Didn't match:\nGot: %s\nExpected: %s",
```
codyoss, line 63: nit: usually we use the words got/want. I am not sure if the extra newline formatting is needed here. See above link or reference other test files in the package for some examples.

jimfulton: @codyoss I'm not sure what's being requested.  I'd be happy to change this to specifically use the got/want language, however, in the previous comment above, you seemed to be suggesting getting rid of the assert function altogether.  Honestly, I'd rather keep it, as I think it avoids a lot of boilerplate.

WRT newlines, I found them helpful for my use cases, as the values being compared were often multi-line text.

telpirion:
I'm going to merge this commit to a dev branch so that I can pull it over to [the dedicated `cbt` repo](https://github.com/googleapis/cloud-bigtable-cbt-cli). After this is merged, we can delete CBT from this repo!

PR #4290, opened 2021-06-21
feat(internal/godocfx): add support for other modules

tbpg:
I chose to process all modules at the same time, rather than in separate
invocations of godocfx, so we wouldn't have to make duplicate requests
to the module index.

The docs.metadata generation will probably need to be updated in a
future PR to support different final URL destinations.

### DIFF
--- a/internal/godocfx/index.go
+++ b/internal/godocfx/index.go
@@ -28,7 +28,7 @@ import (
 
 // indexer gets a limited list of entries from index.golang.org.
 type indexer interface {
-	get(prefix string, since time.Time) (entries []indexEntry, last time.Time, err error)
+	get(prefixes []string, since time.Time) (entries []indexEntry, last time.Time, err error)
 }
 
 // indexClient is used to access index.golang.org.
@@ -49,7 +49,7 @@ type indexEntry struct {
 // versions since the given timestamp.
 //
 // newModules stores the timestamp of the last successful run with tSaver.
-func newModules(ctx context.Context, i indexer, tSaver timeSaver, prefix string) ([]indexEntry, error) {
+func newModules(ctx context.Context, i indexer, tSaver timeSaver, prefixes []string) ([]indexEntry, error) {
 	since, err := tSaver.get(ctx)
 	if err != nil {
 		return nil, err
@@ -61,7 +61,7 @@ func newModules(ctx context.Context, i indexer, tSaver timeSaver, prefix string)
 	for {
 		count++
 		var cur []indexEntry
-		cur, since, err = i.get(prefix, since)
+		cur, since, err = i.get(prefixes, since)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


internal/godocfx/main.go
```diff
@@ -58,42 +58,43 @@ func main() {
...
+				modPath = parts[0]
+				version = parts[1]
 			}
-			modPath = parts[0]
-			version = parts[1]
-		}
-		modPath = strings.TrimSuffix(modPath, "/...") // No /... needed.
-		mods = []indexEntry{
-			{
+			modPath = strings.TrimSuffix(modPath, "/...") // No /... needed.
```
shollyman, line 93: Unclear to me end to end how module strings are moving through the system, do you ever need to sanitize trailing comments?

internal/godocfx/index_test.go
```diff
@@ -49,7 +49,7 @@ func (f *fakeTS) put(context.Context, time.Time) error {
 func TestNewModules(t *testing.T) {
 	ic := fakeIC{}
 	ts := &fakeTS{}
-	entries, err := newModules(context.Background(), ic, ts, "cloud.google.com")
+	entries, err := newModules(context.Background(), ic, ts, []string{"cloud.google.com"})
```
shollyman, line 52: Maybe include an entry with explicit versioning and maybe something that exercises the filter paths?

internal/godocfx/index.go
```diff
@@ -28,7 +28,7 @@ import (
 
 // indexer gets a limited list of entries from index.golang.org.
 type indexer interface {
-	get(prefix string, since time.Time) (entries []indexEntry, last time.Time, err error)
+	get(prefixes []string, since time.Time) (entries []indexEntry, last time.Time, err error)
```
shollyman, line 31: Are overlapping prefixes an issue that may need consideration?

shollyman (approved):
My comments are largely borne of lack of familiarity with the project, feel free to ignore if they're irrelevant.

internal/godocfx/index.go
tbpg: No, I don't think so. We won't process duplicate modules because we don't process each prefix independently -- if a module matches any prefix, it's included.

internal/godocfx/index_test.go
tbpg: This test really just makes sure the paging logic works for the index. It doesn't exercise the prefix logic, which is a gap in coverage. Can try to figure out how to add a test for it (need to refactor the HTTP calls) if you feel strongly.

internal/godocfx/main.go
tbpg: I'm not sure what you mean by trailing comments. This comes from a command line argument. This makes sure that if the user _did_ include the `/...` at the end, we handle it transparently. A previous version of this command passed the argument through directly, and I made use of `/...`, so I didn't want to break that.

tbpg (commented):
Thank you!

codyoss (approved):
LGTM, thanks!

PR #4291, opened 2021-06-21
testing: add an explicit module download step to build_samples.sh

shollyman:
Currently, the work to go mod replace doesn't account for dependency
changes in head.  This adds an explicit step to download modules to
reflect differences in the latest released google-cloud-go vs head.


codyoss (approved):
LGTM, thanks!

PR #4293, opened 2021-06-22
fix(pubsublite): fixes for background partition count updates

tmdiep:
* When refreshing topic partition counts in the background for Publishers, retry slowly upon resource exhausted errors. The last `gax.Retryer` found in call options is used, so we introduce a `compositeRetryer` for error codes with different backoff params.
* `partitionCountWatcher` ignores errors after the first update. This is consistent with implementations in the Lite Java and Python client libraries.

### DIFF
--- a/pubsublite/internal/wire/rpc.go
+++ b/pubsublite/internal/wire/rpc.go
@@ -110,7 +110,7 @@ func isRetryableStreamError(err error, isEligible func(codes.Code) bool) bool {
 	return isEligible(s.Code())
 }
 
-// Wraps an ordered list of retryers.
+// Wraps an ordered list of retryers. Earlier retryers take precedence.
 type compositeRetryer struct {
 	retryers []gax.Retryer
 }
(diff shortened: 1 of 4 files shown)


pubsublite/internal/wire/rpc.go
```diff
@@ -110,24 +110,49 @@ func isRetryableStreamError(err error, isEligible func(codes.Code) bool) bool {
 	return isEligible(s.Code())
 }
 
-// retryableReadOnlyCallOption returns a call option that retries with backoff
-// for ResourceExhausted in addition to other default retryable codes for
-// Pub/Sub. Suitable for read-only operations which are subject to only QPS
+// Wraps an ordered list of retryers.
```
palmere-google, line 113: nit: maybe note that earlier retriers take precedence

palmere-google (approved):

manuelmenzella-google (approved):

pubsublite/internal/wire/rpc.go
tmdiep: Done

tmdiep:
Thanks both for the review.

PR #4297, opened 2021-06-22
feat(internal/godocfx): different metadata for different modules

tbpg:
I refactored the writing of the metadata into a separate function so it can be tested directly.


codyoss (approved):
LGTM

PR #4299, opened 2021-06-22
fix(firestore): return error from NewClient if projectID is not set

### LINKED ISSUE #4298: firestore: Unclear Error when Project ID is not set
**Client**

Firestore

**Environment**

All

**Context**
I recently had an issue porting my code from App Engine to Cloud Run. I found the SDK was throwing a generic and unhelpful error: `rpc error: code = InvalidArgument desc = Invalid resource field value in the request`

Stack Overflow post I created: https://stackoverflow.com/questions/68086047/invalid-resource-field-value-in-the-request

After spending quite a bit of time trying to diagnose the error, I realized the problem was that I was not supplying a project ID into the project ID variable in NewClient.

**Code**
This is the code for NewClient from https://github.com/googleapis/google-cloud-go/blob/firestore/v1.5.0/firestore/client.go
```
// NewClient creates a new Firestore client that uses the given project.
func NewClient(ctx context.Context, projectID string, opts ...option.ClientOption) (*Client, error) {
	var o []option.ClientOption
	// If this environment variable is defined, configure the client to talk to the emulator.
	if addr := os.Getenv("FIRESTORE_EMULATOR_HOST"); addr != "" {
		conn, err := grpc.Dial(addr, grpc.WithInsecure(), grpc.WithPerRPCCredentials(emulatorCreds{}))
		if err != nil {
			return nil, fmt.Errorf("firestore: dialing address from env var FIRESTORE_EMULATOR_HOST: %s", err)
		}
		o = []option.ClientOption{option.WithGRPCConn(conn)}
		if projectID == DetectProjectID {
			projectID, _ = detectProjectID(ctx, opts...)
			if projectID == "" {
				projectID = "dummy-emulator-firestore-project"
			}
		}
	}
	o = append(o, opts...)

	if projectID == DetectProjectID {
		detected, err := detectProjectID(ctx, o...)
		if err != nil {
			return nil, err
		}
		projectID = detected
	}

	vc, err := vkit.NewClient(ctx, o...)
	if err != nil {
		return nil, err
	}
	vc.SetGoogleClientInfo("gccl", version.Repo)
	c := &Client{
		c:          vc,
		projectID:  projectID,
		databaseID: "(default)", // always "(default)", for now
	}
	return c, nil
}
```

Unfortunately, if you do not provide the input projectID, no error is returned. As explained previously, this leads to a difficult to understand error later.

**Expected behavior**
When you provide invalid inputs to a function, it should return an error.

It would also be helpful if the retuned error was not "Invalid resource field value in the request" and instead the returned description for the error described which resource field value was invalid. But since this is probably generated code / sourced outside this project I assume such a request would be out of scope here.

Gregory-Ledray:
To fix #4298

### DIFF
--- a/firestore/client.go
+++ b/firestore/client.go
@@ -60,6 +60,9 @@ type Client struct {
 
 // NewClient creates a new Firestore client that uses the given project.
 func NewClient(ctx context.Context, projectID string, opts ...option.ClientOption) (*Client, error) {
+	if projectID == "" {
+		return nil, errors.New("firestore: projectID was empty")
+	}
 	var o []option.ClientOption
 	// If this environment variable is defined, configure the client to talk to the emulator.
 	if addr := os.Getenv("FIRESTORE_EMULATOR_HOST"); addr != "" {


Gregory-Ledray:
Feel free to abandon this PR.

crwilcox (approved):

PR #4305, opened 2021-06-23
fix(pubsublite)!: do not expose apply for CreateSubscriptionOption

tmdiep:
Since the `createSubscriptionSettings` parameter is not public, also hide the `apply()` func.


tmdiep:
WANT_LGTM=any

Sorry, just noticed this as I was doing a final pass through the API surface.

tmdiep:
apidiff detected a breaking change. Can we suppress/ignore it for this change?

tmdiep:
Replacement (tests pass): https://github.com/googleapis/google-cloud-go/pull/4344

PR #4306, opened 2021-06-23
chore(internal): gen googleapis-discovery compute genproto

noahdietz:
Hardcodes go-genproto generation for google/cloud/compute/v1 from googleapis-discovery. We will need to figure out how we support other protos in that repo in the future.

Note: This **deos not** add gapic gen for compute yet.

Note: `compute.pb.go` is approximately 158k LoC 😮 .

cc: @vchudnov-g @software-dov

### DIFF
--- a/internal/gapicgen/generator/genproto.go
+++ b/internal/gapicgen/generator/genproto.go
@@ -55,19 +55,21 @@ var denylist = map[string]bool{
 
 // GenprotoGenerator is used to generate code for googleapis/go-genproto.
 type GenprotoGenerator struct {
-	genprotoDir   string
-	googleapisDir string
-	protoSrcDir   string
-	forceAll      bool
+	genprotoDir        string
+	googleapisDir      string
+	googleapisDiscoDir string
+	protoSrcDir        string
+	forceAll           bool
 }
 
 // NewGenprotoGenerator creates a new GenprotoGenerator.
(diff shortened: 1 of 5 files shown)


codyoss:
@noahdietz Can can you rebase this PR now that regen should be unblocked.

noahdietz:
> @noahdietz Can can you rebase this PR now that regen should be unblocked.

Done!

internal/gapicgen/generator/genproto.go
```diff
@@ -132,9 +138,15 @@ func (g *GenprotoGenerator) Regen(ctx context.Context) error {
 		fn := fileNames
 		grp.Go(func() error {
 			log.Println("running protoc on", pk)
-			return g.protoc(fn)
+			return g.protoc(fn, true /* grpc */)
 		})
 	}
+	// TODO(noahdietz): This needs to be generalized to support any proto in googleapis-discovery.
```
codyoss, line 144: We can change the format of the file to have several commits to track if necessary as this is all hand-rolled(no pattern to follow).

codyoss (approved):
LGTM, thanks!

internal/gapicgen/generator/genproto.go
noahdietz: Yeah, I think we may have to eventually. I'm not sure the googleapis-discovery pipeline is fully automated yet, so it's not urgent.

PR #4307, opened 2021-06-23
chore(internal/gapicgen): gen googleapis-discovery compute gapic

noahdietz:
Adds generation of a `compute` gapic that uses googleapis-discovery. 

Note: we need to capture some fixes in gapic-generator-go with a release there and update here before this will work.

cc: @software-dov @vchudnov-g


noahdietz:
Needs #4306 merged as well.

noahdietz:
Need #4315 to go in first

noahdietz:
We will need another gapic-generator-go release as there things being fixed before we can start generating compute

noahdietz:
generator v0.21.2 introduced in #4315 should do the trick

noahdietz:
Holding off the merge until #4315 is in.

PR #4308, opened 2021-06-23
chore(internal): disable apidiff script

noahdietz:
The `internal/apidiff` tool is borked so I'm disabling it for now.


codyoss:
@noahdietz Out of curiosity what are you seeing?

noahdietz:
Seems as though they were just local errors...in checking a recent kokoro run, it looks like it worked fine! Let's keep it then. Closing.

PR #4313, opened 2021-06-24
fix(pubsub): retry GOAWAY errors

### LINKED ISSUE #4257: pubsub: non-retryable rpc errors in subscription `Receive()` method
**Client**

PubSub

**Environment**

Ubuntu Focal on GKE

**Go Environment**

$ go version
go version go1.16.5

**Code**

We're using the standard way to receive messages from a subscription, with default configurations (similar to the [code snippet from doc](https://pkg.go.dev/cloud.google.com/go/pubsub#hdr-Receiving)):

```go
 err := subscription.Receive(context.Background(), func(ctx context.Context, m *Message) {
 	log.Printf("Got message: %s", m.Data)
 	m.Ack()
 })
 if err != nil {
	// Handle error.
 }
```

We've managed to use this method for several months in long-living containers, without receiving any (non-retryable) errors.

On June 2, 2021, we started to receive sporadic fatal RPC errors in the following forms:

```
rpc error: code = Unknown desc = closing transport due to: connection error:
  desc = "error reading from server: EOF", received prior goaway: code: NO_ERROR, debug data: server_shutting_down

rpc error: code = Unknown desc = closing transport due to: connection error:
  desc = "error reading from server: EOF", received prior goaway: code: NO_ERROR, debug data: max_age

rpc error: code = Unknown desc = closing transport due to: connection error:
  desc = "error reading from server: read tcp <ip>:54780-><ip>:443: read: connection reset by peer", received prior goaway: code: NO_ERROR, debug data: max_age
```

We've seen these kinds of errors several times per day since then.

**Expected behavior**

Automatic internal retries on these kinds of RPC errors, so that the [Receive()](https://pkg.go.dev/cloud.google.com/go/pubsub#Subscription.Receive) method doesn't error.

I've noticed that there is a piece of retrying logic in [pubsub/service.go](https://github.com/googleapis/google-cloud-go/blob/pubsub/v1.11.0/pubsub/service.go#L51) that detects server shutdowns associated to the `Unavailable` code and the `"Server shutdownNow invoked"` message.

Maybe something recently changed in the way the remote server signals that it is shutting down and this logic needs to be updated?

**Actual behavior**

The [Receive()](https://pkg.go.dev/cloud.google.com/go/pubsub#Subscription.Receive) method from the subscription errors.

**Additional context**

Started on June 2, 2021.

Not long after we upgraded to pubsub [v1.11.0](https://github.com/googleapis/google-cloud-go/releases/tag/pubsub%2Fv1.11.0), although I don't see how it would be related - so I'm mentioning this mostly for completeness.

hongalex:
The frontend server for Pub/Sub might occasionally emit GOAWAY errors which are currently not retried. This is not unique to the Go client, though the categorization as `UNKNOWN` vs `UNVAILABLE` is a golang-GRPC issue. Although UNKNOWN [should not generally be retried](https://google.aip.dev/194), this will unblock users of `Receive` until the grpc library can be changed. See internal cl/377393940 for a similar fix in another Go library.

Fixes #4257.


pubsub/service.go
```diff
@@ -62,6 +62,14 @@ func (r *defaultRetryer) Retry(err error) (pause time.Duration, shouldRetry bool
 			return r.bo.Pause(), true
 		}
 		return 0, false
+	case codes.Unknown:
+		// Retry GOAWAY, see https://github.com/googleapis/google-cloud-go/issues/4257.
+		isGoaway := strings.Contains(s.Message(), "closing error reading from server: EOF") &&
```
tmdiep, line 67: Just checking: does the error message contain exactly "closing error reading from server..."? I looked at the linked issue and that phrase isn't prefixed by closing, so it may not match.

tmdiep (approved):

PR #4316, opened 2021-06-24
feat(pubsublite): support seek subscription in AdminClient

tmdiep:
`SeekSubscription` performs an out-of-band seek for a subscription to a specified target. Adds `SeekSubscriptionOperation`, which wraps the auto-generated gapic type for polling the seek operation.


pubsublite/operations.go
```diff
@@ -0,0 +1,152 @@
...
+const (
+	// End refers to the location past all currently published messages. End
+	// skips the entire message backlog.
+	End BacklogLocation = iota + 1
+
+	// Beginning refers to the location of the oldest retained message.
+	Beginning
+)
+
+func (b BacklogLocation) setRequest(req *pb.SeekSubscriptionRequest) {
```
manuelmenzella-google, line 44: This is really cool! I didn't know Go allowed attaching behavior to types that are effectively ints.

manuelmenzella-google (approved):

tmdiep:
@codyoss, @tbpg - please review this API change.

I couldn't find other examples of handwritten libraries exposing operation polling. We want to allow waiting for the operation to be optional, as users may want it to be asynchronous in most cases. 

Notes:
* `pubsublite.SeekSubscriptionOperation` basically just wraps the gapic generated operation, transforming Timestamp protos to `time.Time`.
* `SeekTarget` implements the equivalent of a `oneof`.
* `SeekSubscriptionOption` is a placeholder for future options (which we've deferred). It looks a little strange to define now. But I wanted to avoid introducing another variant of `SeekSubscription`.

pubsublite/operations.go
```diff
@@ -0,0 +1,152 @@
...
+			Time: &pb.TimeTarget_EventTime{tspb.New(time.Time(e))},
+		},
+	}
+}
+
+// SeekSubscriptionOption is reserved for future options.
+type SeekSubscriptionOption interface{}
+
+// OperationMetadata stores metadata for long-running operations.
+type OperationMetadata struct {
```
codyoss, line 82: You might want to call this SeekOperationMetadata so should other LROs be added in the future this struct is unique.

pubsublite/operations.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return protoToOperationMetadata(m)
+}
+
+// Wait polls until the seek operation is complete and returns one of the
+// following:
+//  - nil if the operation is complete and succeeded.
+//  - failure reason if the operation is complete and failed.
+//  - error if polling the operation status failed due to a non-retryable error.
+func (s *SeekSubscriptionOperation) Wait(ctx context.Context) error {
```
codyoss, line 149: I think this should return a response type, even if that type contains nothing today. I think `pubsub` has used this strategy in the past, @hongalex correct me if I am wrong. This way you can start returning info in the future without the need of a new method/breaking-change.

pubsublite/operations.go
```diff
@@ -0,0 +1,152 @@
...
+	_, err := s.op.Wait(ctx)
+	return err
+}
```
codyoss, line 152: It looks like you have matched the LRO pattern from the gapic layer, which I agree seem right to me. Any reason to not also expose Poll? Then this would be more like every other LRO.

codyoss (commented):
A couple of comments, Looks good overall though!

pubsublite/operations.go
tmdiep: We decided to go with a common `OperationMetadata` proto for all operations to be consistent with a number of other APIs, so just reused the name here.

pubsublite/operations.go
tmdiep: Done. I guess it doesn't hurt, even if it's empty.

pubsublite/operations.go
tmdiep: I wasn't sure how often `Poll` is used in practice. The integration tests just call `Wait` and I noticed that the other handwritten libraries tend to just call `Wait` internally as well. So `Poll` was omitted to minimize the API surface, until a user asks for it.

Thanks for reviewing!

codyoss: SGTM. I think the main use case is when programming in a "single threaded" environment where you want a call that is not long-blocking. It would allow users to get a quick status check as well as metadata for the operation even if the operation is not in a terminal state.

PR #4317, opened 2021-06-25
chore(bigtable): switch to google.golang.org/protobuf

neild:
Update references to github.com/golang/protobuf to use the newer module.
No user-visible changes.

Regenerate bigtable/internal/conformance/tests.pb.go.


codyoss:
@telpirion Would you mind taking a look at this one.

telpirion (approved):

trakhimenok:
Why is this not merging? Would be good to get rid of deps on an obsolete module.

codyoss:
@trakhimenok I think this issue just fell through the cracks. I think at this point we should just make a new PR

PR #4312, opened 2021-06-24
feat(bigquery): enable project autodetection, expose project ids further

shollyman:
PR supersedes: https://github.com/googleapis/google-cloud-go/pull/4076

Related: https://github.com/googleapis/google-cloud-go/issues/1294

With this change, project autodetection is enabled via use of a sentinel
value, and the retained project identifier is now exposed on the Client
and Job resources via the Project() function.

PR #4320, opened 2021-06-25
fix(bigquery): minor rename to feature that's not yet in a release

shollyman:
This adds a minor correction to Job's access to the underlying project.
While the access on the Client is more accurately called `Project()` as
it accepts project ID or project number, in this case the field is part
of the underlying JobReference type.

This was added yesterday in
https://github.com/googleapis/google-cloud-go/pull/4312 and hasn't made
it into a release cut yet, so addressing this now before release.


shollyman:
Broken kokoro is the apidiff warning about the change.

codyoss (approved):

codyoss:
LGTM since this was not yet released.

Issue #1438, opened 2019-05-29
firestore: Add function to convert firestore protobuf document to a struct


bashtian:
Cloud Functions supports Firestore triggers and when a document is updated an event is send to the function with this format:

``` go
pb "google.golang.org/genproto/googleapis/firestore/v1"

// FirestoreEvent is the payload of a Firestore event.
type FirestoreEvent struct {
	OldValue   pb.Document `json:"oldValue"`
	Value      pb.Document `json:"value"`
	UpdateMask struct {
		FieldPaths []string `json:"fieldPaths"`
	} `json:"updateMask"`
}
```
There is currently no way to convert `pb.Document` to a normal Go struct. [setFromProtoValue](https://github.com/googleapis/google-cloud-go/blob/61166ef305533a6c5b0b3275df2cb544d678339a/firestore/from_value.go#L27) could be exported to make it possible or a new helper function based on this [DataTo](https://github.com/googleapis/google-cloud-go/blob/27b5416a80e6a65ee9eebda55f06def51b9b7b88/firestore/document.go#L115) line.

bashtian:
The data is send as JSON but unmarshalling it into a document does not work because OneOf fields are not supported. So this feature request not needed until we can unmarshal the data into a pb.Document.

jadekler:
cc @tbpg

jadekler:
cc @benwhitehead

tbpg:
Seems like a good idea to me.

odeke-em:
Thank you for filing this issue @bashtian!

I don't understand what the feature request is about that isn't already supported. https://godoc.org/google.golang.org/genproto/googleapis/firestore/v1#Document is already a "normal" Go struct and moreover it can be JSON marshaled and then unmarshaled to a map or even another struct for which the JSON tags and fields could be attached.

Could you please show a use case/code sample for where converting to a "normal Go struct" from that pb.Document arises?

tbpg:
Ahh. So, here is the flow:

1. Store some data in Firestore, going in as type `MyModel`.
2. A Firestore Cloud Function is triggered and the data comes out as [`firestorepb.Document`](https://godoc.org/google.golang.org/genproto/googleapis/firestore/v1#Document), where all of the fields from `MyModel` are untyped in the `Fields` field.
[...]

PR #4328, opened 2021-06-28
feat(firestore): add NewDocumentSnapshotFromProto

jsmouret:
See #1438


jsmouret:
Sorry, content of Fields in FirebaseValue is not a map[string]*pb.Value

PR #4335, opened 2021-06-29
feat(firestore): add support for FirebaseEvent

jsmouret:
Add useful tooling to work with Firebase triggers.
See #1438

### DIFF
--- a/firestore/event.go
+++ b/firestore/event.go
@@ -1,0 +1,108 @@
+// Copyright 2017 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


codyoss:
I wonder if https://github.com/googleapis/google-cloudevents-go is the correct place for this, these events might already exist there.

jsmouret:
It seems to match those [structures](https://github.com/googleapis/google-cloudevents-go/blob/master/cloud/firestore/v1/document_event_data.go) but I don't see any logic in the repo. It is only definitions.

jsmouret:
... and they look different from the original [documentation](https://cloud.google.com/functions/docs/calling/cloud-firestore#code_sample)

codyoss:
@jsmouret  Correct, that repo in just meant to have code for the different event types. Would you mind opening up an issue for the "glue code" and describe why you think adding it would be beneficial.

jsmouret:
Can I piggyback on that issue #1438 ? or do you prefer a new one?

codyoss:
@googleapis/api-firestore Can we decide what to do with this PR?

jsmouret:
Not using firebase anymore, so closing this PR

PR #4346, opened 2021-06-29
fix(firestore): correct an issue with returning empty query partitions

crwilcox:


tritone (approved):
Is there a test that would/could have caught this?

crwilcox:
@tritone we have a test for getting all contents of all partitions. The issue is we can't guarantee the partition count, but we can guarantee that all of them together get the contents. A small change we could make: check that the partitions have the shape we expect (as this was they were just empty)

PR #4347, opened 2021-06-29
feat(firestore): Support DocumentRefs in OrderBy, Add Query.Serialize, Query.Deserialize for cross machine serialization

crwilcox:
Changes to the internal workings of query to store things in a proto representation. I wanted to do this for start/end cursors, but due to the allowing of order to happen after startat, it isn't feasible. There is a bug #4448 to track making a breaking change that would allow simplifying the cursors. See https://github.com/googleapis/google-cloud-go/pull/4347/commits/d4f34b0d2e43cc5653be4af1c5cae99891e9f02b for that approach (it is untidy, but should be functional)

This also enables serialization to proto which should make it possible for users to stream the query on another machine/process.

Also extended orderby to support document refs, a long todo that needed to be implemented to support round tripping.


firestore/integration_test.go
```diff
@@ -1674,12 +1674,26 @@ func TestIntegration_ColGroupRefPartitionsLarge(t *testing.T) {
...
 	totalCount := 0
 	for _, query := range partitions {
-
 		allDocs, err := query.Documents(ctx).GetAll()
 		if err != nil {
 			t.Fatalf("GetAll(): received unexpected error: %v", err)
 		}
 		totalCount += len(allDocs)
+
+		// Check that the same results are returned evenv if we use the proto converted query
```
schmidt-sebastian, line 1683:
```suggestion
		// Check that the same results are returned even if we use the proto converted query
```

firestore/query.go
```diff
@@ -250,6 +251,159 @@ func (q *Query) processCursorArg(name string, docSnapshotOrFieldValues []interfa
 
 func (q Query) query() *Query { return &q }
 
+func getSafeCursorValue(vProto *pb.Value, q Query) (interface{}, error) {
+	switch v := vProto.ValueType.(type) {
+	case *pb.Value_ReferenceValue:
+		refVal := vProto.GetReferenceValue()
+		if strings.HasPrefix(refVal, q.path) {
```
schmidt-sebastian, line 258: Is this the database name? If so, it should always be there.

firestore/query.go
```diff
@@ -250,6 +251,159 @@ func (q *Query) processCursorArg(name string, docSnapshotOrFieldValues []interfa
...
+
+	// 	startVals, endVals     []interface{}
+	// 	startDoc, endDoc       *DocumentSnapshot
+	// 	startBefore, endBefore bool
+	if startAt := pbq.GetStartAt(); startAt != nil {
+		if startAt.GetBefore() {
+			q.startBefore = true
+		}
+		for _, v := range startAt.GetValues() {
+			c, err := getSafeCursorValue(v, q)
```
schmidt-sebastian, line 308: I am not sure how difficult it would be, but this conversion would not be needed if the Query class stored the cursor as the Proto.

firestore/query.go
```diff
@@ -250,6 +251,159 @@ func (q *Query) processCursorArg(name string, docSnapshotOrFieldValues []interfa
...
+	// 	selection              []FieldPath
+	if s := pbq.GetSelect(); s != nil {
+		fields := s.GetFields()
+		fieldStrings := make([]string, 0)
+
+		// As part of sending to protos, field paths are escaped. Unescape
+		for _, v := range fields {
+			fieldPath := v.GetFieldPath()
+			if strings.Contains(fieldPath, "`") {
+				fieldPath = strings.ReplaceAll(fieldPath, "`", "")
```
schmidt-sebastian, line 353: Again, this would not be needed if we stored the Proto representation. This also seems to be missing the unescape on `\` (which also allows \\\` to be used to represent \`)

schmidt-sebastian (commented):
I think your life would be simpler if you changed the Query class to be backed by Protos. Inspiration: https://github.com/googleapis/java-firestore/blob/9ff2f41b765c8878c3b3fb7df962f6f1ed537f05/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java#L234

firestore/integration_test.go
```diff
@@ -1674,12 +1674,26 @@ func TestIntegration_ColGroupRefPartitionsLarge(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("FromProto error: %v", err)
+		}
+		protoReturnedDocs, err := q.Documents(ctx).GetAll()
+		if err != nil {
+			t.Fatalf("GetAll error: %v", err)
+		}
+		if len(allDocs) != len(protoReturnedDocs) {
+			t.Fatalf("Expected document count to be the same on both query runs: %v", err)
+		}
```
schmidt-sebastian, line 1696: FYI Since this is a partition query, this does not verify a lot of the query internals (e.g. filters or more complex cursors, orderBys).

firestore/query.go
crwilcox: Moved field path to proto rep.

firestore/query.go
schmidt-sebastian: The change to back cursors by Protos should actually be very simple and remove `getSafeCursorValue()`. Unlike filters and orderBys, you can store the cursor directly and do not need to wrap it in an array.

schmidt-sebastian (commented):
Basically LGTM, but I do think you should also convert the Cursor to Proto which should simplify the new code.

firestore/integration_test.go
crwilcox: Yep. Most of what I am using to test the change is our existing toproto test which I modified to do a roundtrip verification

firestore/query_test.go
```diff
@@ -448,8 +459,41 @@ func TestQueryToProto(t *testing.T) {
...
+
+		want := test.in
+		gotProto, err := got.ToProto()
+		fmt.Println(gotProto)
+		if err != nil {
+			t.Fatalf("%s: %v", test.desc, err)
+		}
+
+		// Compare protos before and after taking to a query. proto -> query -> proto.
+		if diff := cmp.Diff(gotProto, proto, protocmp.Transform()); diff != "" {
```
crwilcox, line 488: @schmidt-sebastian resolving your other comment, but this is where we primarily are testing the serialization. We had a dataset testing to proto. I added a case that goes query -> proto -> query -> proto and compares both protos to verify they are the same.

firestore/query.go
crwilcox: I made these changes and I like them, unfortunately the go library allows orderby after startat, which means we need to have a doc snapshot available to retrieve fields from. As removing this would be breaking I am not going to be simplifying this now.

crwilcox: https://github.com/googleapis/google-cloud-go/issues/4448 to track that breaking change though. I'd like to do this when we need to do a break so we would enable this simplification in the future.

firestore/client.go
```diff
@@ -148,6 +148,12 @@ func (c *Client) Doc(path string) *DocumentRef {
 	return doc
 }
 
+// Query creates a query object related to this client.
+func (c *Client) Query() *Query {
```
codyoss, line 152: It looks like the semantics of the public API all are defined on a value type. Should the return a value too?

codyoss: Also it is not obvious to me how/when this method should be called. Could you provide an example?

firestore/query.go
```diff
@@ -50,6 +51,8 @@ type Query struct {
 	startBefore, endBefore bool
 	err                    error
 
+	structuredQueryProto *pb.StructuredQuery
```
codyoss, line 54: This appears to be unused.

firestore/query.go
```diff
@@ -103,7 +122,12 @@ func (q Query) Where(path, op string, value interface{}) Query {
 		q.err = err
 		return q
 	}
-	q.filters = append(append([]filter(nil), q.filters...), filter{fp, op, value})
+	proto, err := filter{fp, op, value}.toProto()
+	if err != nil {
+		q.err = err
+		return q
+	}
+	q.filters = append(append([]*pb.StructuredQuery_Filter(nil), q.filters...), proto)
```
codyoss, line 130: nit: Can this be two separate lines, then no initialization is needed.

firestore/query.go
```diff
@@ -141,15 +171,15 @@ func (q Query) OrderBy(path string, dir Direction) Query {
 		q.err = err
 		return q
 	}
-	q.orders = append(q.copyOrders(), order{fp, dir})
+	q.orders = append(q.copyOrders(), order{fp, nil, dir})
```
codyoss, line 174: nit: If this new field was added to the end of the struct no code change would be needed here. In general I would avoid using unkeyed struct literals where possible as they can lead to more trouble than they are worth, imo.

firestore/client.go
codyoss: Should this be public?

firestore/query.go
```diff
@@ -250,6 +280,147 @@ func (q *Query) processCursorArg(name string, docSnapshotOrFieldValues []interfa
...
+			q.err = err
+			return q, err
+		}
+		return i, nil
+	}
+}
+
+// FromProto creates a new Query object from a RunQueryRequest. This can be used
+// in combintation with ToProto to serialize Query objects.
+func (q Query) FromProto(pbQuery *pb.RunQueryRequest) (Query, error) {
```
codyoss, line 303: When will a user have a reference to this directly? This is never returned from the public api, correct?

codyoss (commented):
Just took a quick look, will take another after I understand better the why of some of these changes.

firestore/query_test.go
schmidt-sebastian: I saw that too late and forgot to close my previous comment :)

firestore/query.go
```diff
@@ -250,6 +280,147 @@ func (q *Query) processCursorArg(name string, docSnapshotOrFieldValues []interfa
...
+func getSafeCursorValue(vProto *pb.Value, q Query) (interface{}, error) {
+	switch vProto.ValueType.(type) {
+	case *pb.Value_ReferenceValue:
+		refVal := vProto.GetReferenceValue()
+		if strings.HasPrefix(refVal, q.path) {
+			refVal = refVal[len(q.path)+1:]
+		}
+		return refVal, nil
+	default:
+		i, err := createFromProtoValue(vProto, q.c)
```
schmidt-sebastian, line 292: `createFromProtoValue` should handle references as well. Did you try that?

crwilcox: I don't think it does, at least as is. There is a todo item to support docRefs. I think this would clean things up though so I am taking the time to add that support.

crwilcox: added allowing of documentRefs, so this code is greatly simplified.

firestore/query.go
```diff
@@ -549,22 +536,26 @@ func (q *Query) fieldValuesToCursorValues(fieldValues []interface{}) ([]*pb.Valu
...
-			if !ok {
+
+			switch docID := fval.(type) {
+			case string:
+				vals[i] = &pb.Value{ValueType: &pb.Value_ReferenceValue{q.path + "/" + docID}}
+				continue
+			case *DocumentRef:
+				// DocumentRef can be transformed in usual way.
+			default:
 				return nil, fmt.Errorf("firestore: expected doc ID for DocumentID field, got %T", fval)
```
crwilcox, line 548: I could change this err, but it doesn't seem far off since we accept docRefs now in addition. Happy to alter though if anyone has opinions.

firestore/query.go
crwilcox: offline chat. this is to result in a copied filter slice.

firestore/query.go
crwilcox: I agree with this, though adding a fieldRef to the end of a set of things (fieldPath, direction, fieldRef) felt a bit off and since it is internal I could safely refactor. I could alter these to all state the field name though if that is preferred. It would like make it more readable anyway.

firestore/query.go
crwilcox: Users get this via ToProto. (line 390).

firestore/query.go
crwilcox: code deleted when refvals became supported.

firestore/client.go
crwilcox: so, my thought here was to allow getting at Query().FromProto() for users. There may be a better way. This isn't strictly needed mind you, but I thought it made it a bit nicer.

firestore/query.go
```diff
@@ -103,7 +119,12 @@ func (q Query) Where(path, op string, value interface{}) Query {
 		q.err = err
 		return q
 	}
-	q.filters = append(append([]filter(nil), q.filters...), filter{fp, op, value})
+	proto, err := filter{fp, op, value}.toProto()
```
tritone, line 122: Should this be factored out as a function?

tritone (commented):
Seems basically fine, just trying to understand better what this is for.

firestore/query.go
```diff
@@ -250,6 +276,130 @@ func (q *Query) processCursorArg(name string, docSnapshotOrFieldValues []interfa
...
+	// 	path                   string // path to query (collection)
+	// 	parentPath             string // path of the collection's parent (document)
+	parent := pbQuery.GetParent()
+	q.parentPath = parent
+	q.path = parent + "/" + q.collectionID
+
+	// 	startVals, endVals     []interface{}
+	// 	startDoc, endDoc       *DocumentSnapshot
+	// 	startBefore, endBefore bool
+	if startAt := pbq.GetStartAt(); startAt != nil {
```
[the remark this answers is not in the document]
crwilcox: the reverse exists to create the proto request from the query, but the reverse (proto to query) doesn't exist elsewhere.

firestore/query.go
crwilcox: I can look at adding docs. the purpose is to allow a user to fanout work to distributed process. `GetParititionedQueries`  returns a Slice of queries. To aid in serialization, these can be converted to RunQueryRequest which can be more easily serialized

firestore/client.go
crwilcox: A thought: This isn't strictly necessary. 

```
client.Query().FromProto(runQueryRequest)

// vs

client.Collection("collectionId").Query.FromProto(runQueryRequest)
```


We could require the user to place the query into a collection first. I though this way was a bit more 'obvious' but it could also be added at later point

firestore/integration_test.go
```diff
@@ -1675,12 +1675,26 @@ func TestIntegration_ColGroupRefPartitionsLarge(t *testing.T) {
...
 	for _, query := range partitions {
-
 		allDocs, err := query.Documents(ctx).GetAll()
 		if err != nil {
 			t.Fatalf("GetAll(): received unexpected error: %v", err)
 		}
 		totalCount += len(allDocs)
+
+		// Check that the same results are returned even if we use the proto converted query
+		pbStructuredQuery, _ := query.ToProto()
```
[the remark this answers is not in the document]
crwilcox: I am ensuring the ToProto/FromProto properly round trips.

firestore/client.go
```diff
@@ -148,11 +148,12 @@ func (c *Client) Doc(path string) *DocumentRef {
 	return doc
 }
 
+// TODO(crwilcox): Commented out, may be best to not extend surface at client level.
```
tritone, line 151: Nice, I think removing this is good. Best not to add additional ways of doing things without a clear rationale.

firestore/collgroupref.go
```diff
@@ -54,6 +54,9 @@ func newCollectionGroupRef(c *Client, dbPath, collectionID string) *CollectionGr
 //
 // If a Collection Group Query would return a large number of documents, this
 // can help to subdivide the query to smaller working units that can be distributed.
+//
+// If the goal is to run the queries across processes or workers, it may be useful to use
+// `Query.ToProto` and `Query.FromProto` to serialize the query.
```
tritone, line 59: This is good. I think it would be good to have a longer doc (here or elsewhere, not sure where the best place is, maybe the docs for the Query type?) that describes step-by-step how to run queries across workers, at least on a high level. Or if that best practices doc exists in the cloudsite firestore docs, we should link to it.

I guess the flow would be (for multiple worker machines): partition query, serialize to protos, send to workers, get result, deserialize. Or for threads on a single machine: partition query, spin up 1 goroutine per partition, run, collect results (no serialization/deserialization needed). Do I have that right?

crwilcox: That is my thinking. if using goroutines I figure you just use the Query directly. The design calls to make sure we can serialize since this feature is mostly used for large collection groups, so beneficial to be able to partition. That said, I imagine for go, most users won't be doing this. I am working with another on creating user samples and solutions in the near future and I think, to your point, this would benefit from that

firestore/query.go
crwilcox: Great catch. I can use WherePath to replace all that follow.s

firestore/query.go
crwilcox: A small it of docstrings have been added to explain why you might use this.

tritone (approved):
Nice work on this. I think we could still use a longer worked example and explanation for how to use this (in Go and across languages) but that can come later.

tritone (approved):
Nice, I like this new surface!

PR #4373, opened 2021-07-02
test: add pubsublite to local modules

hongalex:
Adds Pub/Sub Lite which was missing from the list of modules of handwritten libraries


shollyman (approved):

codyoss (approved):
Thanks!

Issue #4366, opened 2021-07-01
bigquery: build veneer for bigquery write client


shollyman:
Tracks PRs related to veneer work for new streaming client veneer.

internally b/178808000

shollyman:
Didn't link it during review, but https://github.com/googleapis/google-cloud-go/pull/4729 changed the AppendRows contract to return a single appendresult future rather than one per row.

shollyman:
Didn't get tagged at commit time, but https://github.com/googleapis/google-cloud-go/pull/5102 added variadic appends

shollyman:
Closing this FR at this point.  Tagging relevant the PRs is becoming more noisy.

PR #4375, opened 2021-07-02
feat(bigquery/storage/managedwriter/adapt): add schema -> proto support

shollyman:
This PR adds the ability to generate proto message definitions dynamically based on a table's schema.  The new BQ write API communicates data solely through protocol buffers , so this helps enable the various use cases where users show up without a proto message predefined/compiled.

With this, a user can use the dynamic definitions to construct and serialize messages suitable for use in the write client.  It also enables other features like json -> dynamic proto -> serialized data.

More background: The underlying API expects users to communicate proto schema (in the form of a [DescriptorProto](https://pkg.go.dev/google.golang.org/protobuf@v1.27.1/types/descriptorpb#DescriptorProto)).  Then we can append rows via streaming RPC where the backend acknowledges the writes.


bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,260 @@
...
+	storagepb.TableFieldSchema_BYTES:      ".google.protobuf.BytesValue",
+	storagepb.TableFieldSchema_DATE:       ".google.protobuf.Int32Value",
+	storagepb.TableFieldSchema_DATETIME:   ".google.protobuf.Int64Value",
+	storagepb.TableFieldSchema_DOUBLE:     ".google.protobuf.DoubleValue",
+	storagepb.TableFieldSchema_GEOGRAPHY:  ".google.protobuf.StringValue",
+	storagepb.TableFieldSchema_INT64:      ".google.protobuf.StringValue",
+	storagepb.TableFieldSchema_NUMERIC:    ".google.protobuf.BytesValue",
+	storagepb.TableFieldSchema_STRING:     ".google.protobuf.StringValue",
+	storagepb.TableFieldSchema_TIME:       ".google.protobuf.Int64Value",
+	storagepb.TableFieldSchema_TIMESTAMP:  ".google.protobuf.Int64Value",
```
yirutang, line 70: Is there a Timestamp type in proto3?

bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,260 @@
...
+	storagepb.TableFieldSchema_BIGNUMERIC: ".google.protobuf.BytesValue",
+	storagepb.TableFieldSchema_BOOL:       ".google.protobuf.BoolValue",
+	storagepb.TableFieldSchema_BYTES:      ".google.protobuf.BytesValue",
+	storagepb.TableFieldSchema_DATE:       ".google.protobuf.Int32Value",
+	storagepb.TableFieldSchema_DATETIME:   ".google.protobuf.Int64Value",
+	storagepb.TableFieldSchema_DOUBLE:     ".google.protobuf.DoubleValue",
+	storagepb.TableFieldSchema_GEOGRAPHY:  ".google.protobuf.StringValue",
+	storagepb.TableFieldSchema_INT64:      ".google.protobuf.StringValue",
+	storagepb.TableFieldSchema_NUMERIC:    ".google.protobuf.BytesValue",
+	storagepb.TableFieldSchema_STRING:     ".google.protobuf.StringValue",
```
yirutang, line 68: Do we need this wrapper?

bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,260 @@
...
+var bqTypeToWrapperMap = map[storagepb.TableFieldSchema_Type]string{
+	storagepb.TableFieldSchema_BIGNUMERIC: ".google.protobuf.BytesValue",
+	storagepb.TableFieldSchema_BOOL:       ".google.protobuf.BoolValue",
+	storagepb.TableFieldSchema_BYTES:      ".google.protobuf.BytesValue",
+	storagepb.TableFieldSchema_DATE:       ".google.protobuf.Int32Value",
+	storagepb.TableFieldSchema_DATETIME:   ".google.protobuf.Int64Value",
+	storagepb.TableFieldSchema_DOUBLE:     ".google.protobuf.DoubleValue",
+	storagepb.TableFieldSchema_GEOGRAPHY:  ".google.protobuf.StringValue",
+	storagepb.TableFieldSchema_INT64:      ".google.protobuf.StringValue",
+	storagepb.TableFieldSchema_NUMERIC:    ".google.protobuf.BytesValue",
```
yirutang, line 67: Do we need this wrapper?

bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,260 @@
...
+// proper representation of NULL values, as proto3 semantics would just use default value.
+var bqTypeToWrapperMap = map[storagepb.TableFieldSchema_Type]string{
+	storagepb.TableFieldSchema_BIGNUMERIC: ".google.protobuf.BytesValue",
+	storagepb.TableFieldSchema_BOOL:       ".google.protobuf.BoolValue",
+	storagepb.TableFieldSchema_BYTES:      ".google.protobuf.BytesValue",
+	storagepb.TableFieldSchema_DATE:       ".google.protobuf.Int32Value",
+	storagepb.TableFieldSchema_DATETIME:   ".google.protobuf.Int64Value",
+	storagepb.TableFieldSchema_DOUBLE:     ".google.protobuf.DoubleValue",
+	storagepb.TableFieldSchema_GEOGRAPHY:  ".google.protobuf.StringValue",
+	storagepb.TableFieldSchema_INT64:      ".google.protobuf.StringValue",
```
yirutang, line 66: Why is this mapping?

bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,260 @@
...
+func (dm *dependencyCache) get(schema *storagepb.TableSchema) protoreflect.Descriptor {
+	if dm == nil {
+		return nil
+	}
+	b, err := proto.Marshal(schema)
+	if err != nil {
+		return nil
+	}
+	encoded := base64.StdEncoding.EncodeToString(b)
+	// TODO: should we
```
yirutang: ?

bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,260 @@
...
+	if field.GetMode() != storagepb.TableFieldSchema_NULLABLE {
+		return &descriptorpb.FieldDescriptorProto{
+			Name:   proto.String(name),
+			Number: proto.Int32(idx),
+			Type:   bqTypeToFieldTypeMap[field.GetType()].Enum(),
+			Label:  bqModeToFieldLabelMap[field.GetMode()].Enum(),
+		}, nil
+	}
+	// For NULLABLE, we use the wrapper types.
+	return &descriptorpb.FieldDescriptorProto{
```
yirutang, line 252: Should we default nullable to wrapper? Should we annotate the field with use_defaults instead?
https://screenshot.googleplex.com/3kHa8QbkhmgFsuj

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: https://github.com/protocolbuffers/protobuf/tree/master/src/google/protobuf, but this comes down to what the backend accepts for each column type.  If the backend will convert timestamp protos, we can use them.

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: Error, corrected.

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: I'm hesitant to include a bunch of internal type annotations for driving behavior here.

*  We have no stable source of truth for them.  They're published within the [zetasql project](https://github.com/google/zetasql/blob/master/zetasql/public/proto/type_annotation.proto), but that project doesn't make any guarantees about stability and has no GA release, so the route to a GA launch is...probably problematic.

*  The reason we're using proto3 semantics is external users live in a proto3 ecosystem.  In that world, the wrapper types are how to properly communicate nulls.  I don't know that it matters, but type_annotations.proto is still proto2 based, so it's possible it introduces other issues as a side effect.  We could consider revisiting this route down the line.

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: empty string vs null string?

bigquery/storage/managedwriter/adapt/protoconversion.go
yirutang: Yeah, we haven't supported it yet.

bigquery/storage/managedwriter/adapt/protoconversion.go
yirutang: String field can be not set, which means null?

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: Stale comment, since removed.

bigquery/storage/managedwriter/adapt/protoconversion.go
yirutang: I don't think frontend supports wrapper to nullable field conversion (we probably should). So currently if you convert these to wrapper, they will try map themselves back as a struct. In order to support such converter, the field needs to be annotated as is_wrapper. I understand that zetasql is not usable, but we can open backdoors for your self-defined annotation. Introducing the default behavior of mapping wrapper to nullable field would be a breaking change now.

Note that our API actually accepts proto2 instead of proto3 as the schema descriptor, inside of the converter, we look at the default_value field, if it is set then we set the default value, if not set, then we set null. You can surely set it on the ProtoSchema explicitly without having to introduce this mapping.

shollyman: If I'm understanding correctly, you're making users choose between being unable to send nulls, or being unable to send default values (empty string, 0, etc) when communicating the ProtoSchema?

Do you want me to create an issue for supporting wrapper types on the internal converter, or is there one already?  An advantage of using the well known wrapper types is to avoid the need for special annotations, since both the client and backend have the definitions in place as part of the whole protocol buffer ecosystem.

yirutang: In order to support the wrapper, we would need the is_wrapper annotation, otherwise it is a breaking change to existing conversions.

As to your first question, we don't have to make the choice if we are using proto2 (yeah, going back to that question). If it is proto 3 then there seems to be no other choices.

you can create an issue to support the wrapper, you are even welcomed to just go ahead and fix it!

shollyman: This API is still in beta, so its your opportunity to correct this if you think the right final solution is to do it all without annotations.  In such case, you could add a transition option to the ProtoSchema message (e.g. allow_wrapper_types or something).  Much like the discussion around the ProtoDescriptor being a bytes field containing a protodescriptor rather than a direct reference to a ProtoDescriptor.

shollyman: Created b/193064992 for the wrapper issue

yirutang: I think the final solution would be for zetasql to be published and everything conform with the current googlesql specification (instead of creating our own wheels). If that is not possible, adding some fields to ProtoSchema would be fine, less ideal since it would be per schema instead of per msg/field option.

Since you are applying this to all nullable fields, it would be better for the backend to first support it before you add the wrapper conversion, otherwise, the library would be unusable...

bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,259 @@
...
+
+// filename used by well known types proto
+var wellKnownTypesWrapperName = "google/protobuf/wrappers.proto"
+
+// dependencyCache is used to reduce the number of unique messages we generate by caching based on the tableschema.
+//
+// keys are based on the base64-encoded serialized tableschema value.
+type dependencyCache map[string]protoreflect.Descriptor
+
+func (dm *dependencyCache) get(schema *storagepb.TableSchema) protoreflect.Descriptor {
```
codyoss, line 81: nit: does not need to be a pointer as a map is already.

bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	encoded := base64.StdEncoding.EncodeToString(b)
+	(*dm)[encoded] = descriptor
+	return nil
+}
+
+// StorageSchemaToDescriptor builds a protoreflect.Descriptor for a given table schema.
+func StorageSchemaToDescriptor(inSchema *storagepb.TableSchema, scope string) (protoreflect.Descriptor, error) {
+	dc := make(dependencyCache)
+	return storageSchemaToDescriptorInternal(inSchema, scope, &dc)
```
codyoss, line 112: Just noting that currently the error message here will differ from when this is called recursively.

bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,259 @@
...
+		},
+	}
+	for _, d := range deps {
+		fds.File = append(fds.File, protodesc.ToFileDescriptorProto(d))
+	}
+
+	// Load the set into a registry, then interrogate it for the descriptor corresponding to the top level message.
+	files, err := protodesc.NewFiles(fds)
+	if err != nil {
+		return nil, fmt.Errorf("protodesc.NewFiles: %v", err)
```
codyoss, line 219: This error message does not seem to add too much context, should the original error just be returned?

bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,259 @@
...
+
+// tableFieldSchemaToFieldDescriptorProto builds individual field descriptors for a proto message.
+// We're using proto3 syntax, but BigQuery supports the notion of NULLs which conflicts with proto3 default value
+// behavior.  To enable it, we look for nullable fields in the schema that should be scalars, and use the
+// well-known wrapper types.
+//
+// Messages are always nullable, and repeated fields are as well.
+func tableFieldSchemaToFieldDescriptorProto(field *storagepb.TableFieldSchema, idx int32, scope string) (*descriptorpb.FieldDescriptorProto, error) {
+	name := strings.ToLower(field.GetName())
+	if field.GetType() == storagepb.TableFieldSchema_STRUCT {
```
codyoss, line 232: Maybe use a switch here?

bigquery/storage/managedwriter/adapt/protoconversion.go
```diff
@@ -0,0 +1,259 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to serialize tableschema: %v", err)
+	}
+	encoded := base64.StdEncoding.EncodeToString(b)
+	(*dm)[encoded] = descriptor
+	return nil
+}
+
+// StorageSchemaToDescriptor builds a protoreflect.Descriptor for a given table schema.
+func StorageSchemaToDescriptor(inSchema *storagepb.TableSchema, scope string) (protoreflect.Descriptor, error) {
```
codyoss, line 110: If this is all meant to be dynamic do we need to expose this to the user at all? Could we just call this method for them given a table schema? I might just be missing the usecase.

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: done

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: Blerg, thanks for the spot.  Switched.

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: Added a toggle to suppress generation of wrapper types, so we can revisit this once we have a path forward.

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: Perhaps if the logic expands.  Right now the struct case is the significant deviation from the others, which all differ slightly in specific enum values.

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: Specifically, I'm thinking of cases where the user is reading and writing row data and do want to interact directly with a MessageDescriptor via dynamicpb.  This may warrant more helper functions, but I wanted to get back to that after making progress on the writer itself.

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: I was also looking at the `is_wrapper` annotation you mentioned (googlesql MessageOption extension).  I don't think that's going to be the right way to do this, as the expectation is that you "own" the message and can annotate it appropriately.  These types are provided as part of the standard protocol buffer definitions, so adding annotations seems dodgy.  I could see adding a FieldOption extension where you effectively say "the message in this field is a wrapper", but the "this message is a wrapper" annotation seems mismatched for standard types.

bigquery/storage/managedwriter/adapt/protoconversion.go
shollyman: I suspect the right thing to do is to call out the current scope so the user can work out where in the message hierarchy things blew up.  Do you think it's worth having a custom error type for that, or just ensure all the emitted errors implicate the scope?

bigquery/storage/managedwriter/adapt/protoconversion.go
yirutang: Yeah, the annotation should belong to the field.

bigquery/storage/managedwriter/adapt/protoconversion.go
codyoss: I think an unexpected custom error seems reasonable if multiple errors need some context -- scope.

bigquery/storage/managedwriter/adapt/error.go
```diff
@@ -0,0 +1,36 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package adapt
+
+import "fmt"
+
+type ConversionError struct {
```
codyoss, line 19: nit: I don't think users would have any additional action to take with this error so I think it can be unexported.

bigquery/storage/managedwriter/adapt/error.go
```diff
@@ -0,0 +1,36 @@
...
+	Location string
+	Details  error
+}
+
+func (ce *ConversionError) Error() string {
+	if ce.Location == "" {
+		return ce.Details.Error()
+	}
+	return fmt.Sprintf("Location: %s, Error: %v", ce.Location, ce.Details)
```
codyoss, line 28: nit: Start with a a lowercase letter and not need for the word error.
It also might be nice to make the location part more of a phrase. i.e.

```
fmt.Sprintf("unable to convert schema with location %q: %v", ce.Location, ce.Details)
```

bigquery/storage/managedwriter/adapt/error.go
shollyman: Done.

bigquery/storage/managedwriter/adapt/error.go
shollyman: fmt.Sprintf("conversion error in location %q: %v", ce.Location, ce.Details)

codyoss (approved):
LGTM

tyang020:
@shollyman Could you support `packed = true` annotation in repeated FieldOption? We use StorageSchemaToProto2Descriptor and observed 3 bytes overhead per element on repeated fields in the encoded rows. This would be huge overhead to AppendRows throughput for large arrays.

PR #4401, opened 2021-07-08
chore(storage): implement Read with gRPC

noahdietz:
Adds (unexported) first pass implementation of gRPC-based Reader with integration tests based on the Storage v2 API.

This includes:
- Stream reopening/retry
- CRC32 checksum verification
- Integration tests
- local caching of left over response data

Note: This would normally be a `feat` but we don't want to trigger a semver change.


noahdietz:
cc: @danielduhh as an fyi

storage/integration_test.go
```diff
@@ -744,6 +746,127 @@ func TestIntegration_ObjectsRangeReader(t *testing.T) {
...
+	}
+	defer client.Close()
+	bkt := client.Bucket(grpcBucket)
+
+	objName := uidSpace.New()
+	obj := bkt.Object(objName)
+	contents := []byte("Hello, world this is a grpc request")
+
+	if err := retry(ctx, func() error {
+		w := obj.NewWriter(ctx)
```
tritone, line 766: Should we make a separate client to write with so that this does not silently use gRPC once we have writer implemented? Or I guess it would be fine to do both in one test as long as it's clear.

storage/integration_test.go
```diff
@@ -744,6 +746,127 @@ func TestIntegration_ObjectsRangeReader(t *testing.T) {
...
+		t.Error(err)
+	}
+	defer client.Close()
+	bkt := client.Bucket(grpcBucket)
+
+	objName := uidSpace.New()
+	obj := bkt.Object(objName)
+	contents := []byte("Hello, world this is a grpc request")
+
+	if err := retry(ctx, func() error {
```
tritone, line 765: Should be no need to retry with current write logic since the writer itself retries.

storage/integration_test.go
```diff
@@ -744,6 +746,127 @@ func TestIntegration_ObjectsRangeReader(t *testing.T) {
...
+		}
+	}()
+
+	r, err := obj.NewReader(ctx)
+	if err != nil {
+		t.Error(err)
+	}
+	defer r.Close()
+	buf := make([]byte, len(contents))
+	n, err := r.Read(buf)
```
tritone, line 789: Should probably just use io.Copy on the reader.

storage/integration_test.go
```diff
@@ -744,6 +746,127 @@ func TestIntegration_ObjectsRangeReader(t *testing.T) {
...
+
+	r, err := obj.NewReader(ctx)
+	if err != nil {
+		t.Error(err)
+	}
+	defer r.Close()
+
+	bufSize := len(contents)
+	buf := make([]byte, bufSize)
+	chunk := 3
```
tritone, line 848: I see what you're doing here but really I think we need to read a larger file than the message size to verify that the chunking works correctly.

storage/reader.go
```diff
@@ -77,9 +80,76 @@ type ReaderObjectAttrs struct {
...
+	defer func() { trace.EndSpan(ctx, err) }()
+
+	if o.c.gc == nil {
+		err = fmt.Errorf("handle doesn't have a gRPC client initialized")
+		return
+	}
+	if err = o.validate(); err != nil {
+		return
+	}
+	if length < 0 {
```
tritone, line 102: Negative length should mean read until the end of the object.

storage/reader.go
```diff
@@ -77,9 +80,76 @@ type ReaderObjectAttrs struct {
...
+	// Define a function that initiates a Read with offset and length, assuming we
+	// have already read seen bytes.
+	reopen := func(seen int64) (storagepb.Storage_ReadObjectClient, context.CancelFunc, error) {
+		// If the context has already expired, return immediately without making a
+		// call.
+		if err := ctx.Err(); err != nil {
+			return nil, nil, err
+		}
+
+		cc, cancel := context.WithCancel(ctx)
```
tritone, line 121: Does this continue to respect any cancel that was handed out to users? (i assume yes...)

tritone (commented):
Nice start on this! a couple comments...

storage/storage.go
```diff
@@ -158,13 +161,33 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
...
 
+// newClientWithGRPC creates a new Storage client that initializes a gRPC-based client
+// for media upload and download operations.
+func newClientWithGRPC(ctx context.Context, opts ...option.ClientOption) (*Client, error) {
+	c, err := NewClient(ctx, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	g, err := gapic.NewClient(ctx, opts...)
```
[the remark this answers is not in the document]
noahdietz: Good question. I thought they did, but turns out [they don't](https://github.com/googleapis/google-api-go-client/blob/896f71cdd1e1bb6aca6dcba186a0b401ad44bbee/transport/grpc/dial.go#L112).

Honestly, I think unsupported options should just be ignored...but maybe @codyoss has some thoughts? We will have to think of something here because the GAPIC will only be used for media operations and the HTTP client for metadata operations. I guess the factory method should have as separate `grpcOpts option.ClientOption` field, which kind of sucks and precludes making it variadic (only one of the two will be able to be variadic).

storage/reader.go
```diff
@@ -77,9 +80,76 @@ type ReaderObjectAttrs struct {
 //
 // The caller must call Close on the returned Reader when done reading.
 func (o *ObjectHandle) NewReader(ctx context.Context) (*Reader, error) {
+	if o.c.gc != nil {
+		return o.newRangeReaderWithGRPC(ctx, 0, 0)
+	}
 	return o.NewRangeReader(ctx, 0, -1)
```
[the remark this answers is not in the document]
noahdietz: Yep! Good catch.

storage/integration_test.go
noahdietz: Thanks, I just copied the code from another test, but I'll remove that stuff here.

storage/integration_test.go
noahdietz: Great idea. I figured I'd punt on that to when we implemented Write w/gRPC, but might as well start off correct.

storage/reader.go
noahdietz: Ok, I will coerce it to `0` to read to the end instead of returning an error. Supplying a negative value will [result in an error](https://github.com/googleapis/googleapis/blob/master/google/storage/v2/storage.proto#L128-L129) for the grpc API.

storage/reader.go
noahdietz: If i understand your question, then yes. If the user has a `CancelFunc` for the context they gave the reader, it will still work as expected. And inversely, canceling this derived context will not cancel the user provided context. LMK if I misunderstood.

storage/reader.go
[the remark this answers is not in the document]
noahdietz: Moved this to NewRangeReader entirely b.c NewReader just calls through to NewRangeReader. Done.

storage/reader.go
noahdietz: Coercing length to `0` when it is negative to indicate "read to the end". Done.

storage/storage.go
```diff
@@ -95,6 +96,8 @@ type Client struct {
 	envHost string
 	// ReadHost is the default host used on the reader.
 	readHost string
+
```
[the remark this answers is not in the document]
noahdietz: Done!

storage/storage.go
```diff
@@ -158,13 +161,33 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 	}, nil
 }
 
+// newClientWithGRPC creates a new Storage client that initializes a gRPC-based client
+// for media upload and download operations.
```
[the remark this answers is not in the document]
noahdietz: Done!

storage/integration_test.go
noahdietz: Removed and refactored into a separate function that creates a new HTTP client. Done!

storage/integration_test.go
noahdietz: Removed and refactored into a separate function that creates a new HTTP client. Done!

storage/integration_test.go
noahdietz: Done!

storage/integration_test.go
noahdietz: Agreed. How would you like to see multi-megabyte test data be generated? Should I just write a bunch of random content in a loop until we hit a good size and then expect that the same amount was read?

storage/reader.go
tritone: Yep that's what I meant! Just thought I'd double check.

storage/reader.go
tritone: Yeah, I don't think this is a thing in HTTP either, but it's a convention used in this library-- see the NewRangeReader docs https://pkg.go.dev/cloud.google.com/go/storage#ObjectHandle.NewRangeReader

storage/integration_test.go
tritone: There are various clever ways to do this in Go. I remember in another test I used [bytes.Repeat](https://pkg.go.dev/bytes#Repeat) to make a larger buffer. Obviously if you want more content than can be held in memory, you need to take a different approach, but this is a simple one anyway.

storage/storage.go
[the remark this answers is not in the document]
codyoss: Are there real examples where providing options here for both would cause issues? I know we do have some validation rules around not mixing some options: https://github.com/googleapis/google-api-go-client/blob/896f71cdd1e1bb6aca6dcba186a0b401ad44bbee/internal/settings.go#L115-L123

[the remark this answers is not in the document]
noahdietz: @tritone do you know if there are some HTTP specific options folks use?

That is the code I was thinking about specifically, Cody. 

Do you think we could add some helper to google.golang.org/api/option that can split a slice of options into http & grpc options? Seems a little weird I think, but could make this a non-issue.

`func TransportOptions(opts ...ClientOption) (hOpts []ClientOption, gOpts []ClientOption)`

Those options that would work for both could be included in both.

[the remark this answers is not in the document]
codyoss: I think we could do something like that, but if we did I think it would be in `internaloption` as I don't think a consumer of this API would ever need to interact with that. 

I think there are some options that might overlap and not play nice. For example changing an endpoint. I think we need a way to change both two different values, right? If so, a new method that takes a config struct that can separate options might be in order instead.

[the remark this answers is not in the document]
noahdietz: > I think there are some options that might overlap and not play nice. For example changing an endpoint. 

Yeah In my experience, use of `WithEndpoint` requires a scheme prefix, while gRPC doesn't allow that. That would be tricky.

> a new method that takes a config struct that can separate options might be in order instead.

Yeah, something like this if I'm following?

```go
// Could be in storage, could be in google.golang.org/api/option
type ClientOptions struct {
  // hOpts is the set of options used for a HTTP client.
  hOpts []option.ClientOption

  // gOpts is the set of options used for a gRPC client.
  gOpts []option.ClientOption
}


// In storage...could make it variadic to save the user passing in `nil`, but that would require merging all structs
func NewClientWithGRPC(ctx context.Context, opts *ClientOptions) (Client, error)
```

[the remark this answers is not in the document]
codyoss: Yup, it would be kinda annoying if you are not using auto-detection with ADC to pass auth options twice, but I think that is the only way to get it right 100% of the time.

storage/integration_test.go
noahdietz: `bytes.Repeat` is super useful. 

> Obviously if you want more content than can be held in memory,

WDYM? I've not done this before so, idk what is acceptable. Could I make a 2+ MB buffer in mem using `bytes.Repeat` to upload then read? Or are you recommending writing 2+ MB worth of data using smaller buffers, then reading it all?

storage/integration_test.go
```diff
@@ -759,8 +759,16 @@ func TestIntegration_ObjectReadGRPC(t *testing.T) {
 
 	content := []byte("Hello, world this is a grpc request")
 
-	var name string
-	if name, err = writeTestData(ctx, grpcBucket, content); err != nil {
+	// Create an HTTP client to upload test data.
+	hc, err := NewClient(ctx)
```
tritone, line 763: Use [`testConfig`](https://github.com/googleapis/google-cloud-go/blob/e36b29a3d43bce5c1c044f7daf6e1db00b0a49e0/storage/integration_test.go#L192) for client creation (actually, we should make a version of this method that creates a gRPC client as well).

storage/integration_test.go
```diff
@@ -804,14 +812,22 @@ func TestIntegration_ObjectReadChunksGRPC(t *testing.T) {
 	ctx := context.Background()
 	client, err := newClientWithGRPC(ctx)
 	if err != nil {
-		t.Error(err)
+		t.Fatal(err)
 	}
 	defer client.Close()
 
-	content := []byte("Hello, world this is a grpc request")
+	content := bytes.Repeat([]byte("a"), 4<<10)
```
tritone, line 819: Yep, this is what I had in mind as far as using bytes.Repeat-- maybe add a comment noting how large this file is and why it is this size.

storage/integration_test.go
noahdietz: Done! I haven't done the grpc-flavored `testConfig`, but I put the HTTP `testConfig` call ahead of grpc client creation so that it'll skip if necessary. Can we do the grpc-flavored `testconfig` in a follow on PR?

storage/integration_test.go
noahdietz: Done! I do think we need an even bigger blob though...this isn't big enough atm to trigger backend chunking on read (i.e. multiple stream Recv calls)

tritone: Yep, let's do something like 5MB maybe? The test should still be pretty quick.

storage/integration_test.go
tritone: Sure, that's fine!

storage/storage.go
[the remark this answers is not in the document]
noahdietz: I've added an unexported `clientOptions` type for this.

storage/integration_test.go
noahdietz: Done and it works it seems! Test data is ~5MB and reading it with a ~4KB window.

noahdietz:
@codyoss @tritone what else do you want to see to get this submitted? I've been testing these changes locally against the V1 API, because the v2 API isn't rolled out yet so the tests are skipped atm.

codyoss:
@noahdietz TBH I have not looked at this PR, yet. I will give it a look tmr. Along the same lines as client options I wonder if by adding this we are going to break people? If a person is passing in an endpoint override today it will now not apply to grpc stuff, correctly. Similar if they are doing fancy things and mucking with the HTTP transport to change proxy settings or logging middleware perhaps. I have not thought about what this would mean for this client, but when I take a look tmr these are some thoughts going through my head.

noahdietz:
> TBH I have not looked at this PR, yet

No problem, I just realized that I cc'd you in to get your opinion on one thing haha

> Along the same lines as client options I wonder if by adding this we are going to break people? 

These changes don't actually change the normal behavior of the client. `newClientWithGRPC` is actually not referenced anywhere except for in tests. We'd end up adding a new factory `NewClientWithGRPC` when we want to export it I think. So I'd say its not a concern atm.

noahdietz:
cc: @coryan would you mind doing a once over of this PR? You've said you aren't Go fluent, but your depth of knowledge with Storage client behavior would be useful here. LMK what you think. Thanks.

storage/reader.go
```diff
@@ -392,6 +525,82 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+				return n, nil
+			}
+
+			// Do not retry PermissionDenied.
+			if st, ok := status.FromError(err); ok && st.Code() == codes.PermissionDenied {
+				return n, err
+			}
+
+			// TODO: How many times do we try to read before stopping?
+			// TODO: Should we backoff?
```
coryan, line 570: Yes, you should backoff.

storage/reader.go
```diff
@@ -392,6 +525,82 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+				// Let the next call to readWithGRPC return EOF.
+				return n, nil
+			}
+
+			// Do not retry PermissionDenied.
+			if st, ok := status.FromError(err); ok && st.Code() == codes.PermissionDenied {
+				return n, err
+			}
+
+			// TODO: How many times do we try to read before stopping?
```
coryan, line 569: Presumably `ReadWithRetry` already has an answer to that question, shouldn't you use the same answer?

storage/reader.go
```diff
@@ -392,6 +525,82 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+	for len(p[n:]) > 0 {
+		msg, err := r.stream.Recv()
+		if err != nil {
+			if err == io.EOF && r.seen == r.size {
+				// Let the next call to readWithGRPC return EOF.
+				return n, nil
+			}
+
+			// Do not retry PermissionDenied.
+			if st, ok := status.FromError(err); ok && st.Code() == codes.PermissionDenied {
```
coryan, line 565: There are many other things you should not retry, e.g., `FailedPrecondition` or `Aborted`.  I would guess the go library already has an answer for "what is retryable" that we could reuse?

storage/reader.go
```diff
@@ -392,6 +525,82 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+			if err != nil {
+				// Cannot reinstantiate the stream so return the original error.
+				return n, e
+			}
+
+			// Try to Recv again.
+			continue
+		}
+
+		content := msg.GetChecksummedData().GetContent()
```
coryan, line 588: Consider a TODO or some other indication that we can add CRC32C checksums here.

storage/reader.go
```diff
@@ -392,6 +525,82 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+		// No more room left in the user's buffer or the entire object has been
+		// read.
+		if len(p[cp:]) == 0 || r.size == r.seen {
+			return cp, nil
+		}
+
+		n = cp
+	}
+
+	for len(p[n:]) > 0 {
```
coryan, line 556: Why is the termination condition for this while loop different from the `if()` on line 549?

storage/reader.go
```diff
@@ -80,6 +83,110 @@ func (o *ObjectHandle) NewReader(ctx context.Context) (*Reader, error) {
...
+	reopen := func(seen int64) (storagepb.Storage_ReadObjectClient, context.CancelFunc, error) {
+		// If the context has already expired, return immediately without making a
+		// call.
+		if err := ctx.Err(); err != nil {
+			return nil, nil, err
+		}
+
+		cc, cancel := context.WithCancel(ctx)
+
+		start := offset + seen
```
coryan, line 125: Does the golang library support reading from a negative offset?  It is possible to ask GCS to read "starting at byte -200" which really means "read the lasts 200 bytes".  If that is supported, reading from "offset + 200" is not the same as reading from "-0", I would leave it up to you to figure this out.

storage/reader.go
```diff
@@ -80,6 +83,110 @@ func (o *ObjectHandle) NewReader(ctx context.Context) (*Reader, error) {
...
+	}
+
+	// A negative length means "read to the end of the object", but the
+	// read_limit field it corresponds to uses zero to mean the same thing. Thus
+	// we coerce the length to 0 to read to the end of the object.
+	if length < 0 {
+		length = 0
+	}
+
+	req := &storagepb.ReadObjectRequest{
```
coryan, line 109: Does the golang library support pre-conditions (`ifMetagenerationMatch`, `ifGenerationMatch`) or reading an specific generation of an object?  Do you want to support those for gRPC too?

storage/integration_test.go
```diff
@@ -744,6 +746,129 @@ func TestIntegration_ObjectsRangeReader(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer r.Close()
+
+	bufSize := len(content)
+	buf := make([]byte, bufSize)
+
+	// Read in 4KB chunks.
+	chunk := 4 << 10
```
coryan, line 851: You may want to read in sizes that are *not* a divisor of the received chunk size, just to make sure you can read across a `recv()` boundary.

coryan (commented):
A bunch of nits and comments that could be addressed in future PRs.

storage/reader.go
noahdietz: @tritone ?

storage/reader.go
noahdietz: `if len(r.leftovers) > 0 {` just checks to see if the client's internal buffer has any data stashed in it to use for filling the user-provided buffer. `for len(p[n:]) > 0 {` will continue `Recv`'ing until the user-provided buffer has been filled.

Did that answer the question?

storage/reader.go
noahdietz: @tritone can you point us to where this may be defined?

storage/reader.go
noahdietz: Yep, done!

storage/reader.go
coryan: Your code is probably correct, it just looked odd to me.  Basically I wondered: if there is reason to believe that you should return when `r.size == r.seen` in line 549: do those reasons apply here?  Do you even need that `if()` in line 549?  Since you looked at the code twice, I am fine either way, you know better.

storage/integration_test.go
noahdietz: Good idea. Done.

storage/reader.go
noahdietz: I went ahead and just tried removing the conditional and everything still worked. You were right; the `leftovers` will either have nothing (`copy` returns `0`) and the following code just executes no problem with `0`, or it has something and we use it. Since `leftovers` is filled with the left over bytes from the previous `Recv` msg, if the user's buffer has more space, we should try to `Recv` again. 

Thanks Carlos! Definitely cleaned this up.

storage/reader.go
noahdietz: > Does the golang library support reading from a negative offset?

Yes it should. I had to make some code changes to support this, so thank you. Initializing the `Reader.size` and `Reader.remain` to the `abs(offset)` so that the reader will stop once it has seen that `size`. Haven't pushed the changes yet b.c I'm having trouble with CRCs now.

> "starting at byte -200"...reading from "offset + 200" is not the same as reading from "-0"

I don't think this will be an issue because of the aforementioned logic. We should not attempt to reopen if the `Reader.size` and `Reader.seen` are equal. In your example, the `Reader.size` would be `200` and the `Reader.seen` should never surpass `200`. I could put a check in there, or a note warning of this, but I _think_ the logic is sound.

noahdietz: Ok, I've pushed the changes with a test.

storage/reader.go
noahdietz: I've added support for grpc to the existing `shouldRetry` predicate and started using that machinery where applicable.

storage/reader.go
noahdietz: Worked offline, this is now implemented!

storage/storage.go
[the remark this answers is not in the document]
noahdietz: @broady could you weigh in?

storage/reader.go
tritone: Discussed this offline; retries will continue for retryable errors until the user cancels the context.

noahdietz: Worked this out offline. Done!

storage/reader.go
noahdietz: reopening will now backoff. done!

storage/integration_test.go
noahdietz: I added a `testConfigGRPC` that creates both HTTP and gRPC clients.

codyoss (commented):
LGTM to me overall, but I am no storage expert so I defer to Chris.

coryan:
Why "chore(storage)"?  This is a fun feature, not a "chore" 😀

noahdietz:
> Why "chore(storage)"? This is a fun feature, not a "chore" 😀

Indeed it is fun and exciting :) but using `feat` would trigger a semver minor update to the Go module, and `fix` would be a patch update. `chore` doesn't elicit an automatic version bump, so we will use that until we export the feature. Then it will be `feat` with lots of hoopla 😃

storage/reader.go
```diff
@@ -349,6 +485,25 @@ func setConditionsHeaders(headers http.Header, conds *Conditions) error {
 	return nil
 }
 
+func setRequestConditions(req *storagepb.ReadObjectRequest, conds *Conditions) {
```
tritone, line 488: Let's separate out these utility functions for gRPC and put them at the bottom (with comments marking them as non-production at this point)

storage/reader.go
```diff
@@ -392,6 +574,72 @@ func (r *Reader) Read(p []byte) (int, error) {
 	return n, err
 }
 
+func (r *Reader) readWithGRPC(p []byte) (int, error) {
+	// No stream to read from, either never initiliazed or Close was called.
+	// TODO: should an error be returned here? "reader has been closed"?
```
tritone, line 579: Yeah we should return an error probably; EOF seems bad because it implies that we finished reading the object. Or if "never initialized" somehow we'd want to try to reopen the stream probably?

storage/reader.go
```diff
@@ -392,6 +574,72 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+	if r.size != 0 && r.size == r.seen {
+		return 0, io.EOF
+	}
+
+	n := copy(p, r.leftovers)
+	r.seen += int64(n)
+	r.leftovers = r.leftovers[n:]
+
+	for len(p[n:]) > 0 {
+		msg, err := r.stream.Recv()
```
tritone, line 594: So it looks like our strategy here is to try to fill up `p`. In the Reader interface it says "If some data is available but not len(p) bytes, Read conventionally returns what is available instead of waiting for more." I'm wondering if we should prefer this strategy (which would mean putting off the new Recv call unless leftovers is empty). What do you think?

storage/reader.go
```diff
@@ -392,6 +574,72 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+			// Close existing stream and initialize new stream with updated
+			// offset.
+			r.closeStream()
+
+			// This will immediately attempt to reopen the stream, but will
+			// backoff if further attempts are necessary.
+			var res *readStreamResponse
+			res, r.cancelStream, err = r.reopenWithGRPC(r.seen)
+			if err != nil {
+				// Cannot reinstantiate the stream so return the original error.
```
tritone, line 618: Should we wrap these perhaps? Is there any use in preserving the reopen error?

storage/reader.go
```diff
@@ -392,6 +574,72 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+				return n, e
+			}
+			r.stream = res.stream
+			// Reopened the stream and Recv'd the first message, so carry on.
+			msg = res.response
+		}
+
+		// TODO: Capture incremental CRC32C for this chunk.
```
tritone, line 626: Can we make a longer TODO about what we want to do for per-chunk checksums?

storage/reader.go
```diff
@@ -392,6 +574,72 @@ func (r *Reader) Read(p []byte) (int, error) {
...
+		content := msg.GetChecksummedData().GetContent()
+		cp := copy(p[n:], content)
+		leftover := len(content) - cp
+		if leftover > 0 {
+			// Wasn't able to copy all of the data in the message, store for
+			// future Read calls.
+			r.leftovers = make([]byte, leftover)
```
tritone, line 633: Memory performance wise, do you think it would be problematic to be re-making this buffer (as a custom length) rather than re-using a buffer or having a pool? I don't think this is something we need to address now but maybe another TODO.

tritone (commented):
Looking good mostly! A bunch of comments but I am being picky :D

storage/integration_test.go
```diff
@@ -744,6 +761,197 @@ func TestIntegration_ObjectsRangeReader(t *testing.T) {
...
+		}
+		if err != nil {
+			t.Fatal(err)
+		}
+		offset += n
+	}
+
+	got := string(buf)
+	want := string(content)
+	if diff := cmp.Diff(got, want); diff != "" {
```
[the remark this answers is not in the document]
noahdietz: Yep, indeed. Done!

storage/integration_test.go
```diff
@@ -203,6 +205,21 @@ func testConfig(ctx context.Context, t *testing.T) *Client {
 	return client
 }
 
+// testConfigGPRC returns both a JSON/HTTP-based client and a gRPC-based client
+// to access GCS. testConfigGRPC skips the curent test if credentials are not
+// available or when being run in Short mode.
+func testConfigGRPC(ctx context.Context, t *testing.T) (hc *Client, gc *Client) {
```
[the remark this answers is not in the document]
noahdietz: Sure thing. Done!

storage/storage.go
```diff
@@ -157,13 +163,44 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 	}, nil
 }
 
+// clientOptions carries the set of client options for HTTP and gRPC clients.
```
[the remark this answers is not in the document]
noahdietz: Good idea. Done.

storage/storage.go
```diff
@@ -157,13 +163,44 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
...
+type clientOptions struct {
+	HTTPOpts []option.ClientOption
+	GRPCOpts []option.ClientOption
+}
+
+// newClientWithGRPC creates a new Storage client that initializes a gRPC-based client
+// for media upload and download operations.
+//
+// This is an experimental API and not intended for public use.
+func newClientWithGRPC(ctx context.Context, opts *clientOptions) (*Client, error) {
```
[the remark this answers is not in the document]
noahdietz: Yep, done!

storage/reader.go
noahdietz: Good point. I think we should follow up on it and I've added a TODO. The implementation is simple atm and just shrinks `leftovers` by making subslices as the data is copied over to the read buffer. But we could make use of a "leftovers cursor" instead, and resuse the buffer once we've copied it all, expanding it if necessary.

storage/reader.go
noahdietz: Done! I don't really know how to do this or if it is necessary, because we already have full Object checksum verification in this PR, but I've expanded this comment nonetheless :)

storage/reader.go
noahdietz: I'm going to have it return an error. I don't want to reopen b.c part of the `Close` implementation here will set `r.stream` to `nil` in order to have it be GC'd. So if `r.stream == nil`' the user called `Close` and a subsequent `Read` should be an error.

However there is a possible race condition in the event there are multiple consumers of the Reader and one call to Read encounters an error on `Recv` and the stream is to be reopened. If another consumer calls Read on the same Reader, the stream will either be `nil` or not functional/closed (by the backend). Is that scenario a concern @tritone ?

storage/reader.go
coryan: drive-by: the object checksums don't help for partial downloads (ranges, offsets, etc.) the per-chunk checksums would detect bit errors in that case.

storage/reader.go
tritone: Ah yeah that makes sense. As for the race condition, I think we don't make any thread safety guarantees about the current Reader implementation either -- it would be really weird to have multiple go routines reading from the same reader actually. Maybe we should note this explicitly in the docs.

storage/reader.go
tritone: Sounds good!

storage/reader.go
noahdietz: Good idea. I've returned early after exhausting the `leftovers`. It doesn't change much in the way of buffer management, but it does make understanding the code easier, and I like conforming to the documented Reader semantics.

storage/reader.go
noahdietz: I can wrap it, yes. 

As for the reopen error, I'm not sure. Attempting to `Recv` from the stream should've produced the "real" error  (e.g. BadRequest), so I don't want to lose that. But if we attempt to reopen and _that_ fails, maybe that is useful to know as well. Should be return both? `fmt.errorf("reading stream failed: %s and stream reopen failed: %s", e, err)`

storage/go.mod
```diff
@@ -12,4 +12,5 @@ require (
 	google.golang.org/api v0.52.0
 	google.golang.org/genproto v0.0.0-20210805201207-89edb61ffb67
 	google.golang.org/grpc v1.39.1
+	google.golang.org/protobuf v1.27.1
```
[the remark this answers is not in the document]
noahdietz: There are already dependencies in this `go.mod` that transitively bring in `google.golang.org/protobuf` (genproto and grpc), so I'm not sure this is an issue, but @codyoss please confirm.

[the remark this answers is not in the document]
codyoss: Yeah this dependency is already there because of how our code is structured today sadly. Tested by making a simple file:

```
package main

import (
	"context"

	"cloud.google.com/go/storage"
)

func main() {
	storage.NewClient(context.Background())
}
```

Then I ran `go mod vendor`. Sources for `google.golang.org/protobuf` are already present.

storage/reader.go
tritone: Nice. I think in this case we no longer need a loop here either, right? We can just `Recv` once and if that doesn't fill the buffer up, that's fine.

storage/reader.go
noahdietz: No longer necessary due to refactoring of Recv with retry. Spoke offline.

storage/reader.go
noahdietz: You are right! Except we need _some_ loop to Retry the Recv. I've refactored the Recv into `recvWithRetry`.

tritone: Actually I think this is incorrect-- `reopen` itself does a retry on the `Recv()`. So we shouldn't wrap reopen in a retry as well.

noahdietz: Oops you are correct. Fixed!

storage/integration_test.go
```diff
@@ -203,6 +205,22 @@ func testConfig(ctx context.Context, t *testing.T) *Client {
...
+// skips the curent test when being run in Short mode.
+func testConfigGRPC(ctx context.Context, t *testing.T) (gc *Client) {
+	if testing.Short() {
+		t.Skip("Integration tests skipped in short mode")
+	}
+
+	var err error
+	gc, err = newHybridClient(ctx, nil)
+	if err != nil {
+		log.Fatalf("newClientWithGRPC: %v", err)
```
tritone, line 218: Should be `t.Fatalf`

storage/integration_test.go
```diff
@@ -203,6 +205,22 @@ func testConfig(ctx context.Context, t *testing.T) *Client {
...
 }
 
+// testConfigGPRC returns a gRPC-based client to access GCS. testConfigGRPC
+// skips the curent test when being run in Short mode.
+func testConfigGRPC(ctx context.Context, t *testing.T) (gc *Client) {
+	if testing.Short() {
+		t.Skip("Integration tests skipped in short mode")
+	}
+
+	var err error
```
tritone, line 215: nit: don't need this on a separate line.

tritone (commented):
Couple additional comments on testing. This is looking really good aside from my other few open comments!

storage/integration_test.go
noahdietz: Done!

storage/integration_test.go
noahdietz: Done!

storage/integration_test.go
```diff
@@ -61,6 +62,7 @@ const (
 	// TODO(jba): move to testutil, factor out from firestore/integration_test.go.
 	envFirestoreProjID     = "GCLOUD_TESTS_GOLANG_FIRESTORE_PROJECT_ID"
 	envFirestorePrivateKey = "GCLOUD_TESTS_GOLANG_FIRESTORE_KEY"
+	grpcBucket             = "golang-grpc-test-integration"
```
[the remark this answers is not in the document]
noahdietz: Yep, good call. I've changed things to create a new bucket each test run with our prefix and use that.

storage/storage.go
```diff
@@ -1642,3 +1679,8 @@ func (c *Client) ServiceAccount(ctx context.Context, projectID string) (string,
 	}
 	return res.EmailAddress, nil
 }
+
+// bucket formats the given project ID and bucket ID into a Bucket name.
```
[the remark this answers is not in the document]
noahdietz: Good idea. This is a "newer" API design practice. Done.

storage/reader.go
```diff
@@ -80,6 +82,136 @@ func (o *ObjectHandle) NewReader(ctx context.Context) (*Reader, error) {
...
+	}
+
+	// For now, there are only globally unique buckets, and "_" is the alias
+	// project ID for such buckets.
+	b := bucket("_", o.bucket)
+	req := &storagepb.ReadObjectRequest{
+		Bucket: b,
+		Object: o.object,
+	}
+	if o.gen >= 0 {
```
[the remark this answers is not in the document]
noahdietz: Adding a comment but we resolved this offline.

storage/reader.go
```diff
@@ -80,6 +82,136 @@ func (o *ObjectHandle) NewReader(ctx context.Context) (*Reader, error) {
...
+			}
+
+			msg, err = stream.Recv()
+
+			return err
+		})
+		if err != nil {
+			// Close the stream context we just created to ensure we don't leak
+			// resources.
+			cancel()
```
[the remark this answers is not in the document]
noahdietz: Yep, this is how the client closes their end of the stream for a server-stream RPC

storage/reader.go
```diff
@@ -80,6 +82,136 @@ func (o *ObjectHandle) NewReader(ctx context.Context) (*Reader, error) {
...
+		return nil, err
+	}
+
+	r = &Reader{
+		stream:         res.stream,
+		reopenWithGRPC: reopen,
+		cancelStream:   cancel,
+	}
+
+	// Recv the first message so we can populate the object metadata.
```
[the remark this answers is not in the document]
noahdietz: Yep, updating the comment.

storage/reader.go
```diff
@@ -365,15 +520,42 @@ type Reader struct {
 	wantCRC            uint32 // the CRC32c value the server sent in the header
 	gotCRC             uint32 // running crc
 	reopen             func(seen int64) (*http.Response, error)
+
+	stream         storagepb.Storage_ReadObjectClient
```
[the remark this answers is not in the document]
noahdietz: Adding a comment! And this specific field is an interface so we don't need to also make it a pointer.

storage/reader.go
noahdietz: I've moved everything gRPC related to a block at the bottom.

storage/reader.go
```diff
@@ -80,6 +82,136 @@ func (o *ObjectHandle) NewReader(ctx context.Context) (*Reader, error) {
...
+	}
+
+	// Recv the first message so we can populate the object metadata.
+	msg := res.response
+	obj := msg.GetMetadata()
+	size := obj.GetSize()
+
+	// If offset is negative, we read the remaining bytes of content starting
+	// at the offset relative to the end. Thus the absolute value of the offset
+	// is the size of the portion to be read.
```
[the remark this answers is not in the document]
noahdietz: Good catch! Size was ambiguous here and as we discussed offline, it is the size of the content to be read by the reader. Done.

storage/reader.go
noahdietz: Adding a note about this.

storage/storage.go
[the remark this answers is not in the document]
noahdietz: We've resolve this offline and are sticking with the struct.

tritone (approved):
After a bunch of wrangling, this LGTM. Thanks for all your work on this @noahdietz !!

@coryan @codyoss @broady feel free to take another look; if we don't hear from you we'll merge tomorrow afternoon.

PR #4403, opened 2021-07-09
feat(spanner/spansql): support ALTER DATABASE

### LINKED ISSUE #4390: spanner/spansql: Support ALTER DATABASE (SET OPTIONS) statements

Some softwares use `spansql` package as a DDL parser but it doesn't support `ALTER DATABASE` statements.
If the DDL contains `ALTER DATABASE`, parsing will fail.
 
https://cloud.google.com/spanner/docs/data-definition-language?hl=en#alter-database
```
ALTER DATABASE `dbname` SET OPTIONS (enable_key_visualizer=true, optimizer_version=2, version_retention_period='7d')
```

spansql package should support `ALTER DATABASE` with `SET OPTIONS`.

olavloite:
Adds support for ALTER DATABASE dbname SET OPTIONS (..) statements.

Resolves #4390


spanner/spansql/parser.go
```diff
@@ -1286,6 +1289,50 @@ func (p *parser) parseAlterTable() (*AlterTable, *parseError) {
 	}
 }
 
+func (p *parser) parseAlterDatabase() (*AlterDatabase, *parseError) {
+	debugf("parseAlterDatabase: %v", p)
+
+	/*
+			ALTER DATABASE database_id
```
hengfengli, line 1296: The indentation for this entire comment is a bit weird. Please take a look at `parseAlterTable`.

spanner/spansql/parser.go
```diff
@@ -1286,6 +1289,50 @@ func (p *parser) parseAlterTable() (*AlterTable, *parseError) {
...
+
+	if err := p.expect("ALTER"); err != nil {
+		return nil, err
+	}
+	pos := p.Pos()
+	if err := p.expect("DATABASE"); err != nil {
+		return nil, err
+	}
+	// The following method is also valid for database names.
+	dbname, err := p.parseTableOrIndexOrColumnName()
```
hengfengli, line 1313: It's a bit different, but I think we should bother with that. The database ID cannot be ended with underscore. Also, it supports hyphens. https://cloud.google.com/spanner/docs/data-definition-language#database-id-names

spanner/spansql/parser.go
```diff
@@ -1286,6 +1289,50 @@ func (p *parser) parseAlterTable() (*AlterTable, *parseError) {
...
+		return nil, err
+	}
+	a := &AlterDatabase{Name: dbname, Position: pos}
+
+	tok := p.next()
+	if tok.err != nil {
+		return nil, tok.err
+	}
+	switch {
+	default:
```
hengfengli, line 1324: nits: can we move default to the last one in `switch`?

spanner/spansql/parser.go
```diff
@@ -1495,6 +1542,98 @@ func (p *parser) parseColumnOptions() (ColumnOptions, *parseError) {
 	return co, nil
 }
 
+func (p *parser) parseDatabaseOptions() (DatabaseOptions, *parseError) {
+	debugf("parseDatabaseOptions: %v", p)
+	/*
+			options_def:
```
hengfengli, line 1548: Same here. I think it should move a tab left.

spanner/spansql/parser.go
```diff
@@ -1495,6 +1542,98 @@ func (p *parser) parseColumnOptions() (ColumnOptions, *parseError) {
...
+	// We ignore case for the key (because it is easier) but not the value.
+	var opts DatabaseOptions
+	for {
+		if p.eat("enable_key_visualizer", "=") {
+			tok := p.next()
+			if tok.err != nil {
+				return DatabaseOptions{}, tok.err
+			}
+			enableKeyVisualizer := new(bool)
+			switch tok.value {
```
hengfengli, line 1570: if the case is `false`, what would happen here?

spanner/spansql/parser.go
```diff
@@ -1495,6 +1542,98 @@ func (p *parser) parseColumnOptions() (ColumnOptions, *parseError) {
...
+				return DatabaseOptions{}, p.errorf("invalid enable_key_visualizer_value: %v", tok.value)
+			}
+			opts.EnableKeyVisualizer = enableKeyVisualizer
+		} else if p.eat("optimizer_version", "=") {
+			tok := p.next()
+			if tok.err != nil {
+				return DatabaseOptions{}, tok.err
+			}
+			optimizerVersion := new(int)
+			switch tok.value {
```
hengfengli, line 1585: TBH, is `if-else` better to replace this switch? `default` has too many logics.

spanner/spansql/parser.go
```diff
@@ -1495,6 +1542,98 @@ func (p *parser) parseColumnOptions() (ColumnOptions, *parseError) {
...
+				optimizerVersion = &version
+			}
+			opts.OptimizerVersion = optimizerVersion
+		} else if p.eat("version_retention_period", "=") {
+			tok := p.next()
+			if tok.err != nil {
+				return DatabaseOptions{}, tok.err
+			}
+			retentionPeriod := new(string)
+			switch tok.value {
```
hengfengli, line 1605: Same here. Use `if-else` to replace `switch`.

spanner/spansql/types.go
```diff
@@ -208,6 +208,37 @@ const (
...
+func (*AlterDatabase) isDDLStmt()        {}
+func (ad *AlterDatabase) Pos() Position  { return ad.Position }
+func (ad *AlterDatabase) clearOffset()   { ad.Position.Offset = 0 }
+
+type DatabaseAlteration interface {
+	isDatabaseAlteration()
+	SQL() string
+}
+
+func (SetDatabaseOptions) isDatabaseAlteration() {}
```
hengfengli, line 230: Can we move this after the definition of `SetDatabaseOptions`?

spanner/spansql/sql_test.go
```diff
@@ -230,6 +230,32 @@ func TestSQL(t *testing.T) {
...
+			&AlterDatabase{
+				Name: "dbname",
+				Alteration: SetDatabaseOptions{Options: DatabaseOptions{
+					VersionRetentionPeriod: func(s string) *string { return &s }("7d"),
+					OptimizerVersion:       func(i int) *int { return &i }(2),
+					EnableKeyVisualizer:    func(b bool) *bool { return &b }(true),
+				}},
+				Position: line(1),
+			},
+			"ALTER DATABASE dbname SET OPTIONS (optimizer_version=2, version_retention_period='7d', enable_key_visualizer=true, )",
```
hengfengli, line 243: Do we need to remove the trailing `,`? Is this statement valid?

hengfengli (changes requested):
LGTM. Thanks for making this change. Just left a few comments.

spanner/spansql/parser.go
olavloite: I've added a TODO for it. Currently, spansql does not apply these restrictions for any identifiers, and it would make more sense to pick that up for all identifiers in a separate PR.

spanner/spansql/parser.go
olavloite: Done.

spanner/spansql/parser.go
olavloite: I would prefer to keep it as the first element in the switch for the sake of consistency. This is how all parsing is implemented in spansql. See for example [ALTER TABLE](https://github.com/googleapis/google-cloud-go/blob/2cb9ad72416c84742cca4073e7583d4e0b02cb4f/spanner/spansql/parser.go#L1217).

This structure also makes it clear at the start of the switch block what can be expected in the rest of the block.

spanner/spansql/parser.go
olavloite: Done

spanner/spansql/parser.go
olavloite: It will fall through to the default statement, which indicates that it is an invalid value, so it will return an error. The documentation currently does not include this option, so I'm actually not 100% sure what the valid values are, but [looking at the `allow_commit_timestamp` boolean option for columns](https://cloud.google.com/spanner/docs/data-definition-language#alter_table) it seems that boolean options (with `false/null` as the default value) only accept `true` or `null` as values.

spanner/spansql/parser.go
olavloite: Yes, that is indeed better. Changed to `if-else`.

spanner/spansql/parser.go
olavloite: Done

spanner/spansql/types.go
olavloite: Done

spanner/spansql/sql_test.go
olavloite: You're right, Cloud Spanner does not support a trailing comma in an options list. I've updated that here and in the parsing and SQL printing.

olavloite:
> LGTM. Thanks for making this change. Just left a few comments.

Thanks for the review. I've addressed / answered your comments. PTAL.

hengfengli (approved):

PR #4405, opened 2021-07-09
feat(bigquery): add more dml statistics to query statistics

shollyman:
This adds the new DMLStats submessage, which indicates the number of
rows inserted/updated/deleted by a DML statement.

It also makes a change to test utility functions to pass back query
statistics.

### DIFF
--- a/bigquery/job.go
+++ b/bigquery/job.go
@@ -436,6 +436,10 @@ type QueryStatistics struct {
 	// statements INSERT, UPDATE or DELETE.
 	NumDMLAffectedRows int64
 
+	// DMLStatistics provides statistics about the row mutations performed by
+	// DML statements.
+	DMLStats *DMLStats
+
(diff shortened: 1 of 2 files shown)


bigquery/job.go
```diff
@@ -436,6 +436,10 @@ type QueryStatistics struct {
 	// statements INSERT, UPDATE or DELETE.
 	NumDMLAffectedRows int64
 
+	// DMLStatistics provides statistics about the row mutations performed by
```
codyoss, line 439: nit: DMLStatistics and DMLStats don't match. Either is fine by me.

codyoss (approved):
Just a small thing, else LGTM

bigquery/job.go
shollyman: Matched the underlying pattern.  Object is now DMLStatistics, fieldname in querystatistics is DMLStats.

PR #4407, opened 2021-07-09
feat(bigquery/storage/managedwriter): add state tracking

shollyman:
This PR introduces two new types:

AppendResult - tracks the progress of an individual row append to
completion, either success or error.  Successful appends _may_ have
an associated offset, failed appends will have an associated error.
The AppendResult has a blocking method users can interrogate.

pendingWrite - handles the state management for a set of rows appended
as a group.  There's a 1:many relationship between
pendingWrite:AppendResult(s), so as a pendingWrite completes all
associated AppendResult references should be updated.


bigquery/storage/managedwriter/appendresult.go
```diff
@@ -0,0 +1,133 @@
...
+	// We compute the size now for flow controller purposes, though
+	// the actual request size may be slightly larger (e.g. the first
+	// request in a new stream bears schema and stream id).
+	pw.reqSize = proto.Size(pw.request)
+	return pw
+}
+
+// noOffsetSpecified is a sentinel value for signalling we're not tracking
+// stream offset (e.g. a default stream which allows simultaneous append streams).
+var noOffsetSpecified int64 = -1
```
codyoss, line 111: I think this could be a const and moved to top of file.

bigquery/storage/managedwriter/appendresult.go
```diff
@@ -0,0 +1,133 @@
...
+	return &AppendResult{
+		ready:   make(chan struct{}),
+		rowData: data,
+	}
+}
+
+// Ready blocks until the append request is completed.
+func (ar *AppendResult) Ready() <-chan struct{} { return ar.ready }
+
+// GetResult returns the optional offset of this row, or the associated
```
codyoss, line 49: When this says optional offset does that mean sometimes the offset will be 0 even if there is not an error?

bigquery/storage/managedwriter/appendresult.go
```diff
@@ -0,0 +1,133 @@
...
+// error.
+func (ar *AppendResult) GetResult(ctx context.Context) (int64, error) {
+	select {
```
codyoss, line 52: I wonder if this is necessary, the calling function would also have to do something like this were it ignores the context the first time, else you would just get a context deadline error at that level.

codyoss: If we want to do this we should add a test to show what this does.

bigquery/storage/managedwriter/appendresult.go
```diff
@@ -0,0 +1,133 @@
...
+	for _, ar := range pw.results {
+		if err != nil {
+			ar.err = err
+			close(ar.ready)
+			continue
+		}
+
+		ar.offset = curOffset
+		// only advance curOffset if we were given a valid starting offset.
+		if startOffset >= 0 {
```
codyoss, line 126: Should there be a test for checking offsets?

codyoss (commented):
A couple of comments

bigquery/storage/managedwriter/appendresult.go
shollyman: Yeah, this should be exported.

bigquery/storage/managedwriter/appendresult.go
shollyman: There's certain kinds of streams (default, committed) where sending offset with an append isn't supported.  Additionally, for (default) streams the backend responses never tell you the current offset because the stream always exists, can can be written to by multiple streams simultaneously.  In this case, you only care about error presence.

My intent was to return 0, but perhaps the exported constant is better.

bigquery/storage/managedwriter/appendresult.go
shollyman: I'll verify this in the test.

bigquery/storage/managedwriter/appendresult.go
shollyman: I'll simplify this one for now.  We may need to reintroduce it, but fine with losing a result to a context cancellation.

bigquery/storage/managedwriter/appendresult.go
```diff
@@ -22,6 +22,10 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
+// NoOffset is a sentinel value for signalling we're not tracking
+// stream offset (e.g. a default stream which allows simultaneous append streams).
+const NoOffset int64 = -1
```
codyoss, line 27: nit: maybe NoStreamOffset. As the package grows I could see this being harder to identify what it is for.

codyoss (approved):
LGTM, have one small comment to consider.

bigquery/storage/managedwriter/appendresult.go
shollyman: Done.

PR #4421, opened 2021-07-12
chore(compute): add integration tests for compute

georgiyekkert:


compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+		t.Error(err)
+	}
+	items := list.GetItems()
+	found := false
+	for _, element := range items {
+		if element.GetName() == name {
+			found = true
+		}
+	}
+	if found == false{
```
noahdietz, line 139:
```suggestion
	if !found {
```

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	deleteInstanceRequest := &computepb.DeleteInstanceRequest{
+		Project: projectId,
+		Zone: defaultZone,
+		Instance: name,
+	}
+	_, err = c.Delete(ctx, deleteInstanceRequest)
+	if err != nil {
+		fmt.Println(err)
```
noahdietz, line 150: Print the error is unnecessary because `t.Error` will do the same thing but in the context of the failure.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+	assert.Equal(t, "тест", get.GetDescription())
+
+	listRequest := &computepb.ListInstancesRequest{
+		Project: projectId,
+		Zone: defaultZone,
+	}
+
+	list, err := c.List(ctx, listRequest)
+	if err != nil {
+		fmt.Println(err)
```
noahdietz, line 129: Please remove this print.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+	defer ForceDeleteInstance(name, ctx, c)
+
+	getRequest := &computepb.GetInstanceRequest{
+		Project: projectId,
+		Zone: defaultZone,
+		Instance: name,
+	}
+	get, err := c.Get(ctx, getRequest)
+	if err != nil {
+		fmt.Println(err)
```
noahdietz, line 116: Please remove this print.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+	createRequest := &computepb.InsertInstanceRequest{
+		Project: projectId,
+		Zone: defaultZone,
+		InstanceResource: instance,
+	}
+
+	insert, err := c.Insert(ctx, createRequest)
+	if err != nil {
+		fmt.Println(err)
```
noahdietz, line 93: Please remove this print.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+// To run these tests, set GCLOUD_TESTS_GOLANG_PROJECT_ID env var to your GCP projectID
+
+package compute
+
+import (
+	"cloud.google.com/go/internal/testutil"
+	"context"
+	"fmt"
+	"github.com/stretchr/testify/assert"
```
noahdietz, line 23: We try not to use third party libs when possible. This does seem to be in our `go.sum` tree, but I'm not sure we should be using it. If you could, can you switch your `assert.Equals` to `if` comparisons and report failures with `t.Errorf`? Take a look at how other packages with hand-written tests do this (https://github.com/googleapis/google-cloud-go/blob/master/storage/acl_test.go#L61-L63)

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+		Rules: rules,
+	}
+
+	insertRequest := &computepb.InsertSecurityPolicyRequest{
+		Project: projectId,
+		SecurityPolicyResource: securityPolicy,
+	}
+	insert, err := c.Insert(ctx, insertRequest)
+	if err != nil {
+		fmt.Println(err)
```
noahdietz, line 224: Please remove this print statement.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+		t.Error(err)
+	}
+
+	waitGlobalRequest := &computepb.WaitGlobalOperationRequest{
+		Project: projectId,
+		Operation: insert.GetName(),
+	}
+	_, err = globalCLient.Wait(ctx, waitGlobalRequest)
+	if err != nil {
+		fmt.Println(err)
```
noahdietz, line 234: Please remove this print statement.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+	removeRuleRequest := &computepb.RemoveRuleSecurityPolicyRequest{
+		Priority: &priority,
+		Project: projectId,
+		SecurityPolicy: name,
+	}
+
+	rule, err := c.RemoveRule(ctx, removeRuleRequest)
+	if err != nil {
+		fmt.Println(err)
```
noahdietz, line 248: Please remove this print statement.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+		t.Error(err)
+	}
+	waitGlobalRequestRemove := &computepb.WaitGlobalOperationRequest{
+		Project: projectId,
+		Operation: rule.GetName(),
+	}
+	_, err = globalCLient.Wait(ctx, waitGlobalRequestRemove)
+	if err != nil {
+		fmt.Println(err)
```
noahdietz, line 257: Please remove this print statement.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+		t.Error(err)
+	}
+
+	getRequest := &computepb.GetSecurityPolicyRequest{
+		Project: projectId,
+		SecurityPolicy: name,
+	}
+	get, err := c.Get(ctx, getRequest)
+	if err != nil {
+		fmt.Println(err)
```
noahdietz, line 267: Please remove this print statement.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+	assert.Equal(t, 1, len(get.GetRules()))
+
+	deleteRequest := &computepb.DeleteSecurityPolicyRequest{
+		Project: projectId,
+		SecurityPolicy: name,
+	}
+	_, err = c.Delete(ctx, deleteRequest)
+	if err != nil {
+		fmt.Println(err)
```
noahdietz, line 278: Please remove this print statement.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+func ForceDeleteInstance(name string, ctx context.Context, client *InstancesClient){
+	deleteInstanceRequest := &computepb.DeleteInstanceRequest{
+		Project: projectId,
+		Zone: defaultZone,
+		Instance: name,
+	}
+	_, err := client.Delete(ctx, deleteInstanceRequest)
+	if err != nil {}
```
noahdietz, line 162: Should we be swallowing the error? If so, just remove the `_, err :=` and this line. If you think it should be reported, please pass in the `*testing.T` instance.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+		Instance: name,
+	}
+	_, err = c.Delete(ctx, deleteInstanceRequest)
+	if err != nil {
+		fmt.Println(err)
+		t.Error(err)
+	}
+}
+
+func ForceDeleteInstance(name string, ctx context.Context, client *InstancesClient){
```
noahdietz, line 155: We prefer `ctx context.Context` to be the first parameter to any function: https://pkg.go.dev/context

compute/apiv1/smoke_test.go
codyoss, line 23: nit: Let's avoid pulling in a new dependency. In general we follow [Useful Test Failures](https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures).

noahdietz: Mmm indeed this is why the tests are failing, this dependency is not in our `go.mod`. Please remove its use.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+	"math/rand"
+	"testing"
+	"time"
+)
+
+var projectId = testutil.ProjID()
+var defaultZone = "us-central1-a"
+
+func TestCreateGetListInstance(t *testing.T){
+	rand.Seed(time.Now().UTC().UnixNano())
```
codyoss, line 35: May want to look at cloud.google.com/go/internal/uid. There are several examples in this repo for generating a names.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+	name := fmt.Sprintf("gotest%d", rand.Int())
+	description := "тест"
```
codyoss, line 37: Honest question, do our apis support these characters?

compute/apiv1/smoke_test.go
codyoss, line 162: Empty conditional. Should this be a t.Helper?

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+}
+
+func ForceDeleteSecurityPolicy(name string, ctx context.Context, client *SecurityPoliciesClient){
+	deleteRequest := &computepb.DeleteSecurityPolicyRequest{
+		Project: projectId,
+		SecurityPolicy: name,
+	}
+	_, err := client.Delete(ctx, deleteRequest)
+	if err != nil {}
```
codyoss, line 289: Empty conditional. Should this be a t.Helper?

compute/apiv1/smoke_test.go
codyoss, line 93: All of the `fmt.Println`s can be removed.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	waitZonalRequest := &computepb.WaitZoneOperationRequest{
+		Project: projectId,
+		Zone: defaultZone,
+		Operation: insert.GetName(),
+	}
+	_, err = zonesClient.Wait(ctx, waitZonalRequest)
+	if err != nil {
+		return
```
codyoss, line 104: Should `t.Error()`.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,290 @@
...
+	"google.golang.org/protobuf/proto"
+	"math/rand"
+	"testing"
+	"time"
+)
+
+var projectId = testutil.ProjID()
+var defaultZone = "us-central1-a"
+
+func TestCreateGetListInstance(t *testing.T){
```
codyoss, line 34: If these are going to be run as integration tests they should be guarded by something like:

```go
if testing.Short() {
		t.Skip("Integration tests skipped in short mode")
}
```

codyoss (changes requested):
Thanks for adding these, some initial feedback.

codyoss:
Whoops had an old browser tab open that had not updated. Sorry for any dupes that Noah may have mentioned.

compute/apiv1/smoke_test.go
georgiyekkert: We should swallow the error here as it works like tear down.

compute/apiv1/smoke_test.go
georgiyekkert: Didn't know that, thanks!

compute/apiv1/smoke_test.go
georgiyekkert: Yep, they do :)

compute/apiv1/smoke_test.go
georgiyekkert: I removed empty conditional, we want to swallow error here

compute/apiv1/smoke_test.go
georgiyekkert: I've tried uuid but it generated names that API didn't accept. uid works as expected, thanks

compute/apiv1/smoke_test.go
georgiyekkert: Done

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,280 @@
...
+	c, err := NewInstancesRESTClient(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	zonesClient, err := NewZoneOperationsRESTClient(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	configName := "default"
+	accessConfig := computepb.AccessConfig{
```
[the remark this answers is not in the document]
georgiyekkert: done

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,280 @@
...
+		AutoDelete:       proto.Bool(true),
+		Boot:             proto.Bool(true),
+		Type:             &diskType,
+		InitializeParams: initializeParams,
+	}
+	disks := []*computepb.AttachedDisk{
+		&disk,
+	}
+	instance := &computepb.Instance{
+		Name:              &name,
```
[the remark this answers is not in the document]
georgiyekkert: I left "name" variable as it used in multiple places but I removed vars where it was used only once.

noahdietz (approved):
LGTM but @codyoss needs to Approve. Thanks Georgii.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,257 @@
...
+	ctx := context.Background()
+	c, err := NewInstancesRESTClient(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	zonesClient, err := NewZoneOperationsRESTClient(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	diskType := computepb.AttachedDisk_PERSISTENT
```
codyoss, line 47: nit: this can be inlined by adding `.Enum()`

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,257 @@
...
+						Name: proto.String("default"),
+					},
+				},
+			},
+		},
+	}
+	createRequest := &computepb.InsertInstanceRequest{
+		Project:          projectId,
+		Zone:             defaultZone,
+		InstanceResource: instance,
```
codyoss, line 75: nit: inline variable.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	insert, err := c.Insert(ctx, createRequest)
+	if err != nil {
+		t.Error(err)
```
codyoss, line 80: nit: Please add a little extra information to error message about what the failure is. Also, should this be a t.Fatal? (The rest of the test would stop running.)

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,257 @@
...
+	waitZonalRequest := &computepb.WaitZoneOperationRequest{
+		Project:   projectId,
+		Zone:      defaultZone,
+		Operation: insert.GetName(),
+	}
+	_, err = zonesClient.Wait(ctx, waitZonalRequest)
+	if err != nil {
+		t.Error(err)
+	}
+	fmt.Printf("Inserted instance named %s\n", name)
```
codyoss, line 92: nit: remove. This could be switched to a t.Log but I am not don't think it needs to be.

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,257 @@
...
+	getRequest := &computepb.GetInstanceRequest{
+		Project:  projectId,
+		Zone:     defaultZone,
+		Instance: name,
+	}
+	get, err := c.Get(ctx, getRequest)
+	if err != nil {
+		t.Error(err)
+	}
+	testutil.Equal(name, get.GetName())
```
codyoss, line 104: These just do a compare under the hood, they don't "assert". Still need a t.Error in a conditional. In this case the values are just strings so a normal comparison should be fine. For suggestions on testing errors please see: [Useful Test Failures](https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures).

compute/apiv1/smoke_test.go
```diff
@@ -0,0 +1,257 @@
...
+	for _, element := range items {
+		if element.GetName() == name {
+			found = true
+		}
+	}
+	if !found {
+		t.Error("Couldn't find the instance in list response")
+	}
+
+	deleteInstanceRequest := &computepb.DeleteInstanceRequest{
```
codyoss, line 126: Isn't this done by the defer already?

codyoss (commented):
Only commented on the first test but similar patterns are seen in other tests.

compute/apiv1/smoke_test.go
georgiyekkert: logic is:
1) we want to test delete() method here and the test fails if there are errors from API.
2) defer delete is for cases where we inserted an instance but test execution was aborted before it ended.

compute/apiv1/smoke_test.go
georgiyekkert: done

codyoss (approved):

codyoss:
@georgiyekkert Thanks for working on these!

codyoss:
@georgiyekkert Please `gofmt` the file to make CI happy.

georgiyekkert:
@codyoss I gofmt'ed files but CI still fails. I was running `gofmt -w .`  tcommand

codyoss:
@georgiyekkert It not looks like goimports is failing:

```
+ goimports -l .
+ tee /dev/stderr
+ read
compute/apiv1/smoke_test.go
```

You can run
```
go install golang.org/x/tools/cmd/goimports@latest
goimports -w .
```

georgiyekkert:
@codyoss could you rerun CI jobs please? looks like CLA job got stuck

PR #4422, opened 2021-07-12
feat(bigquery/storage/managedwriter): add base client

shollyman:
This PR adds a base client and managed stream abstraction, and
implements some of the associated surfaces.  All the
streaming client abstractions are elided and will be introduced in
subsequent PRs, but this PR does include non-streaming RPC methods

Alongside the client, we introduce an option type (WriterOption) for
constructing a client in a variadic fashion.

The client contains an internal settings type, streamSettings, which
contains fields of note for both the streaming client abstraction and
its flow controller.

Testing: This PR contains unit tests, but doesn't include integration
tests.  I'll start hoisting that in soon.


bigquery/storage/managedwriter/client.go
```diff
@@ -0,0 +1,222 @@
...
+	return &streamSettings{
+		streamType:          DefaultStream,
+		MaxInflightRequests: 1000,
+		MaxInflightBytes:    0,
+		TracePrefix:         "defaultManagedWriter",
+	}
+}
+
+// NewManagedWriter instantiates a new managed writer.
+func NewManagedWriteClient(ctx context.Context, client *storage.BigQueryWriteClient, table *bigquery.Table, opts ...WriterOption) (*ManagedWriteClient, error) {
```
codyoss, line 92: I am not sure this should take another client as input. That means there is some pre-setup needed before calling this method. I think a function like this would be easier to use and follow the same style as other clients:

```go
func NewClient(ctx context.Context, settings SomeConfigStruct, opts ...option.ClientOption)
```

- I changed the name
- Because of dialing in this fashion we need to take clientoptions so that means writer options could not be passed as well. Used config struct.
- I left out Table. I am not sure what to do here, feedback welcome. I know for some clients we just take a string called parent and is documented and expected to be in a certain form, maybe we could do that here. It could be rolled into the settings as well I suppose. Using table make this package require the base package which I am not sure is desirable. It also has an embedded and is used in different type of semantics in the other client.

bigquery/storage/managedwriter/client.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	return resp.GetOffset(), nil
+}
+
+// Finalize is used to mark a stream as complete, and thus ensure no further data can
+// be appended to the stream.  You cannot finalize a DefaultStream, as it always exists.
+//
+// Finalizing does not advance the current offset of a BufferedStream, nor does it commit
+// data in a PendingStream.
+func (mw *ManagedWriteClient) Finalize(ctx context.Context) (int64, error) {
```
codyoss, line 173: Once this is called can the client still be used? Should this be a Close for consistency if so?

codyoss: Should client be able to be reused somehow but on a different stream?

shollyman: This shouldn't close a client, because Finalize isn't the end of a stream's interactions.  You still can choose to commit (in the case of pending streams).

There's also some future considerations around whether we do traffic balancing by creating multiple streams under the hood, but that's not currently in scope for this first round.

bigquery/storage/managedwriter/client.go
shollyman: Hmm, good feedback.  Perhaps we should have something like a NewManagedStream( opts ...WriterOption) hung off the client, which gives us multiple streams on the same client, and lets us do the client options and the writer options.  Then we move the commit off the stream and into the base client.  Let me play with this some more.

codyoss: SGTM

bigquery/storage/managedwriter/client.go
```diff
@@ -0,0 +1,222 @@
...
+// StreamType indicates the type of stream this write client is managing.
+type StreamType string
+
+var (
+	// DefaultStream most closely mimics the legacy bigquery
+	// tabledata.insertAll semantics.  Successful inserts are
+	// committed immediately, and there's no tracking offsets as
+	// all writes go into a "default" stream that always exists
+	// for a table.
+	DefaultStream StreamType = "DEFAULT"
```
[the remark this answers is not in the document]
shollyman: I went with the strings to support `getStreamType()` as a human readable thing out of the box.  Since we're not exposing the protos directly they can't easily invoke the proto's method to get at it.

bigquery/storage/managedwriter/client.go
shollyman: As discussed, latest change breaks it into a Client and a ManagedStream.  ManagedStream retains the settings, Client contains methods that aren't tied to a specific stream (mostly just batch commit streams, and getStream for retreiving stats.

bigquery/storage/managedwriter/client.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	return &Client{
+		rawClient: rawClient,
+		projectID: projectID,
+	}, nil
+}
+
+// NewManagedStream establishes a new stream for writing.
+func (c *Client) NewManagedStream(ctx context.Context, table *bigquery.Table, opts ...WriterOption) (*ManagedStream, error) {
```
codyoss, line 58: I think table could just be datasetID and tableID. I worry about this client reaching for code in both directions up and down the file structure. This means the core bigquery package could never rely on this package due to cycles.

bigquery/storage/managedwriter/client.go
```diff
@@ -0,0 +1,135 @@
...
+		c:              c,
+	}
+
+	// apply writer options
+	for _, opt := range opts {
+		opt(ms)
+	}
+
+	if ms.streamSettings.streamID == "" && ms.streamSettings.streamType == "" {
+		return nil, fmt.Errorf("TODO insufficient validation")
```
codyoss, line 71: Is there more to do right now?

bigquery/storage/managedwriter/client.go
shollyman: Fine switching it to string, now I'm wondering if it should be a variadic argument.  We don't do anything with the reference other than for the construct a stream case, but if we allow users to pass in an optional stream ID perhaps we let them treat table as optional as well?  Will switch it and augment the validation (your other comment) and see how it plays.

bigquery/storage/managedwriter/client.go
shollyman: added a validation method.

bigquery/storage/managedwriter/client.go
```diff
@@ -0,0 +1,163 @@
...
+
+	return &Client{
+		rawClient: rawClient,
+		projectID: projectID,
+	}, nil
+}
+
+// NewManagedStream establishes a new stream for writing.
+//
+// destinationTable should be of the format: projects/{project}/datasets/{dataset}/tables/{table}
```
codyoss, line 58: Is destinationTable required? Should it be a param vs an optional option?

shollyman: Stale comment from refactor, corrected.

codyoss (approved):

Issue #4380, opened 2021-07-07
compute: fix bugs in alpha release


noahdietz:
The `compute/apiv1` alpha release in v0.86.0 has a few known bugs. They are as follows:

- [x] missing scheme on default endpoint
- [x] `GET`/`DELETE` requests are generated with a body
- [x] remove `EmitUnpopulated` encoding option

First discovered post-release and then reported in #4377 by users.

This is a tracking bug.

noahdietz:
https://github.com/googleapis/gapic-generator-go/pull/698 fixes the missing scheme.

noahdietz:
https://github.com/googleapis/gapic-generator-go/pull/699 fixes the extraneous request body.

noahdietz:
Regeneration of `compute/apiv1` with the above fixes passes CI: https://github.com/googleapis/google-cloud-go/pull/4364.

codyoss:
@wanglong001 Thanks for the report, I am going to close this issue in favor of the one linked above.

noahdietz:
Local integration tests succeed.

noahdietz:
Fixes carried in https://github.com/googleapis/gapic-generator-go/releases/tag/v0.21.6.

noahdietz:
Changes carried in #4383

noahdietz:
To fix the third issue, need https://github.com/googleapis/gapic-generator-go/pull/702

noahdietz:
Released generator fix in https://github.com/googleapis/gapic-generator-go/releases/tag/v0.21.7

noahdietz:
Pending the outcome of some more testing before making a new release. Should have an update today midday PST.

noahdietz:
Integration tests uncovered another issue, which was fixed with https://github.com/googleapis/gapic-generator-go/pull/710. Updating generator version in #4425.

noahdietz:
All mentioned bugs were fixed in the release: https://github.com/googleapis/google-cloud-go/releases/tag/v0.87.0

PR #4426, opened 2021-07-13
fix(compute): properly generate PUT requests

noahdietz:
Changes regenerated locally with generator v0.21.8.

This is the fix for https://github.com/googleapis/google-cloud-go/issues/4380#issuecomment-879218739.


codyoss (approved):
LGTM

PR #2599, opened 2020-07-14
feat(spanner): support of client-level custom retry settings

hengfengli:
Changes included are: 

- Add a field `CallOptions` in `ClientConfig`.
- Add a field `callOptions` in `sessionClient`.
- Add a method `mergeCallOptions` to merge two CallOptions.
- Update `nextClient` in sessionclient.go so that if a custom CallOptions is provided, merge it with the default one. 

Fixes #2594

PR #4428, opened 2021-07-13
feat(pubsub): expose CallOptions for pub/sub retries and timeouts

### LINKED ISSUE #4162: pubsub: Allow to override PublisherClient default CallOptions in pubsub.NewClient
At the moment I'm trying to troubleshoot a issue with high latencies when publishing events and I would like to have more visibility on the retries the internal PublisherClient is doing, like logging the error codes, but because pubsub.Client.pubc is not exposed in any way I can't change its CallOptions nor pass those in methods like Topic.Publish.

I guess that the easier way would be to pass my own retry policy when calling pubsub.NewClient as an option

hongalex:
In a similar vein to the spanner PR (#2599), this PR allows users to specify their own retry / timeout configurations for clients for the high level manual library. Default topic/subscription configs are configured in all the GAPIC libraries and are exposed there (`cloud.google.com/go/pubsub/apiv1`), but currently cannot be configured in the handwritten/manual library. This differs from the Spanner PR in that Pub/Sub requires specifying the publisher and subscriber options separately, requiring separate helper functions. Otherwise these functions are identical to spanner.

Closes #4162


pubsub/pubsub.go
```diff
@@ -51,8 +53,75 @@ type Client struct {
...
-// NewClient creates a new PubSub client.
+// ClientConfig has configurations for the client.
+type ClientConfig struct {
+	PublisherCallOptions  *vkit.PublisherCallOptions
+	SubscriberCallOptions *vkit.SubscriberCallOptions
+}
+
+// mergePublisherCallOptions merges two PublisherCallOptions into one and the first argument has
+// a lower order of precedence than the second one. If either is nil, return the other.
+func mergePublisherCallOptions(a *vkit.PublisherCallOptions, b *vkit.PublisherCallOptions) *vkit.PublisherCallOptions {
```
broady, line 64: I think you could write a generic "merge two structs of the same type" function, but this works fine. 👍

broady (approved):

PR #4431, opened 2021-07-14
feat(spanner): add row.String() and refine error message for decoding a struct array

hengfengli:
Two changes: 
- Add row.String()
- Refine the error message for decoding a struct array


olavloite (approved):
LGTM, but I think the commit message should start with `feat` instead of `fix`. This is not really a bug that is being fixed, but a feature that was missing that is being added.

hengfengli:
@skuruppu Can I get an approval from you?

skuruppu (approved):

PR #4445, opened 2021-07-15
fix(internal/godocfx): set exit code, print cmd output, no go get ...

tbpg:
The command was failing, but the job was succeeding. These changes:
* Fix the underlying issue (`...`).
* Print `cmd` output to make debugging easier.
* Set an exit code.


codyoss (approved):
LGTM, thanks!

PR #4450, opened 2021-07-16
test(spanner): fix the failed TestColumnTypeErr test

hengfengli:
Fix #4443

PR #4451, opened 2021-07-16
test(spanner): fix errors not reporting when emulator test passes

codyoss:
Emulator tests were overwritting the results of the main test
run which can cause errors that occured in main test run not
to be reported when the emulator tests pass. Now emulator tests
append to the test log.

Updates: #4450


olavloite (approved):
Thanks for looking into this so quickly. The fact that the build is failing, is a good sign in this case, as the failure is caused by a known failure that was previously missed because the build turned green even though the test failed.

codyoss:
I will force merge this and let that test be fixed in another PR.

PR #4452, opened 2021-07-16
feat(bigquery/storage/managedwriter): add append stream plumbing

shollyman:
This PR adds enough of the wiring to the client to being testing via integration tests.  It adapts a similar pattern to the pullstream in pubsub, in that it abstracts individual calls from stream state management.

There's two significant units of future work that may yield changes here:

* For traffic efficiency sake, we only want to add things like the stream ID, schema, and trace ID to the first append on any stream.

* For stream connection retry, we may want to re-send writes that were sent but we didn't get an acknowledgement back.  For default/committed streams, this behavior may yield additional writes (at least once semantics).  For buffered/pending streams, it means either the library or user should know to expect "data already present" for these resent-writes.


bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -140,3 +159,173 @@ func (ms *ManagedStream) Finalize(ctx context.Context) (int64, error) {
...
+	if ms.err != nil {
+		return nil, nil, ms.err
+	}
+	ms.err = ms.ctx.Err()
+	if ms.err != nil {
+		return nil, nil, ms.err
+	}
+
+	// Always return the retained ARC if the arg differs.
+	if arc != ms.arc {
```
codyoss, line 176: This will differ when ms.arc is nil, does this need an extra check?

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -62,10 +68,23 @@ func streamTypeToEnum(t StreamType) storagepb.WriteStream_Type {
...
 	streamSettings   *streamSettings
+	schemaDescriptor *descriptorpb.DescriptorProto
 	destinationTable string
 	c                *Client
+
+	// aspects of the stream client
+	ctx     context.Context // retained context for the stream
+	cancel  context.CancelFunc
+	open    func() (storagepb.BigQueryWrite_AppendRowsClient, error) // how we get a new connection
+	arc     *storagepb.BigQueryWrite_AppendRowsClient                // current stream connection
```
codyoss, line 79: This is an interface so I don't think it needs to be a pointer, right?

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -140,3 +159,173 @@ func (ms *ManagedStream) Finalize(ctx context.Context) (int64, error) {
...
+	if ms.err != nil {
+		return nil, nil, ms.err
+	}
+
+	// Always return the retained ARC if the arg differs.
+	if arc != ms.arc {
+		return ms.arc, ms.pending, nil
+	}
+
+	ms.arc = new(storagepb.BigQueryWrite_AppendRowsClient)
```
codyoss, line 180: I don't quite understand how this is useful?

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -140,3 +159,173 @@ func (ms *ManagedStream) Finalize(ctx context.Context) (int64, error) {
...
+			ms.err = err
+			ms.mu.Unlock()
+		}
+		return err
+	}
+}
+
+func (ms *ManagedStream) append(pw *pendingWrite) error {
+	return ms.call(func(arc storagepb.BigQueryWrite_AppendRowsClient, ch chan *pendingWrite) error {
+		// TODO: we should only send stream ID and schema for the first message in a new stream, but
```
codyoss, line 249: https://pkg.go.dev/sync#Once.Do ?

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -140,3 +159,173 @@ func (ms *ManagedStream) Finalize(ctx context.Context) (int64, error) {
...
+		}
+		err := arc.Send(pw.request)
+		if err == nil {
+			ch <- pw
+		}
+		return err
+	})
+}
+
+func (ms *ManagedStream) CloseSend() error {
```
codyoss, line 264: Maybe just Close so this is a [Closer](https://pkg.go.dev/io#Closer)

codyoss: Also docs.

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -140,3 +159,173 @@ func (ms *ManagedStream) Finalize(ctx context.Context) (int64, error) {
...
+	})
+	ms.mu.Lock()
+	ms.err = io.EOF
+	close(ms.pending)
+	ms.mu.Unlock()
+	return err
+}
+
+// AppendRows sends the append requests to the service, and returns one AppendResult per row.
+func (ms *ManagedStream) AppendRows(data [][]byte, offset int64) ([]*AppendResult, error) {
```
codyoss, line 280: We should eventually add godoc examples for some of these.

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -140,3 +159,173 @@ func (ms *ManagedStream) Finalize(ctx context.Context) (int64, error) {
...
+		case nextWrite, ok := <-ch:
+			if !ok {
+				// Channel closed, all elements processed.
+				return
+			}
+
+			// block until we get a corresponding response or err from stream.
+			resp, err := arc.Recv()
+			if err != nil {
+				log.Printf("recv got err: %#v", err)
```
codyoss, line 314: nit: remove logging

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -140,3 +159,173 @@ func (ms *ManagedStream) Finalize(ctx context.Context) (int64, error) {
...
+				nextWrite.markDone(NoStreamOffset, err)
+			}
+
+			if status := resp.GetError(); status != nil {
```
codyoss, line 318: Is this needed? Is the error not already a grpc.Status in this case?

codyoss (commented):
Skipped reviewing the test for now, will look at those with the next pass incase refactors are needed. Just a handful of comments.

bigquery/storage/managedwriter/integration_test.go
```diff
@@ -0,0 +1,206 @@
...
+	defer mwClient.Close()
+	defer bqClient.Close()
+
+	dataset, cleanup, err := setupTestDataset(context.Background(), t, bqClient)
+	if err != nil {
+		t.Fatalf("failed to init test dataset: %v", err)
+	}
+	defer cleanup()
+
+	ctx, _ := context.WithTimeout(context.Background(), defaultTestTimeout)
```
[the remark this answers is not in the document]
shollyman: Good point, done.

bigquery/storage/managedwriter/client.go
```diff
@@ -53,41 +55,65 @@ func NewClient(ctx context.Context, projectID string, opts ...option.ClientOptio
 	}, nil
 }
 
+// Close releases resources held by the client.
+func (c *Client) Close() error {
+	// TODO: we should retain a references to instantiated clients, or have a client-local context.
+	return fmt.Errorf("not implemented")
```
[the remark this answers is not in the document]
shollyman: I'm also worried about running goroutines that may hang waiting on updates.  Will do as you suggested and leave a note for propagating a cancellation.

bigquery/storage/managedwriter/client.go
```diff
@@ -53,41 +55,65 @@ func NewClient(ctx context.Context, projectID string, opts ...option.ClientOptio
...
+}
+
+func (c *Client) buildManagedStream(ctx context.Context, streamFunc streamClientFunc, skipSetup bool, opts ...WriterOption) (*ManagedStream, error) {
+
+	ctx, cancel := context.WithCancel(ctx)
 
 	ms := &ManagedStream{
 		streamSettings: defaultStreamSettings(),
 		c:              c,
+		ctx:            ctx,
```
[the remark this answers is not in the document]
shollyman: I may be a bit wonky here due to the BQ client here having a bunch of retained context.

bigquery/storage/managedwriter/managed_stream.go
shollyman: Oooh, good point.  Will think about that.

bigquery/storage/managedwriter/managed_stream.go
shollyman: the response is a oneof that can either signal append error of append success.

bigquery/storage/managedwriter/managed_stream.go
shollyman: printf debugging is my jam, removed.

bigquery/storage/managedwriter/managed_stream.go
shollyman: Yeah, there will need to be extensive docs on this thing, even with the simplification.

bigquery/storage/managedwriter/managed_stream.go
shollyman: Swapped name.

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -62,10 +68,23 @@ func streamTypeToEnum(t StreamType) storagepb.WriteStream_Type {
...
+	schemaDescriptor *descriptorpb.DescriptorProto
 	destinationTable string
 	c                *Client
+
+	// aspects of the stream client
+	ctx     context.Context // retained context for the stream
+	cancel  context.CancelFunc
+	open    func() (storagepb.BigQueryWrite_AppendRowsClient, error) // how we get a new connection
+	arc     *storagepb.BigQueryWrite_AppendRowsClient                // current stream connection
+	mu      sync.Mutex
```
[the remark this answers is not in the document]
shollyman: Done.

bigquery/storage/managedwriter/managed_stream.go
codyoss: Oh streaming RPC, duh. My bad.

bigquery/storage/managedwriter/managed_stream.go
shollyman: Plumbed this in.  I ended up removing the intermediate call() abstraction, as we're only using it for sending appends and for issuing the CloseSend() and making those two methods call getStream directly.  This ensures the sync.Once fires only for appends.

bigquery/storage/managedwriter/client.go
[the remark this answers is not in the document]
shollyman: Decided to stick with retained context for now and document.  We could potentially get clever using the contexts passed in to AppendRows, but then the lifetime of the contexts become even more ambiguous (e.g. a single append context gets used for the receive processor, etc).

bigquery/storage/managedwriter/managed_stream.go
shollyman: The only time ms.arc will be nil is when arc is nil (we have never opened a connection), so we'll skip this conditional.

codyoss (approved):
LGTM

PR #4457, opened 2021-07-19
feat(spanner/spansql): support table_hint_expr at from_clause on query_statement

nakatamixi:
This fixes parse error when query statement includes table hint expr.
This add function `parseHints` and use in parsing table hints and join hints.


hengfengli:
@olavloite Can you please review this PR? Thanks.

spanner/spansql/parser.go
```diff
@@ -2111,6 +2111,13 @@ func (p *parser) parseSelectFrom() (SelectFrom, *parseError) {
 		return nil, err
 	}
 	sf := SelectFromTable{Table: tname}
+	if p.eat("@") {
+		hint, err := p.parseHints(map[string]string{}, false)
```
olavloite, line 2115: Hmm... It seems that the [documentation](https://cloud.google.com/spanner/docs/query-syntax#from_clause) is slightly off here. It is actually possible to have more than one table hint, so the variable here should be renamed to `hints` and the `parseHints(..)` function should be called with `true` rather than `false`.

For example the following query is accepted by Cloud Spanner:

```sql
select *
from albums@{FORCE_INDEX=_BASE_TABLE, GROUPBY_SCAN_OPTIMIZATION=FALSE}
```

spanner/spansql/parser.go
```diff
@@ -2111,6 +2111,13 @@ func (p *parser) parseSelectFrom() (SelectFrom, *parseError) {
...
+		if err != nil {
+			return nil, err
+		}
+		sf.Hint = hint
```
olavloite, line 2119: nit: same as above, rename to `sf.Hints`

spanner/spansql/parser_test.go
```diff
@@ -114,6 +114,20 @@ func TestParseQuery(t *testing.T) {
 				},
 			},
 		},
+		// with table hint
+		{`SELECT * FROM Packages@{FORCE_INDEX=PackagesIdx} WHERE package_idx=@packageIdx`,
```
olavloite, line 118: nit: add an additional test that verifies that the parser also accepts multiple table hints.

spanner/spansql/sql.go
```diff
@@ -365,6 +365,14 @@ func (sel Select) addSQL(sb *strings.Builder) {
 
 func (sft SelectFromTable) SQL() string {
 	str := sft.Table.SQL()
+	if len(sft.Hint) > 0 {
+		for k, v := range sft.Hint {
+			str += fmt.Sprintf("@{%s=%s}", k, v)
+			// table hint exists only one key
```
olavloite, line 371: Same as above, it seems that the documentation is not completely clear on this, but Cloud Spanner does accept more than one table hint.

spanner/spansql/types.go
```diff
@@ -394,6 +394,7 @@ type SelectFrom interface {
 type SelectFromTable struct {
 	Table ID
 	Alias ID // empty if not aliased
+	Hint  map[string]string
```
olavloite, line 397: nit: rename to `Hints`

olavloite (commented):
Thanks for your contribution. This looks mostly good to me, but it seems that the documentation is not 100% correct on whether multiple table hints are allowed or not. Cloud Spanner actually does support more than one table hint, so we should try to implement it in the same way in this emulator. See my comments and example below.

nakatamixi:
Thanks for your kind review.
I checked multiple table hint keys can use at my spanner instance.
And I fixed to `parseHints` always parse multiple keys.

olavloite (approved):
Thanks, LGTM

olavloite:
@hengfengli I think we need an approval from either you or @skuruppu in order to merge this.

hengfengli (approved):

spanner/spansql/sql.go
```diff
@@ -365,6 +365,18 @@ func (sel Select) addSQL(sb *strings.Builder) {
 
 func (sft SelectFromTable) SQL() string {
 	str := sft.Table.SQL()
+	if len(sft.Hints) > 0 {
+		str += "@{"
+		kvs := make([]string, len(sft.Hints))
+		i := 0
+		for k, v := range sft.Hints {
+			kvs[i] = fmt.Sprintf("%s=%s", k, v)
+			i += 1
```
hengfengli, line 374: Can you please update to `i++`? I got an error from the linter: 

```
spanner/spansql/sql.go:374:4: should replace i += 1 with i++
```

nakatamixi: Thanks.

fixed at 83d63c5

hengfengli:
The integration test fails due to #4450. I will merge this PR after it is merged.

hengfengli:
@nakatamixi Thank you for your contribution!

nakatamixi:
Thank you!

PR #4459, opened 2021-07-19
feat(pubsub/pstest): add ability to create a pstest server listening on a specified port

caod123:
In some of our tests we handle port allocations manually and would prefer to pass in an allocated port for our pstest Server dependency rather than having the `pstest.NewServer` call allocate an available port.

### DIFF
--- a/pubsub/pstest/fake.go
+++ b/pubsub/pstest/fake.go
@@ -114,9 +114,14 @@ type GServer struct {
 
 // NewServer creates a new fake server running in the current process.
 func NewServer(opts ...ServerReactorOption) *Server {
-	srv, err := testutil.NewServer()
+	return NewServerWithPort(0, opts...)
+}
+
+// NewServerWithPort creates a new fake server running in the current process at the specified port.
+func NewServerWithPort(port int, opts ...ServerReactorOption) *Server {
+	srv, err := testutil.NewServerWithPort(port)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


caod123:
@googlebot I signed it!

pubsub/pstest/fake.go
```diff
@@ -114,7 +114,12 @@ type GServer struct {
...
 func NewServer(opts ...ServerReactorOption) *Server {
-	srv, err := testutil.NewServer()
+	return NewServerWithPort(0, opts...)
+}
+
+// NewServerWithPort creates a new fake server running in the current process at the specified port.
+func NewServerWithPort(port int, opts ...ServerReactorOption) *Server {
+	srv, err := testutil.NewServerWithPort(port)
 	if err != nil {
 		panic(fmt.Sprintf("pstest.NewServer: %v", err))
```
hongalex, line 119: nit: change this to NewServerWithPort. Perhaps add the port number if that's not logged in the error, I don't know if it is or not.

caod123: `testutil.NewServerWithPort` just calls a basic `net.Listen` which should return an error containing the port

hongalex (approved):
LGTM pending tests

dc185333:
Rebased on master, may need to force-run kokoro again

hongalex:
Thanks for the fix @dc185333, merged.

Issue #4447, opened 2021-07-15
all: repo metadata incorrect for submodules


tbpg:
For example:

https://github.com/googleapis/google-cloud-go/blob/60160af8f815122ef1c26297a78b4707912374c2/internal/.repo-metadata-full.json#L565

PR #4460, opened 2021-07-19
fix(internal): detect module for properly generating docs URLs

codyoss:
Fixes: #4447


tbpg (approved):
Thank you!

I'm not sure how we should handle unreleased packages. We also need to handle packages that move from one module to another. But, those are separate issues.

PR #4468, opened 2021-07-20
chore: remove sample that does not compile

codyoss:


codyoss:
Pushing this to unblock regen.

PR #4471, opened 2021-07-21
feat(spanner/spansql): fix (date|timestamp) comparison parse.

nakatamixi:
This PR fixes two nit things.

- fix (date|timestamp) comparison parse.
- fix unstable SelectFromTable SQL

Second was made unstable tests and SQL output by #4457 (sorry.)


nakatamixi:
The first one is imcomplete, so I'll modify this PR only include failed test.

> fix (date|timestamp) comparison parse.

So, I close this and made new PR https://github.com/googleapis/google-cloud-go/pull/4473 only include fix test.

PR #4476, opened 2021-07-21
chore: manually regen cloudbuild do to breaking changes

codyoss:
I had to manually manipulate the old code directory so that it
could compile. In the future this makes me question if v2 directories
make sense for generated code were we must always match the proto
spec for code to actually work.

I also include a few changes to the generator the helped me regen
and debug what was wrong.
Changes:

feat(cloudbuild/apiv1): Implementation of Build Failure Info: - Added message FailureInfo field
  PiperOrigin-RevId: 385847244
  Source-Link: googleapis/googleapis@f84d1e2

feat(cloudbuild/apiv1): add a WorkerPools API
  BREAKING CHANGE: Proto had a prior definitions of WorkerPool resources which were never supported. This change replaces those resources with definitions that are currently supported.

  Committer: @mwiczer
  PiperOrigin-RevId: 385610622
  Source-Link: googleapis/googleapis@7a1fb83


codyoss:
@broady I am starting to doubt that v2 directories make sense for generated proto sources for reasons like this.

internal/gapicgen/generator/gapics.go
```diff
@@ -100,11 +100,11 @@ func (g *GapicGenerator) Regen(ctx context.Context) error {
...
-	if !g.onlyGenerateGapic {
-		if err := g.regenSnippets(ctx); err != nil {
-			return err
-		}
-	}
+	// if !g.onlyGenerateGapic {
+	// 	if err := g.regenSnippets(ctx); err != nil {
+	// 		return err
+	// 	}
+	// }
```
noahdietz, line 107: Please uncomment this :)

codyoss: Whoops, thanks!

noahdietz:
It's a bummer because we made the `v2` directory after adding operation_info annotations to the RPCs I think which was a breaking change. Now the original client is the same as the v2 one, right?

codyoss:
@noahdietz It is not the same as it is missing a handful of RPCs as of right now, but it could be. But yeah I think the v2 pattern for generated clients might not make sense after seeing all of the trouble it causes.

codyoss:
The kokoro failures are due to the breaking changes in this API which are expected as they fix the code to return the proper types that the API exposes. Force pushing.

PR #4479, opened 2021-07-22
fix(storage): revise Reader to send XML preconditions

### LINKED ISSUE #4470: storage: Reader uses the xml api but sends json preconditions
* `NewRangerReader` uses the XML api to download files.
* `conditionsQuery()` uses the JSON API to configure preconditions
* `NewRangerReader` calls `conditionsQuery()` rather than setting the expected precondition headers.

---

`ObjectHandle.NewReader()` calls `NewRangeReader`: https://github.com/googleapis/google-cloud-go/blob/5ac09d911f9e65fd15acb7c0c7acb0dd635f364f/storage/reader.go#L79-L81

This eventually sets `req.URL.RawQuery = conditionsQuery(gen, o.conds)`:

https://github.com/googleapis/google-cloud-go/blob/5ac09d911f9e65fd15acb7c0c7acb0dd635f364f/storage/reader.go#L151-L152

`conditionsQuery()` appends the applicable [JSON API URL parameters](https://cloud.google.com/storage/docs/generations-preconditions#_JSONAPI):
https://github.com/googleapis/google-cloud-go/blob/5ac09d911f9e65fd15acb7c0c7acb0dd635f364f/storage/storage.go#L1620-L1633

However, the `readHost` is not the json api, it is the xml api:
https://github.com/googleapis/google-cloud-go/blob/5ac09d911f9e65fd15acb7c0c7acb0dd635f364f/storage/storage.go#L118

The [xml API expects](https://cloud.google.com/storage/docs/generations-preconditions#_XMLAPI) an `x-goog-if-generation-match` header. It does not recognize the JSON API's url parameters.

Thus the following will always work and never return the expected preconditon error:

```golang
client := storage.NewClient()
r, err := client.Bucket("whatever").Object("some/object").If(&storage.Conditions{GenerationMatch: 666}).NewReader(context.Background())
if err == nil {
  panic("should have failed")
}
gerr, ok := err.(*googleapi.Error)
if !ok {
  panic("should have returned a googleapi error")
}
if gerr.Code != http.PreconditionFailed {
  panic("should have return a precondition failed error")
}
// congrats! This bug no longer repros.
```

cojenco:
This updates `Reader` to send precondition request headers to fit the XML API used for downloads. 

- adds a new function `setConditionsHeaders()` that parses `Conditions` and sets the corresponding precondition headers
- revises `NewRangerReader` to call `setConditionsHeaders()`
- if an object version is specified via `generation`, include `generation` as query string parameters
- updates tests

Fixes #4470


storage/reader.go
```diff
@@ -149,7 +149,13 @@ func (o *ObjectHandle) NewRangeReader(ctx context.Context, offset, length int64)
 			req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", start, offset+length-1))
 		}
 		// We wait to assign conditions here because the generation number can change in between reopen() runs.
-		req.URL.RawQuery = conditionsQuery(gen, o.conds)
```
tritone, line 152: I think conditionsQuery is no longer used anywhere in the library? If so we can remove it since it's not exported.

storage/storage.go
```diff
@@ -1634,6 +1634,24 @@ func conditionsQuery(gen int64, conds *Conditions) string {
 	return string(buf)
 }
 
+// setConditionsHeaders sets precondition request headers for downloads
+// using the XML API. It assumes that the conditions have been validated.
```
tritone, line 1638: Let's maybe move this to reader.go since it's only used there?

storage/integration_test.go
```diff
@@ -744,6 +744,36 @@ func TestIntegration_ObjectsRangeReader(t *testing.T) {
...
+
+	wc := o.NewWriter(ctx)
+	wc.ContentType = "text/plain"
+	h.mustWrite(wc, []byte("foo"))
+
+	gen := wc.Attrs().Generation
+	metaGen := wc.Attrs().Metageneration
+
+	if _, err := o.Generation(gen + 1).NewReader(ctx); err == nil {
+		t.Fatalf("Unexpected successful download with nonexistent Generation")
```
tritone, line 764: nit: these can start with lowercase letters.

storage/reader.go
```diff
@@ -149,7 +149,13 @@ func (o *ObjectHandle) NewRangeReader(ctx context.Context, offset, length int64)
...
+		if err := setConditionsHeaders(req.Header, o.conds); err != nil {
+			return nil, err
+		}
+		if gen >= 0 {
```
tritone, line 155: This possibly deserves a comment making the distinction clearer between generation vs. conditions that operate on generation

tritone (commented):
Nice fix! Side note, the docs for this in the XML API reference are kind of terrible...

storage/reader.go
cojenco: Thanks for catching that. Removed.

storage/integration_test.go
cojenco: Ack, I'll leave it as is for consistency in the integration test file.

storage/storage.go
cojenco: Done. That makes sense, thanks!

tritone (approved):
Nice work on this!!

PR #4480, opened 2021-07-22
fix(spanner/spansql): fix DATE and TIMESTAMP parsing.

nakatamixi:
This enables to parse below cases.
- DATE, TIMESTAMP used as identifier.(these words does not reserved
  keyword.)
- DATE, TIMESTAMP used as function.


spanner/spansql/parser.go
```diff
@@ -2714,11 +2714,15 @@ func (p *parser) parseLit() (Expr, *parseError) {
 		p.back()
 		return p.parseArrayLit()
 	case tok.caseEqual("DATE"):
-		p.back()
-		return p.parseDateLit()
+		if !p.sniff("=") && !p.sniff(",") && !p.sniff("FROM") && !p.sniff("AS") {
```
olavloite, line 2717: I think this should be implemented using a positive look-ahead instead of a negative-look ahead. In other words: If the `DATE` keyword is followed by a string literal, then it is a date literal. Otherwise it is not. The current implementation does not work for for example the following valid query:

```sql
SELECT UNIX_DATE(DATE "2008-12-25")
```

Also, adding additional exceptions to the current if-not list will not solve that permanently, as there might be changes to the query syntax that would add additional keywords (i.e. not only `AS` and `FROM`) that could follow an identifier.

olavloite (changes requested):

spanner/spansql/keywords.go
```diff
@@ -158,6 +158,7 @@ var allFuncs = []string{
 	"BYTE_LENGTH", "CHAR_LENGTH", "CHARACTER_LENGTH",
 	"CODE_POINTS_TO_BYTES", "CODE_POINTS_TO_STRING",
 	"CONCAT",
+	"DATE",
```
olavloite, line 161: nit: Do not add this to the list of string functions (see comment above the list)

nakatamixi: Thanks!
Fixed and added more date timestamp functions at 0cb3b33e20d8b6add71d7f8e52c5d7fd60926a7e

spanner/spansql/parser.go
nakatamixi: Agreeed.

Fixed at 007bcf2db5643fe597ae34cf5b0b9411b76fbd36

olavloite (approved):
LGTM

hengfengli (approved):

nakatamixi:
Thank you!

PR #4482, opened 2021-07-22
feat(spanner): allow untyped nil values in parameterized queries

### LINKED ISSUE #4481: spanner: Allow untyped nil parameter values
Cloud Spanner allows query and DML statements to include named parameter. Each parameter __must__ be assigned a value and __may__ be assigned a type when the statement is executed. The Go client will automatically detect and set the type of a parameter based on the value that is passed in. This works for all non-nil values and for typed nil values. It however refuses to accept untyped nil values, although an untyped nil could be translated to a simple untyped `NULL` which will be accepted by Cloud Spanner in most cases.

Consider for example the following example:

```sql
INSERT INTO Singers (SingerId, Name, BirthDate) VALUES (@id, @name, @birth_date)
```

Executing this statement using non-nil value for `@id` and `@name` and an untyped `nil` value for `@birth_date` will fail in the Go client, as it requires a typed `nil` to be given. Cloud Spanner however accepts such a statement without problems if it is given an untyped `NULL` parameter value for `@birth_date`.

This restriction makes it very hard to support ORM frameworks like gorm, as these frameworks tend to generate untyped nil values for many statements.

olavloite:
Removes the restriction that nil valued parameters must be typed. The current restriction prevents the usage of DML statements with one or more untyped nil parameter values to be used to insert/update a value to NULL. Cloud Spanner does not require all parameters to be typed, only when it would otherwise be ambiguous what type should be used (e.g. non-NULL STRING/BYTES values).

This restriction also prevents the usage of the go/sql driver with gorm, as gorm will generate statements with untyped nil parameter values.

Fixes #4481

### DIFF
--- a/spanner/integration_test.go
+++ b/spanner/integration_test.go
@@ -1736,7 +1736,7 @@ func TestIntegration_BasicTypes(t *testing.T) {
 	// Write rows into table first using DML. Only do this on real Spanner
 	// as the emulator does not support untyped parameters.
 	// TODO: Remove when the emulator supports untyped parameters.
-	if isEmulatorEnvSet() {
+	if !isEmulatorEnvSet() {
 		statements := make([]Statement, 0)
 		for i, test := range tests {
 			stmt := NewStatement(fmt.Sprintf("INSERT INTO Types (RowId, `%s`) VALUES (@id, @value)", test.col))
--- a/spanner/integration_test.go
+++ b/spanner/integration_test.go
@@ -1715,9 +1715,59 @@ func TestIntegration_BasicTypes(t *testing.T) {
 		{col: "NumericArray", val: []NullNumeric(nil)},
 		{col: "NumericArray", val: []NullNumeric{}},
 		{col: "NumericArray", val: []NullNumeric{{n1, true}, {n2, true}, {}}},
+		{col: "String", val: nil, want: NullString{}},
+		{col: "StringArray", val: nil, want: []NullString(nil)},
+		{col: "Bytes", val: nil, want: []byte(nil)},
+		{col: "BytesArray", val: nil, want: [][]byte(nil)},
+		{col: "Int64a", val: nil, want: NullInt64{}},
+		{col: "Int64Array", val: nil, want: []NullInt64(nil)},
+		{col: "Bool", val: nil, want: NullBool{}},
+		{col: "BoolArray", val: nil, want: []NullBool(nil)},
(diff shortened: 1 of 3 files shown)


spanner/integration_test.go
```diff
@@ -1715,9 +1715,63 @@ func TestIntegration_BasicTypes(t *testing.T) {
...
+		{col: "Timestamp", val: nil, want: NullTime{}},
+		{col: "TimestampArray", val: nil, want: []NullTime(nil)},
+		{col: "Numeric", val: nil, want: NullNumeric{}},
+		{col: "NumericArray", val: nil, want: []NullNumeric(nil)},
+	}
+
+	// Write rows into table first using DML. Only do this on real Spanner
+	// as the emulator does not support untyped parameters.
+	// TODO: Remove when the emulator supports untyped parameters.
+	if isEmulatorEnvSet() {
```
hengfengli, line 1739: Shouldn't this be `!isEmulatorEnvSet()` if the emulator env is not set?

olavloite: Yep :+1:

spanner/statement_test.go
```diff
@@ -159,6 +159,12 @@ func TestConvertParams(t *testing.T) {
 			listProto(listProto(intProto(10)), listProto(intProto(20))),
 			listType(structType(mkField("field", intType()))),
 		},
+		// Untyped null
+		{
+			nil,
+			nullProto(),
+			nil,
```
hengfengli, line 166: Just curious: would this work as it expects a *sppb.Type but we give nothing. E.g., 

```
func numericType() *sppb.Type {
	return &sppb.Type{Code: sppb.TypeCode_NUMERIC}
}
```

hengfengli (approved):
LGTM.

spanner/statement_test.go
olavloite: In this test it would fail, as the test will verify that the param type is not set (or set to nil). That comparison will fail if we set it to `TypeCode_NUMERIC`. If we were to do something like that on production, it will also fail. I tried it out with the integration test, and it will refuse to insert a `NULL` value with type `NUMERIC` into a `STRING` column:

```
failed to insert values using DML: spanner: code = "InvalidArgument", desc = "Value has type NUMERIC which cannot be inserted into column String, which has type STRING [at 1:50]\\nINSERT INTO Types (RowId, `String`) VALUES (@id, @value)
```

PR #4489, opened 2021-07-22
feat: add schema support

hongalex:
This was previously reviewed in #3820 but was merged into a branch and released as a beta as the feature wasn't GA yet.


anguillanneuf (approved):
Thank you!

PR #4496, opened 2021-07-26
feat(spanner/spansql): add ROW DELETION POLICY parsing

nakatamixi:
This enables to parse `ROW DELETION POLICY` at `CREATE TABLE`, `ALTER
TABLE` clauses.

https://cloud.google.com/spanner/docs/ttl

(This feature is currently preview.)


hengfengli:
@olavloite It's worth to note that this feature is currently preview. So it is not open to the public yet.

spanner/spansql/parser.go
```diff
@@ -1236,6 +1245,15 @@ func (p *parser) parseAlterTable() (*AlterTable, *parseError) {
 			return a, nil
 		}
 
+		if p.eat("ROW") {
```
olavloite, line 1248: Suggestion: Would it maybe make more sense to check if you can eat `ROW DELETION POLICY` here, and then only parse the actual expression. The same would then also apply to the `REPLACE ROW DELETION POLICY`?

spanner/spansql/sql.go
```diff
@@ -130,6 +133,16 @@ func (ac AlterColumn) SQL() string {
...
 }
 
+func (ardp AddRowDeletionPolicy) SQL() string {
+	return "ADD " + ardp.RowDeletionPolicy.SQL()
+}
+
+func (rrdp ReplaceRowDeletionPolicy) SQL() string {
+	return "REPLACE " + rrdp.RowDeletionPolicy.SQL()
+}
+func (drdp DropRowDeletionPolicy) SQL() string {
```
olavloite, line 143: nit: add empty line between methods

olavloite (approved):
Looks good to me. I have a question/suggestion on where to start the parsing of the row deletion policy expression, but feel free to leave it as is if you think that is better.

spanner/spansql/sql.go
nakatamixi: fixed at 6091cab77c0e0cf90ecbe8c3799449b9d7b20309

spanner/spansql/parser.go
nakatamixi: Thanks!

Certainly, this makes easier to switch if another syntax is added after the ROW token.

I fixed at d83fb538d9bd852a452a8db1811612b11eefac43

olavloite (approved):

hengfengli (approved):

PR #4502, opened 2021-07-26
fix(bigquery/storage/managedwriter): fix double-close error, add tests

shollyman:
Writing tests picked up the error, so hooray.

BufferedStream integration test exposed that while the API surface is in preview without special requirements, advanced features such as the FlushRows rpc used by BufferedStream does.  This has been whitelisted for test projects, but we'll want to add this to doc.go when I start that PR.


bigquery/storage/managedwriter/integration_test.go
```diff
@@ -205,3 +208,161 @@ func TestIntegration_ManagedWriter_BasicOperation(t *testing.T) {
...
+	)
+	if err != nil {
+		t.Fatalf("NewManagedStream: %v", err)
+	}
+
+	info, err := ms.c.getWriteStream(ctx, ms.streamSettings.streamID)
+	if err != nil {
+		t.Errorf("couldn't get stream info: %v", err)
+	}
+	log.Printf("info id: %s type: %s", info.GetName(), info.GetType().String())
```
codyoss, line 328: If you want to log here I would recommend `t.Log`

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -139,6 +140,7 @@ func (ms *ManagedStream) FlushRows(ctx context.Context, offset int64) (int64, er
 			Value: offset,
 		},
 	}
+	log.Printf("flush req: %s", req.String())
```
codyoss, line 143: nit: remove

codyoss (changes requested):
Just a couple of small things

bigquery/storage/managedwriter/integration_test.go
shollyman: Sorry, this (and the other comment) was due to some debugging I was doing, which exposed the need to backend whitelist some functionality.  Cleared it out.

bigquery/storage/managedwriter/managed_stream.go
shollyman: done

codyoss (approved):

PR #4507, opened 2021-07-28
test(bigquery/storage/managedwriter): test all stream types

shollyman:
This PR adds testing to support all stream types, and refactors the
integration test to support parallel testing and reduce some of the
boilerplate code.

### DIFF
--- a/bigquery/storage/managedwriter/integration_test.go
+++ b/bigquery/storage/managedwriter/integration_test.go
@@ -40,6 +40,11 @@ var (
 	defaultTestTimeout = 15 * time.Second
 )
 
+var testSimpleSchema = bigquery.Schema{
+	{Name: "name", Type: bigquery.StringFieldType, Required: true},
+	{Name: "value", Type: bigquery.IntegerFieldType, Required: true},
+}
+
 func getTestClients(ctx context.Context, t *testing.T, opts ...option.ClientOption) (*Client, *bigquery.Client) {
 	if testing.Short() {
 		t.Skip("Integration tests skipped in short mode")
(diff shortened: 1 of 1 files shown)


bigquery/storage/managedwriter/integration_test.go
```diff
@@ -66,28 +71,28 @@ func getTestClients(ctx context.Context, t *testing.T, opts ...option.ClientOpti
...
-func validateRowCount(ctx context.Context, t *testing.T, client *bigquery.Client, tbl *bigquery.Table, expectedRows int64) {
+func validateRowCount(ctx context.Context, t *testing.T, client *bigquery.Client, tbl *bigquery.Table, expectedRows int64, description string) {
 
 	// Verify data is present in the table with a count query.
 	sql := fmt.Sprintf("SELECT COUNT(1) FROM `%s`.%s.%s", tbl.ProjectID, tbl.DatasetID, tbl.TableID)
 	q := client.Query(sql)
 	it, err := q.Read(ctx)
 	if err != nil {
-		t.Errorf("failed to issue validation query: %v", err)
+		t.Errorf("failed to issue validation query %s: %v", description, err)
```
codyoss, line 81: It may be nice to use %q for strings.

bigquery/storage/managedwriter/integration_test.go
```diff
@@ -136,13 +141,35 @@ func TestIntegration_ManagedWriter_DefaultStream(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTestTimeout)
 	defer cancel()
 
+	t.Run("group", func(t *testing.T) {
```
codyoss, line 144: I don't think this t.Run is necessary, right? I this `go test` would group these all under TestIntegration_ManagedWriter

codyoss (commented):
LGTM, just a couples that you can choose to do if you want.

codyoss (approved):

bigquery/storage/managedwriter/integration_test.go
shollyman: It's a context nuance mostly.  The group runs non-parallel, which prevents the deferred context cancellation to fire before it completes.  If you add the parallel tests directly, the deferred calls fire while subtests are still running.

bigquery/storage/managedwriter/integration_test.go
shollyman: swapped.

bigquery/storage/managedwriter/integration_test.go
codyoss: Oops, I had not noticed!

PR #4511, opened 2021-07-28
feat: wrap compute Operation

noahdietz:


compute/apiv1/addresses_client.go
```diff
@@ -51,9 +51,9 @@ type internalAddressesClient interface {
 	setGoogleClientInfo(...string)
 	Connection() *grpc.ClientConn
 	AggregatedList(context.Context, *computepb.AggregatedListAddressesRequest, ...gax.CallOption) (*computepb.AddressAggregatedList, error)
-	Delete(context.Context, *computepb.DeleteAddressRequest, ...gax.CallOption) (*computepb.Operation, error)
+	Delete(context.Context, *computepb.DeleteAddressRequest, ...gax.CallOption) (*Operation, error)
 	Get(context.Context, *computepb.GetAddressRequest, ...gax.CallOption) (*computepb.Address, error)
-	Insert(context.Context, *computepb.InsertAddressRequest, ...gax.CallOption) (*computepb.Operation, error)
+	Insert(context.Context, *computepb.InsertAddressRequest, ...gax.CallOption) (*Operation, error)
```
vchudnov-g, line 56: Where is `Operation` defined? Sorry if I'm missing something obvious.

(BTW, for companion PRs, consider using compute_small.proto...or doing a Showcase-based companion. Both for easier review)

noahdietz: > Where is `Operation` defined? Sorry if I'm missing something obvious.

It is defined as an "auxiliary type", like an Iterator, so the entire package uses it and it is generated in just 1 of the files in the package, not alongside each method. It seems that it is in [this file](https://github.com/googleapis/google-cloud-go/pull/4511/files#diff-b23791b960a18d989f0c1e1600e5401a76ab799895a0b07c1eb94b6a711e816e). There is al ong standing improvement I think we could make that moves all shared auxiliary types into a generated `aux.go` file instead of having them randomly dispersed across the client files.

> (BTW, for companion PRs, consider using compute_small.proto...or doing a Showcase-based companion. Both for easier review)

Those things don't have smoke tests already that will run in CI, where as the full compute proto/client does here.

vchudnov-g: I'm not seeing the comment "represents a long running operation for this API" that I see in the generator PR.

noahdietz: image

vchudnov-g: Ah, OK. Yes, I support putting all these auxiliary files shared among services into one central place. It's very arbitrary otherwise. Thanks for the pointer!

> Those things don't have smoke tests already that will run in CI, where as the full compute proto/client does here.
True, but is the point of the companion PR to make it easier for humans to review the generated output or to validate against CI? Maybe we should have two companions, one for each of these purposes.

PR #4512, opened 2021-07-28
feat(bigquery/storage/managedwriter): add opencensus instrumentation

shollyman:
This starts to plumb in oc instrumentation, as we're already using it in
bigquery proper and other veneers.

Testing instrumentation helped catch another double-close in the recv processor, so this
addresses that as well.


codyoss (approved):
LGTM

PR #4517, opened 2021-07-29
testing(bigquery/storage/managedwriter): extend table validation functionality

shollyman:
Soon we'll start tackling null values and complex schemas, so this PR augments the existing table validator with one that can be passed multiple validation constraints for evaluation (how many rows, how many nulls, cardinality, etc).  It updates the existing integration tests to use the new validator, and tightens validation to ensure that tests are propagating the appended values as expected.


codyoss (approved):
LGTM

PR #4520, opened 2021-07-30
feat(pubsub): add topic message retention duration

hongalex:
This PR adds support for topic retention duration. This is a new field that can be set on topic creation, and can also be changed by calling `topic.Update`. If set to a negative value for an `Update`, this clears the retention duration from the topic.


pubsub/subscription.go
```diff
@@ -256,6 +256,14 @@ type SubscriptionConfig struct {
...
 	// the endpoint will not be made.
 	Detached bool
+
+	// TopicMessageRetentionDuration indicates the minimum duration for which a message is retained
+	// after it is published to the subscription's topic. If this field is set,
+	// messages published to the subscription's topic in the last
+	// `TopicMessageRetentionDuration` are always available to subscribers.
+	//
+	// This field is set only in responses from the server and otherwise ignored.
+	TopicMessageRetentionDuration time.Duration
```
noahdietz, line 266: Noob question: how is this different from the `RetentionDuration time.Duration` field? It might be worth documenting the difference.

pubsub/subscription.go
```diff
@@ -256,6 +256,14 @@ type SubscriptionConfig struct {
 	// FAILED_PRECONDITION. If the subscription is a push subscription, pushes to
 	// the endpoint will not be made.
 	Detached bool
+
+	// TopicMessageRetentionDuration indicates the minimum duration for which a message is retained
+	// after it is published to the subscription's topic. If this field is set,
+	// messages published to the subscription's topic in the last
+	// `TopicMessageRetentionDuration` are always available to subscribers.
+	//
+	// This field is set only in responses from the server and otherwise ignored.
```
noahdietz, line 265: To be clear: this is an "output only" field? I know this isn't a proto field, but just want to clarify with https://google.aip.dev/203#output-only

pubsub/topic.go
```diff
@@ -100,6 +101,10 @@ type PublishSettings struct {
 	//
 	// Defaults to DefaultPublishSettings.BufferedByteLimit.
 	BufferedByteLimit int
+
+	// MessageRetentionDuration indicates the minimum duration to retain a message after
+	// it is published to the topic.
```
noahdietz, line 106: Does this mean that the message may or may not be dropped after the `MinimumRetentionDuration` has passed? Is there any behavior that can be documented for end users to help them make an educated choice for this field?

pubsub/topic.go
```diff
@@ -316,6 +337,15 @@ func (t *Topic) updateRequest(cfg TopicConfigToUpdate) *pb.UpdateTopicRequest {
 		pt.MessageStoragePolicy = messageStoragePolicyToProto(cfg.MessageStoragePolicy)
 		paths = append(paths, "message_storage_policy")
 	}
+	if cfg.RetentionDuration != 0 {
+		if cfg.RetentionDuration > 0 {
+			pt.MessageRetentionDuration = durationpb.New(cfg.RetentionDuration)
+		} else {
+			// If set to a negative value, clear MessageRetentionDuration.
+			pt.MessageRetentionDuration = nil
+		}
```
noahdietz, line 346: Nit, total matter of preference: invert the if/else to make the positive-value case the "green path" and the negative-value case a special case (thus indented).
```suggestion
              pt.MessageRetentionDuration = durationpb.New(cfg.RetentionDuration)
		if cfg.RetentionDuration < 0 {
			// If set to a negative value, clear MessageRetentionDuration.
			pt.MessageRetentionDuration = nil
		}
```

pubsub/subscription.go
hongalex: Yeah, this is an output only field, correspondig to https://github.com/googleapis/googleapis/blob/ba30d8097582039ac4cc4e21b4e4baa426423075/google/pubsub/v1/pubsub.proto#L761.

Should I explicitly add this to the comments here?

pubsub/subscription.go
hongalex: This is essentially the same thing but you can have both set and the maximum value will be applied (https://cloud.google.com/pubsub/docs/replay-overview#configuring_message_retention).

I'll add this as a comment

pubsub/topic.go
hongalex: Messages are dropped after all of a topic's subscriptions have acknowledge the message. I'll add a link to the cloud docs.

pubsub/topic.go
hongalex: Done.

pubsub/subscription.go
noahdietz: > Should I explicitly add this to the comments here?

It's fine the way it is, but maybe it would be clearer to say `This is an output only field, meaning it will only appear in responses from the backend and will be ignored if sent in a request.` You could include the link too if you feel it would be helpful.

No change is really needed, just wanted to clarify for myself.

pubsub/topic.go
```diff
@@ -337,11 +350,11 @@ func (t *Topic) updateRequest(cfg TopicConfigToUpdate) *pb.UpdateTopicRequest {
...
 	}
-	if cfg.RetentionDuration != 0 {
-		if cfg.RetentionDuration > 0 {
-			pt.MessageRetentionDuration = durationpb.New(cfg.RetentionDuration)
-		} else {
-			// If set to a negative value, clear MessageRetentionDuration.
+	if cfg.RetentionDuration != nil {
+		r := optional.ToDuration(cfg.RetentionDuration)
+		pt.MessageRetentionDuration = durationpb.New(r)
+		if r <= 0 {
```
noahdietz, line 356: You've switched to `0` being meaning "clear the retention duration", are you sure? If so, please also update the field documentation (that right now just says "if negative"

pubsub/topic.go
```diff
@@ -238,17 +247,21 @@ type TopicConfigToUpdate struct {
...
-	return TopicConfig{
+	tc := TopicConfig{
 		Labels:               pbt.Labels,
 		MessageStoragePolicy: protoToMessageStoragePolicy(pbt.MessageStoragePolicy),
 		KMSKeyName:           pbt.KmsKeyName,
 		SchemaSettings:       protoToSchemaSettings(pbt.SchemaSettings),
-		RetentionDuration:    pbt.MessageRetentionDuration.AsDuration(),
 	}
+	if pbt.MessageRetentionDuration != nil {
+		tc.RetentionDuration = pbt.MessageRetentionDuration.AsDuration()
```
noahdietz, line 262: A super nit: It is best practice to use the accessor methods when _reading_ a proto field value i.e. `pbt.GetMessageRetentionDuration()`. For example, if a field is a `*string`, `GetFoo` would deref `if Foo != nil` or return `""` `if Foo == nil`.

hongalex: Done

pubsub/topic.go
hongalex: Oops, this was actually the case before too. I kept the behavior as "just negative".

pubsub/subscription.go
noahdietz: SGTM

noahdietz (approved):
LGTM for general review.

pubsub/topic.go
```diff
@@ -200,6 +198,33 @@ type TopicConfig struct {
...
+	// RetentionDuration configures the minimum duration to retain a message
+	// after it is published to the topic. If this field is set, messages published
+	// to the topic in the last `message_retention_duration` are always available to subscribers.
+	// For instance, it allows any attached subscription to [seek to a
+	// timestamp](https://cloud.google.com/pubsub/docs/replay-overview#seek_to_a_time)
+	// that is up to `message_retention_duration` in the past. If this field is
+	// not set, message retention is controlled by settings on individual
+	// subscriptions. Cannot be more than 7 days or less than 10 minutes.
+	//
+	// For more information, see https://cloud.google.com/pubsub/docs/replay-overview#topic_retention.
```
lahuang4, line 211: We just changed the relative URL in the docs to #topic_message_retention.

pubsub/topic.go
```diff
@@ -200,6 +198,33 @@ type TopicConfig struct {
 	// Schema defines the schema settings upon topic creation. This cannot
 	// be modified after a topic has been created.
 	SchemaSettings *SchemaSettings
+
+	// RetentionDuration configures the minimum duration to retain a message
+	// after it is published to the topic. If this field is set, messages published
+	// to the topic in the last `message_retention_duration` are always available to subscribers.
```
lahuang4, line 204: Does it make sense to s/message_retention_duration/RetentionDuration in this comment?

lahuang4 (approved):

PR #4525, opened 2021-07-30
feat(internal/godocfx): mark status of deprecated items

tbpg:
A separate PR will update the template to use this new field.

A future PR here may include a `status` for the overall package and in the TOC.

### DIFF
--- a/internal/godocfx/parse.go
+++ b/internal/godocfx/parse.go
@@ -93,6 +93,7 @@ type item struct {
 	Examples []example `yaml:"codeexamples,omitempty"`
 	Children []child   `yaml:"children,omitempty"`
 	AltLink  string    `yaml:"alt_link,omitempty"`
+	Status   string    `yaml:"status,omitempty"`
 }
 
 func (p *page) addItem(i *item) {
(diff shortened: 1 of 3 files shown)


noahdietz (approved):

internal/godocfx/parse.go
```diff
@@ -310,6 +319,16 @@ func parse(glob string, workingDir string, optionalExtraFiles []string, filter [
 	}, nil
 }
 
+// getStatus returns a possibly empty status string for the given
+// docs.
+func getStatus(doc string) string {
+	deprecated := "\nDeprecated:"
```
dandhlee, line 325: Would it always work with a newline prefix? Is it possible that it's embedded within a line and only has a whitespace prefix instead?

dandhlee (approved):
LGTM, one minor concern below.

internal/godocfx/parse.go
tbpg: We may even be able to strengthen this to be two newlines before, but I think this is fine. See https://github.com/golang/go/wiki/Deprecated.

tbpg (commented):
Thank you!

broady (approved):

PR #4537, opened 2021-07-30
feat(.github): support dynamic submodule detection

codyoss:
Today the sub-modules are are coded into a couple of places to
work with release-please. This change is the first of three to
make release-please automatically support newly added modules.

Also, the workflow step was refactored to be in Go which makes
the code easier to test and debug locally in my opinion.


internal/actions/cmd/changefinder/main.go
```diff
@@ -0,0 +1,149 @@
...
+		rootDir = os.Args[1]
+	}
+	log.Printf("Root dir: %q", rootDir)
+	var modDirs []string
+	// Find all external modules
+	filepath.WalkDir(rootDir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.Name() == "go.mod" && !strings.Contains(path, "/internal/") {
```
tbpg, line 44: Can check if `Name` is `internal` and return `SkipDir` if so.

internal/actions/go.mod
```diff
@@ -0,0 +1,3 @@
+module cloud.google.com/go/internal/actions
```
tbpg, line 1: Why do we need a separate module for this?

codyoss: Technically no, but if I want to keep using newer io.FS apis, yes.

internal/actions/cmd/changefinder/main.go
codyoss: Done

tbpg (approved):

PR #4542, opened 2021-08-02
feat(compute)!: add pagination and an Operation wrapper

noahdietz:
This PR does multiple things due to the breaking nature of the code changes for `compute`.

Breaking changes:
- paginated APIs now return Iterators
- long running APIs now return a wrapped `Operation` type instead of the protobuf-go generated type directly

Integration tests are added for all of the new changes.

The gapic generator is updated with the aforementioned changes. To enable these changes, `genbot` will now supply the `diregapic` flag to the gapic generator based on if the API is generated via `googleapis-discovery`.


codyoss:
Force merging. This broke samples in a known way. This code is currently considered alpha.

PR #4544, opened 2021-08-02
chore(storage): remove Go 1.9 code

tritone:
We no longer support Go 1.9 so the code removed here is dead
and build tags are no longer required.


noahdietz (approved):

codyoss (approved):
LGTM

PR #4547, opened 2021-08-03
feat(internal/godocfx): add status to packages and TOCs

tbpg:
Follow up from https://github.com/googleapis/google-cloud-go/pull/4525.


codyoss (approved):
LGTM, thanks!

dandhlee (commented):
If we add other `Status` other than `Deprecated`, would that cause any of the other statuses to show because of https://github.com/googleapis/doc-templates/pull/247? I'm guessing not because we specifically look for `{{#deprecated}}`?

tbpg:
> If we add other Status other than Deprecated, would that cause any of the other statuses to show

Yes. The TOC plumbs through any status value. However, the `alpha` and `beta` statuses are not used for main doc content.

PR #4551, opened 2021-08-04
fix(spanner/spansql): only add comma after other option

olavloite:
Only add a comma to the options list if at least one other preceding option has been set.


hengfengli (approved):

olavloite:
The build error is caused by a failure in one of the samples for Google Compute, and so is unrelated to this change.

hengfengli:
@codyoss An unrelated test fails. Can we fix it to unblock this PR?

codyoss:
@hengfengli I just raised https://github.com/GoogleCloudPlatform/golang-samples/pull/2174 to unblock this.

PR #4558, opened 2021-08-05
fix(spanner): allow decoding null values to spanner.Decoder

### LINKED ISSUE #4552: spanner: Can not decode Null value to user defined custom type with DecodeSpanner
**Client**

Spanner

**Environment**

macOS

**Go Environment**

$ go version
`go version go1.16.6 darwin/amd64`
$ go env

**Code**

https://github.com/googleapis/google-cloud-go/blob/v0.90.0/spanner/value.go#L1330-L1338

```go
		// Check if the pointer is a custom type that implements spanner.Decoder
		// interface.
		if decodedVal, ok := ptr.(Decoder); ok {
			x, err := getGenericValue(v)
			if err != nil {
				return err
			}
			return decodedVal.DecodeSpanner(x)
		}
```

https://github.com/googleapis/google-cloud-go/blob/v0.90.0/spanner/value.go#L1911-L1923

```go
// getGenericValue returns the interface{} value encoded in proto3.Value.
func getGenericValue(v *proto3.Value) (interface{}, error) {
	switch x := v.GetKind().(type) {
	case *proto3.Value_NumberValue:
		return x.NumberValue, nil
	case *proto3.Value_BoolValue:
		return x.BoolValue, nil
	case *proto3.Value_StringValue:
		return x.StringValue, nil
	default:
		return 0, errSrcVal(v, "Number, Bool, String")
	}
}
```

**Expected behavior**

With user defined custom type that implements spanner.Decoder interface, how the Null value is handled can be defined in the `DecodeSpanner` function, and the expected behavior is that `DecodeSpanner` function is invoked when decode Null value.

**Actual behavior**

As the code logic listed, `getGenericValue` function will return error when the value type is not `Number`, `Bool`, `String`.
And when the value is Null, the value type is `*structpb.Value_NullValue`, an error returned, so the `DecodeSpanner` in custom type will not be invoked and an error returned.

```
cannot use null_value:NULL_VALUE(Kind: *structpb.Value_NullValue) as Number, Bool, String Value
```

olavloite:
Allow NULL values from the database to be passed in to the DecodeSpanner method
of a struct that implements spanner.Decoder.

Fixes #4552


hengfengli (approved):
LGTM. Thanks for the fix.

PR #4561, opened 2021-08-05
chore(tests): install kubectl using apt-get in test script

daniel-sanche:
Our [environment tests scripts](go/cdpe-ops-testgrid) are currently failing with the error message:

```
You cannot perform this action because the Cloud SDK component manager 
is disabled for this installation. You can run the following command 
to achieve the same result for this installation: 
```

This PR attempts to fix this by switching to the apt-get installation method


simonz130 (approved):

codyoss (approved):
LGTM, might be nice to move this to our base image so it does not need to be installed everytime.

PR #4565, opened 2021-08-06
test(bigtable): print hex using %x, sprintf to be valid utf-8

crwilcox:
casting byte vals to string was causing some chars to be (randomly) invalid for identifiers


codyoss (approved):
LGTM

PR #4570, opened 2021-08-09
Update reader.go

CarlJi:


codyoss:
Closing this PR as it does not seem to be adding anything. If you have a suggestion @CarlJi I would recommend starting with an issue first. Thanks.

Issue #1294, opened 2019-01-30
all: provide project ID detection


jadekler:
We should provide something like a `google.DefaultProjectID` method that attempts to discover project ID from ADC. This should probably live in `x/oauth2/google` (much of the logic for ADC lives there). This should be accompanied by documentation in clients that require project ID showing them how to use this.

Reference internal bug: 123588169

jba:
Not sure that's the right place. Is it possible to get the project ID from compute metadata even if the creds are explicitly supplied (e.g. a JSON file)? If so, and if that's something you want to support, then the function isn't really about creds. Maybe somewhere under google.golang.org/api?

jadekler:
That's a good point. google.golang.org/api makes sense. :+1:

zaddok:
How about a DefaultLocationId as well?

(The tasks api needs it as well as the project id.)

broady:
I don't think this can/should be implemented as a standalone function.

For example, if the user has provided a path to a service account file (using `WithCredentialsFile`) and that credentials file includes the project ID, that project ID should be used.

If this is implemented as a standalone function, it should also include the same API options as client creation.

I would lean towards being a sentinel value and the actual detection happening after API options have been processed.

While chatting with @jadekler today, I also came up with an idea to mitigate the issue of ambiguous project IDs (e.g., metadata server, environment variable, and credentials file with different IDs). The obvious solution is to introduce an ordering, similar to default credentials. However, connecting project IDs to default credentials is still dangerous and can still lead to unexpected data being mistakenly sent to the wrong project.

Since this is a convenience function, and there are many edge cases, I think it makes most sense to only allow detection when there is no ambiguity on detected project ID. That is, all mechanisms for detecting project ID should match. If they don't match (for example, using a service account credential from a different project within a Cloud Function), then an error is returned - the user must explicitly set their project ID.

A list of places to look for project IDs (non-exhaustive, please comment if there are more):

* project_id field in service account key file (from default credentials or set via client options)
* GOOGLE_CLOUD_PROJECT env var (used by GAE Flex)
* GCLOUD_PROJECT env var (used by GCF)
* metadata server (GAE, GCF, GKE, GCE)

broady:
Oh, to clear up any confusion, detection via metadata already exists:

https://godoc.org/cloud.google.com/go/compute/metadata#ProjectID

If you're looking for simple detection and you know you want the project ID from the metadata server, use this.

broady:
[...]

PR #4582, opened 2021-08-09
feat(internal/detect): add helper to detect projectID from env

codyoss:
This is meant to be used by certain veneers so that we can detect the projectID in a consistent manner. Future PRs will use this logic.
Fixes: #1294


codyoss:
cc @crwilcox Some of your libraries can use this in the future. I think you brought this up not too long ago...

noahdietz (approved):
LGTM but is there a way to add a test?

crwilcox:
How does this differ from https://github.com/googleapis/google-cloud-go/blob/master/firestore/client.go#L83 / https://github.com/googleapis/google-cloud-go/blob/master/firestore/client.go#L104

Is the goal here to centralize this as most veneers have this (more or less) duplicated code?

codyoss:
@crwilcox 

> Is the goal here to centralize this as most veneers have this (more or less) duplicated code?

Yes. Just wanted to have a consistent implementation for all to use.

codyoss:
@noahdietz Done.

PR #4583, opened 2021-08-09
chore(storage): add proto converters for metadata

noahdietz:


storage/acl.go
```diff
@@ -314,6 +345,15 @@ func (r ACLRule) toRawObjectAccessControl(bucket string) *raw.ObjectAccessContro
 	}
 }
 
+func (r ACLRule) toProtoObjectAccessControl(bucket string) *storagepb.ObjectAccessControl {
+	return &storagepb.ObjectAccessControl{
+		Entity: string(r.Entity),
+		Role:   string(r.Role),
+		// The other fields are not settable.
+		// TODO: Is that the case for gRPC?
```
tritone, line 353: Yes, only these fields can be updated in JSON https://cloud.google.com/storage/docs/json_api/v1/objectAccessControls/patch

...and really the other fields only make sense as read-only.

storage/storage.go
```diff
@@ -1132,6 +1135,39 @@ func (o *ObjectAttrs) toRawObject(bucket string) *raw.Object {
...
+	if !o.RetentionExpirationTime.IsZero() {
+		ret = timestamppb.New(o.RetentionExpirationTime)
+	}
+	var ct *timestamppb.Timestamp
+	if !o.CustomTime.IsZero() {
+		ct = timestamppb.New(o.CustomTime)
+	}
+	// For now, there are only globally unique buckets, and "_" is the alias
+	// project ID for such buckets.
+	b = bucket("_", b)
```
tritone, line 1150: Missed this in the read PR but we should probably rename this to indicate that it's the proto bucket path.

storage/storage.go
```diff
@@ -1132,6 +1135,39 @@ func (o *ObjectAttrs) toRawObject(bucket string) *raw.Object {
...
+		ContentType:         o.ContentType,
+		ContentEncoding:     o.ContentEncoding,
+		ContentLanguage:     o.ContentLanguage,
+		CacheControl:        o.CacheControl,
+		ContentDisposition:  o.ContentDisposition,
+		StorageClass:        o.StorageClass,
+		Acl:                 toProtoObjectACL(o.ACL),
+		Metadata:            o.Metadata,
+		CustomTime:          ct,
+		KmsKey:              o.KMSKeyName,
```
tritone, line 1167: Why is it necessary to add this field here and not in the apiary object?

storage/storage.go
```diff
@@ -1333,6 +1377,40 @@ func newObject(o *raw.Object) *ObjectAttrs {
 	}
 }
 
+func newObjectFromProto(r *storagepb.WriteObjectResponse) *ObjectAttrs {
+	o := r.GetResource()
+	if r == nil || o == nil {
+		return nil
+	}
+	return &ObjectAttrs{
+		Bucket:                  o.Bucket,
```
tritone, line 1386: Will this now return the path including the _ for projectID? If so we should remove that probably and return bucket name only.

storage/storage.go
```diff
@@ -1333,6 +1377,40 @@ func newObject(o *raw.Object) *ObjectAttrs {
...
+		Generation:              o.Generation,
+		Metageneration:          o.Metageneration,
+		StorageClass:            o.StorageClass,
+		CustomerKeySHA256:       o.GetCustomerEncryption().GetKeySha256(),
+		KMSKeyName:              o.GetKmsKey(),
+		Created:                 convertProtoTime(o.GetCreateTime()),
+		Deleted:                 convertProtoTime(o.GetDeleteTime()),
+		Updated:                 convertProtoTime(o.GetUpdateTime()),
+		CustomTime:              convertProtoTime(o.GetCustomTime()),
+	}
```
tritone, line 1411: I noticed ETag is missing, is that no longer provided via gRPC?

tritone (commented):
Good start! Two more thoughts

* I think we have unit tests for raw object encoding/decoding-- we should add the same for proto.
* This PR only covers object metadata and not bucket metadata. Let's change the title to indicate that.

storage/acl.go
noahdietz: Ok so I will remove the TODO.

storage/storage.go
noahdietz: Oop, sure thing.

storage/storage.go
noahdietz: Idk if it's necessary or not, but it's in the proto so I included it here. Should I remove it?

storage/storage.go
noahdietz: Good point, it will. Done!

storage/storage.go
noahdietz: There isn't an etag field in the protos so I guess not.

noahdietz:
@tritone I've added test cases PTAL!

storage/storage.go
```diff
@@ -1137,24 +1137,21 @@ func (o *ObjectAttrs) toRawObject(bucket string) *raw.Object {
 
 // toProtoObject copies the editable attributes from o to the proto library's Object type.
 func (o *ObjectAttrs) toProtoObject(b string) *storagepb.Object {
-	var ret *timestamppb.Timestamp
-	if !o.RetentionExpirationTime.IsZero() {
-		ret = timestamppb.New(o.RetentionExpirationTime)
-	}
-	var ct *timestamppb.Timestamp
-	if !o.CustomTime.IsZero() {
```
tritone: What happened here? We definitely need to be able to get/set CustomTime.

storage/storage.go
```diff
@@ -1137,24 +1137,21 @@ func (o *ObjectAttrs) toRawObject(bucket string) *raw.Object {
...
-		ct = timestamppb.New(o.CustomTime)
+	var checksums *storagepb.ObjectChecksums
+	if len(o.MD5) > 0 {
```
tritone, line 1141: Is it possible to have both checksums?

storage/storage.go
tritone: Hmm, looks like it might be an oversight in the current implementation which is concerning!

tritone (commented):
thanks for adding tests!

storage/storage.go
noahdietz: I moved them down to use an inline helper `toProtoTimestamp`

storage/storage.go
noahdietz: Yes, just checked, I for whatever reason thought it would be one or the other. Done.

storage/acl_test.go
```diff
@@ -63,3 +64,51 @@ func TestSetACL(t *testing.T) {
...
+
+func TestToProtoObjectACL(t *testing.T) {
+	for i, tst := range []struct {
+		rules []ACLRule
+		want  []*storagepb.ObjectAccessControl
+	}{
+		{nil, nil},
+		{
+			rules: []ACLRule{
+				{Entity: "foo", Role: "bar"},
```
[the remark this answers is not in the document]
noahdietz: Good idea, done!

PR #4587, opened 2021-08-10
chore(tests): fix google-cloud-sdk-app-engine-go installation in test script

daniel-sanche:
Related to https://github.com/googleapis/google-cloud-go/pull/4561. It looks like the GAE test is also failing for a similar reason.

This PR removes the `gcloud components install` line to install using `apt-get` instead. It also adds a ` | cat` so that the script will continue if installtion fails, in case the package ends up being unnecessary.

Note that this impacts the `environment.sh` test script only currently used by the logging library.


codyoss (approved):
LGTM, again, once these changes settle I think it would be good to move them up to the source image: https://github.com/googleapis/testing-infra-docker/blob/master/go/go116/Dockerfile

PR #4596, opened 2021-08-11
docs(storage): add comment to clarify BucketHandle.Object()

BrennaEpp:
Clarifies that the call does not create an object.


storage/bucket.go
```diff
@@ -131,6 +131,7 @@ func (b *BucketHandle) DefaultObjectACL() *ACLHandle {
 
 // Object returns an ObjectHandle, which provides operations on the named object.
 // This call does not perform any network operations.
+// The object is only created in the bucket (if it does not already exist) once you write to the object.
```
tritone, line 134: I think this is basically implied by the previous line? Maybe change it to this:

```
// This call does not perform any network operations such as fetching the object or verifying its existence.
```

How does that sound?

tritone (approved):

PR #4597, opened 2021-08-11
chore(tests): fix kubectl installation in environment test script

daniel-sanche:
Related to https://github.com/googleapis/google-cloud-go/pull/4561

I had attempted to add manual kubectl installation to the environment test script, but there were a couple issues, and the test was still failing. I fixed the issues, and confirmed it works this time by finding and and running the changes inside the `gcr.io/cloud-devrel-kokoro-resources/go111` test container

Note that this impacts the environment.sh test script only currently used by the logging library.

@codyoss, you mentioned we should consider making these changes in the base image. Do you know where I could submit that change?


codyoss:
https://github.com/googleapis/testing-infra-docker/tree/master/go are all of the version that we use in the client libraries.

codyoss (approved):

PR #4601, opened 2021-08-11
feat(bigquery/storage/managedwriter): more instrumentation support

shollyman:
This PR adds a bytes metric to the list of defined instrumentation
metrics, and adds an additional key to track data origin.  Ability
for users to set the data origin comes a new WithDataOrigin option
that can be passed to the managed stream constructor.

This also does some minor refactoring of how opencensus view creation
is handled.


codyoss (approved):
LGTM

PR #4602, opened 2021-08-12
fix(pubsub): mitigate race in checking ordering config

### LINKED ISSUE #3626: pubsub: TestIntegration_OrderedKeys_JSON failed
This test failed!

I will stop commenting.

---

commit: 98637e089776292232bb7c039844680627ddade1
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/94fa7e9c-3713-4d19-9a3d-271588f72995), [Sponge](http://sponge2/94fa7e9c-3713-4d19-9a3d-271588f72995)
status: failed
<details><summary>Test output</summary><br><pre>integration_test.go:1282: timed out after 60s waiting for all messages to be received</pre></details>

hongalex:
We use `sync.once.Do(s.checkOrdering)` to perform one config check for a subscription's message ordering field when a message with an ordering key is received. This way, we know if we should deliver messages in order. However, since `sub.Receive` generates multiple goroutines, `s.enableOrdering` is not guaranteed to be read after the first invocation of `s.checkOrdering`. This PR makes this race happen less often by making the config check closer to when the message is received.

This cannot be fully eliminated without making the config call at the start of every single `Receive`. This behavior is less desirable, as we want to make the config call only when we have to, that is when we receive a message with an ordering key. 

Fixes #3626

### DIFF
--- a/pubsub/subscription.go
+++ b/pubsub/subscription.go
@@ -897,6 +897,9 @@ func (s *Subscription) Receive(ctx context.Context, f func(context.Context, *Mes
 				}
 				for i, msg := range msgs {
 					msg := msg
+					if msg.OrderingKey != "" {
+						s.once.Do(s.checkOrdering)
+					}
 					// TODO(jba): call acquire closer to when the message is allocated.
 					if err := fc.acquire(ctx, len(msg.Data)); err != nil {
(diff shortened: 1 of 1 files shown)


codyoss (approved):
LGTM

PR #4604, opened 2021-08-12
feat(storage): SignedUrl can use existing creds to authenticate

BrennaEpp:
Part of the effort to improve Signed URLs in the Go Storage Client.

Users can omit filling in GoogleAccessID, PrivateKey, and SignBytes if they are okay with default values for these.

### DIFF
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -16,15 +16,20 @@ package storage
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"reflect"
 	"time"
 
+	"cloud.google.com/go/compute/metadata"
 	"cloud.google.com/go/internal/optional"
 	"cloud.google.com/go/internal/trace"
 	"google.golang.org/api/googleapi"
+	"google.golang.org/api/iamcredentials/v1"
 	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
 	raw "google.golang.org/api/storage/v1"
 )
 
@@ -220,6 +225,93 @@ func (b *BucketHandle) newPatchCall(uattrs *BucketAttrsToUpdate) (*raw.BucketsPa
 		req.UserProject(b.userProject)
 	}
 	return req, nil
+}
+
+// todo  add
+func (b *BucketHandle) SignedURL(object string, opts *SignedURLOptions) (string, error) {
+	if opts.GoogleAccessID != "" && (opts.SignBytes != nil || len(opts.PrivateKey) > 0) {
+		return SignedURL(b.name, object, opts)
+	}
+	// Make a copy of opts so we don't modify the pointer parameter.
+	newopts := &SignedURLOptions{
+		GoogleAccessID:  opts.GoogleAccessID,
+		SignBytes:       opts.SignBytes,
+		PrivateKey:      opts.PrivateKey,
+		Method:          opts.Method,
+		Expires:         opts.Expires,
+		ContentType:     opts.ContentType,
+		Headers:         opts.Headers,
+		QueryParameters: opts.QueryParameters,
+		MD5:             opts.MD5,
+		Style:           opts.Style,
+		Insecure:        opts.Insecure,
+		Scheme:          opts.Scheme,
+	}
+
+	if newopts.GoogleAccessID == "" {
+		id, err := b.detectDefaultGoogleAccessID()
+		if err != nil {
+			return "", err
+		}
+		newopts.GoogleAccessID = id
+	}
+	if newopts.SignBytes == nil && len(newopts.PrivateKey) == 0 {
+		if len(b.c.creds.JSON) > 0 {
+			var sa struct {
+				PrivateKey string `json:"private_key"`
+			}
+			// Don't error out if we can't marshal, fallback to GCE check.
+			err := json.Unmarshal(b.c.creds.JSON, &sa)
+			if err == nil && sa.PrivateKey != "" {
+				newopts.PrivateKey = []byte(sa.PrivateKey)
+			}
+		}
+		if len(newopts.PrivateKey) == 0 {
+			newopts.SignBytes = b.defaultSignBytesFunc(newopts.GoogleAccessID)
+		}
+	}
+	return SignedURL(b.name, object, newopts)
+}
+
+func (b *BucketHandle) detectDefaultGoogleAccessID() (string, error) {
+	if len(b.c.creds.JSON) > 0 {
+		var sa struct {
+			ClientEmail string `json:"client_email"`
+		}
+		// Don't error out if we can't marshal, fallback to GCE check.
+		err := json.Unmarshal(b.c.creds.JSON, &sa)
+		if err == nil && sa.ClientEmail != "" {
+			return sa.ClientEmail, nil
+		}
+	}
+	if metadata.OnGCE() {
+		email, err := metadata.Email("default")
+		if err == nil && email != "" {
+			return email, nil
+		}
+	}
+	return "", fmt.Errorf("unable to detect default GoogleAccessID")
+}
+
+func (b *BucketHandle) defaultSignBytesFunc(email string) func([]byte) ([]byte, error) {
+	return func(in []byte) ([]byte, error) {
+		ctx := context.Background()
+		svc, err := iamcredentials.NewService(ctx, option.WithHTTPClient(b.c.hc))
+		if err != nil {
+			return nil, fmt.Errorf("unable to create iamcredentials client: %v", err)
+		}
+		resp, err := svc.Projects.ServiceAccounts.SignBlob(fmt.Sprintf("projects/-/serviceAccounts/%s", email), &iamcredentials.SignBlobRequest{
+			Payload: base64.StdEncoding.EncodeToString(in),
+		}).Do()
+		if err != nil {
+			return nil, fmt.Errorf("unable to sign bytes: %v", err)
+		}
+		out, err := base64.StdEncoding.DecodeString(resp.SignedBlob)
+		if err != nil {
+			return nil, fmt.Errorf("unable to base64 decode response: %v", err)
+		}
+		return out, nil
+	}
 }
 
 // BucketAttrs represents the metadata for a Google Cloud Storage bucket.
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -3694,6 +3694,42 @@ func TestIntegration_Scopes(t *testing.T) {
 
 }
 
+func TestBucketSignURL(t *testing.T) {
+	ctx := context.Background()
+	client := testConfig(ctx, t)
+	defer client.Close()
+	bkt := client.Bucket(bucketName)
+
+	obj := "testBucketSignedURL"
+	contents := []byte("test")
+	if err := writeObject(ctx, bkt.Object(obj), "text/plain", contents); err != nil {
+		t.Fatalf("writing: %v", err)
+	}
+
+	url, err := bkt.SignedURL(obj, &SignedURLOptions{
+		Method:  "GET",
+		Expires: time.Now().Add(30 * time.Second),
+	})
+	if err != nil {
+		t.Fatalf("unable to create signed URL: %v", err)
+	}
+	resp, err := http.Get(url)
+	if err != nil {
+		t.Fatalf("http.Get(%q) errored: %q", url, err)
+	}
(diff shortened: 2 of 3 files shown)


storage/bucket.go
```diff
@@ -223,6 +228,96 @@ func (b *BucketHandle) newPatchCall(uattrs *BucketAttrsToUpdate) (*raw.BucketsPa
 	return req, nil
 }
 
+// SignedURL returns a URL for the specified object. Signed URLs allow anyone
+// access to a restricted resource for a limited time without needing a
+// Google account or signing in. For more information about signed URLs, see
+// https://cloud.google.com/storage/docs/accesscontrol#signed_urls_query_string_authentication
```
tritone, line 234: Let's add docs discussing the difference between this and the original `SignedURL` function. Users should know that the auth arguments are optional and have a high-level view into how we fill them in.

storage/bucket.go
```diff
@@ -223,6 +228,96 @@ func (b *BucketHandle) newPatchCall(uattrs *BucketAttrsToUpdate) (*raw.BucketsPa
...
+func (b *BucketHandle) SignedURL(object string, opts *SignedURLOptions) (string, error) {
+	if opts.GoogleAccessID != "" && (opts.SignBytes != nil || len(opts.PrivateKey) > 0) {
+		return SignedURL(b.name, object, opts)
+	}
+	// Make a copy of opts so we don't modify the pointer parameter.
```
tritone, line 239: Add a copy/clone method to SignedURLOptions instead?

storage/bucket.go
```diff
@@ -223,6 +228,96 @@ func (b *BucketHandle) newPatchCall(uattrs *BucketAttrsToUpdate) (*raw.BucketsPa
...
+			return sa.ClientEmail, nil
+		}
+	}
+	if metadata.OnGCE() {
+		email, err := metadata.Email("default")
+		if err == nil && email != "" {
+			return email, nil
+		}
+	}
+	return "", fmt.Errorf("unable to detect default GoogleAccessID")
```
tritone, line 297: This message will be returned to the user, so let's make it more actionable. Users should be able to distinguish between the following cases:

* If the JSON is present but has a problem with unmarshalling.
* If we're on GCE but get an error from the metadata service.
* If JSON is not present and also we're not on GCE.

storage/bucket.go
```diff
@@ -223,6 +228,96 @@ func (b *BucketHandle) newPatchCall(uattrs *BucketAttrsToUpdate) (*raw.BucketsPa
...
+			return "", err
+		}
+		newopts.GoogleAccessID = id
+	}
+	if newopts.SignBytes == nil && len(newopts.PrivateKey) == 0 {
+		if len(b.c.creds.JSON) > 0 {
+			var sa struct {
+				PrivateKey string `json:"private_key"`
+			}
+			// Don't error out if we can't marshal, fallback to GCE check.
```
tritone, line 267: This comment probably belongs on L272?

storage/bucket.go
```diff
@@ -223,6 +228,96 @@ func (b *BucketHandle) newPatchCall(uattrs *BucketAttrsToUpdate) (*raw.BucketsPa
...
+			err := json.Unmarshal(b.c.creds.JSON, &sa)
+			if err == nil && sa.PrivateKey != "" {
+				newopts.PrivateKey = []byte(sa.PrivateKey)
+			}
+		}
+		if len(newopts.PrivateKey) == 0 {
+			newopts.SignBytes = b.defaultSignBytesFunc(newopts.GoogleAccessID)
+		}
+	}
```
tritone, line 276: We probably want an error here if for whatever reason PrivateKey or SignBytes hasn't been populated at this point.

storage/bucket.go
```diff
@@ -223,6 +228,96 @@ func (b *BucketHandle) newPatchCall(uattrs *BucketAttrsToUpdate) (*raw.BucketsPa
...
+			return email, nil
+		}
+	}
+	return "", fmt.Errorf("unable to detect default GoogleAccessID")
+}
+
+func (b *BucketHandle) defaultSignBytesFunc(email string) func([]byte) ([]byte, error) {
+	return func(in []byte) ([]byte, error) {
+		ctx := context.Background()
+		svc, err := iamcredentials.NewService(ctx, option.WithHTTPClient(b.c.hc))
```
tritone, line 303: Is it costly to re-create this service each time we want to generate a signed URL? Is there some way we can maintain this at the client level instead and reuse?

storage/storage.go
```diff
@@ -120,10 +124,19 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 	// need to account for STORAGE_EMULATOR_HOST override when setting the default endpoints.
 	if host := os.Getenv("STORAGE_EMULATOR_HOST"); host == "" {
 		// Prepend default options to avoid overriding options passed by the user.
-		opts = append([]option.ClientOption{option.WithScopes(ScopeFullControl), option.WithUserAgent(userAgent)}, opts...)
+		opts = append([]option.ClientOption{option.WithScopes(ScopeFullControl, "https://www.googleapis.com/auth/cloud-platform"), option.WithUserAgent(userAgent)}, opts...)
```
tritone, line 127: Could this create issues for users who don't have this scope enabled currently and are not trying to use Signed URL features?

storage/bucket.go
```diff
@@ -223,6 +228,96 @@ func (b *BucketHandle) newPatchCall(uattrs *BucketAttrsToUpdate) (*raw.BucketsPa
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to sign bytes: %v", err)
+		}
+		out, err := base64.StdEncoding.DecodeString(resp.SignedBlob)
+		if err != nil {
+			return nil, fmt.Errorf("unable to base64 decode response: %v", err)
+		}
+		return out, nil
+	}
+}
```
tritone, line 319: Let's make a TODO to add a wrapper for GenerateSignedPostPolicyV4 as well (can go in a separate PR).

storage/integration_test.go
```diff
@@ -3994,6 +3995,69 @@ func TestIntegration_Scopes(t *testing.T) {
 
 }
 
+func TestBucketSignURL(t *testing.T) {
```
tritone, line 3998: This is nice-- we can probably just replace existing signed URL integration tests with tests that use the wrapper method.

storage/integration_test.go
```diff
@@ -3994,6 +3995,69 @@ func TestIntegration_Scopes(t *testing.T) {
...
+
+	if testing.Short() && !replaying {
+		t.Skip("Integration tests skipped in short mode")
+	}
+
+	creds, err := findTestCredentials(ctx, "GCLOUD_TESTS_GOLANG_KEY", ScopeFullControl, "https://www.googleapis.com/auth/cloud-platform")
+	if err != nil {
+		t.Fatalf("unable to find test credentials: %v", err)
+	}
+	client, err := newTestClient(ctx, option.WithCredentials(creds))
```
tritone, line 4009: Is this creds magic necessary? What happens if you run the tests with a test client as-is?

tritone (commented):
Nice start on this! A few questions and comments, mainly around docs and error cases.

storage/bucket.go
codyoss: The costly part usually to creating a client is the auth/transport layer, since we are passing in a client here that cost is circumvented. If you would like we could cache the client, but I don't see to much benefit.

storage/storage.go
codyoss: I don't think so, scopes work hand-in-hand with IAM for access to things. The scope just says it is allowed to talk, IAM says you may talk.

This does bring up a documentation point though. Users potentially have to enable the IAM credentials API if they are using compute creds and that flow.

storage/bucket.go
tritone: Gotcha, that's good to know. Maybe this deserves a comment to that effect.

storage/storage.go
tritone: Yep for sure, we should definitely document that.

storage/integration_test.go
codyoss: So here we are explicitly testing the SA use-case and signing with it IIRC. By default we are authed with a TS so don't have the context to read some of the fields from the keyfile. For completeness we could also have a test that does the iamcredentials flow where we auth like normal but set the email address field. Maybe this bit of code deserves a comment 😛

storage/bucket.go
```diff
@@ -232,25 +233,21 @@ func (b *BucketHandle) newPatchCall(uattrs *BucketAttrsToUpdate) (*raw.BucketsPa
...
 // https://cloud.google.com/storage/docs/accesscontrol#signed_urls_query_string_authentication
+//
+// This method only requires the Method and Expires fields in the specified
+// SignedURLOptions opts to be non-nil. If not provided, it attempts to fill the
+// GoogleAccessID and PrivateKey from the GOOGLE_APPLICATION_CREDENTIALS environment variable.
+// If no private key is found, it attempts to use the GoogleAccessID to sign the URL.
+// This requires the IAM credentials API to be enabled and iam.serviceAccounts.signBlob
+// permissions on the service account.
+// If you do not want these fields set for you, you may use
+// SignedURL(bucket, name string, opts *SignedURLOptions) instead.
```
tritone, line 244: Alternatively, they can still use this method and provide the auth options themself-- we only attempt to fill in if there is something missing, correct?

storage/bucket.go
tritone: Hmm yeah I guess that's correct-- so then if there is an issue with signBytes, it will occur when it is called to do the signing. That seems fine. Can you verify that the error looks sensible if, for example, we don't have correct permissions and so signBytes fails?

storage/bucket.go
```diff
@@ -264,46 +261,68 @@ func (b *BucketHandle) SignedURL(object string, opts *SignedURLOptions) (string,
...
 		var sa struct {
 			ClientEmail string `json:"client_email"`
 		}
-		// Don't error out if we can't marshal, fallback to GCE check.
 		err := json.Unmarshal(b.c.creds.JSON, &sa)
 		if err == nil && sa.ClientEmail != "" {
 			return sa.ClientEmail, nil
+		} else {
+			returnErr = errors.New("storage: empty client email in credentials")
+			if err != nil {
```
tritone, line 294: Can we rearrange this if/else to remove the nesting? I think there are 3 cases, this seems clearer to me:

```
if err == nil && sa.ClientEmail != "" {
  return ...
} else if err != nil {
  returnErr = err
} else {
  returnErr = errors.New...
}
```

Same thing on L302 and following below.

tritone (commented):
Looking good!

storage/bucket.go
BrennaEpp: Correct, I'll fix the comment to make that clearer.

storage/bucket.go
BrennaEpp: Error if the service account email is not found by the IAM client (for example can occur if you use your email instead of a service account email or there's a typo in your email): 
`unable to sign bytes: googleapi: Error 404: Requested entity was not found.`
This error isn't the best but it's what we receive from the IAM api: https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/signBlob

Error if the service account does not have permissions:
`unable to sign bytes: googleapi: Error 403: The caller does not have permission`

Error if the Service Account Credentials API is not enabled: 
```
unable to sign bytes: googleapi: Error 403: IAM Service Account Credentials API has not been used in project 1074109021461 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/iamcredentials.googleapis.com/overview?project=1074109021461 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
Details:
[
  {
    "@type": "type.googleapis.com/google.rpc.Help",
    "links": [
      {
        "description": "Google developers console API activation",
        "url": "https://console.developers.google.com/apis/api/iamcredentials.googleapis.com/overview?project=1074109021461"
      }
    ]
  },
  {
    "@type": "type.googleapis.com/google.rpc.ErrorInfo",
    "domain": "googleapis.com",
    "metadata": {
      "consumer": "projects/1074109021461",
      "service": "iamcredentials.googleapis.com"
    },
    "reason": "SERVICE_DISABLED"
  }
]
```

storage/integration_test.go
BrennaEpp: I added a couple comments and the extra test case

PR #4605, opened 2021-08-12
feat(bigquery/storage/managedwriter): add project autodetection

shollyman:


codyoss:
If this is in not a big rush I suggest we wait to land this until https://github.com/googleapis/google-cloud-go/pull/4582 lands. Once that lands and is released I would also like to refactor all other veneers so that this logic is consistent.

shollyman:
> If this is in not a big rush I suggest we wait to land this until #4582 lands. Once that lands and is released I would also like to refactor all other veneers so that this logic is consistent.

updated logic to use internal/detect mechanism, but tests will fail here until we have a release with the change.

codyoss (approved):

PR #4606, opened 2021-08-13
fix(pubsub): always make config check to prevent race

### LINKED ISSUE #3626: pubsub: TestIntegration_OrderedKeys_JSON failed

hongalex:
In the previous iteration of this PR, #4602, the assumption was made that making the config check earlier will prevent race conditions in the test, which turned out to be incorrect. In addition, although previously ruled out as a solution, this PR makes it so that a config check is always made on the first call to `Receive`. I originally thought this would result in a poor experience for those users who don't have the `roles/pubsub.viewer` permission, but we default `enableOrdering` to `true` anyway if the config call fails due to lack of permissions. Using a default of `true` only negatively impacts those who don't want ordering on their subscription, and this behavior is [already documented](https://github.com/googleapis/google-cloud-go/blob/pubsub/v1.14.0/pubsub/subscription.go#L235-L239).

Fixes #3626

### DIFF
--- a/pubsub/subscription.go
+++ b/pubsub/subscription.go
@@ -48,7 +48,6 @@ type Subscription struct {
 	mu            sync.Mutex
 	receiveActive bool
 
-	once           sync.Once
 	enableOrdering bool
 }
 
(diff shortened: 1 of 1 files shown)


codyoss (approved):
LGTM

PR #4607, opened 2021-08-13
feat(pubsub): add list configs for topic & sub

### LINKED ISSUE #4559: pubsub: list subscription configs, not just names
I have a use case for listing subscriptions where I want to check information in the `SubscriptionConfig` object. The underlying gRPC `ListSubscriptions` call includes this information (or at least the portion I need -- I'm not entirely sure if it includes everything needed to form a SubscriptionConfig?),  but the way it is exposed via the `Subscriptions` API on `pubsub.Client` and the resulting iterator it returns discards all this information (https://github.com/googleapis/google-cloud-go/blob/master/pubsub/subscription.go#L90-L96), and requires a second gRPC call for each subscription to re-acquire it!

A couple possible solutions present themselves:
1. A `ListSubscriptionConfigs` API on `pubsub.Client` that returns an iterator exposing `SubscriptionConfig`s instead of `Subscription`s
2. A `NextConfig()` method on `SubscriptionIterator`

The workaround of calling `Config()` on every `Subscription` returned works, but is inefficient

hongalex:
When calling `client.Topics` and `client.Subscriptions`, you get back an iterator that allows you to iterate over all resources. However, this gives you the actual `Topic` and `Subscription` struct that's defined by the library and does not list configs, which currently need to be retrieved from an additional `config` call. This adds `NextConfig` to both `Topic/SubscriptionIterator` so that the config can be viewed instead of `Topic/Subscription`. This is the desired behavior, and is used for `SchemaConfig` as well as topic/subscription configs in Pub/Sub Lite.

Thanks to @mgabeler-lee-6rs for suggesting this.

Closes #4559


pubsub/subscription.go
```diff
@@ -111,6 +115,19 @@ func (subs *SubscriptionIterator) Next() (*Subscription, error) {
 	return &Subscription{c: subs.c, name: subName}, nil
 }
 
+// NextConfig returns the next subscription config. If there are no more subscriptions, iterator.Done will be returned.
```
noahdietz, line 118: I'm not really sure how the iterator is used elsewhere, but I can see that `it.Next` is captured in a closure for the `next` field. Is there any concern that a user will mix calls of `Next` (for the Subscription) and `NextConfig` (for the SubscriptionConfig) that will pull from the same iterator, thereby "losing" items?

hongalex: I sort of see your point. I'll add a note to `NextConfig` that this is shared with `Next` and make sure to create separate iterator instances if you want to use both.

noahdietz: Sounds good. You could also make it return both, like `NextWithConfig() (*Subscription, *SubscriptionConfig, error)`. That doesn't help with using `Next` and "losing" the config part - although that might be OK TBH, with a comment to use the `NextWithConfig` version if they want that.

hongalex: I'm fairly certain that most users probably want to retrieve the config only, and that the existing `Next` feature was probably unused. Retrieving `*Subscription` is only useful if you want to start multiple `Receive` calls, which I don't think is a common use case. I'm tempted to keep this as is for simplicity (retrieving both `Subscription` and `SubscriptionConfig` is too much for users).

noahdietz: SGTM!

noahdietz (approved):
LGTM but another pubsub reviewer might want to approve

kamalaboulhosn (approved):

mgabeler-lee-6rs:
@hongalex I just tried to pick this up in my app, and ... I realized I missed something pretty important in my original suggestion: the `SubscriptionConfig` and `TopicConfig` objects returned from these don't have any information about the _identity_ of the returned objects, which makes them nigh useless :man_facepalming:

What do you think about adding a `name` field and `String()` and `ID()` methods to `TopicConfig` and `SubscriptionConfig` to match `Topic` and `Subscription`?

hongalex:
Oof, yeah you're totally right. I think the original intent for Topic/SubscriptionConfig was purely for storing config (irregardless of the actual topic/config it is associated with). I'll draft up a fix for this today. Thanks for reporting as always!

mgabeler-lee-6rs:
@hongalex guessing this got lost in the shuffle (no hard feelings!). Want me to put an issue up for the followup fix?

PR #3509, opened 2021-01-07
storage: fix scheme lost when use endpoint

3pointer:
Client should choose the correct scheme when using the endpoint with scheme (http://localhost:8080). or it will return 
`Get https://localhost:8080/xxx: http: server gave HTTP response to HTTPS client"` when client try to use [NewRangeReader](https://github.com/googleapis/google-cloud-go/blob/c66c5b8799db244755c353f47d159830c8ee4e7c/storage/reader.go#L110) to read objects.

PR #4609, opened 2021-08-13
fix(storage): preserve supplied endpoint's scheme

BrennaEpp:
Fixes #3509

`option.WithEndpoint()` completely overrides `STORAGE_EMULATOR_HOST` environment variable.

Built on #4608

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -90,8 +90,6 @@ type Client struct {
 	raw *raw.Service
 	// Scheme describes the scheme under the current host.
 	scheme string
-	// EnvHost is the host set on the STORAGE_EMULATOR_HOST variable.
-	envHost string
 	// ReadHost is the default host used on the reader.
 	readHost string
 }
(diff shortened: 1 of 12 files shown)


storage/storage.go
```diff
@@ -139,12 +139,13 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
...
 		return nil, fmt.Errorf("storage client: %v", err)
 	}
-	// Update readHost with the chosen endpoint.
+	// Update readHost and scheme with the chosen endpoint.
 	u, err := url.Parse(ep)
 	if err != nil {
 		return nil, fmt.Errorf("supplied endpoint %q is not valid: %v", ep, err)
 	}
 	readHost = u.Host
+	scheme = u.Scheme
```
tritone, line 148: It looks like we set `scheme` in L114 & L123 above-- I guess this no longer does anything?

Also, is it possible for the returned endpoint from `htransport.NewClient` to have an empty scheme?

I'm thinking we should keep L114 & L123 as is and only set scheme here if `u.Scheme != ""`, just out of caution. Does that make sense?

tritone (commented):
New test cases look good.

storage/storage.go
BrennaEpp: The endpoint for `htransport.NewClient` should always have a scheme as long as we send in the `WithDefaultEndpoint` option. But yes that makes sense, since it does not actually promise to.

tritone (approved):

PR #4611, opened 2021-08-13
feat(spannertest): Add validation of duplicated column names

### LINKED ISSUE #4610: spannertest: Add validation of the duplicate column names
It seems `spannertest` doesn't check if there are the duplicated column names in new columns in a DDL or between new columns in a DDL and the existing columns.
Actually, the `UpdateDDL` method in `spannertest` doesn't return error against the following DDLs.

Case1: the duplicated columns in a DDL
ex) 
```.sql
CREATE TABLE Users (
    Id INT64 NOT NULL,
    Name STRING(MAX),
    Name STRING(MAX),
) PRIMARY KEY (Id);
```

Case2: the duplicated columns between the existing column and a newly added column.
ex) 
Apply this first (valid DDL),
```
CREATE TABLE Users (
    Id INT64 NOT NULL,
    Name STRING(MAX),
) PRIMARY KEY (Id);
```
Then, apply this (The new column is a duplicate of the existing one),
```
ALTER TABLE Users ADD COLUMN Name STRING(MAX);`
```

Jusr add that validation in the `addColumn` method.

sryoya:
Fixes #4610


spanner/spannertest/db.go
```diff
@@ -622,6 +622,10 @@ func (t *table) addColumn(cd spansql.ColumnDef, newTable bool) *status.Status {
 		return status.Newf(codes.InvalidArgument, "new non-key columns cannot be NOT NULL")
 	}
 
+	if _, ok := t.colIndex[cd.Name]; ok {
+		return status.Newf(codes.InvalidArgument, "duplicated column names: %s", cd.Name)
```
olavloite, line 626: Violation of the uniqueness of a column name for a table should use error code `AlreadyExists`:

```suggestion
		return status.Newf(codes.AlreadyExists, "column %s already exists", cd.Name)
```

sryoya: I see. Thanks, fixed now.

olavloite (approved):
LGTM

hengfengli (approved):

PR #4616, opened 2021-08-13
fix(storage): accept emulator env var without scheme

### LINKED ISSUE #4444: Storage: empty readHost when STORAGE_EMULATOR_HOST is set to host:port
**Client**

Storage

**Environment**

Darwin Kernel Version 20.5.0

**Go Environment**

$ go version
```
go version go1.16 darwin/amd64
```

$ go env
```
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOOS="darwin"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOVCS=""
GOVERSION="go1.16"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/dev/null"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/_r/pyfmdgr15fz3_x7hhfk7gb3r0000gp/T/go-build2794454919=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

**Code**

```go
package main

func main() {
	os.Setenv("STORAGE_EMULATOR_HOST", "localhost:4444")
	if err != nil {
		log.Fatal(err)
	}

	r, err := client.Bucket("bucket").Object("object").NewReader(context.TODO())
	if err != nil {
		log.Fatal(err)
	}
}
```

**Expected behavior**

return a reader to the object.

**Actual behavior**

Fatal with error: `http: no Host in request URL`,
URL: `http:///bucket/object`

**Additional context**

I believe this happens after [this call](https://github.com/googleapis/google-cloud-go/blob/master/storage/storage.go#L141). since `WithEndpoint` and `WithDefaultEndpoint` have different behaviour inside [this function](https://github.com/googleapis/google-api-go-client/blob/master/transport/internal/dca/dca.go#L98). There might be a modification there.

BrennaEpp:
You can now use the client library with an emulator by setting the environment variable to a host, with or without a scheme
You also no longer have to supply an endpoint to use the emulator; the endpoint is built for you

Fixes #4444


storage/storage.go
```diff
@@ -120,13 +120,26 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 		opts = append(opts, internaloption.WithDefaultEndpoint("https://storage.googleapis.com/storage/v1/"))
 		opts = append(opts, internaloption.WithDefaultMTLSEndpoint("https://storage.mtls.googleapis.com/storage/v1/"))
 	} else {
-		scheme = "http"
-		readHost = host
+		// Add scheme for user if not supplied in STORAGE_EMULATOR_HOST
+		var hostURL *url.URL
+		if strings.Contains(host, "://") {
```
tritone, line 125: Is this check necessary? Can you just parse the supplied URL and then add the scheme if it's not present already?

storage/storage.go
```diff
@@ -120,13 +120,26 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 		opts = append(opts, internaloption.WithDefaultEndpoint("https://storage.googleapis.com/storage/v1/"))
 		opts = append(opts, internaloption.WithDefaultMTLSEndpoint("https://storage.mtls.googleapis.com/storage/v1/"))
 	} else {
-		scheme = "http"
-		readHost = host
```
tritone, line 124: Is readHost no longer set here? Why not?

BrennaEpp: As discussed offline, we are removing these temporary variables as the scheme and readHost are set later anyway from the endpoint provided by `htransport`, which should be valid

storage/storage.go
BrennaEpp: Added a comment to clarify this - if the scheme is not supplied but the host is, then `url.Parse()` has undefined behaviour

storage/storage.go
```diff
@@ -110,23 +109,35 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
...
+		} else {
+			// Add scheme for user if not supplied in STORAGE_EMULATOR_HOST
+			// URL is only parsed correctly if it has a scheme, so we build it ourselves
+			hostURL = &url.URL{Scheme: "http", Host: host}
+		}
+
+		hostURL.Path = "storage/v1/"
+		endpoint := hostURL.String()
+
+		// Prepend the emulator host as default endpoint for the user
```
tritone, line 136: Looks like we are appending here, not prepending.

tritone (commented):
Just fix the comment, otherwise looks good!

tritone (approved):

PR #4625, opened 2021-08-16
feat(bigquery): switch to centralized project autodetect logic

shollyman:
Switches from a local implementation to the logic in internal/detect.


codyoss (approved):
LGTM

PR #4626, opened 2021-08-16
docs: move example inline into godoc

codyoss:
This change is needed, at least temporarily, to facilitate the
splitting of modules. The parent module can't depend on a module
that is about to be split out as tidying will fail to resolve
import paths as the code no longer exists.


doc.go
```diff
@@ -28,18 +28,96 @@ Authentication and Authorization
 
 All the clients in sub-packages support authentication via Google Application Default
 Credentials (see https://cloud.google.com/docs/authentication/production), or
-by providing a JSON key file for a Service Account. See the authentication examples
-in this package for details.
+by providing a JSON key file for a Service Account. See examples below.
+
+Google Application Default Credentials(ADC) is the recommended way to authorize
```
tbpg, line 33:
```suggestion
Google Application Default Credentials (ADC) is the recommended way to authorize
```

tbpg (changes requested):
We can't do a 2-step process to avoid this? Publish the new module then update the root module?

codyoss:
@tbpg I don't think so. Root module needs to be tagged first so that the sub-module can depend on it. Once tagged it can no longer reference the new child mods code without a new dependency on the child mod. I think we should avoid having the root depend on any gapic packages if possible.

tbpg (approved):
Ah, got it. Thanks for the explanation.

noahdietz (approved):

PR #4641, opened 2021-08-18
test(storage): add test for uploading with an emulator

BrennaEpp:
This test should catch an error such as #2476
Also added tests to ensure endpoint overrides emulator host when host is specified as host:port or scheme://host:port - this adds extra coverage to ensure something like #4444 does not occur again


storage/storage_test.go
```diff
@@ -1344,3 +1360,149 @@ func TestWithEndpoint(t *testing.T) {
...
+		originalScheme := c.scheme
+
+		op := "bucket creation"
+
+		c.Bucket("test-bucket").Create(ctx, "pid", nil)
+		if err != nil {
+			t.Errorf("%s: %v", tc.desc, err)
+		}
+
+		if c.raw.BasePath != originalRawBasePath {
```
tritone, line 1422: I think you can just do this set of checks once at the end of the test case.

storage/storage_test.go
```diff
@@ -1344,3 +1360,149 @@ func TestWithEndpoint(t *testing.T) {
...
+			StorageEmulatorHost: "",
+		},
+		{
+			desc:                "both emulator and endpoint specified",
+			CustomEndpoint:      "https://" + addr,
+			StorageEmulatorHost: "https://" + addr,
+		},
+	}
+	ctx := context.Background()
+	for _, tc := range testCases {
```
tritone, line 1404: Use t.Run for subtests maybe?

storage/storage_test.go
```diff
@@ -1344,3 +1360,149 @@ func TestWithEndpoint(t *testing.T) {
...
+			t.Errorf("%s: readHost changed after %s\n\tgot %v, original %v", tc.desc, op, c.readHost, originalReadHost)
+		}
+		if c.scheme != originalScheme {
+			t.Errorf("%s: scheme changed after %s\n\tgot %v, original %v", tc.desc, op, c.scheme, originalScheme)
+		}
+	}
+}
+
+func newTestServerAt(handler func(w http.ResponseWriter, r *http.Request)) (string, *http.Client, func()) {
+	ts := httptest.NewTLSServer(http.HandlerFunc(handler))
```
tritone, line 1495: Can you have the test server verify that the expected endpoint is being hit? That is probably more important than validating that the client fields are not being changed.

storage/storage_test.go
```diff
@@ -1344,3 +1360,149 @@ func TestWithEndpoint(t *testing.T) {
...
+		},
+		{
+			desc:                "both emulator and endpoint specified",
+			CustomEndpoint:      "https://" + addr,
+			StorageEmulatorHost: "https://" + addr,
+		},
+	}
+	ctx := context.Background()
+	for _, tc := range testCases {
+		os.Setenv("STORAGE_EMULATOR_HOST", tc.StorageEmulatorHost)
```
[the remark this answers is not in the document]
BrennaEpp: I think that's what I'm doing on line 1368 and 1369

storage/storage_test.go
BrennaEpp: Changed! My original idea was to provide information on which operation caused the change, but that's probably overkill and can easily be figured out if necessary by debugging.

storage/storage_test.go
BrennaEpp: good idea :)

storage/storage_test.go
BrennaEpp: I added some code for checking that the same endpoint is hit after completing the operations on the client, is that what you had in mind?

tritone: No, I'd like to have the server verify that the calls made to it are as expected (for example create bucket resulted in a call to `POST https://<hostname>/storage/v1/b` and not to some other endpoint). This will ensure that something problematic isn't happening at the apiary layer for example.

storage/storage_test.go
```diff
@@ -1361,20 +1361,23 @@ func TestWithEndpoint(t *testing.T) {
...
+	gotURL := make(chan string, 1)
+	gotMethod := make(chan string, 1)
 
-	addr, hClient, close := newTestServerAt(func(w http.ResponseWriter, r *http.Request) {
+	hClient, close := newTestServer(func(w http.ResponseWriter, r *http.Request) {
 		io.Copy(ioutil.Discard, r.Body)
 		fmt.Fprintf(w, "{}")
-		req, _ := url.Parse(r.RequestURI)
-		lastRequestEndpoint = req.RawPath
+		gotURL <- r.RequestURI
```
tritone, line 1379: Can you have this send the scheme and hostname to be checked as well?

storage/storage_test.go
```diff
@@ -1417,76 +1420,90 @@ func TestOperationsWithEndpoint(t *testing.T) {
...
 
-			if lastRequestEndpoint != createBucketEndpoint {
-				t.Errorf("unexpected change in endpoint for creating a bucket\n\tgot:\t\t%v\n\toriginal:\t%v", lastRequestEndpoint, createBucketEndpoint)
+			// Check that the calls made to the server are as expected
+			// given the operations performed
+			for _, op := range operations {
+				if err := op.runOp(); err != nil {
+					t.Errorf("%s: %v", op.desc, err)
+				}
+				u, method := <-gotURL, <-gotMethod
```
tritone, line 1483: What happens if the channel never sends? Should we add some kind of timeout to each subtest?

tritone (commented):
Looks good! Two small things.

storage/storage_test.go
BrennaEpp: added the hostname - not sure how to get the scheme here - the test server is a TLS server so the requests should always be https and it will fail if you try using http

storage/storage_test.go
BrennaEpp: Good point, I added a timeout.

Issue #4642, opened 2021-08-18
compute/metadata: add retries


codyoss:
Add retries consistent to what is being done in other languages.

codyoss:
Can probably borrow the logic used in apiaries today: https://github.com/googleapis/google-api-go-client/blob/8258338b24dd72029fd1705ff1a5fa15d55ce17c/internal/gensupport/resumable.go#L233-L258. I think that should cover the different edge cases well enough.

codyoss:
@hostirosti This change is on HEAD. I will cut a release that includes this change on Monday, waiting for some other stuff to land first.

valyala:
FYI, the commit, which fixes the issue, led to significant binary size increase for apps which import `compute/metadata` package. See https://github.com/googleapis/google-cloud-go/issues/4783 .

PR #4648, opened 2021-08-19
feat(compute/metadata): retry error when talking to metadata service

codyoss:
This was reported internally to be causing issues. This package is
used by some of our auth flows so it should be good to make this
package more resilient to transient failures. Implementation inspired
by what we do for some of our http based services.

Because this package is currently not context aware I needed to add
attempts so retrying does not happen forever. Five attempts was
arbitrarily chosen.

Fixes: #4642

Release-As: 0.94.0

### DIFF
--- a/compute/metadata/metadata.go
+++ b/compute/metadata/metadata.go
@@ -284,6 +284,7 @@ func NewClient(c *http.Client) *Client {
 // getETag returns a value from the metadata service as well as the associated ETag.
 // This func is otherwise equivalent to Get.
 func (c *Client) getETag(suffix string) (value, etag string, err error) {
+	ctx := context.TODO()
 	// Using a fixed IP makes it very difficult to spoof the metadata service in
 	// a container, which is an important use-case for local testing of cloud
 	// deployments. To enable spoofing of the metadata service, the environment
@@ -315,7 +316,7 @@ func (c *Client) getETag(suffix string) (value, etag string, err error) {
 			break
 		}
 		if delay, shouldRetry := retryer.Retry(res.StatusCode, err); shouldRetry {
-			if err := gax.Sleep(context.Background(), delay); err != nil {
+			if err := gax.Sleep(ctx, delay); err != nil {
 				return "", "", err
 			}
 			continue
--- a/compute/metadata/metadata.go
+++ b/compute/metadata/metadata.go
@@ -32,6 +32,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/googleapis/gax-go/v2"
 )
 
 const (
@@ -304,9 +306,21 @@ func (c *Client) getETag(suffix string) (value, etag string, err error) {
 	}
 	req.Header.Set("Metadata-Flavor", "Google")
 	req.Header.Set("User-Agent", userAgent)
-	res, err := c.hc.Do(req)
-	if err != nil {
-		return "", "", err
+	var res *http.Response
+	retryer := newRetryer()
+	for {
+		var err error
+		res, err = c.hc.Do(req)
+		if err != nil {
+			if delay, shouldRetry :=retryer.Retry(res.StatusCode, err); shouldRetry {
+				if err := gax.Sleep(context.Background(), delay); err != nil {
+					return "", "", err
+				}
+				continue
+			}
+			return "", "", err
+		}
+		break
 	}
 	defer res.Body.Close()
 	if res.StatusCode == http.StatusNotFound {
--- a/compute/metadata/metadata.go
+++ b/compute/metadata/metadata.go
@@ -311,16 +311,16 @@ func (c *Client) getETag(suffix string) (value, etag string, err error) {
 	for {
 		var err error
 		res, err = c.hc.Do(req)
-		if err != nil {
-			if delay, shouldRetry :=retryer.Retry(res.StatusCode, err); shouldRetry {
-				if err := gax.Sleep(context.Background(), delay); err != nil {
-					return "", "", err
-				}
-				continue
-			}
-			return "", "", err
+		if err == nil {
+			break
 		}
-		break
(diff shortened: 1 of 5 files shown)


codyoss:
Temporarily adding don't merge while I deploy and test this code on the platform.

compute/metadata/retry.go
```diff
@@ -0,0 +1,73 @@
...
+	attempts int
+}
+
+func (r *metadataRetryer) Retry(status int, err error) (time.Duration, bool) {
+	retryOk := shouldRetry(status, err)
+	if !retryOk {
+		return 0, false
+	}
+	r.attempts++
+	if r.attempts == 6 {
```
noahdietz, line 47: Nit: can you put the magic number in a `const attemptLimit`  or something?

compute/metadata/metadata.go
```diff
@@ -304,9 +306,21 @@ func (c *Client) getETag(suffix string) (value, etag string, err error) {
...
 	req.Header.Set("User-Agent", userAgent)
-	res, err := c.hc.Do(req)
-	if err != nil {
-		return "", "", err
+	var res *http.Response
+	retryer := newRetryer()
+	for {
+		var err error
+		res, err = c.hc.Do(req)
+		if err != nil {
```
noahdietz, line 314: Nit: You could invert the `if` here. Matter of style preference I think though.
```go
res, err = c.hc.Do(req)
if err == nil {
  break
}

if delay, shouldRetry := ...
```

codyoss: I like it!

compute/metadata/retry.go
codyoss: done.

codyoss:
everything seems to be in working order, tested on the platform.

compute/metadata/metadata.go
```diff
@@ -304,8 +306,20 @@ func (c *Client) getETag(suffix string) (value, etag string, err error) {
...
+	var res *http.Response
+	retryer := newRetryer()
+	for {
+		var err error
+		res, err = c.hc.Do(req)
+		if err == nil {
+			break
+		}
+		if delay, shouldRetry := retryer.Retry(res.StatusCode, err); shouldRetry {
+			if err := gax.Sleep(context.Background(), delay); err != nil {
```
tbpg, line 318: This seems like we should declare a `ctx := context.TODO()` or something at the root of the function.

compute/metadata/retry_linux_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"testing"
+)
+
+func TestMetadataRetryerLinux(t *testing.T) {
+	retryer := metadataRetryer{bo: constantBackoff{}}
+
+	t.Run("retry on syscall.ECONNRESET", func(t *testing.T) {
+		_, shouldRetry := retryer.Retry(400, syscall.ECONNRESET)
+		if !shouldRetry {
+			t.Fatal("retryer.Retry(500, nil) = false, want true")
```
tbpg, line 30: 400

compute/metadata/retry_linux_test.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+	})
+	t.Run("retry on syscall.ECONNREFUSED", func(t *testing.T) {
+		_, shouldRetry := retryer.Retry(400, syscall.ECONNREFUSED)
+		if !shouldRetry {
+			t.Fatal("retryer.Retry(500, nil) = false, want true")
```
tbpg, line 36: 400

compute/metadata/retry_test.go
```diff
@@ -0,0 +1,108 @@
...
+		retryer := metadataRetryer{bo: constantBackoff{}}
+		delay, shouldRetry := retryer.Retry(500, nil)
+		if !shouldRetry {
+			t.Fatal("retryer.Retry(500, nil) = false, want true")
+		}
+		if delay != 100 {
+			t.Fatalf("retryer.Retry(500, nil) = %d, want 100", delay)
+		}
+	})
+	t.Run("don't retry 400", func(t *testing.T) {
```
tbpg, line 52: Seems like these could be a table test?

compute/metadata/metadata.go
```diff
@@ -304,8 +306,20 @@ func (c *Client) getETag(suffix string) (value, etag string, err error) {
 	}
 	req.Header.Set("Metadata-Flavor", "Google")
 	req.Header.Set("User-Agent", userAgent)
-	res, err := c.hc.Do(req)
-	if err != nil {
+	var res *http.Response
+	retryer := newRetryer()
```
tbpg, line 310: This appears to be race free because each call creates its own retryer. Are there any other concerns?

compute/metadata/metadata.go
tbpg: An alternative would be putting the timeout in here and removing the retry counting from the retryer. Not sure what is better.

noahdietz (approved):

tbpg (changes requested):

compute/metadata/metadata.go
codyoss: correct, with how our gax backoff works today I think you always need to return a new one per call so that the backoff can work accordingly.

compute/metadata/metadata.go
codyoss: TODO it is!

compute/metadata/retry_linux_test.go
codyoss: This is what happens when I copy paste too much, good catch!

compute/metadata/retry_linux_test.go
codyoss: ack

compute/metadata/retry_test.go
codyoss: Yeah, good idea. Done.

tbpg (approved):

PR #4655, opened 2021-08-20
fix(spanner/spannertest): Fix the "LIKE" clause handling for forward and backward matches

### LINKED ISSUE #4654: spannertest: Forward and backward matches using the "LIKE" clause include partially-matched records in the result
**Client**

spanner/spannertest

**Environment**

macOS BigSur11.4
(But I think it's not environment-dependent)

**Go Environment**

$ go version
1.17.0
(But I think it's not environment-dependent)

**Code**
Here is the example table and query that can reproduce the bug when applied to `spannertest`

- Table: Cities

| ID  |  Name  |
| ---- | ---- |
|  1  |  PARIS  |
|  2  |  ISTANBUL  |
|  3  |  TOKYO  |

```sql
SELECT Name FROM Cities WHERE Name LIKE ("IS%")
```

**Expected behavior**

The above query should return only this record.

```
ISTANBUL
```

**Actual behavior**

The actual result includes the partially matched record.

```
PARIS
ISTANBUL
```

That's because `spannertest` returns all values including "IS", not only in the prefixes.
And, same applies to the backward match.

**Cause**

I guess that's due to this func in `spannertest`.
It should add "^" for forwarding match and "$" for backward match to the regular expression before `regexp.MatchString`

```
func evalLike(str, pat string) bool {
	/*
		% matches any number of chars.
		_ matches a single char.
		TODO: handle escaping
	*/

	// Lean on regexp for simplicity.
	pat = regexp.QuoteMeta(pat)
	pat = strings.Replace(pat, "%", ".*", -1)
	pat = strings.Replace(pat, "_", ".", -1)
	match, err := regexp.MatchString(pat, str)
	if err != nil {
		panic(fmt.Sprintf("internal error: constructed bad regexp /%s/: %v", pat, err))
	}
	return match
}
```

sryoya:
Fixes #4654


olavloite (approved):
Nice find, LGTM

hengfengli (approved):

PR #4656, opened 2021-08-20
docs(storage): add comment in doc on how to use STORAGE_EMULATOR_HOST

### LINKED ISSUE #4576: storage: add docs or example for using STORAGE_EMULATOR_HOST
Supersedes: #1617

BrennaEpp:
Fixes #4576

### DIFF
--- a/storage/doc.go
+++ b/storage/doc.go
@@ -48,6 +48,13 @@ an unauthenticated client with
 
     client, err := storage.NewClient(ctx, option.WithoutAuthentication())
 
+To use an emulator with this library, you can set the STORAGE_EMULATOR_HOST
+environment variable to the address at which your emulator is running. This will
+send requests to that address instead of to Cloud Storage. You can then create a
+client as normal (see above) or, if you want, manually specify the endpoint:
+
+    client, err := storage.NewClient(ctx, option.WithEndpoint(emulatorHost + "/storage/v1/"))
+
 Buckets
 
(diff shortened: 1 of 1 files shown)


storage/doc.go
```diff
@@ -48,6 +48,13 @@ an unauthenticated client with
 
     client, err := storage.NewClient(ctx, option.WithoutAuthentication())
 
+To use an emulator with this library, you can set the STORAGE_EMULATOR_HOST
```
tritone, line 51: We should probably recommend that people use STORAGE_EMULATOR_HOST and give that as an example rather than the endpoint. I would give an example kind of like this:

```
// Set STORAGE_EMULATOR_HOST environment variable.
err := os.Setenv("STORAGE_EMULATOR_HOST", "localhost:9000")
if err != nil {
  // TODO: Handle error.
}

// Create client as usual.
if err != nil {
  // TODO: Handle error.
}

// This request is now directed to http://localhost:9000/storage/v1/b
// instead of https://storage.googleapis.com/storage/v1/b
if err := client.Bucket("my-bucket").Create(ctx, projectID, nil); err != nil {
    // TODO: Handle error.
}
```

storage/doc.go
```diff
@@ -48,6 +48,13 @@ an unauthenticated client with
...
+environment variable to the address at which your emulator is running. This will
```
tritone, line 52: Also note that there is no official emulator for Cloud Storage.

PR #4662, opened 2021-08-23
chore(storage): fix gRPC reader size/remain logic

tritone:
Correctly sets the size to the size of the whole object and
remain to the number of bytes to be read. Adds checks for
Size and Remain to the integration tests.

I used "chore" because this is still experimental, non-exported
code.


noahdietz (approved):
LGTM thank you for fixing this

PR #4670, opened 2021-08-24
feat(spanner/spansql): add support for STARTS_WITH function

### LINKED ISSUE #4661: spanner/spannertest: support STARTS_WITH
I suppose `spannertest` doesn't support the `STARTS_WITH` clause yet. 
It would be great if it's supported.
As [the official doc](https://cloud.google.com/spanner/docs/sql-best-practices#param-like) recommends using it rather than `LIKE` for a forward match, there should be certain motivation to use it.

I tried to implement it, but, due to the following characteristic of `STARTS_WITH`, it looks there is a certain point to consider in `spansql`.

- `STARTS_WITH` is a kind of comparison clause same as `LIKE` and `BETWEEN`
But,
- It takes both comparison targets as arguments unlike the other comparison clauses take one of them before the clause.

Due to the above, we cannot simply parse it as with the other comparison clauses even though what it does is the same as them.

I think there are two implementation options to handle that (I suppose the first one is better).

- Create another `Op` struct and the parsing func for `STARTS_WITH`.
- Put the parsing code for `STARTS_WITH` in the same func with the other comparison clauses,  and add code in the parent parsing funcs so that `STARTS_WITH` won't be parsed in the different code.

olavloite:
- Adds support for the STARTS_WITH function.
- Adds an example for how further functions can be implemented.

Fixes #4661

### DIFF
--- a/spanner/spannertest/db_eval.go
+++ b/spanner/spannertest/db_eval.go
@@ -28,6 +28,8 @@ import (
 
 	"cloud.google.com/go/civil"
 	"cloud.google.com/go/spanner/spansql"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // evalContext represents the context for evaluating an expression.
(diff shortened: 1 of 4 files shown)


hengfengli (approved):
LGTM.

PR #4700, opened 2021-08-30
fix(bigtable): detect version at runtime for user-agent (1.12+)

crwilcox:


bigtable/doc.go
```diff
@@ -118,7 +122,23 @@ const (
 
 // clientUserAgent identifies the version of this package.
 // It should be the same as https://pkg.go.dev/cloud.google.com/go/bigtable.
-const clientUserAgent = "cbt-go/v1.6.0"
+// const clientUserAgent = "cbt-go/v1.6.0"
+
+func getVersion() string {
```
codyoss, line 127: I think we need to guard this with a build constraint today. Having one impl pre 1.12 and one for after.

PR #4714, opened 2021-09-01
fix(compute/metadata): fix retry logic to not panic on error

codyoss:
In the case of an error res will be nil. Also, we should not break on nil err
as it might be eligible for a retry from the status code.

Fixes: #4713

### DIFF
--- a/compute/metadata/retry.go
+++ b/compute/metadata/retry.go
@@ -43,6 +43,9 @@ type metadataRetryer struct {
 }
 
 func (r *metadataRetryer) Retry(status int, err error) (time.Duration, bool) {
+	if status == 200 {
+		return 0, false
+	}
 	retryOk := shouldRetry(status, err)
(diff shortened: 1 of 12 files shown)


compute/metadata/retry.go
```diff
@@ -43,6 +43,9 @@ type metadataRetryer struct {
 }
 
 func (r *metadataRetryer) Retry(status int, err error) (time.Duration, bool) {
+	if status == 200 {
```
noahdietz, line 46: Nit: should we use `http.StatusOK`?

noahdietz (approved):

compute/metadata/retry.go
codyoss: Good idea, also added a test that calls the wrapping code for good measure.

tbpg (approved):

Issue #4727, opened 2021-09-05
civil: functions to check if Date, Time and DateTime is empty is not present


Qalifah:
At some point in my code, i needed to check if Date and Time is empty (all the fields in it have the zero value of their type) and i realized that there are no implementation for that, i would need to check all the fields in the Date and Time structs which i feel isn't the best option. I think there should be a function to handle that.

PR #4728, opened 2021-09-05
feat(civil):  add IsEmpty function to time, date and datetime

Qalifah:


Qalifah:
> @Qalifah Can you please raise an issue for this feature request for discussion. We generally like to talk things through before accepting PRs, thanks!

I did that https://github.com/googleapis/google-cloud-go/issues/4727

civil/civil.go
```diff
@@ -107,6 +107,11 @@ func (d1 Date) After(d2 Date) bool {
 	return d2.Before(d1)
 }
 
+// IsEmpty reports whether date fields are empty
+func (d Date) IsEmpty() bool {
```
codyoss, line 111: I think we should match the API of the time package and use [IsZero](https://pkg.go.dev/time#Time.IsZero).

civil/civil.go
```diff
@@ -107,6 +107,11 @@ func (d1 Date) After(d2 Date) bool {
 	return d2.Before(d1)
 }
 
+// IsEmpty reports whether date fields are empty
```
codyoss, line 110: nit: Please end sentence with a period, applies below as well.

codyoss (changes requested):
Just a few suggestions, thanks for the PR.

civil/civil.go
Qalifah: done

civil/civil.go
Qalifah: done

civil/civil.go
```diff
@@ -107,6 +107,11 @@ func (d1 Date) After(d2 Date) bool {
 	return d2.Before(d1)
 }
 
+// IsZero reports whether date fields are empty.
```
codyoss, line 110: nit: instead of saying empty for all of these could we say something like ".. are set to their default value"

codyoss (commented):
Just one more small thing.

civil/civil.go
Qalifah: okay. Done

codyoss (approved):
LGTM, thank you for your contribution!

PR #4730, opened 2021-09-08
chore(spanner): modify benchwrapper's proto to use Singers/Albums

zoercai:


spanner/go.mod
```diff
@@ -12,5 +12,6 @@ require (
 	google.golang.org/api v0.54.0
 	google.golang.org/genproto v0.0.0-20210825212027-de86158e7fda
 	google.golang.org/grpc v1.40.0
+	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0 // indirect
```
hengfengli, line 15: I don't think we should add this dependency. We probably should remove the changes in go.mod and go.sum.

spanner/internal/benchwrapper/main.go
```diff
@@ -80,8 +81,9 @@ func (s *server) Read(ctx context.Context, req *pb.ReadQuery) (*pb.EmptyResponse
 
 func (s *server) Insert(ctx context.Context, req *pb.InsertQuery) (*pb.EmptyResponse, error) {
 	var muts []*spanner.Mutation
-	for _, i := range req.Users {
-		muts = append(muts, spanner.Insert("sometable", []string{"name", "age"}, []interface{}{i.Name, i.Age}))
+	fmt.Print(req.Singers)
```
hengfengli, line 84: is this for debugging?

spanner/internal/benchwrapper/proto/spanner.proto
```diff
@@ -15,31 +15,42 @@
...
-message UpdateQuery{
+message UpdateQuery {
   // The queries to use in the update call.
-  repeated string Queries = 1;
+  repeated string queries = 1;
 }
 
-message EmptyResponse{
-}
+// TODO(deklerk): Replace with Google's canonical Empty.
```
hengfengli, line 50: why do we add this TODO as deklerk is not working on this any more?

spanner/internal/benchwrapper/main.go
zoercai: ah yep thanks for spotting - removed!

spanner/go.mod
zoercai: I've removed both

spanner/internal/benchwrapper/proto/spanner.proto
zoercai: Oh I copied this proto from the one in google3, which had this todo, I've removed it now.

hengfengli (approved):
LGTM.

zoercai:
@codyoss do you know why the kokoro CI build is failing here? There doesn't seem to be anything in the logs that show what's causing the failure.

codyoss:
@zoercai It looks like you need to run `goimports -w .` in the directory of the updated proto. Linting error

PR #4733, opened 2021-09-09
fix(pubsub): prevent draining error return for Receive

### LINKED ISSUE #4718: pubsub: receive returns an error when canceled
**Client**

Pub/Sub

**Expected behavior**

Cancelling `Receive` should never return any errors.

**Actual behavior**

Sometimes a `draining` error is returned when cancelling a `Receive` as we try to process an incoming message.

**Additional Notes**
This is mostly caused by the receive scheduler. We should either ignore this error (and allow the message to be redelivered) or the better thing to do would be to somehow make sure it never happens.

hongalex:
In [`receive_scheduler.go`](https://github.com/hongalex/google-cloud-go/blob/d50ab295185c8e85688796acbf0c9d587a1bbc33/pubsub/internal/scheduler/receive_scheduler.go), adding a new message after the context passed into `Receive` is cancelled will result in a library-defined `draining` error. This is poor behavior as the condition which this happens is out of the user's control (simultaneous `Shutdown` + `ReceiveScheduler.Add()`). This PR mitigates this by 

1. Stop message pulling if the context is cancelled
2. Nack messages that have have already been received (don't extend their lease)

In addition, since it's unlikely for messages to be added, this PR also removes the draining error entirely and allows messages to be added. This is very unlikely and is still preferable to surfacing an unactionable error to the user.

Fixes #4718


pubsub/subscription.go
```diff
@@ -925,13 +925,29 @@ func (s *Subscription) Receive(ctx context.Context, f func(context.Context, *Mes
 						}
 					}
 				}
+				// If the context is done, don't pull more messages.
+				select {
+				case <-ctx.Done():
+					return nil
```
codyoss, line 931: Should we be returning the ctx.Err() in these cases?

hongalex: I think in general we don't want `Receive` to return any errors, including context cancellation errors.

codyoss: Ack

codyoss (approved):

PR #4742, opened 2021-09-10
feat(spanner/spannertest): Support generated columns

psytale:


spanner/spannertest/README.md
```diff
@@ -22,7 +22,7 @@ by ascending esotericism:
 - more aggregation functions
 - SELECT HAVING
 - more literal types
-- generated columns
+- generated columns referencing other generated columns
```
olavloite, line 25:
```suggestion
- generated columns referencing other generated columns or columns containing NULL values
```

It seems that this implementation does not support generating a value based on other columns that contain a NULL value. Is there a specific reason why that cannot be supported by this implementation? If so, we should add a short notice here, but my impression is that there's nothing special here that makes it impossible to generate a value if one of the dependent columns contains a null value.

spanner/spannertest/db.go
```diff
@@ -414,6 +424,45 @@ func (d *database) writeValues(tx *transaction, tbl spansql.ID, cols []spansql.I
...
+			if col.Generated.expr != nil {
+				allNotNull := true
+				// We should only generate if all dependent columns are not null
+				for _, dependency := range col.Generated.dependentCols {
+					if row[t.colIndex[dependency]] == nil {
+						allNotNull = false
+						break
+					}
+				}
+				// We skip this column if any of its dependencies are null
```
olavloite, line 454: Why do we need to do that?

spanner/spannertest/db.go
```diff
@@ -414,6 +424,45 @@ func (d *database) writeValues(tx *transaction, tbl spansql.ID, cols []spansql.I
...
+		if !found {
+			return status.Error(codes.Internal, "row failed to be inserted")
+		}
+		row := t.rows[rowNum]
+		ec := evalContext{
+			cols: t.cols,
+			row:  row,
+		}
+
+		// TODO: We would need to do a topoligical sort on dependencies to ensure we can
```
olavloite, line 442: nit:

```suggestion
		// TODO: We would need to do a topological sort on dependencies to ensure we can
```

spanner/spannertest/db.go
```diff
@@ -634,6 +683,10 @@ func (t *table) addColumn(cd spansql.ColumnDef, newTable bool) *status.Status {
 			// TODO: what happens in this case?
 			return status.Newf(codes.Unimplemented, "can't add NOT NULL columns to non-empty tables yet")
 		}
+		if cd.Generated != nil {
```
olavloite, line 686: nit: This check should be extended to check whether the table is empty or not. Adding a generated column to an empty tables should be safe in this case.

This PR should also add a check to `dropColumn` to prevent a column that is a dependent of a generated column from being dropped.

spanner/spannertest/db.go
```diff
@@ -1090,3 +1147,40 @@ func parseAsDate(s string) (civil.Date, error) { return civil.ParseDate(s) }
 func parseAsTimestamp(s string) (time.Time, error) {
 	return time.Parse("2006-01-02T15:04:05.999999999Z", s)
 }
+
+// getIDsForGeneratedExpression returns a list of column names the expression depends on
+func getIDsForGeneratedExpression(e spansql.Expr) []spansql.ID {
```
olavloite, line 1152: I'm not sure this is the best way to do this from a maintenance and completeness perspective. Instead I think it might be better to edit the `evalContext.evalExpr` function to also return the id's that the expression needs to be evaluated. That would keep the logic for what is needed for a certain expression in one place, and reduce the chance of missing something, and also ensure that whenever we add support for new expressions, those are also automatically included in this array.
This implementation currently does not detect any dependencies on other columns if they are used in:
1. Function calls
2. Array expressions

spanner/spannertest/db_test.go
```diff
@@ -361,6 +362,83 @@ func TestConcurrentReadInsert(t *testing.T) {
...
+	) PRIMARY KEY (Id);`
+	var db database
+
+	ddl, err := spansql.ParseDDL("filename", sql)
+	if err != nil {
+		t.Fatalf("%s: Bad DDL", err)
+	}
+	for _, stmt := range ddl.List {
+		if st := db.ApplyDDL(stmt); st.Code() != codes.OK {
+			t.Fatal()
```
olavloite, line 380: nit: Add error message

spanner/spannertest/db_test.go
```diff
@@ -361,6 +362,83 @@ func TestConcurrentReadInsert(t *testing.T) {
...
+		[]*structpb.ListValue{
+			listV(stringV("3")),
+		}, kr, 0)
+	if err != nil {
+		t.Fatalf("failed to read: %v", err)
+	}
+	rows := slurp(t, iter)
+	// 2 * A * B = 2 * 3 * 3
+	if rows[0][1].(int64) != 18 {
+		t.Fatal("Generated value for C should have been 18")
```
olavloite, line 413: nit: use got/want format, so something like `Generated value for C mismatch\n Got: %v\nWant: %v`

spanner/spannertest/db_test.go
```diff
@@ -361,6 +362,83 @@ func TestConcurrentReadInsert(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to read: %v", err)
+	}
+	rows = slurp(t, iter)
+	// Should be nil as only A was set not B
+	if rows[0][1] != nil {
+		t.Fatal("Generated value for C should be nil")
+	}
+
+	t.Fail()
```
olavloite, line 439: I don't think this belongs here.

spanner/spannertest/db_test.go
```diff
@@ -361,6 +362,83 @@ func TestConcurrentReadInsert(t *testing.T) {
...
+		}, kr, 0)
+	if err != nil {
+		t.Fatalf("failed to read: %v", err)
+	}
+	rows = slurp(t, iter)
+	// Should be nil as only A was set not B
+	if rows[0][1] != nil {
+		t.Fatal("Generated value for C should be nil")
+	}
+
```
olavloite, line 438: Could we also add tests for update and delete?

spanner/spannertest/funcs.go
```diff
@@ -52,6 +52,14 @@ var functions = map[string]function{
 			return strings.HasPrefix(s, prefix), spansql.Type{Base: spansql.Bool}, nil
 		},
 	},
+	"LOWER": {
```
olavloite, line 55: I don't think this change belongs in this PR.

spanner/spannertest/db_test.go
```diff
@@ -361,6 +362,83 @@ func TestConcurrentReadInsert(t *testing.T) {
 	}
 }
 
+func TestGeneratedColumn(t *testing.T) {
```
olavloite, line 365: Could we also add tests for:
1. Adding a generated column to an existing table.
2. Altering a generated column.
3. Dropping a column that is a dependant of a generated column.

olavloite (changes requested):
Thanks for this change, I think it is a good step in the right direction, but there are a couple of things that need some looking into:
1. I get two test failures when I run the tests locally. One seems to simply be a leftover `t.Fail()` call in one of your tests. The other is in an existing test.
2. I think it would be better to try to integrate the check which columns a generated column depends on with the existing expression evaluation. This reduces the chances of having differences between the two.

spanner/spannertest/README.md
psytale: I struggled to find decent documentation but I was playing with live spanner it appeared to not generate any value if the column it depended on was null.

spanner/spannertest/db.go
psytale: Good point, thanks

spanner/spannertest/db.go
psytale: Yeah I really didn't like this approach, it felt like it was prone to error and future failure but couldn't figure out something smarter. I think your suggestion sounds solid I can check into that, So adding a 3rd return argument to evalExpr, potentially returning and error and the list of dependencies?

spanner/spannertest/funcs.go
psytale: Can move it to a separate one, just smuggled it in here as I was testing with it

psytale:
> Thanks for this change, I think it is a good step in the right direction, but there are a couple of things that need some looking into:
> 
> 1. I get two test failures when I run the tests locally. One seems to simply be a leftover `t.Fail()` call in one of your tests. The other is in an existing test.
> 2. I think it would be better to try to integrate the check which columns a generated column depends on with the existing expression evaluation. This reduces the chances of having differences between the two.

Thanks for the quick review, I think that suggestion is great, I was hoping someone could offer a better suggestion!

spanner/spannertest/README.md
psytale: My problem was mostly how do I determine when we should calculate a value for a generated column, I guess you are correct that sometimes basing it on a column containing null is in the column, I just didn't want to hide errors, i.e. when trying to multiply a null by 2 I'll get a null and be unable to store it, I wanted to differentiate that from if something else went wrong but it felt non trivial. Perhaps that is better, if the evaluating the expression fails I just store nothing.

psytale: In that case I don't even care about which columns I'm dependent on as I'm making no promise to resolve generated columns in a sensible order to enable them to depend on each other.

olavloite: The evaluation should normally not fail, but `NULL` should be stored in the generated column. Most expressions that include at least one `NULL` value will evaluate to null, and as long as the generated column is not marked as `NOT NULL` it should not be a problem. Consider the following example:

```
create table Singers (SingerId INT64, FirstName STRING(100), LastName STRING(100), FullName STRING(200) AS (FirstName || ' ' || LastName) STORED) PRIMARY KEY (SingerId);

insert into Singers (SingerId, FirstName, LastName) VALUES (1, 'First', 'Last');
insert into Singers (SingerId, FirstName, LastName) VALUES (2, 'First', NULL);
```

The `Singers` table will now contain two rows, and the `FullName` value for row 2 will be `NULL`, as string concatenation produces `NULL` if one of the concatenated values is null.

spanner/spannertest/db.go
olavloite: Yes, something like that was my thought as well.

spanner/spannertest/README.md
psytale: So in that case I could just run the expression and as long as the type matches we store it, dump all this logic looking for dependent columns, what do you think?

olavloite: Yes, I think that would be better. If the evaluation of the expression returns an error instead of a value, then of course the error should be returned, but otherwise if there is no error then we should just accept the outcome of the evaluation and store it.

spanner/spannertest/db.go
psytale: This is already checking the table is non empty, as for checking if its a dependency of a generated column do you still think its worth adding that as we talked about dropping all dependency calculations above?

olavloite: Sorry, I missed that it was already checking it.

As for the dependency checking:
1. I don't think you need any dependency checking here.
2. We should preferably use the dependency checking in `dropColumn` to prevent columns that are used by a generated column from being dropped. But preferably that dependency checking can be done using the method that also does the evaluation, so that we don't have to duplicate that code. Would it for example be possible to use default values for the evaluation context when we only want to check the dependencies?

spanner/spannertest/funcs.go
olavloite: Ah, ok. I would prefer to have it in a separate PR, unless it is actually used in one of the test cases (I don't think it is at this moment, right?)

spanner/spannertest/db.go
psytale: I think that definitely sounds like the right approach, having a way to run the evaluator in with default values in some "complete" way (i.e. we'd have to check in an OR what is required to evaluate the RHS even if the LHS evaluates to true). It feels out of scope for me right now, given it will be a fair bit of extra work to get going well and I'm not overly confident with the evaluation code and typing. For now I'll just come up with the solution for adding columns as any other features should be able to  be built on top without changes so doesn't feel wasteful

psytale:
@olavloite I did a rewrite dropping all the column dependency stuff, pretty happy with it now, I don't see why that can't be added on top later when we want to support dropping columns that may be dependencies or resolving co-dependent generated columns in a reliable way.  It felt like a tricky problem to solve though, needing default values and ensuring complete execution etc, a good rewrite of the db_eval stuff.

spanner/spannertest/db.go
```diff
@@ -414,6 +419,40 @@ func (d *database) writeValues(tx *transaction, tbl spansql.ID, cols []spansql.I
...
+			row:  row,
+		}
+
+		// TODO: We would need to do a topological sort on dependencies (i.e. what other columns the expression references)
+		// to ensure we can handle generated columns which reference other generated columns
+		for i, col := range t.cols {
+			if col.Generated != nil {
+				res, err := ec.evalExpr(col.Generated)
+				if err != nil {
+					// We assume that if the expression ended up with a type error a
```
olavloite, line 443: I looked into the evaluation of the different expressions, and the problem that you are running into is that those evaluations do not take into account that some of the input parameters could be NULL. So I think that this is a reasonable workaround for this for now, but we should probably try to fix the evaluation of expressions that may return null values and then remove this.

olavloite:
@googlebot I consent.

olavloite (approved):
LGTM

spanner/spannertest/db.go
psytale: Yeah exactly, its very much a best effort but dumb approach. Cheers

olavloite:
@psytale The build error is caused by the fact that the implementation uses `[Errors.Is](https://pkg.go.dev/errors#Is)`, which is only supported from Go 1.13 and further. The Spanner client library supports Go 1.11 and onwards, so we need to change that bit as well. There are two files in the repository (errors112.go and errors113.go) that are conditionally added to the build depending on the Go version. You could add the method that you need there.

Feel free to let me know if you want me to look into it for you.

psytale:
> @psytale The build error is caused by the fact that the implementation uses `[Errors.Is](https://pkg.go.dev/errors#Is)`, which is only supported from Go 1.13 and further. The Spanner client library supports Go 1.11 and onwards, so we need to change that bit as well. There are two files in the repository (errors112.go and errors113.go) that are conditionally added to the build depending on the Go version. You could add the method that you need there.
> 
> Feel free to let me know if you want me to look into it for you.

Think I got something going with it but feel free to push any changes if you want to do it in a different way, I'm not overly confident with go. I also fixed (and added a test case) for functions and null columns, I wasn't happy with removing the codes.InvalidArgument but felt it was okay, shout if you think of a better way to do it (or feel free to push commits)

olavloite:
@psytale I decided to remove the extra error wrapping and checking, and instead add NULL checks to the most used evaluations. That should ensure that generated columns that evaluate to NULL will (in most cases) work as intended. The cases where they don't are not really related to the generated columns implementation, but to the fact that those evaluations would fail anyways before this change. So if you were to for example do a `SELECT a*b FROM Foo` where some values of a or b are null, the select would fail.

@hengfengli Would you mind taking a quick look at this as well?

spanner/spannertest/db.go
```diff
@@ -414,6 +418,32 @@ func (d *database) writeValues(tx *transaction, tbl spansql.ID, cols []spansql.I
 		if err := f(t, colIndexes, r); err != nil {
 			return err
 		}
+
+		// Get row again after potential update merge to ensure we compute
+		// generated columns with fresh data.
+		pk := r[:t.pkCols]
+		rowNum, found := t.rowForPK(pk)
+		// this should never fail as the row was just inserted
```
hengfengli, line 426: `this` -> `This`

hengfengli: Also, the missing period at the end `... inserted.`

spanner/spannertest/db.go
```diff
@@ -414,6 +418,32 @@ func (d *database) writeValues(tx *transaction, tbl spansql.ID, cols []spansql.I
...
+		if !found {
+			return status.Error(codes.Internal, "row failed to be inserted")
+		}
+		row := t.rows[rowNum]
+		ec := evalContext{
+			cols: t.cols,
+			row:  row,
+		}
+
+		// TODO: We would need to do a topological sort on dependencies (i.e. what other columns the expression references)
```
hengfengli, line 436: I think we have a convention to limit the line length of comments to 80 characters.

spanner/spannertest/db.go
```diff
@@ -643,6 +677,9 @@ func (t *table) addColumn(cd spansql.ColumnDef, newTable bool) *status.Status {
 		Name:    cd.Name,
 		Type:    cd.Type,
 		NotNull: cd.NotNull,
+		// TODO: We should figure out what columns the Generator expression relies on and check validate it at this time
```
hengfengli, line 680: Please well format the comment.

spanner/spannertest/db_test.go
hengfengli: Please add the error message.

spanner/spannertest/integration_test.go
```diff
@@ -1262,6 +1262,106 @@ func TestIntegration_ReadsAndQueries(t *testing.T) {
...
+
+	ri = client.Single().Query(ctx, spanner.NewStatement(
+		`SELECT CanonicalName, TotalSales FROM `+tableName+` ORDER BY Name`,
+	))
+	all, err = slurpRows(t, ri)
+	if err != nil {
+		t.Errorf("Read rows failed: %v", err)
+	}
+
+	// Great writer has nil because NumSongs is nil
```
hengfengli, line 1354: The comment should be `Poor writer`, right?

hengfengli:
@olavloite I saw some of your comments in the tests are not resolved yet. Do you think they should be fixed before merging?

olavloite:
> @olavloite I saw some of your comments in the tests are not resolved yet. Do you think they should be fixed before merging?

I've added the missing error message and missing delete test.

My comment on tests for dropping and altering columns is void, as we have removed the check altogether and added it to the list of limitations in the README.

So as far as I'm concerned, this is now good to be merged once CI is green.

hengfengli (approved):

Issue #4745, opened 2021-09-11
bigquery: No way to get job statistics from Query


jameshartig:
**Is your feature request related to a problem? Please describe.**
I'd like to use `Query.Read` to query and read the rows which can use the fast path queries API however the returned iterator does not give me any statistics on the query.

**Describe the solution you'd like**
Can the job's Statistics be put on the iterator so we could get them for metrics/billing purposes without having to resort to using the jobs.insert API?

**Describe alternatives you've considered**
Alternatively a new method could be added to only do the fast path and return an error if its not supported but this seems more complicated than just adding another field/method onto the iterator.

shollyman:
What stats are you interested in?  A specific subset, or all?

jameshartig:
> What stats are you interested in? A specific subset, or all?

Ideally the whole `JobStatistics` but for now just `TotalBytesProcessed`,  `TotalBytesBilled`, `CacheHit`.

shollyman:
Would it be sufficient to expose the RowIterator source I wonder?  With the job reference you could then get the job metadata at your leisure.

jameshartig:
In the case of the slow path the source would be a `Job` and in the fast path the source would be the `QueryResponse`? That sounds good to me!

shollyman:
Yeah, both paths retain a reference to the job identifier/reference.  However, a RowIterator can be instantiated from a table directly, or by a query.  So we either have a job reference or a table reference.

shollyman:
Take a look at https://github.com/googleapis/google-cloud-go/pull/4748 and see if that would address your needs.

jameshartig:
Yep! That should work perfectly. Thanks for such a quick turnaround! I wasn't originally aware that the `queries` BQ endpoint actually still generated a job that you could get the status of but I see now that it returns a `JobReference`. Great

jameshartig:
@shollyman thank you so much for the quick resolution! 🙏

shollyman:
Working on that this week actually.  bigquery is its own submodule and released independently.  Keep an eye on https://github.com/googleapis/google-cloud-go/pull/4738 for updates.

shollyman:
https://github.com/googleapis/google-cloud-go/releases/tag/bigquery/v1.23.0

PR #4748, opened 2021-09-13
feat: expose the query source of a rowiterator via SourceJob()

shollyman:
RowIterator can currently come from calling Read() on a Table,
Read() on a Job, and Read() directly from a Query.  The third invocation
is used for fast path queries and doesn't return a query identifier
suitable for gathering statistics.

This PR exposes a SourceJob() method on RowIterator to address this
issue.  Users still get the benefits of optimized query execution, but
can get the reference to a Job for looking up additional statistics
should they need it.

Fixes: https://github.com/googleapis/google-cloud-go/issues/4745

### DIFF
--- a/bigquery/integration_test.go
+++ b/bigquery/integration_test.go
@@ -2197,6 +2197,29 @@ func TestIntegration_LegacyQuery(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


codyoss (approved):
LGTM

PR #4754, opened 2021-09-15
feat(bigquery): add session and connection support

shollyman:
Adds:
* CreateSession to Query config
* ConnectionProperties to specify connection keys/values in Query config
* SessionInfo in JobStatistics to record session stats

Also contains some minor test refactoring to make it easier to test query execution.


bigquery/query.go
```diff
@@ -132,6 +132,12 @@ type QueryConfig struct {
 	// Allows the schema of the destination table to be updated as a side effect of
 	// the query job.
 	SchemaUpdateOptions []string
+
+	// CreateSession will trigger creation of a new session when true.
+	CreateSession bool
+
+	// ConnectionProperties are optional key-values settings.
+	ConnectionProperties []*ConnectionProperty
```
codyoss, line 140: Question: Should this just be a map[string]string?

shollyman: The issue is the lack of enforcement of key uniqueness.  If keys were required to be unique, the map representation would make sense.  But if you want a 1:many representation for a key, you can just send several properties with the same key and difference values.

codyoss: `map[string][]string`? Maybe that is worse, I am not sure.

shollyman: I _suspect_ map to a slice of strings is worse given most things are 1:1 and so the dominant case you're dealing with extra logic for the first element of the slice.  Is it worse than dealing with a slice of k-v pairs?  That's the part that's unclear to me.

shollyman: Reviewing prior art, there's not a great set of examples either way; most are using another level of indirection.  JDBC/ODBC support a connection string, which string with delimiters, and key-values are parsed from the single string.  Programatically, the wrapper is typed so each individual property can have a custom getter-setter.

It's likely we'll see the same pattern emerge, so I'm less concerned about this one, and think the raw slice of k-v pairs is sufficient.

stephaniewang526 (approved):

PR #4756, opened 2021-09-15
fix(internal/gapicgen): add a temporary import

codyoss:
The new compute beta makes use of xerrors which is causing regen
to fail since we don't depend on this yet. This change will be
blown away by regen, which is intended.


noahdietz (approved):
Thanks! Sorry should've given you a heads up or done this myself in the original PR.

PR #4764, opened 2021-09-16
docs(storage): replace misleading project ID

briandealwis:
The `client.Buckets()` example uses `my-bucket` as the project ID, which seems more like a bucket ID.


tritone (dismissed):
Thanks for the fix!

tritone (approved):

PR #4769, opened 2021-09-16
test(compute): fix a flaky test

georgiyekkert:
attempt to fix https://github.com/googleapis/google-cloud-go/issues/4741


compute/apiv1/smoke_test.go
```diff
@@ -570,7 +571,16 @@ func TestCapitalLetter(t *testing.T) {
 				Firewall: name,
 			})
 		if err != nil {
-			t.Error(err)
+			// sometimes the resource is not "ready" for deletion, sleep and retry.
+			time.Sleep(60 * time.Second)
```
codyoss, line 575: Instead of a single hardcoded sleep I think we should make use of internal.Retry:

https://github.com/googleapis/google-cloud-go/blob/master/internal/retry.go#L30-L32

codyoss (changes requested):

compute/apiv1/smoke_test.go
georgiyekkert: done

codyoss:
Looks like formatting errors, please run `gomimports -w .` to fix.

codyoss (approved):

PR #4771, opened 2021-09-17
docs(datastore): clarify documentation of omitempty

benjaminjkraft:
Previously the documentation said that omitempty omits "zero values"
(linking to the spec), and then says structs aren't (but I didn't find
that part clear the first 5 times I read it).  I found this pretty
unclear, especially since it's burying the fact that it differs from
`encoding/json` in its special handling of `time.Time`.  Hopefully being
more explicit is clearer.


crwilcox:
Thanks @benjaminjkraft for this. I think it does improve the docs.

crwilcox (approved):

PR #4777, opened 2021-09-20
feat(pubsublite): notify subscriber clients on partition reassignment

tmdiep:
* Adds `ReceiveSettings.ReassignmentHandler` for the `SubscriberClient` to receive notifications when the server sends a new partition reassignment to the client. The handler receives previous and new partition numbers.
* Discards NACKs for messages whose partitions have been assigned away. This prevents unnecessary termination of the `SubscriberClient` if the default nack handler is used.


pubsublite/pscompat/settings.go
```diff
@@ -164,6 +164,28 @@ type NackHandler func(*pubsub.Message) error
 // error and terminate.
 type ReceiveMessageTransformerFunc func(*pb.SequencedMessage, *pubsub.Message) error
 
+// ReassignmentHandlerFunc is called any time a new partition assignment is
+// received from the server. It will be called with both the previous and new
+// partition numbers as decided by the server.
+//
+// When this handler is called, partitions that are being assigned away are
```
manuelmenzella-google, line 171: Do we offer any stronger guarantees? For example: can we say something like "when this is invoked, no more messages from the removed partitions will be delivered to the callback?" Can the message callback be invoked concurrently to this callback?

pubsublite/internal/wire/subscriber.go
```diff
@@ -492,17 +510,22 @@ func (as *assigningSubscriber) handleAssignment(partitions partitionSet) error {
 	return nil
 }
 
-func (as *assigningSubscriber) doHandleAssignment(partitions partitionSet) ([]*singlePartitionSubscriber, error) {
+func (as *assigningSubscriber) doHandleAssignment(nextPartitions PartitionSet) (PartitionSet, []*singlePartitionSubscriber, error) {
```
manuelmenzella-google, line 513: Nit; can you add a comment explaining that this returns the older partition set and the removed subscribers?

pubsublite/internal/wire/subscriber.go
```diff
@@ -537,6 +560,15 @@ func (as *assigningSubscriber) Terminate() {
 	}
 }
 
+// PartitionActive returns whether the partition is still active.
```
manuelmenzella-google, line 563: Have you considered returning a slice of active partitions instead? It's almost as simple for clients, and provides a superset of functionality.

pubsublite/internal/wire/subscriber_test.go
```diff
@@ -929,6 +935,15 @@ func TestSinglePartitionSubscriberStopDuringAdminSeek(t *testing.T) {
 	}
 }
 
+func verifyPartitionsActive(t *testing.T, sub Subscriber, want bool, partitions ...int) {
+	t.Helper()
+	for _, p := range partitions {
+		if got := sub.PartitionActive(p); got != want {
```
manuelmenzella-google, line 941: If the Subscriber can return the entire set of partitions, verifying expectations is simpler; i.e. you do not need to check for assigned & unassigned partitions. You can just do got == wanted.

pubsublite/pscompat/subscriber.go
```diff
@@ -58,6 +59,11 @@ func (ah *pslAckHandler) OnNack() {
 		return
 	}
 
+	// Ignore nacks for partitions that have been assigned away.
```
manuelmenzella-google, line 62: Is this behavior replicated in the other clients? I think it's reasonable, but we should try to be consistent.

pubsublite/pscompat/settings.go
tmdiep: I added "message deliveries may still be in flight" to the comments. For Go, we can't offer strong guarantees. The Java implementation does, but Python doesn't.

I spent some time trying to offer the strong guarantee for Go, but it requires placing a mutex around the invocation of the message callback [here](https://github.com/googleapis/google-cloud-go/blob/320f4970ccd396d6af69486862730cff1b4b3e28/pubsublite/internal/wire/subscriber.go#L120), as well as reading/writing a "shutdown" state. But the downside is that a message receiver blocking for too long can delay acking the new assignment, which I think is what the [Java implementation](https://github.com/googleapis/java-pubsublite/blob/00b6863fc2521919048f0d582fa963e98a13c915/google-cloud-pubsublite/src/main/java/com/google/cloud/pubsublite/cloudpubsub/internal/AssigningSubscriber.java#L92) has.

pubsublite/internal/wire/subscriber.go
tmdiep: Done

pubsublite/pscompat/subscriber.go
tmdiep: Yes, both Java and Python do this. It was easier on those client libraries, due to their structure, but required exposing `PartitionActive()` from the wire layer here.

pubsublite/internal/wire/subscriber.go
tmdiep: Yes, I almost did that, but it didn't work well with the existing data structures of `assigningSubscriber` and `multiPartitionSubscriber`, and so thought it wasn't worth implementing the superset of functionality that isn't required yet. Nothing outside this library can access anything in `internal/`, so we can easily change the interface later.

pubsublite/internal/wire/subscriber_test.go
tmdiep: Ack; as above.

The previous validation essentially did as you suggested, but I changed it to call the new `PartitionActive()` function.

manuelmenzella-google:
Thanks for addressing the comments!

manuelmenzella-google (approved):

tmdiep:
Thanks for reviewing :)

PR #4781, opened 2021-09-21
feat(bigquery): don't generate random insertId

erikdubbelboer:
Random insertIds are basically the same as not setting the insertId, except that not setting it allows for a lot more inserts per second.

The code for the random insertIds was still something from before empty insertIds was supported.

### DIFF
--- a/bigquery/inserter.go
+++ b/bigquery/inserter.go
@@ -26,7 +26,7 @@ import (
 
 // NoDedupeID indicates a streaming insert row wants to opt out of best-effort
 // deduplication.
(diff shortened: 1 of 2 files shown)


codyoss:
@erikdubbelboer Thanks for the PR! Would you mind creating an associated issue for discussion, thanks!

loferris (approved):
LGTM

erikdubbelboer:
This was decided against here so closing it: https://github.com/googleapis/google-cloud-go/issues/4787

PR #4794, opened 2021-09-22
fix(firestore): preconditions should have microsecond precision for timestamps

### LINKED ISSUE #4229: firestore: `LastUpdateTime` is `time.Time` must be microsecond precision
See: https://github.com/googleapis/google-cloud-go/blob/d979830c5d1423eb5befd33d03ba7adf3ec4c217/firestore/options.go#L59

The [documentation](https://firebase.google.com/docs/firestore/reference/rest/v1/Precondition) does not fully explain how this works and it took me time to grok it.

> string (Timestamp&nbsp;format)
>
> When set, the target document must exist and have been last updated at that time.
>
> A timestamp in RFC3339 UTC "Zulu" format, with nanosecond resolution and up to nine fractional digits. 
>
> Examples: "2014-10-02T15:01:23Z" and "2014-10-02T15:01:23.045123456Z".

Golang [time.Time](https://golang.org/pkg/time/#Time) is nanosecond precision and so we should be good. Even though `LastUpdateTime` takes a `time.Time`, the nanosecond resolution (last 3 digits) must be zeros for the Golang function to work.

This seems to be an implementation detail for the client library (and its documentation) rather than something that needs to be learned by the developer.

```Golang
now := time.Now()
log.Printf("[Now] %s (%d)\n", now, now.UnixNano())

preconds := firestore.LastUpdateTime(now)
_, err := ref.Delete(ctx, preconds)
if err == nil {
	t.Errorf("Expected failure")
}

errStatus, _ := status.FromError(err)
log.Printf("[Now] %s (%s)", errStatus.Code(), errStatus.Message())
```
Yields:
```
[Now] 2021-06-07 13:54:23.851441895 -0700 PDT m=+0.662380388 (1623099263851441895)
[Now] InvalidArgument (timestamp cannot have more than microseconds precision)
```

The error is incorrect because, Golang `time.Time` is always nanosecond precision. I realized that what it needs is a `time.Time` that's rounded to microsecond (1000 nanoseconds) units. And so, the code needs to be something of the form:

```Golang
roundNow := time.Unix(0, (now.UnixNano()/int64(time.Microsecond))*int64(time.Microsecond))
log.Printf("[RoundNow] %s (%d)\n", roundNow, roundNow.UnixNano())

preconds := firestore.LastUpdateTime(roundNow)
_, err := ref.Delete(ctx, preconds)
if err == nil {
	t.Errorf("Expected failure")
}

errStatus, _ := status.FromError(err)
log.Printf("[RoundNow] %s (%s)", errStatus.Code(), errStatus.Message())
```
Which fails correctly because we're not giving it a time that matches its last updated:
```
[RoundNow] 2021-06-07 13:54:23.851441 -0700 PDT (1623099263851441000)
[RoundNow] InvalidArgument (Invalid base version, it is greater than the stored version: projects/my-proj/databases/(default)/documents/tests/valid)
```

The error is likely uncommon because, the developer is likely to `Get`'ting the snapshot and then usings its `Updated` property to set the `LastUpdateTime(snap.Update)` but still...

crwilcox:
Fix #4229 by ensuring times are round to microsecond precision. Related, this documentation may be incorrect. https://firebase.google.com/docs/firestore/reference/rest/v1/Precondition

### DIFF
--- a/firestore/options.go
+++ b/firestore/options.go
@@ -54,9 +54,13 @@ func (e exists) String() string {
 }
 
 // LastUpdateTime returns a Precondition that checks that a resource must exist and
-// must have last been updated at the given time. If the check fails, the write
-// does not occur.
-func LastUpdateTime(t time.Time) Precondition { return lastUpdateTime(t) }
+// must have last been updated at the given time, rounded to microsecond precision.
+// If the check fails, the write does not occur.
+func LastUpdateTime(t time.Time) Precondition {
+	// UpdateTime preconditions have microsecond precision.
+	t = t.Round(time.Microsecond)
+	return lastUpdateTime(t)
+}
 
 type lastUpdateTime time.Time
 


firestore/options.go
```diff
@@ -54,9 +54,13 @@ func (e exists) String() string {
 }
 
 // LastUpdateTime returns a Precondition that checks that a resource must exist and
-// must have last been updated at the given time. If the check fails, the write
-// does not occur.
-func LastUpdateTime(t time.Time) Precondition { return lastUpdateTime(t) }
+// must have last been updated at the given time, rounded to microsecond precision.
+// If the check fails, the write does not occur.
```
tritone, line 58: I'm a little confused how this bug got noticed. The example [here](https://pkg.go.dev/cloud.google.com/go/firestore#hdr-Preconditions) shows pulling the `UpdateTime` from a document snapshot and then using it immediately in a precondition. If this workflow is followed, I don't think `UpdateTime` would ever have more than microsecond precision? Is there some other way you would calculate an `UpdateTime`?

In any case, this seems fine; it might be worth checking for additional incongruities between `time.Time` and firestore time.

tritone (approved):

firestore/options.go
crwilcox: In order to see this, you have to do some slightly strange things. Any time returned on a document for update already is microsecond aligned. I also couldn't make it happen with time.Now().

However, if I create a time from nanoseconds and ensure it isn't aligned, that will result in the failure.

PR #4802, opened 2021-09-23
feat(storage): returned wrapped error for timeouts

### LINKED ISSUE #4197: storage: Uninformative error messages after errors cause timeout
**Client**

Storage

**Environment**

Local Ubuntu 18.04 machine

**Go Environment**
```
$ go version
go version go1.15.5 linux/amd64
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/roger.curley/.cache/go-build"
GOENV="/home/roger.curley/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/roger.curley/go/pkg/mod"
GOOS="linux"
GOPATH="/home/roger.curley/go"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build136821508=/tmp/go-build -gno-record-gcc-switches"
```
(Some lines deleted that reference info internal to my company. If needed I can reproduce this on a non-work machine, but I don't expect my local environment to be very relevant here)
**Code**

e.g.
```go
package main

import (
	"context"
	"fmt"
	"net/http"
	"time"

	"cloud.google.com/go/storage"
)

func main() {
	err := mainWrapped()
	if err != nil {
		fmt.Printf("%+v\n", err)
	}
}
func mainWrapped() error {
	ctx := context.Background()
	httpClient :=
		http.Client{
			Transport: transport{},
		}
	client, err := storage.NewClient(ctx, option.WithHTTPClient(&httpClient))
	if err != nil {
		return err
	}
	bucket := client.Bucket("foo")
	obj := bucket.Object("bar")
	ctx, _ = context.WithTimeout(ctx, time.Second)
	_, err = obj.Attrs(ctx)
	if err != nil {
		return err
	}
	return nil
}

type transport struct{}

func (t transport) RoundTrip(req *http.Request) (*http.Response, error) {
	fmt.Println("Returning 500 error")
	return &http.Response{
		Status:     "500 Internal Server Error",
		StatusCode: 500,
	}, nil
}
```

**Expected behavior**

It should be possible to recover the underlying error that led to the timeout.

**Actual behavior**

The only error information returned is that the context timed out:
```
Returning 500 error
Returning 500 error
context deadline exceeded
```
Or using `spew.Dump` instead of `fmt.Printf`:
```
Returning 500 error
Returning 500 error
(context.deadlineExceededError) {
}
```

**Additional context**

I care about the specific error because some retryable errors can be mitigated on my end: for example, this issue masked an issue caused by attempting to delete the same object many times at once.

I considered filing this as a feature request, but it looks like there's some attempt to do this in the code:
https://github.com/googleapis/google-cloud-go/blob/f97bdfaa3ea3431eb6b5cb73a7fc0ed1b3730a1d/internal/retry.go#L42-L45

tritone:
This piggybacks on #4797 to allow storage to expose wrapped
service errors when a call retries without success until timeout
or cancellation.

I also updated all checks for context sentinel errors in storage
to use xerrors.Is instead of strict equality. Users of this
package should do the same. I'll update the documentation on
errors from this package in a subsequent PR.

We will have to bump the dependency on the root module before
merging this change I believe.

Fixes #4197


storage/integration_test.go
```diff
@@ -3023,7 +3025,8 @@ func TestIntegration_PublicBucket(t *testing.T) {
 	}
 
 	errCode := func(err error) int {
-		err2, ok := err.(*googleapi.Error)
+		var err2 *googleapi.Error
+		ok := xerrors.As(err, &err2)
 		if !ok {
 			return -1
 		}
```
BrennaEpp, line 3032: nit: Can this be inlined as follows?
```suggestion
		var err2 *googleapi.Error
		if ok := xerrors.As(err, &err2); !ok {
			return -1
		}
```

Also I don't love the variable name `err2` but I realize that was there already.

BrennaEpp (approved):
Very excited about this work!

storage/integration_test.go
tritone: Good idea, fixed.

noahdietz (approved):

Issue #4804, opened 2021-09-24
Storage: Bucket.go doesn't export projectNumber available in the REST API 


anrajitha:
**Is your feature request related to a problem? Please describe.**
I intend to verify the project where bucket exists, in my 'CheckIfBucketExists' functionality. 
For me to do this check, it would be great if bucket.go would export the projectNumber field available in the REST API: 
https://cloud.google.com/storage/docs/json_api/v1/buckets/get

**Describe the solution you'd like**
bucket.attrs would give me the projectNumber associated with the project as well. 

**Describe alternatives you've considered**
Other alternatives would for checking the existence of a bucket in a project is : to list all the buckets in the project check whether the bucket is one of them.  (I dont intend to do this as it adds latency).

**Additional context**
https://cloud.google.com/storage/docs/json_api/v1/buckets#resource-representations --> projectNumber

PR #4805, opened 2021-09-24
feat(storage): add projectNumber field to bucketAttrs.

anrajitha:
Fixes: https://github.com/googleapis/google-cloud-go/issues/4804

Adds ProjectNumber field to bucketAttrs.


storage/bucket.go
```diff
@@ -337,6 +337,9 @@ type BucketAttrs struct {
 	// Typical values are "multi-region", "region" and "dual-region".
 	// This field is read-only.
 	LocationType string
+
+	// The project number of the project the bucket belongs to.
```
tritone, line 341: Note that the field is read-only (as in L339 above).

storage/bucket_test.go
```diff
@@ -642,10 +642,11 @@ func TestNewBucket(t *testing.T) {
...
 		},
-		LocationType: "dual-region",
-		Encryption:   &raw.BucketEncryption{DefaultKmsKeyName: "key"},
-		Logging:      &raw.BucketLogging{LogBucket: "lb", LogObjectPrefix: "p"},
-		Website:      &raw.BucketWebsite{MainPageSuffix: "mps", NotFoundPage: "404"},
+		LocationType:  "dual-region",
+		Encryption:    &raw.BucketEncryption{DefaultKmsKeyName: "key"},
+		Logging:       &raw.BucketLogging{LogBucket: "lb", LogObjectPrefix: "p"},
+		Website:       &raw.BucketWebsite{MainPageSuffix: "mps", NotFoundPage: "404"},
+		ProjectNumber: 123231313,
```
tritone, line 649: Let's also add a check to the integration test for buckets to ensure that this is received. https://github.com/googleapis/google-cloud-go/blob/81fda19b595a4fab7d2ff322d8524d2c7aced16e/storage/integration_test.go#L255

tritone (commented):
Generally looks good!

storage/bucket.go
anrajitha: Done. Added the comment.

storage/bucket_test.go
anrajitha: Done., added a check on projectNumber

storage/integration_test.go
```diff
@@ -265,6 +265,9 @@ func TestIntegration_BucketMethods(t *testing.T) {
 	if attrs.LocationType == "" {
 		t.Error("got an empty LocationType")
 	}
+	if attrs.ProjectNumber == 0 {
+		t.Errorf("got a zero ProjectNumber")
+	}
```
BrennaEpp, line 270: Can we add this check for creating/deleting with attributes in this same test as well?

anrajitha: Done. Added this check in there as well.

BrennaEpp (approved):
LGTM

PR #4808, opened 2021-09-24
feat(spanner/spansql): fill in missing hash functions

dfinkel:
The `spansql` package is quite useful for parsing and managing schemas,
however, with the addition of `CHECK` constraints to the spanner DDL,
the full set of SQL functions can now appear in schemas.

Fill in the missing Hash functions to support their use.

List of hash functions pulled from: https://cloud.google.com/spanner/docs/hash_functions


olavloite (approved):
LGTM.

Note that this will only add these functions to the list of functions that are recognized as being a function. They will still not return any values, so they cannot be used in queries.

dfinkel:
Thanks @olavloite!

Yeah, when I glanced at which functions `spannertest` had implemented, I only saw one string-function (and it had to do some type-resolution that would have required a bit more work, so I didn't bother trying to implement any of these functions (looks like another's been added since then).

For my use-case, I want to include `SHA256` in a `CHECK` constraint's expression. Adding the function to the list of recognized functions unblocks me adding that constraint, even if it's mostly ignored by `spannertest`. (I'll run most of my tests against the C++ emulator (and a real Cloud Spanner instance) anyway, because of other deficiencies in `spannertest`)

hengfengli (approved):

dfinkel:
Thanks @hengfengli !

hengfengli:
@dfinkel No worries. Thank you for your contribution!

PR #4814, opened 2021-09-24
docs(bigquery): use xerror.As with googleapi.Error

tritone:
Fix up type assertions with googleapi.Error to use xerrors.As.
In bigquery, this was only required in integration tests and in
the package docs.

Follows from #4797 and similar to some of the fixes in storage
in #4802 (storage required additional changes though).


codyoss:
LGTM

bigquery/doc.go
```diff
@@ -303,9 +303,10 @@ Errors
 
 Errors returned by this client are often of the type googleapi.Error: https://godoc.org/google.golang.org/api/googleapi#Error
 
-These errors can be introspected for more information by type asserting to the richer *googleapi.Error type. For example:
+These errors can be introspected for more information by using `errors.As` with the richer *googleapi.Error type. For example:
```
shollyman, line 306: s/error.As/xerrors.As/ here and below

shollyman (approved):

bigquery/doc.go
tritone: Ah, I wrote this because this is sample user code and for Go 1.13+ they should really use `errors` instead of `xerrrors`... however I guess xerrors works for both, so I can change it.

PR #4896, opened 2021-09-27
chore(storage): replace Canceled status with context.Canceled

### LINKED ISSUE #4772: storage: TestIntegration_CancelWriteGRPC failed
This test failed!

To configure my behavior, see [the Flaky Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/main/packages/flakybot).

I will stop commenting.

---

commit: 0e7ccf1cc95c84838cc3de722ff4cc2e5098f579
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/56a826f4-e7cf-4998-91a7-a219532c664d), [Sponge](http://sponge2/56a826f4-e7cf-4998-91a7-a219532c664d)
status: failed
<details><summary>Test output</summary><br><pre>integration_test.go:1138: got <nil>, wanted context.Canceled</pre></details>

noahdietz:
The gRPC Storage client replaces `Canceled` statuses with `context.Canceled` to conform to what JSON does.

Unskip the `CancelWriteGRPC` test by fixing the `Canceled` status handling and updating the test to provoke network calls using after cancelation, resulting in `Write` errors. See notes in #4772 WRT the context cancelation race condition behavior. 

Fixes #4772


storage/writer.go
```diff
@@ -637,3 +642,11 @@ func read(r io.Reader, buf []byte) (int, bool, error) {
 	}
 	return recvd, done, err
 }
+
+func checkCanceled(err error) error {
+	if status.Code(err) == codes.Canceled {
```
tritone, line 647: This seems like the right thing to do, but just to check: are we ever losing any information here by obscuring the difference between a user-side context cancelation and the RPC call returning `codes.Canceled`?

tritone (approved):

storage/writer.go
noahdietz: We _might_ be, but I'm not entirely sure. I don't think we would be since this error is induced by client-side request settings i.e. the user sets a deadline on the context, and the server notices that the deadline it received has expired. Nothing for the server to add there imo.

One other option would be to change how `gax.APIError` handles `codes.Canceled` and instead of storing the original grpc-go error, it stores a `context.Canceled` after parsing the `Status`. But that's a bit more far reaching and I'd be hesitant to do that.

PR #4899, opened 2021-09-28
feat(internal/godocfx): detect preview versions

tbpg:
See b/197007831.

I added unit tests. But, I also confirmed this works on `google.golang.org/appengine/v2` (`status: preview` shows up in the output YAML).


codyoss (approved):
LGTM

PR #4909, opened 2021-09-29
fix(storage): update PAP to use inherited instead of unspecified

BrennaEpp:


storage/bucket.go
```diff
@@ -376,23 +376,29 @@ const (
 	// not set in a call to GCS.
 	PublicAccessPreventionUnknown PublicAccessPrevention = iota
 
-	// PublicAccessPreventionUnspecified corresponds to a value of "unspecified"
-	// and is the default for buckets.
+	// PublicAccessPreventionUnspecified corresponds to a value of "inherited".
```
shaffeeullah, line 379: this should still be unspecified i think

shaffeeullah (commented):
Let's wait for review from @tritone  as well, but looks good to me. Thanks, Brenna!

storage/bucket.go
tritone: +1, keep `unspecified` here.

storage/bucket_test.go
```diff
@@ -257,11 +257,22 @@ func TestBucketAttrsToRawBucket(t *testing.T) {
 	}
 
 	// Test that setting PublicAccessPrevention to "unspecified" leads to the
-	// setting being propagated in the proto.
+	// undeprecrated setting being propagated in the proto.
```
tritone, line 260: I would say "inherited" rather than "undeprecated"

storage/bucket.go
```diff
@@ -376,23 +376,29 @@ const (
...
 
 	// PublicAccessPreventionEnforced corresponds to a value of "enforced". This
 	// enforces Public Access Prevention on the bucket.
 	PublicAccessPreventionEnforced
 
-	publicAccessPreventionUnknown     string = ""
-	publicAccessPreventionUnspecified        = "unspecified"
-	publicAccessPreventionEnforced           = "enforced"
+	// PublicAccessPreventionInherited corresponds to a value of "inherited"
+	// and is the default for buckets.
```
tritone, line 388: @shaffeeullah maybe we should include a sentence about how this means that the value for PAP is inherited from the project, folder, or org? Do we have something like this in other languages?

tritone (commented):
Generally seems good, a couple small comments

storage/bucket.go
shaffeeullah: i do not believe we have something like that in other languages

PR #4928, opened 2021-09-30
chore(storage): use xerrors for Canceled check

noahdietz:


codyoss (commented):
LGTM, but I'll let the storage folk sign off for real.

noahdietz:
> LGTM, but I'll let the storage folk sign off for real.

Thanks!

PR #4929, opened 2021-09-30
chore(storage): refactor read conditions

noahdietz:
Refactors setting read object conditions to use the generic `applyCondsProto` helper instead of the type specific function.


codyoss (commented):
LGTM, but I will let the storage team sign off.

tritone (approved):
Nice, thank you!

PR #4932, opened 2021-09-30
chore(storage): reuse grpc buffer for leftovers

noahdietz:
Simplifies management of leftover bytes from a read message by just reusing the sub-slice of unread content as the `leftovers` instead of maintaining a copy of it. gRPC will allocate the memory to deserialize a message no matter what, so we should just keep referencing that block until we've read it all. Doing some simple profiling shows that this reduces memory allocations even in simple cases because the Reader now longer allocates a new block for leftovers.


codyoss (approved):
Nice find, I like it!

PR #4933, opened 2021-09-30
test(compute): add headers test

georgiyekkert:
This test is a requirement by vchudnov

### DIFF
--- a/compute/apiv1/smoke_test.go
+++ b/compute/apiv1/smoke_test.go
@@ -600,9 +600,6 @@ func TestCapitalLetter(t *testing.T) {
 }
 
 func TestHeaders(t *testing.T) {
-	if testing.Short() {
-		t.Skip("skipping integration test in short mode")
-	}
 	ctx := context.Background()
 
 	svr := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


compute/apiv1/smoke_test.go
```diff
@@ -593,3 +598,35 @@ func TestCapitalLetter(t *testing.T) {
 		t.Fatalf("got(-),want(+):\n%s", diff)
 	}
 }
+
+func TestHeaders(t *testing.T) {
+	if testing.Short() {
```
codyoss, line 603: nit: you can remove the skip. This test does not hit any backend resources so it should be fine to run all the time.

codyoss (changes requested):
Just one minor thing

compute/apiv1/smoke_test.go
noahdietz: I think the client factory method will try to load ADC & fail if not present, so if this test is using an insecure endpoint we may want to also pass `WithoutAuthentication` [option](https://pkg.go.dev/google.golang.org/api/option#WithoutAuthentication).

georgiyekkert: Done

codyoss (approved):
Thanks!

Issue #4663, opened 2021-08-23
bigquery: INTERVAL not a recognized type


Thomasdezeeuw:
**Client**

BigQuery

**Environment**

N/A

**Go Environment**

$ go version
go version go1.16.6 darwin/arm64
$ go env
```go
GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/Users/thomas/Library/Caches/go-build"
GOENV="/Users/thomas/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/thomas/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/thomas/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/opt/homebrew/Cellar/go/1.16.6/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/opt/homebrew/Cellar/go/1.16.6/libexec/pkg/tool/darwin_arm64"
GOVCS=""
GOVERSION="go1.16.6"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/thomas/src/slight/sql-runner/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/lv/qbz3lh7j6gdc43dd1p22bcc40000gn/T/go-build1478395030=/tmp/go-build -gno-record-gcc-switches -fno-common"

```

**Code**

e.g.
```go
package main

func main() {
        client, _ := bigquery.NewClient(ctx, "my-project-ID")
	q := client.Query("SELECT interval_field FROM some_table_with_interval_field")
        rows, _ := q.Run(..)
        var nullable_values []bigquery.Value
        err := rows.Next(&nullable_values)
        // The `error` above will be set.
}
```

**Expected behavior**

[...]

PR #4943, opened 2021-10-04
feat(bigquery): add minimal INTERVAL support

shollyman:
This PR simply adds the interval type to the allowed schema values.  It
does not include any type conversion support between go<-->bigquery
types, nor does it include features like query parameters.

Towards: https://github.com/googleapis/google-cloud-go/issues/4663

### DIFF
--- a/bigquery/schema.go
+++ b/bigquery/schema.go
@@ -226,6 +226,10 @@ const (
 	// BigNumericFieldType is a numeric field type that supports values of larger precision
 	// and scale than the NumericFieldType.
 	BigNumericFieldType FieldType = "BIGNUMERIC"
+	// IntervalFieldType is a representation of a duration or amount of time.
+	//
+	// This type is in preview.  It is EXPERIMENTAL and subject to change or removal without notice.
+	IntervalFieldType FieldType = "INTERVAL"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


codyoss (approved):
LGTM

bigquery/schema.go
```diff
@@ -226,6 +226,10 @@ const (
 	// BigNumericFieldType is a numeric field type that supports values of larger precision
 	// and scale than the NumericFieldType.
 	BigNumericFieldType FieldType = "BIGNUMERIC"
+	// IntervalFieldType is a representation of a duration or amount of time.
+	//
+	// This type is in preview.  It is EXPERIMENTAL and subject to change or removal without notice.
```
tswast, line 231: Is this sufficient to give you enough wiggle room to return something other than string in the future?

tswast (approved):

bigquery/schema.go
shollyman: I believe it is; users read data using custom marshaling into their own data structures or for the dynamic case use a type switch.  When we have a new go type for handling interval values we can augment the docs to indicate such.

shollyman:
Closing this in favor of 5907

Issue #4665, opened 2021-08-24
pubsub: native OpenTelemetry support


ianonavy:
cc @aabmass

**Is your feature request related to a problem? Please describe.**
As developer using PubSub via the Golang client lib, I would like to propagate OpenTelemetry span contexts via messages so that I can better understand the entire lifecycle of a message.

**Describe the solution you'd like**
I would expect a working solution to look similar to the [NodeJS implementation](https://github.com/googleapis/nodejs-pubsub/pull/1078).

Maybe something like this?

```golang
client, _ := pubsub.NewClientWithConfig(ctx, PROJECT_ID, pubsub.ClientConfig{
    OpenTelemetryEnabled: true,
})
```

Then I would expect the result to look something like this when configured with a Cloud Trace exporter:

![Screenshot of trace waterfall showing spans across publisher and consumer](https://miro.medium.com/max/688/1*QDbHoSvE7xrtd-nnWErsew.png)

Image is from [this article](https://medium.com/google-cloud/integrating-opentelemetry-into-cloud-pub-sub-19aacd83692a) by the former intern who implemented the NodeJS and Python versions of this request.

**Describe alternatives you've considered**
[OpenTelemetry-OpenCensus Bridge](https://pkg.go.dev/go.opentelemetry.io/otel/bridge/opencensus) would be great if PubSub were already instrumented to propagate OpenCensus spans through messages like this, but unfortunately the behavior I describe is orthogonal to the [existing issue](https://github.com/googleapis/google-cloud-go/issues/4237) to port the existing tracing from OpenCensus to OpenTelemetry.

hongalex:
Hiya,

Thanks for filing this issue. OT tracing support is definitely something that's on our roadmap and we're actively planning to support this. At the moment, it's mostly blocked on standardization across Pub/Sub client library languages and [waiting on the Go implementation for 1.0](https://github.com/orgs/open-telemetry/projects/15). I'll post here for updates.

Also appreciate the detail info. If you have specific asks or features you want, I'd be happy to relay them to the rest of the client library maintainers while we work on this.

hongalex:
Hey! Yeah we (myself and the Node library maintainer) are in contact with @aabmass. Yeah, that's one of the few things that we're trying to standardize though we're not 100% sure how we're going to fix that for Node since it's already released. We're discussing the merits of making a breaking change/major version bump or just supporting both attributes.

[...]

PR #5034, opened 2021-10-26
feat(pubsub): add open telemetry trace support

hongalex:
This PR adds the foundation for tracing of a publisher and subscriber for Pub/Sub. This PR attempts to mimic existing messaging system tracing and follows the semantic conventions [defined here](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/trace/semantic_conventions/messaging.md). This is a draft and is subject to backwards incompatible changes.

[Here's a sample](https://gist.github.com/hongalex/929383213155a64b30435b1bc42171dd) that demonstrates using this otel enabled library w/ Google Cloud Tracing as the exporter. The output looks something like the following:
image


pubsub/topic.go
```diff
@@ -522,19 +528,32 @@ type PublishResult = ipubsub.PublishResult
...
+	}
+
 	// Calculate the size of the encoded proto message by accounting
 	// for the length of an individual PubSubMessage and Data/Attributes field.
 	msgSize := proto.Size(&pb.PubsubMessage{
 		Data:        msg.Data,
 		Attributes:  msg.Attributes,
 		OrderingKey: msg.OrderingKey,
 	})
+	span.SetAttributes(semconv.MessagingMessagePayloadSizeBytesKey.Int(msgSize))
```
quartzmo, line 556: Is this already set in `getSpanAttributes` on line 532, above?

quartzmo: I think I understand: After injecting the span into the message attributes, the size of the message needs to be updated.

pubsub/trace.go
```diff
@@ -236,3 +242,61 @@ func withSubscriptionKey(ctx context.Context, subName string) context.Context {
...
+		OrderingKey: msg.OrderingKey,
+	})
+	ss := []trace.SpanStartOption{
+		trace.WithAttributes(
+			semconv.MessagingSystemKey.String("pubsub"),
+			semconv.MessagingDestinationKey.String(topic),
+			semconv.MessagingDestinationKindTopic,
+			semconv.MessagingMessageIDKey.String(msg.ID),
+			semconv.MessagingMessagePayloadSizeBytesKey.Int(msgSize),
+			attribute.String("pubsub.ordering_key", msg.OrderingKey),
```
quartzmo, line 296: Should this be `"messaging.pubsub.ordering_key"`, per the design doc?

pubsub/topic.go
```diff
@@ -697,6 +728,7 @@ func (t *Topic) publishMessageBundle(ctx context.Context, bms []*bundledMessage)
 			ipubsub.SetPublishResult(bm.res, "", err)
 		} else {
 			ipubsub.SetPublishResult(bm.res, res.MessageIds[i], nil)
+			bm.span.SetAttributes(semconv.MessagingMessageIDKey.String(res.MessageIds[i]))
```
quartzmo, line 731: Does the "publish RPC" span also need to be updated with the message ID?

mbn18:
Hey, is there any expectation to when this will be supported? thanks

mbn18:
@hongalex , thanks for letting me know.

For the mean while I took your propagator design and implemented it in my code. Will add it here so other can use it till a final solution exist
```
package whatever

import (
	"cloud.google.com/go/pubsub"
	"context"
	"go.opentelemetry.io/otel"
)

const pubsubAttribPrefix = "whatever"

// PubsubMessageCarrier injects and extracts traces from a pubsub.Message.
type PubsubMessageCarrier struct {
	msg *pubsub.Message
}

// NewPubsubMessageCarrier creates a new PubsubMessageCarrier.
func NewPubsubMessageCarrier(msg *pubsub.Message) PubsubMessageCarrier {
	return PubsubMessageCarrier{msg: msg}
}

// Get retrieves a single value for a given key.
func (c PubsubMessageCarrier) Get(key string) string {
	return c.msg.Attributes[pubsubAttribPrefix+"_"+key]
}

// Set sets an attribute.
func (c PubsubMessageCarrier) Set(key, val string) {
	c.msg.Attributes[pubsubAttribPrefix+"_"+key] = val
}

// Keys returns a slice of all keys in the carrier.
func (c PubsubMessageCarrier) Keys() []string {
	i := 0
	out := make([]string, len(c.msg.Attributes))
	for k := range c.msg.Attributes {
		out[i] = k
		i++
	}
	return out
}

func PubSubMessageInjectContext(ctx context.Context, msg *pubsub.Message) {
	otel.GetTextMapPropagator().Inject(ctx, NewPubsubMessageCarrier(msg))
}

func PubSubMessageExtractContext(ctx context.Context, msg *pubsub.Message) context.Context {
	return otel.GetTextMapPropagator().Extract(ctx, NewPubsubMessageCarrier(msg))
}
```

Mistic92:
Hi, what is the current status?

panperla:
Are there any alternative thread? Or it just died and nobody is looking into it?

meredithslota:
We are still tracking this here: https://github.com/googleapis/google-cloud-go/issues/4665

PR #5038, opened 2021-10-28
feat(spanner/spansql): add support for CAST functions

yokoyama10:
The changes make the spansql library possible to parse type conversion functions (`CAST` and `SAFE_CAST`).

https://cloud.google.com/spanner/docs/conversion_functions

- added `CastFunc` struct  for handling these functions.
- added a bool argument `noParam` to `parseType` function, because parameterized typename (e.g. `STRING(MAX)`) is not supported in cast function.

### DIFF
--- a/spanner/spansql/parser.go
+++ b/spanner/spansql/parser.go
@@ -1538,7 +1538,7 @@ func (p *parser) parseColumnDef() (ColumnDef, *parseError) {
 
 	cd := ColumnDef{Name: name, Position: p.Pos()}
 
-	cd.Type, err = p.parseType()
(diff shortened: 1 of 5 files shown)


olavloite:
@yokoyama10 I've added an [alternative here](https://github.com/googleapis/google-cloud-go/pull/5057) that implements the CAST and SAFE_CAST functions as actual functions instead of a special kind of literal. Would you mind taking a look at that if that would also satisfy your requirement?

yokoyama10:
Thank you for your work #5057 and it is very applicable to our requirements!

Issue #4228, opened 2021-06-07
bigquery: allow construction of arbitrary job references


shollyman:
BQ veneer doesn't currently allow construction of job references with foreign projectID.  We have `JobFromID` and `JobFromIDLocation`, so we should consider adding the ability to fully qualify the job.

This also warrants more testing given reliance on embedded projectID in retained client references.

PR #5048, opened 2021-10-29
feat(bigquery): allow construction of jobs from other projects

shollyman:
Adds a `JobfromProject()` method to allow users to get metadata from a
job that was created within another project.

Fixes: https://github.com/googleapis/google-cloud-go/issues/4228

### DIFF
--- a/bigquery/job.go
+++ b/bigquery/job.go
@@ -46,17 +46,24 @@ type Job struct {
 // For jobs whose location is other than "US" or "EU", set Client.Location or use
 // JobFromIDLocation.
 func (c *Client) JobFromID(ctx context.Context, id string) (*Job, error) {
-	return c.JobFromIDLocation(ctx, id, c.Location)
+	return c.JobFromProject(ctx, c.projectID, id, c.Location)
 }
 
 // JobFromIDLocation creates a Job which refers to an existing BigQuery job. The job
 // need not have been created by this package (for example, it may have
 // been created in the BigQuery console), but it must exist in the specified location.
 func (c *Client) JobFromIDLocation(ctx context.Context, id, location string) (j *Job, err error) {
-	ctx = trace.StartSpan(ctx, "cloud.google.com/go/bigquery.JobFromIDLocation")
+	return c.JobFromProject(ctx, c.projectID, id, location)
+}
+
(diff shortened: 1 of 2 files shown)


bigquery/job.go
```diff
@@ -46,17 +46,24 @@ type Job struct {
...
 // been created in the BigQuery console), but it must exist in the specified location.
 func (c *Client) JobFromIDLocation(ctx context.Context, id, location string) (j *Job, err error) {
-	ctx = trace.StartSpan(ctx, "cloud.google.com/go/bigquery.JobFromIDLocation")
+	return c.JobFromProject(ctx, c.projectID, id, location)
+}
+
+// JobFromProject creates a Job which refers to an existing BigQuery job.  The job
+// need not have been created by this package, nor does it need to reside within the same
+// project or location as the instantiated client.
+func (c *Client) JobFromProject(ctx context.Context, projectID, jobID, location string) (j *Job, err error) {
```
codyoss, line 62: This is fine to me. Another option would be to have a config struct here and let users selectively override particular parameters than what they have set with their client.

codyoss (approved):

bigquery/job.go
shollyman: This pattern is used mostly because its in line with how we expose similar out-of-project resources like datasets.  There's a minor preposition inconsistency though, in jobs we do JobFrom<X>, but datasets are DatasetInProject.  Seemed better to be consistent with the other job constructor names.

tswast (commented):
Design LGTM. Are there other places that use c.projectID that need to be updated? Might be good to see a unit test or two that verifies when the job project != client project sets the expect HTTP path.

shollyman:
> Design LGTM. Are there other places that use c.projectID that need to be updated? Might be good to see a unit test or two that verifies when the job project != client project sets the expect HTTP path.

Other callsites that reference the client's projectID directly is job insertion, either the standard jobs.insert or the optimized jobs.query path.  There's other things like dataset iterator Datasets(), but under the hood it calls DatasetsInProject() which allows remote enumeration.

shollyman:
Getting this to a unit test is surprisingly annoying given the lack of intermediate interfaces, so I wrote a small integration test for now.

tswast (approved):

PR #5059, opened 2021-11-01
fix(bigquery): address issue from job construction feature

shollyman:
As part of the changes for
https://github.com/googleapis/google-cloud-go/pull/5048 one callsite of
getJobInternal was missed.  Normally this would easily get identified
due to the change in signature, but getJobInternal has a set of expected
string arguments, followed by variadic string args.  This got picked up
by integration testing, but I failed to recall that presubmit doesn't run
integration tests so it was caught after submit.

Mostly this one's a cautionary tale for having a mix of mandatory and
variadic functions that share the same type.

Fixes: https://github.com/googleapis/google-cloud-go/issues/5058


tswast (approved):
Good catch.

codyoss (approved):

PR #5060, opened 2021-11-01
test(storage): use current googler email in integration test

### LINKED ISSUE #5051: storage: TestIntegration_ACL failed
This test failed!

To configure my behavior, see [the Flaky Bot documentation](https://github.com/googleapis/repo-automation-bots/tree/main/packages/flakybot).

I will stop commenting.

---

commit: d15b1e55bb2812d65ad397b53a39d771a5743122
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/afa6d53f-aebd-4b14-855b-046a7a7b09d7), [Sponge](http://sponge2/afa6d53f-aebd-4b14-855b-046a7a7b09d7)
status: failed
<details><summary>Test output</summary><br><pre>    integration_test.go:2238: Error while deleting bucket ACL rule: googleapi: Error 400: Unknown user email addresss: jbd@google.com, invalid</pre></details>

crwilcox:
Fixes #5051

### DIFF
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -2213,8 +2213,8 @@ func TestIntegration_ACL(t *testing.T) {
 		t.Errorf("default ACL: could not delete entity %s", entity)
 	}
 
-	entity2 := ACLEntity("user-jbd@google.com")
-	rule2 := ACLRule{Entity: entity2, Role: RoleReader, Email: "jbd@google.com"}
+	entity2 := ACLEntity("user-crwilcox@google.com")
+	rule2 := ACLRule{Entity: entity2, Role: RoleReader, Email: "crwilcox@google.com"}
 	if err := bkt.ACL().Set(ctx, entity2, RoleReader); err != nil {
 		t.Errorf("Error while putting bucket ACL rule: %v", err)
 	}
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -2213,8 +2213,8 @@ func TestIntegration_ACL(t *testing.T) {
 		t.Errorf("default ACL: could not delete entity %s", entity)
 	}
 
-	entity2 := ACLEntity("user-crwilcox@google.com")
-	rule2 := ACLRule{Entity: entity2, Role: RoleReader, Email: "crwilcox@google.com"}
+	entity2 := AllAuthenticatedUsers
+	rule2 := ACLRule{Entity: entity2, Role: RoleReader}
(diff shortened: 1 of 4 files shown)


codyoss (approved):

storage/integration_test.go
```diff
@@ -2213,8 +2213,8 @@ func TestIntegration_ACL(t *testing.T) {
 		t.Errorf("default ACL: could not delete entity %s", entity)
 	}
 
-	entity2 := ACLEntity("user-jbd@google.com")
-	rule2 := ACLRule{Entity: entity2, Role: RoleReader, Email: "jbd@google.com"}
+	entity2 := ACLEntity("user-crwilcox@google.com")
```
tritone, line 2216: Is there another more stable test identity/email that is used for these tests elsewhere? Not to imply that you won't be around indefinitely 😄 

Also wondering if it's worth having a test with a specific email-- we might be able to just use another predefined role.

tritone: I see https://github.com/googleapis/google-cloud-go/blob/master/storage/integration_test.go#L452 , let's do the same here?

crwilcox: Hey now, I am gonna work at google and live *forever*. Obviously 😅 


https://github.com/googleapis/google-cloud-go/blob/master/storage/acl.go#L41 does define a few options, perhaps we can use a group instead?

frankyn: Could we use `project-owners-TEST_PROJECT_NUMBER` ==> `project-owners-123456789` instead? 

https://cloud.google.com/storage/docs/access-control/lists#scopes

crwilcox: I set it to allAuthenticatedUsers, since that should be more-or-less permanent. are you implying though we could use an arbitrary product, or that we could use a known project?

frankyn: I would prefer specific to the test project which you should be able to get the project number: using `allAuthenticatedUsers` is almost like `allUsers` except `This special scope identifier represents anyone who is authenticated with a Google account`.

crwilcox: As part of test setup we do require the following var

```
- `GCLOUD_TESTS_GOLANG_PROJECT_ID`: Developers Console project's ID (e.g.
bamboo-shift-455) for the general project.
```

I can make it use that as the principal.

frankyn: @tritone clarified that it's acceptable to keep allAuthenticatedUsers in this case. Please ignore my concern.

PR #5065, opened 2021-11-02
chore(storage): add default gRPC connection pool

noahdietz:
Adds a default gRPC connection pool of 4 for the gRPC client library. This is the same default used by BigTable, BigQuery Storage, and PubSub.


storage/storage.go
```diff
@@ -81,6 +81,10 @@ const (
 	// ScopeReadWrite grants permissions to manage your
 	// data in Google Cloud Storage.
 	ScopeReadWrite = raw.DevstorageReadWriteScope
+
+	// defualtConnPoolSize is the default number of connections
```
crwilcox, line 85: nit: typo

noahdietz: Done! Thanks :)

storage/storage.go
```diff
@@ -81,6 +81,10 @@ const (
 	// ScopeReadWrite grants permissions to manage your
 	// data in Google Cloud Storage.
 	ScopeReadWrite = raw.DevstorageReadWriteScope
+
+	// defaultConnPoolSize is the default number of connections
+	// to initialize in the GAPIC gRPC connection pool.
```
tritone, line 86: Maybe for later but do we want to add a comment about how a larger pool may be required for high-throughput operations?

tritone (commented):
Generally looks good!

storage/storage.go
noahdietz: I will add something here, but the proper documentation I think should be written when we take this public.

noahdietz: Added a bit.

tritone (approved):

PR #5072, opened 2021-11-03
feat(bigtable/cbt): cbt 'import' cmd to parse a .csv file and write to CBT

markduffett:
Add 'import' command to 'cbt' and unit/emulator tests
- parses and imports .csv files only for now
- should be accompanied/shortly followed by a tutorial + sample.csv
- csv file formatting assumes a column family row is present unless flag 'column-family' is set, then it will use that family for all columns
- configurable batch-size(500 rows) and worker threads(1) defaults

Runtime expectations:
2MB file with 12588 rows
- 1s with 20 workers, 5s with the default 1

190MB file in 4s @ 47.5MB/s when using 'batch-size=5' 'workers=20'
- 50%+ of 1 Node/tabletservers' CPU
- 1288 rows, 1000 cols, 5kB data populated in 3% of cells
- 200MB/s+ depending on mutation size and number of CBT nodes


bigtable/cmd/cbt/cbt_test.go
```diff
@@ -171,3 +178,438 @@ func TestParseColumnsFilter(t *testing.T) {
...
+		}
+		colMap[fams[i]+":"+cols[i]] = col
+	}
+	for rowIdx, data := range rowData {
+		if row, err := tbl.ReadRow(ctx, data[0]); err != nil {
+			t.Errorf("Error %s", err)
+		} else {
+			for _, fam := range fams {
+				for _, column := range row[fam] {
+					colId := string(column.Column)
```
crwilcox, line 410: Seems the linter is upset about this var. in Go, Id => ID.

```
bigtable/cmd/cbt/cbt_test.go:410:6: var colId should be colID
bigtable/cmd/cbt/cbt_test.go:482:5: var colId should be colID
bigtable/cmd/cbt/cbt_test.go:597:7: var colId should be colID
```

markduffett: ok, I have a cleaner way to do this verification logic too, will update

crwilcox:
Looks like another linter error showed up. 
```
bigtable/cmd/cbt/cbt_test.go:384:10: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
```

Also, if you'd like, the script that is running is `/internal/kokoro/vet.sh`. You can get pretty close by running `golint ./...`, `gofmt`, `goimports`, though there are a few linter instances where we ignore failures (linting was added later)

bigtable/cmd/cbt/cbt.go
```diff
@@ -1534,6 +1552,180 @@ func doDeleteAppProfile(ctx context.Context, args ...string) {
...
+			ia.fams = append(ia.fams, strings.Split(arg, "=")[1])
+		case strings.HasPrefix(arg, "batch-size="):
+			ia.sz, err = strconv.Atoi(strings.Split(arg, "=")[1])
+			if err != nil || ia.sz <= 0 || ia.sz >= 100000 {
+				return ia, fmt.Errorf("batch-size must be > 0 and <= 100000")
+			}
+		case strings.HasPrefix(arg, "workers="):
+			ia.workers, err = strconv.Atoi(strings.Split(arg, "=")[1])
+			if err != nil || ia.workers <= 0 {
+				return ia, fmt.Errorf("Workers must be > 0, err:%s", err)
```
crwilcox, line 1607: Typically, error strings shouldn't be capitalized. I think there are a few others in here as well.
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

bigtable/cmd/cbt/cbt_test.go
```diff
@@ -171,3 +178,401 @@ func TestParseColumnsFilter(t *testing.T) {
...
+			len(got.fams) != len(tc.out.fams) ||
+			got.sz != tc.out.sz ||
+			got.workers != tc.out.workers {
+			t.Errorf("parseImportArgs(%q) did not fail, out: %q", tc.in, got)
+			continue
+		}
+		for i, f := range got.fams {
+			if f != tc.out.fams[i] {
+				t.Errorf("parseImportArgs(%q) incorrect column families, out: %q", tc.in, got)
+				continue
```
crwilcox, line 226: I think this continue is a no-op?

bigtable/cmd/cbt/cbt_test.go
```diff
@@ -171,3 +178,401 @@ func TestParseColumnsFilter(t *testing.T) {
...
+			}
+		}
+	}
+}
+
+func writeAsCSV(data [][]string) ([]byte, error) {
```
crwilcox, line 232: at first I thought this was a mock, maybe we should add a comment that this is to transform test data for the test?

bigtable/cmd/cbt/cbt_test.go
```diff
@@ -171,3 +178,401 @@ func TestParseColumnsFilter(t *testing.T) {
...
+		t.Fatalf("parseAndWrite() failed unexpectedly")
+	}
+
+	if err := validateData(ctx, tbl, fams, cols, rowData); err != nil {
+		t.Fatalf("Read back validation error:%s", err)
+	}
+}
+
+func TestCsvParseAndWriteBadFamily(t *testing.T) {
+	ctx, client := setupEmulator(t, []string{"my-table"}, []string{"my-family"})
```
crwilcox, line 431: I wonder if this couldn't use the NewEmulateEnv? 
https://github.com/googleapis/google-cloud-go/blob/6be4cae4555c128f1dee7d778f0b8358d20c6cee/bigtable/export_test.go#L150


Related, but not likely relevant to these tests: 

Also, we have some tests that don't work in emulator or prod. 

```
if !testEnv.Config().UseProd {
		t.Skip("emulator doesn't support IAM Policy creation")
	}
```

Unless these tests wouldn't work against prod, in which case it may be relevant in reverse?

bigtable/cmd/cbt/cbt.go
```diff
@@ -1534,6 +1552,180 @@ func doDeleteAppProfile(ctx context.Context, args ...string) {
...
+				log.Fatalf("Error: %s", e)
+			}
+		}(i)
+	}
+	wg.Wait()
+	log.Printf("Done importing %d rows.\n", sr.t)
+}
+
+func parseCsvHeaders(r *csv.Reader, fams []string) ([]string, []string, error) {
+	var err error
```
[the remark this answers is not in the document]
markduffett: I wasn't sure how else to handle the nested :=, open to feedback. It has been refactored slightly now too since the fams [] string parameter was not intuitive with a balnk first value then using the second. It should now just be "" or have a family.

bigtable/cmd/cbt/cbt_test.go
markduffett: I don't touch anything admin based, so fine on the IAM front. Looking into the emulator now.

crwilcox: :+1: sorry for the confusion, I was just showing that our testEnv bits should be able to detect if it was requested we use prod for the test run or not. If you know tests work in emulator and/or prod you can guard them :)

markduffett: For NewEmulatedEnv I think many of the steps in setupEmulator() will still be required as seen here https://github.com/googleapis/google-cloud-go/blob/6be4cae4555c128f1dee7d778f0b8358d20c6cee/bigtable/bigtable_test.go#L181

I tried to replace the bttest.NewServer() call with it and for some reason NewEmulatedEnv isn't being discovered.

bigtable/cmd/cbt/cbt_test.go
markduffett: reworded to "transformToCsvBuffer()" to be more descriptive, can add more comments if required

bigtable/cmd/cbt/cbt_test.go
markduffett: We are ok running the cbt commands against the emulator as the behaviour should be as expected. It could be another feature to add integration test support for prod, however the current setup has this tucked in integration_test.go and exporter_test.go files in the parent package, and setting up compatibility with downstream cbt methods (cyclic imports) and is outside scope currently.

bigtable/cmd/cbt/cbt_test.go
```diff
@@ -171,3 +178,390 @@ func TestParseColumnsFilter(t *testing.T) {
...
+		if !tc.fail && err != nil {
+			t.Errorf("parseCsvHeaders() failed. input:%+v, error:%s", tc, err)
+			continue
+		}
+		if tc.fail && err == nil {
+			t.Errorf("parseImportArgs() did not fail. input:%+v, error:%s", tc, err)
+			continue
+		}
+		if tc.fail {
+			continue
```
crwilcox, line 288: nit: I think more typically the tests just catch this via an OR or two separate conditionals, collapsing this a bit? I may be missing nuance here

markduffett: This case would allow continuation since the fail is expected, but we don't want to do any comparisons below. It can't be paired with any of the above because they add errors were as this is expected.

I can put the if err != nil {t.Error} inside the tc.fail block,  just wasn't sure if go-style preferred the nested ifs.
```
if tc.fail {
    if err != nil {
	t.Errorf("parseImportArgs() did not fail. input:%+v, error:%s", tc, err)
    }
    continue
}
```

crwilcox: this to me seems like we are trying to check if fail end err match. I don't see why they would have different messages. Also one of these has 'parseImportArgs' which I assume is leftover from a copy paste?

Are we in essence checking this
```
if tc.fail != (err != nil)
```

Alternatively, rather than storing tc.fail as a bool, store the err string to do a deeper check: possibly kind of like https://github.com/googleapis/google-cloud-go/blob/master/datastore/datastore_test.go#L1987

markduffett: updated to error matching pattern.

PR #5075, opened 2021-11-03
test(bigtable): allow multiple tests to create app profiles

### LINKED ISSUE #5046: bigtable: TestIntegration_InstanceAdminClient_AppProfile failed
Note: #4895 was also for this test, but it was closed more than 10 days ago. So, I didn't mark it flaky.

----

commit: 910e11a9979900cb3f2ff8bc12de1274529e9d4a
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/22792bdd-0d8e-4921-aeb8-36edf0c4316b), [Sponge](http://sponge2/22792bdd-0d8e-4921-aeb8-36edf0c4316b)
status: failed
<details><summary>Test output</summary><br><pre>    integration_test.go:2323: Delete app profile: rpc error: code = NotFound desc = Failed to read: projects/{36639933145}/instances/gc-bt-it-instance/app_profiles/app_profile1</pre></details>

crwilcox:
fixes #5046

### DIFF
--- a/bigtable/integration_test.go
+++ b/bigtable/integration_test.go
@@ -2262,14 +2262,21 @@ func TestIntegration_InstanceAdminClient_AppProfile(t *testing.T) {
 		t.Fatalf("List app profile: %v", err)
 	}
 
+	fmt.Print(profiles)
+
+	// Ensure the profiles we require exist. profiles ⊂ allProfiles
+	verifyProfilesSubset := func(allProfiles []*btapb.AppProfile, profiles map[string]struct{}) {
+		for _, profile := range allProfiles {
+			delete(profiles, profile.Name)
+		}
+		if len(profiles) > 0 {
+			t.Fatalf("Initial app profile list missing profile: %v", profiles)
+		}
+	}
+
 	// App Profile list should contain default, app_profile1
 	wantProfiles := map[string]struct{}{"default": {}, profileID: {}}
-	for _, profile := range profiles {
-		delete(wantProfiles, profile.Name)
-	}
-	if len(wantProfiles) > 0 {
-		t.Fatalf("Initial app profile list missing profile: %v", wantProfiles)
-	}
+	verifyProfilesSubset(profiles, wantProfiles)
 
 	for _, test := range []struct {
 		desc   string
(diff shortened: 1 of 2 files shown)


codyoss (approved):

bigtable/integration_test.go
```diff
@@ -2263,8 +2263,12 @@ func TestIntegration_InstanceAdminClient_AppProfile(t *testing.T) {
 	}
 
 	// App Profile list should contain default, app_profile1
-	if got, want := len(profiles), 2; got != want {
-		t.Fatalf("Initial app profile list len: %d, want: %d", got, want)
+	wantProfiles := map[string]struct{}{"default": {}, profileID: {}}
+	for _, profile := range profiles {
```
enocom, line 2267: This took me a moment to figure out what's going on (without reading the file). The more verbose version might be:

``` golang
equalProfiles := func(a, b map[string]struct{}) {
  if len(a) != len(b) {
    return false
  }
  for k, _ := range a {
    if _, ok := b[k]; !ok {
      return false
    }
  }
  return true
}

wantProfiles := map[string]struct{}{"default": {}, profileID: {}}
if !equalProfiles(wantProfiles, profiles) {
  t.Fatalf("...")
}
```

But if `equalProfiles` weren't useful anywhere else, this more concise version is probably better.

enocom (approved):

bigtable/integration_test.go
crwilcox: So this is the interesting bit, it isn't checking for equality. 

The test creates profiles ('a', 'b')
The live service has existing profile ('c')

we need to verify all profiles we created exist, but we don't care about 'c'.

That said, I can certainly
1) add a comment explaining this
2) extract the slightly odd code to a func as you did with a name like "ensureProfiles" or something of the sort.

enocom: Extracting the check into a function seems useful and a comment, if only for the tired, future reader. :-)

tritone (approved):
+1 to Eno's comment, otherwise looks good.

bigtable/integration_test.go
crwilcox: I restructured this. I am not sure that brought much clarity, but I could be wrong. I did add a comment explaining what I am checking here though. lmk what you think.

enocom: I'm a sucker for intention revealing code and adding the function makes a big difference for me. ^^;

PR #5078, opened 2021-11-03
feat(bigquery/storage/managedwriter): extend managedstream to support call options

shollyman:
BREAKING CHANGE:  changes function signatures to add variadic call options

This plumbs the ability to pass gax.CallOption opts to the
underlying client underpinning the ManagedStream.  It also
adds a WithAppendRowsCallOption option to the constructor,
as well as adding direct option passing for operations like
Finalize() and FlushRows().


codyoss (approved):
LGTM. It might we worth calling out in the commit message that this is a breaking change though. I would not use the feat! thing though.(To avoid confusing release please)

tswast (approved):

shollyman:
> LGTM. It might we worth calling out in the commit message that this is a breaking change though. I would not use the feat! thing though.(To avoid confusing release please)

Added the "BREAKING CHANGE:" stanza to the description

PR #5079, opened 2021-11-04
feat(storage): post policy can be signed with a fn that takes raw bytes

### LINKED ISSUE #4752: storage: PostPolicyV4Options.SignBytes is not compatible with Credentials API
**Client**

Cloud Storage

**Environment**

Reproduced in local and remote environment

**Go Environment**

$ go version
go version go1.16.5 darwin/amd64

**Code**

The definition of `PostPolicyV4Options.SignBytes` and `SignedURLOptions.SignBytes` differ.

The input of `PostPolicyV4Options.SignBytes` is hashedBytes.

https://github.com/googleapis/google-cloud-go/blob/27c3ed022c83e598bacf34fbce47911985e0af35/storage/post_policy_v4.go#L221-L233
https://github.com/googleapis/google-cloud-go/blob/27c3ed022c83e598bacf34fbce47911985e0af35/storage/post_policy_v4.go#L310-L311

The input of `SignedURLOptions.SignBytes` is raw bytes.
https://github.com/googleapis/google-cloud-go/blob/27c3ed022c83e598bacf34fbce47911985e0af35/storage/storage.go#L699-L715

This difference seems to make `PostPolicyV4Options.SignBytes` incompatible with `projects.serviceAccounts.signBlob` of Service Account Credentials API.

**Code**

```go
email := os.Getenv("IMPERSONATE_SERVICE_ACCOUNT")
var signBytesFn func(b []byte) ([]byte, error)
var privateKey []byte

if email != "" {
	credCli, err := credentials.NewIamCredentialsClient(ctx)
	if err != nil {
		 panic(err)
	}
	defer credCli.Close()

	signBytesFn = func(b []byte) ([]byte, error) {
		 req := &credentialspb.SignBlobRequest{
			 Payload: b,
			 Name:    email,
		 }
		 resp, err := credCli.SignBlob(ctx, req)
		 if err != nil {
			 panic(err)
		 }
		 return resp.SignedBlob, err
	}
} else {
	cred, _ := google.FindDefaultCredentials(ctx)
	jwtConfig, _ := google.JWTConfigFromJSON(cred.JSON)
	email = jwtConfig.Email
	privateKey = jwtConfig.PrivateKey
}

opts := &storage.PostPolicyV4Options{
	GoogleAccessID: email,
	SignBytes:      signBytesFn,
	PrivateKey:     privateKey,
	Expires:        time.Now().Add(10 * time.Minute),
	Fields: &storage.PolicyV4Fields{
		 Metadata: map[string]string{},
	},
}

policy, err := storage.GenerateSignedPostPolicyV4(bucket, object, opts)
```

**Expected behavior**

Signed policy document are valid in both cases of `IMPERSONATE_SERVICE_ACCOUNT ` or `GOOGLE_APPLICATION_CREDENTIALS`.

**Actual behavior**

Signed policy document only valid in case of `GOOGLE_APPLICATION_CREDENTIALS`.

**Additional context**

Possible patch
https://github.com/googleapis/google-cloud-go/compare/storage/v1.16.1...apstndb:fix-policy-document

BrennaEpp:
* Adds a `SignRawBytes` field to `PostPolicyV4Options`. This field works like the `SignedURLOptions` `SignBytes` field.  This function expects raw bytes as opposed to the `SignBytes` field which expects the bytes to be hashed before called.
* Deprecates the `SignBytes` field of `PostPolicyV4Options`
* Adds a test to ensure that both `bucket.SignedURL` and `GenerateSignedPostPolicyV4` work with the same signing function for `SignBytes` and `SignRawBytes` respectively.
* Refactors some test code to reuse checks for signed urls and post policies; adds a check to `TestIntegration_SignedURL` to error on an expected failure not failing. 

Fixes #4752


codyoss:
@googlebot I consent.

storage/post_policy_v4.go
```diff
@@ -52,22 +52,30 @@ type PostPolicyV4Options struct {
...
 	//     		_, signedBytes, err := appengine.SignBytes(ctx, b)
 	//     		return signedBytes, err
 	//     	},
 	//     	// etc.
 	//     })
 	//
-	// Exactly one of PrivateKey or SignBytes must be non-nil.
-	SignBytes func(hashBytes []byte) (signature []byte, err error)
+	// Exactly one of PrivateKey or SignRawBytes must be non-nil.
+	// If previously using SignBytes, add the following to use SignRawBytesInstead:
```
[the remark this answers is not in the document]
BrennaEpp: Woops! Forgot to commit some changes (:

storage/integration_test.go
```diff
@@ -4171,77 +4171,17 @@ func TestIntegration_PostPolicyV4(t *testing.T) {
 	}
 
 	objectName := "my-object.txt"
```
tritone, line 4173: Make the object name specific to the test; no reason to cause rate limit issues by reusing.

storage/integration_test.go
```diff
@@ -4329,24 +4269,181 @@ func TestBucketSignURL(t *testing.T) {
 		bkt := test.client.Bucket(bucketName)
 		url, err := bkt.SignedURL(obj, &test.opts)
 		if err != nil {
-			t.Fatalf("unable to create signed URL: %v", err)
+			t.Fatalf("%s: unable to create signed URL: %v", test.desc, err)
```
tritone, line 4272: Probably best to use subtests to achieve this, but either way is fine.

storage/integration_test.go
```diff
@@ -4329,24 +4269,181 @@ func TestBucketSignURL(t *testing.T) {
...
-			t.Fatalf("resp.StatusCode = %v, want 200: %v", resp.StatusCode, err)
+
+		if err := verifySignedURL(url, nil, contents); err != nil {
+			t.Fatalf("%s: %v", test.desc, err)
 		}
-		b, err := ioutil.ReadAll(resp.Body)
+	}
+}
+
+// Tests that the same SignBytes function works for both GeneratePostPolicyV4 and SignedURL
```
tritone, line 4281: Specify that this is `SignRawBytes` for post policy. Also, nice test addition!

storage/post_policy_v4.go
```diff
@@ -52,22 +52,31 @@ type PostPolicyV4Options struct {
 	// Exactly one of PrivateKey or SignBytes must be non-nil.
 	PrivateKey []byte
 
-	// SignBytes is a function for implementing custom signing. For example, if
+	// SignBytes is a function for implementing custom signing.
```
tritone, line 55: Maybe include a sentence about how it differs from `SignRawBytes` even though we are deprecating it.

storage/post_policy_v4.go
```diff
@@ -52,22 +52,31 @@ type PostPolicyV4Options struct {
...
+	//
+	// Deprecated: Use SignRawBytes. If both SignBytes and SignRawBytes are defined,
+	// SignBytes will be ignored.
+	// Add the following to the top of your signing function to use SignRawBytes instead:
+	//		shaSum := sha256.Sum256(bytes)
+	//		bytes = shaSum[:]
+	SignBytes func(hashBytes []byte) (signature []byte, err error)
+
+	// SignRawBytes is a function for implementing custom signing. For example, if
```
tritone, line 64: I would note explicitly that the same function that works for `SignBytes` for signed URLs will work for `SignRawBytes` for post policy. Obviously it's confusing but still better to state explicitly.

storage/post_policy_v4.go
```diff
@@ -307,10 +318,18 @@ func GenerateSignedPostPolicyV4(bucket, object string, opts *PostPolicyV4Options
...
 
 	b64Policy := base64.StdEncoding.EncodeToString(condsAsJSON)
-	shaSum := sha256.Sum256([]byte(b64Policy))
-	signature, err := signingFn(shaSum[:])
-	if err != nil {
-		return nil, err
+	var signature []byte
+	var signErr error
+	// SignBytes is used only if SignRawBytes is not defined
+	if opts.SignRawBytes == nil && opts.SignBytes != nil {
```
tritone, line 324: I worry about the fact that we are checking this condition multiple times. Maybe we just add an unexported field to the options about whether the signed bytes should be hashed or not, set that once when parsing the options, then make subsequent decisions accordingly?

tritone (commented):
Generally looks good, a few comments mostly on docs.

storage/integration_test.go
```diff
@@ -4329,24 +4269,181 @@ func TestBucketSignURL(t *testing.T) {
...
+	}
+	if !bytes.Equal(got, expectedFileBody) {
+		return fmt.Errorf("got %q, want %q", got, expectedFileBody)
+	}
+	return nil
+}
+
+// verifyPostPolicy uploads a file to the obj using the provided post policy and
+// verifies that it was uploaded correctly
+func verifyPostPolicy(pv4 *PostPolicyV4, obj *ObjectHandle, bytesToWrite []byte, statusCodeOnSuccess int) error {
```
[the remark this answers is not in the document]
BrennaEpp: It can be different depending on what you specify in `PostPolicyV4Options.PolicyV4Fields.StatusCodeOnSuccess`

storage/integration_test.go
BrennaEpp: each test is on a different bucket, but doesn't hurt to change the object name as well

tritone (approved):
Nice improvements, looks really good!

storage/integration_test.go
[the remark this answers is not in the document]
tritone: I think currently it is 200 in all these tests, so we shouldn't create more constants unnecessarily if possible.

PR #5089, opened 2021-11-04
ci: filter presubmit tests to run only changed directories

BrennaEpp:
* Filters presubmit tests based on changed directories
* Adds emulator tests to be run when all module tests run, such as during regular nightly jobs


codyoss (approved):
LGTM, thank you!

PR #5090, opened 2021-11-05
feat(bigtable/spanner): add google-c2p dependence to bigtable and spanner

mohanli-ml:
According to discussion in https://github.com/googleapis/google-api-go-client/issues/1283, we plan to move the google-c2p dependence to bigtable/spanner, which are services that will use DirectPath.

feat(bigtable): add google-c2p dependence

feat(spanner): add google-c2p dependence


codyoss:
Please note that this does seem to carry a large amount of new dependencies. Open to other suggestions but I think it is at least better to limit the scope of these to the clients using the feature rather than being on the hot-path for all clients.

mohanli-ml:
As the plan is that CBT/Spanner will use DirectPath by default, we will eventually add the google-c2p dependencies to CBT/Spanner client repo, so I guess it does not harm to do it now?

hengfengli:
I don't see this PR introduces more dependencies in Spanner. LGTM.

X0Xoo:
Stop please

tamird:
Hello. Is this ready to be merged? It appears to be required for https://github.com/googleapis/google-api-go-client/issues/1283.

mohanli-ml:
@codyoss I run `go mod tidy` to both bigtable and spanner, but only bigtable/go.sum is updated, while spanner/go.sum is still the same. Do you know why?

codyoss:
@mohanli-ml I think are Spanner lib does its own gRPC connection handling to support session/transaction affinity IIRC. It may be that the client already relied on an additional gRPC package so it does not have more new code to pull in. This is a guess without looking at the code.

tamird:
Hello. Any progress here?

mohanli-ml:
@codyoss Can you help to approve and merge this PR? Thanks!

codyoss (approved):

PR #5102, opened 2021-11-09
feat(bigquery/storage/managedwriter): support variadic appends

shollyman:
BREAKING CHANGE: adds a variadic option to the AppendRows() method, removes offset argument

This updates the call signature to allow variadic appends, and introduces two new AppendOption options: one for setting the offset in an optional fashion (`WithOffset()`), and one for updating the schema (`UpdateSchemaDescriptor()`).

Due to current API limitations, this means that we need to close/reconnect the open connection when this option is passed.  However, this should be resolved in the backend eventually.  Internal issue 205756033 tracks this.

In practice this means the following changes need attention by consumers of this library:

For the "don't set an offset" behavior:
* Old: `mystream.AppendRows(ctx, data, managedwriter.NoStreamOffset)`
* New: `mystream.AppendRows(ctx, data)`

For the "set an an offset" behavior:
* Old: `mystream.AppendRows(ctx, data, offset)`
* New: `mystream.AppendRows(ctx, data, managedwriter.WithOffset(offset))`


codyoss (commented):
LGTM from a Go perspective. I will let Tim sign off from the BQ perspective.

bigquery/storage/managedwriter/appendresult.go
```diff
@@ -66,7 +67,9 @@ func (ar *AppendResult) GetResult(ctx context.Context) (int64, error) {
 // append request.
 type pendingWrite struct {
 	request *storagepb.AppendRowsRequest
-	result  *AppendResult
+	// for schema evolution cases, accept a new schema
+	newSchema *descriptorpb.DescriptorProto
```
tswast, line 71: Will this property need to be removed once 205756033 is resolved? It might be redundant with `AppendRowsRequest.proto_rows.writer_schema.proto_descriptor`.

Or am I understanding that email thread that we won't be exposing the underlying AppendRowsRequest to users?

bigquery/storage/managedwriter/integration_test.go
```diff
@@ -513,6 +517,85 @@ func testInstrumentation(ctx context.Context, t *testing.T, mwClient *Client, bq
...
+	if err := testTable.Create(ctx, &bigquery.TableMetadata{Schema: testdata.SimpleMessageSchema}); err != nil {
+		t.Fatalf("failed to create test table %s: %v", testTable.FullyQualifiedName(), err)
+	}
+
+	m := &testdata.SimpleMessageProto2{}
+	descriptorProto := protodesc.ToDescriptorProto(m.ProtoReflect().Descriptor())
+
+	// setup a new stream.
+	ms, err := mwClient.NewManagedStream(ctx,
+		WithDestinationTable(fmt.Sprintf("projects/%s/datasets/%s/tables/%s", testTable.ProjectID, testTable.DatasetID, testTable.TableID)),
```
tswast, line 531: Interesting. So the users don't see the stream ID, either? I guess that gives us enough flexibility for the current schema evolution workaround.

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -189,9 +189,13 @@ func (ms *ManagedStream) getStream(arc *storagepb.BigQueryWrite_AppendRowsClient
 	}
 
 	// Always return the retained ARC if the arg differs.
-	if arc != ms.arc {
+	if arc != ms.arc && !forceReconnect {
 		return ms.arc, ms.pending, nil
 	}
+	if arc != ms.arc && forceReconnect && ms.arc != nil {
+		// TODO: is closing send sufficient?
```
tswast, line 196: TODO was verified by integration test?

tswast: Or do we need to create a new stream if it's not the default stream?

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -327,15 +331,29 @@ func (ms *ManagedStream) Close() error {
 // AppendRows sends the append requests to the service, and returns a single AppendResult for tracking
 // the set of data.
 //
-// The format of the row data is binary serialized protocol buffer bytes, and and the message
-// must adhere to the format of the schema Descriptor passed in when creating the managed stream.
-func (ms *ManagedStream) AppendRows(ctx context.Context, data [][]byte, offset int64) (*AppendResult, error) {
+// The format of the row data is binary serialized protocol buffer bytes.  The message must be compatible
+// with the schema currently set for the stream.
+//
+// Use the sentinel value NoStreamOffset to omit sending of the offset value.
```
tswast, line 337: From my testing, they must send an offset if using PENDING mode. And they can't if using default stream, right? I didn't test with BUFFERED, so maybe it's optional there? Maybe there's a better way to communicate when NoStreamOffset should be set?

tswast (commented):
A few questions.

bigquery/storage/managedwriter/appendresult.go
shollyman: In this veneer I'm wrapping the append request.  My expectation is we'll end up with per-stream caches of schema and per-stream append queues for the multiplexing case.  Currently we just retain a single schema and append queue (though in go it's a channel rather than a queue).

bigquery/storage/managedwriter/integration_test.go
shollyman: There's effectively two paths to getting a managed stream: allow `NewManagedStream()` to deal with the stream construction by specifying table/type/etc, or do it yourself and pass it in via `WithStreamName()` option.

bigquery/storage/managedwriter/managed_stream.go
shollyman: Interestingly enough, this gets to the other breaking change I've been considering here.  Remove offset as a static argument from the AppendRows() function and add a `WithOffset()` AppendOption.

The origin of the `NoStreamOffset` was to simplify the AppendResult, as Go's lack of null vs default value makes it more complex to deal with the optional offset.  I didn't want to do `*int64`, but it's an option to consider as well.

bigquery/storage/managedwriter/integration_test.go
shollyman: I expect users who explicitly do stream construction to be more likely to be doing dynamic proto schema stuff, as one of the things you get back from stream creation is table schema.

bigquery/storage/managedwriter/appendresult.go
GlenDC: Is this also the reason why you can optionally have a new schema updated as part of a return error. I was puzzled how one can automatically fix schema updates on runtime though? Is there any realistic example of this?

bigquery/storage/managedwriter/integration_test.go
```diff
@@ -513,6 +517,85 @@ func testInstrumentation(ctx context.Context, t *testing.T, mwClient *Client, bq
...
+	//       In the continuous case, we'd get it from the response, but the change-and-wait case needs something more.
+	time.Sleep(6 * time.Second)
+
+	// ready descriptor, send an additional append
+	m2 := &testdata.SimpleMessageEvolvedProto2{
+		Name:  proto.String("evolved"),
+		Value: proto.Int64(180),
+		Other: proto.String("hello evolution"),
+	}
+	descriptorProto = protodesc.ToDescriptorProto(m2.ProtoReflect().Descriptor())
```
GlenDC, line 578: This I find one of the more harder parts as a beginner to start making use of proto models for streaming into the storage API. It's quite a chain of commands that aren't really something you would ever figure out by just looking at the API. Only way I learned how to do this was by checking these examples. I wonder if somehow there isn't an easier way we can make it just pass in a the generated proto type somehow. Dunno. Haven't found an easy way myself for my `bqwriter` wrapper, otherwise I would have already done it.

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -189,9 +189,13 @@ func (ms *ManagedStream) getStream(arc *storagepb.BigQueryWrite_AppendRowsClient
 	}
 
 	// Always return the retained ARC if the arg differs.
-	if arc != ms.arc {
+	if arc != ms.arc && !forceReconnect {
 		return ms.arc, ms.pending, nil
 	}
+	if arc != ms.arc && forceReconnect && ms.arc != nil {
+		// TODO: is closing send sufficient?
+		(*ms.arc).CloseSend()
```
GlenDC, line 197: this smells fishy, or is it just me?

bigquery/storage/managedwriter/managed_stream.go
GlenDC: Given this is already a kind of managed writer i wonder if this cannot be abstracted away completely? But I guess to not force how to handle errors in a retry-able manner it is needed to expose it? Either way, I do not find the NoStreamOffset sentient value a big deal, works fine.

bigquery/storage/managedwriter/appendresult.go
shollyman: Yeah, the returned schema is the notification of schema extension.  It's a case where you do something like an ALTER TABLE ADD COLUMN or extend schema via tables.update while streaming data, and the change gets acknowledged by the streaming backend by setting the new schema.  My expectation is we'll add a callback registration for this as well, but not in this change.

bigquery/storage/managedwriter/integration_test.go
shollyman: Two reasons:  this is a bit of a standard in the APIs (see aip.dev for more context), and we want to avoid some circular dependencies by having the managedwriter depend on cloud.google.com/go/bigquery directly. 

https://github.com/googleapis/google-cloud-go/pull/5017 will make it easier to generate the string for the table resources if you end up using this option from a bigquery resource.

bigquery/storage/managedwriter/managed_stream.go
shollyman: Went ahead and made the offset part of the variadic options.

This changes the signatures for appending to:
* No offset set: `<ManagedStream>.AppendRows(ctx, data)`
* Offset set: `<ManagedStream>.AppendRows(ctx, data, WithOffset(offset))`

tswast (commented):
AppendOption design LGTM, but one worry:

By hiding the fact that we're recreating the stream, does it make the use of offset harder to understand / make it a breaking change when the API no longer requires a stream reset? Though, perhaps there's some signal we send at the end of the stream so that folks don't need to know specifically that a schema change could cause that?

shollyman:
> AppendOption design LGTM, but one worry:
> 
> By hiding the fact that we're recreating the stream, does it make the use of offset harder to understand / make it a breaking change when the API no longer requires a stream reset? Though, perhaps there's some signal we send at the end of the stream so that folks don't need to know specifically that a schema change could cause that?

Management of the network stream connection is abstracted from the user.  This PR uses a CloseSend() on the network stream to cleanly signal a new connection: existing appends in flight will still process on the recv side, and the next append (either new or due to a retry) will pick up a new connection.  Schema change notification from the backend isn't currently in this veneer (but will come in a future PR).

If you change the schema in a compatible way, retrying an old proto message with a new schema that has additional fields/tags shouldn't be an issue as that's the power of proto extension in a nutshell.  If the table is changed to an incompatible schema, the stream itself is invalid (for explicitly created streams).  Default streams are a special case here, but essentially a similar metadata inconsistency like the existing tabledata.insertall when schema changes arrive.

tswast (approved):
LGTM after clarifying offline that it's reconnecting but not creating a new (backend) stream. Hooray for ambiguous streams 🙄

bigquery/storage/managedwriter/managed_stream.go
shollyman: It's a temporary issue until the backend allows schema change on an already open stream connection; I'm not enamored of it but this will get cleaned up.

PR #5103, opened 2021-11-09
ci: fix variable for presubmits

BrennaEpp:
`$KOKORO_GIT_COMMIT` is not set in the presubmit invocation, so the presubmit filtering is not working. Use `$KOKORO_GIT_COMMIT_google_cloud_go` which is set instead.


codyoss (approved):
Ah, good catch. I forget that we have multiple sources so the environment variable is qualified with a suffix.

PR #5105, opened 2021-11-10
feat(storage): GenerateSignedPostPolicyV4 can use existing creds to authenticate

BrennaEpp:
Parallel change to #4604 - this implements the same functionality for generating signed post policies.

Users can omit filling in GoogleAccessID, PrivateKey, and SignRawBytes if they are okay with default values for these.

* adds a test for this method
* changes existing tests to use this method as it wraps the original
* adds a bit more logging to `mustDeleteObject`


storage/post_policy_v4.go
```diff
@@ -116,6 +116,20 @@ type PostPolicyV4Options struct {
 	shouldHashSignBytes bool
 }
 
+func (opts *PostPolicyV4Options) clone() *PostPolicyV4Options {
+	return &PostPolicyV4Options{
+		GoogleAccessID:      opts.GoogleAccessID,
+		PrivateKey:          opts.PrivateKey,
+		SignBytes:           opts.SignBytes,
```
tritone, line 123: Missing `SignRawBytes`?

storage/integration_test.go
```diff
@@ -4335,7 +4407,7 @@ func TestIntegration_PostPolicyV4_SignedURL_WithSignBytes(t *testing.T) {
 		},
 	}
 
-	pv4, err := GenerateSignedPostPolicyV4(bucketName, objectName, ppv4Opts)
+	pv4, err := bucket.GenerateSignedPostPolicyV4(objectName, ppv4Opts)
```
tritone, line 4410: I don't think we have to switch for this test since it's explicitly about using the `SignRawBytes` option.

tritone (commented):
Looking good!

storage/post_policy_v4.go
BrennaEpp: oh no!! good catch!

storage/integration_test.go
BrennaEpp: makes sense 👍

tritone (approved):
Looks good! @codyoss could you take a look as well before we merge?

PR #5110, opened 2021-11-10
test(storage): add scenario 6 to retry conformance tests

BrennaEpp:
some of the scenario 6 tests are conditionally idempotent and thus do not yet pass


tritone (approved):
Looks good!

PR #5118, opened 2021-11-11
feat(spanner): remove dedefault dial option that disables service config

mohanli-ml:
DirectPath based on Traffic Director can not work with `grpc.WithDisableServiceConfig()`. 
Background: https://github.com/googleapis/google-api-go-client/pull/1260
Bug: b/202169093#comment8

### DIFF
--- a/spanner/apiv1/spanner_client.go
+++ b/spanner/apiv1/spanner_client.go
@@ -63,7 +63,6 @@ func defaultGRPCClientOptions() []option.ClientOption {
 		internaloption.WithDefaultAudience("https://spanner.googleapis.com/"),
 		internaloption.WithDefaultScopes(DefaultAuthScopes()...),
 		internaloption.EnableJwtWithScope(),
-		option.WithGRPCDialOption(grpc.WithDisableServiceConfig()),
 		option.WithGRPCDialOption(grpc.WithDefaultCallOptions(
 			grpc.MaxCallRecvMsgSize(math.MaxInt32))),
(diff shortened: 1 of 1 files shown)


noahdietz:
This option is part of the generated codebase. This file should not be changed manually.

This was added to all cloud clients to prevent APIs from specifying retry policies that would retry in the grpc layer as well as in the gapic layer. 

We can remove it from the generator if you believe this is no longer a concern.

noahdietz:
Is this what you want? https://github.com/googleapis/gapic-generator-go/pull/812

mohanli-ml:
@noahdietz Thanks Noah!

Yes I do want to remove `grpc.WithDisableServiceConfig()` for Spanner, as once it is set I do not have a way to remove it, so it must be removed.

However, I do not have the context about why `grpc.WithDisableServiceConfig()` is set by default. I tried to check, but from blame this option is set 2 years ago. So maybe the secure way is just removing this option for Spanner. Is this feasible? If not then we will need your https://github.com/googleapis/gapic-generator-go/pull/812.

noahdietz:
I've merged the generator change. I will cut a release and update the generation pipeline today (hopefully). You can probably close this PR.

The option was added by default in anticipation of some changes that never came to fruition so I think its OK to remove it at this point.

menghanl:
There's an option to disable retry: https://pkg.go.dev/google.golang.org/grpc#WithDisableRetry

PR #5125, opened 2021-11-12
chore: update microgen to v0.23.4

noahdietz:
Changes include:
* remove disable service config dial option
* ensure that mix-in protos present in the input file set aren't generated as separate clients when only present for mix-in generation
* dependency updates


noahdietz:
cc @mohanli-ml

codyoss (approved):

mohanli-ml:
@noahdietz Hey Noah, when can we expect a new `chore(all): auto-regenerate gapics` PR like #4553, which will actually update the config in spanner_client.go?

noahdietz:
@mohanli-ml yes it runs as a cron job. Check in with @codyoss for updates on that.

codyoss:
@mohanli-ml Another PR should be proposed later this evening.

PR #5147, opened 2021-11-15
chore: bulk gen

codyoss:
This commit will be tagged:

- ids/v0.1.0
- vmmigration/v0.1.0


noahdietz (approved):
`ids` is kind of an unfortunate pkg name, but 🤷🏻‍♂️

Issue #5131, opened 2021-11-15
spanner: ToStruct can ignore the missing fields instead of throwing error


git-hulk:
`ToStruct` requires all select result fields MUST be matched in the struct receiver and it would cause trouble when using the `*` as the select result. Does it make sense that ignore those fields
that didn't exist instead of throwing an error? Or create another new API to workaround?

Example 

DDL:
```sql
CREATE TABLE entries (
    key           STRING NOT NULL,
    value         STRING
) PRIMARY KEY (key)
```

Firstly, we have only two fields `key` and `value`:
```go
struct Entry {
  Key   string
  Value string
}

func GetValue(key string) (*Entry, error) {
  ...
  spannerCli.Single().Query(ctx, spanner.Statement {
      SQL: 'SELECT * FROM entries WHERE key=@key',
      Params: map[string]string{"key": key},
  ).Do(func(r *spanner.Row) error {
       entry := new(Entry)
       r.ToStruct(entry)
  )
}
```
Then we add a new field like `expire` into the table `entries`, the old code would NOT work, so
we should update the fields carefully that DDL shouldn't be updated before changing the code. Another workaround was to specify all fields explicitly but I don't think is a good way when the table fields grow.

I would be happy to send the PR to resolve this issue if it makes sense.

hengfengli:
@rahul2393 Can you please take a look at this issue? Thanks.

PR #5153, opened 2021-11-17
feat(spanner): add ToStructLenient method to decode to struct fields with no error return in case of un-matched row's column with struct's exported fields

rahul2393:
The PR adds method `ToStructLenient` which decodes the row's column into the struct's exported fields, the difference with the existing `ToStruct` methods are:
* Decoding from row's column to struct with no matching exported fields is allowed.

ToStructLenient should pass all existing and future test cases for ToStruct method so tests are updated to make sure same test cases run for both methods.

ToStructLenient is created to handle below example:

DDL:
```
CREATE TABLE Singers (
    SingerId	INT64 NOT NULL,
    FirstName	STRING(1024)
) PRIMARY KEY (SingerId)
```
and client code looks like
```
struct Singer {
  SingerId     int64
  FirstName string
}

func GetSinger(ID int64) (*Singer, error) {
  ...
  spannerCli.Single().Query(ctx, spanner.Statement {
      SQL: 'SELECT * FROM singers WHERE SingerId=@id',
      Params: map[string]string{"id": ID},
  ).Do(func(r *spanner.Row) error {
       singer := new(Singer)
       r.ToStruct(singer)
  )
}
```

Now, let's say we update the singers table to have new column `LastName`, ToStruct will start returning errNoOrDupGoField after schema update to fix this ToStructLenient can be used instead of ToStruct.
  
Fixes: https://github.com/googleapis/google-cloud-go/issues/5131

### DIFF
--- a/spanner/row.go
+++ b/spanner/row.go
@@ -288,6 +288,10 @@ func errToStructArgType(p interface{}) error {
 //   2. Otherwise, if the name of a field matches the name of a column (ignoring case),
 //      decode the column into the field.
 //
+//   3. The number of columns in the row must match the number of exported fields in the struct.
+//      There must be exactly one match for each column in the row. The method will return an error
+//      if a column in the row cannot be assigned to a field in the struct.
+//
 // The fields of the destination struct can be of any type that is acceptable
 // to spanner.Row.Column.
 //
@@ -311,5 +315,48 @@ func (r *Row) ToStruct(p interface{}) error {
 		&sppb.StructType{Fields: r.fields},
 		&proto3.ListValue{Values: r.vals},
 		p,
+		false,
+	)
+}
+
+// ToStructLenient fetches the columns in a row into the fields of a struct.
+// The rules for mapping a row's columns into a struct's exported fields
+// are:
+//
+//   1. If a field has a `spanner: "column_name"` tag, then decode column
+//      'column_name' into the field. A special case is the `spanner: "-"`
+//      tag, which instructs ToStruct to ignore the field during decoding.
+//
+//   2. Otherwise, if the name of a field matches the name of a column (ignoring case),
+//      decode the column into the field.
+//
+//   3. The number of columns in the row and exported fields in the struct do not need to match.
+//      Any field in the struct that cannot not be assigned a value from the row is assigned its default value.
+//      Any column in the row that does not have a corresponding field in the struct is ignored.
+//
+// The fields of the destination struct can be of any type that is acceptable
+// to spanner.Row.Column.
+//
+// Slice and pointer fields will be set to nil if the source column is NULL, and a
+// non-nil value if the column is not NULL. To decode NULL values of other types, use
+// one of the spanner.NullXXX types as the type of the destination field.
+//
+// If ToStructLenient returns an error, the contents of p are undefined. Some fields may
+// have been successfully populated, while others were not; you should not use any of
+// the fields.
+func (r *Row) ToStructLenient(p interface{}) error {
+	// Check if p is a pointer to a struct
+	if t := reflect.TypeOf(p); t == nil || t.Kind() != reflect.Ptr || t.Elem().Kind() != reflect.Struct {
+		return errToStructArgType(p)
+	}
+	if len(r.vals) != len(r.fields) {
+		return errFieldsMismatchVals(r)
+	}
+	// Call decodeStruct directly to decode the row as a typed proto.ListValue.
+	return decodeStruct(
+		&sppb.StructType{Fields: r.fields},
+		&proto3.ListValue{Values: r.vals},
+		p,
+		true,
 	)
 }
--- a/spanner/value.go
+++ b/spanner/value.go
@@ -2871,6 +2871,11 @@ func errNilSpannerStructType() error {
 	return spannerErrorf(codes.FailedPrecondition, "unexpected nil StructType in decoding Cloud Spanner STRUCT")
 }
 
+// errDupGoField returns error for duplicated Go STRUCT field names
+func errDupGoField(s interface{}, name string) error {
+	return spannerErrorf(codes.InvalidArgument, "Go struct %+v(type %T) has duplicate fields for GO STRUCT field %s", s, s, name)
+}
+
 // errUnnamedField returns error for decoding a Cloud Spanner STRUCT with
 // unnamed field into a Go struct.
 func errUnnamedField(ty *sppb.StructType, i int) error {
@@ -2905,7 +2910,7 @@ func errDecodeStructField(ty *sppb.StructType, f string, err error) error {
 // decodeStruct decodes proto3.ListValue pb into struct referenced by pointer
 // ptr, according to
 // the structural information given in sppb.StructType ty.
-func decodeStruct(ty *sppb.StructType, pb *proto3.ListValue, ptr interface{}) error {
+func decodeStruct(ty *sppb.StructType, pb *proto3.ListValue, ptr interface{}, lenient bool) error {
 	if reflect.ValueOf(ptr).IsNil() {
 		return errNilDst(ptr)
 	}
@@ -2921,6 +2926,15 @@ func decodeStruct(ty *sppb.StructType, pb *proto3.ListValue, ptr interface{}) er
 	if err != nil {
 		return ToSpannerError(err)
 	}
+	// return error if lenient is true and destination has duplicate exported columns
+	if lenient {
+		fieldNames := getAllFieldNames(v)
+		for _, f := range fieldNames {
+			if fields.Match(f) == nil {
+				return errDupGoField(ptr, f)
+			}
+		}
+	}
 	seen := map[string]bool{}
 	for i, f := range ty.Fields {
 		if f.Name == "" {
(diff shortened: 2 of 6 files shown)


git-hulk:
Cool, many thanks to @rahul2393

spanner/value.go
```diff
@@ -2871,6 +2871,11 @@ func errNilSpannerStructType() error {
 	return spannerErrorf(codes.FailedPrecondition, "unexpected nil StructType in decoding Cloud Spanner STRUCT")
 }
 
+// errDupGoField returns error for duplicated Go STRUCT field names
+func errDupGoField(s interface{}) error {
+	return spannerErrorf(codes.InvalidArgument, "Go struct %+v(type %T) has duplicate fields for GO STRUCT field", s, s)
```
olavloite, line 2876: nit: Could this error also include the field that is duplicate? I think that would be helpful to any user who needs to debug code that returns this error.

spanner/value.go
```diff
@@ -2921,14 +2926,21 @@ func decodeStruct(ty *sppb.StructType, pb *proto3.ListValue, ptr interface{}) er
 	if err != nil {
 		return ToSpannerError(err)
 	}
+	fieldNames := getAllFieldNames(v)
+	for _, f := range fieldNames {
+		sf := fields.Match(f)
+		if sf == nil {
+			return errDupGoField(ptr)
```
olavloite, line 2933: Is this really a duplicate field? Or is it a missing field? I would expect `fields.Match(f)` to return nil if the field name is not in the list, and that this scenario is that the field is missing and not duplicate.

spanner/value.go
```diff
@@ -2995,6 +3007,38 @@ func decodeStructArray(ty *sppb.StructType, pb *proto3.ListValue, ptr interface{
 	return nil
 }
 
+func getAllFieldNames(v reflect.Value) []string {
+	var names []string
+	s := v
```
olavloite, line 3012: Why do we need this?

olavloite (commented):
There seems to be some integration tests that are failing. Would you mind taking a look at those?

olavloite:
> Fixes: #5131

Also, would you mind extending the PR description to describe what the PR does. I understand that it solves the issue, but adding a short description of what/how makes it easier to review and will automatically be included in the release notes.

spanner/value.go
rahul2393: We need this for cases when GO struct has more than one common spanner tags, In that case we want to return duplicate error. There was no way to get all exported struct field names so created one here. please let me know if you think there is another good way to do this.

rahul2393:
Thanks @olavloite, Integreation tests passed and updated the PR description.

spanner/value.go
olavloite: Sorry, I should have been more specific. I meant literally why do we need this specific line:

`s := v`

Why can't the method just use `v`?

spanner/value.go
rahul2393: Here it is duplicate field, it would be missing if we would have tried matching the spanner.Row field name, but here the same GO struct field names are matched from which the fields object was created.

spanner/value.go
rahul2393: Done

olavloite:
@rahul2393 Thanks for the changes. The implementation itself looks good to me.

I however have a couple of general concerns / questions regarding this:
1. First of all, I think this PR should be labelled `feat` and not `fix`. This is not fixing a bug, but relaxing a restriction that appears to have been added quite deliberately when the feature was created. Would you mind extending the PR description a little bit more to explain what is possible now that was not possible before (preferably with an example)?
2. I'm also wondering whether we need to mark this as a breaking change, or otherwise at least flag it more prominently in the release notes than a 'normal' feature, as we are changing the behavior of the method. The `ToStruct` method would previously return an error in certain cases, and after this change, it will stop returning an error. In theory, there could be users who rely on an error being returned in those cases, and whose code would break after this change.
3. We should update the documentation of the `ToStruct` [method here](https://github.com/googleapis/google-cloud-go/blob/90e2868a3d220aa7f897438f4917013fda7a7c59/spanner/row.go#L280). As far as I can tell, this change does not contradict with anything in the documentation of that method, but that is because the documentation is not explicit in what happens if the result set contains more/less columns than there are fields in the struct. The documentation should be extended to explicitly state what the method does if the exported fields of the struct and the columns in the result set do not match.
4. The documentation also does not explicitly state that it will only fill exported fields. From what I can tell, both the current implementation and the one in this PR will only fill exported fields, and I think the documentation should reflect that.

rahul2393:
@olavloite I have updated the PR to add new method instead of updating the existing ToStruct method so it won't be a breaking change, also updated the PR description. same tests should be run for ToStructLenient which were created for ToStruct so tests are updated, this will make sure that functionality wise ToStruct and ToStructLenient works same with just one difference of relaxing restriction on matching row's column with struct exported field.
Please help review again.

spanner/row.go
```diff
@@ -311,5 +312,43 @@ func (r *Row) ToStruct(p interface{}) error {
...
+//
+// The fields of the destination struct can be of any type that is acceptable
+// to spanner.Row.Column. No errors will be returned if the row's column is not
+// mapped to a struct's exported field or destination struct has extra exported fields,
+// unmatched struct's exported fields will have default values.
+//
+// Slice and pointer fields will be set to nil if the source column is NULL, and a
+// non-nil value if the column is not NULL. To decode NULL values of other types, use
+// one of the spanner.NullXXX types as the type of the destination field.
+func (r *Row) ToStructLenient(p interface{}) error {
```
olavloite, line 339:
```suggestion
// If ToStruct returns an error, the contents of p are undefined. Some fields may
// have been successfully populated, while others were not; you should not use any of
// the fields.
func (r *Row) ToStructLenient(p interface{}) error {
```

This warning is still true for this method. Although this method is less likely to return an error, it can still fail if for example the type of one of the fields is not compatible with the datatype of the corresponding column.

spanner/row.go
```diff
@@ -311,5 +312,43 @@ func (r *Row) ToStruct(p interface{}) error {
...
+// The rules for mapping a row's columns into a struct's exported fields
+// are:
+//
+//   1. If a field has a `spanner: "column_name"` tag, then decode column
+//      'column_name' into the field. A special case is the `spanner: "-"`
+//      tag, which instructs ToStruct to ignore the field during decoding.
+//
+//   2. Otherwise, if the name of a field matches the name of a column (ignoring case),
+//      decode the column into the field.
+//
```
olavloite, line 329:
```suggestion
//
// 3. The number of columns in the row and exported fields in the struct do not need to match.
// Any field in the struct that cannot not be assigned a value from the row is assigned its default value.
// Any column in the row that does not have a corresponding field in the struct is ignored.
```

spanner/value.go
```diff
@@ -2986,7 +3003,7 @@ func decodeStructArray(ty *sppb.StructType, pb *proto3.ListValue, ptr interface{
 			return errDecodeArrayElement(i, pv, "STRUCT", err)
 		}
 		// Decode proto3.ListValue l into struct referenced by s.Interface().
-		if err = decodeStruct(ty, l, s.Interface()); err != nil {
+		if err = decodeStruct(ty, l, s.Interface(), false); err != nil {
```
olavloite, line 3006: Shouldn't this be:
```suggestion
		if err = decodeStruct(ty, l, s.Interface(), lenient); err != nil {
```

spanner/value.go
```diff
@@ -2921,13 +2926,25 @@ func decodeStruct(ty *sppb.StructType, pb *proto3.ListValue, ptr interface{}) er
 	if err != nil {
 		return ToSpannerError(err)
 	}
+	// return error if linent is true and destination has duplicate exported columns
```
olavloite, line 2929: nit:
```suggestion
	// return error if lenient is true and destination has duplicate exported columns
```

spanner/value.go
rahul2393: No, this is called from decodeStructArray and we don't want to change that function.

spanner/row.go
```diff
@@ -295,7 +295,8 @@ func errToStructArgType(p interface{}) error {
 // non-nil value if the column is not NULL. To decode NULL values of other types, use
 // one of the spanner.NullXXX types as the type of the destination field.
 //
-// If ToStruct returns an error, the contents of p are undefined. Some fields may
+// If ToStruct returns an error, the contents of p are undefined or if the there is un-matching
+// column from row's columns into struct's exported fields. Some fields may
```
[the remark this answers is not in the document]
rahul2393: > if an exported field in the struct cannot not be assigned a value from the row.
I think this is not the case with ToStruct now, it won't return error.

spanner/value.go
olavloite: Ah, yeah, good point. I missed that.

olavloite (approved):

spanner/value.go
```diff
@@ -2921,13 +2926,25 @@ func decodeStruct(ty *sppb.StructType, pb *proto3.ListValue, ptr interface{}) er
 	if err != nil {
 		return ToSpannerError(err)
 	}
+	// return error if lenient is true and destination has duplicate exported columns
```
hengfengli, line 2929: nit: return -> Return

hengfengli (approved):

PR #5170, opened 2021-11-26
feat(storage): add retry config to BucketHandle

BrennaEpp:
* Adds Retryer configurability to BucketHandle
* Adds retrying to BucketHandle.Update, including defaulting to retry only when idempotency conditions are present
* Adds retry config to all direct methods on BucketHandle
* Adds integration test for retry configs

Bucket config will merge with object config, with the object's config overriding the options it sets.


storage/bucket.go
```diff
@@ -1367,6 +1376,19 @@ func (b *BucketHandle) Objects(ctx context.Context, q *Query) *ObjectIterator {
 	return it
 }
 
+// Retryer returns a bucket handle that is configured with custom retry
+// behavior as specified by the options that are passed to it. All operations
+// on the new handle will use the customized retry configuration.
+func (b *BucketHandle) Retryer(opts ...RetryOption) *BucketHandle {
```
tritone, line 1382: This looks like the right surface.

I realized we probably also need to account for resolving the options between BucketHandle and ObjectHandle-- for example, if the user calls `client.Bucket(...).Retryer(...).Object(...)`, I guess the retry config should be inherited from the bucket. We should probably discuss what the resolution strategy should look like.

storage/bucket.go
```diff
@@ -210,12 +211,20 @@ func (b *BucketHandle) Update(ctx context.Context, uattrs BucketAttrsToUpdate) (
 	if uattrs.PredefinedDefaultObjectACL != "" {
 		req.PredefinedDefaultObjectAcl(uattrs.PredefinedDefaultObjectACL)
 	}
-	// TODO(jba): retry iff metagen is set?
-	rb, err := req.Context(ctx).Do()
-	if err != nil {
+
+	isIdempotent := b.conds != nil && b.conds.MetagenerationMatch != 0
```
tritone, line 215: This looks good good for BucketHandle.Update. Could you add the logic to use the retry config to the other direct methods on `BucketHandle` as well? It will be confusing to have it work only on some methods in the mean time.

BrennaEpp: Added!

storage/storage.go
```diff
@@ -1897,6 +1904,29 @@ type retryConfig struct {
 	shouldRetry func(err error) bool
 }
 
+func (r *retryConfig) clone() *retryConfig {
+	if r == nil {
+		return nil
+	}
+
+	var bo *gax.Backoff
+	if r.backoff == nil {
```
tritone, line 1913: Only the `!= nil` case is necessary since `bo` is nil by default.

storage/integration_test.go
```diff
@@ -4448,6 +4449,147 @@ func findTestCredentials(ctx context.Context, envVar string, scopes ...string) (
 	return transport.Creds(ctx, opts...)
 }
 
+func TestIntegration_RetryConfig(t *testing.T) {
```
tritone, line 4452: I feel like this test is a bit hacky since it depends on an unrealistic shouldRetry method. Do we really need an integration test? I feel like it's probably sufficient to just to write a unit test that verifies that the configs are what we expect them to be, if options are set on either or both of the BucketHandle and ObjectHandle.

storage/storage.go
```diff
@@ -1790,9 +1790,16 @@ func setConditionField(call reflect.Value, name string, value interface{}) bool
 // Retryer returns an object handle that is configured with custom retry
 // behavior as specified by the options that are passed to it. All operations
 // on the new handle will use the customized retry configuration.
+// These retry options will merge with the bucket's retryer (if set) for the
+// returned handle. Options passed into this method will take precedence over
+// options on the bucket's retryer.
 func (o *ObjectHandle) Retryer(opts ...RetryOption) *ObjectHandle {
 	o2 := *o
 	retry := &retryConfig{}
```
tritone, line 1798: Let's only assign this in the `o.retry == nil` case.

```
var retry *retryConfig
if o.retry != nil {
  retry = o.retry
} else {
  retry = &retryConfig{}
}
```

storage/integration_test.go
BrennaEpp: good point. switched it to unit test

tritone (approved):

PR #5171, opened 2021-11-26
feat(storage): retry objects.insert on writer open

BrennaEpp:
* Since the internals automatically retry some errors without checking for idempotency, one of the retry conformance tests fails for objects.insert


tritone (changes requested):
This is not going to accomplish anything unfortunately. I'm working on a PR in the Apiary library which will allow retry options to  be passed into the call. Until that is present, there's nothing we can do at the manual client layer to provide idempotency.

PR #5172, opened 2021-11-26
feat(storage): retry copy funcs on idempotent conds

BrennaEpp:
retry `objects.compose` and `objects.rewrite` on idempotent conditions


tritone (commented):
@frankyn could you double check the idempotency conditions on these? Just wanted to confirm that checking these conditions specifically on the destination object is correct for rewrite and compose.

tritone (approved):
Actually yes confirmed that this is correct-- see googleapis/python-storage#454 for the python equivalent.

PR #5173, opened 2021-11-26
feat(spanner/spannertest): support JSON_VALUE function

### LINKED ISSUE #5167: spanner(spansql): DDL Parser errors on JSON_VALUE syntax
**Client**

Spanner v1.27.0

**Environment**

Alpine Linux (also locally on Mac OS)

**Go Environment**

Observed on go 1.14 and 1.16

**Code**

e.g.
```go
ddlText := "ALTER TABLE products ADD COLUMN item STRING(MAX) AS (JSON_VALUE(itemDetails, '$.itemDetails')) STORED;"
parsed, err := spansql.ParseDDL("", ddlText)
if err != nil {
     // err is non nil
}
```

**Expected behavior**

No errors

**Actual behavior**

`error: :3: got "(" while expecting ")"`

Whereas running the DDL in the GCP console works without issue.

**Screenshots**

N/A

**Additional context**

DDL is 1:1 with the example here https://cloud.google.com/spanner/docs/working-with-json#indexing

olavloite:
Adds support for the JSON_VALUE function. The function always returns an empty string,
as there is no XPath query engine available in the Spanner client library. This does
however enable the usage of computed columns that use the function.

This change also fixes the TODO that generated columns cannot be added to non-empty
tables.

Fixes #5167

### DIFF
--- a/spanner/spannertest/db.go
+++ b/spanner/spannertest/db.go
@@ -700,12 +700,20 @@ func (t *table) addColumn(cd spansql.ColumnDef, newTable bool) *status.Status {
 			// TODO: what happens in this case?
 			return status.Newf(codes.Unimplemented, "can't add NOT NULL columns to non-empty tables yet")
 		}
-		if cd.Generated != nil {
-			// TODO: should backfill the data to maintain behaviour with real spanner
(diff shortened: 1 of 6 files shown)


rahul2393 (approved):
LGTM

hengfengli (approved):

Issue #5174, opened 2021-11-26
compute: unable to resize instance group to 0


matihost:
https://github.com/googleapis/google-cloud-go/blob/main/compute/apiv1/instance_group_managers_client.go#L1301

When size is 0, the size parameter is not added to HTTP param - and server returns 400 bad request - leading to unability to resize intance group to 0.

The gcloud cli works:
```
gcloud compute instance-groups managed resize NAME  --zone=ZONE --size=0
```

codyoss:
cc @noahdietz WDYT? A bug perhaps?

noahdietz:
Opened https://github.com/googleapis/gapic-generator-go/issues/819 to track the bug fix.

noahdietz:
https://pkg.go.dev/cloud.google.com/go@v0.99.0 should have the fix https://github.com/googleapis/google-cloud-go/blob/5563382a8560ea851e6d54962ea71165b2807b24/compute/apiv1/instance_group_managers_client.go#L1297-L1302

PR #5181, opened 2021-12-02
chore(internal/gapicgen): update microgen to v0.23.5

noahdietz:
In REST clients, singular, primitive required fields will be added to query params regardless of if they are default empty value or not.

This is the fix for #5174 which will come in the next regen.


codyoss:
LGTM, I will land this later after I unstick the current regen

codyoss (approved):

PR #5185, opened 2021-12-03
feat(storage): add retry config to ACL handle methods

BrennaEpp:
* Adds retryConfig to ACLs when set on a bucket or object handle


storage/storage_test.go
```diff
@@ -1007,17 +1007,47 @@ func TestRetryer(t *testing.T) {
...
+					want: b.retry,
+				},
+				{
+					name: "bucket.DefaultObjectACL()",
+					r:    b.DefaultObjectACL().retry,
+					want: b.retry,
+				},
+			}
+			for _, ac := range configHandleCases {
+				s.Run(ac.name, func(ss *testing.T) {
```
tritone, line 1037: Didn't realize these could be nested... nice!

storage/acl.go
```diff
@@ -174,12 +175,12 @@ func (a *ACLHandle) bucketDelete(ctx context.Context, entity ACLEntity) error {
...
 	var acls *raw.ObjectAccessControls
 	var err error
-	err = runWithRetry(ctx, func() error {
+	err = run(ctx, func() error {
 		req := a.c.raw.ObjectAccessControls.List(a.bucket, a.object)
 		a.configureCall(ctx, req)
 		acls, err = req.Do()
 		return err
-	})
+	}, a.retry, true)
```
tritone, line 183: You are missing a bunch of calls in this file (mostly stuff that is not retried currently)-- could you add `run()` wrappings for those as well?

tritone (commented):
Generally looks good, just need to finish extending the retry wrapper to the remaining ACLHandle methods.

storage/acl.go
BrennaEpp: Whoops - that brings to mind, should/can we test to check that all calls use the retry wrapping? Perhaps using the same list of functions as the retry conformance tests?

tritone: Hmm, this is a good question. I'm not sure of a great way to test this! Let's think about it though (and we should definitely do an audit after the work is complete).

tritone (approved):

PR #5193, opened 2021-12-06
feat(storage): add retry config to the Client and HmacKey operations

BrennaEpp:


storage/storage.go
```diff
@@ -1810,6 +1812,27 @@ func (o *ObjectHandle) Retryer(opts ...RetryOption) *ObjectHandle {
 	return &o2
 }
 
+// Retryer returns an Client that is configured with custom retry behavior as
+// specified by the options that are passed to it. All operations on the new
+// handle will use the customized retry configuration.
+// Retry options set on a bucket or object handle will take precedence over
+// these options.
```
tritone, line 1819: Maybe note that this should be configured before calling any network ops on the client. Otherwise there could be indeterminate behavior with operations in progress.

storage/storage.go
```diff
@@ -1810,6 +1812,27 @@ func (o *ObjectHandle) Retryer(opts ...RetryOption) *ObjectHandle {
...
+func (c *Client) Retryer(opts ...RetryOption) *Client {
```
tritone, line 1820: Let's not have this return a new client, but instead configure the existing one. We can call it `SetRetry` maybe.

I used that pattern on BucketHandle/ObjectHandle because there were existing methods that used the same pattern (e.g. `If`, `Key`, `Generation`). But client methods generally don't follow the same pattern. And we want users to set this once before using the client if they plan on using it, not setting it using chained calls while calling a method.

tritone (commented):
Let me know if you'd like to discuss more about what surface makes sense here.

storage/storage.go
BrennaEpp: Good note

storage/hmac.go
```diff
@@ -257,10 +263,10 @@ func (h *HMACKeyHandle) Update(ctx context.Context, au HMACKeyAttrsToUpdate, opt
 
 	var metadata *raw.HmacKeyMetadata
 	var err error
-	err = runWithRetry(ctx, func() error {
+	err = run(ctx, func() error {
 		metadata, err = call.Context(ctx).Do()
 		return err
-	})
+	}, h.retry, true)
```
tritone, line 269: This should be conditionally idempotent based on whether the Etag is present in the metadata-- the user may pass it in HMACKeyAttrsToUpdate or not.

Are conformance tests catching this?

storage/storage_test.go
```diff
@@ -875,11 +875,102 @@ func TestObjectRetryer(t *testing.T) {
 	}
 }
 
-// Test the interactions between ObjectHandle and BucketHandle Retryers and that
-// they correctly configure the retry configuration for objects and ACLs
+// Test that Client.Retryer correctly configures the retry configuration
```
tritone, line 878: Should be Client.SetRetry here and in the test name.

storage/hmac.go
BrennaEpp: It should be good now

storage/retry_conformance_test.go
```diff
@@ -252,12 +252,18 @@ var methods = map[string][]retryFunc{
 	"storage.hmacKey.update": {
 		func(ctx context.Context, c *Client, fs *resources, preconditions bool) error {
 			key := c.HMACKeyHandle(projectID, fs.hmacKey.AccessID)
+			uattrs := HMACKeyAttrsToUpdate{State: "INACTIVE"}
 
-			_, err := key.Update(ctx, HMACKeyAttrsToUpdate{State: "INACTIVE"})
-			if err != nil {
-				return err
+			if preconditions {
+				k, err := key.Get(ctx)
```
tritone, line 258: I think you can get Etag from `fs.hmacKey` rather than having another `Get`

tritone (approved):
One more nit, otherwise LGTM

storage/retry_conformance_test.go
BrennaEpp: good note!

tritone (approved):

PR #5199, opened 2021-12-07
feat(spanner): Adding GFE Latency and Header Missing Count Metrics

asthamohta:
Benchmarking:
Each function was performed 100 times in a test and the time for it is noted . This is averaged out across 50 runs. Below is the summarisation of the numbers:. Following are the results:

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-cddf1b73-7fff-2115-e458-c98f328fd63d"><div dir="ltr" style="margin-left:0pt;" align="left">

Functionality | ReadRow | InsertOrUpdate | Query
-- | -- | -- | --
Time without adding GFE Latency Feature | 18.47518476 | 23.79237414 | 17.92703158
Time with GFE Latency added and Disabled | 18.84231938 | 21.38502053 | 23.93358834
Time with GFE Latency added and Enabled | 19.45244371 | 22.03212286 | 21.28579487

</div><br /></b>

For full details of each run, check [here](https://docs.google.com/spreadsheets/d/1igkw5VGK8jRWRLGg-CUgRlytUEo6xSXAFH8SM2_bLYI/edit#gid=0)


spanner/oc_test.go
```diff
@@ -261,6 +261,57 @@ func TestOCStats_SessionPool_GetSessionTimeoutsCount(t *testing.T) {
...
+
+	_, client, teardown := setupMockedTestServer(t)
+	defer teardown()
+
+	client.Single().ReadRow(context.Background(), "Users", Key{"alice"}, []string{"email"})
+
+	waitErr := &Error{}
+	waitFor(t, func() error {
+		select {
+		case stat := <-te.Stats:
```
olavloite, line 278: This is where the test case is waiting indefinitely, as there will never be any stats that are exported. That is fixed by adding the additional view at the top of the test. However, the stats that will be exported will at this moment be a `GFEHeaderMissingCount` and not a `GFELatency` value.

spanner/sessionclient.go
```diff
@@ -230,11 +248,29 @@ func (sc *sessionClient) executeBatchCreateSessions(client *vkit.Client, createC
...
 		}
+		var mdForGFELatency metadata.MD
 		response, err := client.BatchCreateSessions(ctx, &sppb.BatchCreateSessionsRequest{
 			SessionCount:    remainingCreateCount,
 			Database:        sc.database,
 			SessionTemplate: &sppb.Session{Labels: labels},
-		})
+		}, gax.WithGRPCOptions(grpc.Header(&mdForGFELatency)))
+
+		if GFELatencyOrHeaderMissingCountEnabled && md != nil{
```
olavloite, line 258: This if statement should check that `mdForGFELatency != nil`:

```suggestion
		if GFELatencyOrHeaderMissingCountEnabled && mdForGFELatency != nil{
```

spanner/sessionclient.go
```diff
@@ -230,11 +248,29 @@ func (sc *sessionClient) executeBatchCreateSessions(client *vkit.Client, createC
...
+				break
+			}
+			// Errors should not prevent initializing the session pool.
+			ctxGFE, err := tag.New(ctx,
+				tag.Upsert(tagKeyClientID, sc.id),
+				tag.Upsert(tagKeyDatabase, database),
+				tag.Upsert(tagKeyInstance, instance),
+				tag.Upsert(tagKeyLibVersion, version.Repo),
+			)
+			captureGFELatencyStats(ctxGFE, md, "executeBatchCreateSessions")
```
olavloite, line 272:
```suggestion
			captureGFELatencyStats(ctxGFE, mdForGFELatency, "executeBatchCreateSessions")
```

olavloite (commented):
You also need to change the mock server so it actually returns a server-timing header. To do this, add the following lines to just before the return statement of the `BatchCreateSessions` method in `internal/testutil/inmem_spanner_server.go` ([before line 690](https://github.com/googleapis/google-cloud-go/blob/828a9d24c482854281a829ae6c960e6d953926d9/spanner/internal/testutil/inmem_spanner_server.go#L690)):

```golang
	header := metadata.New(map[string]string{"server-timing": "123"})
	if err := grpc.SendHeader(ctx, header); err != nil {
		return nil, gstatus.Errorf(codes.Internal, "unable to send 'server-timing' header")
	}
```

(And this also needs to be added to all the other RPCs that should also return a server timing value)

spanner/integration_test.go
```diff
@@ -3264,6 +3266,85 @@ func TestIntegration_DirectPathFallback(t *testing.T) {
 	}
 }
 
+func TestIntegration_GFE_Latency(t *testing.T) {
+	t.Parallel()
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
+
+	te := testutil.NewTestExporter(GFEHeaderMissingCountView, GFELatencyView)
+	GFELatencyOrHeaderMissingCountEnabled = true
```
olavloite, line 3275: This is the variable that users must set to `true` to enable this feature, right? If so, I think we should rename the variable to simply `GFELatencyMetricsEnabled`. The fact that we also count the number of times that the header is missing, is a technical detail that users do not need to know about.

spanner/integration_test.go
```diff
@@ -3264,6 +3266,85 @@ func TestIntegration_DirectPathFallback(t *testing.T) {
...
+				case *view.LastValueData:
+					data = fmt.Sprintf("%v", row.Data.(*view.LastValueData).Value)
+				case *view.DistributionData:
+					data = fmt.Sprintf("%v", row.Data.(*view.DistributionData).Count)
+				}
+				if got, want := fmt.Sprintf("%v", data), "0"; got <= want {
+					t.Fatalf("Incorrect data: got %v, wanted more than %v for metric %v", got, want, stat.View.Measure.Name())
+				}
+			}
+		case <-time.After(2 * time.Second):
```
olavloite, line 3339: The test is green, but 2 seconds seem to me like a very optimistic timeout value that might cause a flaky failure now and then. Or is it really reasonable to assume that the stats will always be available within 2 seconds?

spanner/oc_test.go
```diff
@@ -22,6 +22,10 @@ import (
 	"testing"
 	"time"
 
+	structpb "github.com/golang/protobuf/ptypes/struct"
+	"google.golang.org/api/iterator"
+	spannerpb "google.golang.org/genproto/googleapis/spanner/v1"
+
```
olavloite, line 28: nit: merge with import block below

spanner/sessionclient.go
```diff
@@ -24,6 +24,9 @@ import (
 	"sync"
 	"time"
 
+	"go.opencensus.io/tag"
+	"google.golang.org/grpc"
```
olavloite, line 28: nit: merge with block below

spanner/stats.go
```diff
@@ -16,6 +16,12 @@ package spanner
 
 import (
 	"context"
+	"strconv"
+	"strings"
+	"testing"
+
+	"cloud.google.com/go/internal/version"
```
olavloite, line 23: nit: merge with block below

spanner/transaction.go
```diff
@@ -22,6 +22,12 @@ import (
 	"sync/atomic"
 	"time"
 
+	"cloud.google.com/go/internal/version"
+	"github.com/googleapis/gax-go/v2"
+	"go.opencensus.io/tag"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
```
olavloite, line 29: nit: merge with block below

spanner/transaction.go
```diff
@@ -75,6 +81,28 @@ type txReadOnly struct {
...
+	database string
+	// Instance ID
+	instance string
+	// Library Version
+	libVersion string
+}
+
+func (ct *CommonTags) init() {
+	// setting default value of set as false
+	ct.set = false
```
olavloite, line 105: The initial value of `ct.set` is already false when you create a `CommonTags`.

spanner/integration_test.go
asthamohta: Makes sense. Changing it to 10 seconds

spanner/batch.go
```diff
@@ -251,12 +290,30 @@ func (t *BatchReadOnlyTransaction) Cleanup(ctx context.Context) {
...
 	sid, client := sh.getID(), sh.getClient()
-	err := client.DeleteSession(contextWithOutgoingMetadata(ctx, sh.getMetadata()), &sppb.DeleteSessionRequest{Name: sid})
-	if err != nil {
-		var logger *log.Logger
-		if sh.session != nil {
-			logger = sh.session.logger
+	var logger *log.Logger
+	if sh.session != nil {
+		logger = sh.session.logger
+	}
```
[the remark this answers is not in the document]
asthamohta: logger is used in the GFELatency block as well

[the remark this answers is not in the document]
olavloite: Ah, good point. I missed that.

spanner/batch.go
```diff
@@ -136,7 +140,18 @@ func (t *BatchReadOnlyTransaction) PartitionReadUsingIndexWithOptions(ctx contex
...
+	if GFELatencyMetricsEnabled && md != nil && t.txReadOnly.CommonTags != nil {
+		ctxGFE, errGFE := createContextForGFELatencyMetrics(ctx, t.txReadOnly)
+		if errGFE != nil {
+			return nil, errGFE
+		}
+		errGFE = captureGFELatencyStats(ctxGFE, md, "PartitionReadUsingIndexWithOptions")
+		if errGFE != nil {
+			return nil, errGFE
+		}
+	}
```
olavloite, line 154: This block of code (with some small variations) seem to be duplicated very many times in this PR. Would it be possible to extract that to a separate method with the RPC name as an argument and that returns an error (and possibly other arguments that I'm missing here now)?

spanner/batch.go
```diff
@@ -291,10 +332,27 @@ func (t *BatchReadOnlyTransaction) Execute(ctx context.Context, p *Partition) *R
 				RequestOptions: p.rreq.RequestOptions,
 				ResumeToken:    resumeToken,
 			})
+			if client != nil {
```
olavloite, line 335: I think that the following is more idiomatic, and it also reduces nesting:

```golang
if err != nil` {
    return client, err
}
```

spanner/batch.go
```diff
@@ -305,6 +363,24 @@ func (t *BatchReadOnlyTransaction) Execute(ctx context.Context, p *Partition) *R
 				RequestOptions: p.qreq.RequestOptions,
 				ResumeToken:    resumeToken,
 			})
+			if client != nil {
```
olavloite, line 366: Same as above (replace with `if err != nil`)

spanner/client.go
```diff
@@ -280,6 +281,17 @@ func (c *Client) Single() *ReadOnlyTransaction {
 		t.sh = sh
 		return nil
 	}
+	if c.sc != nil {
+		_, instance, database, err := parseDatabaseName(c.sc.database)
```
olavloite, line 285: I don't think that we should parse this information every time a read-only transaction (or any other transaction) is requested from the client. Instead, it should be possible to just store the `CommonTags` in the `Client` (or the `sessionClient`) and assign that to each transaction that is created. The information in the `CommonTags` won't change during the lifetime of a client.

spanner/internal/testutil/inmem_spanner_server.go
```diff
@@ -25,6 +25,9 @@ import (
 	"sync"
 	"time"
 
+	"google.golang.org/grpc"
```
olavloite, line 28: nit: merge with block below

spanner/transaction.go
```diff
@@ -184,6 +203,23 @@ func (t *txReadOnly) ReadWithOptions(ctx context.Context, table string, keys Key
 					Limit:          int64(limit),
 					RequestOptions: createRequestOptions(prio, requestTag, t.txOpts.TransactionTag),
 				})
+			if client != nil {
```
olavloite, line 206: replace with `if err != nil { return client, err }`

spanner/transaction.go
```diff
@@ -373,7 +409,24 @@ func (t *txReadOnly) query(ctx context.Context, statement Statement, options Que
 		func(ctx context.Context, resumeToken []byte) (streamingReceiver, error) {
 			req.ResumeToken = resumeToken
 			req.Session = t.sh.getID()
-			return client.ExecuteStreamingSql(ctx, req)
+			client, err := client.ExecuteStreamingSql(ctx, req)
```
olavloite, line 412: Replace with `if err != nil { return client, err }`

spanner/batch.go
asthamohta: You mean I check if err==nil then only go into the GFE Latency block?

olavloite: No, the idiomatic thing to do in Go is to return early if an error is returned by a function that is called. See https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

Also, normally only one of `client` and `err` should be non-nil at this point. An idiomatic Go function should normally not return both a 'normal' return value and an error. So that means that if `err != nil` is true, then `client == nil` should also be true.

spanner/client.go
```diff
@@ -280,6 +280,9 @@ func (c *Client) Single() *ReadOnlyTransaction {
 		t.sh = sh
 		return nil
 	}
+	if c.sc != nil && commonTags.clientID == "" {
```
olavloite, line 283: Sorry, I think I was a little unclear on what I meant with where the `commonTags` should go:

1. `commonTags` will never change for a given `sessionClient` or `Client`. So one of the `sessionClient` or `Client` structs should therefore have a field `commonTags` (or just the values that are needed for common tags). `commonTags` should not be a global variable. The current implementation means that `commonTags` will be filled with the values of the first client that executes a transaction. If the application opens a second client for a different database, `commonTags` will continue to contain the values for the first client.
2. The initialization of `commonTags` should not happen when a transaction is created. It should be done when the `Client` / `sessionClient` is created, as it will never change after that. So in this case somewhere around line 194 in this file. That means that these checks and `setCommonTags(..)` calls can be removed from all the transaction methods.

spanner/batch.go
```diff
@@ -291,10 +320,24 @@ func (t *BatchReadOnlyTransaction) Execute(ctx context.Context, p *Partition) *R
 				RequestOptions: p.rreq.RequestOptions,
 				ResumeToken:    resumeToken,
 			})
+			if err != nil {
+				return client, err
+			}
+			md, errGFE := client.Header()
```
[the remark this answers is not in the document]
asthamohta: Hi Knut, the reason why I did this was because in someplaces I was worries that the error was getting returned at the end of the function, example [here ](https://github.com/googleapis/google-cloud-go/blob/5a2ed6b2cd1c304e0f59daa29959863bff9b5c29/spanner/batch.go#L131). So will have to use a separate error in such places. Hence to seem consistent I used errGFE

spanner/batch.go
```diff
@@ -136,7 +140,13 @@ func (t *BatchReadOnlyTransaction) PartitionReadUsingIndexWithOptions(ctx contex
 		Columns:          columns,
 		KeySet:           kset,
 		PartitionOptions: opt.toProto(),
-	})
+	}, gax.WithGRPCOptions(grpc.Header(&md)))
+
+	if GFELatencyMetricsEnabled && md != nil && t.ct != nil {
+		if errGFE := createContextAndCaptureGFELatencyMetrics(ctx, t.ct, md, "PartitionReadUsingIndexWithOptions"); errGFE != nil {
```
olavloite, line 146: nit: I would prefer `err` instead of `errGFE` here and elsewhere whenever possible. I think that sticking to the Golang default whenever possible is better than deviating from it in all cases to be consistent with the places where it is not possible and you have to use a different name.

spanner/batch.go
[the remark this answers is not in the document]
olavloite: Yeah, I understand that. But I think that it is better to follow the standard naming of Golang when possible, and only make an exception when you need to. In this case for example, you no longer need to return `err` at the end of the function anymore, as you know that it is `nil` because of the `if` statement that was added above. So the last statement in the function can now be changed to `return client, nil`

spanner/client.go
```diff
@@ -280,6 +281,10 @@ func (c *Client) Single() *ReadOnlyTransaction {
 		t.sh = sh
 		return nil
 	}
+	if c.sc != nil && c.ct == nil {
+		setCommonTags(c)
+	}
```
olavloite, line 286: This is not necessary here (and also not in the other methods that start a transaction). Instead, this initialization should be done in the `NewClientWithConfig` method, right after the `sessionClient` has been created and before returning the new client.

spanner/batch.go
```diff
@@ -136,7 +140,13 @@ func (t *BatchReadOnlyTransaction) PartitionReadUsingIndexWithOptions(ctx contex
 		Columns:          columns,
 		KeySet:           kset,
 		PartitionOptions: opt.toProto(),
-	})
+	}, gax.WithGRPCOptions(grpc.Header(&md)))
+
```
olavloite, line 144: Here and elsewhere: I think it is better to add an `if err != nil` { return err }` statement before checking for the latency values. I worry that we otherwise might start returning errors that are harder to interpret for users if the following happens:
1. The `client.PartitionRead` returns an error. That error is not checked here at the moment.
2. The RPC did return a header (so `md != nil`), but that header does not contain a GFE latency value. So instead of the error that was returned from `PartitionedRead`, we return an error that indicates that it could not find a GFE Latency value.

I worry that the above would be confusing for users, and make it harder to debug errors that occur because of for example invalid requests.

spanner/batch.go
```diff
@@ -190,7 +201,13 @@ func (t *BatchReadOnlyTransaction) partitionQuery(ctx context.Context, statement
 		Params:           params,
 		ParamTypes:       paramTypes,
 	}
-	resp, err := client.PartitionQuery(contextWithOutgoingMetadata(ctx, sh.getMetadata()), req)
+	resp, err := client.PartitionQuery(contextWithOutgoingMetadata(ctx, sh.getMetadata()), req, gax.WithGRPCOptions(grpc.Header(&md)))
+
```
olavloite, line 205: Here also: Check the error from the previous call before trying to extract the metadata

spanner/pdml.go
```diff
@@ -114,7 +118,13 @@ func executePdml(ctx context.Context, sh *sessionHandle, req *sppb.ExecuteSqlReq
 	req.Transaction = &sppb.TransactionSelector{
 		Selector: &sppb.TransactionSelector_Id{Id: res.Id},
 	}
-	resultSet, err := sh.getClient().ExecuteSql(contextWithOutgoingMetadata(ctx, sh.getMetadata()), req)
+	resultSet, err := sh.getClient().ExecuteSql(contextWithOutgoingMetadata(ctx, sh.getMetadata()), req, gax.WithGRPCOptions(grpc.Header(&md)))
+	if GFELatencyMetricsEnabled && md != nil && sh.session.pool != nil {
```
olavloite, line 122: Also here: Check the error from the `ExecuteSql` method before trying to extract the metadata to prevent us from returning a potentially confusing error to the user.

spanner/sessionclient.go
```diff
@@ -130,10 +132,32 @@ func (sc *sessionClient) createSession(ctx context.Context) (*session, error) {
 		return nil, err
 	}
 	ctx = contextWithOutgoingMetadata(ctx, sc.md)
+	var md metadata.MD
 	sid, err := client.CreateSession(ctx, &sppb.CreateSessionRequest{
 		Database: sc.database,
 		Session:  &sppb.Session{Labels: sc.sessionLabels},
-	})
+	}, gax.WithGRPCOptions(grpc.Header(&md)))
+
```
olavloite, line 140: Check the error from `CreateSession` and return that if it is non-nil before trying to extract the metadata.

spanner/sessionclient.go
```diff
@@ -230,11 +254,39 @@ func (sc *sessionClient) executeBatchCreateSessions(client *vkit.Client, createC
...
 			break
 		}
+		var mdForGFELatency metadata.MD
 		response, err := client.BatchCreateSessions(ctx, &sppb.BatchCreateSessionsRequest{
 			SessionCount:    remainingCreateCount,
 			Database:        sc.database,
 			SessionTemplate: &sppb.Session{Labels: labels},
-		})
+		}, gax.WithGRPCOptions(grpc.Header(&mdForGFELatency)))
+
```
olavloite, line 263: Also here: Check the error from the `BatchCreateSessions` call before trying to extract the metadata. That also means that we don't need consider the session creation to have failed if extracting the GFE latency fails (see below)

spanner/sessionclient.go
```diff
@@ -230,11 +254,39 @@ func (sc *sessionClient) executeBatchCreateSessions(client *vkit.Client, createC
...
+			ctxGFE, errGFE := tag.New(ctx,
+				tag.Upsert(tagKeyClientID, sc.id),
+				tag.Upsert(tagKeyDatabase, database),
+				tag.Upsert(tagKeyInstance, instance),
+				tag.Upsert(tagKeyLibVersion, version.Repo),
+			)
+			if errGFE != nil {
+				trace.TracePrintf(ctx, nil, "Error in adding tags in BatchCreateSessions for GFE Latency: %v", errGFE)
+				consumer.sessionCreationFailed(ToSpannerError(errGFE), remainingCreateCount)
+				break
```
olavloite, line 281: This can be removed if you move the error check for the session creation to above this entire block, as it means that we at this point will know that the session creation succeeded and only the extraction of the latency failed. Logging the error from that should be enough, we don't need to report to the client that the creation itself failed (as it didn't in this case).

spanner/sessionclient.go
```diff
@@ -230,11 +254,39 @@ func (sc *sessionClient) executeBatchCreateSessions(client *vkit.Client, createC
...
+			}
+			errGFE = captureGFELatencyStats(ctxGFE, mdForGFELatency, "executeBatchCreateSessions")
+			if errGFE != nil {
+				trace.TracePrintf(ctx, nil, "Error in Capturing GFE Latency and Header Missing count. Try disabling and rerunning. Error: : %v", errGFE)
+				consumer.sessionCreationFailed(ToSpannerError(errGFE), remainingCreateCount)
+				break
```
olavloite, line 287: Same as above; this can also be removed if the error check for session creation is moved to before this entire block.

spanner/transaction.go
```diff
@@ -184,6 +190,19 @@ func (t *txReadOnly) ReadWithOptions(ctx context.Context, table string, keys Key
 					Limit:          int64(limit),
 					RequestOptions: createRequestOptions(prio, requestTag, t.txOpts.TransactionTag),
 				})
+			if err != nil {
+				return client, err
+			}
+			md, errGFE := client.Header()
```
olavloite, line 196: nit: prefer `err` over `errGFE` when possible.

spanner/transaction.go
```diff
@@ -533,14 +565,22 @@ func (t *ReadOnlyTransaction) begin(ctx context.Context) error {
...
 			Session: sh.getID(),
 			Options: &sppb.TransactionOptions{
 				Mode: &sppb.TransactionOptions_ReadOnly_{
 					ReadOnly: buildTransactionOptionsReadOnly(t.getTimestampBound(), true),
 				},
 			},
-		})
+		}, gax.WithGRPCOptions(grpc.Header(&md)))
+
+		if GFELatencyMetricsEnabled && md != nil && t.ct != nil {
```
olavloite, line 578: Also here: Move the error handling code for the `BeginTransaction` call to before the block that tries to extract the GFE latency.

spanner/transaction.go
```diff
@@ -886,7 +926,14 @@ func (t *ReadWriteTransaction) update(ctx context.Context, stmt Statement, opts
 	if err != nil {
 		return 0, err
 	}
-	resultSet, err := sh.getClient().ExecuteSql(contextWithOutgoingMetadata(ctx, sh.getMetadata()), req)
+	var md metadata.MD
+	resultSet, err := sh.getClient().ExecuteSql(contextWithOutgoingMetadata(ctx, sh.getMetadata()), req, gax.WithGRPCOptions(grpc.Header(&md)))
+
```
olavloite, line 931: Also here: Move the error check for the `ExecuteSql` call here.

spanner/transaction.go
```diff
@@ -948,13 +995,20 @@ func (t *ReadWriteTransaction) batchUpdateWithOptions(ctx context.Context, stmts
...
+	var md metadata.MD
 	resp, err := sh.getClient().ExecuteBatchDml(contextWithOutgoingMetadata(ctx, sh.getMetadata()), &sppb.ExecuteBatchDmlRequest{
 		Session:        sh.getID(),
 		Transaction:    ts,
 		Statements:     sppbStmts,
 		Seqno:          atomic.AddInt64(&t.sequenceNumber, 1),
 		RequestOptions: createRequestOptions(opts.Priority, opts.RequestTag, t.txOpts.TransactionTag),
-	})
+	}, gax.WithGRPCOptions(grpc.Header(&md)))
+
```
olavloite, line 1006: Check the error returned by `ExecuteBatchDml` before trying to extract the latency

spanner/transaction.go
```diff
@@ -1210,6 +1264,10 @@ func NewReadWriteStmtBasedTransactionWithOptions(ctx context.Context, c *Client,
 	t.txReadOnly.txReadEnv = t
 	t.txReadOnly.qo = c.qo
 	t.txOpts = options
+	if c.sc != nil && c.ct == nil {
+		setCommonTags(c)
+	}
```
olavloite, line 1269: Remove, as this can be replaced by initialization code in `NewClientWithConfig` (see above)

spanner/batch.go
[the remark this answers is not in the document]
asthamohta: Actually I looked into it further and I realised if I define err inside the loop even in the case I described I can use it as the scope of that variable will be limited to the if condition loop. So i think it should be okay to do that. Wdyt?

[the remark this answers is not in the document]
olavloite: I'm not really sure I understand what you mean in this case. But if you mean that you can define `err` only inside the if-block, then that would be good.

spanner/batch.go
asthamohta: Actually if there is no GFE Latency value it won't throw an error. Instead it will add to the header missing count. It only throws an error if it got a garbage value in the value for the key value pair "server-timing". Also currently in the Java implementation, they intercept every rpc call(even when there is a failed operation) and use it to increase the header missing count. I feel if we don't add to the Header missing count in this case then across languages the header missing count might be different in someway

spanner/batch.go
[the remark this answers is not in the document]
asthamohta: Yes that's what I mean

spanner/sessionclient.go
asthamohta: Do you mean we can simply add a trace but don't need to actually break the fxn in case of GFE Latency error?

spanner/batch.go
olavloite: > I feel if we don't add to the Header missing count in this case then across languages the header missing count might be different in someway

That's a good point. So agreed to keep this as is on that.

I still worry a little bit that we might be returning an error from extracting the metadata instead of the error from the actual operation. I think we had a discussion about that earlier, and agreed on returning the error from those methods. I think I'm starting to change my mind on that. It is not an error that the user can do anything about, and it might just cover other potentially more important errors. So I would suggest then ignoring those errors. What do you think?

spanner/sessionclient.go
olavloite: Yes, I think that would be reasonable. It is not an error that the user can do anything with, and the session creation itself succeeded.

spanner/batch.go
asthamohta: Should we log these errors instead then? How do we let the user know that an error has occurred?

olavloite: If we have access to the logger, then logging it is probably a good idea. I don't think users can do anything with these errors anyways, so I don't think we need to actively inform them any more than that (and the chance that it happens should also be minimal, as it would mean that the data that is sent back is invalid)

asthamohta: As per our conversation going to add a trace instead of returning an error

spanner/batch.go
```diff
@@ -291,10 +318,23 @@ func (t *BatchReadOnlyTransaction) Execute(ctx context.Context, p *Partition) *R
 				RequestOptions: p.rreq.RequestOptions,
 				ResumeToken:    resumeToken,
 			})
+			if err != nil {
+				return client, err
+			}
+			md, err := client.Header()
+			if err != nil {
+				return nil, err
+			}
```
olavloite, line 327: nit: this early return is no longer needed now that we don't return an error if getting the latency stats fails

spanner/batch.go
```diff
@@ -305,6 +345,20 @@ func (t *BatchReadOnlyTransaction) Execute(ctx context.Context, p *Partition) *R
 				RequestOptions: p.qreq.RequestOptions,
 				ResumeToken:    resumeToken,
 			})
+			if err != nil {
+				return client, err
+			}
+			md, err := client.Header()
+			if err != nil {
+				return nil, err
+			}
```
olavloite, line 354: nit: this early return is no longer needed now that we don't return an error if getting the latency stats fails

spanner/sessionclient.go
```diff
@@ -130,10 +132,32 @@ func (sc *sessionClient) createSession(ctx context.Context) (*session, error) {
...
+			tag.Upsert(tagKeyLibVersion, version.Repo),
+		)
+		if err != nil {
+			trace.TracePrintf(ctx, nil, "Error in recording GFE Latency. Try disabling and rerunning. Error: %v", ToSpannerError(err))
+		}
+		err = captureGFELatencyStats(ctxGFE, md, "createSession")
+		if err != nil {
+			trace.TracePrintf(ctx, nil, "Error in recording GFE Latency. Try disabling and rerunning. Error: %v", ToSpannerError(err))
+		}
+
```
olavloite, line 159: nit: remove empty line

spanner/stats.go
```diff
@@ -167,3 +210,108 @@ func EnableStatViews() error {
...
+	ctx = tag.NewContext(ctx, tag.FromContext(ctx))
+	ctx, err = tag.New(ctx, tag.Insert(tagKeyMethod, keyMethod))
+	if err != nil {
+		return err
+	}
+	recordStat(ctx, GFELatency, int64(gfeLatency))
+	return nil
+}
+
+func checkCommonTagsGFELatency(t *testing.T, m map[tag.Key]string) {
```
olavloite, line 264: nit: consider moving this method to one of the `_test.go` files, as it seems to be a test method.

spanner/stats.go
```diff
@@ -167,3 +210,108 @@ func EnableStatViews() error {
...
+	return &CommonTags{
+		clientID:   sc.id,
+		database:   database,
+		instance:   instance,
+		libVersion: version.Repo,
+	}
+}
+
+// CommonTags are common key-value pairs of data associated with the GFELatency measure
+type CommonTags struct {
```
olavloite, line 308: nit: Could this struct be unexported (that is; `commonTags` instead of `CommonTags`)?

spanner/transaction.go
```diff
@@ -184,6 +190,19 @@ func (t *txReadOnly) ReadWithOptions(ctx context.Context, table string, keys Key
 					Limit:          int64(limit),
 					RequestOptions: createRequestOptions(prio, requestTag, t.txOpts.TransactionTag),
 				})
+			if err != nil {
+				return client, err
+			}
+			md, err := client.Header()
+			if err != nil {
+				return nil, err
+			}
```
olavloite, line 199: nit: this early return is no longer needed

spanner/transaction.go
```diff
@@ -373,7 +392,20 @@ func (t *txReadOnly) query(ctx context.Context, statement Statement, options Que
...
 			req.Session = t.sh.getID()
-			return client.ExecuteStreamingSql(ctx, req)
+			client, err := client.ExecuteStreamingSql(ctx, req)
+			if err != nil {
+				return client, err
+			}
+			md, err := client.Header()
+			if err != nil {
+				return nil, err
+			}
```
olavloite, line 402: nit: this early return is no longer needed

olavloite (approved):

spanner/batch.go
asthamohta: I think this is still needed as there might be an error in getting the Header, and in that case we return the error

olavloite: Yes, but that happens on line 333. The reason that we added the early return above in the first place, was that we would return the error that might be returned `createContextAndCaptureGFELatencyMetrics`. As we don't do that anymore, we can safely wait until line 333 before returning the error from the `client.Header()` method.

asthamohta: But then won't md, err := client.Header() override the value of the original error. I will have to add a separate check as md, errGFE : = client.Header()
			if errGFE != nil {
				return nil, errGFE
			}

rahul2393 (approved):

spanner/batch.go
olavloite: I'm not sure we mean the same return in this case. I think you can safely return the `if err != nil` block starting at line 325. The `err` on line 329 will not override the `err` from the call to `client.Header()`, as that is a different variable, although it does have the same name. That `err` is only visible in the statement on line 329.

So to summarize:
1. The `if err != nil` on line 321 should remain.
2. The `if err != nil` on line 325 can be removed. (And yes, the `md, err := client.Header()` does override the previous value of `err`, but that is not a problem as we know that `err` at that point is `nil`)
3. The `if err := createContextAndCaptureGFELatencyMetrics` on line 329 can safely remain. That will not override anything, as it is a separate variable only defined on line 329.
4. The `return client, err` on line 333 will therefore always return the value of `err` that was returned by `client.Header()`.

asthamohta: Okay makes sense. My bad. I thought you meant remove the error return on line 321. Will do that.

hengfengli (approved):

PR #5201, opened 2021-12-08
docs(bigquery): use equality operator in Errors example

mesge:
I believe the intention of this example is to show how to introspect errors for more information (such as the error code). This updates the example to use the equality operator instead of assignment operator.


shollyman (approved):
Thanks for the contribution, good spot!

PR #5208, opened 2021-12-09
docs(bigquery/storage/managedwriter): augment docs

shollyman:
Towards: #4366

### DIFF
--- a/bigquery/storage/managedwriter/doc.go
+++ b/bigquery/storage/managedwriter/doc.go
@@ -146,7 +146,7 @@ an error.
 	}
 
 
-Advanced Stream Management - Buffered Streams
+Buffered Stream Management
 
 For Buffered streams, users control when data is made visible in the destination table/stream
 independently of when it is written.  Use FlushRows on the ManagedStream to advance the flush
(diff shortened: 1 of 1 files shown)


codyoss (approved):
LGTM, nice docs!

bigquery/storage/managedwriter/doc.go
```diff
@@ -24,12 +24,157 @@ feature-rich successor to the classic BigQuery streaming interface, which is pre
...
+	client, err := managedwriter.NewClient(ctx, projectID)
+	if err != nil {
+		// TODO: Handle error.
+	}
+
+
+Defining the Protocol Buffer Schema
+
+The write functionality of BigQuery Storage requires data to be sent using encoded
+protocol buffer messages using proto2 syntax.  As the protocol buffer is not
```
tswast, line 42: Nit: [Encoding / binary wire format](https://developers.google.com/protocol-buffers/docs/encoding) is the thing the API needs. I think of syntax as the language definition, which doesn't always apply given we can create dynamic messages in some languages.

```suggestion
protocol buffer messages using proto2 wire format.  As the protocol buffer is not
```

tswast (approved):
Looks like you covered the main topics. Just a minor nit

bigquery/storage/managedwriter/doc.go
shollyman: Good callout.  I was considering "encoding" before, but was unhappy with it.  Wire format is cleaner.

PR #5210, opened 2021-12-09
feat(storage): configurable retries for uploads

tritone:
Depends on merge and release of googleapis/google-api-go-client#1324


storage/writer.go
```diff
@@ -172,6 +172,25 @@ func (w *Writer) open() error {
 			// call to set up the upload as well as calls to upload individual chunks
 			// for a resumable upload (as long as the chunk size is non-zero). Hence
 			// there is no need to add retries here.
+
+			// Retry only when the operation is idempotent or the retry policy is RetryAlways.
+			var isIdempotent bool
+			if w.o.conds != nil && (w.o.conds.GenerationMatch >= 0 || w.o.conds.DoesNotExist == true) {
+				isIdempotent = true
+			}
```
BrennaEpp, line 180: Here and below, isn't setting the variable to true redundant? 
```suggestion
				isIdempotent := w.o.conds != nil && (w.o.conds.GenerationMatch >= 0 || w.o.conds.DoesNotExist == true)
```

BrennaEpp (commented):
Nice! Just a couple comments.

This should make the objects.insert conformance tests pass, correct (once everything is merged)? If so then we should update the ci as well

storage/writer.go
tritone: Yep! I'll change this.

BrennaEpp (approved):

Issue #5209, opened 2021-12-09
spanner/spansql: v1.28.0 generated columns


geerttouquet:
**Client**
golang
**Environment**

**Go Environment**
$ go version go version go1.16.9 

**Code**

sql code that can be parsed by cloudspanner but it gives an error on spansql

CREATE TABLE test (
	patientId STRING(MAX) NOT NULL,
	starttime TIMESTAMP NOT NULL,
	cetDate DATE AS (EXTRACT(DATE FROM startTime AT TIME ZONE "CET")) STORED,
) PRIMARY KEY (patientId);

_, err := spansql.ParseDDL("filename", test.sql) gives following error 
ParseDDL("CREATE TABLE test (\n\tpatientId STRING(MAX) NOT NULL,\n\tstarttime TIMESTAMP NOT NULL,\n\tcetDate DATE AS (EXTRACT(DATE FROM startTime AT TIME ZONE \"CET\")) STORED,\n) PRIMARY KEY (patientId);"): filename:4: got "FROM", want ")" or ","

rahul2393:
Thanks for opening the request @geerttouquet I was able to replicate the issue, will create a PR but from the documentation of https://cloud.google.com/spanner/docs/reference/standard-sql/date_functions#extract return type of Extract is always Int64 and we cannot extract Date from a Date

PR #5218, opened 2021-12-15
feat(spanner/spansql): support EXTRACT

rahul2393:
This adds support for EXTRACT func and similar constructions

Fixes: https://github.com/googleapis/google-cloud-go/issues/5209

### DIFF
--- a/spanner/spansql/parser.go
+++ b/spanner/spansql/parser.go
@@ -1901,6 +1901,39 @@ func (p *parser) parseType() (Type, *parseError) {
 	return p.parseBaseOrParameterizedType(true)
 }
 
+var extractPartTypes = map[string]TypeBase{
+	"NANOSECOND":  Int64,
+	"MICROSECOND": Int64,
+	"MILLISECOND": Int64,
+	"SECOND":      Int64,
+	"MINUTE":      Int64,
+	"HOUR":        Int64,
+	"DAYOFWEEK":   Int64,
+	"DAY":         Int64,
+	"DAYOFYEAR":   Int64,
+	"WEEK":        Int64,
+	"ISOWEEK":     Int64,
+	"MONTH":       Int64,
+	"QUARTER":     Int64,
+	"YEAR":        Int64,
+	"ISOYEAR":     Int64,
+	"DATE":        Date,
+}
+
+func (p *parser) parseExtractType() (Type, *parseError) {
+	var t Type
+	tok := p.next()
+	if tok.err != nil {
+		return Type{}, tok.err
+	}
+	base, ok := extractPartTypes[strings.ToUpper(tok.value)] // valid part types for EXTRACT is keyed by upper case strings.
+	if !ok {
+		return Type{}, p.errorf("got %q, want valid EXTRACT types", tok.value)
+	}
+	t.Base = base
+	return t, nil
+}
+
 func (p *parser) parseBaseOrParameterizedType(withParam bool) (Type, *parseError) {
 	debugf("parseBaseOrParameterizedType: %v", p)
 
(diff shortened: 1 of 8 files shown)


spanner/spansql/parser.go
```diff
@@ -2482,6 +2515,33 @@ var typedArgParser = func(p *parser) (Expr, *parseError) {
...
+// Special argument parser for EXTRACT
+var extractArgParser = func(p *parser) (Expr, *parseError) {
+	partType, err := p.parseExtractType()
+	if err != nil {
+		return nil, err
+	}
+	if err := p.expect("FROM"); err != nil {
+		return nil, err
+	}
+	e, err := p.parseTableOrIndexOrColumnName()
```
olavloite, line 2527: This is formally not correct, but I'm fine with this being a limitation (that we in that case should add to the list of limitations): The definition of the `EXTRACT` function specifies that any timestamp expression may be given at this position. That also includes any timestamp literal or function call that returns a timestamp. This implementation only allows the use of a column that contains a timestamp.

So it means that the following statement would not parse correctly:

```sql
select extract(date from timestamp '2020-01-01T10:00:00+01:00' at time zone 'Europe/Amsterdam') as d
```

spanner/spansql/parser.go
```diff
@@ -2482,6 +2515,33 @@ var typedArgParser = func(p *parser) (Expr, *parseError) {
...
+	if err != nil {
+		return nil, err
+	}
+	// AT TIME ZONE is optional
+	if p.eat("AT", "TIME", "ZONE") {
+		tok := p.next()
+		if tok.err != nil {
+			return nil, err
+		}
+		return TypedExpr{Expr: Func{Name: "AT TIME ZONE", Args: []Expr{e, StringLiteral(tok.string)}}, Type: partType}, nil
```
olavloite, line 2537: This is modelling `AT TIME ZONE` as a function. I don't think that is correct (or at least it will be very difficult to work with when you want to implement it), as function names do normally not contain spaces. Instead, I think this should be considered as a timezone expression, and then be passed in as an optional argument to the `EXTRACT` function.

In order to get an idea of what works best, it might be good to add (an) empty implementation(s) of the function(s) in spannertest, and add tests for that as well. The implementation(s) can just return a fixed value, but then you know that you have chosen a parsing strategy that will work.

spanner/spansql/parser_test.go
```diff
@@ -524,7 +528,9 @@ func TestParseDDL(t *testing.T) {
 		CREATE TABLE users (
 		  user_id      STRING(36) NOT NULL,
 		  some_string  STRING(16) NOT NULL,
+          some_time TIMESTAMP NOT NULL,
```
olavloite, line 531: nit: fix indentation

spanner/spansql/parser_test.go
```diff
@@ -524,7 +528,9 @@ func TestParseDDL(t *testing.T) {
...
 		  number_key   INT64 AS (SAFE_CAST(SUBSTR(some_string, 2) AS INT64)) STORED,
+          generated_date DATE AS (EXTRACT(DATE FROM some_time AT TIME ZONE "CET")) STORED,
```
olavloite, line 533: nit: fix indentation

spanner/spansql/sql.go
```diff
@@ -589,6 +589,20 @@ func (te TypedExpr) addSQL(sb *strings.Builder) {
...
+func (ee ExtractExpr) addSQL(sb *strings.Builder) {
+	sb.WriteString(ee.Part)
+	sb.WriteString(" FROM ")
+	ee.Expr.addSQL(sb)
+}
+
+func (aze AtTimeZoneExpr) SQL() string { return buildSQL(aze) }
+func (aze AtTimeZoneExpr) addSQL(sb *strings.Builder) {
+	aze.Expr.addSQL(sb)
+	sb.WriteString(" AT TIME ZONE  ")
```
olavloite, line 602: super nit: remove extra space

```suggestion
	sb.WriteString(" AT TIME ZONE ")
```

olavloite (approved):
LGTM

PR #5232, opened 2021-12-21
feat(bigtable): add support for autoscaling

enocom:
See https://github.com/googleapis/googleapis/commit/0fd6a324383fdd1220c9a937b2eef37f53764664 for details.

Note: this work depends on https://github.com/googleapis/googleapis/commit/0fd6a324383fdd1220c9a937b2eef37f53764664 (from Nov. 15) being available in [go-genproto.](https://github.com/googleapis/go-genproto/tree/main/googleapis/bigtable/admin/v2) (last updated Nov. 2).


codyoss:
https://github.com/googleapis/go-genproto/pull/733 has been merged. Should now just need to update the version of genproto being pulled in.

enocom:
Thanks @codyoss!

bigtable/admin.go
```diff
@@ -1167,14 +1254,21 @@ func (iac *InstanceAdminClient) Clusters(ctx context.Context, instanceID string)
...
 			Name:        nameParts[len(nameParts)-1],
 			Zone:        locParts[len(locParts)-1],
 			ServeNodes:  int(c.ServeNodes),
 			State:       c.State.String(),
 			StorageType: storageTypeFromProto(c.DefaultStorageType),
 			KMSKeyName:  kmsKeyName,
-		})
+		}
+		// Use type assertion to handle protobuf oneof type
+		if cfg, ok := c.Config.(*btapb.Cluster_ClusterConfig_); ok {
```
tritone, line 1266: Can you use this instead? https://github.com/googleapis/go-genproto/blob/d629cc9a93d58a60b9f166bf2a825dd80df3879d/googleapis/bigtable/admin/v2/instance.pb.go#L547

bigtable/admin.go
```diff
@@ -1206,15 +1300,39 @@ func (iac *InstanceAdminClient) GetCluster(ctx context.Context, instanceID, clus
...
+	// Use type assertion to handle protobuf oneof type
+	if cfg, ok := c.Config.(*btapb.Cluster_ClusterConfig_); ok {
+		if asc, ok := autoscalingConfig(cfg); ok {
+			ci.AutoscalingConfig = asc
+		}
+	}
+	return ci, nil
+}
+
+func autoscalingConfig(c *btapb.Cluster_ClusterConfig_) (AutoscalingConfig, bool) {
```
tritone, line 1320: Is there a naming convention for these veneer type to/from proto type funcs? In storage/apiary we say to/from raw usually.

bigtable/admin.go
```diff
@@ -1033,6 +1045,41 @@ func (iac *InstanceAdminClient) InstanceInfo(ctx context.Context, instanceID str
...
+	// CPUTargetPercent sets the CPU utilization target for your cluster's
+	// workload.
+	CPUTargetPercent int
+}
+
+func (a AutoscalingConfig) isZero() bool {
+	return a.MinNodes == 0 && a.MaxNodes == 0 && a.CPUTargetPercent == 0
+}
+
+func (a AutoscalingConfig) proto() *btapb.Cluster_ClusterConfig_ {
```
tritone, line 1064: It seems a bit odd to have an `AutoscalingConfig` convert to a new `Cluster_ClusterConfig` rather than say modifying an existing one. Is it possible that other cluster config fields will be added in the future?

tritone (commented):
Couple thoughts and questions, hopefully some of this is helpful!

bigtable/admin.go
```diff
@@ -808,6 +808,10 @@ type InstanceConf struct {
 	StorageType  StorageType
 	InstanceType InstanceType
 	Labels       map[string]string
+
+	// AutoscalingConfig configures the autoscaling properties on the cluster
+	// created with the instance. It is optional.
+	AutoscalingConfig AutoscalingConfig
```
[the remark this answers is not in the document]
enocom: Whether this field should be a pointer or not is one of the key design decisions here. A zero value AutoscalingConfig isn't valid in the API, but whether we should catch that in the client or not is the question. Making it a pointer would alleviate the ambiguity while departing from the other value objects. @telpirion what do you think?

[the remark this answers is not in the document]
enocom: And to answer your question -- this config will be for just the initial cluster.

bigtable/admin.go
```diff
@@ -1033,6 +1045,41 @@ func (iac *InstanceAdminClient) InstanceInfo(ctx context.Context, instanceID str
 	}, nil
 }
 
+// AutoscalingConfig contains autoscaling configuration for a cluster.
+// For details, see https://cloud.google.com/bigtable/docs/autoscaling.
```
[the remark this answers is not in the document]
enocom: Good point. I'll update the documentation to call this out.

bigtable/admin.go
enocom: That's a good point. I'll update this function to just return the corresponding type. If future fields are added, I think this method will be open for extension.

bigtable/admin.go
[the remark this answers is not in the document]
telpirion: If I may, I think it would be best to keep this as a value field. For one, there's the parallelism with the other fields. For another, I think it's helpful to have the `isZero()` method to check whether the developer accidentally passed in a meaningless `AutoscalingConfig` (where all fields were set to zero).

bigtable/admin.go
```diff
@@ -1129,13 +1185,44 @@ func (iac *InstanceAdminClient) DeleteCluster(ctx context.Context, instanceID, c
...
+		},
+		Cluster: cluster,
+	})
+	if err != nil {
+		return err
+	}
+	return longrunning.InternalNewOperation(iac.lroClient, lro).Wait(ctx, nil)
+}
+
+// UpdateCluster updates attributes of a cluster. If Autoscaling is configured
```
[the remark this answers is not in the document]
enocom: Unfortunately, this method only changes the number of serve nodes. I'll update the documentation to make that clear. Since we can't change this function, we're effectively stuck with the name. As an alternative to this approach here (SetAutoscaling turns it on, UpdateCluster turns it off), we could introduce a more generic function, e.g., UpdateClusterConfig or similar.

bigtable/admin.go
```diff
@@ -1129,13 +1185,44 @@ func (iac *InstanceAdminClient) DeleteCluster(ctx context.Context, instanceID, c
...
 	cluster := &btapb.Cluster{
 		Name:       "projects/" + iac.project + "/instances/" + instanceID + "/clusters/" + clusterID,
-		ServeNodes: serveNodes}
-	lro, err := iac.iClient.UpdateCluster(ctx, cluster)
+		ServeNodes: serveNodes,
+		// Explicitly removing autoscaling config (and including it in the field
+		// mask below)
+		Config: nil,
+	}
+	lro, err := iac.iClient.PartialUpdateCluster(ctx, &btapb.PartialUpdateClusterRequest{
```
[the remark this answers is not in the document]
enocom: Yes, they're synonymous. PartialUpdateCluster is the preferred update method now.

bigtable/admin.go
enocom: Yes! Nice find.

bigtable/admin.go
telpirion: This method makes sense to me (but maybe I'm thinking about this wrong). I think about proto message objects as being like value-only structs--they're a way to pass structured data around.

I'm not sure what we gain by having this method modify an existing `Cluster_ClusterConfig_` proto message.

bigtable/admin.go
enocom: to/from seems like the way to go. I'll update here.

bigtable/admin.go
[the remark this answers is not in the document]
telpirion: I agree. Can we put a link to the official autoscaling documentation? Maybe to here:

https://cloud.google.com/bigtable/docs/autoscaling

or

https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#autoscalinglimits

bigtable/admin.go
telpirion: I think the convention is `toProto()` and `fromProto()`, but I can't swear to it.

bigtable/admin.go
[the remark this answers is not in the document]
tritone: I think generally, in principle, it's best to let the service catch "bad" values rather than attempting to validate on the client side (been burned by this too many times!). However, I definitely see the rationale for keeping this field parallel to the existing fields. Either way is fine with me ultimately.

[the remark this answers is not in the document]
enocom: I've been on both sides of this. I think a pointer aligns best with what a Go developer would expect. So I'll make this a pointer.

bigtable/admin.go
enocom: Updated to fromClusterConfigProto (since fromProto would be ambiguous).

bigtable/admin.go
enocom: Tightened this up to have it return the corresponding proto type.

bigtable/admin.go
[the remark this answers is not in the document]
enocom: I think introducing a more general UpdateClusterConfig (or similar) is a good idea, but has its own gotcha's. So I propose we punt on that work for now, especially since such a future change could simply refactor the internals of SetAutoscaling to use the more general update method.

bigtable/admin.go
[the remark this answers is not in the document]
enocom: Done. I've added the basic expectations for Min and Max Nodes.

enocom:
@telpirion and @tritone PTAL. I've addressed all the comments above.

bigtable/admin_test.go
```diff
@@ -0,0 +1,571 @@
+package bigtable
```
kolea2, line 1: missing header

enocom: fixed

bigtable/admin.go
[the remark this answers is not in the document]
tritone: Specifically this section seems the most important for the config: https://cloud.google.com/bigtable/docs/autoscaling#parameters

bigtable/admin.go
```diff
@@ -1119,19 +1113,27 @@ type ClusterConfig struct {
...
 		EncryptionConfig: &btapb.Cluster_EncryptionConfig{
 			KmsKeyName: cc.KMSKeyName,
 		},
-		Config: cc.AutoscalingConfig.proto(),
 	}
+
+	if asc := cc.AutoscalingConfig.proto(); asc != nil {
+		cl.Config = &btapb.Cluster_ClusterConfig_{
+			ClusterConfig: &btapb.Cluster_ClusterConfig{
+				ClusterAutoscalingConfig: cc.AutoscalingConfig.proto(),
```
tritone, line 1132: You can just set it equal to `asc` here correct?

bigtable/admin.go
```diff
@@ -1186,12 +1188,16 @@ func (iac *InstanceAdminClient) DeleteCluster(ctx context.Context, instanceID, c
...
+// UpdateCluster. See AutoscalingConfig documentation for deatils.
 func (iac *InstanceAdminClient) SetAutoscaling(ctx context.Context, instanceID, clusterID string, conf AutoscalingConfig) error {
 	ctx = mergeOutgoingMetadata(ctx, iac.md)
 	cluster := &btapb.Cluster{
-		Name:   "projects/" + iac.project + "/instances/" + instanceID + "/clusters/" + clusterID,
-		Config: conf.proto(),
+		Name: "projects/" + iac.project + "/instances/" + instanceID + "/clusters/" + clusterID,
+		Config: &btapb.Cluster_ClusterConfig_{
+			ClusterConfig: &btapb.Cluster_ClusterConfig{
+				ClusterAutoscalingConfig: conf.proto(),
```
tritone, line 1198: This will now be a `Cluster_ClusterConfig` object with a nil `ClusterAutoscalingConfig`, rather than just a nil `Cluster_ClusterConfig`. Does that make a difference?

bigtable/admin.go
```diff
@@ -1309,22 +1314,22 @@ func (iac *InstanceAdminClient) GetCluster(ctx context.Context, instanceID, clus
...
+		if asc, ok := fromClusterConfigProto(cfg); ok {
+			ci.AutoscalingConfig = &asc
 		}
 	}
 	return ci, nil
 }
 
-func autoscalingConfig(c *btapb.Cluster_ClusterConfig_) (AutoscalingConfig, bool) {
-	if c.ClusterConfig == nil {
+func fromClusterConfigProto(c *btapb.Cluster_ClusterConfig) (AutoscalingConfig, bool) {
```
tritone, line 1325: Based on how you're using this, maybe simpler just to have this return *AutoscalingConfig which may be nil if it's not present in the proto?

tritone (commented):
couple more minor comments/questions, overall LGTM. I will leave it to @kolea2 and @telpirion to approve!

bigtable/admin.go
```diff
@@ -1478,8 +1601,12 @@ func max(x, y int) int {
 //   and the given ClusterConfig.
 // - Any cluster missing from conf.Clusters but present in the instance will be removed from the instance
 //   using DeleteCluster.
-// - Any cluster in conf.Clusters that also exists in the instance will be updated to contain the
-//   provided number of nodes if set.
+// - Any cluster in conf.Clusters that also exists in the instance will be
+//   updated either to contain the provided number of nodes or to use the
+//   provided autoscaling config. If both the number of nodes and autoscaling
+//   are configured, autoscaling takes precedence. If the number of nodes is zero
```
kolea2, line 1607: In this condition (zero and not provided), should we instead surface an error?

bigtable/integration_test.go
```diff
@@ -1985,90 +1982,142 @@ func TestIntegration_AdminUpdateInstanceAndSyncClusters(t *testing.T) {
...
-func (iacm *instanceAdminClientMock) GetAppProfile(ctx context.Context, in *btapb.GetAppProfileRequest, opts ...grpc.CallOption) (*btapb.AppProfile, error) {
-	return nil, nil
-}
+	clusterID := instanceToCreate + "-cluster"
 
-func (iacm *instanceAdminClientMock) ListAppProfiles(ctx context.Context, in *btapb.ListAppProfilesRequest, opts ...grpc.CallOption) (*btapb.ListAppProfilesResponse, error) {
-	return nil, nil
-}
+	serveNodes := 1
+	t.Logf("creating an instance with serve nodes = %v (autoscaling OFF)", serveNodes)
```
kolea2, line 2013: is there a test creating an instance with autoscaling on?

kolea2 (commented):
Mostly reviewing from an API perspective but looks really good! A few questions/comments

bigtable/admin.go
enocom: Woops. Yes.

enocom: Actually, I removed the `return nil` statement inside `proto` and cleaned this up as well.

bigtable/admin.go
enocom: I checked -- and both are rejected without a valid autoscaling config.

bigtable/admin.go
enocom: Yes, that's better and consistent with other similar functions. Done.

bigtable/admin.go
```diff
@@ -1047,7 +1088,8 @@ type ClusterConfig struct {
 	Zone string
 
 	// NumNodes specifies the number of nodes allocated to this cluster. More
-	// nodes enable higher throughput and more consistent performance. Required.
+	// nodes enable higher throughput and more consistent performance. One of
+	// NumNodes or AutoscalingConfig is required.
```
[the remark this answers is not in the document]
enocom: Done.

bigtable/integration_test.go
enocom: It's probably better to start with creating an instance with autoscaling, disable it, and then re-enable. I'll update the test to do that. How does that sound?

bigtable/admin.go
enocom: It's a good question. The existing behavior was to ignore `NumNodes <= 0`, so if we change the code to return an error, we'd be changing the behavior. Since I'm new to the code here, I'm inclined to maintain the status quo and leave it for a future pass.

bigtable/admin.go
```diff
@@ -1033,6 +1045,32 @@ func (iac *InstanceAdminClient) InstanceInfo(ctx context.Context, instanceID str
...
+	// MaxNodes sets the maximum number of nodes in a cluster. MaxNodes must be
+	// equal to or greater than MinNodes.
+	MaxNodes int
+	// CPUTargetPercent sets the CPU utilization target for your cluster's
+	// workload.
+	CPUTargetPercent int
+}
+
+func (a *AutoscalingConfig) proto() *btapb.Cluster_ClusterAutoscalingConfig {
+	return &btapb.Cluster_ClusterAutoscalingConfig{
```
tritone, line 1063: I think you should probably keep the `nil` case for safety.

tritone (approved):
one nit, otherwise LGTM

bigtable/admin.go
kolea2: sgtm!

kolea2 (approved):

bigtable/integration_test.go
kolea2: perfect thanks!

bigtable/admin.go
enocom: Done.

PR #5240, opened 2021-12-21
feat(spanner): Add ReadRowWithOptions method

shuheiktgw:
Currently, there are no ways users can provide ReadOptions to the ReadRow method. It would be great if we have a method for that purpose so I've added a new method called `ReadRowWithOptions`.  I could not find tests for the ReadRow method so would you guide me on where to add tests for the new method? Thank you!


rahul2393:
Thanks @shuheiktgw for creating the PR, you can add test by referring to this test https://github.com/googleapis/google-cloud-go/blob/main/spanner/client_test.go#L2839

shuheiktgw:
@rahul2393 thanks for pointing it out! I've added a test for the ReadRowWithOptions so would you review the PR please? 🙏

rahul2393 (approved):
LGTM

Issue #5246, opened 2021-12-22
firestore: GetAll panics if iterator has error


jba:
**Client**

Firestore

**Environment**

any
**Go Environment**

any

**Code**

After writing in a transaction,

```go
    err = c.RunTransaction(ctx, func(_ context.Context, tx *Transaction) error {
        if err := tx.Delete(c.Doc("C/a")); err != nil {
            return err
        }
        _, err := tx.Documents(c.Collection("C").Select("x")).GetAll()
        return err
    })
```

**Expected behavior**

"read after write" error

**Actual behavior**

panic

PR #5247, opened 2021-12-22
fix(firestore): DocumentIterator.GetAll: check iterator error

jba:
Check to see if the iterator is in an error state first, and
return the error if it is.

Otherwise, the function panics when accessing `q.limitToLast`, because
q is nil.

Fixes googleapis/google-cloud-go#5246.


tritone (approved):
Thanks for the fix!

jba:
I don't see the kokoro failure ([logs](https://source.cloud.google.com/results/invocations/ba0a9ea8-2504-42f5-90be-2ca7fe501536/targets/cloud-devrel%2Fclient-libraries%2Fgo%2Fgoogle-cloud-go%2Fpresubmit%2Fgo117/log)) but I assume it's a flake. How can I re-run?

tritone:
Looks like there was a gofmt error on a file not in your PR somehow-- should be fixed now in any case. I'll retrigger.

Issue #5244, opened 2021-12-21
bigquery/storage/managedwriter: support schema change notifications


shollyman:
`cloud.google.com/go/bigquery/storage/managedwriter` doesn't yet include support for dealing with schema change notifications from the storage write backend.

When schema changes are propagated from the v1 storage write backend (e.g. from a `tables.update` or `ALTER TABLE` query), the backend service populates `UpdatedSchema` as part of the [`AppendRowsResponse`](https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/bigquery/storage/v1#AppendRowsResponse), which is part of the core AppendRows bidi stream.  There's no other notification path, barring doing something horrible like calling the unary `GetWriteStream` and polling for changes.

The managedwriter veneer package already allows users to set a new proto descriptor (schema) when appending by passing the [`UpdateSchemaDescriptor`](https://pkg.go.dev/cloud.google.com/go/bigquery/storage/managedwriter#UpdateSchemaDescriptor) option to the AppendRows method on a ManagedStream instance.

The question to tackle here is *how the notifications should be surfaced to users*.

The **least effort** solution is to add the updated schema to the [`AppendResult`](https://pkg.go.dev/cloud.google.com/go/bigquery/storage/managedwriter#AppendResult) future, and simply return the schema as a method there.  We'd use the same channel block as we use for Ready/GetResult, so to be performant users would need to go through extra hoops to check every response

However, another option is to define an interface for propagating the notifications, and allowing users to register a listener capable of receiving the callback.  This obviates the need for users to monitor responses for the notification directly, but it's unclear what should be passed back.

### defining the contract 

Clearly, the **new schema** needs to be part of the message.  This is yet another instance of where the v1 proto types need to either be exposed directly from managedwriter, or we need to wrap; I'm inclined to give up on encapsulating the v1 proto messages and simply couple managedwriter to the v1 API directly.

Another component of the notification likely should be the **stream name** itself.  The backend does not currently support stream multiplexing, but it is expected to in the future.  To disambiguate between write streams, we should callback with the identifier.

I not sure **context** should be part of the notification at this point.  The context we'd have available for use is the retained context inside ManagedStream responsible for network stream management, and it seems odd to use that to notify the likely owner/controller of the ManagedStream instance.  Then again, if we want to end up propagating things like trace spans in the future, it's probably necessary.  We could generate a new context, or derive one from the retained context if we end up sending one.

[...]

PR #5253, opened 2021-12-23
feat(bigquery/storage/managedwriter): support schema change notification

shollyman:
This adds a way for user to get schema changes by checking AppendResult.

Related: https://github.com/googleapis/google-cloud-go/issues/5244
Related: https://github.com/googleapis/google-cloud-go/issues/4366


tswast (approved):
LGTM. It's the "least effort solution" described in https://github.com/googleapis/google-cloud-go/issues/5244, right?

> add the updated schema to the AppendResult future, and simply return the schema as a method there.

shollyman:
> LGTM. It's the "least effort solution" described in #5244, right?
> 
> > add the updated schema to the AppendResult future, and simply return the schema as a method there.

That's correct.  It may make sense to have a callback in the future, but this allows users to roll their own monitor.

By comparison, java effectively exposes this as a property of the StreamWriter, but there's no way to signal/clear it; users end up needing to recreate a new one.

PR #5273, opened 2022-01-04
chore: carve out compute module

codyoss:
This will be tagged:

- v0.100.2
- compute/v0.1.0


noahdietz (approved):
I'm guessing that the go mod download issues are expected given that this **creates** the compute v0.1.0 module?

codyoss:
Yes, the tests here are expected to fail :)

PR #5275, opened 2022-01-05
fix: change to solve issue 5266

asthamohta:
Fixes: https://github.com/googleapis/google-cloud-go/issues/5266


rahul2393:
Changes already included in https://github.com/googleapis/google-cloud-go/pull/5276

PR #5276, opened 2022-01-05
test(spanner)!: fix data race in spanner integration tests

rahul2393:
* Update GFELatencyMetricsEnabled flag private to the package, to avoid possibility of data race from customers.
* Added setter/getter method to update the GFELatencyMetricsEnabled flag.
* Added lock to make GFELatencyMetricsEnabled thread safe.

Fixes https://github.com/googleapis/google-cloud-go/issues/5272


spanner/stats.go
```diff
@@ -213,28 +216,40 @@ func EnableStatViews() error {
...
 func EnableGfeLatencyAndHeaderMissingCountViews() error {
-	GFELatencyMetricsEnabled = true
+	SetGFELatencyMetricsFlag(true)
 	return view.Register(
 		GFELatencyView,
 		GFEHeaderMissingCountView,
 	)
 }
 
+func GetGFELatencyMetricsFlag() bool {
```
olavloite, line 238: Could this be unexported? And same for the setter?

asthamohta (approved):

spanner/stats.go
```diff
@@ -267,12 +282,6 @@ func checkCommonTagsGFELatency(t *testing.T, m map[tag.Key]string) {
 	if !strings.HasPrefix(m[tagKeyClientID], "client") {
 		t.Fatalf("Incorrect client ID: %v", m[tagKeyClientID])
 	}
-	if !strings.HasPrefix(m[tagKeyInstance], "gotest") {
```
olavloite, line 270: Good catch 👍

spanner/stats.go
rahul2393: done

olavloite:
By the way: The build error seems to be unrelated to this change. AFAICT, it is caused by a formatting error in the unedited `iam/go_mod_tidy_hack.go` file.

@codyoss It seems that #5269 introduced a formatting error that was hidden by a different build error in that PR, but that it is a formatting error that only seems to be picked up by Go 1.17. At least, if I run `gofmt -s -d` on that file using Go 1.16, it thinks it's ok.

olavloite (approved):
If I understand it correctly, the GFE Latency feature had not yet been released. We should make sure that this change then is included in the next release, as it would otherwise be a breaking change (we are removing an exported variable).

rahul2393:
No its not yet released so its safe for now, here is the release PR https://github.com/googleapis/google-cloud-go/pull/5219

rahul2393:
@olavloite build passed after changing iam/go_mod_tidy_hack.go and adding the unused exported GFELatencyMetricsEnabled field back to this PR. is it good to merge now?
cc: @codyoss @hengfengli

hengfengli (approved):
LGTM.

PR #5301, opened 2022-01-06
test(storage): fill out invoke tests

BrennaEpp:


storage/invoke_test.go
```diff
@@ -121,8 +169,94 @@ func TestInvoke(t *testing.T) {
...
+			if test.expectFinalErr && got != test.finalErr {
+				s.Errorf("got %v, want %v", got, test.finalErr)
+			} else if !test.expectFinalErr && got != test.initialErr {
+				s.Errorf("got %v, want %v", got, test.initialErr)
+			}
+		})
+	}
+}
+
+func TestShouldRetry(t *testing.T) {
```
tritone, line 181: Separating this out is much nicer, thanks!

storage/invoke_test.go
```diff
@@ -121,8 +169,94 @@ func TestInvoke(t *testing.T) {
...
+		{
+			desc:        "googleapi.Error{Code: 400}",
+			inputErr:    &googleapi.Error{Code: 400},
+			shouldRetry: false,
+		},
+		{
+			desc:        "googleapi.Error{Code: 408}",
+			inputErr:    &googleapi.Error{Code: 408},
+			shouldRetry: false, // to be changed
+		},
```
tritone, line 253: Could you add cases for retryable/non-retryable gRPC statuses?

tritone (commented):
Couple little things, thanks for adding this!

storage/invoke_test.go
BrennaEpp: done! let me know if you think we should add more/all possible gRPC statuses

tritone: looks good!

tritone (approved):

PR #5313, opened 2022-01-07
storage: bucket exists func implemented

eftalyurtseven:
I didn't find any method for checking bucket exists or not,
This PR simply checks the whether bucket exists or not. 
Stackoverflow: https://stackoverflow.com/a/52935830/5905905

### DIFF
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -74,6 +74,20 @@ func (c *Client) Bucket(name string) *BucketHandle {
 	}
 }
 
+// BucketExists checks whether bucket exists or not with given bucket object
+// https://stackoverflow.com/a/52935830/5905905
+func (c *Client) BucketExists(ctx context.Context, bucket *BucketHandle) (bool, error) {
+	_, err := bucket.Attrs(ctx)
+	if err != nil {
+		if err == ErrBucketNotExist {
+			return false, nil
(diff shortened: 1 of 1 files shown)


tritone:
Hi @eftalyurtseven , thanks very much for contributing.

I am inclined not to merge this because it's just a wrapper for the existing `BucketHandle.Attrs` method. In general we try to keep the exported surface for this library as small as possible, unless we have a strong need to expand it.

However, I'll leave the PR open for now in case others have thoughts. (I know some of the other language GCS clients have a similar bucket.Exists method).

tritone (commented):
Hi @eftalyurtseven , thanks very much for contributing.

I am inclined not to merge this because it's just a wrapper for the existing `BucketHandle.Attrs` method. In general we try to keep the exported surface for this library as small as possible, unless we have a strong need to expand it.

However, I'll leave the PR open for now in case others have thoughts. (I know some of the other language GCS clients have a similar bucket.Exists method).

eftalyurtseven:
Hi @tritone,thank you very much for your quick and friendly response.

I saw different pattern in python client. If I migrate to golang this func with .Query, will it be possible solution for this?

codyoss:
Closing based on @tritone comment above.

Issue #5319, opened 2022-01-08
pubsub: fake server does not handle deadletters and retry counts


yottta:
**Client**

PubSub

**Go Environment**

go version go1.16.6 darwin/amd64

**Description**

The PubSub library offers a testing server that is still in beta and therefore does not have all the functionalities.
This ticket is about improving the PubSub testing server to handle correctly Deadletters and retries.

**Expected behavior**

When a Subscription is created with a DeadLetter and retries configuration, an unacked message is received from the subscription only as many times as specified in the retries configuration.

**Actual behavior**

When a Subscription is created with a DeadLetter configuration, an unacked message is never sent to deadletter and is received from the subscription indefinitely.

yottta:
Fixed by #5320

burdzwastaken:
@yottta could the PR #5374 or #5463 get merged soon so this functionality is available? will help with our testing strategy. thank you so much

burdzwastaken:
sorry @yottta had a brainfart & didn't realize you were the contributor. thanks for the approval anyway <3

PR #5320, opened 2022-01-08
fix(pubsub): add deadletter and retries handling in the fake pubsub

yottta:
**Client**

PubSub

**Go Environment**

go version go1.16.6 darwin/amd64

**Description**

The PubSub library offers a testing server that is still in beta and therefore does not have all the functionalities.
This ticket is about improving the PubSub testing server to handle correctly Deadletters and retries.

**Expected behavior**

When a Subscription is created with a DeadLetter and retries configuration, an unacked message is received from the subscription only as many times as specified in the retries configuration.

**Actual behavior**

When a Subscription is created with a DeadLetter configuration, an unacked message is never sent to deadletter and is received from the subscription indefinitely.


yottta:
Solves #5319

pubsub/pstest/fake.go
```diff
@@ -954,10 +979,15 @@ func (s *subscription) pull(max int) []*pb.ReceivedMessage {
...
-	for _, m := range s.msgs {
+	for id, m := range s.msgs {
 		if m.outstanding() {
 			continue
 		}
+		if s.deadletterCandidate(m) {
+			s.ack(id)
+			s.publishToDeadletter(m)
+			continue
+		}
```
hongalex, line 990: Sync delivery attempt field here before incrementing if dead letter policy is set.

```suggestion
		}
		if s.proto.DeadLetterPolicy != nil {
			m.proto.DeliveryAttempt = int32(*m.deliveries)
		}
```

pubsub/pstest/fake.go
```diff
@@ -426,6 +426,14 @@ func (s *GServer) DeleteTopic(_ context.Context, req *pb.DeleteTopicRequest) (*e
 	if t == nil {
 		return nil, status.Errorf(codes.NotFound, "topic %q", req.Topic)
 	}
+	for _, sub := range s.subs {
+		if sub.deadletterTopic == nil {
```
hongalex, line 430: nit: prefer deadLetterTopic as dead lettering is two words. 

in general, s/deadletter/deadLetter/g
s/Deadletter/DeadLetter/g

pubsub/pstest/fake_test.go
```diff
@@ -219,6 +278,138 @@ func TestSubscriptionErrors(t *testing.T) {
...
+			Subscription: sub.Name,
+			MaxMessages:  10,
+		})
+		if err != nil {
+			t.Fatalf("failed during pull")
+		}
+		if i < retries {
+			if len(pull.ReceivedMessages) != 1 {
+				t.Fatalf("expected 1 message received a different number %d", len(pull.ReceivedMessages))
+			}
```
hongalex, line 346: After adding the fix above to sync message.DeliveryAttempts, add verification for `m.proto.GetDeliveryAttempts()` here.

hongalex (commented):
Thanks for drafting this PR. I had a few minor suggestions but otherwise looks good!

pubsub/pstest/fake_test.go
yottta: Check [fake_test.go#349](https://github.com/googleapis/google-cloud-go/pull/5320/files#diff-c6a74920cef99f2ab6281118b9eb61b39c99f9b17ee2961758db6688e0c6af9eR349)

pubsub/pstest/fake.go
```diff
@@ -1074,6 +1113,33 @@ func (s *subscription) deleteStream(st *stream) {
...
+func (s *subscription) publishToDeadLetter(m *message) {
+	acks := 0
+	if m.acks != nil {
+		acks = *m.acks
+	}
+	deliveries := 0
+	if m.deliveries != nil {
+		deliveries = *m.deliveries
+	}
+	s.deadLetterTopic.publish(m.proto.Message, &Message{
```
yottta, line 1136: Here I doubt that I need to use the original message properties (`deliveries` and `acks`) since the `publish` method is not allowing the proto values as well for these properties.
This is about the changes that suggested around `m.proto.DeliveryAttempt`.
Also you can check [fake_test.go#382](https://github.com/googleapis/google-cloud-go/pull/5320/files#diff-c6a74920cef99f2ab6281118b9eb61b39c99f9b17ee2961758db6688e0c6af9eR382)

hongalex (approved):

PR #5335, opened 2022-01-11
chore(.github): revert to v2 of release-please

codyoss:
With the big refactor the logic to selectively filter commits on
scope was removed. Until that logic is reinstated we will move
back to v2.


codyoss:
Force merging so no bad release go out.

PR #5348, opened 2022-01-12
chore: fix snippet module up to resolve imports

codyoss:


tbpg (approved):
I'd expect the generator to handle all of this. But, LGTM if not.

PR #5375, opened 2022-01-19
chore(storage): support emulator for gRPC

tritone:
Previously, newHybridClient ignored the emulator env var when making
requests on the gRPC transport. With this change, we support the
emulator by adding a separate env var for gRPC server requests.


storage/storage.go
```diff
@@ -218,6 +219,25 @@ func newHybridClient(ctx context.Context, opts *hybridClientOptions) (*Client, e
 		return nil, err
 	}
 
+	// Set emulator options for gRPC if an emulator was specified. Note that in a
+	// hybrid client, STORAGE_EMULATOR_HOST will set the host to use for HTTP and
+	// STORAGE_EMULATOR_HOST_GRPC will set the host to use for gRPC (when using a
+	// local emulator, HTTP and gRPC must use different ports, so this is
+	// necessary).
+	if host := os.Getenv("STORAGE_EMULATOR_HOST_GRPC"); host != "" {
```
codyoss, line 227: Has this been agreed upon cross lang? Just making sure we have consistency 😸 (I realize this this gRPC work is only happening here for now)

noahdietz: I _believe_ this is what the emulator actually supports today, so Go isn't doing anything special here just following what the emulator dictated.

noahdietz (approved):

storage/storage.go
tritone: The library currently supports `STORAGE_EMULATOR_HOST`; this PR adds `STORAGE_EMULATOR_HOST_GRPC` as well. I'll confer with folks to verify that this name is acceptable. The env vars don't matter to the emulator directly at all.

tritone: Decided to make this a temporary measure since it's only required for the hybrid client and can be removed when we have a full gRPC client.

PR #5376, opened 2022-01-19
fix(storage): fix nil check in gRPC Reader

tritone:
GetObjectChecksums() can return nil if no checksums are present,
so we need to check for nil.

This might just be an emulator artifact but it's probably not a bad idea
to have this check regardless.


noahdietz:
Thanks for fixing this!

PR #5378, opened 2022-01-19
chore(bigtable): upgrade some deps

codyoss:
- cloud.google.com/go
- google.golang.org/genproto
- google.golang.org/api

This is needed for some direct path changes.


codyoss:
cc @mohanli-ml

tbpg (approved):

PR #5382, opened 2022-01-20
fix(pubsublite): mitigate gRPC stream connection issues

tmdiep:
Mitigates hanging streams by detecting idle streams and reconnecting after a timeout (default 10min for partition assignment streams, 2min for all others). If the user has specified a lower timeout in settings, this will be used for publish, subscribe and commit streams.

The `StreamIdleTimer` is restarted when the client receives a response on the stream. The stream is reinitialized when the timeout expires. For publish and commit streams, the timeout will still expire even if there is no user activity.


tmdiep:
This is a port of the logic in the Java lib (https://github.com/googleapis/java-pubsublite/pull/1038), which has been tested internally for over a week. It seems to have eliminated the stream hanging issues since rollout.

hongalex (approved):

PR #5383, opened 2022-01-20
fix: add http request to validate access to metadata server

### LINKED ISSUE #5381: compute/metadata: `OnGCE` returns true if request to metadata server returns `ECONNREFUSED`
**Client**

Metadata & Profiler

**Environment**

Buildkite runners running on GCE. The metadata server responds to a ping, but not to a curl (returns ECONNREFUSED error). I can also reproduce locally by simply adding the hostname and IP to my `/etc/hosts` file (obviously pointing to nothing). I'm using the `profiler` package, previously didn't have any issue until #5063 was merged. When the profiler starts, the `OnGCE` call returns true, and before the bugfix, the call to `Zone()` would simply return an empty string, now it's returning an error (as it should). I think the issue stems from the `OnGCE` call, since the metadata server can't be reached it should probably return `false`.

**Go Environment**

$ go version
go version go1.17.3 linux/amd64

**Expected behavior**

I expect `OnGCE` to return `false` so that the profiler doesn't attempt to get instance metadata when it obviously can't.

**Actual behavior**

`OnGCE` returns `true`, causing the profiler to attempt retrieving instance metadata.

**Additional Context:**

[this check here](https://github.com/googleapis/google-cloud-go/blob/main/compute/metadata/metadata.go#L143-L148) is likely the issue; The test above it will not only fail in the IP doesn't resolve, but also if the connection is refused, whereas the `LookupHost` test only fails if the DNS lookup operation fails, it doesn't check to see if the server responds correctly. Since it only takes one of the two tests to pass for this function to return true, we get into this case where the `OnGCE` call returns true even if the server doesn't respond.

wperron:
Fixes #5381

cc @gravelg

### DIFF
--- a/compute/metadata/metadata.go
+++ b/compute/metadata/metadata.go
@@ -145,7 +145,23 @@ func testOnGCE() bool {
 			resc <- false
 			return
 		}
-		resc <- strsContains(addrs, metadataIP)
+
+		// If the host name cannot be resolved, send false and exit early.
+		if !strsContains(addrs, metadataIP) {
+			resc <- false
+			return
+		}
+
+		// Try to poke the metadata server to see if it can be reached.
+		req, _ := http.NewRequest("GET", "http://"+addrs[0], nil)
+		req.Header.Set("User-Agent", userAgent)
+		res, err := defaultClient.hc.Do(req.WithContext(ctx))
+		if err != nil {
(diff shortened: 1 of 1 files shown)


wperron:
Closing this following discussion in the linked issue.

PR #5408, opened 2022-01-27
chore(internal/gapicgen): update microgenerator to v0.24.0

noahdietz:
The only changes that are relevant are for `compute`, including use of `gax.Invoke`, support for header injection via context metadata, and support for URIs with template variables. [Release notes](https://github.com/googleapis/gapic-generator-go/releases/tag/v0.24.0).


noahdietz:
FYI @FrodoTheTrue this will introduce some changes to the internals of `compute/apiv1`. I don't think any integration tests/samples need to be updated after regeneration with this new version.

codyoss (approved):

PR #5426, opened 2022-02-02
chore(storage): internal client interface design

noahdietz:
Introduces an internal interface for the Storage client implementation. This is a transport-agnostic client surface. Additionally, internal, supporting types like `settings` and `storageOption` are added to help the client layer convey API call-related configuration (e.g. retry config).


storage/client.go
```diff
@@ -0,0 +1,145 @@
...
+
+package storage
+
+import (
+	"context"
+
+	gax "github.com/googleapis/gax-go/v2"
+)
+
+// storageClient is an internal-only interface designed to separate the
```
tritone, line 23: Add a TODO to note that this is not yet used-- it may be confusing for people trying to read this code otherwise.

Also, should we consider moving the public client constructors in storage.go to this file as well?

tritone (commented):
Generally looks good with a few comments. I assume we'll have to make further revisions once we try implementing some methods, but since this is unexported I guess we're not married to this surface.

storage/client.go
noahdietz: Done.

> should we consider moving the public client constructors in storage.go to this file as well?

Yeah I think that could be good.

noahdietz: I'm going to save moving things for another PR. Will add a TODO.

storage/client.go
```diff
@@ -0,0 +1,145 @@
...
+	UpdateObjectACL(ctx context.Context, bucket, object string, entity ACLEntity, role ACLRole, opts ...storageOption) (ACLRule, error)
+
+	// Media operations.
+
+	ComposeObject(ctx context.Context, req composeObjectRequest, opts ...storageOption) (*ObjectAttrs, error)
+	RewriteObject(ctx context.Context, req rewriteObjectRequest, opts ...storageOption) (rewriteObjectResponse, error)
+
+	OpenReader(ctx context.Context, r *Reader, opts ...storageOption) error
+	OpenWriter(ctx context.Context, w *Writer, opts ...storageOption) error
+}
```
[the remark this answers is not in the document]
noahdietz: I've added all of these, but I will need some help figuring out how to integrate the `HMACKeyOptions`, not really familiar with this interface. Added a TODO for that.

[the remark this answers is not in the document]
tritone: Hmm, it looks like the options are used to create a config object called [`hmacKeyDesc`](https://github.com/googleapis/google-cloud-go/blob/876964fe4b2ae2ba9ef862f749df3d6b38dcc638/storage/hmac.go#L394).

I think we could create this object from opts above the interface and then pass a pointer to it through to set configs on the call. Alternatively you could just pass through just the fields needed for a particular call, but probably simpler to just pass the config.

storage/client.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// Object ACL methods.
+
+	DeleteObjectACL(ctx context.Context, bucket, object string, entity ACLEntity, opts ...storageOption) error
+	ListObjectACLs(ctx context.Context, bucket, object string, opts ...storageOption) ([]ACLRule, error)
+	UpdateObjectACL(ctx context.Context, bucket, object string, entity ACLEntity, role ACLRole, opts ...storageOption) (ACLRule, error)
+
+	// Media operations.
+
+	ComposeObject(ctx context.Context, req composeObjectRequest, opts ...storageOption) (*ObjectAttrs, error)
```
tritone, line 81: Also, maybe better to pass a pointer to these new request structs rather than copying?

noahdietz: Good catch, done.

storage/client.go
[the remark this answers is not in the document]
noahdietz: Yep, passing `*hmacKeyDesc` in. Thanks Chris.

PR #5433, opened 2022-02-02
chore(compute): update smoke_tests with operation Wait helper

noahdietz:
Removes explicit use of the operation clients and switches to using the operation wrapper helper for `Wait` to work with operations. Also adds some failed test exits where it would cause a weird panic or error if the test continued to execute. LMK if I should use `t.Fatal` instead.

### DIFF
--- a/compute/apiv1/smoke_test.go
+++ b/compute/apiv1/smoke_test.go
@@ -54,10 +54,6 @@ func TestCreateGetPutPatchListInstance(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	zonesClient, err := NewZoneOperationsRESTClient(ctx)
-	if err != nil {
-		t.Fatal(err)
-	}
 	createRequest := &computepb.InsertInstanceRequest{
 		Project: projectId,
 		Zone:    defaultZone,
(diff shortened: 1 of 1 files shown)


compute/apiv1/smoke_test.go
```diff
@@ -92,16 +88,12 @@ func TestCreateGetPutPatchListInstance(t *testing.T) {
...
-		Project:   projectId,
-		Zone:      defaultZone,
-		Operation: insert.Proto().GetName(),
-	}
-	_, err = zonesClient.Wait(ctx, waitZonalRequest)
+	err = insert.Wait(ctx)
+	defer ForceDeleteInstance(ctx, name, c)
 	if err != nil {
 		t.Error(err)
+		t.FailNow()
```
codyoss, line 95: nit: there are a couple of these spots, they can be summarized as `t.Fatal(err)`

codyoss (approved):
Just a small nit, else looks good.

compute/apiv1/smoke_test.go
noahdietz: Haha sorry had auto-merge on. I will follow up :)

Issue #5428, opened 2022-02-02
bigquery/storage/managedwriter: investigate possible panic


shollyman:
Mentioned in an old PR wiring in the flow controller by @donners:

https://github.com/googleapis/google-cloud-go/blob/main/bigquery/storage/managedwriter/managed_stream.go#L353

> @shollyman Is there a missing return here? We've observed the following panic, which seems like it could be related to markDone resetting pw.request above before the append call proceeds below.

```runtime error: invalid memory address or nil pointer dereference
at cloud.google.com/go/bigquery/storage/managedwriter.(*ManagedStream).append ( /go/src/github.com/ScriptRock/webscan-golang/vendor/cloud.google.com/go/bigquery/storage/managedwriter/managed_stream.go:284 )
at cloud.google.com/go/bigquery/storage/managedwriter.(*ManagedStream).AppendRows ( /go/src/github.com/ScriptRock/webscan-golang/vendor/cloud.google.com/go/bigquery/storage/managedwriter/managed_stream.go:363 )
```

shollyman:
Fix out for review.  @donners thanks for the report!

donners:
Awesome thanks for the quick fix!

PR #5436, opened 2022-02-03
fix(bigquery/storage/managedwriter): possible panic due to flowcontrol

shollyman:
An exhausted flow controller will emit error rather than block.  When
this occured, we were marking the write done, but allowing the append to
proceed, getting us into another double-close situation, which yields a
panic.

Fixes: https://github.com/googleapis/google-cloud-go/issues/5428


codyoss (approved):
LGTM

PR #5439, opened 2022-02-03
fix(compute/metadata): init a HTTP client in OnGCE check

codyoss:
Initialize a resolver and http.Client locally in function to force
GC to cleanup background resources such as cached connections. This
can falsely report as leaked goroutines otherwise. This kind of issue
will still persist for calling methods like ProjectID, but we are
intentionally caching a client and its connections for these occasions
today. If this ends up causing users issues in the future we can
reevaluate at that time.

Fixes: #5430


codyoss:
Note: this is a hunch kinda fix, but I have ran into errors like this in the past and I am pretty sure this was the solution.

noahdietz (approved):
After reading the issue, this fix makes sense to me. Initializing a client in the scope of the function should result in the client being tagged for GC clean up once the function is finished.

PR #5442, opened 2022-02-03
test(spanner): enable more tests on emulator

olavloite:
The emulator now supports all data types that are also supported by
Cloud Spanner. The integration tests also unnecessarily skipped a large
number of tests that tried to insert untyped null values as it was
assumed that this was not supported. That is however supported for most
data types, and only DATE and TIMESTAMP need to be skipped.

Updates #https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/31


spanner/integration_test.go
```diff
@@ -1569,23 +1569,21 @@ func TestIntegration_BasicTypes(t *testing.T) {
 
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	defer cancel()
-	stmts := singerDBStatements
```
olavloite, line 1572: GitHub renders this in an almost unreadable way, but what it does is it removes the `if !isEmulatorEnvSet()` block and always uses the same data model for both the emulator and real Spanner. The if-block was previously needed as the emulator did not support `JSON`.

spanner/integration_test.go
```diff
@@ -1647,7 +1644,6 @@ func TestIntegration_BasicTypes(t *testing.T) {
 		{col: "String", val: "foo", want: NullString{"foo", true}},
 		{col: "String", val: NullString{"bar", true}, want: "bar"},
 		{col: "String", val: NullString{"bar", false}, want: NullString{"", false}},
-		{col: "String", val: nil, want: NullString{}},
```
olavloite, line 1650: Removed as it was a duplicate (the same test is on line 1737). This also applies to a couple of other tests below.

spanner/emulator_test.sh
```diff
@@ -24,7 +24,8 @@ export GCLOUD_TESTS_GOLANG_PROJECT_ID=emulator-test-project
 echo "Running the Cloud Spanner emulator: $SPANNER_EMULATOR_HOST";
 
 # Download the emulator
-EMULATOR_VERSION=1.2.0
+# TODO: Find a way to use 'latest' here.
+EMULATOR_VERSION=1.4.0
```
snehashah16, line 28: isnt there a 1.4.1 version ?

spanner/integration_test.go
snehashah16: ack, I made this exact change in my repo last night so I know what u mean here.. thank u for clarifying

snehashah16 (approved):

rahul2393 (approved):

spanner/emulator_test.sh
olavloite: You're right. I've bumped the version to 1.4.1.

olavloite:
Build error is caused by a missing library on the test server that seems to be required by the latest version of the emulator:

`./emulator_main: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by ./emulator_main)`

PR #5454, opened 2022-02-04
feat(bigquery/storage/managedwriter): support schema tracking

shollyman:
This PR attempts to provide a mechanism to make it easier to get the
current schema for a given stream.  It registers schema when calling
CreateWriteStream on behalf of users, listens for UpdatedSchema messages
in AppendRowsResponse messages, and also leverages GetWriteStream for
explicit checking.


shollyman:
Closing this, this depends on the API features not expected anytime soon.

PR #5456, opened 2022-02-04
build: enable release-please app with manifest config

chingor13:
This configuration allows release-please to more efficiently propose pull releases from the GitHub app.

Requires: https://github.com/googleapis/release-please/pull/1287


.github/release-please.yml
```diff
@@ -0,0 +1,9 @@
+handleGHRelease: true
```
codyoss, line 1: This will use the app now right? If so, could you remove the two release please actions.

.release-please-manifest-submodules.json
```diff
@@ -0,0 +1,111 @@
+{
+  "accessapproval": "1.0.1",
```
noahdietz, line 2: Will this need to be manually updated when a new submodule is added? Or is this something that is auto-updated?

release-please-config-submodules.json
```diff
@@ -0,0 +1,335 @@
+{
+  "release-type": "go-yoshi",
+  "separate-pull-requests": true,
+  "include-component-in-tag": true,
+  "tag-separator": "/",
+  "packages": {
```
noahdietz, line 6: Ditto for management of this set.

.release-please-manifest-submodules.json
chingor13: Yes, this will need to be updated for new modules.

release-please-config-submodules.json
chingor13: Yes, this will need to be updated for new modules.

Issue #5459, opened 2022-02-06
bigquery: Possible deadlock in managedwriter


jsabbatini-upguard:
**Client**

BigQuery Storage API - ManagedWriter

**Environment**

Alpine Docker on GKE

**Go Environment**

go version go1.17.2

**Code**

e.g.
```go
result, err := stream.AppendRows(ctx, encoded)
if err != nil {
    return errors.Wrap(err, "failed to append rows to BigQuery stream")
}

_, err = result.GetResult(ctx)
if err != nil {
    return errors.Wrap(err, "failed to get BigQuery result")
}
```

**Expected behavior**

Rows should be appended and written to BigQuery.

**Actual behavior**

We noticed that in certain situations the managed writer seems to enter deadlock. 

Goroutine analysis seems to point to the fact that:
1. The `recvProcessor` is blocked trying to receive data from the `storagepb.BigQueryWrite_AppendRowsClient`, https://github.com/googleapis/google-cloud-go/blob/main/bigquery/storage/managedwriter/managed_stream.go#L397
2.  The `append` method of `ManagedStream` is blocked trying to add a new `pendingWrite` onto the pending channel, https://github.com/googleapis/google-cloud-go/blob/main/bigquery/storage/managedwriter/managed_stream.go#L303

Beside the deadlock, the context passed to the `AppendRows` method is not used in the `append` method of `ManagedStream` and that method does not check for context expiry when adding a `pendingWrite` to the `ch` channel so it is possible for the overall call to `AppendRows` to extend well beyond a context expiry time.

shollyman:
Thanks for the report.  I'll get this addressed.

shollyman:
Sent a prospective fix PR out for review, feel free to chime in if you have feedback.

jameshartig:
We're noticing a deadlock as well but I'm not seeing it block on `append`. We have the following 2 goroutines that are blocked:

```
1 @ 0x43a2f6 0x44a0d2 0xa599ea 0xa59727 0xa5a5f2 0x4cc4da 0xa5a565 0xa5a53b 0xabc067 0xabc8c6 0xabcfae 0xacd050 0xacc405 0xacb296 0xacc13f 0xadadac 0xb65614 0x46ac81
#       0xa599e9        google.golang.org/grpc/internal/transport.(*recvBufferReader).readClient+0xa9                                   google.golang.org/grpc@v1.43.0/internal/transport/transport.go:190
[...]

PR #5465, opened 2022-02-07
feat(bigquery/storage/managedwriter): append improvements

shollyman:
Fixes: https://github.com/googleapis/google-cloud-go/issues/5459

This PR addresses two issues.  The receiver channel for processing
asyncronous updates is switched to a buffer channel, based on the
allowed append depth.

The second change is that this allows for better context
expiry/cancellation when invoking AppendRows on a managed stream.

This also improves testing with some test refactors, as well as
shaking out some timing issues due the larger queue depth.

### DIFF
--- a/bigquery/storage/managedwriter/integration_test.go
+++ b/bigquery/storage/managedwriter/integration_test.go
@@ -211,7 +211,14 @@ func testDefaultStream(ctx context.Context, t *testing.T, mwClient *Client, bqCl
 	}
 	// wait for the result to indicate ready, then validate again.  Our total rows have increased, but
 	// cardinality should not.
-	result.Ready()
+	// wait for the result to indicate ready, then validate.
+	o, err := result.GetResult(ctx)
+	if err != nil {
+		t.Errorf("result error for last send: %v", err)
+	}
+	if o != NoStreamOffset {
+		t.Errorf("offset mismatch, got %d want %d", o, NoStreamOffset)
+	}
 	validateTableConstraints(ctx, t, bqClient, testTable, "after second send round",
 		withExactRowCount(int64(2*len(testSimpleData))),
 		withDistinctValues("name", int64(len(testSimpleData))),
@@ -267,7 +274,13 @@ func testDefaultStreamDynamicJSON(ctx context.Context, t *testing.T, mwClient *C
 	}
 
 	// wait for the result to indicate ready, then validate.
-	result.Ready()
+	o, err := result.GetResult(ctx)
+	if err != nil {
+		t.Errorf("result error for last send: %v", err)
+	}
+	if o != NoStreamOffset {
+		t.Errorf("offset mismatch, got %d want %d", o, NoStreamOffset)
+	}
 	validateTableConstraints(ctx, t, bqClient, testTable, "after send",
 		withExactRowCount(int64(len(sampleJSONData))),
 		withDistinctValues("name", int64(len(sampleJSONData))),
--- a/bigquery/storage/managedwriter/managed_stream.go
+++ b/bigquery/storage/managedwriter/managed_stream.go
@@ -229,7 +229,13 @@ func (ms *ManagedStream) openWithRetry() (storagepb.BigQueryWrite_AppendRowsClie
 		if err == nil {
 			// The channel relationship with its ARC is 1:1.  If we get a new ARC, create a new chan
 			// and fire up the associated receive processor.
-			ch := make(chan *pendingWrite)
+			depth := 1000 // default backend queue limit
+			if ms.streamSettings != nil {
+				if ms.streamSettings.MaxInflightRequests > 0 {
+					depth = ms.streamSettings.MaxInflightRequests
+				}
+			}
+			ch := make(chan *pendingWrite, depth)
 			go recvProcessor(ms.ctx, arc, ms.fc, ch)
 			// Also, replace the sync.Once for setting up a new stream, as we need to do "special" work
 			// for every new connection.
@@ -360,11 +366,29 @@ func (ms *ManagedStream) AppendRows(ctx context.Context, data [][]byte, opts ...
 			ProtoDescriptor: pw.newSchema,
 		}
 	}
-	// proceed to call
-	if err := ms.append(pw); err != nil {
-		// pending write is DOA.
-		pw.markDone(NoStreamOffset, err, ms.fc)
-		return nil, err
+	// Call the underlying append.  The stream has it's own retained context and will surface expiry on
+	// it's own, but we also need to respect any deadline for the provided context.
+	errCh := make(chan error)
+	var appendErr error
+	go func() {
+		err := ms.append(pw)
+		errCh <- err
+		close(errCh)
+	}()
+	select {
+	case <-ctx.Done():
+		// It is incorrect to simply mark the request done, as it's potentially in flight in the bidi stream
+		// where we can't propagate a cancellation.  Our options are to return the pending write even though
+		// it's in an ambiguous state, or to return the error and simply drop the pending write on the floor.
+		//
+		// This API expresses request idempotency through offset management, so users who care to use offsets
+		// can deal with the dropped request.
+		return nil, ctx.Err()
+	case appendErr = <-errCh:
+		if appendErr != nil {
+			pw.markDone(NoStreamOffset, appendErr, ms.fc)
+			return nil, appendErr
+		}
 	}
 	return pw.result, nil
 }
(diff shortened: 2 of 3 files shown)


bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -360,11 +366,29 @@ func (ms *ManagedStream) AppendRows(ctx context.Context, data [][]byte, opts ...
...
 		}
 	}
-	// proceed to call
-	if err := ms.append(pw); err != nil {
-		// pending write is DOA.
-		pw.markDone(NoStreamOffset, err, ms.fc)
-		return nil, err
+	// Call the underlying append.  The stream has it's own retained context and will surface expiry on
+	// it's own, but we also need to respect any deadline for the provided context.
+	errCh := make(chan error)
```
codyoss, line 371: nit: Not sure if you import or use it already but this looks like a great usecase for: [errorgroup](https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext)

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -229,7 +229,13 @@ func (ms *ManagedStream) openWithRetry() (storagepb.BigQueryWrite_AppendRowsClie
 		if err == nil {
 			// The channel relationship with its ARC is 1:1.  If we get a new ARC, create a new chan
 			// and fire up the associated receive processor.
-			ch := make(chan *pendingWrite)
+			depth := 1000 // default backend queue limit
```
codyoss, line 232: Does each call to receive guaranteed to only move the offset by one? Lets say the offset is N. We send two request to append rows before we receive. Will receive always be N+1 then N+2? Or could it bump to N+2 right away? Just want to make sure we will not build up too many requests that can never be processed with this new paradigm.

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -360,11 +366,29 @@ func (ms *ManagedStream) AppendRows(ctx context.Context, data [][]byte, opts ...
...
-	// proceed to call
-	if err := ms.append(pw); err != nil {
-		// pending write is DOA.
-		pw.markDone(NoStreamOffset, err, ms.fc)
-		return nil, err
+	// Call the underlying append.  The stream has it's own retained context and will surface expiry on
+	// it's own, but we also need to respect any deadline for the provided context.
+	errCh := make(chan error)
+	var appendErr error
+	go func() {
```
jsabbatini-upguard, line 373: There is the potential here that `append` will block waiting to add a pending write to the pending writes channel. When that happens the context below could expire returning an error from `AppendRows`. However this goroutine is still running and as soon as the pending channel is freed the `append` will in fact succeed even if an error was returned.

Furthermore this goroutine not being stopped when the caller context expires means that if `append` becomes slow due a full pending write channel, rapid calls to `AppendRows`, even with a short timeouts, could result in a very large number of gorountines left running.

bigquery/storage/managedwriter/managed_stream.go
shollyman: its based on the rows, not the number of calls.  append can send N rows per call.

Example:
*  ​append 4 rows @ offset 0: offset is now 4 (0,1,2,3 written)
* append 2 rows @ offset 0: error, already exists
* append 2 rows @ offset 4: offset is now 6

bigquery/storage/managedwriter/managed_stream.go
shollyman: Good point.  I was reasoning through backend behaviors, didn't think through this scenario.

shollyman: Spend some time refactoring this, as there were other correctness issues as well primarily around potential races.

bigquery/storage/managedwriter/managed_stream.go
shollyman: This particular use case is handling a single request on a bidi stream and supporting cancellation, but it's certainly something we may end up with if we end up with write multiplexing.

bigquery/storage/managedwriter/managed_stream.go
shollyman: Also, added a test explicitly to watch for goroutine leak.

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -272,36 +285,53 @@ func (ms *ManagedStream) append(pw *pendingWrite, opts ...gax.CallOption) error
...
-			ms.mu.Unlock()
+		} else {
+			// Subsequent requests need no modification.
+			err = (*arc).Send(pw.request)
 		}
 		if err == nil {
+			// Compute numRows, once we pass ownership to the channel the request may be
+			// cleared.
+			numRows := int64(len(pw.request.GetProtoRows().Rows.GetSerializedRows()))
 			ch <- pw
```
jsabbatini-upguard, line 302: Should this guard against the `otherCtx` expiring as well?

Something like 

```go
select {
    case ch <- pw:
        // We've passed ownership of the pending write to the channel.
        // It's now responsible for marking the request done, we're done
        // with the critical section.
        ms.mu.Unlock()

        // Record stats and return.
        recordStat(ms.ctx, AppendRequests, 1)
        recordStat(ms.ctx, AppendRequestBytes, int64(pw.reqSize))
        recordStat(ms.ctx, AppendRequestRows, numRows)
        return nil
  case <-otherCtx.Done():
       ms.mu.Unlock()
       return otherCtx.Err()
}
```

That way the call to `append` is guaranteed not to lock if `ch` is full and it is guaranteed to return within the given context timeout (within some margins).

And once calls to `append` are guaranteed to return within the timeout the goroutine in `AppendRows` could probably be removed and replaced with a synchronous call I believe.

shollyman: The issue is that the select allows the writer to get out of sync processing append responses, which arrive in the same order as append requests were accepted.  We use the channel for maintaining the ordering when processing the responses.

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -240,7 +246,7 @@ func (ms *ManagedStream) openWithRetry() (storagepb.BigQueryWrite_AppendRowsClie
 	}
 }
 
-func (ms *ManagedStream) append(pw *pendingWrite, opts ...gax.CallOption) error {
+func (ms *ManagedStream) append(otherCtx context.Context, pw *pendingWrite, opts ...gax.CallOption) error {
```
codyoss, line 249: Would you mind adding a comment here about how otherCtx is to be used and the difference between this and ms.ctx. Just want to avoid potential future misuse.

codyoss (approved):

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -229,7 +229,13 @@ func (ms *ManagedStream) openWithRetry() (storagepb.BigQueryWrite_AppendRowsClie
...
 			// The channel relationship with its ARC is 1:1.  If we get a new ARC, create a new chan
 			// and fire up the associated receive processor.
-			ch := make(chan *pendingWrite)
+			depth := 1000 // default backend queue limit
+			if ms.streamSettings != nil {
+				if ms.streamSettings.MaxInflightRequests > 0 {
+					depth = ms.streamSettings.MaxInflightRequests
+				}
+			}
+			ch := make(chan *pendingWrite, depth)
```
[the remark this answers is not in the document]
shollyman: Yeah, we have a flow controller already wired in, but the unbuffered channel bounds the queue length to 1.  This is the "how many outstanding writes are we waiting for responses on".

bigquery/storage/managedwriter/integration_test.go
```diff
@@ -211,7 +217,14 @@ func testDefaultStream(ctx context.Context, t *testing.T, mwClient *Client, bqCl
 	}
 	// wait for the result to indicate ready, then validate again.  Our total rows have increased, but
 	// cardinality should not.
-	result.Ready()
+	// wait for the result to indicate ready, then validate.
```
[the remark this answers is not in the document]
shollyman: Cleaned this up with some other minor comment issues.

bigquery/storage/managedwriter/managed_stream.go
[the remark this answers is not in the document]
shollyman: Augmented the comments for this to help add breadcrumbs on whats going on.

tswast (approved):

PR #5467, opened 2022-02-07
feat(video/transcoder): remove apiv1beta1, turned down

codyoss:


noahdietz (approved):
Repo metadata json will get updated on its own, right?

codyoss:
Yup, I was just lazy :P Will update automatically /w next regen

PR #5482, opened 2022-02-09
feat(storage): add Writer.ChunkRetryDeadline

tritone:
Expose the ChunkRetryDeadline MediaOption through the manual
client layer. This allows users to set a longer deadline for
chunk retries in resumable uploads if desired.

I also considered this as a RetryOption, but 1. it's only
relevant for uploads and 2. it should be configured in
conjunction with ChunkSize which is a field on Writer.

Updates googleapis/google-api-go-client#685


tritone:
Presubmits seem to be failing for everything due to issues with the new Go 1.15 kokoro build; @codyoss PTAL

storage/writer.go
```diff
@@ -76,6 +77,23 @@ type Writer struct {
...
+	//
+	// For uploads of larger files, the Writer will attempt to retry if the
+	// request to upload a particular chunk fails with a transient error.
+	// If a single chunk has been attempting to upload for longer than this
+	// deadline and the request fails, it will no longer be retried, and the error
+	// will be returned to the caller. This is only applicable for files which are
+	// large enough to require a multi-chunk resumable upload. The default value
+	// is 32s. Users may want to pick a longer deadline if they are using larger
+	// values for ChunkSize or if they expect to have a slow or unreliable
+	// internet connection.
```
BrennaEpp, line 91: Should we include this info elsewhere as well (in Retry docs and/or samples)?

BrennaEpp (approved):
Awesome!

PR #5511, opened 2022-02-10
feat(compute/metadata): add HTTPClient Interface for metadata client

singku:
Add HTTP Client interface to metadata client so that consumers can mock http client behaviors and test the correctness of metadata client, e.g., acquire GCE metadata properties.

### DIFF
--- a/compute/metadata/metadata.go
+++ b/compute/metadata/metadata.go
@@ -61,7 +61,8 @@ var (
 	instID  = &cachedValue{k: "instance/id", trim: true}
 )
 
-var defaultClient = &Client{hc: newDefaultHTTPClient()}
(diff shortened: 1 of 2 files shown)


tbpg (commented):
Is there an open issue with a discussion about this? You can define interfaces in calling code; they don't need to be defined here. In particular, I don't think we want to export/support `HTTPClient`, `SetHTTPClient`, or `ResetToDefaultHTTPClient`.

If you file an issue with your goal and what you've tried so far, we can find the right solution.

singku:
It looks like the code used in our code base is imported and old, we can't inject a client to perform what we need. And here is this repo the client is injectable. I will close this PR then. Thanks.

PR #5617, opened 2022-02-12
feat(storage): allow specifying includeTrailingDelimiter

gaul:
References fsouza/fake-gcs-server#676.


storage/storage.go
```diff
@@ -1588,6 +1588,11 @@ type Query struct {
 	// which returns all properties. Passing ProjectionNoACL will omit Owner and ACL,
 	// which may improve performance when listing many objects.
 	Projection Projection
+
+	// If true, objects that end in exactly one instance of delimiter have
```
tritone, line 1592: I see this is copied from the API docs, but it should be rephrased based on how this is exposed via ObjectIterator in this library and based on godoc standards. I would say something like:

> IncludeTrailingDelimiter controls how objects which end in a single instance of Delimiter (for example, if Query.Delimiter = "/" and the object name is "foo/bar/") are included in the results. By default, these objects only show up as prefixes. If IncludeTrailingDelimiter is set to true, they will also be included as objects and their metadata will be populated in the returned ObjectAttrs.

storage/storage.go
```diff
@@ -1588,6 +1588,11 @@ type Query struct {
...
+	// their metadata included in items[] in addition to the relevant part of
+	// the object name appearing in prefixes[].
+	IncludeTrailingDelimiter bool
 }
```
tritone, line 1596: Can you also write an integration test to verify that this behaves as expected with real objects? Example for Query.Projection is here: https://github.com/googleapis/google-cloud-go/blob/9f79ac4f5ff32fc99fb3d49ec59cf7dd87ce21f9/storage/integration_test.go#L1902

tritone (commented):
Thanks for the contribution! A couple comments.

storage/storage.go
gaul: I read through the [integration test scrolls](https://github.com/googleapis/google-cloud-go/blob/main/CONTRIBUTING.md#testing) but the evil wizard frightened me.  I tried again with some magical potions 🍷 but this quest seems to require some serious XP.  Is there a cheat code to avoid grinding?

gaul: Added test.

storage/storage.go
gaul: Done.

tritone (approved):
LGTM, thanks very much for the contribution!

PR #5618, opened 2022-02-13
fix(pubsublite): use pubsublite internal version

tmdiep:
Pass the version number in `pubsublite/internal/version.go` to the server in stream metadata.


codyoss (approved):
nice!

tmdiep:
@codyoss Thanks for providing the internal version!

PR #5619, opened 2022-02-13
docs(pubsublite): more updates for regional topics

tmdiep:
A few more outdated comments about zonal topics.


anguillanneuf (approved):
Thank you!

PR #5620, opened 2022-02-14
test(profiler): make integration test use Go 1.17.7

amchiclet:
Successfully built profiler/integration_test.go locally with Go 1.17.7.


nolanmar511 (approved):

profiler/kokoro/integration_test.sh
```diff
@@ -33,6 +33,15 @@ cd $(dirname $0)/..
 export GOOGLE_APPLICATION_CREDENTIALS="${KOKORO_KEYSTORE_DIR}/72935_cloud-profiler-e2e-service-account-key"
 export GCLOUD_TESTS_GOLANG_PROJECT_ID="cloud-profiler-e2e"
 
+# Ensure a newer version of Go is used so it is compatible with newer libraries.
```
codyoss, line 36: We always test on the lastest version of go and a min version. Instead of installing a version of Go in the test you could check if it is 1.17 or greater in this script.

codyoss: @amchiclet

PR #5655, opened 2022-02-14
chore: bulk release yoshi managed libs

codyoss:
All yoshi managed libs will be released as one PR. Modules owned
by other teams will continue to be seperate PRs. This will cut down
repo noise and cause less pressure on CI systems


noahdietz (approved):
Great idea, folks.

.release-please-manifest-submodules.json
```diff
@@ -1,127 +1,13 @@
 {
-  "accessapproval": "1.1.0",
```
chingor13, line 2: non-blocking: I _think_ we can point both configs to the same version file

chingor13 (changes requested):
You'll need to add an entry to `.github/release-please.yml` to tell it to find the new config file.

Also, after https://github.com/googleapis/release-please/pull/1300 rolls out, we can combine the root module one with either of the other configs.

PR #5666, opened 2022-02-16
feat(bigquery): support authorized datasets

shollyman:
This augments what you can express in a dataset ACL, and can include a
dataset with a set of allowed resource types.  In short, this allows for
simplifying authorized views by allowing an entry that says "all views
in this dataset are authorized" vs the current behavior, which requires
views to be enumerated individually.

Also addresses a gotcha in that client was plumbed all the way to the top
of the dataset metadata converter chain, which is needed as in Go references
like dataset and table retain a local reference to the client.


bigquery/dataset_test.go
```diff
@@ -325,7 +327,16 @@ func TestDatasetToBQ(t *testing.T) {
 			},
 			Location: "EU",
 			Labels:   map[string]string{"x": "y"},
-			Access:   []*AccessEntry{{Role: OwnerRole, Entity: "example.com", EntityType: DomainEntity}},
+			Access: []*AccessEntry{
+				{Role: OwnerRole, Entity: "example.com", EntityType: DomainEntity},
+				{Role: ReaderRole,
```
tswast, line 332: I see in https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#Dataset.FIELDS.inlinedField_8 that "The role field is not required when this field is set. "

Do we want a test that verifies role can be omitted from Go?

shollyman: Work adjusting the unit test expectations, good spot.

shollyman:
Blocked on internal issue 211633697

shollyman:
necessary discovery fixes present in google.golang.org/api v0.71.0 or later

PR #5674, opened 2022-02-18
chore(internal/gapicgen): fix logging yaml name

noahdietz:
The serivce config yaml for logging was renamed to conform, so we must rename it in the generation config.


codyoss (approved):

codyoss:
Rolled this into #5687, closinig.

PR #5685, opened 2022-02-22
chore(storage): stub out grpc/json interfaces with factory methods

noahdietz:
This PR does a couple of things to the internal `storageClient` interface:
* changes any `struct` value to a pointer in the interface (e.g. `ACLRule` -> `*ACLRule`)
* adds a `clientOption` property to `settings` for capturing the `option.ClientOption` used during client init
* adds `initSettings` and `resolveOptions` helpers for applying `storageOption` to `settings`
* adds both `jsonStorageClient` and `grpcStorageClient` stub implementations of `storageClient`
* adds (internal) factory methods for both implementations


BrennaEpp:
Looking good! 
Just a concern, why is it necessary to change struct values to pointers?

noahdietz:
> why is it necessary to change struct values to pointers

It's not necessary, per say, but it helps avoid accidental copies being made if the return value is assigned to multiple different variables for example.

noahdietz:
@tritone I've added a sanity check test for the client factory methods.

storage/grpc_client.go
```diff
@@ -0,0 +1,217 @@
...
+	defaults := []option.ClientOption{
+		option.WithGRPCConnectionPool(defaultConnPoolSize),
+	}
+
+	// Set emulator options for gRPC if an emulator was specified. Note that in a
+	// hybrid client, STORAGE_EMULATOR_HOST will set the host to use for HTTP and
+	// STORAGE_EMULATOR_HOST_GRPC will set the host to use for gRPC (when using a
+	// local emulator, HTTP and gRPC must use different ports, so this is
+	// necessary).
+	// TODO: when full gRPC client is available, remove STORAGE_EMULATOR_HOST_GRPC
```
tritone, line 48: I guess we can do this TODO once we have removed NewHybridClient

storage/grpc_client.go
```diff
@@ -0,0 +1,217 @@
...
+// Note: This is experimental and should not be used in production yet.
+type grpcStorageClient struct {
+	raw      *gapic.Client
+	settings *settings
+}
+
+// newGRPCStorageClient initializes a new storageClient that uses the gRPC
+// Storage API.
+//
+// Note: This is experimental and should not be used in production yet.
```
tritone, line 79: I liked the wording you had in the media ops: "// This is an experimental API and not intended for public use."

Let's keep it consistent so it's easy to grep and remove or edit all these tags.

storage/grpc_client.go
```diff
@@ -0,0 +1,217 @@
...
+	return nil, ErrMethodNotSupported
+}
+
+// HMAC Key methods.
+
+func (c *grpcStorageClient) GetHMACKey(ctx context.Context, desc *hmacKeyDesc, opts ...storageOption) (*HMACKey, error) {
+	return nil, ErrMethodNotSupported
+}
+func (c *grpcStorageClient) ListHMACKey(ctx context.Context, desc *hmacKeyDesc, opts ...storageOption) *HMACKeysIterator {
+	return nil
```
tritone, line 207: Let's return an empty iterator rather than nil, I think this would make the example panic.

storage/json_client.go
```diff
@@ -0,0 +1,251 @@
...
+)
+
+// jsonStorageClient is the HTTP-JSON API implementation of the transport-agnostic
+// storageClient interface.
+//
+// Note: This is experimental and should not be used in production yet.
+type jsonStorageClient struct {
+	creds    *google.Credentials
+	hc       *http.Client
+	host     string
```
tritone, line 41: Should be readHost (only used for the reader).

storage/storage.go
```diff
@@ -66,6 +65,8 @@ var (
 	ErrBucketNotExist = errors.New("storage: bucket doesn't exist")
 	// ErrObjectNotExist indicates that the object does not exist.
 	ErrObjectNotExist = errors.New("storage: object doesn't exist")
+	// ErrMethodNotSupported indicates that the method called is not currently supported by the client.
+	ErrMethodNotSupported = errors.New("storage: method is not currently supported")
```
tritone, line 69: Unexported for now please.

storage/json_client.go
```diff
@@ -0,0 +1,251 @@
...
+	"google.golang.org/api/transport"
+	htransport "google.golang.org/api/transport/http"
+	iampb "google.golang.org/genproto/googleapis/iam/v1"
+)
+
+// jsonStorageClient is the HTTP-JSON API implementation of the transport-agnostic
+// storageClient interface.
+//
+// Note: This is experimental and should not be used in production yet.
+type jsonStorageClient struct {
```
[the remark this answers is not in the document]
noahdietz: Good call. Going to rename the file as well.

storage/json_client.go
noahdietz: Done.

storage/storage.go
noahdietz: oops thanks, done.

storage/grpc_client.go
noahdietz: Good idea, done.

storage/grpc_client.go
noahdietz: Cool, changed.

storage/client_test.go
```diff
@@ -28,11 +28,11 @@ func TestNewGRPCStorageClient(t *testing.T) {
 	}
 }
 
-func TestNewJSONStorageClient(t *testing.T) {
+func TestnewHTTPStorageClient(t *testing.T) {
```
tritone, line 31: nit: capital N in New

tritone (approved):
Thanks Noah!

PR #5700, opened 2022-02-25
feat(spanner): add support of PGNumeric with integration tests for PG dialect

rahul2393:
Two changes:

* PGNumeric support
* Integration tests with PG dialect


hengfengli (approved):
LGTM.

hengfengli:
@rahul2393 For the TestKey, can we fix the test instead of skipping it? Do we know why it fails?

hengfengli (changes requested):
Can we fix the TestKey test for PG?

hengfengli (approved):

PR #5701, opened 2022-02-26
feat(spanner): add database roles

varuncnaik:
Adds the ability to specify a database role on requests.


ansh0l:
Adding do not merge label to avoid accidental merges.

rahul2393 (commented):
LGTM, any reasons why we did not dd the role in BatchCreateSessions requests?

spanner/client.go
```diff
@@ -116,6 +116,8 @@ type ClientConfig struct {
 	// override the default values.
 	CallOptions *vkit.CallOptions
 
+	DatabaseRole string
```
hmehrotra, line 119: Add a comment.

"DatabaseRole specifies the role to be assumed for all operations on the database by this client."

spanner/client_test.go
```diff
@@ -960,6 +960,35 @@ func TestClient_ReadWriteTransaction_DoNotLeakSessionOnPanic(t *testing.T) {
...
+	server, client, teardown := setupMockedTestServerWithConfig(t, ClientConfig{SessionPoolConfig: sc, DatabaseRole: "test"})
+	defer teardown()
+
+	// Wait until all sessions have been created, so we know that those requests will not interfere with the test.
+	sp := client.idleSessions
+	waitFor(t, func() error {
+		sp.mu.Lock()
+		defer sp.mu.Unlock()
+		if uint64(sp.idleList.Len()) != 1 {
+			return fmt.Errorf("num open sessions mismatch.\nWant: %d\nGot: %d", sp.MinOpened, sp.numOpened)
```
hmehrotra, line 978: Standard unit test terminology is 'Expected' and 'Actual' (with Expected mentioned before Actual) instead of Got/Want, but I see existing tests are already using got/want, so this is fine.

Existing tests have also consistently specified Got before want. Let's do the same for consistency.

spanner/integration_test.go
```diff
@@ -2476,6 +2476,366 @@ func TestIntegration_TransactionRunner(t *testing.T) {
 	}
 }
 
+func TestIntegration_QueryWithRoles(t *testing.T) {
+	t.Parallel()
+
+	// Set up testing environment.
+	var (
+		row                 *Row
+		client, client2     *Client
```
hmehrotra, line 2485: Use clientWithRole instead of client2?

spanner/integration_test.go
```diff
@@ -2476,6 +2476,366 @@ func TestIntegration_TransactionRunner(t *testing.T) {
...
+	}
+}
+
+func TestIntegration_ReadWithRoles(t *testing.T) {
+	t.Parallel()
+
+	// Set up testing environment.
+	var (
+		row                 *Row
+		client, client2     *Client
```
hmehrotra, line 2583: Same as above. clientWithRole here and in below tests.

spanner/integration_test.go
```diff
@@ -2476,6 +2476,366 @@ func TestIntegration_TransactionRunner(t *testing.T) {
...
+		if err == nil {
+			t.Fatal("expected err, got nil")
+		}
+		if msg, ok := matchError(err, codes.PermissionDenied, test.errMsg); !ok {
+			t.Fatal(msg)
+		}
+	}
+}
+
+func TestIntegration_MutationWithRoles(t *testing.T) {
```
hmehrotra, line 2758: Could you also add a test for Insert/Delete mutation?

hmehrotra (commented):
Other TODO's 

1. Add tests that use DROP ROLE, REVOKE. (We don't need to test role inheritance, thats sufficiently covered in backend unit tests.

2. Implement/Add tests for ListDatabaseRoles.

spanner/client_test.go
olavloite: Sorry to wade in on this PR, but I feel that this one is a little too important to ignore: The idiomatic naming for testing actual/expected in Go is actually `got, want := ...`, and not `actual, expected := ...`.

The point about `Got` before `Want` is valid.

varuncnaik: I actually copied this line from a different test case in this file. But I agree, we should be consistent with other similar checks.

spanner/integration_test.go
varuncnaik: Added these to `TestIntegration_DMLWithRoles` and `TestIntegration_MutationWithRoles`.

varuncnaik:
> any reasons why we did not dd the role in BatchCreateSessions requests?

I did add it, it's part of `SessionTemplate`.

> Add tests that use DROP ROLE, REVOKE.

I added these to `TestIntegration_QueryWithRoles` and `TestIntegration_ReadWithRoles`.

I'll work on ListDatabaseRoles soon.

rahul2393 (approved):

PR #5706, opened 2022-02-28
chore(storage): update to go 1.15 chores

BrennaEpp:
- Remove build tags for go1.114 test
- Change xerrors to errors package for .As() and .Is()


codyoss (commented):
Woohoo!

storage/integration_test.go
```diff
@@ -3417,6 +3417,7 @@ func TestIntegration_UpdateRetentionExpirationTime(t *testing.T) {
 	h.mustWrite(obj.NewWriter(ctx), randomContents())
 
 	defer func() {
+		t.Helper()
```
tritone, line 3420: What effect does this have?

tritone (approved):

storage/integration_test.go
BrennaEpp: Helper marks the calling function as a test helper function. When printing file and line information, that function will be skipped. Basically it replaces `loc()` here  - I removed that now too

Issue #3613, opened 2021-01-26
bigquery: expose jobTimeoutMs field


jaketf:
**Is your feature request related to a problem? Please describe.**
The BigQuery API has a [jobTimeoutMs](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfiguration.FIELDS.job_timeout_ms:~:text=undefined.-,jobTimeoutMs,%5BOptional%5D%20Job%20timeout%20in%20milliseconds.%20If%20this%20time%20limit%20is%20exceeded%2C%20BigQuery%20might%20attempt%20to%20stop%20the%20job) 

**Describe the solution you'd like**
I want to revive an API similar to the one in this [PR](https://github.com/googleapis/google-cloud-go/pull/1392) closed for not being a gerrit CL?

**Describe alternatives you've considered**
Alternatively, client side one can implement a timer and use jobs.cancel.

shollyman:
I've deferred on adding this given the narrow use case:  you need to have strong understanding of the latency bounds of your work before starting the job, and you can't alter the deadline once the job is inserted.  It also implies the setter understands how other work using the same resources will impact the run time of the job.

Is there a specific idiom you're trying to support with this?

shollyman:
We've gotten additional feedback about supporting this, so I'll revive this once more.

PR #5707, opened 2022-03-01
feat(bigquery): add job timeout support

shollyman:
This feature allows callers to configure a job timeout.  If a BigQuery
job runs longer than this duration, BigQuery may cancel the job.

We expose JobTimeout in each of the job config types, given the existing pattern.

Fixes: https://github.com/googleapis/google-cloud-go/issues/3613

### DIFF
--- a/bigquery/copy.go
+++ b/bigquery/copy.go
@@ -16,6 +16,7 @@ package bigquery
 
 import (
 	"context"
+	"time"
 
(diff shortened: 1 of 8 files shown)


steffnay (approved):

Thomasdezeeuw:
@shollyman would it be possible to use the context deadline if `JobTimeout` is zero?

PR #5718, opened 2022-03-03
docs(pubsub): clarify ack/nack cannot be called from goroutine

hongalex:
For more information, see https://github.com/GoogleCloudPlatform/golang-samples/issues/2456


codyoss (approved):
A good call out, thanks for adding this to the docs!

PR #5720, opened 2022-03-03
feat(video/stitcher): start generating apiv1

quartzmo:


noahdietz:
I'll let Cody review/approve, but I think this your first one of these? Good job Chris! :)

quartzmo:
Uh oh, I left in a test assertion that I didn't want to commit. Fixing the build now.

video/stitcher/apiv1/doc.go
```diff
@@ -0,0 +1,152 @@
...
+	"google.golang.org/api/option"
+	"google.golang.org/grpc/metadata"
+)
+
+// For more information on implementing a client constructor hook, see
+// https://github.com/googleapis/google-cloud-go/wiki/Customizing-constructors.
+type clientHookParams struct{}
+type clientHook func(context.Context, clientHookParams) ([]option.ClientOption, error)
+
+const versionClient = "UNKNOWN"
```
codyoss, line 87: This appears to have been generated with an older version of the generator. Can you pull in the lastest version and update.

codyoss: @noahdietz I think protoc always generates in what version was used to generate the file as a comment. I wonder if we should do the same with the plugin. I only caught this because I had touched the generator recently.

codyoss (changes requested):
Just one thing in the code. Also, for the conventional commit use "feat(video/stitcher)". The video directory prefix is what will trigger release please to propose a release for the video module.

video/stitcher/apiv1/doc.go
noahdietz: Yeah that would be a good thing to have (the version of the generator used to generate the file). It would be part of the "Generated by..." comment at the top. We use release-please, but not using a version.go file. We could switch to that... @quartzmo want to get into the generator? 😈

codyoss:
This actually also makes me think we should really just have a job that we feed a couple variables to generate libraries... Maybe that is something work looking into in the future. The current approach is makes it so you really need an up-to-date local env...

video/stitcher/apiv1/doc.go
quartzmo: @codyoss I squashed and edited the original two commits to update the message to "feat(video/stitcher)". Then I re-ran generation with the most recent generator versions and committed the updates separately so you can see the diff.

codyoss (approved):

PR #5721, opened 2022-03-03
fix(datastore): initialize map fields in structs

jesper-lindberg:
This initializes map fields that are part of a struct. This solves an issue where trying to load data using the `Load` function of `PropertyLoadSaver` would fail as maps are not initialized prior to calling `Load`.


codyoss:
@jesper-lindberg Thanks for the PR! Would you mind opening an issue associated with this PR for discussion and showing the current issue. Also if this PR does move forward a test would be nice 😄

bhshkh:
@jesper-lindberg can you please link the issue that demonstrates the bug ?

bhshkh:
No reply from PR author. Closing the PR.
Please re-open if issue is seen.

PR #5725, opened 2022-03-04
docs(storage): clarify upload checksum docs

tritone:
I recently reviewed code for someone who tried setting this
field at the end of the upload rather than the beginning--
unfortunately, this causes the checksum to be ignored (and there
is no way to fix this since sending the checksum at the end of
an upload is not supported in the GCS API currently). We
should clarify the docs to include this warning.


codyoss (approved):

BrennaEpp (approved):
Thanks Chris!

PR #5726, opened 2022-03-04
fix(internal/gapicgen): add generation of internal/version.go for new client modules

quartzmo:
This was discovered attempting to generate certificatemanager.
New client generation failed with `apiv1/version.go: no such file or directory`

* Move `genVersionFile` before `genModule `.
* Fix path for generation of `apivX/version.go` file.
* Add template and creation of the client's `internal/version.go` file, conditional on `genModule`.


quartzmo:
Thanks for helping me with every step of this PR @codyoss !!!

codyoss (approved):
LGTM

PR #5728, opened 2022-03-04
fix(pubsub): prevent infinite retry with publishing invalid utf-8 chars

### LINKED ISSUE #5268: pubsub: infinite retry because of invalid utf-8 characters in attributes
**Client**

PubSub

**Go Environment**

$ go version 1.17.5

**Code**

Steps to reproduce:
```go
package foo_test

import (
	"context"
	"testing"

	"cloud.google.com/go/pubsub"
)

func TestPublish_Attribute_Invalid_UTF8(t *testing.T) {
	pc, err := pubsub.NewClient(context.Background(), "project")
	if err != nil {
		panic(err)
	}

	if _, err := pc.Topic("topic").Publish(context.Background(), &pubsub.Message{
		Data: []byte("foo"),
		Attributes: map[string]string{
			"attr": "a\xc5z",
		},
	}).Get(context.Background()); err != nil {
		t.Fatal(err)
	}
}
```

**Expected behavior**

Receive an error with details regarding invalid utf-8 characters in attributes.

**Actual behavior**

Retries forever until context is canceled.

**Additional context**

Looks related to https://github.com/googleapis/google-cloud-go/issues/2443

hongalex:
`INTERNAL` is a retryable code for `Publish` but we should not indefinitely retry this error code when message attributes contains invalid utf-8 characters, which the `grpc-go` library returns as an `INTERNAL` error.

Fixes #5268

### DIFF
--- a/pubsub/service.go
+++ b/pubsub/service.go
@@ -91,3 +91,24 @@ func (r *streamingPullRetryer) Retry(err error) (pause time.Duration, shouldRetr
 		return r.defaultRetryer.Retry(err)
 	}
 }
+
+type publishRetryer struct {
+	defaultRetryer gax.Retryer
+}
+
+func (r *publishRetryer) Retry(err error) (pause time.Duration, shouldRetry bool) {
+	s, ok := status.FromError(err)
+	if !ok { // call defaultRetryer so that its backoff can be used
+		return r.defaultRetryer.Retry(err)
+	}
+	switch s.Code() {
+	case codes.Internal:
+		invalid := strings.Contains(s.Message(), "string field contains invalid UTF-8")
+		if invalid {
+			return 0, false
+		}
+		return 0, true
+	default:
+		return 0, false
+	}
+}
--- a/pubsub/topic.go
+++ b/pubsub/topic.go
@@ -678,7 +678,8 @@ func (t *Topic) publishMessageBundle(ctx context.Context, bms []*bundledMessage)
 		res, err = t.c.pubc.Publish(ctx, &pb.PublishRequest{
 			Topic:    t.name,
 			Messages: pbMsgs,
-		}, gax.WithGRPCOptions(grpc.MaxCallSendMsgSize(maxSendRecvBytes)))
+		}, gax.WithGRPCOptions(grpc.MaxCallSendMsgSize(maxSendRecvBytes)),
+			gax.WithRetry(func() gax.Retryer { return &publishRetryer{} }))
 	}
 	end := time.Now()
 	if err != nil {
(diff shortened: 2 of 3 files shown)


pubsub/service.go
```diff
@@ -91,3 +91,23 @@ func (r *streamingPullRetryer) Retry(err error) (pause time.Duration, shouldRetr
...
+type publishRetryer struct {
+	defaultRetryer gax.Retryer
+}
+
+func (r *publishRetryer) Retry(err error) (pause time.Duration, shouldRetry bool) {
+	s, ok := status.FromError(err)
+	if !ok {
+		return r.defaultRetryer.Retry(err)
+	}
+	switch s.Code() {
```
codyoss, line 104: nit: This can be simplified to `if s.Code == codes.Internal && strings.Contains(...)`

codyoss (approved):
Just one minor nit, or else LGTM

hongalex:
It's cheaper to punt the behavior down to the gRPC level and have it bubble up the marshaling error, rather than add another check for every publish call to detect if there are UTF-8 characters.

Did this cause an issue for you recently?

AndreasBergmeier6176:
> It's cheaper to punt the behavior down to the gRPC level and have it bubble up the marshaling error, rather than add another check for every publish call to detect if there are UTF-8 characters.
>
> Did this cause an issue for you recently?

Well, it might be cheaper but figuring out which of the attributes is the actual problem is a PITA. Especially if you have multiple layers of software that modifies the attributes. If the gRPC level would at least report, what the failing attribute key was...

hongalex:
> If the gRPC level would at least report, what the failing attribute key was...

If you get back the `string field contains invalid UTF-8` error, consider iterating over the attributes and call `utf8.ValidString` to see which attribute failed.

AndreasBergmeier6176:
> > If the gRPC level would at least report, what the failing attribute key was...
> 
> If you get back the `string field contains invalid UTF-8` error, consider iterating over the attributes and call `utf8.ValidString` to see which attribute failed.

I think I would do that if the gRPC module would export said error. Currently nothing is preventing the gRPC implementation from just doing a rug-pull and use yet another error string.

hongalex:
Yeah 100% agree that catching this based on the error string is brittle (and is what we're currently doing as well in the library). Are you willing to open an issue against `grpc-go` for this issue (needing to export the error)?

AndreasBergmeier6176:
> Yeah 100% agree that catching this based on the error string is brittle (and is what we're currently doing as well in the library). Are you willing to open an issue against `grpc-go` for this issue (needing to export the error)?

See here: https://github.com/golang/protobuf/issues/1642

PR #5742, opened 2022-03-09
chore(internal/gapicgen): disable storage gapic retry

noahdietz:
We should disable generation of GAPIC retries for Storage as the GAPIC will be used inside of the Veneer that will instead handle the retry of API calls in a uniform way between transports.

### DIFF
--- a/internal/gapicgen/generator/config.go
+++ b/internal/gapicgen/generator/config.go
@@ -1401,11 +1401,12 @@ var microgenGapicConfigs = []*microgenConfig{
 		releaseLevel:          "ga",
 	},
 	{
-		inputDirectoryPath:    "google/storage/v2",
-		pkg:                   "storage",
-		importPath:            "cloud.google.com/go/storage/internal/apiv2",
-		gRPCServiceConfigPath: "storage_grpc_service_config.json",
-		apiServiceConfigPath:  "storage_v2.yaml",
-		releaseLevel:          "alpha",
+		inputDirectoryPath: "google/storage/v2",
+		pkg:                "storage",
+		importPath:         "cloud.google.com/go/storage/internal/apiv2",
+		// The retry configuration is explicitly excluded in order to disable
+		// GAPIC-level retries and allow the veneer layer to handle retries.
+		apiServiceConfigPath: "storage_v2.yaml",
+		releaseLevel:         "alpha",
 	},
 }


internal/gapicgen/generator/config.go
```diff
@@ -1401,11 +1401,12 @@ var microgenGapicConfigs = []*microgenConfig{
...
-		pkg:                   "storage",
-		importPath:            "cloud.google.com/go/storage/internal/apiv2",
-		gRPCServiceConfigPath: "storage_grpc_service_config.json",
-		apiServiceConfigPath:  "storage_v2.yaml",
-		releaseLevel:          "alpha",
+		inputDirectoryPath: "google/storage/v2",
+		pkg:                "storage",
+		importPath:         "cloud.google.com/go/storage/internal/apiv2",
+		// The retry configuration is explicitly excluded in order to disable
+		// GAPIC-level retries and allow the veneer layer to handle retries.
```
tritone, line 1408: Does gRPCServiceConfigPath ever have anything other than retry-related configs?

tritone (approved):
One question, otherwise LGTM

internal/gapicgen/generator/config.go
noahdietz: Nope not for GAPICs

tritone: Well, we don't set timeouts on most ops at the veneer layer, but I guess so.

noahdietz: So then even more reason to not have default timeouts because that is non-standard behavior relative to the client today :)

tritone (approved):

codyoss:
I wonder if instead of ignoring the defaults we just override the ones we want. We do a little reflection magic in pubsub and one or two other libs to get this effect: https://github.com/googleapis/google-cloud-go/blob/main/pubsub/pubsub.go#L62-L88. Thoughts?

noahdietz:
> I wonder if instead of ignoring the defaults we just override the ones we want. We do a little reflection magic in pubsub and one or two other libs to get this effect: https://github.com/googleapis/google-cloud-go/blob/main/pubsub/pubsub.go#L62-L88. Thoughts?

That is a different use case. That is attempting to allow users to modify the default (gapic) client options at client creation time. This is just that we want to disable GAPIC retries altogether all the time because the default & user-provided retry resolution is done in the Veneer, then passed down to the GAPIC (and JSON client) on each request.

codyoss (approved):

PR #5744, opened 2022-03-10
chore(storage): add bucket adapters

noahdietz:
Add adapters for Bucket-related types used for converting between proto type and client type. Most of this code was copied from the existing adapters for the Apiary types, and refactored to use the proto types.

Also adds dependency on pkg.go.dev/github.com/googleapis/go-type-adapters in order to convert between `time.Time` and `datepb.Date`.


storage/acl.go
```diff
@@ -361,6 +399,16 @@ func toBucketProjectTeam(p *raw.BucketAccessControlProjectTeam) *ProjectTeam {
 	}
 }
 
+func projectTeamFromProto(p *storagepb.ProjectTeam) *ProjectTeam {
```
BrennaEpp, line 402: This functions seems to be the same as the func `fromProtoToObjectProjectTeam ` on L422?

storage/acl.go
```diff
@@ -259,6 +259,14 @@ func toBucketACLRules(items []*raw.BucketAccessControl) []ACLRule {
 	return rs
 }
 
+func toBucketACLRulesFromProto(items []*storagepb.BucketAccessControl) []ACLRule {
```
BrennaEpp, line 262: I see some mix of `xxxFromProto`, `toXxxFromProto` and `fromProtoToXxx` in this file. Is there a reason for this or can we make them consistent?

storage/acl.go
noahdietz: You're right! Getting rid of `fromProtoToObjectProjectTeam` in favor of the generic method I've added.

storage/acl.go
```diff
@@ -351,6 +381,14 @@ func (r ACLRule) toProtoObjectAccessControl(bucket string) *storagepb.ObjectAcce
 	}
 }
 
+func (r ACLRule) toProtoBucketAccessControl() *storagepb.BucketAccessControl {
+	return &storagepb.BucketAccessControl{
+		Entity: string(r.Entity),
```
tritone, line 386: Is this because these are enums currently and will need to change after the breaking change?

storage/bucket.go
```diff
@@ -801,6 +805,36 @@ func newBucket(b *raw.Bucket) (*BucketAttrs, error) {
...
+	if b == nil {
+		return nil
+	}
+	return &BucketAttrs{
+		Name:                     parseBucketName(b.GetName()),
+		Location:                 b.GetLocation(),
+		MetaGeneration:           b.GetMetageneration(),
+		DefaultEventBasedHold:    b.GetDefaultEventBasedHold(),
+		StorageClass:             b.GetStorageClass(),
+		Created:                  b.GetCreateTime().AsTime(),
```
tritone, line 818: I am wondering if some of these get methods with chained method calls need nil checks. Based on my reading of the generated code and protobuf code it sounds like maybe not in this case, but I'm wondering if we have a general guarantee/contract.

https://github.com/googleapis/go-genproto/blob/main/googleapis/storage/v2/storage.pb.go#L4186 (no nil check)
https://github.com/protocolbuffers/protobuf-go/blob/v1.27.1/types/known/timestamppb/timestamp.pb.go#L206
https://github.com/protocolbuffers/protobuf-go/blob/b92717ecb630d4a4824b372bf98c729d87311a4d/types/known/durationpb/duration.pb.go#L283 (but underlying proto code seems to treat nil as 0).

storage/bucket.go
```diff
@@ -854,6 +888,62 @@ func (b *BucketAttrs) toRawBucket() *raw.Bucket {
...
+	if b.UniformBucketLevelAccess.Enabled || b.BucketPolicyOnly.Enabled || b.PublicAccessPrevention != PublicAccessPreventionUnknown {
+		bktIAM = &storagepb.Bucket_IamConfig{}
+		if b.UniformBucketLevelAccess.Enabled || b.BucketPolicyOnly.Enabled {
+			bktIAM.UniformBucketLevelAccess = &storagepb.Bucket_IamConfig_UniformBucketLevelAccess{
+				Enabled: true,
+			}
+		}
+		// TODO(noahdietz): This will be switched to a string.
+		//
+		// if b.PublicAccessPrevention != PublicAccessPreventionUnknown {
```
tritone, line 922: When is this breaking change going to land? Wondering if we should hold off until then, since this is not that great to have to leave commented out.

storage/bucket.go
```diff
@@ -854,6 +888,62 @@ func (b *BucketAttrs) toRawBucket() *raw.Bucket {
...
+		labels = make(map[string]string, len(b.Labels))
+		for k, v := range b.Labels {
+			labels[k] = v
+		}
+	}
+
+	// Ignore VersioningEnabled if it is false. This is OK because
+	// we only call this method when creating a bucket, and by default
+	// new buckets have versioning off.
+	var v *storagepb.Bucket_Versioning
```
tritone, line 904: All of this repeated code makes me wish we had some common setter interface that was satisfied by both the proto object and the apiary object

tritone (commented):
Few thoughts, thanks for putting this together!

storage/acl.go
noahdietz: No, it's because `r.Entity` is an `ACLEntity` - a type alias for `string`. The proto field is a `string` and the compiler won't accept an `ACLEntity` so we must coerce it to a `string`

storage/bucket.go
noahdietz: > but I'm wondering if we have a general guarantee/contract.

I'm not sure what that means concretely, but your understanding of the code is correct - the default empty value for a field is returned in the case where receiver (proto message) is `nil`, so no need to do `nil` checks generally, because if left unset, the default empty value of the field in structure we are building would be set to that anyways.

An unset `time.Time` field is the same as a `time.Time` being created with `0`s.

storage/bucket.go
noahdietz: I don't have a concrete date, it is still in the works. I'd prefer not to block other work on this if possible, these adapters are necessary for implementing any gRPC methods.

storage/bucket.go
noahdietz: Yeah, it's a bummer 🤷🏻  wish I didn't have to write all of this either :)

storage/acl.go
noahdietz: I've made everything conform to `toXxxFromProto` - basically the same as the apiary adpaters but with `FromProto` tacked on the end. Thanks for the comment.

storage/bucket.go
tritone: Sure, sounds good.

storage/bucket.go
tritone: Sounds good, yeah that answers my question.

storage/acl.go
tritone: Ahh of course, sorry I wasn't tracking this very closely haha.

tritone (approved):

Issue #5710, opened 2022-03-02
spanner: Enabling TrackSessionHandles causes deadlock


volicx:
**Client**

Spanner

**Environment**

My local environment.

**Go Environment**

```
$ go version
go version go1.16.13 linux/arm64
```

<details>
<summary>$ go env</summary>
<pre><code>
GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/home/lima.linux/.cache/go-build"
GOENV="/home/lima.linux/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/lima.linux/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/lima.linux/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_arm64"
GOVCS=""
GOVERSION="go1.16.13"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/lima.linux/go/src/github.com/volicx/spanner/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build351475225=/tmp/go-build -gno-record-gcc-switches"
</code></pre>
</details>

<details>
<summary>$ cat go.mod</summary>
<pre><code>
module github.com/volicx/spanner

go 1.16

require cloud.google.com/go/spanner v1.29.0
</code></pre>
</details>

**Code**

Here is the code to reproduce this issue.

```golang
package main

import (
        "context"
        "fmt"
        "log"
        "time"

        "cloud.google.com/go/spanner"
)

const (
        TrackSessionHandles = true

        ProjectID    = "{{ PLEASE REWRITE THIS }}"
[...]

PR #5745, opened 2022-03-10
fix(spanner): remove session handle lock during dumping the tracked session handles.

yfuruyama:
Fixes https://github.com/googleapis/google-cloud-go/issues/5710.

### DIFF
--- a/spanner/session.go
+++ b/spanner/session.go
@@ -838,11 +838,12 @@ func (p *sessionPool) getTrackedSessionHandleStacksLocked() string {
 	element := p.trackedSessionHandles.Front()
 	for element != nil {
 		sh := element.Value.(*sessionHandle)
-		sh.mu.Lock()
-		if sh.stack != nil {
-			stackTraces = fmt.Sprintf("%s\n\nSession %d checked out of pool at %s by goroutine:\n%s", stackTraces, i, sh.checkoutTime.Format(time.RFC3339), sh.stack)
+		// NOTE: No need to lock sessionHandle since we don't modify it.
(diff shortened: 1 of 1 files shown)


rahul2393:
I tried running this snippet with "go run -race ." https://go.dev/play/p/VV1SeNJfjq6 which is reading a shard object like sessionHandle and has list.Element just as we have trackedSessionHandle, it reports data race, and if I add lock in read the warning goes away. @olavloite Do you think its good idea to remove lock here.

olavloite:
I don't think it's a good idea to remove taking the lock there. Even though we don't change it there, Go will detect a data race if were are trying to read it while it is being changed by another goroutine. And that could happen if for example the session is returned to the session pool while we are reading it here.

yfuruyama:
Yeah I realized that `stack` is a slice and it could be modified while reading the slice. So this is not a good idea.

Then what we can do would be changing the order of two mutex locking to fix the [current lock order](https://github.com/googleapis/google-cloud-go/issues/5710#issuecomment-1063799494). Is it feasible?

rahul2393:
I would rather try RWMutex if it helps here avoid the deadlock, @yfuruyama were you able to replicate the issue, if yes can you share the script?

yfuruyama:
@rahul2393 I pasted a code snippet in an original issue: https://github.com/googleapis/google-cloud-go/issues/5710#issuecomment-1066255978 Let's discuss in that thread.

rahul2393:
Closing, will be fixed via https://github.com/googleapis/google-cloud-go/pull/5777

PR #5753, opened 2022-03-14
fix(bigquery): Query.Read fails with dry-run queries

### LINKED ISSUE #5752: bigquery: query.Read does not fail on dry-run queries, but evaluates them
**Client**

bigquery

**Environment**

macOS Big Sur 11.6.4

**Go Environment**

go version go1.17.6 darwin/amd64

**Code**

Dear all,
I notice discrepancies between the documentation and the code behavior. Actually, if one calls `query.Read()` with `query.Run = true` the [function is supposed to fail](https://github.com/googleapis/google-cloud-go/blob/main/bigquery/query.go#L121):
```
        // If true, don't actually run this job. A valid query will return a mostly
	// empty response with some processing statistics, while an invalid query will
	// return the same error it would if it wasn't a dry run.
	//
	// Query.Read will fail with dry-run queries. Call Query.Run instead, and then
	// call LastStatus on the returned job to get statistics. Calling Status on a
	// dry-run job will fail.
	DryRun bool
```

Instead `query.Read()` evaluates totally the query and ignores the `dryRun` option when it performs [`probeFastPath`](https://github.com/googleapis/google-cloud-go/blob/main/bigquery/query.go#L121). 

In addition, if `probeFastPath` fails, the query evaluation fallbacks into `query.Run()` which handles the `dryRun` option.

I will open a pull request to fix this current error.

Best,
Noël

**Expected behavior**

`bigquery.query.Read()` must fail, as detailed in documentation, when the dry-run option is activated

**Actual behavior**

`bigquery.query.Read()` evaluates totally the query, unless `probeFastPath()` fails also

NoelM:
Fixes #5752. Query.Read must fail with dry-run queries, as the documentation mentions


shollyman (approved):
Thanks for the contribution!

PR #5762, opened 2022-03-16
chore(storage): add emulated tests for transport-clients 

noahdietz:
Adds an implementation for `CreateBucket` in both gRPC and HTTP clients.

Seeds test files for both gRPC and HTTP clients targeting the emulator.

Updates `emulator_test.sh` to run on both linux and macos (pending a fix for https://github.com/googleapis/storage-testbench/issues/295) by toggling the network mapping flags accordingly, use `-d` (detached mode) instead of `&` so that we know if the `docker run` fails, trap/cleanup immediately after `docker run` incase something fails in the `curl` commands, add gRPC server start command, and update the `-run` flag to include the emulated gRPC/HTTP tests.


storage/http_client_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestCreateBucketEmulated_HTTP(t *testing.T) {
+	if os.Getenv("STORAGE_EMULATOR_HOST") == "" {
+		t.Skip("Emulator tests skipped without emulator running")
+	}
+
+	c, err := newHTTPStorageClient(context.Background())
```
tritone, line 32: So, the test code is basically identical and both constructors return the interface... can we just write once and run twice in a loop, swapping out the constructor?

tritone (commented):
Overall looks good! @BrennaEpp feel free to approve/merge while I'm OOO.

storage/grpc_client.go
```diff
@@ -101,8 +102,30 @@ func (c *grpcStorageClient) GetServiceAccount(ctx context.Context, project strin
 	return "", errMethodNotSupported
 }
 func (c *grpcStorageClient) CreateBucket(ctx context.Context, project string, attrs *BucketAttrs, opts ...storageOption) (*BucketAttrs, error) {
-	return nil, errMethodNotSupported
+	s := callSettings(c.settings, opts...)
```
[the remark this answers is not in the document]
noahdietz: Totally! Done!

storage/http_client_test.go
noahdietz: Yeah, we can do this. Should the tests be added to `client_test.go` instead? @BrennaEpp WDYT?

The only concern I have is if future tests/methods will have transport-specific things we want to test does it make sense to have two files?

noahdietz: Ok I combined the tests in `client_test.go` and integrated an `initEmulatedClients` helper with the `TestMain` in `integration_test.go`. Works well.

storage/emulator_test.sh
```diff
@@ -32,23 +32,40 @@ if [ "$minor_ver" -lt "$min_minor_ver" ]; then
...
-# Start the testbench
 # Note: --net=host makes the container bind directly to the Docker host’s network, 
 # with no network isolation. If we were to use port-mapping instead, reset connection errors 
 # would be captured differently and cause unexpected test behaviour.
 # The host networking driver works only on Linux hosts.
 # See more about using host networking: https://docs.docker.com/network/host/
-docker run --name $CONTAINER_NAME --rm --net=host $DOCKER_IMAGE &
+DOCKER_NETWORK="--net=host"
+if [ `go env GOOS` == 'darwin' ]; then
+    DOCKER_NETWORK="-p 9000:9000 -p 8888:8888"
```
BrennaEpp, line 49: Do the retry conformance tests pass on mac with this?

storage/client_test.go
```diff
@@ -16,23 +16,81 @@ package storage
...
 	}
-	if _, err := newHTTPStorageClient(context.Background()); err != nil {
-		t.Fatal(err)
+}
+
+// checkEmulatorEnvironment skips the test if the emulator environment variables
+// are not set.
+func checkEmulatorEnvironment(t *testing.T) {
+	if !isEmulatorEnvironmentSet() {
+		t.Skip("Emulator tests skipped without emulator running")
```
BrennaEpp, line 89: nit: this is not entirely true - the tests are skipped if the env variable is not set, regardless of whether the emulator is running or not.

storage/http_client.go
```diff
@@ -126,14 +126,46 @@ func newHTTPStorageClient(ctx context.Context, opts ...storageOption) (storageCl
...
+	// the location. This is a GCS quirk/bug.
+	if bkt.Location == "" && bkt.Lifecycle != nil {
+		bkt.Location = "US"
+	}
+	req := c.raw.Buckets.Insert(project, bkt)
+	setClientHeader(req.Header())
+	if attrs.PredefinedACL != "" {
+		req.PredefinedAcl(attrs.PredefinedACL)
+	}
+	if attrs.PredefinedDefaultObjectACL != "" {
```
BrennaEpp, line 154: Why were the nil checks removed?

storage/client_test.go
noahdietz: Fair enough! Done.

storage/emulator_test.sh
BrennaEpp: Not sure why, I think it's something to do with how errors differ when the networking is different - we see an EOF with the port forwarding; when it's on the same network, it's a connection error (which is what we are expecting). 

Either way, these tests weren't working using docker on mac before; imo it's fine as long as it works on linux. I would just add a comment mentioning that we expect the retry conformance tests to fail on darwin since we are specifically changing the network here.

BrennaEpp (approved):
LGTM once a comment is added in on the retry conformance tests failing on mac :)

storage/emulator_test.sh
noahdietz: Ok, good idea. I've added such a comment, thanks for the review.

PR #5770, opened 2022-03-18
chore(docs): add APIError to debugging docs

noahdietz:
Expand the existing Go debugging docs to include the "new" `apierror.APIError` and the potential for `googleapi.Error` REST transport errors. 

cc: @summer-ji-eng


codyoss:
Do you think we still need the old gRPC docs? I wonder if we try to start guiding users to just the new way. Thoughts?

noahdietz:
> Do you think we still need the old gRPC docs? I wonder if we try to start guiding users to just the new way. Thoughts?

I think we want to keep it for now. If they want to inspect the gRPC status Code, they will want to unwrap it, same with `googleapi.Error` and the HTTP status code. I'm hesitant to add a `Code` API to the `APIError` because the numbers are different between transports.

doc.go
```diff
@@ -165,9 +165,35 @@ For HTTP logging, set the GODEBUG environment variable to "http2debug=1" or "htt
...
-`grpc.Status`. Converting your errors to this type can be a useful to get
-more information about what went wrong while debugging.
+Most of the errors returned by the generated clients are wrapped in an
+`apierror.APIError` (https://pkg.go.dev/github.com/googleapis/gax-go/v2/apierror)
+and can be further unwrapped into a `grpc.Status` or `googleapi.Error` depending
+on the transport used to make the call (gRPC or REST). Converting your errors to
+these types can be a useful way to get more information about what went wrong
+while debugging.
+
+Converting to an `apierror.APIError` gives access to specific details in the
```
quartzmo, line 175: "Converting" in this context sounds to me like an action that can taken by the reader.

doc.go
```diff
@@ -165,9 +165,35 @@ For HTTP logging, set the GODEBUG environment variable to "http2debug=1" or "htt
...
+error. The transport-specific errors can still be unwrapped using the
+`apierror.APIError`.
+ if err != {
```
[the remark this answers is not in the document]
noahdietz: Yep! Copy pasta issue - it was already like this! https://pkg.go.dev/cloud.google.com/go#hdr-Inspecting_errors

doc.go
noahdietz: I'm now sure what you mean - it is something that the user would do?

quartzmo: Right, to me this sounds like something the user would do. So maybe just drop "Converting to an ":

```
`apierror.APIError` gives access to specific details in the
```

noahdietz: Oh oh gotchya. Done!

codyoss (approved):

PR #5775, opened 2022-03-21
chore(storage): add Bucket Get & Delete implementations

noahdietz:
Implements both gRPC & HTTP Bucket Get & Delete methods, including basic emulator tests.

Also adds a helper, `applyBucketCondsProto`, that applies the given `BucketConditions` to an arbitrary protobuf Message that satisfies the documented field requirements. This uses the same strategy as the `applyCondsProto` helper for the `Conditions` type.


storage/client_test.go
```diff
@@ -37,7 +37,7 @@ func TestCreateBucketEmulated(t *testing.T) {
 		}
 		got, err := client.CreateBucket(context.Background(), project, want)
 		if err != nil {
-			t.Fatal(err)
+			t.Fatalf("%s, %v", transport, err)
```
BrennaEpp, line 40: Maybe this is a good case for using subtests (`t.Run()`)? Same for the bucket tests below -  then you can separate each transport

storage/http_client.go
```diff
@@ -178,10 +180,45 @@ func (c *httpStorageClient) ListBuckets(ctx context.Context, project string, opt
...
+	req := c.raw.Buckets.Delete(bucket)
+	setClientHeader(req.Header())
+	if err := applyBucketConds("httpStorageClient.DeleteBucket", conds, req); err != nil {
+		return err
+	}
+	if s.userProject != "" {
+		req.UserProject(s.userProject)
+	}
+
+	return run(ctx, func() error { return req.Context(ctx).Do() }, s.retry, s.idempotent)
```
BrennaEpp, line 193: How do the settings know the default idempotency value for this delete operation?

storage/http_client.go
```diff
@@ -178,10 +180,45 @@ func (c *httpStorageClient) ListBuckets(ctx context.Context, project string, opt
...
+	}
+
+	var resp *raw.Bucket
+	err = run(ctx, func() error {
+		resp, err = req.Context(ctx).Do()
+		return err
+	}, s.retry, s.idempotent)
+
+	var e *googleapi.Error
+	if ok := xerrors.As(err, &e); ok && e.Code == http.StatusNotFound {
```
BrennaEpp, line 215: We should be able to use `errors.As()` here, right?

noahdietz: Yeah, this was a copy paste issue, thanks!

storage/http_client.go
noahdietz: Communicating the idempotency will the responsibility of the transport-agnostic client layer (i.e. the layer above these transport-specific implementations that will call them). it will communicated to this transport-specific layer via the `storageOption.idempotent(bool)`. This is basically the same as it is today, where the transport-agnostic client layer supplies `true` or `false` to the `run` call. Swap out the `run` call for the transport-client interface call with the option `idempotent(true)` and we are good to do.

storage/client_test.go
noahdietz: Yeah we can do that! Can I do that in a separate PR because there is an existing test in here (and in some other local branches I have) I'd like to update at once.

BrennaEpp: Sounds good 👍

BrennaEpp (approved):
LGTM

PR #5777, opened 2022-03-22
fix(spanner): early unlock of session pool lock during dumping the tracked session handles to avoid deadlock

rahul2393:
Fixes https://github.com/googleapis/google-cloud-go/issues/5710.


spanner/session.go
```diff
@@ -832,10 +832,13 @@ func (p *sessionPool) errGetSessionTimeoutWithTrackedSessionHandles(code codes.C
 // requires the caller to have locked p.mu.
 func (p *sessionPool) getTrackedSessionHandleStacksLocked() string {
 	p.mu.Lock()
-	defer p.mu.Unlock()
+	element := p.trackedSessionHandles.Front()
+	// NOTE: No need to keep the sessionPool lock since we don't modify it,
+	// this will return the snapshot of current elements when error getting sessions
```
yfuruyama, line 837: `trackedSessionHandles` could be being modified in different places like here. https://github.com/googleapis/google-cloud-go/blob/spanner/v1.30.0/spanner/session.go#L83

How do we guarantee that we will dump a consistent snapshot of `trackedSessionHandles` without locking `p.mu`?

rahul2393: I think getTrackedSessionHandleStacksLocked is only called when the client gets `errGetSessionTimeout`, so even if trackedSessionHandles later gets updated as a customer I want to know what happen when errGetSessionTimeout was thrown and not the current snapshot, so I think we should release the lock as soon as we read trackedSessionHandles.Front()

rahul2393: @codyoss can you please help review this.

yfuruyama: @rahul2393 If I understand correctly, this could print invalid state as stack traces.

Let's say there are the following session handle chains in `trackedSessionHandles` when the client gets `errGetSessionTimeout`. And same time, let's say `shA` was [recycled](https://github.com/googleapis/google-cloud-go/blob/spanner/v1.30.0/spanner/session.go#L83) by other goroutine.

```
shA -> shB -> shC
```

Then valid state would be either `shA -> shB -> shC` or `shB -> shC` depending on timinig.

But if we allow other goroutine to get `shA` removed from `trackedSessionHandles`, then the returned value from `p.trackedSessionHandles.Front()` would loose `next` pointer as Go's [list.List.Remove](https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/container/list/list.go;l=108-115;drc=refs%2Ftags%2Fgo1.18) removes the next pointer from the element. So the stack trace would be just `shA` for this case.

rahul2393: you are right, I updated the PR @yfuruyama can you check if its ok

spanner/session.go
```diff
@@ -72,7 +72,9 @@ func (sh *sessionHandle) recycle() {
 	}
 	p := sh.session.pool
 	tracked := sh.trackedSessionHandle
+	sh.mu.Unlock()
 	sh.session.recycle()
```
yfuruyama, line 76: I might be wrong, but is there any possibility that [sessionHandle.destory()](https://github.com/googleapis/google-cloud-go/blob/spanner/v1.30.0/spanner/session.go#L135) is called during `sh.session.recycle()` is being called?

rahul2393: No, as per my understanding destroy is only called when isSessionNotFoundError is returned from server.

yfuruyama: No, I was just wondering if it's ok to release a lock during `sh.session.recycle()` call. It looks good to me.

ansh0l (approved):
lgtm

spanner/session.go
olavloite: I think it would be safer and clearer if you would just store `sh.session` in a local variable `s` while the lock is still taken, and then after (original) line 80, call `s.recycle()`. So you get something like:

```go
	p := sh.session.pool
	tracked := sh.trackedSessionHandle
	s := sh.session
	sh.session = nil
	sh.trackedSessionHandle = nil
	sh.checkoutTime = time.Time{}
	sh.stack = nil
	sh.mu.Unlock()
        s.recycle()
```

olavloite (approved):
The solution itself LGTM, but take a look at my suggestion to reorganize the function a little to avoid releasing and taking the lock multiple times.

PR #5792, opened 2022-03-23
test(storage): use t.Cleanup for requesterpays test

BrennaEpp:
Buckets will be cleaned up anyway, so I added the error check back to the cleanup


noahdietz (approved):

codyoss:
Whoo 1.14 features!

PR #5796, opened 2022-03-24
fix(pubsub): ignore invalid argument errors in ack/modack

hongalex:
With the upcoming exactly-once delivery feature, acks/modacks can now return unrecoverable errors. However, if the user is not using exactly-once, these errors should be ignored as ack/modack should be fire-and-forget. When exactly-once is supported, we will also need to check if the user has set exactly-once from StreamingPull responses.

For more information, see discussion b/225881793 and b/226593754

### DIFF
--- a/pubsub/iterator.go
+++ b/pubsub/iterator.go
@@ -419,6 +419,13 @@ func (it *messageIterator) sendAck(m map[string]bool) bool {
 				if err := gax.Sleep(cctx, bo.Pause()); err != nil {
 					return nil
 				}
+			case codes.InvalidArgument:
+				// TODO(b/226593754): this error should be ignored unless exactly once is enabled
+				// since modacks are "fire and forget". Once EOS feature is out, retry this error
+				// if exactly-once is enabled (from StreamingPull response).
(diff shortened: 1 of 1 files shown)


maheshgattani (approved):

pubsub/iterator.go
```diff
@@ -486,18 +474,17 @@ func (it *messageIterator) sendModAck(m map[string]bool, deadline time.Duration)
...
 				return nil
 			default:
-				if err == nil {
-					return nil
-				}
 				// This addresses an error where `context deadline exceeded` errors
 				// not captured by the previous case causes fatal errors.
 				// See https://github.com/googleapis/google-cloud-go/issues/3060
-				if strings.Contains(err.Error(), "context deadline exceeded") {
+				if err != nil && strings.Contains(err.Error(), "context deadline exceeded") {
```
jskeet, line 480: Does the RPC error include an ErrorInfo? We should really be using that rather than relying on the error message.

hongalex: So this was a strange error that wasn't coming from the server (as far as I could tell) so I'm not entirely sure if it has `ErrorInfo`.

PR #5825, opened 2022-03-30
feat(run): start generating apiv2

quartzmo:


run/internal/version.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Code generated by gapicgen. DO NOT EDIT.
+
+package internal
+
+// Version is the current tagged release of the library.
+const Version = "0.1.0"
```
codyoss, line 20: Would you mind manually changing this to 0.0.0. We should also update the template in a future CL to do the same

codyoss (changes requested):
Just one small thing that I think will make release-please a little happier

run/CHANGES.md
```diff
@@ -0,0 +1,5 @@
+# Changes
+
+## v0.1.0
```
codyoss, line 3: manually update to remove this entry

quartzmo: Done, thanks!

codyoss (approved):

PR #5834, opened 2022-03-31
chore(internal/gapicgen): Change internal version to 0.0.0

quartzmo:
@codyoss Does this resolve the issue you raised in [#5825](https://github.com/googleapis/google-cloud-go/pull/5825#discussion_r839926591)?

> Would you mind manually changing this to 0.0.0. We should also update the template in a future CL to do the same

Should `CHANGES.md` also be modified?

```
# Changes

## v0.1.0

- feat(run): start generating clients
```

Answer from @codyoss: Yes, remove v0.1.0 from CHANGES.md.


codyoss (approved):

codyoss (changes requested):
We no longer have a need for this to be a template, be parsed etc. There are no longer variables. Let's just move this to a string that we write.

codyoss (approved):

PR #5843, opened 2022-04-04
chore(bigtable): removes cbt source code

telpirion:
CBT development as moved [here](https://github.com/googleapis/cloud-bigtable-cbt-cli).


codyoss (approved):
Thanks!

tritone (commented):
I noticed that pkg.go.dev still links to google-cloud-go. Is there something that needs to happen to point it to the new location?

telpirion:
> I noticed that pkg.go.dev still links to google-cloud-go. Is there something that needs to happen to point it to the new location?

Good catch. IIRC, @codyoss and I discussed this (but I don't recall what the action item is). Cody, would you mind refreshing my memory as to whether I need to do something special to update pkg.dev.go?

codyoss:
There is likely a little process that is needed to break the chain for deps when carving out a submodule. I have this in a doc internally but there is some external guidance as well. https://github.com/golang/go/wiki/Modules#is-it-possible-to-add-a-module-to-a-multi-module-repository You need to depend on a future version of existing module for a release to break the cycle. Let me know if you wanna chat offline.

PR #5844, opened 2022-04-04
chore(bigtable): Table.ReadRow returns empty map for row

### LINKED ISSUE #5828: bigtable: table.ReadRow returns nil instead of zero-length map
**Client**
Bigtable

**Environment**
MacOS

**Go Environment**

go version go1.17 darwin/amd64

**Code**
```go
...init bigtable client
tbl := client.Open(tbl)
row, err := tbl.ReadRow(ctx, "randomkey", bigtable.RowFilter(bigtable.LatestNFilter(2)))
```

**Expected behavior**
`table.ReadRow()` should return a zero-length map and `nil` error according to the docs.
`ReadRow is a convenience implementation of a single-row reader. A missing row will return a zero-length map and a nil error.`

**Actual behavior**
A `nil` map and `nil` error is returned

**Additional context**
This can be either solved by initialising an empty map in `ReadRow()`, or modifiying the docs to specifiy it returns a nil instead.

telpirion:
Fixes #5828


bigtable/bigtable.go
```diff
@@ -258,6 +258,12 @@ func (t *Table) ReadRow(ctx context.Context, row string, opts ...ReadOption) (Ro
 		r = rr
 		return true
 	}, opts...)
+
+	// If ReadRows() returns a nil row, return an empty map
+	if r == nil {
```
tritone, line 263: I would imagine this should only happen if the error is nil as well? If the error is non-nil, nil seems right for r.

telpirion: Acknowledged.

I disagree: I can imagine customers checking for a nil error and then attempting to iterate over a nil return value (without thinking to check whether the return value is nil, too).

tbpg: Generally Go doesn't distinguish between `nil` maps and slices versus empty ones. See https://go.dev/play/p/Yaeo1V-J8zA.

So, I'd argue we should update the docs rather than return an empty map.

tbpg: Expanded example that includes a `for` loop (which works just fine on a `nil` map): https://go.dev/play/p/xBR7WliRi0U.

See "For statements with range clause" in https://go.dev/ref/spec#For_statements.

telpirion: > Generally Go doesn't distinguish between `nil` maps and slices versus empty ones. See https://go.dev/play/p/Yaeo1V-J8zA.
> 

I didn't know that! Thank you for the tip.

> So, I'd argue we should update the docs rather than return an empty map.
>

Re-reading this thread, it makes sense to change the documentation rather than the behavior. Change made.

tritone: There is one gotcha with a nil map vs. an empty map: inserting in a nil map will cause a panic. I don't think it is the most likely thing for someone to do with the return value of this function, but obviously someone noticed enough to file the bug.

IMO, it makes most sense for the function to behave as follows:
1. `err` is non-nil: return a nil map and `err`.
2. `err` is nil and `r` is nil: return an empty map and `err`
3. `err` is nil and `r` is non-nil: return `r` and `nil`

The reason I commented in the first place is that your original change didn't do (1) as I would have expected.

However, I recognize that this is pretty low stakes so up to you what you do with it.

tritone (approved):

bigtable/bigtable.go
telpirion: Acknowledged. I would prefer not to change the behavior of the method at this point.

PR #5848, opened 2022-04-05
chore: start generating bigquery/migration v2

shollyman:
I wanted to get generation of the new v2 endpoint for bigquery/migration bootstrapped.

If I need to do more to bootstrap the initial PR, please let me know.  I also noticed the fusion link for the genbot run appears to be terminating in failure due to go 1.18 mismatches of some kind.


codyoss:
@shollyman I think you will also just want to generate the client locally here too or else it won't generate until there is a generation with this library being updated. 

I opened a seperate PR to fix the 1.18 issues you mentioned, thanks for the FYI

bigquery/go.mod
```diff
@@ -2,19 +2,22 @@ module cloud.google.com/go/bigquery
...
+	google.golang.org/grpc v1.45.0
+	google.golang.org/protobuf v1.28.0
+)
+
 require (
 	cloud.google.com/go v0.100.2
 	cloud.google.com/go/datacatalog v1.3.0
 	cloud.google.com/go/iam v0.3.0
-	cloud.google.com/go/storage v1.21.0
+	cloud.google.com/go/storage v1.10.0
```
quartzmo, line 17: Any idea why this dependency version was lowered to v1.10.0?

shollyman: That one's surprising.  I'll bump it up and tidy again.

quartzmo (approved):
LGTM

PR #5851, opened 2022-04-05
chore: fixup 1.17 -> 1.18 things

codyoss:


codyoss:
It appears we need to rebuild the 1.18 image so that staticcheck has proper support for 1.18

quartzmo (approved):
LGTM

PR #5856, opened 2022-04-05
chore(storage): add benchmarking script

BrennaEpp:


storage/internal/benchmarks/utils.go
```diff
@@ -0,0 +1,131 @@
...
+
+func randomObjectName() string {
+	// GCS accepts object name up to 1024 characters, but 128 seems long enough to
+	// avoid collisions.
+	maxLen := 128
+
+	return randomString(maxLen, lowercaseLettersAndNumbers+uppercaseLetters)
+}
+
+// random string between 1 and maxLen
```
tritone, line 66: Is it problematic memory-wise to use this for larger objects? Also, to compare with other languages, I'm wondering if measuring the whole round trip from disk would be preferable.

storage/internal/benchmarks/utils.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+}
+
+func initializeClient(ctx context.Context, api benchmarkAPI) (*storage.Client, benchmarkAPI, benchmarkAPI, error) {
+	var readAPI, writeAPI benchmarkAPI
+	var client *storage.Client
+	var err error
+
+	if api == random {
```
tritone, line 107: Hopefully it shouldn't be random!

storage/internal/benchmarks/utils.go
```diff
@@ -0,0 +1,131 @@
...
+
+	switch api {
+	case xml, json:
+		client, err = storage.NewClient(ctx, option.WithCredentialsFile(credentialsFile))
+		readAPI = json
+		writeAPI = xml
+	case grpc:
+		client, err = storage.NewHybridClient(ctx, &storage.HybridClientOptions{
+			HTTPOpts: []option.ClientOption{option.WithCredentialsFile(credentialsFile)},
+			GRPCOpts: []option.ClientOption{option.WithCredentialsFile(credentialsFile)},
```
tritone, line 123: Looks like this uses the default connection pool size. It would be interesting to see if performance improves using a larger pool size (maybe only for concurrent ops though). To set it: https://pkg.go.dev/google.golang.org/api/option#WithGRPCConnectionPool

storage/internal/benchmarks/utils.go
```diff
@@ -0,0 +1,131 @@
...
+
+	return sb.String()
+}
+
+func randomObjectName() string {
+	// GCS accepts object name up to 1024 characters, but 128 seems long enough to
+	// avoid collisions.
+	maxLen := 128
+
+	return randomString(maxLen, lowercaseLettersAndNumbers+uppercaseLetters)
```
tritone, line 63: total nit but it's more normal to use something like UUIDs for this purpose.

storage/internal/benchmarks/upload_benchmark.go
```diff
@@ -0,0 +1,83 @@
...
+	objectSize     int64
+	chunkSize      int
+}
+
+func uploadBenchmark(ctx context.Context, u uploadParams) (time.Duration, error) {
+	elapsedTime := time.Duration(0)
+	start := time.Now()
+
+	u.o.If(storage.Conditions{
+		DoesNotExist: true,
```
tritone, line 39: It is interesting that we end up including retried requests in the benchmark. I'm not sure what other languages do in that regard.

storage/internal/benchmarks/upload_benchmark.go
```diff
@@ -0,0 +1,83 @@
...
+	})
+
+	objectWriter := u.o.NewWriter(ctx)
+	objectWriter.ChunkSize = u.chunkSize
+
+	var written int64
+	for written < u.objectSize {
+		// stop timer here
+		elapsedTime += time.Since(start)
+		rewindReader(u, written)
```
tritone, line 49: Can you speak to why pausing time here is necessary? Also again, I'm wondering if writing the data to a rewindable reader is really what we want to be measuring here.

tritone (commented):
A bunch of thoughts, hopefully we can discuss more later on Fri. Thanks for putting this together!

storage/internal/benchmarks/utils.go
BrennaEpp: Default is to randomly select between GRPC and JSON/XML to upload/download. We can then compare the APIs when analyzing the results. 

This follows the C++ guidance, although I can make the default something else.

storage/internal/benchmarks/README.md
```diff
@@ -0,0 +1,43 @@
...
+## Run example:
+This runs 1000 iterations on 0 to 2Gib files in the background, sending program output (errors and such) to `output.log`:
+
+`go run main -t 72h -max_size 2097152 -max_samples 1000 -o results/2GibFiles1k.csv -c &> output.log &`
+
+
+
+## CLI parameters:
+```
+  -api string
```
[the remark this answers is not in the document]
BrennaEpp: 👍  currently adding a table for the cli parameters to explain them better. This was ripped directly from code output when trying to put a wrong parameter in.

storage/internal/benchmarks/README.md
```diff
@@ -0,0 +1,43 @@
...
+        api used to upload/download objects; JSON or XML values will use JSON to upload and XML to download (default "RANDOM")
+  -c    concurrent
```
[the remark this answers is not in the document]
BrennaEpp: added default value to read me. allowing number of workers is a good idea; would need some refactoring. I see another comment on this below so I'll close this one.

storage/internal/benchmarks/README.md
```diff
@@ -0,0 +1,43 @@
...
+  -creds string
+        path to credentials file
+  -gc_f
```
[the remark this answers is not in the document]
BrennaEpp: I was interested in seeing the difference between running "on a blank slate" (garbage collecting before each run) vs organically. There does seem to be a small difference but I don't know if it's statistically significant.

storage/internal/benchmarks/README.md
```diff
@@ -0,0 +1,43 @@
...
+        force garbage collection at the beginning of each upload
+  -max_cs int
+        max chunksize in kib (default 16384)
+  -max_samples int
+        maximum number of objects to upload (default 10000)
+  -max_size int
+        maximum object size in kib (default 16)
```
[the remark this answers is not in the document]
BrennaEpp: Yeah. Number of samples should be around at least 3000 for non-concurrent runs and between 4000-8000 for concurrent runs if we consider a difference of 5 MiB/s to be significant.
I guess it depends also on the amount of time we want the default to be, but we can control that with the timeout as well. 

As for max size, I've been using `2097152` (2GiB) but never bothered updating the code.

storage/internal/benchmarks/main.go
```diff
@@ -0,0 +1,358 @@
...
+	recordResultGroup, _ := errgroup.WithContext(context.Background())
+	startRecordingResults(file, recordResultGroup)
+
+	benchmarkRunGroup, _ := errgroup.WithContext(context.Background())
+
+	for i := 0; i < opts.maxSamples && (i < opts.minSamples || time.Since(start) < opts.timeout); i++ {
+		if opts.concurrent {
+			benchmarkRunGroup.Go(func() error {
+				if err := benchmarkRun(context.Background(), opts, bucketName); err != nil {
+					log.Printf("run failed: %v", err)
```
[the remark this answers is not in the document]
BrennaEpp: We don't want to stop all benchmarks on one error, so this logs an error and lets the other benchmark runs continue.

storage/internal/benchmarks/utils.go
BrennaEpp: Memory-wise, this is trying to emulate a buffer of "AppSizeBuffer". I'm not sure if other languages are getting from disk. My understanding of the C++ implementation is that it does something similar

storage/internal/benchmarks/upload_benchmark.go
BrennaEpp: I paused time here to remove the time to rewind the reader (which admittedly is probably not that significant) from the benchmark, as that would not be a typical workload.

I used a rewindable reader here on a random string buffer to avoid reallocating a buffer. Perhaps that would be a more typical workflow?

storage/internal/benchmarks/upload_benchmark.go
BrennaEpp: Yep. We should probably discuss with the larger team so we implement it consistently.

BrennaEpp: Added it as a topic for next team meeting.

storage/internal/benchmarks/download_benchmark.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+	}()
+
+	objectReader, err = o.NewReader(ctx)
+	if err != nil {
+		return elapsedTime, fmt.Errorf("NewReader on object %s/%s : %v", o.BucketName(), o.ObjectName(), err)
+	}
+
+	var read int64
+	if read, err = io.Copy(io.Discard, objectReader); err != nil {
```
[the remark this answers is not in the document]
BrennaEpp: That is a good point. We probably should include the data being written locally, which would use the `q_read` to determine local buffer size

storage/internal/benchmarks/README.md
```diff
@@ -0,0 +1,43 @@
...
+        min chunksize in kib (default 16384)
+  -min_samples int
+        minimum number of objects to upload (default 10)
+  -min_size int
+        minimum object size in kib
+  -o string
+        file to output results to (default "res.csv")
+  -p string
+        projectID
+  -q_read int
```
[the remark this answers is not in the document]
tritone: TODO: We think this should maybe be https://pkg.go.dev/net/http#Transport.ReadBufferSize . Let's sync with other language owners and figure out how this is being used. Note in docs that it is not being applied currently

storage/internal/benchmarks/utils.go
BrennaEpp: Added this in the updated code

storage/internal/benchmarks/main.go
```diff
@@ -0,0 +1,358 @@
...
+	defer file.Close()
+
+	recordResultGroup, _ := errgroup.WithContext(context.Background())
+	startRecordingResults(file, recordResultGroup)
+
+	benchmarkRunGroup, _ := errgroup.WithContext(context.Background())
+
+	for i := 0; i < opts.maxSamples && (i < opts.minSamples || time.Since(start) < opts.timeout); i++ {
+		if opts.concurrent {
+			benchmarkRunGroup.Go(func() error {
```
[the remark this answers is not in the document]
BrennaEpp: There is now control over number of goroutines

storage/internal/benchmarks/upload_benchmark.go
```diff
@@ -0,0 +1,168 @@
...
+	md5Hash hash.Hash
+	crcHash hash.Hash32
+	w       *storage.Writer
+	md5     bool
+	crc     bool
+}
+
+// Uploads the wrong md5 hash on large objects and
+// doesn't always send the last crc32 sum to the object
+// not sure why
```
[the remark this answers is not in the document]
BrennaEpp: I think I figured it out - I think we are only sending the checksums on the first write, so naturally setting them after has no effect. 

That means that to do these checks, we need to write two separate times for uploads, once to calculate the hashes in memory, and the second to actually upload the object. I added that in but I imagine this is going to have an impact on the performance.

storage/internal/benchmarks/main.go
```diff
@@ -0,0 +1,382 @@
...
+	"runtime"
+	"strconv"
+	"sync"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"golang.org/x/sync/errgroup"
+)
+
+const codeVersion = 2.3 // to keep track of which version of the code a benchmark ran on
```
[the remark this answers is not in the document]
BrennaEpp: I've been updating it based on changes I make - major changes update the number and minor changes I've updated the decimal portion.

This has mainly been for my own tracking on results so I can a. compare between code versions and b. don't accidentally use the results from a previous iteration on the code for my analysis

We could probably reset this once everything is ready to merge as we will be re-taking the results anyway.

storage/internal/benchmarks/utils.go
```diff
@@ -0,0 +1,176 @@
...
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		WriteBufferSize:       int(writeBufferSize),
+		ReadBufferSize:        int(readBufferSize),
+	}
+	trans, err := htransport.NewTransport(ctx, base, option.WithScopes("https://www.googleapis.com/auth/devstorage.full_control"),
+		option.WithUserAgent("custom-user-agent"), option.WithCredentialsFile(credentialsFile))
```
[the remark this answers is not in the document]
BrennaEpp: Scope is required to be able to upload/download to gcs. Usually we set this internally but since we are using a custom http client here we must set it manually. 
CredentialsFile is required to pass in custom credentials (which is what I've been doing on my compute instance). We could remove that and require credentials to be set in a different way but I like the simplicity of this.
I removed user agent option as it is not necessary.

storage/internal/benchmarks/download_benchmark.go
```diff
@@ -0,0 +1,142 @@
...
+
+	for bytesLeftToRead > 0 {
+		shouldRead := int64(d.appBufferSize)
+		if int64(d.appBufferSize) > bytesLeftToRead {
+			shouldRead = bytesLeftToRead
+		}
+
+		var read int64
+		if read, err = io.CopyN(w, objectReader, shouldRead); err != nil {
+			return elapsedTime, fmt.Errorf("io.Copy: %v", err)
```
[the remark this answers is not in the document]
BrennaEpp: After changing to files, I removed explicit buffering into memory on the application level. 
Now, we just have the transport buffers for read and write. I have options to allow those to be tweaked but the defaults of the application is the Go default.

storage/internal/benchmarks/download_benchmark.go
```diff
@@ -0,0 +1,142 @@
...
+
+		elapsedTime = time.Since(start)
+
+		readAllBytes := bytesLeftToRead == 0
+		if err == nil && !readAllBytes {
+			err = fmt.Errorf("did not read all bytes; read: %d, expected to read: %d", d.objectSize-bytesLeftToRead, d.objectSize)
+		}
+	}()
+
+	objectReader, err = d.o.NewReader(ctx)
```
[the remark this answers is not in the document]
BrennaEpp: Yes, I hadn't seen that - in this case, I will modify the code to only to md5 checks on reads, and always set the crc32c value to true for reads as we are always doing those checks.

storage/internal/benchmarks/utils.go
BrennaEpp: +1 I haven't gotten to it, but will do so shortly

storage/internal/benchmarks/upload_benchmark.go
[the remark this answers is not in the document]
frankyn: You can take an alternate approach which is what is used in other languages (we've hit this before); basically you compute the hash on upload and then after upload is complete you review the responseAttrs to validate the checksum is equivalent.

frankyn (changes requested):

storage/internal/benchmarks/upload_benchmark.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	}()
+
+	f, err := os.Open(uopts.fileName)
+	if err != nil {
+		return elapsedTime, fmt.Errorf("os.Open: %v", err)
+	}
+	defer f.Close()
+
+	if uopts.crc32c || uopts.md5 {
```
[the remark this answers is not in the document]
BrennaEpp: Done

storage/internal/benchmarks/main.go
```diff
@@ -0,0 +1,381 @@
...
+		}
+		// give some time before checking again
+		time.Sleep(time.Millisecond * 100)
+	}
+
+	// Read.
+	for i := 0; i < 3; i++ {
+		if opts.forceGC {
+			runtime.GC()
+			// debug.FreeOSMemory()
```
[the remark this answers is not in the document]
BrennaEpp: Done, lmk if that's what you had in mind

storage/internal/benchmarks/download_benchmark.go
```diff
@@ -0,0 +1,90 @@
...
+		return
+	}
+	defer func() {
+		err := objectReader.Close()
+		if rerr == nil {
+			rerr = err
+		}
+	}()
+
+	w := newHashWriter(f, dopts.md5, false) // crc32c checks are performed automatically
```
frankyn, line 69: I missed this on first pass; In this case Go doesn't have the option to disable crc32c and IIUC if we introduce md5 validation on download we will be performing double checksum validation per object when md5 is selected; I say let's always default to crc32c in this case and don't enable md5 as the integrity validation only needs one.

frankyn (commented):
Overall LGTM, just two comments;

storage/internal/benchmarks/hashwriter.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	if hw.crc32c {
+		if got := hw.crcHash.Sum32(); got != expectedCRCChecksum {
+			return fmt.Errorf("crc checksum does not match; got: %d, expected: %d", got, expectedCRCChecksum)
+		}
+	}
+	return nil
+}
+
+func (hw *hashWriter) Write(p []byte) (n int, err error) {
```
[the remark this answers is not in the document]
BrennaEpp: I didn't know about that function! I'm guessing it does the same as this implementation, and it reduces the code somewhat so I'll make use of that.

storage/internal/benchmarks/download_benchmark.go
BrennaEpp: Makes sense; modified!

storage/internal/benchmarks/main.go
```diff
@@ -0,0 +1,367 @@
...
+	os.Remove(objectName)
+	// Do not attempt to read from a failed upload
+	if err != nil {
+		return fmt.Errorf("failed upload: %v", err)
+	}
+
+	// Wait for the object to be available.
+	timedCtx, cancelTimedCtx := context.WithTimeout(ctx, time.Second*3)
+	defer cancelTimedCtx()
+	if _, err := o.Retryer(storage.WithPolicy(storage.RetryAlways)).Attrs(timedCtx); err != nil {
```
frankyn, line 245: Not sure I understand this line; but does it always retry for any failures on subsequent storage requests? (which in this case would be reads which are always retriable)

frankyn (commented):
One last question, thanks for your patience

storage/internal/benchmarks/client_pool.go
```diff
@@ -0,0 +1,196 @@
...
+			c.Close()
+		}
+		for _, c := range gRPCClients.clients {
+			c.Close()
+		}
+	}
+}
+
+func canUseClientPool(opts *benchmarkOptions) bool {
+	return opts.useDefaults || (opts.maxReadSize == opts.minReadSize && opts.maxWriteSize == opts.minWriteSize)
```
tritone, line 104: I don't really understand why this is the criterion for being able to use a client pool; could you add a comment?

storage/internal/benchmarks/client_pool.go
```diff
@@ -0,0 +1,196 @@
...
+		Proxy:                 http.ProxyFromEnvironment,
+		DialContext:           dialer.DialContext,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		WriteBufferSize:       writeBufferSize,
+		ReadBufferSize:        readBufferSize,
+	}
```
tritone, line 161: The HTTP/2 ReadIdleTimeout should also be set at this stage; see https://github.com/googleapis/google-api-go-client/blob/main/transport/http/configure_http2_go116.go

storage/internal/benchmarks/client_pool.go
```diff
@@ -0,0 +1,196 @@
...
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}
+
+	// These are the default parameters with write and read buffer sizes modified
+	base := &http.Transport{
+		Proxy:                 http.ProxyFromEnvironment,
+		DialContext:           dialer.DialContext,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
```
tritone, line 155: This could also be a command-line param (doesn't have to be set now though). Same for https://pkg.go.dev/net/http#Transport.DisableKeepAlives

storage/internal/benchmarks/client_pool.go
```diff
@@ -0,0 +1,196 @@
...
+func (p *clientPool) Put(c *storage.Client) {
+	p.clients = append(p.clients, c)
+}
+
+// we can share clients as long as the app buffer sizes are constant
+var httpClients, gRPCClients *clientPool
+
+var nonBenchmarkingClients = clientPool{
+	New: func() *storage.Client {
+		// we don't care if it's grpc or http, so we don't need mutex
```
tritone, line 61: Might be good to just make this HTTP always for debuggability's sake

storage/internal/benchmarks/client_pool.go
```diff
@@ -0,0 +1,196 @@
...
+	clientMu.Lock()
+	client, err := storage.NewClient(ctx, option.WithHTTPClient(&http.Client{Transport: trans}))
+	clientMu.Unlock()
+
+	return client, err
+}
+
+func initializeGRPCClient(ctx context.Context, writeBufferSize, readBufferSize int, connPoolSize int, useDefaults bool) (*storage.Client, error) {
+	if useDefaults {
+		clientMu.Lock()
```
tritone, line 178: This mutex solution is ugly but seems fine for now

storage/internal/benchmarks/client_pool.go
```diff
@@ -0,0 +1,196 @@
...
+	"sync"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/option"
+	htransport "google.golang.org/api/transport/http"
+	"google.golang.org/grpc"
+)
+
+type clientPool struct {
```
tritone, line 33: Can we add some overall docs for this file? In particular would be good to clarify the purpose of having client pools, the number of clients we'd expect to be spun up, and the relationship between this and `-workers`.

storage/internal/benchmarks/download_benchmark.go
```diff
@@ -18,39 +18,72 @@ import (
...
+	client     *storage.Client
+	objectSize int64
+	bucket     string
+	object     string
+}
+
+func downloadBenchmark(ctx context.Context, dopts downloadOpts) (elapsedTime time.Duration, rerr error) {
+	// Set timer
 	start := time.Now()
+	defer func() { elapsedTime = time.Since(start) }()
```
tritone, line 37: Might be obvious but this could probably use a comment about the order in which the `defer` statements execute, because it's highly relevant here :)

storage/internal/benchmarks/main.go
```diff
@@ -30,275 +30,365 @@ import (
...
 	"cloud.google.com/go/storage"
 	"golang.org/x/sync/errgroup"
-)
 
-const (
-	codeVersion = 1.0                 // to keep track of which version of the code a benchmark ran on
-	prefix      = "golang-grpc-test-" // prefix must be this for GRPC (for now)
+	// Install google-c2p resolver, which is required for direct path.
+
+	_ "google.golang.org/grpc/xds/googledirectpath"
```
tritone, line 36: Can we confirm that this does not affect the size of binaries depending on the storage root module? Just want to be doubly sure that this does not get pulled in unless you are building/running the benchmark

storage/internal/benchmarks/main.go
```diff
@@ -30,275 +30,365 @@ import (
...
+var (
+	projectID, credentialsFile, outputFile string
 
-var results chan benchmarkResult
+	opts    = &benchmarkOptions{}
+	results chan benchmarkResult
+
+	goVersion          string
+	dependencyVersions = map[string]string{
+		"cloud.google.com/go/storage": "",
```
tritone, line 51: Maybe just move this list of keys to the function where they are populated-- the indirection is a bit confusing.

storage/internal/benchmarks/w1r3.go
```diff
@@ -0,0 +1,160 @@
...
+	"log"
+	"os"
+	"path"
+	"runtime"
+	"time"
+
+	"cloud.google.com/go/storage"
+)
+
+type w1r3 struct {
```
tritone, line 30: Add a general doc describing what this benchmark is and what it does

tritone (commented):
Overall looking very good. A few more comments to address but hopefully we can get this merged in ASAP

storage/internal/benchmarks/client_pool.go
BrennaEpp: Added to taskflow (see b/260416164)

storage/internal/benchmarks/w1r3.go
BrennaEpp: Done 👍

storage/internal/benchmarks/client_pool.go
BrennaEpp: I added some - lmk if you see any gaps or think we should change how it currently works

storage/internal/benchmarks/main.go
```diff
@@ -30,275 +30,365 @@ import (
...
+		chunkSize:     randomInt64(opts.minChunkSize, opts.maxChunkSize),
+		crc32cEnabled: doCRC32C,
+		md5Enabled:    doMD5,
+		api:           api,
+	}
 
-	o := client.Bucket(bucketName).Object(objectName)
-	defer o.Delete(ctx)
+	if opts.useDefaults {
+		// get a writer on an object to check the default chunksize
```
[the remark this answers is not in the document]
tritone: Ah I see, sounds good.

storage/internal/benchmarks/client_pool.go
tritone: Discussed offline -- we probably want to have a more deterministic number of clients spun up to reduce variability between runs, rather than using a non-fixed size pool. @BrennaEpp will file an issue to make this change, but it's not a blocker for this PR.

tritone (approved):

frankyn (approved):

PR #5857, opened 2022-04-05
chore(storage): update genproto dep

codyoss:


codyoss:
@tritone After this PR lands we should cut a release right away as the proto changes contain a breaking change.

PR #5869, opened 2022-04-07
chore(storage): refactor emulator tests, resolve some TODOs

noahdietz:
Resolve TODOs around use of predefined ACL strings in proto-based code.

Refactor emulator tests to run as sub-tests, and push boilerplate code into helper function.

### DIFF
--- a/storage/client_test.go
+++ b/storage/client_test.go
@@ -29,8 +29,6 @@ import (
 var emulatorClients map[string]storageClient
 
 func TestCreateBucketEmulated(t *testing.T) {
-	checkEmulatorEnvironment(t)
-
 	transportClientTest(t, func(t *testing.T, project, bucket string, client storageClient) {
 		want := &BucketAttrs{
 			Name: bucket,
@@ -50,8 +48,6 @@ func TestCreateBucketEmulated(t *testing.T) {
 }
 
 func TestDeleteBucketEmulated(t *testing.T) {
-	checkEmulatorEnvironment(t)
-
 	transportClientTest(t, func(t *testing.T, project, bucket string, client storageClient) {
 		b := &BucketAttrs{
 			Name: bucket,
@@ -70,8 +66,6 @@ func TestDeleteBucketEmulated(t *testing.T) {
 }
 
 func TestGetBucketEmulated(t *testing.T) {
-	checkEmulatorEnvironment(t)
-
 	transportClientTest(t, func(t *testing.T, project, bucket string, client storageClient) {
 		want := &BucketAttrs{
 			Name: bucket,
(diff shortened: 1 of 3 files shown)


storage/client_test.go
```diff
@@ -171,6 +148,22 @@ func initEmulatorClients() func() error {
 	}
 }
 
+// transportClienttest executes the given function with a sub-test, a project name
+// based on the transport, a unique bucket name also based on the transport, and
+// the transport-specific client to run the test with. It also checks the environment
+// to ensure it is suitable for emulator-based tests, or skips.
+func transportClientTest(t *testing.T, test func(*testing.T, string, string, storageClient)) {
```
tritone, line 155: This is great!

storage/client_test.go
```diff
@@ -29,113 +29,90 @@ import (
...
 		want := []string{"storage.foo", "storage.bar"}
 		perms, err := client.TestIamPermissions(context.Background(), battrs.Name, want)
 		if err != nil {
-			t.Errorf("%s: on TestIamPermissions %v", transport, err)
-			continue
+			t.Fatalf("error on TestIamPermissions %v", err)
 		}
 		if diff := cmp.Diff(perms, want); diff != "" {
-			t.Errorf("%s: got(-),want(+):\n%s", transport, diff)
+			t.Fatalf("got(-),want(+):\n%s", diff)
```
tritone, line 113: Comparing results can probably just stay as Errorf right? (doesn't make a difference either way in this case though).

storage/client_test.go
```diff
@@ -29,113 +29,90 @@ import (
...
+			Name: bucket,
 		}
 		// Create the bucket that will be deleted.
 		_, err := client.CreateBucket(context.Background(), project, b)
 		if err != nil {
-			// Flag the error and continue so the Delete is skipped and the next
-			// transport can run its test.
-			t.Errorf("%s: %v", transport, err)
-			continue
+			t.Fatalf("error on CreateBucket %v", err)
```
tritone, line 58: nit, here and following: I'd just say `client.CreateBucket: %v`. Or include the bucket name if you think it would be helpful for debugging. We already know it's an error :)

tritone (commented):
Nice refactor!

storage/client_test.go
noahdietz: Yeah fair enough, I'll change it.

storage/client_test.go
noahdietz: Good point, done!

storage/client_test.go
noahdietz: Thanks! put all the things in a func!

tritone (approved):

PR #5874, opened 2022-04-08
fix(spanner/spannertest): Improve DDL application delay cancellation.

### LINKED ISSUE #5406: spanner/spannertest: delay during UpdateDatabaseDdl calls
I've been looking into why some of our tests have a suspiciously long setup time for being all in-memory by using the `spannertest.Server`, and I've found a few things, or a flow rather, that I suspect might not be intended behavior.

Looking at this code here it suggests that there's been some work into not delaying a waiting caller for too long during simulated long-running operations:

https://github.com/googleapis/google-cloud-go/blob/0ea8993ee2e0d56fd77a3a57226cca3124382a1c/spanner/spannertest/inmem.go#L304-L327

However since `UpdateDatabaseDdl` runs that code async and returns immediately here:

https://github.com/googleapis/google-cloud-go/blob/0ea8993ee2e0d56fd77a3a57226cca3124382a1c/spanner/spannertest/inmem.go#L295-L301

It will cause a waiting admin client's first `longrunning.Operation.Poll` to see if the DDL update operation is done to be false. Triggering an initial backoff of 1 sec sleep by `longrunning.Operation.WaitWithInterval`:

https://github.com/googleapis/google-cloud-go/blob/0ea8993ee2e0d56fd77a3a57226cca3124382a1c/longrunning/longrunning.go#L134-L138

Our current workaround is to just call `spannertest.Server.UpdateDDL` instead, which is fine, but since the `spannertest`s own integration tests also have these delays I guess other people than us may be using the admin client to setup DDL for their tests as well.

dsymonds:
Client code that applies a DDL change and waits for it should get a fast
response. To avoid triggering the standard polling delay, give the LRO a
chance to complete when the first status RPC (GetOperation) occurs on
any particular LRO.

Fixes #5406.

### DIFF
--- a/spanner/spannertest/inmem.go
+++ b/spanner/spannertest/inmem.go
@@ -162,10 +162,14 @@ func newLRO(initState *lropb.Operation) *lro {
 	}
 }
 
-func (l *lro) noWait() {
+// noWait causes the LRO to stop the artificial delay when applying the operation.
+// It returns whether this was the first invocation of this method.
+func (l *lro) noWait() bool {
 	if atomic.CompareAndSwapInt32(&l.waitatom, 0, 1) {
 		close(l.waitc)
+		return true
 	}
+	return false
 }
 
 func (l *lro) State() *lropb.Operation {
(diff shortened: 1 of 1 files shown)


dsymonds:
@rahul2393

rahul2393:
Thanks @dsymonds for opening the PR, LGTM

rahul2393 (approved):

Issue #5867, opened 2022-04-07
bigquery: SchemaFromJSON doesn't support policyTags


Nabil372:
Hi there!

[SchemaFromJSON](https://pkg.go.dev/cloud.google.com/go/bigquery#SchemaFromJSON) ignores policyTags in JSONs which are generated by [protoc-gen-bq-schema](https://github.com/GoogleCloudPlatform/protoc-gen-bq-schema).
This function no longer does what it's described to do:
"SchemaFromJSON takes a JSON BigQuery table schema definition (as generated by https://github.com/GoogleCloudPlatform/protoc-gen-bq-schema) and returns a fully-populated Schema."

Example JSON taken from [protoc-gen-bq-schema](https://github.com/GoogleCloudPlatform/protoc-gen-bq-schema#example-with-policy-tags):

```json
[
 {
  "name": "a",
  "type": "INTEGER",
  "mode": "REQUIRED",
  "policyTags": {
   "names": [
    "private"
   ]
  }
 },
 {
  "name": "b",
  "type": "STRING",
  "mode": "NULLABLE",
  "policyTags": {
   "names": [
    "public"
   ]
  }
 },
 {
  "name": "nested",
  "type": "RECORD",
  "mode": "REQUIRED",
  "fields": [
   {
    "name": "a",
    "type": "INTEGER",
    "mode": "REQUIRED",
    "policyTags": {
     "names": [
      "private"
     ]
    }
   },
   {
    "name": "b",
    "type": "STRING",
    "mode": "NULLABLE"
   }
  ]
 }
]
```

The function is still able to produce a `bigquery.Schema` but it ignores the `policyTags`.

shollyman:
Will take a look at this as a FR.

Issue #5833, opened 2022-03-31
bigquery: how to create a BQ schema


laurentsimon:
Hi

I've been trying to create a BQ schema to use with the `bq` from a Go structure:
```
func GenerateBQSchema(t interface{}) (string, error) {
}
```

and use it as input to the bq command to create/update tables:
```
bq mk --table   ...   myschema.schema

```

The only way I found to create this schema is to generate it myself. The schema needed seems to be the non-exported structure defined in 
https://github.com/googleapis/google-cloud-go/blob/5a2ed6b2cd1c304e0f59daa29959863bff9b5c29/bigquery/schema.go#L544

I was wondering if I missed something and it's already feasible to generate the schema, or whether this is a missing feature which would be nice to have.

Can you advise?

shollyman:
Not sure I understand the question.  Take a look at `InferSchema` if you want to generate a schema for a Go datastructure: https://pkg.go.dev/cloud.google.com/go/bigquery#InferSchema

If you're trying to mapping unexported fields, this likely needs custom code.

laurentsimon:
`InferSchema` generates some different schema when I tried, and was not accepted by the `bq` command to generate tables. Unless theres a different command to use?

shollyman:
Oh, I think I understand now.  The "schema file" used by CLI is just the underlying API's representation of the schema.  We have support  via `SchemaFromJSON` for parsing the API representation into a `Schema` type, but don't expose the converse.

If you're asking about how to create a table with the library and InferSchema, take a look at the examples:

```
func main() {
	ctx := context.Background()
	// Infer table schema from a Go type.
	schema, err := bigquery.InferSchema(Item{})
	if err != nil {
		// TODO: Handle error.
	}
	client, err := bigquery.NewClient(ctx, "project-id")
	if err != nil {
		// TODO: Handle error.
	}
	t := client.Dataset("my_dataset").Table("new-table")
	if err := t.Create(ctx,
		&bigquery.TableMetadata{
			Name:           "My New Table",
			Schema:         schema,
			ExpirationTime: time.Now().Add(24 * time.Hour),
		}); err != nil {
		// TODO: Handle error.
	}
}

type Item struct {
	Name  string
	Size  float64
	Count int
}
```

If you mean using this in other tools, could you elaborate more about the workflow you're trying to build?  It seems odd to use the library for all the schema detection and then use the CLI for the actual resource creation, but I may just not be understanding.

[...]

PR #5877, opened 2022-04-08
feat(bigquery): enhance SchemaFromJSON

shollyman:
This PR does two things: It enhances SchemaFromJSON to
work directly with the underlying TableFieldSchema messages
from the discovery API definition, and adds a FormatJSONFields
method to Schema to export the same format consumed by SchemaFromJSON.

With this, we're able clear up the existing internal duplicate logic
for this special case, and we manage to address two different feature
requests at the same time.

Fixes: https://github.com/googleapis/google-cloud-go/issues/5833
Fixes: https://github.com/googleapis/google-cloud-go/issues/5867


shollyman:
I'm also happy to entertain alternative suggestions for the `FormatJSONFields()` method.  It's not quite a schema, but we don't have a more commonly used term for "give me column definitions wrapped in a list rather than a schema message".

steffnay (approved):

bigquery/schema_test.go
```diff
@@ -1204,3 +1233,66 @@ func TestSchemaFromJSON(t *testing.T) {
...
+		description  string
+		inSchema     Schema
+		expectedJSON []byte
+	}{
+		{
+			description: "basic schema",
+			inSchema: Schema{
+				fieldSchema("foo", "strfield", "STRING", false, false, nil),
+				fieldSchema("bar", "intfield", "INTEGER", false, true, nil),
+				fieldSchema("baz", "bool_arr", "INTEGER", true, false, []string{"tag1"}),
```
crwilcox, line 1260: What happens if both required and repeated are true?

shollyman: Disallowed.  `Required` is ignored if `Repeated` is true.  See FieldSchema comments: https://pkg.go.dev/cloud.google.com/go/bigquery#FieldSchema

Under the hood this is controlled by a single "mode" value: required, optional, repeated.

PR #5884, opened 2022-04-12
chore(storage): add GetServiceAccount implementation

cojenco:
add GetServiceAccount implementation
- [x] gRPC GetServiceAccount method
- [x] HTTP GetServiceAccount method
- [x] basic emulator test


noahdietz (commented):
Looks great @cojenco, I'll let @tritone Approve this one.

storage/grpc_client.go
```diff
@@ -110,8 +110,22 @@ func (c *grpcStorageClient) Close() error {
...
+	s := callSettings(c.settings, opts...)
+	req := &storagepb.GetServiceAccountRequest{
+		Project: toProjectResource(project),
+	}
+	var resp *storagepb.ServiceAccount
+	err := run(ctx, func() error {
+		var err error
+		resp, err = c.raw.GetServiceAccount(ctx, req, s.gax...)
+		return err
+	}, s.retry, s.idempotent)
```
tritone, line 122: This is an idempotent operation, so we need to use `true` here (same in the HTTP impl). I believe s.idempotent is false by default and I can't see anywhere that `callSettings` would be able to resolve this otherwise unless we tell it explicitly that the op is idempotent.

@noahdietz would the pattern to use here to be including an idempotency option on L113, or just manually passing in true on L122?

tritone (commented):
One issue with idempotency, otherwise LGTM!

storage/grpc_client.go
noahdietz: > would the pattern to use here to be including an idempotency option on L113

Exactly, the veneer layer (which already makes decisions on idempotency today) will pass in the idempotency option when it invokes the transport-agnostic client, which gets set in the options on L113. So we should use  `callSettings.idempotent` here so that it can be configured.

noahdietz: >  including an idempotency option on L113

Correction: not _exactly_, the transport-specific clients shouldn't decide whether something is idempotent or not, the veneer should (and already is). That we don't have to code such logic twice (one for each transport).

tritone: Ahh, okay. So we should have this code go in as-is, but add the idempotency option when we shift the implementation of the public method to use the interface. I think I knew this already but thanks for reminding me 😄

tritone (approved):

PR #5892, opened 2022-04-14
chore(storage): add gRPC client constructor

tritone:
Adds newGRPCClient to replace newHybrid client, as well as an
env var to trigger calling this from the exported NewClient.
This will unblock performance benchmark testing.

Also switched over integration tests to use the new constructor and fixed a couple calls that were using the wrong client.


storage/storage.go
```diff
@@ -200,28 +207,18 @@ type hybridClientOptions struct {
 	GRPCOpts []option.ClientOption
 }
 
-// newHybridClient creates a new Storage client that initializes a gRPC-based client
-// for media upload and download operations.
+// newGRPCClient creates a new Storage client that initializes a gRPC-based
+// client. Calls that have not been implemented in gRPC will panic.
 //
 // This is an experimental API and not intended for public use.
-func newHybridClient(ctx context.Context, opts *hybridClientOptions) (*Client, error) {
```
noahdietz, line 207: Can you also remove the `type hybridClientOptions struct`? Or is that being used somewhere still?

tritone: Good point, I missed this! Will remove as well.

noahdietz (approved):
Thanks for doing this!

PR #5908, opened 2022-04-18
chore(storage): implement ListBuckets

noahdietz:
Adds `ListBuckets` implementation for HTTP and gRPC transport clients. The `ListBuckets` interface only needs to return a `BucketIterator`, no need to return an error, as no RPC is made until results are requested.

Needs https://github.com/googleapis/storage-testbench/pull/316 for emulator tests to pass.


noahdietz:
cc @cojenco

storage/grpc_client.go
```diff
@@ -156,8 +157,52 @@ func (c *grpcStorageClient) CreateBucket(ctx context.Context, project string, at
...
-	return nil, errMethodNotSupported
+func (c *grpcStorageClient) ListBuckets(ctx context.Context, project string, opts ...storageOption) *BucketIterator {
+	s := callSettings(c.settings, opts...)
+	it := &BucketIterator{
+		ctx:       ctx,
+		projectID: project,
+	}
+
+	var gitr *gapic.BucketIterator
+	fetch := func(pageSize int, pageToken string) (token string, err error) {
```
tritone, line 168: I'm wondering if, to integrate this with the production code path, we should make `fetch` a field on the iterator rather than a method? Or can we just remove the fetch method entirely? It might be good to include a TODO about how you foresee this integrating because it's not totally obvious to me.

storage/client_test.go
```diff
@@ -124,6 +126,41 @@ func TestGetSetTestIamPolicyEmulated(t *testing.T) {
...
 }
 
+func TestListBucketsEmulated(t *testing.T) {
+	transportClientTest(t, func(t *testing.T, project, bucket string, client storageClient) {
+		prefix := time.Now().Nanosecond()
+		want := []*BucketAttrs{
+			{Name: fmt.Sprintf("%d-%s-%d", prefix, bucket, time.Now().Nanosecond())},
+			{Name: fmt.Sprintf("%d-%s-%d", prefix, bucket, time.Now().Nanosecond())},
+		}
+		// Create the buckets that will be listed.
```
tritone, line 136: Is prefix taken into account by the emulator? Can we create a bucket that should not be included in the listed objects?

For a really robust test, I'd think we'd need to have more buckets than the page limit as well to test calling fetch more than once. But this can also wait until we are ready to do integration tests.

storage/grpc_client.go
noahdietz: > Or can we just remove the fetch method entirely? It might be good to include a TODO about how you foresee this integrating...

That was exactly my thinking and I will add a comment to that affect.

The existing `fetch` method references some fields of the `BucketIterator` i.e. the `Prefix`, but we can capture that reference in the function closure created here when we define `fetch`. Also, the existing  `BucketIterator.fetch` is just supplied to `iterator.NewPageInfo` and not referenced elsewhere. Making it a field is not entirely necessary.

storage/grpc_client.go
```diff
@@ -156,8 +157,52 @@ func (c *grpcStorageClient) CreateBucket(ctx context.Context, project string, at
 	return battrs, err
 }
 
-func (c *grpcStorageClient) ListBuckets(ctx context.Context, project string, opts ...storageOption) (*BucketIterator, error) {
-	return nil, errMethodNotSupported
+func (c *grpcStorageClient) ListBuckets(ctx context.Context, project string, opts ...storageOption) *BucketIterator {
+	s := callSettings(c.settings, opts...)
+	it := &BucketIterator{
```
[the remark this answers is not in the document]
noahdietz: Wdym by that? I think we need to create it here in order initialize the BucketIterator's PageInfo and nextFunc, and we also need to be able to capture it in the function closure created when defining `fetch`.

storage/client_test.go
noahdietz: > Is prefix taken into account by the emulator? 

Yes it is (I had to add it though).

> Can we create a bucket that should not be included in the listed objects?

Yes I can add that! That's better than depending on the fact that other tests create buckets thus creating extra buckets to ignore with the prefix.

storage/grpc_client.go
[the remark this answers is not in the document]
tritone: Ah, I meant that since this code is duplicated we could just create it in the top level method and then pass `it *BucketIterator` as a param to the interface method.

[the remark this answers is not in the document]
noahdietz: Ah ok I misunderstood. I think I'd like to keep this code in the transport-specific implementations just to keep the interface a little cleaner. It feels like a proper separation of concerns with this code here. Happy to change it though if you'd prefer.

storage/grpc_client.go
noahdietz: Added a comment to `fetch`. Done.

storage/client_test.go
noahdietz: Added an extra, non-prefixed bucket to the test data. Done.

storage/grpc_client.go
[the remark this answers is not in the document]
tritone: That sounds reasonable!

tritone (approved):

PR #5909, opened 2022-04-18
fix(pubsub): nack messages properly with error from receive scheduler

### LINKED ISSUE #5911: pubsub: messages in receive scheduler should be nacked if draining
**Client**

PubSub

**Expected behavior**

`ReceiveScheduler` should return an error up to `subscription.go` if the scheduler is no longer processing messages (draining). In this case, no more messages should be processed and the library should nack messages received this way for redelivery.

**Actual behavior**

Messages are being processed even if RecieveScheduler is shutdown.

hongalex:
This PR fixes a change made in #4733 which was supposed to prevent errors in the receive scheduler from bubbling up to user code (when the user had no control over this behavior). Specifically, when `Shutdown` is called on the receive scheduler and new messages come in, draining errors were being returned to the user. The limitation of #4733 is that the shutdown signal wasn't being used at all and messages _might_ be added to the scheduler even after shutdown is called. This is not a user reported issue, just something I noticed when I was reading through `receive_scheduler.go`.

Fixes #5911


codyoss (approved):
LGTM

PR #5913, opened 2022-04-19
fix(logging): revise monitored resource detection implementation

minherz:
Add deterministic method for detection monitored resource environment using file system.
Remove requirement for the specific call order of the resource detection methods.
Refactor code to improve readability.
Verify parent parameter to match the expected format.
Align global resource creation with the existing monitored resource types.


logging/go.mod
```diff
@@ -13,7 +12,11 @@ require (
 	go.opencensus.io v0.23.0
 	golang.org/x/oauth2 v0.0.0-20220411215720-9780585627b5
 	google.golang.org/api v0.74.0
+<<<<<<< HEAD
 	google.golang.org/genproto v0.0.0-20220413183235-5e96e2839df9
+=======
+	google.golang.org/genproto v0.0.0-20220407144326-9054f6ed7bac
+>>>>>>> 7ad0521ca (chore(logging): align module dependencies with other versions)
```
losalex, line 19: I believe you need to resolve those conflicts

logging/metadata/metadata.go
```diff
@@ -0,0 +1,208 @@
...
+// defined to be the empty string.
+func ProjectAttributeValue(attr string) (string, error) {
+	return defaultClient.get(projectPathPrefix + attr)
+}
+
+func pathLastValue(val string, err error) (string, error) {
+	if err == nil {
+		return val[strings.LastIndex(val, "/")+1:], nil
+	}
+	return val, err
```
losalex, line 182: Should `nil` be used instead of returned `val`?

logging/metadata/metadata.go
```diff
@@ -0,0 +1,208 @@
...
+	"net"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+)
+
+const (
+	// metadataIP is the documented metadata server IP address.
+	metadataIP = "169.254.169.254"
```
losalex, line 35: If `metadataIP` is not used anywhere else, perhaps you could assign "169.254.169.254" to `host` below and put all relevant comments there.

losalex (approved):
Left some comments. Please resolve conflicts before merge.

logging/logging.go
```diff
@@ -134,11 +134,19 @@ type Client struct {
 // By default NewClient uses WriteScope. To use a different scope, call
 // NewClient using a WithScopes option (see https://godoc.org/google.golang.org/api/option#WithScopes).
 func NewClient(ctx context.Context, parent string, opts ...option.ClientOption) (*Client, error) {
-	parent = makeParent(parent)
+	var (
+		c *vkit.Client
```
codyoss, line 138: declare these just in time with `:=`

logging/logging.go
```diff
@@ -167,11 +175,15 @@ func NewClient(ctx context.Context, parent string, opts ...option.ClientOption)
...
-func makeParent(parent string) string {
+func makeParent(parent string) (string, error) {
 	if !strings.ContainsRune(parent, '/') {
-		return "projects/" + parent
+		return "projects/" + parent, nil
 	}
-	return parent
+	prefix := strings.Split(parent, "/")[0]
+	if prefix != "projects" && prefix != "folders" && prefix != "billingAccounts" && prefix != "organizations" {
+		return parent, fmt.Errorf("Parent parameter should start with 'projects/', 'folders/', 'billingAccounts/' or 'organizations/'")
```
codyoss, line 184: Errors should start with a lowercase letter. Remove all comma delimitation and just put spaces between different resources path prefixes.

logging/logging_unexported_test.go
```diff
@@ -302,7 +302,7 @@ func TestMonitoredResource(t *testing.T) {
 		{
 			"projects/P",
 			&mrpb.MonitoredResource{
-				Type:   "project",
+				Type:   "global",
```
codyoss, line 305: Is this going to be a breaking change for customers, seems like a behaviour changes.

logging/metadata/metadata.go
```diff
@@ -0,0 +1,208 @@
...
+
+// Package metadata provides access to metadata server adoported for needs
+// of the resource detection algorithm. The package partially duplicates
+// implementation of "cloud.google.com/go/compute/metadata" package.
+//
+// This package is a wrapper around the GCE metadata service,
+// as documented at https://developers.google.com/compute/docs/metadata.
+package metadata // import "cloud.google.com/go/logging/metadata"
+
+import (
```
codyoss, line 23: A lot of the logic you added here there is already connivence methods for in the compute/metadata package. Can you just use those instead of duplicating things?

logging/metadata/metadata.go
```diff
@@ -0,0 +1,208 @@
...
+	metadataHostEnv = "GCE_METADATA_HOST"
+
+	userAgent = "gcloud-golang/0.1"
+
+	instancePathPrefix = "instance/attributes/"
+	projectPathPrefix  = "project/attributes/"
+)
+
+// NotDefinedError is returned when requested metadata is not defined.
+type NotDefinedError string
```
codyoss, line 52: I don't know if a named error here provided users much value. Sentinel , typed, errors should be a special actionable case. There is no special action a user should take if there error is this type I think, so I would just make this private.

logging/metadata/metadata.go
```diff
@@ -0,0 +1,208 @@
...
+
+func (suffix NotDefinedError) Error() string {
+	return fmt.Sprintf("metadata: attribute %q not defined", string(suffix))
+}
+
+// Error contains an error response from the server.
+type Error struct {
```
codyoss, line 59: Same comment as above.

logging/resource.go
```diff
@@ -53,49 +77,33 @@ func detectAppEngineResource() *mrpb.MonitoredResource {
 	if err != nil {
 		return nil
 	}
-	if projectID == "" {
```
codyoss, line 56: Why is this removed?

logging/resource.go
```diff
@@ -53,49 +77,33 @@ func detectAppEngineResource() *mrpb.MonitoredResource {
...
-	}
-
+	zone, _ := metadata.InstanceZone()
 	return &mrpb.MonitoredResource{
 		Type: "gae_app",
 		Labels: map[string]string{
 			"project_id":  projectID,
 			"module_id":   os.Getenv("GAE_SERVICE"),
 			"version_id":  os.Getenv("GAE_VERSION"),
-			"instance_id": os.Getenv("GAE_INSTANCE"),
```
codyoss, line 70: Why are these removed?

codyoss (changes requested):

logging/logging.go
minherz: The problem there is with `err` which has to be declared twice if I use `:=`

logging/logging_unexported_test.go
minherz: Sending `"project"` as a resource type is just invalid. This function generates a global resource type structure which has to have resource type to be set to "global". For folder, organization and billing account I left the resource type "as is" because it was originally wrong to expose this functionality to end user. And although these resource types are invalid, they can remain invalid because they have not be used in the first place.
However, if a user declare the properly formatted project's parent for a global resource, it should be reflected as "global" resource type and not as "project".

logging/metadata/metadata.go
minherz: It is correct. I put a comment explaining why I duplicated the functionality here. The core problem is that customers complain that logging works really slow when used outside of GCP. It happens because the existing code is using compute/metadata implementation with a long timeout and a retrying mechanism. All this is internal to the package. I did not want introducing this change for a special use case. I remembered that the recommended way in go is to duplicate this. The duplication also removed a need for the compute package which, theoretically, should slightly reduce overall size of the dependent packages.

logging/metadata/metadata.go
minherz: I will check if the use of this error is strictly internal and refactor the code if it is true.

logging/metadata/metadata.go
minherz: I will check if the use of this error is strictly internal and refactor the code if it is true.

logging/logging.go
```diff
@@ -167,11 +175,15 @@ func NewClient(ctx context.Context, parent string, opts ...option.ClientOption)
...
 }
 
-func makeParent(parent string) string {
+func makeParent(parent string) (string, error) {
 	if !strings.ContainsRune(parent, '/') {
-		return "projects/" + parent
+		return "projects/" + parent, nil
 	}
-	return parent
+	prefix := strings.Split(parent, "/")[0]
```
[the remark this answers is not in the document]
minherz: I think we can expect application to provide a valid parent for the destination. it should not be the job of the library to decide whether or not a misspelled value (even as small as padded with space) is neglectable or not

logging/metadata/metadata.go
minherz: I am going to revise the metadata implementation. This code probably will be removed.
In general, my reason was that returning value i provide extra info for troubleshooting in a case of `err != nil`

logging/metadata/metadata.go
minherz: This way it is more verbose

logging/metadata/metadata.go
minherz: as discussed, i will implement a quick validation for metadata server existence and for attribute retrieval i will use the compute/metadata package

minherz:
Closing in favor of #6022

PR #5918, opened 2022-04-20
chore(storage): implement ListObjects

noahdietz:
Implements ListObjects in the transport-specific clients with emulated tests.

Also adds a `veneerClient` for the emulator tests to use for populating test data, until the transport-specific clients can populate test data on their own.


noahdietz:
cc @cojenco just an FYI

storage/http_client.go
```diff
@@ -237,8 +238,65 @@ func (c *httpStorageClient) UpdateBucket(ctx context.Context, uattrs *BucketAttr
...
-	return nil, errMethodNotSupported
+func (c *httpStorageClient) ListObjects(ctx context.Context, bucket string, q *Query, opts ...storageOption) *ObjectIterator {
+	s := callSettings(c.settings, opts...)
+	it := &ObjectIterator{
+		ctx: ctx,
+	}
+	if q != nil {
+		it.query = *q
+	}
+	fetch := func(pageSize int, pageToken string) (string, error) {
```
tritone, line 249: As with ListBuckets(), would be good to clarify the plan for how this will replace the existing fetch method.

storage/grpc_client.go
```diff
@@ -216,8 +217,54 @@ func (c *grpcStorageClient) UpdateBucket(ctx context.Context, uattrs *BucketAttr
...
+		it.query = *q
+	}
+	req := &storagepb.ListObjectsRequest{
+		Parent:             bucketResourceName(globalProjectAlias, bucket),
+		Prefix:             it.query.Prefix,
+		Delimiter:          it.query.Delimiter,
+		Versions:           it.query.Versions,
+		LexicographicStart: it.query.StartOffset,
+		LexicographicEnd:   it.query.EndOffset,
+		// TODO(noahietz): Convert a projection to a FieldMask.
```
tritone, line 235: Good call-out. I'm also curious if projection and FieldMask work exactly the same? It does make a difference performance-wise for JSON to set Projection to NoACL.

storage/client_test.go
```diff
@@ -124,18 +127,126 @@ func TestGetSetTestIamPolicyEmulated(t *testing.T) {
...
+				t.Fatalf("closing object: %v", err)
+			}
+		}
+
+		// Simple list, no query.
+		it := client.ListObjects(context.Background(), bucket, nil)
+		var o *ObjectAttrs
+		for i := 0; err == nil && i <= len(want); i++ {
+			o, err = it.Next()
+			if err != nil {
```
tritone, line 169: Don't totally understand why we are continuing to the next iteration on a non-nil error. In general I would expect this test to check that the number of objects iterated lines up correctly with what's expected based on input.

tritone (commented):
Few little things. overall looks good, thanks!

storage/grpc_client.go
noahdietz: No it doesn't work the same. We have to enumerate the fields that we *want* in a FieldMask. The existing "projection" implementation for JSON is almost like a "View" (https://google.aip.dev/157#view-enumeration). In FieldMask, `*` is like "full" projection. "NoACL" is gonna be a tad messier (have to enumerate every field _except_ the ACL field(s)).

tritone: Ah yeah, I guess my main concern is whether the equivalent of NoACL will give the same performance advantage for gRPC.

noahdietz: Oh Idk! That's a great question for the broader team.

storage/client_test.go
noahdietz: Yeah I will just `break` here. 

> I would expect this test to check that the number of objects iterated lines up correctly

For sure, I'll add this.

noahdietz: Adding this actually caught a bug in the emulator list buckets impl for http prefix filtering! lol need https://github.com/googleapis/storage-testbench/pull/317 to land for CI to pass

noahdietz: I'm going to move this change to the list buckets test to a separate PR to not block this one

storage/http_client.go
noahdietz: Done!

tritone (approved):

PR #5923, opened 2022-04-20
chore(storage): implement UpdateBucket

cojenco:
This adds UpdateBucket implementation
- [x] gRPC implementation 
- adds an adapter to convert `BucketAttrsToUpdate` to a proto Bucket
- TODO - handle deleteLabels, pending discussion on support for nulling fields
- [x] HTTP implementation
- [x] basic emulator test


storage/http_client.go
```diff
@@ -231,9 +231,33 @@ func (c *httpStorageClient) GetBucket(ctx context.Context, bucket string, conds
...
+	if uattrs != nil && uattrs.PredefinedDefaultObjectACL != "" {
+		req.PredefinedDefaultObjectAcl(uattrs.PredefinedDefaultObjectACL)
+	}
+
+	var rawBucket *raw.Bucket
+	err = run(ctx, func() error {
+		rawBucket, err = req.Context(ctx).Do()
+		return err
+	}, s.retry, s.idempotent)
+	return newBucket(rawBucket)
```
noahdietz, line 258: Is the err from `run()` being handled here? I think there should be an `if err != nil` afterwards.

storage/bucket.go
```diff
@@ -945,6 +945,82 @@ func (b *BucketAttrs) toProtoBucket() *storagepb.Bucket {
 	}
 }
 
+func (ua *BucketAttrsToUpdate) toProtoBucket() *storagepb.Bucket {
+	if ua == nil {
+		return &storagepb.Bucket{}
+	}
+
+	var labels map[string]string
```
tritone, line 953: As discussed, the label keys will need to go into the updateMask as well (and the functionality isn't actually working). Let's leave this as a TODO.

storage/grpc_client.go
```diff
@@ -210,8 +211,81 @@ func (c *grpcStorageClient) GetBucket(ctx context.Context, bucket string, conds
...
+		req.CommonRequestParams = &storagepb.CommonRequestParams{
+			UserProject: toProjectResource(s.userProject),
+		}
+	}
+
+	var paths []string
+	fieldMask := &fieldmaskpb.FieldMask{
+		Paths: paths,
+	}
+	if uattrs.CORS != nil {
```
tritone, line 236: This logic may be something we can factor out at some point. But no need to fix now.

tritone (approved):

Issue #5916, opened 2022-04-20
spanner/spansql: support a default value for a column when creating or altering a table


neglect-yp:
**Is your feature request related to a problem? Please describe.**

Cloud Spanner is supporting `DEFAULT` keyword now. So, I want to use that via google-cloud-go.

release note: https://cloud.google.com/spanner/docs/release-notes#April_14_2022
document: https://cloud.google.com/spanner/docs/reference/standard-sql/data-definition-language#ddl_syntax 

**Describe the solution you'd like**
A clear and concise description of what you want to happen. What should the API be? How would you use it?

Add `Default` field with `Expr` type  to `ColumnDef`.

https://github.com/googleapis/google-cloud-go/blob/a195835f65190c73e06d738f9efe3ac9747acff6/spanner/spansql/types.go#L318-L328

And use `Default` to build query in `SQL()`.

https://github.com/googleapis/google-cloud-go/blob/a195835f65190c73e06d738f9efe3ac9747acff6/spanner/spansql/sql.go#L252

In addition, parseColumnDef should support `DEFAULT` .

https://github.com/googleapis/google-cloud-go/blob/3da94d59019e366361e53ee46a17b9131580aa88/spanner/spansql/parser.go#L1526

**Describe alternatives you've considered**

no idea.

**Additional context**

nothing.

PR #5932, opened 2022-04-22
feat(spanner/spansql): support DEFAULT keyword

neglect-yp:
close: https://github.com/googleapis/google-cloud-go/issues/5916

- support `DEFAULT` in `CREATE TABLE` and `ALTER TABLE ... ADD COLUMN`
- support `DEFAULT` in `ALTER TABLE ... ALTER COLUMN`
    - type, `SET DEFAULT` and `DROP DEFAULT`


spanner/spansql/sql.go
```diff
@@ -254,6 +265,9 @@ func (cd ColumnDef) SQL() string {
 	if cd.NotNull {
 		str += " NOT NULL"
 	}
+	if cd.Default != nil {
```
neglect-yp, line 268: Both of `Default` and `Generated` are not nil, `(cd ColumnDef) SQL()` will generate invalid SQL like `... DEFAULT (...) AS (...) STORED`. I thought about adding a enum-like const that has Default, Generated or None as a value but I didn't implement that because it will break compatibility.

neglect-yp:
I noticed my mistakes and fixed them.

rahul2393 (approved):

PR #5943, opened 2022-04-26
fix(pubsublite): retry Cancelled stream errors

tmdiep:
Add `CANCELLED` to retryable errors. Internal reference: b/228824524.

Updated integration tests (CancelPublisherContext, CancelSubscriberContext). Fixed an unrelated flaky integration test (IncreasePartitions).


pubsublite/internal/wire/streams_test.go
```diff
@@ -310,7 +345,7 @@ func TestRetryableStreamConnectTimeout(t *testing.T) {
 	if pub.Stream.currentStream() != nil {
 		t.Error("Client stream should be nil")
 	}
-	if gotErr := pub.Stream.Error(); !xerrors.Is(gotErr, ErrBackendUnavailable) {
+	if gotErr := pub.Stream.Error(); !test.ErrorEqual(gotErr, ErrBackendUnavailable) {
```
hongalex, line 348: Technically unrelated, but you can probably switch the `xerrors` in this helper function to just `errors` since `errors.Is` is supported in go1.13+ and we recently bumped the min version of our libraries.

hongalex (approved):

pubsublite/internal/wire/streams_test.go
tmdiep: Ah, excellent! There are many usages of `xerrors` in pubsublite, so I might update everything in a separate PR.

Issue #5938, opened 2022-04-25
spanner: row.ToStructLenient doesn't work for STRUCT data type


yfuruyama:
**Client**

Cloud Spanner

**Environment**

macOS

**Go Environment**

```
$ go version
go version go1.16 darwin/amd64
```

**Code**

```go
package main

import (
	"context"
	"flag"
	"fmt"
	"log"
	"os"

	"cloud.google.com/go/spanner"
)

type Result struct {
	C []*Inner `spanner:"C"`
}

type Inner struct {
	A int64 `spanner:"A"`
	//B int64 `spanner:"B"`
}

func main() {
	var projectID, instanceID, databaseID string

	flag.StringVar(&projectID, "project", "", "GCP Project ID")
	flag.StringVar(&instanceID, "instance", "", "Cloud Spanner Instance ID")
	flag.StringVar(&databaseID, "database", "", "Cloud Spanner Database ID")
	flag.Parse()

	if projectID == "" || instanceID == "" || databaseID == "" {
		flag.Usage()
		os.Exit(1)
	}

	dbPath := fmt.Sprintf("projects/%s/instances/%s/databases/%s", projectID, instanceID, databaseID)
	ctx := context.Background()
	client, err := spanner.NewClient(ctx, dbPath)
	if err != nil {
		log.Fatalf("failed to create spanner client: %v", err)
	}
	defer client.Close()

	query := "SELECT ARRAY(SELECT STRUCT(1 AS A, 2 AS B)) AS C"
	if err := client.Single().Query(ctx, spanner.NewStatement(query)).Do(func(r *spanner.Row) error {
		var result Result
		return r.ToStructLenient(&result)
	}); err != nil {
		log.Fatalf("failed to run query: %v", err)
	}
}
```

**Expected behavior**

`row.ToStructLenient` should decode the row successfully.

**Actual behavior**

Got the following decoding error.

```
[...]

PR #5944, opened 2022-04-26
feat(spanner): enable row.ToStructLenient to work with STRUCT data type

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/5938


spanner/value.go
```diff
@@ -986,7 +986,7 @@ func parseNullTime(v *proto3.Value, p *NullTime, code sppb.TypeCode, isNull bool
 
 // decodeValue decodes a protobuf Value into a pointer to a Go value, as
 // specified by sppb.Type.
-func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}) error {
+func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...bool) error {
```
olavloite, line 989: Would it maybe be better to create a `decodeValueOption` interface and pass that in instead of booleans? This means that we can only have boolean options, which both limit the types of options that we can have, but also makes them very 'magical'. Imagine that we support 3 options; no-one will understand what `decodeValue(v, t, true, false, false)` means.

rahul2393: Thanks, updated the PR please take a look again.

spanner/value.go
```diff
@@ -3043,6 +3049,22 @@ func errDecodeStructField(ty *sppb.StructType, f string, err error) error {
 	return se
 }
 
+// DecodeSetting contains all the settings for decoding from spanner struct
+type DecodeSetting struct {
```
olavloite, line 3053: nit: I think we can make this unexported

olavloite (approved):

PR #5687, opened 2022-02-23
feat(internal/gapicgen): change versionClient for gapics

codyoss:
With this commit we will add an init function that will set the
versionClient var to the module version. This will improve version
reporting for all calls.

PR #5945, opened 2022-04-26
fix(internal/gapicgen): properly update modules that have no gapic changes

codyoss:
Previously this was happening thanks to some logic the revolved around
looking for UNKNOWN date in a generated gapic client. This logic was
removed around a month ago and hense PRs that have genproto changes
but no gapic changes have needed some manual updates. With this change
the update process now parses out what kind of changes happened based
on context aware commit messages. This means even if there we not chnages
to the gapic client genproto will still get updated to the latest version
if there were changes to message types.


internal/gapicgen/generator/gapics.go
```diff
@@ -617,33 +616,6 @@ func ParseAPIShortnames(googleapisDir string, confs []*MicrogenConfig, manualEnt
...
-		dirs[dir] = true
-	}
-
-	var dirList []string
-	for dir := range dirs {
-		dirList = append(dirList, dir)
-	}
-	return dirList, nil
-}
-
```
quartzmo, line 646: I believe this was orphaned code that should have been removed in a previous PR. Maybe @codyoss can link back to that PR just for the record...

quartzmo (approved):
LGTM

internal/gapicgen/generator/gapics.go
codyoss: Correct this is follow up removal that should of happened in https://github.com/googleapis/google-cloud-go/pull/5687

PR #5946, opened 2022-04-26
feat(firestore): adds Bulkwriter support to Firestore client

telpirion:
Adds support for BulkWriter to the Golang client.

+ [Python impl.](https://github.com/googleapis/python-firestore/blob/main/google/cloud/firestore_v1/bulk_writer.py)
+ [Java impl.](https://github.com/googleapis/java-firestore/blob/main/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java)


firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+	"sync"
+	"time"
+
+	vkit "cloud.google.com/go/firestore/apiv1"
+	pb "google.golang.org/genproto/googleapis/firestore/v1"
+)
+
+const (
+	// MaxBatchSize is the max number of writes to send in a request
+	MaxBatchSize = 20
```
codyoss, line 17: At least some of this consts, at least this one, seem to be refferenced only internally and are not defaults that a user can change in a way. For such values I would recommend just keeping them private.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+const (
+	writesProvided bulkWriterStats = iota
+	writesSent
+	writesReceived
+)
+
+// The BulkWriterStatus provides details about the BulkWriter, including the
+// number of writes processed, the number of requests sent to the service, and
+// the number of writes in the queue.
+type BulkWriterStatus struct {
```
codyoss, line 58: I did not look, but is this lib already instrumented with metrics? This seems like these should just be meterics on first glance.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+// request can contain many different document writes: create, delete, update,
+// and set are all supported. Only one operation per document is allowed.
+// Each call to Create, Update, Set, and Delete can return a value and error.
+type BulkWriter struct {
+	database        string                      // the database as resource name: projects/[PROJECT]/databases/[DATABASE]
+	reqs            int64                       // total number of requests sent
+	start           time.Time                   // when this BulkWriter was started
+	vc              *vkit.Client                // internal client
+	isOpen          bool                        // signal that the BulkWriter is closed
+	startFlush      chan struct{}               // signal the beginning of flushing operation
```
codyoss, line 76: There seems to be quite a bit of sync complexity in this PR. Did you consider or look at using the bundler package to encapsalte this complexity? I think it may support many of the features you may need in this PR. We have examples in other libs how it is used internally: https://pkg.go.dev/google.golang.org/api/support/bundler

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+	received        chan bulkWriterJob          // incoming write requests
+	scheduled       chan bulkWriterRequestBatch // scheduled requests
+	sendingQueue    []bulkWriterJob             // queue of requests to send
+	backlogQueue    []bulkWriterJob             // queue of requests to store in memory during flush operation
+	maxOpsPerSecond int                         // number of requests that can be sent per second.
+	docUpdatePaths  sync.Map                    // document paths with corresponding writes in the queue.
+	waitTime        float64                     // time to wait in between requests; increase exponentially
+	writesLog       sync.Map                    // stores writes provided, sent, and received
+	cancel          context.CancelFunc          // context to send cancel message
+	mu              sync.Mutex                  // locks concurrent access
```
codyoss, line 88: nit: Anytime a mu is used in a struct seperate with whitespace the mutext and under it put the items the mutex is protecting. This makes it clear which fields are gauraded by what mutex.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+	isOpen          bool                        // signal that the BulkWriter is closed
+	startFlush      chan struct{}               // signal the beginning of flushing operation
+	isFlushing      bool                        // determines whether we're in a flush state
+	endFlush        chan struct{}               // signal the end of flushing
+	received        chan bulkWriterJob          // incoming write requests
+	scheduled       chan bulkWriterRequestBatch // scheduled requests
+	sendingQueue    []bulkWriterJob             // queue of requests to send
+	backlogQueue    []bulkWriterJob             // queue of requests to store in memory during flush operation
+	maxOpsPerSecond int                         // number of requests that can be sent per second.
+	docUpdatePaths  sync.Map                    // document paths with corresponding writes in the queue.
```
codyoss, line 84: I am not sure this is a usecase sync.Map is optimized for: https://pkg.go.dev/sync#Map

"The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex."

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+	go bw.receiver(withResourceHeader(ct, c.path()))
+	go bw.scheduler(withResourceHeader(ct, c.path()))
+
+	// TODO(telpirion): Create a retry thread? Or separate retry queue?
+
+	return &bw
+}
+
+// Close releases all of the internal resources (go routines) held by the BulkWriter.
+func (bw *BulkWriter) Close() {
```
codyoss, line 133: Have you ran tests with the race detector over this code?

codyoss: Feels like Close and End could be combined? Are there different use cases?

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+	WritesSentCount     int  // number of requests sent to the service
+	WritesReceivedCount int  // number of WriteResults received from the service
+}
+
+// A BulkWriter allows multiple document writes in parallel. The BulkWriter
+// submits document writes in maximum batches of 20 writes per request. Each
+// request can contain many different document writes: create, delete, update,
+// and set are all supported. Only one operation per document is allowed.
+// Each call to Create, Update, Set, and Delete can return a value and error.
+type BulkWriter struct {
```
codyoss, line 70: nit: maybe something like BulkDocUploader? When I see Reader and Writer in Go I look for uses with the io interfaces.

codyoss (commented):
Just to general comments. Did not looks closely at the code as some of my comments may change impl details.

firestore/bulkwriter.go
telpirion: The feature itself is called "BulkWriter" in the CGC documentation as well as in the other libraries. Unfortunately we're kinda locked into this one.

firestore/bulkwriter.go
telpirion: > Have you ran tests with the race detector over this code?

Yep! Thanks for checking. The race detector is why I have the Mutex field in the struct.

> Feels like Close and End could be combined? Are there different use cases?

This is a really good point. Ideally, yes, they would be combined. However, I ran into an issue where a request sent to the BW can be received AFTER `End()` has been called, which means that the internal `write()` function might attempt to send on a closed channel.

I'm fully open to suggestions on how to avoid this.

firestore/bulkwriter.go
telpirion: Okay, thank you for following up on that. I wasn't sure whether this was a good use case and so decided to use it anyway 😄 . Do you recommend a regular 'ol `map` instead?

firestore/bulkwriter.go
telpirion: I only looked at it cursorily (apologies--I know you sent me in that direction!). I thought it would be good to have an initial version to start the conversation with.

firestore/bulkwriter.go
telpirion: Interesting! Will take a look. 

This `Status()` is to allow the caller to wait until certain conditions are met--e.g. number of writes the caller sent is equal to the number provided, as an example. Rather than trying to interpret how the caller might want to manage their `BulkWriter`, I figured that I would give them the data to make decisions themselves.

As always, I'm open to alternative options!

firestore/bulkwriter.go
telpirion: Good point. I exposed these constants because the Java impl. did the same thing. Having them public isn't a hard requirement (and you're not the first person to point this out). I'll hide them in the next rev.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+	writesProvided bulkWriterStats = iota
+	writesSent
+	writesReceived
+)
+
+// The BulkWriterStatus provides details about the BulkWriter, including the
+// number of writes processed, the number of requests sent to the service, and
+// the number of writes in the queue.
+type BulkWriterStatus struct {
+	WritesProvidedCount int  // number of write requests provided by caller
```
enocom, line 59: GoDoc comments should go above the fields and follow the form "Field ...".

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+	}
+
+	// Start the call receiving thread and request sending thread
+	// NOTE: We are using memory to get more speed, be aware of memory usage
+	go bw.receiver(withResourceHeader(ct, c.path()))
+	go bw.scheduler(withResourceHeader(ct, c.path()))
+
+	// TODO(telpirion): Create a retry thread? Or separate retry queue?
+
+	return &bw
```
enocom, line 129: nit: it's more common to initialize a type with the `&` operator, instead of declaring it first, and then taking its address later.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+// the communication routes in between the caller and calls to the service.
+func (bw *BulkWriter) receiver(ctx context.Context) {
+	for {
+		var flushOp bool
+		var bs []bulkWriterBatch
+
+		select {
+		case bwj := <-bw.received:
+			bw.sendingQueue = append(bw.sendingQueue, bwj)
+			if bw.isFlushing {
```
enocom, line 337: This looks like a race condition as other goroutines will be changing isFlushing. Can we just keep appending to a new slice of write requests at this point?

firestore/bulkwriter.go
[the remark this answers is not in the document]
telpirion: I think that we want to make sure that clients are aware of these limitations, but they are unlikely to use them. This is probably better handled in the BulkWriter struct/class documentation rather than as exposed consts. I'll hide them.

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+func newBulkWriter(ctx context.Context, c *Client, database string) *BulkWriter {
+
+	var sm sync.Map
+	sm.LoadOrStore(writesProvided, 0)
+	sm.LoadOrStore(writesReceived, 0)
+	sm.LoadOrStore(writesSent, 0)
+
+	var dsm sync.Map
+
+	ct, cancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
telpirion: Done

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+	bw.cancel()
+
+	// Close the internal channels
+	close(bw.startFlush)
+	close(bw.endFlush)
+	close(bw.received)
+	close(bw.scheduled)
+}
+
+// End sends all enqueued writes in parallel and closes the BulkWriter to new requests.
```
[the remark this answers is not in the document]
telpirion: Removed.

firestore/bulkwriter.go
telpirion: [Outdated]

firestore/bulkwriter.go
```diff
@@ -0,0 +1,521 @@
...
+	return wr, nil
+}
+
+// receiver gets write requests from the caller and sends BatchWriteRequests to the scheduler.
+// The receiver method is the main "event loop" of the BulkWriter. It maintains
+// the communication routes in between the caller and calls to the service.
+func (bw *BulkWriter) receiver(ctx context.Context) {
+	for {
+		var flushOp bool
+		var bs []bulkWriterBatch
```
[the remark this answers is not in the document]
telpirion: [Outdated]

firestore/bulkwriter.go
```diff
@@ -0,0 +1,324 @@
...
+	"fmt"
+	"log"
+	"time"
+
+	vkit "cloud.google.com/go/firestore/apiv1"
+	pb "google.golang.org/genproto/googleapis/firestore/v1"
+)
+
+const (
+	MAX_BATCH_SIZE                          = 20
```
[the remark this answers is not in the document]
telpirion: Done

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+	// Check whether we need to increase the rate of requests to the service
+	// after each result
+	bw.rateLock.Lock()
+	bw.increaseRate()
+	bw.rateLock.Unlock()
+}
+
+// increaseRate updates the number of bundler requests that can be concurrently open
+func (bw *BulkWriter) increaseRate() {
+	elapsed := time.Now().Sub(bw.start).Seconds()
```
codyoss, line 320: Can use Minutes()

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+
+	// Ideally, we would determine the qps before increasing, but instead
+	// we simply increase the number of bundler handler invocations we have
+	// running at once.
+	mins := elapsed / 60
+	if math.Mod(mins, 5) <= 0.1 {
+		newOps := float64(bw.maxOpsPerSecond) * rateLimiterMultiplier
+		bw.maxOpsPerSecond = int(newOps)
+		bw.bundler.HandlerLimit = int(newOps)
```
codyoss, line 329: I am not sure this can be dynmaically changed. Would need testing. I suspect this is racy though because I don't think it is meant to be changed, so no mutexs

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+		maxOpsPerSecond: defaultStartingMaximumOpsPerSecond,
+		docUpdatePaths:  make(map[string]bool),
+		cancel:          cancel,
+		bundler:         &bundler.Bundler{},
+		retryBundler:    &bundler.Bundler{},
+		ctx:             ctx,
+		rateLock:        rateLock,
+	}
+
+	bw.bundler = bundler.NewBundler(bulkWriterOperation{}, bw.send)
```
codyoss, line 102: This and code below could all be moved up into stuct init

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+	bw.bundler.HandlerLimit = bw.maxOpsPerSecond
+	bw.bundler.BundleCountThreshold = maxBatchSize
+	bw.bundler.BundleByteLimit = 0 // unlimited size
+
+	bw.retryBundler = bundler.NewBundler(bulkWriterOperation{}, bw.send)
```
codyoss, line 107: Have not looked to see how or why this is used yet, but I wonder if the calls themselves should be retried and not have a seperate queue for this. May be uneeded complexity.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+// in the queue.
+func (b *BulkWriter) End() {
+	b.Flush()
+	b.cancel()
+	b.isOpen = false
+}
+
+// Flush commits all writes that have been enqueued up to this point in parallel.
+// This method blocks execution.
+func (bw *BulkWriter) Flush() {
```
codyoss, line 126: Does this method need to be exposed if End is also a blocking operation?

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+		return nil, err
+	}
+
+	w, err := doc.newCreateWrites(datum)
+	if err != nil {
+		return nil, fmt.Errorf("firestore: cannot create %v with %v", doc.ID, datum)
+	}
+
+	return &BulkWriterJob{
+		op: bw.write(w[0]),
```
codyoss, line 152: Why only the first item?

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+// Flush commits all writes that have been enqueued up to this point in parallel.
+// This method blocks execution.
+func (bw *BulkWriter) Flush() {
+	// Ensure that the backlogQueue is empty
+	bw.bundler.Flush()
+	bw.retryBundler.Flush()
+}
+
+// Status gets the current open or closed state of the BulkWriter.
+func (bw *BulkWriter) Status() BulkWriterStatus {
```
codyoss, line 133: Is the idea that this could be more than just open and closed? If not this could just return a bool

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+	j := bulkWriterOperation{
+		result: r,
+		write:  w,
+		err:    e,
+		size:   0, // ignore operation size constraints; can't be inferred at compile time
+	}
+
+	err := bw.bundler.Add(j, j.size)
+	if err != nil {
+		j.err <- err
```
codyoss, line 242: I wonder if it is better to let users subscribe to errors with a function? I know some of our other cliets do this sort of things for async workloads.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+		j.result <- nil
```
codyoss, line 243: Close channels? no need to write nil I think.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	return j
+}
+
+// processResults checks for errors returned from send() and packages up the
+// results as WriteResult objects
+func processResults(o bulkWriterOperation) (*WriteResult, error) {
```
codyoss, line 250: Seems like this should be a method on the operation.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+	wpb := <-o.result
```
codyoss, line 251: Almost any time channels are being read and other operations happening is a good usecase for select.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+		// Iterate over the response. Match successful requests with unsuccessful
+		// requests.
+		for i, res := range resp.WriteResults {
+			s := resp.Status[i]
+
+			c := s.GetCode()
+
+			if c != 0 { // Should we do an explicit check against rpc.Code enum?
+				bw.retryBundler.Add(bwj[i], bwj[i].size)
```
codyoss, line 301: This seems like it things could just keep getting sent to the retry infinitly, is that true?

codyoss (commented):
Just focused on the bulkwriter for now and not other code. A few suggestions and questions. Thanks for the refactor things are much clearer now.

firestore/bulkwriter.go
telpirion: Sorry, premature "done" ... will follow up with you on this topic.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,332 @@
...
+	database        string             // the database as resource name: projects/[PROJECT]/databases/[DATABASE]
+	start           time.Time          // when this BulkWriter was started; used to calculate qps and rate increases
+	vc              *vkit.Client       // internal client
+	isOpen          bool               // flag that the BulkWriter is closed
+	maxOpsPerSecond int                // number of requests that can be sent per second
+	docUpdatePaths  map[string]bool    // document paths with corresponding writes in the queue
+	cancel          context.CancelFunc // context to send cancel message
+	bundler         *bundler.Bundler   // handle bundling up writes to Firestore
+	retryBundler    *bundler.Bundler   // handle bundling up retries of writes to Firestore
+	ctx             context.Context    // context for canceling BulkWriter operations
```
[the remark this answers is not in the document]
telpirion: That's an interesting idea. That said, I think that the caller is more likely to want to cancel everything all at once than want to cancel each operation one-at-a-time.

firestore/bulkwriter.go
telpirion: Per our video chat, I'll check with the Firestore team about the reasoning here.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+	w, err := doc.newCreateWrites(datum)
+	if err != nil {
+		return nil, fmt.Errorf("firestore: cannot create %v with %v", doc.ID, datum)
+	}
+
+	if len(w) > 1 {
+		return nil, fmt.Errorf("firestore: too many document writes sent to bulkwriter")
+	}
+
+	j := bw.write(w[0])
```
enocom, line 154: I think `write` should just return a pointer. It looks strange to me to return a value object and then take its address.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+func (o *BulkWriterJob) processResults() (*WriteResult, error) {
+	//  TODO(telpirion): Refactor this to use select/case
+	wpb := <-o.result
+	err := <-o.err
+
+	if err != nil {
+		return nil, err
+	}
+
+	wr, err := writeResultFromProto(wpb)
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+	return wr, err
+}
+
+// A BulkWriter allows multiple document writes in parallel. The BulkWriter
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+	database        string             // the database as resource name: projects/[PROJECT]/databases/[DATABASE]
+	start           time.Time          // when this BulkWriter was started; used to calculate qps and rate increases
+	vc              *vkit.Client       // internal client
+	isOpen          bool               // flag that the BulkWriter is closed
+	maxOpsPerSecond int                // number of requests that can be sent per second
+	docUpdatePaths  map[string]bool    // document paths with corresponding writes in the queue
+	cancel          context.CancelFunc // context to send cancel message
+	bundler         *bundler.Bundler   // handle bundling up writes to Firestore
+	ctx             context.Context    // context for canceling BulkWriter operations
+	openLock        sync.Mutex         // guards against setting isOpen concurrently
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+
+	j := bw.write(w[0])
+	return &j, nil
+}
+
+// checkConditions determines whether this write attempt is valid. It returns
+// an error if either the BulkWriter has already been closed or if it
+// receives a nil document reference.
+func (bw *BulkWriter) checkWriteConditions(doc *DocumentRef) error {
+	if !bw.isOpen {
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	bw.docUpdatePaths[doc.shortPath] = true
+
+	return nil
+}
+
+// write packages up write requests into bulkWriterJob objects.
+func (bw *BulkWriter) write(w *pb.Write) BulkWriterJob {
+	r := make(chan *pb.WriteResult, 1)
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+	e := make(chan error, 1)
+
+	j := BulkWriterJob{
+		result: r,
+		write:  w,
+		err:    e,
+		size:   0, // ignore operation size constraints; can't be inferred at compile time
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+
+	bwr := pb.BatchWriteRequest{
+		Database: bw.database,
+		Writes:   ws,
+		Labels:   map[string]string{},
+	}
+
+	select {
+	case <-bw.ctx.Done():
+		break
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+	if len(bwj) == 0 {
+		return
+	}
+
+	var ws []*pb.Write
+	for _, w := range bwj {
+		ws = append(ws, w.write)
+	}
+
+	bwr := pb.BatchWriteRequest{
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+			// Do we need to be selective about what kind of errors we send?
+			for _, j := range bwj {
+				j.result <- nil
+				j.err <- err
+			}
+			return
+		}
+		// Iterate over the response. Match successful requests with unsuccessful
+		// requests.
+		for i, res := range resp.WriteResults {
```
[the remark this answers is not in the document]
telpirion: No, I asked this very question of the folks who implemented this feature in Python and Java. The reference docs also clarify this point: https://cloud.google.com/firestore/docs/reference/rpc/google.firestore.v1#batchwriteresponse 

(Although note that the comments in the previous message muddies the waters a bit.)

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+				j.attempts++
+
+				// Do we need separate retry bundler
+				if j.attempts < maxRetryAttempts {
+					bw.bundler.Add(j, j.size)
+				}
+				continue
+			}
+
+			bwj[i].result <- res
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
telpirion: Now ... done!

firestore/bulkwriter.go
```diff
@@ -0,0 +1,317 @@
...
+	if j.wr == nil && j.e == nil {
+		j.wr, j.e = j.processResults() // cache the results for additional calls
+	}
+	return j.wr, j.e
+}
+
+// processResults checks for errors returned from send() and packages up the
+// results as WriteResult objects
+func (o *BulkWriterJob) processResults() (*WriteResult, error) {
+	//  TODO(telpirion): Refactor this to use select/case
```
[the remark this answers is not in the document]
telpirion: Done!

telpirion:
Responding to [this comment](https://github.com/googleapis/google-cloud-go/pull/5946#discussion_r907606768):

Discussed this over chat. Unfortunately, because the sending operation (that callers would want to cancel) happens within a handler function (`BulkWriter.send()`) passed to the Bundler, we have to store a context. 

Also, after playing around with per-write contexts, I realized that this wouldn't be ideal either. This could lead to unpredictable write behavior, where some writes fail or succeed without the caller receiving notice about success/failure. Imagine having 20 writes in a bundle, each with its own context. The caller invokes `Context.Done()` on one of them. 
-  Should the entire bundle of writes be cancelled? If all of them should be cancelled, what do we return to the writes that didn't have `Done()` invoked on them? 
- Or should we just cancel that one write? Because of how context is passed to the Bundler, we can't cancel just one write in a bundle of 20. Rather, the request would still be sent and the putatively canceled write would actually be applied.

firestore/client.go
```diff
@@ -290,10 +290,20 @@ func (c *Client) Collections(ctx context.Context) *CollectionIterator {
...
 // Batch returns a WriteBatch.
+//
+// Deprecated: The WriteBatch API has been replaced with the transaction and
+// the bulk writer API. For atomic transaction operations, use `Transaction`.
+// For bulk read and write operations, use `BulkWriter`.
 func (c *Client) Batch() *WriteBatch {
 	return &WriteBatch{c: c}
 }
 
+// BulkWriter returns a BulkWriter instance.
```
codyoss, line 301: Any time a context is going to be stored and held and not acted upon immediately it is best practice to outline this in the docs and how it will be used.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,319 @@
...
+	e        error                // (cached) any errors that occurred
+	ctx      context.Context      // context for canceling/timing out results
+}
+
+// Results gets the results of the BulkWriter write attempt.
+// This method blocks if the results for this BulkWriterJob haven't been
+// received.
+func (j *BulkWriterJob) Results() (*WriteResult, error) {
+	if j.wr == nil && j.e == nil {
+		j.wr, j.e = j.processResults() // cache the results for additional calls
```
codyoss, line 42: In general our clients are thread safe, but calling this function is not because of the caching feature. Is there a need to cache the values? What is the purpopse of being able to call this method more than once?

firestore/bulkwriter.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+	return j.wr, j.e
+}
+
+// processResults checks for errors returned from send() and packages up the
+// results as WriteResult objects
+func (j *BulkWriterJob) processResults() (*WriteResult, error) {
+	select {
+	case <-j.ctx.Done():
+		return nil, fmt.Errorf("bulkwriter: early write cancellation")
```
codyoss, line 52: I would suggest passing back the raw error here. I dont think a custom error adds value in this case and many people code special logic around context.DeadlineExceeded.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,319 @@
...
+
+	select {
+	case <-bw.ctx.Done():
+		return
+	default:
+		resp, err := bw.vc.BatchWrite(bw.ctx, bwr)
+		if err != nil {
+			// Do we need to be selective about what kind of errors we send?
+			for _, j := range bwj {
+				j.setError(err)
```
codyoss, line 294: What if all jobs error out. The buffer is set to 1 and will deadlock, right?

firestore/bulkwriter.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+}
+
+// setError ensures that an error is returned on the error channel of BulkWriterJob.
+func (j *BulkWriterJob) setError(e error) {
+	j.err <- e
+}
+
+// setSuccess ensures that a WriteResult is returned to the result channel of BulkWriterJob.
+func (j *BulkWriterJob) setResult(r *pb.WriteResult) {
```
codyoss, line 66: This is only used in one spot, I would suggest just inlining.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,319 @@
...
+		docUpdatePaths:  make(map[string]bool),
+		cancel:          cancel,
+		ctx:             ctx,
+	}
+
+	// can't initialize with struct above; need instance reference to BulkWriter.send()
+	bw.bundler = bundler.NewBundler(BulkWriterJob{}, bw.send)
+	bw.bundler.HandlerLimit = bw.maxOpsPerSecond
+	bw.bundler.BundleCountThreshold = maxBatchSize
+	bw.bundler.BundleByteLimit = 0 // unlimited size
```
codyoss, line 115: Can be removed since 0 is the default value for this type.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,319 @@
...
+}
+
+// Flush commits all writes that have been enqueued up to this point in parallel.
+// This method blocks execution.
+func (bw *BulkWriter) Flush() {
+	bw.bundler.Flush()
+}
+
+// IsOpen gets the current open or closed state of the BulkWriter.
+func (bw *BulkWriter) IsOpen() bool {
```
codyoss, line 138: Is this method or the idea of being open needed? Only gets set to false when calling End which is a blocking call.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,319 @@
...
+// Note: You cannot write to (Create, Update, Set, or Delete) the same document more than once.
+func (bw *BulkWriter) Delete(doc *DocumentRef, preconds ...Precondition) (*BulkWriterJob, error) {
+	err := bw.checkWriteConditions(doc)
+	if err != nil {
+		return nil, err
+	}
+
+	w, err := doc.newDeleteWrites(preconds)
+	if err != nil {
+		return nil, fmt.Errorf("firestore: cannot delete doc %v", doc.ID)
```
codyoss, line 176: Should the underlying error be wrapped for context? Applies for other methods too.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,319 @@
...
+
+	err := bw.bundler.Add(j, 0) // ignore operation size constraints; can't be inferred at compile time
+	if err != nil {
+		j.setError(err)
+	}
+	return &j
+}
+
+// send transmits writes to the service and matches response results to job channels.
+func (bw *BulkWriter) send(i interface{}) {
```
codyoss, line 268: Maybe document why this type takes an empty interface when you know the type(bundler handler)

firestore/bulkwriter.go
enocom: It would be nice to avoid the caching, but I suspect `Results` could get called a few times.

Putting `j.wr` and `j.e` behind a mutex would be reasonable.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,319 @@
...
+
+	bw.docUpdatePaths[doc.shortPath] = true
+
+	return nil
+}
+
+// write packages up write requests into bulkWriterJob objects.
+func (bw *BulkWriter) write(w *pb.Write) *BulkWriterJob {
+
+	j := BulkWriterJob{
```
enocom, line 253: It's a little odd to declare a value object only to take its address later. Normally, I'd expect `j := &BulkWriterJob{}`, unless you care about value vs point semantics of course.

firestore/bulkwriter.go
enocom: As written, I agree. If the intention is to also close the error channel (or the result channel above), then it might be work keeping these around. In any case, don't we need to close the other channel when we set a result or an error?

firestore/bulkwriter_test.go
```diff
@@ -0,0 +1,233 @@
...
+				return b.Update(c.Doc("C/f"), []Update{{FieldPath: []string{"*"}, Value: 3}})
+			},
+		},
+	}
+
+	for i, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			j, err := tc.test(bw)
+			if err != nil {
+				t.Errorf("bulkwriter: cannot call %s for document\n", tc.name)
```
enocom, line 175: The trailing newline is unnecessary here. Same below.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,319 @@
...
+	bw.bundler.BundleByteLimit = 0 // unlimited size
+
+	return bw
+}
+
+// End sends all enqueued writes in parallel and closes the BulkWriter to new requests.
+// After calling End(), calling any additional method automatically returns
+// with an error. This method completes when there are no more pending writes
+// in the queue.
+func (b *BulkWriter) End() {
```
enocom, line 124: Two things:

1. Why isn't this called `Close`?
2. I think the intention is that `End` should shutdown the background goroutines. So shouldn't we call `b.cancel()` here too?

firestore/bulkwriter.go
enocom: I have the same thought. What if `End` set `isOpen` to false and then any subsequent write just returned an error if a client tried to submit a request after closing the writer? You'd be able to remote this check here.

firestore/go.sum
```diff
@@ -324,6 +324,7 @@ golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208/go.mod h1:RxMgew5VJxzue5/jJ
 golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
telpirion: Yeah, I'm thinking about the possibility of `Results()` being called more than once. I'll put these fields behind a mutex

firestore/bulkwriter.go
telpirion: I don't quite understand; it looks like `Done()` returns a struct. What error should I return?

firestore/bulkwriter.go
telpirion: I think I added these based upon a previous round of feedback 😄 . You're right, I need to close the other channels when I return.

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter.go
telpirion: In an earlier revision, we weren't actually _closing_ anything when we called `End()`, and attempting to defer it could cause a deadlock. For those reasons, I decided against calling it `Close()`.

I can be persuaded to change it to `Close()` 😄

firestore/bulkwriter.go
telpirion: Yeah, we need the idea of being open so that callers don't continually attempt to send to a BW.

I'll remove the `IsOpen()` method tho.

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter.go
telpirion: I'm not quite sure what you mean by "buffer": the buffer of the channel being written to in `setError()`?

We will only follow this code path if the request itself is not successful. Each job would remain unresolved (because we won't have a response back from the server).

telpirion: Put another way: `BatchWrite()` requests have a one-to-many relationship with BulkWriterJobs.

firestore/bulkwriter_test.go
telpirion: It makes the test output easier to read (for me). That said, I don't have strong feelings about it and can be persuaded otherwise.

firestore/client.go
telpirion: Done.

firestore/bulkwriter.go
enocom: `ctx.Err()` would probably work

firestore/bulkwriter_test.go
enocom: Your call.

firestore/bulkwriter.go
enocom: Close seems idiomatic to me, especially if it returns an error.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,320 @@
...
+	// maxRetryAttempts is the max number of times to retry a write
+	maxRetryAttempts = 10
+	// defaultStartingMaximumOpsPerSecond is the starting max number of requests to the service per second
+	defaultStartingMaximumOpsPerSecond = 500
+)
+
+// BulkWriterJob provides read-only access to the results of a BulkWriter write attempt.
+type BulkWriterJob struct {
+	err        chan error           // send error responses to this channel
+	result     chan *pb.WriteResult // send the write results to this channel
```
enocom, line 29: nit: resultCh here.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,320 @@
...
+type BulkWriter struct {
+	database        string             // the database as resource name: projects/[PROJECT]/databases/[DATABASE]
+	start           time.Time          // when this BulkWriter was started; used to calculate qps and rate increases
+	vc              *vkit.Client       // internal client
+	maxOpsPerSecond int                // number of requests that can be sent per second
+	docUpdatePaths  map[string]bool    // document paths with corresponding writes in the queue
+	cancel          context.CancelFunc // context to send cancel message
+	bundler         *bundler.Bundler   // handle bundling up writes to Firestore
+	ctx             context.Context    // context for canceling all BulkWriter operations
+	openLock        sync.Mutex         // guards against setting isOpen concurrently
```
enocom, line 93: nit: since there's only one mutex here, `mu` would be fine in my opinion.

firestore/bulkwriter.go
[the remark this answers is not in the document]
enocom: This should be:

``` golang
bw.openLock.Lock()
open := bw.isOpen
bw.openLock.Unlock()

if !open { /* ... */ }
```

As written, the lock isn't unlocked if we hit the early return.

firestore/bulkwriter.go
enocom: Shouldn't `b.cancel()` be called here?

firestore/bulkwriter_test.go
```diff
@@ -0,0 +1,233 @@
...
+			name: "cannot write to same document twice",
+			test: func(b *BulkWriter) (*BulkWriterJob, error) {
+				b.Create(c.Doc("C/a"), testData)
+				return b.Delete(c.Doc("C/a"))
+			},
+		},
+		{
+			name: "cannot ask a closed BulkWriter to write",
+			test: func(b *BulkWriter) (*BulkWriterJob, error) {
+				b.cancel()
```
enocom, line 218: Shouldn't this be called within `b.End()`?

firestore/bulkwriter_test.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	for i, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			j, err := tc.test(bw)
+			if err != nil {
+				t.Errorf("bulkwriter: cannot call %s for document\n", tc.name)
+			}
+			if j == nil {
+				t.Errorf("bulkwriter: got nil WriteResult for call to %s\n", tc.name)
```
enocom, line 178: Do you intend for the test to continue if we hit this?

firestore/bulkwriter_test.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func TestBulkWriter(t *testing.T) {
+	c, srv, cleanup := newMock(t)
+	defer cleanup()
+
+	docPrefix := c.Collection("C").Path + "/"
+
+	// Create
+	srv.addRPC(
```
enocom, line 24: nit: pulling all this into a setup function that describes what it's doing might help make this test more readable.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,320 @@
...
+			// Get the status code for this WriteResult
+			s := resp.Status[i]
+			c := s.GetCode()
+			if c != 0 { // Should we do an explicit check against rpc.Code enum?
+				j := bwj[i]
+				j.attempts++
+
+				// Do we need separate retry bundler
+				if j.attempts < maxRetryAttempts {
+					bw.bundler.Add(j, 0) // ignore operation size constraints; can't be inferred at compile time
```
enocom, line 311: There's an error here that is checked above, but not here. Is that intentional?

enocom (changes requested):

firestore/bulkwriter.go
```diff
@@ -0,0 +1,320 @@
...
+	retryMaxBatchSize = 10
+	// maxRetryAttempts is the max number of times to retry a write
+	maxRetryAttempts = 10
+	// defaultStartingMaximumOpsPerSecond is the starting max number of requests to the service per second
+	defaultStartingMaximumOpsPerSecond = 500
+)
+
+// BulkWriterJob provides read-only access to the results of a BulkWriter write attempt.
+type BulkWriterJob struct {
+	err        chan error           // send error responses to this channel
```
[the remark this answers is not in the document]
telpirion: Done.

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter_test.go
telpirion: I don't think so. `Bundler.Flush()`, which is called from within `End()`, isn't blocking. If we attempt to call `cancel()` from within `End()`, then we have a race condition that results in the last few write attempts being canceled.

firestore/bulkwriter_test.go
telpirion: No, it probably shouldn't. Good catch.

firestore/bulkwriter.go
telpirion: Yes. It's important to keep in mind that we have two related, but separate operations happening here:

+ Writes to the database, represented in memory by a `BulkWriterJob`
+ Requests to the service

The err check above fails when requests (which contain multiple writes) fail. The client returns an error object that we then pass back as the results to each individual write.

In this block, the request has succeeded (we got a response from the service), but an individual write has failed. In this case, we don't get an error object back from the client. Instead, we have to inspect the status code for each individual write and then either report a success, add a retry, or report a failure.

This code is missing the report a failure part, so I've added that.

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter_test.go
telpirion: Okay, ack.

firestore/bulkwriter.go
telpirion: `End()` doesn't return an error and shouldn't be deferred, which was why I didn't call it `Close()`. Does that change things any?

Unfortunately, calling `bw.cancel()` here creates a race condition with the call to `bw.bundler.Flush()`. `Bundler.Flush()` isn't blocking, so we could potentially call `cancel()` before the Bundler has sent the last bundle.

firestore/bulkwriter_test.go
telpirion: Done.

firestore/bulkwriter_test.go
telpirion: Actually, wait a minute. Correcting myself ...

I'm not sure whether `Bundler.Flush()` blocks or not. Local tests seem to indicate that it doesn't: when I put `cancel()` within `End()` and then call `End()`, I see that the integration tests fail with "context canceled" errors for each write.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,312 @@
...
+			}
+			return
+		}
+		// Match write results with BulkWriterJob objects
+		for i, res := range resp.WriteResults {
+			s := resp.Status[i]
+			c := s.GetCode()
+			if c != 0 { // Should we do an explicit check against rpc.Code enum?
+				j := &bwj[i]
+				(*j).attempts++
```
enocom, line 298: This deref is unnecessary and same with below.

firestore/bulkwriter.go
enocom: Are we talking about `bw.bundler.Add`? It returns an error that looking at the code is mostly unrelated. Perhaps this should just be:

``` golang
_ = bw.bundler.Add(j, 0) // ignoring the error because REASON
```

firestore/bulkwriter.go
```diff
@@ -0,0 +1,312 @@
...
+				j.setError(err)
+			}
+			return
+		}
+		// Match write results with BulkWriterJob objects
+		for i, res := range resp.WriteResults {
+			s := resp.Status[i]
+			c := s.GetCode()
+			if c != 0 { // Should we do an explicit check against rpc.Code enum?
+				j := &bwj[i]
```
enocom, line 297: Could this just be `j := bwj[i]`?

firestore/bulkwriter.go
```diff
@@ -0,0 +1,312 @@
...
+				// Do we need separate retry bundler?
+				if (*j).attempts < maxRetryAttempts {
+					bw.bundler.Add(j, 0) // ignore operation size constraints; can't be inferred at compile time
+				} else {
+					(*j).errChan <- fmt.Errorf("firestore: write failed with status: %v", s)
+				}
+				continue
+			}
+
+			bwj[i].resultChan <- res
```
enocom, line 309: Do we need to close the error chan when this happens?

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,325 @@
...
+	database        string           // the database as resource name: projects/[PROJECT]/databases/[DATABASE]
+	start           time.Time        // when this BulkWriter was started; used to calculate qps and rate increases
+	vc              *vkit.Client     // internal client
+	maxOpsPerSecond int              // number of requests that can be sent per second
+	docUpdatePaths  map[string]bool  // document paths with corresponding writes in the queue
+	limiter         rate.Limiter     // limit requests to server to <= 500 qps
+	bundler         *bundler.Bundler // handle bundling up writes to Firestore
+	ctx             context.Context  // context for canceling all BulkWriter operations
+	openLock        sync.Mutex       // guards against setting isOpen concurrently
+	isOpen          bool             // flag that the BulkWriter is closed
```
enocom, line 100: nit: `open` to match `openLock` above.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,325 @@
...
+
+	return bw
+}
+
+// End sends all enqueued writes in parallel and closes the BulkWriter to new requests.
+// After calling End(), calling any additional method automatically returns
+// with an error. This method completes when there are no more pending writes
+// in the queue.
+func (bw *BulkWriter) End() {
+	bw.Flush()
```
enocom, line 133: Shouldn't this should go after setting `isOpen` to false (e.g., prevent additional write requests, and then flush)?

firestore/bulkwriter.go
```diff
@@ -0,0 +1,325 @@
...
+					// ignore operation size constraints and related errors; job size can't be inferred at compile time
+					// Bundler is set to accept an unlimited amount of bytes
+					_ = bw.bundler.Add(j, 0)
+				} else {
+					j.setError(fmt.Errorf("firestore: write failed with status: %v", s))
+				}
+				continue
+			}
+
+			bwj[i].resultChan <- bulkWriterResult{err: nil, result: res}
```
enocom, line 321: nit: Either approach seems fine, but this should probably match `setError` above, or vice versa.

enocom (commented):
One last thing about closing the bulk writer.

firestore/bulkwriter.go
telpirion: Ack. I prefer using 'verb+Noun' naming conventions for bool fields.

firestore/bulkwriter.go
telpirion: IIRC, @codyoss suggested getting rid of `setResult()` since it's only used once.

firestore/bulkwriter.go
telpirion: Done.

enocom (approved):

firestore/bulkwriter.go
```diff
@@ -0,0 +1,325 @@
...
+	bw.bundler.BundleCountThreshold = maxBatchSize
+
+	return bw
+}
+
+// End sends all enqueued writes in parallel and closes the BulkWriter to new requests.
+// After calling End(), calling any additional method automatically returns
+// with an error. This method completes when there are no more pending writes
+// in the queue.
+func (bw *BulkWriter) End() {
```
enocom, line 132: I think there might be a minor race condition here. Here's the idea:

* Goroutine 1 in `checkWriteCondition` acquires the lock, sees that `isOpen` is true, and then releases the lock
* Goroutine 2 in `End` acquires the lock, sets `isOpen` to false, releases the lock, and then flushes
* Goroutine 1 meanwhile starts a write

The question is: does the write success or not in this case?

One possible option would be to use a read write mutex, where the read mutex wraps the entire write attempt and the write mutex wraps `isOpen` here in `End`.

telpirion: Very good point.

Here's what I've done to address this:

- Changed `BulkWriter.isOpenLock` to a [`RWMutex`](https://pkg.go.dev/sync#RWMutex).
- Added a `Lock()` and deferred `Unlock()` at the top of each DB mutator/write function (`Create()`, `Delete()`, etc).

Each mutator will hold the lock while queueing up the write for the Bundler. If `End()` is called during this time, it will will block while the mutator holds the lock.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,331 @@
...
+// Flush commits all writes that have been enqueued up to this point in parallel.
+// This method blocks execution.
+func (bw *BulkWriter) Flush() {
+	bw.bundler.Flush()
+}
+
+// Create adds a document creation write to the queue of writes to send.
+// Note: You cannot write to (Create, Update, Set, or Delete) the same document more than once.
+func (bw *BulkWriter) Create(doc *DocumentRef, datum interface{}) (*BulkWriterJob, error) {
+	bw.isOpenLock.Lock()
```
enocom, line 148: RLock and RUnlock. ditto below

enocom (approved):

firestore/bulkwriter.go
telpirion: Done.

firestore/bulkwriter.go
```diff
@@ -0,0 +1,331 @@
...
+	j := bw.write(w[0])
+	return j, nil
+}
+
+// checkConditions determines whether this write attempt is valid. It returns
+// an error if either the BulkWriter has already been closed or if it
+// receives a nil document reference.
+func (bw *BulkWriter) checkWriteConditions(doc *DocumentRef) error {
+	open := bw.isOpen
+	if !open {
```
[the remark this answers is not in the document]
telpirion: Done.

enocom (approved):

PR #5974, opened 2022-05-02
chore(storage): add UpdateBucketACL implementation

cojenco:
This adds UpdateBucketACL implementation
- [x] gRPC implementation per discussion with Chris and Noah
- - Add un-exported fields `acl` and `defaultObjectACL` to `BucketAttrsToUpdate`
- - Let UpdateBucketACL to directly utilize UpdateBucket with BucketAttrsToUpdate. This UPDATE operation requires making a GET call first to retrieve existing ACL
- [x] HTTP implementation
- [x] basic emulator test


storage/client_test.go
```diff
@@ -417,6 +417,32 @@ func TestListDefaultObjectACLsEmulated(t *testing.T) {
...
+			Name: bucket,
+		}
+		// Create the bucket that will be retrieved.
+		if _, err := client.CreateBucket(ctx, project, attrs); err != nil {
+			t.Fatalf("client.CreateBucket: %v", err)
+		}
+		entity := AllUsers
+		role := RoleReader
+		want := &ACLRule{Entity: entity, Role: role}
+		got, err := client.UpdateBucketACL(ctx, bucket, entity, role)
```
tritone, line 433: Could we write a test that also verifies that updating does not nuke existing rules? It would be interesting to use PredefinedACL in the bucket creation to get this, maybe.

storage/grpc_client.go
```diff
@@ -425,8 +431,23 @@ func (c *grpcStorageClient) ListBucketACLs(ctx context.Context, bucket string, o
...
-	return nil, errMethodNotSupported
+	// Get the existing acl from BucketAttrs
+	attrs, err := c.GetBucket(ctx, bucket, nil, opts...)
+	if err != nil {
+		return nil, err
+	}
+	var acl []ACLRule
+	acl = append(attrs.ACL, ACLRule{Entity: entity, Role: role})
+	// Set acl in BucketAttrsToUpdate
+	uattrs := &BucketAttrsToUpdate{acl: acl}
```
tritone, line 444: As discussed in Carlos's doc, we need to set an ifMetagenerationMatch precondition here to prevent a data race. We can discuss offline.

storage/http_client.go
```diff
@@ -423,6 +423,28 @@ func (c *httpStorageClient) ListBucketACLs(ctx context.Context, bucket string, o
...
+	}
+	var aclRule ACLRule
+	var err error
+	err = run(ctx, func() error {
+		req := c.raw.BucketAccessControls.Update(bucket, string(entity), acl)
+		configureACLCall(ctx, s.userProject, req)
+		acl, err = req.Do()
+		aclRule = toBucketACLRule(acl)
+		return err
+	}, s.retry, true)
```
tritone, line 441: Idempotency should be taken from s (and will be false ultimately)

tritone (commented):
Good start! A few comments.

storage/http_client.go
cojenco: Right, it's not idempotent. Fixed.

storage/grpc_client.go
cojenco: Added the MetagenerationMatch precondition. I know we talked about it, forgot to add it in >< Thanks for catching this!

storage/client_test.go
cojenco: Good idea! Added PredefinedACL in the bucket creation and asserting with ListBucketACLs

storage/bucket.go
```diff
@@ -1126,6 +1132,13 @@ type BucketAttrsToUpdate struct {
 	// more information.
 	RPO RPO
 
+	// acl is the list of access control rules on the bucket.
```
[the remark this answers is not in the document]
cojenco: Updated

storage/bucket.go
```diff
@@ -986,11 +986,17 @@ func (ua *BucketAttrsToUpdate) toProtoBucket() *storagepb.Bucket {
 		lifecycle = *ua.Lifecycle
 	}
 	var bktACL []*storagepb.BucketAccessControl
+	if ua.acl != nil {
+		bktACL = toProtoBucketACL(ua.acl)
+	}
 	if ua.PredefinedACL != "" {
```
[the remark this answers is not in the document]
cojenco: IIUC, one cannot provide both a predefinedAcl and an acl while patching/updating a bucket. I tried out buckets.patch and buckets.update with the JSON API, and it results in a 409 conflict error. 

So in the library code [here,](https://github.com/googleapis/google-cloud-go/blob/main/storage/bucket.go#L1231-L1240), PredefinedACL and PredefinedDefaultObjectACL take precedence and therefore clears ACLs.

That being said, I believe I need to handle the updateMask with more care under the update semantics. Let me know if this clarifies your question.

storage/http_client.go
tritone: As I mentioned, you should use the setting in s (`s.idempotent`)

tritone (commented):
Nice work! Just a couple more little things

storage/http_client.go
cojenco: Updated to use `s.idempotent` 🙈

storage/bucket.go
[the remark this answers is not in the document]
cojenco: Appreciate the review, Chris! Added some inline comments. Hopefully that's clearer for readers and our future selves.

storage/grpc_client.go
```diff
@@ -425,8 +431,24 @@ func (c *grpcStorageClient) ListBucketACLs(ctx context.Context, bucket string, o
...
+	}
+	var acl []ACLRule
+	acl = append(attrs.ACL, ACLRule{Entity: entity, Role: role})
+	uattrs := &BucketAttrsToUpdate{acl: acl}
+	// Call UpdateBucket with a MetagenerationMatch precondition set.
+	b, err := c.UpdateBucket(ctx, bucket, uattrs, &BucketConditions{MetagenerationMatch: attrs.MetaGeneration}, opts...)
+	if err != nil {
+		return nil, err
+	}
+	aclRule := b.ACL[len(b.ACL)-1]
```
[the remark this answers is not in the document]
cojenco: Done

tritone (approved):

PR #5975, opened 2022-05-02
feat(datastore): adds in, not-in, and != query operations

telpirion:


datastore/query.go
```diff
@@ -171,25 +177,47 @@ func (q *Query) Transaction(t *Transaction) *Query {
...
-// followed by an operator, one of ">", "<", ">=", "<=", or "=".
+// followed by an operator, one of ">", "<", ">=", "<=", "=", and "!=".
 // Fields are compared against the provided value using the operator.
 // Multiple filters are AND'ed together.
 // Field names which contain spaces, quote marks, or operator characters
 // should be passed as quoted Go string literals as returned by strconv.Quote
 // or the fmt package's %q verb.
 func (q *Query) Filter(filterStr string, value interface{}) *Query {
+	// TODO( #5977 ): Add better string parsing (or something)
 	q = q.clone()
```
tritone, line 193: Looks like this clone can be removed if you are cloning in FilterWithArgs as well.

datastore/query.go
```diff
@@ -171,25 +177,47 @@ func (q *Query) Transaction(t *Transaction) *Query {
...
+	q = q.clone()
+
 	f := filter{
-		FieldName: strings.TrimRight(filterStr, " ><=!"),
+		FieldName: fieldName,
 		Value:     value,
 	}
-	switch op := strings.TrimSpace(filterStr[len(f.FieldName):]); op {
+
+	switch o := strings.TrimSpace(strings.ToLower(operation)); o {
```
tritone, line 220: Probably fine not to bother with ToLower, given that this is a new surface. Users will get a sensible error if they put the wrong string in.

datastore/query.go
```diff
@@ -171,25 +177,47 @@ func (q *Query) Transaction(t *Transaction) *Query {
...
+
+// FilterWithArgs returns a derivative query with a field-based filter.
+// The operation parameter takes the following strings: ">", "<", ">=", "<=",
+// "=", "!=", "in", and "not-in".
+// Fields are compared against the provided value using the operator.
+// Multiple filters are AND'ed together.
+// Field names which contain spaces, quote marks, or operator characters
+// should be passed as quoted Go string literals as returned by strconv.Quote
+// or the fmt package's %q verb.
+func (q *Query) FilterWithArgs(fieldName, operation string, value interface{}) *Query {
```
tritone, line 212: I don't love this function name but I can't think of a better one, haha... too bad `Filter` was already taken by something not as good.

tritone (approved):
Overall LGTM, a couple minor ideas/suggestions.

datastore/query.go
telpirion: Good point. Done.

datastore/query.go
tbpg: It does. I think that's OK (more OK to me than `WithArgs` - `Filter` also has args).

datastore/query.go
telpirion: I had wanted to help out folks migrating from SQL `IN` and `NOT-IN`. That said, it's probably better to enforce opinionated syntax.

datastore/query.go
telpirion: A quick straw poll indicates preference for `FilterField()`. Done!

tritone (approved):

datastore/query.go
```diff
@@ -171,25 +177,46 @@ func (q *Query) Transaction(t *Transaction) *Query {
 }
 
 // Filter returns a derivative query with a field-based filter.
+//
+// DEPRECATED. Use the FilterWithArgs() method instead, which supports the same
+// set of operations (and more).
```
tbpg, line 182:
```suggestion
// Deprecated: Use the FilterField method instead, which supports the same
// set of operations (and more).
```

https://github.com/golang/go/wiki/Deprecated

datastore/query_test.go
```diff
@@ -456,6 +459,86 @@ func TestFilterParser(t *testing.T) {
 	}
 }
 
+func TestFilterWithArgs(t *testing.T) {
```
tbpg, line 462:
```suggestion
func TestFilterField(t *testing.T) {
```

datastore/query_test.go
```diff
@@ -456,6 +459,86 @@ func TestFilterParser(t *testing.T) {
...
+	testCases := []struct {
```
tbpg, line 463: Can we declare this separately so we can reuse between the two tests? Or test both `Filter` functions in the same test?

tbpg (changes requested):

datastore/query_test.go
telpirion: We could, but we'd need to break it up into multiple smaller test cases first (since some should fail in `TestFilter()` but pass in `TestFilterField()`) or add multiple test case fields.

tbpg: That makes sense. The downside right now is the list is really long and hard to compare between the two. Perhaps `TestFilterField` tests all of the `TestFilter` cases, but not the other way around? Doesn't seem too onerous. Fine to have more than one group of tests.

telpirion: Eh, you're right. It's better to have fewer test cases to maintain.

datastore/query.go
```diff
@@ -171,25 +177,46 @@ func (q *Query) Transaction(t *Transaction) *Query {
...
-// followed by an operator, one of ">", "<", ">=", "<=", or "=".
+// followed by an operator, one of ">", "<", ">=", "<=", "=", and "!=".
 // Fields are compared against the provided value using the operator.
 // Multiple filters are AND'ed together.
 // Field names which contain spaces, quote marks, or operator characters
 // should be passed as quoted Go string literals as returned by strconv.Quote
 // or the fmt package's %q verb.
 func (q *Query) Filter(filterStr string, value interface{}) *Query {
-	q = q.clone()
+	// TODO( #5977 ): Add better string parsing (or something)
```
tbpg, line 192: Why do we need this?

tbpg (approved):

datastore/query.go
telpirion: The comment? The `TrimRight()` technique of parsing the input `filterStr` seems like it could introduce unwanted side effects.

PR #5983, opened 2022-05-03
fix(pubsub): mark ignore option default for publish flow control

### LINKED ISSUE #5976: pubsub: incorrect default value for PublishSettings LimitExceededBehavior
```golang
// FlowControlSettings controls flow control for messages while publishing or subscribing.
type FlowControlSettings struct {
	// MaxOutstandingMessages is the maximum number of bufered messages to be published.
	// If less than or equal to zero, this is disabled.
	MaxOutstandingMessages int

	// MaxOutstandingBytes is the maximum size of buffered messages to be published.
	// If less than or equal to zero, this is disabled.
	MaxOutstandingBytes int

	// LimitExceededBehavior configures the behavior when trying to publish
	// additional messages while the flow controller is full. The available options
	// include Block (default), Ignore (disable), and SignalError (publish
	// results will return an error).
	LimitExceededBehavior LimitExceededBehavior
}
```
As you can see in the docs, it says that the default behavior is to block but below it is set to ignore.

```golang
// DefaultPublishSettings holds the default values for topics' PublishSettings.
var DefaultPublishSettings = PublishSettings{
	DelayThreshold: 10 * time.Millisecond,
	CountThreshold: 100,
	ByteThreshold:  1e6,
	Timeout:        60 * time.Second,
	// By default, limit the bundler to 10 times the max message size. The number 10 is
	// chosen as a reasonable amount of messages in the worst case whilst still
	// capping the number to a low enough value to not OOM users.
	BufferedByteLimit: 10 * MaxPublishRequestBytes,
	FlowControlSettings: FlowControlSettings{
		MaxOutstandingMessages: 1000,
		MaxOutstandingBytes:    -1,
		LimitExceededBehavior:  FlowControlIgnore,
	},
}
```

hongalex:
Fixes #5976

### DIFF
--- a/pubsub/flow_controller.go
+++ b/pubsub/flow_controller.go
@@ -27,10 +27,10 @@ import (
 type LimitExceededBehavior int
 
 const (
-	// FlowControlBlock signals to wait until the request can be made without exceeding the limit.
-	FlowControlBlock LimitExceededBehavior = iota
 	// FlowControlIgnore disables flow control.
-	FlowControlIgnore
+	FlowControlIgnore LimitExceededBehavior = iota
+	// FlowControlBlock signals to wait until the request can be made without exceeding the limit.
+	FlowControlBlock
 	// FlowControlSignalError signals an error to the caller of acquire.
 	FlowControlSignalError
 )
@@ -47,7 +47,7 @@ type FlowControlSettings struct {
 
 	// LimitExceededBehavior configures the behavior when trying to publish
 	// additional messages while the flow controller is full. The available options
-	// include Block (default), Ignore (disable), and SignalError (publish
+	// are Ignore (disable, default), Block, and SignalError (publish
 	// results will return an error).
 	LimitExceededBehavior LimitExceededBehavior
 }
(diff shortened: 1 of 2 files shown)


shollyman (commented):
Re-ordering the values via the iota is suspicious to me.  If users are using the names consistently it is reasonable, but if they've been setting the int values directly or define their own enum-like things this will potentially cause a behavior change.

hongalex:
> Re-ordering the values via the iota is suspicious to me. If users are using the names consistently it is reasonable, but if they've been setting the int values directly or define their own enum-like things this will potentially cause a behavior change.

Yeah I didn't really want to do this but due to a bug/happenstance this is not a breaking change. There was a bug in the code where we compare the `t.PublishSettings.FlowControlSettings.LimitExceededBehavior` field (default Block since it's the zero value) to LimitExceededBlock. However, since the default value is actually `Ignore`, users have been unable to set this to `Block`.

With this change, `Ignore` should be the default option both in the `DefaultReceiveSettings` struct as well with the iota.

shollyman (approved):
LGTM given the additional context.

feywind (approved):
Approving on behalf of others not in the groups.

Issue #5954, opened 2022-04-29
assuredworkloads: add v1 and release as GA


quartzmo:
> Assured Workloads for Government is now generally available for the FedRAMP Moderate compliance regime.

https://cloud.google.com/assured-workloads/docs/release-notes

quartzmo:
Closing. GA release will be performed per normal procedure after [this date](https://github.com/googleapis/google-cloud-go/blob/41540c47fdf433b00a9ded3fa2a23f1b933ed1be/internal/gapicgen/generator/config.go#L1418) in `config.go`.

PR #5985, opened 2022-05-04
feat(assuredworkloads): start generating apiv1

quartzmo:
refs: #5954


codyoss:
@quartzmo Does this relate to one of the issues you opened? If so, can you link it.

PR #5993, opened 2022-05-04
feat(bigtable): add GC policy to FamilyInfo.

### LINKED ISSUE #6122: bigtable: add full GCPolicy object with PolicyType to bigtable's FamilyInfo
We want to read the GC Policy from bigtable from the external TF provider, so that we can set the resource correctly on read. Current implementation only returns a string policy.

Instead of return a string policy, return the whole object with policy type.

Return just a proto should works, but it's discouraged to pass proto around.

**Additional context**
This item is needed for setting Terraform resource on read.

hoangpham95:
We need the proto of GC policy exposed so that Terraform can read max_version and max_age from the policy, instead of a fully parsed string. Fix #6122.


bigtable/admin.go
```diff
@@ -310,8 +310,9 @@ type TableInfo struct {
 
 // FamilyInfo represents information about a column family.
 type FamilyInfo struct {
-	Name     string
-	GCPolicy string
+	Name          string
+	GCPolicy      string
+	GCPolicyProto *btapb.GcRule
```
tritone, line 315: Have you considered just adding the fields you need directly on this object, rather than exposing the whole proto?

In general we try to avoid exposing proto-types directly via the manual clients. Though, I see a few cases already where they are in this library, so I'll just have this as a question/suggestion rather than being a purist about it.

hoangpham95: We're interested in recursively iterate through this proto to generate a valid JSON field for Terraform. The one we're interested in is  the Rule object in *btapb.GcRule, but the type of that object is not exported, so I opt in to pass the entire proto here.

telpirion: I agree with @tritone. We typically try to provide a veneer over proto message types.

telpirion (approved):

telpirion (commented):
Does this have a corresponding Issue?

bigtable/admin.go
hoangpham95: I see, so how about adding a new struct in the bigtable client and a parser function to convert the proto to the new struct, and pass that struct back from the admin?

tritone (commented):
Generally

bigtable/gc.go
```diff
@@ -25,10 +25,21 @@ import (
...
+	PolicyType_Intersection
+)
+
+type PolicyType = int
+
 // A GCPolicy represents a rule that determines which cells are eligible for garbage collection.
 type GCPolicy interface {
 	String() string
 	proto() *bttdpb.GcRule
+	PolicyType() PolicyType
```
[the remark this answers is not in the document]
hoangpham95: Apology for my late update, it's been busy in the last few days. I have added a wrapper struct called TypedGCPolicy so that it won't break existing user of this client.

bigtable/gc.go
```diff
@@ -158,10 +189,47 @@ func GCRuleToString(rule *bttdpb.GcRule) string {
 	}
 }
 
+func GCRuleToPolicy(rule *bttdpb.GcRule) GCPolicy {
```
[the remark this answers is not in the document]
hoangpham95: Done.

bigtable/gc.go
```diff
@@ -25,10 +25,21 @@ import (
...
 
+const (
+	PolicyType_None = iota
+	PolicyType_MaxAge
+	PolicyType_MaxVersion
+	PolicyType_Union
+	PolicyType_Intersection
+)
+
+type PolicyType = int
```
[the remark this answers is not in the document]
hoangpham95: Done.

bigtable/gc.go
```diff
@@ -25,12 +25,27 @@ import (
 	bttdpb "google.golang.org/genproto/googleapis/bigtable/admin/v2"
 )
 
+type PolicyType = int
+
+const (
+	PolicyTypeNone PolicyType = iota
```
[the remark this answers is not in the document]
hoangpham95: Done.

telpirion:
I still don't see a corresponding issue for this work. Please add.

bigtable/gc.go
```diff
@@ -25,12 +25,27 @@ import (
 	bttdpb "google.golang.org/genproto/googleapis/bigtable/admin/v2"
 )
 
+type PolicyType int
+
+const (
+	PolicyTypeUnknown PolicyType = iota
```
telpirion, line 31: When would the PolicyType be unknown?

telpirion (approved):
This looks good, just one question.

bigtable/gc.go
hoangpham95: It'll be for `NoGcPolicy`. I think the naming could be better, like PolicyTypeUnspecified to match with no policy, but follow the suggestion by tritone@.

hoangpham95:
> I still don't see a corresponding issue for this work. Please add.

Should I link from Buganizer or create a new issue here?

telpirion (commented):
I have concerns about how these API changes will be surfaced & used by customers. Although these changes are useful for the Terraform context, I have concerns about the impact that this can have on customers outside of Terraform.

telpirion (approved):

telpirion:
It looks like something similar to this feature is already available in the Java BT client, so it's reasonable to include here.

PR #5996, opened 2022-05-04
fix: downgrade dep for compatability

codyoss:
It seems like the latest version of github.com/google/martian/v3
is having issues playing nice with Go 1.15. I downgraded the dep
back to where it was and froze it with renovate.

Fixes: #5984
Fixes: #5927
Fixes: #5926


noahdietz (approved):

codyoss:
Failure is unrelated, force pushing this through. There is already an issue open for the failed test.

PR #6013, opened 2022-05-06
chore(storage): add retry metric headers [WIP]

tritone:


storage/invoke.go
```diff
@@ -17,28 +17,38 @@ package storage
...
 )
 
 var defaultRetry *retryConfig = &retryConfig{}
+var xGoogDefaultHeader = fmt.Sprintf("gl-go/%s gccl/%s", version.Go(), sinternal.Version)
 
 // run determines whether a retry is necessary based on the config and
 // idempotency information. It then calls the function with or without retries
 // as appropriate, using the configured settings.
-func run(ctx context.Context, call func() error, retry *retryConfig, isIdempotent bool) error {
+func run(ctx context.Context, call func() error, retry *retryConfig, isIdempotent bool, req interface{ Header() http.Header }) error {
```
noahdietz, line 43: Thinking about how this gets integrated with transport-agnostic clients:

Perhaps `run` accepts not this interface, but rather a function that takes an ID.

```suggestion
func run(ctx context.Context, call func() error, retry *retryConfig, isIdempotent bool, setAttempt func(string))error {
```

Then in transport specific call sites of `run` we just use a closure to define this `func`:

```go
// HTTP
call := c.raw.Buckets.GetIamPolicy(resource)

err = run(ctx, func() error {
  // do stuff
}, c.retry, true, func(attempt string){ setAttemptHeaderHTTP(call, attempt) })
```

```go
// gRPC we set the header using gRPC Metadata on the call context given to the GAPIC
err = run(ctx, func() error {
  // do stuff
}, c.retry, true, func(attempt string){ setAttemptHeaderGRPC(ctx, attempt) })
```

Alternatively, if you want to try to align both HTTP and gRPC on using the call context to communicate Headers/metadata between API layers, maybe there is something that we do in HTTP that pulls the metadata out of the context and adds them to the http.Request.Header 🤷🏻 

I didn't include both the # of attempts and the ID in my example, but you can imagine 😉

noahdietz: Even better, we have helpers that create the closure for us:

```go
func httpAttempts(req interface{  Header() http.Header }) func(id string, attempts int) {
  return func(id string, attempts int) {
    setAttemptHeaderHTTP(req, id, attempts)
  }
}

/* inside a method */ 
// HTTP
call := c.raw.Buckets.GetIamPolicy(resource)

err = run(ctx, func() error {
  // do stuff
}, c.retry, true, httpAttempts(call))
```

tritone: Took your suggestion with the helper func, it works nicely!

storage/invoke.go
```diff
@@ -63,21 +63,30 @@ func run(ctx context.Context, call func() error, retry *retryConfig, isIdempoten
...
-	if req == nil {
+func setRetryHeaderHTTP(req interface{ Header() http.Header }) func(string, int) {
+	return func(invocationID string, attempts int) {
+		if req == nil {
+			return
+		}
+		header := req.Header()
+		invocationHeader := fmt.Sprintf("gccl-invocation-id/%v gccl-attempt-count/%v", invocationID, attempts)
+		xGoogHeader := strings.Join([]string{invocationHeader, xGoogDefaultHeader}, " ")
+		header.Set("x-goog-api-client", xGoogHeader)
```
noahdietz, line 81: I know that we "own" the x-goog-api-client header, but is there a way for this to just _append_ the invocation k-v-pairs to it rather than _set_ it and also be the function that sets the `xGoogDefaultHeader`. If any other part of the stack wants to modify this header to add more info to it, things get complicated.

storage/invoke.go
noahdietz: Great!

tritone:
> Do we have these new headers documented internally? Are they already standardized? If not I am not sure we should add things to this header without more convo

Yes, this has already been standardized and approved internally. Will send you a doc.

storage/invoke.go
tritone: Discussed offline -- it's very difficult to do this with code that retries without appending the same information to the header multiple times. Seems best to take this approach for now.

noahdietz (approved):

storage/invoke_test.go
```diff
@@ -171,16 +174,34 @@ func TestInvoke(t *testing.T) {
...
 			}
-			got := run(ctx, call, test.retry, test.isIdempotentValue)
+			req := &fakeApiaryRequest{header: http.Header{}}
+			got := run(ctx, call, test.retry, test.isIdempotentValue, setRetryHeaderHTTP(req))
 			if test.expectFinalErr && got != test.finalErr {
 				s.Errorf("got %v, want %v", got, test.finalErr)
 			} else if !test.expectFinalErr && got != test.initialErr {
 				s.Errorf("got %v, want %v", got, test.initialErr)
 			}
+			wantHeader := fmt.Sprintf("gccl-invocation-id/.{36} gccl-attempt-count/%v gl-go/.* gccl/", counter)
```
ddelgrosso1, line 184: More so a question of my understanding... just wanted to verify that this test does test a retry and that the invocation ID remains consistent?

tritone: This just checks that the header has the expected format and that the attempt count is the expected number of retries. It doesn't specifically test that the ID remains the same between retries.

I guess I could write a separate test that checks that the ID remains the same.

tritone: Figured out how to make the original test capture the original UUID and make sure that it's consistent.

PR #6017, opened 2022-05-09
feature(datastore): adds in, not-in, and != query operators

telpirion:
This PR was previously [merged](https://github.com/googleapis/google-cloud-go/pull/5975) and then [reverted](https://github.com/googleapis/google-cloud-go/pull/6016).

Waiting for feature to be launched.


telpirion:
Per @kolea2 : Be sure to add integration tests

enocom (approved):

telpirion:
Logged #6184 to track enabling the integration tests for this feature.

lixin9311:
Hi, thanks for releasing this feature.

Is it possible that we could change the signature of the `FilterField(fieldName, operator string, value interface{})` into `FilterField(fieldName, operator string, value ...interface{})`?
So that we could make our code much clear if we want to use the `IN` query.

```go
baseQuery.FilterField("I", "not-in", []interface{}{1, 3, 5, 7})

baseQuery.FilterField("I", "not-in", 1, 3, 5, 7)
```

As for equality queries, we can take the first element of the value, or panic if there is no given value.

Eldelorotiktok:
@

telpirion:
Hi @lixin9311 ! Thank you for your message.

In the future, it's probably best to log an issue in the repo to ask for a feature request. That said, I don't think that I'll accept your requested change. I think that adding a variadic value to account for array values isn't going to be useful for most customers.

PR #6021, opened 2022-05-10
feat(bigquery/storage/managedwriter/adapt): normalize proto3 presence

shollyman:
This PR changes the normalizer to clear the proto3 optional presence
field in the FieldDescriptorProto.  The backend can treat this case
as an instance of proto2 optional.

More background information: https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md


shollyman:
> LGTM, I think it would be even better if you have a e2e so that we have greater confidence that e2e it works.

https://github.com/googleapis/google-cloud-go/pull/6023 will add the round trip testing

PR #6022, opened 2022-05-10
fix(logging): revise automatic resource detection for ingested logs

minherz:
Add deterministic method for detection monitored resource environment using file system.
Remove requirement for the specific call order of the resource detection methods.
Refactor code to improve readability.
Verify parent parameter to match the expected format.
Align global resource creation with the existing monitored resource types.


logging/internal/environment.go
```diff
@@ -0,0 +1,75 @@
...
+
+// ResourceAtttributesGetter declares interface to lookup environment variables and metadata attributes
+type ResourceAtttributesGetter interface {
+	EnvVar(name string) (string, bool)
+	Metadata(path string) (string, bool)
+	ReadAll(path string) (string, error)
+}
+
+var getter ResourceAtttributesGetter = &defaultResourceGetter{
+	metaClient: metadata.NewClient(&http.Client{
```
losalex, line 36: Is `http,Client` used to download a file from remote location?

logging/resource.go
```diff
@@ -34,165 +33,172 @@ type commonResource struct{ *mrpb.MonitoredResource }
...
+var detectedResource = struct {
+	pb    *mrpb.MonitoredResource
+	attrs internal.ResourceAtttributesGetter
+	once  *sync.Once
+}{
+	attrs: internal.ResourceAttributes(),
+	once:  new(sync.Once),
+}
+
+func metadataProjectID() (string, bool) {
```
losalex, line 45: What happens if metadata file is corrupted? Is there a fallback for lookup in env variables?

logging/resource.go
```diff
@@ -34,165 +33,172 @@ type commonResource struct{ *mrpb.MonitoredResource }
...
+	_, versionOK := detectedResource.attrs.EnvVar("GAE_VERSION")
+	_, instanceOK := detectedResource.attrs.EnvVar("GAE_INSTANCE")
+	return serviceOK && versionOK && instanceOK
 }
 
 func detectAppEngineResource() *mrpb.MonitoredResource {
-	projectID, err := metadata.ProjectID()
-	if err != nil {
+	projectID, ok := metadataProjectID()
+	if !ok {
```
losalex, line 75: Correct me if I wrong, but failed metadata reading will make code to bail out, right? If thats a case, why not to use alternate way to retrieve projectID like `os.Getenv("GOOGLE_CLOUD_PROJECT")`?

logging/internal/environment.go
minherz: relevance?

logging/resource.go
minherz: There is no metadata file. Please, see https://cloud.google.com/compute/docs/metadata/overview

losalex: I believe adding this link to code would be a great reference!

logging/resource.go
losalex: Talked offline

logging/internal/environment.go
losalex: Talked offline

losalex (approved):

logging/internal/environment.go
```diff
@@ -0,0 +1,75 @@
...
+	"net"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"cloud.google.com/go/compute/metadata"
+)
+
+// ResourceAtttributesGetter declares interface to lookup environment variables and metadata attributes
```
codyoss, line 28: nit: no need to say it's is an interface the type says that. Please end godoc with a persiod.(This applies elsewhere too)

logging/internal/environment.go
```diff
@@ -0,0 +1,75 @@
...
+// Metadata uses metadata package Client.Get() to lookup for metadata attributes by path
+func (g *defaultResourceGetter) Metadata(path string) (string, bool) {
+	val, err := g.metaClient.Get(path)
+	if err != nil {
+		return val, false
+	}
+	return strings.TrimSpace(val), true
+}
+
+// ReadAll read all content of the file as a string
```
codyoss, line 68: nit: reads

logging/internal/environment.go
```diff
@@ -0,0 +1,75 @@
...
+	"time"
+
+	"cloud.google.com/go/compute/metadata"
+)
+
+// ResourceAtttributesGetter declares interface to lookup environment variables and metadata attributes
+type ResourceAtttributesGetter interface {
+	EnvVar(name string) (string, bool)
+	Metadata(path string) (string, bool)
+	ReadAll(path string) (string, error)
```
codyoss, line 32: Feels odd that this one returns an error and the others are bools. I wonder if you need a second return value at all. Could yoou cue off the string being "" to be a failed lookup? That should never be valid for you I think.

logging/resource.go
```diff
@@ -34,165 +33,172 @@ type commonResource struct{ *mrpb.MonitoredResource }
...
+}{
+	attrs: internal.ResourceAttributes(),
+	once:  new(sync.Once),
+}
+
+func metadataProjectID() (string, bool) {
+	return detectedResource.attrs.Metadata("project/project-id")
+}
+
+func metadataZone() (string, bool) {
```
codyoss, line 49: nit: all of these functions that use detectedResource feel like they should instead be methods on that type. Having functions work on global

logging/resource_test.go
```diff
@@ -0,0 +1,247 @@
...
+	"sync"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	mrpb "google.golang.org/genproto/googleapis/api/monitoredres"
+)
+
+const (
+	there               = "non-empty-value"
```
codyoss, line 28: nit: there seems out of place. Maybe something?

codyoss (changes requested):
Just a few small things, overall looks good. Thanks!

logging/resource_test.go
minherz: this is used as a non empty value. the value itself is validate but is not use. I will place `"anyvalue"` instead

PR #6023, opened 2022-05-11
test(bigquery/storage/managedwriter): data validation testing

shollyman:
This PR does round trip validation of data for the various combinations of protocol buffer syntaxes and features.  The types that aren't tested (uint64, fixed64) are due to BigQuery not supporting unsigned 64 bit values.

We test the following:
* proto2 optional fields - null/unset values
* proto2 optional fields - explicitly set values
* proto2 required fields - explicitly set values
* proto2 optional fields w/default value annotations - null/unset values
* proto3 w/ optional presence - null/unset values

There's tests commented out for two other cases:
* proto3 with non-presence - unset values
* proto3 w/ well-known wrapper types.

As backend features are enabled and we make additional change to the descriptor normalizer, we can augment these validations.


steffnay (approved):

yirutang (approved):
Looks good to me, FYI I have a data type test here:
https://source.corp.google.com/piper///depot/google3/cloud/helix/testing/e2e/bqcli/streaming_api/v2/writeapi/writeapi_data_test_base.py;l=28

I will add more wrapper type support once the change is on test stack.

PR #6026, opened 2022-05-11
chore(storage): add LockBucketRetentionPolicy implementation

noahdietz:
Implements the `LockBucketRetentionPolicy` RPC in the transport-specific clients and adds a test.

Also, this fixes a small bug in `toRetentionPolicy(*raw.BucketRetentionPolicy)` that didn't account for if the `EffectiveTime` was unset i.e. `""`, it would still attempt to parse it as a `time` resulting in a cryptic error. In practice, the API _should_ always set the `EffectiveTime` in the response, but in my opinion we should not return an error if it fails to do so (which the emulator does fail to do).


storage/http_client.go
```diff
@@ -302,7 +302,23 @@ func (c *httpStorageClient) UpdateBucket(ctx context.Context, bucket string, uat
...
+	var metageneration int64
+	if conds != nil {
+		metageneration = conds.MetagenerationMatch
+	}
+	req := c.raw.Buckets.LockRetentionPolicy(bucket, metageneration)
+
+	setClientHeader(req.Header())
+	if s.userProject != "" {
+		req.UserProject(s.userProject)
+	}
```
tritone, line 316: L313-316 is not in the original `BucketHandle.LockRetentionPolicy`; does it need to be included here? Since this is an admin operation, I'm not sure that Requester Pays is really relevant.

tritone (commented):
Looks good, one question

storage/http_client.go
noahdietz: It's your call, I just assumed (and assumed incorrectly) :) What say ye?

tritone: Let's keep it consistent with what's in the http client currently. Can always make changes later if need be.

noahdietz: SGTM! Removed!

tritone (approved):

PR #6028, opened 2022-05-12
test(pubsublite): verify that large message batches can be received by subscribers

tmdiep:
Large Pub/Sub messages can be received by subscribers. Adds a test to verify that one 10 MiB message can be received in a batch.


dpcollins-google (approved):

dpcollins-google:
Nit. Consider setting this to the max value, which is what python and java already do.

tmdiep:
@dpcollins-google I believe we also don't need to do anything for Go, as the generated GAPICs already set the max recv bytes to MaxInt32: https://github.com/googleapis/google-cloud-go/blob/da573ef94c758ac525a82b9d4f1d758df60529c0/pubsublite/apiv1/subscriber_client.go#L47.

I will remove all changes and just keep the unit test to ensure large messages can be received.

PR #6029, opened 2022-05-13
feat(logging): support structured logging functionality

### LINKED ISSUE #3824: logging: add a support of PartialSuccess
Cloud Logging API provides an option to set a field `PartialSuccess`. When it is set to `true`, then we allow backend to write correct log entries and drop incorrect. In case of `false` we either write the whole batch or drop it completely.

Create a new `LoggerOption` for setting `PartialSuccess` field. For example:

```go
func PartialSuccess(v bool) LoggerOption { return partialSuccess{v} }

type partialSuccess struct{ ps bool }

func (r partialSuccess) set(l *Logger) { l.partialSuccess = r.partialSuccess }
```

Then during creation of Logger, user of library can pass option: `logging.PartialSuccess(true)` if they want to enable it.

Set `PartialSuccess` to `true` all the time. Problems:

1. It might break expectations of current users of library.
2. In case when `insertID` is not specified, retry logic might create tons of duplicates.

### LINKED ISSUE #5829: logging: align resource detection algorithm with logging libraries in other languages
tag: structured logging 
Revise the implementation of the automatic resource detection algorithm and align it with implementation in the logging client libraries for other languages (Java and Python).

Requirements:
- minimize dependency on environment variables to avoid unexpected behavior due to involuntary compromising of the runtime environments
- use generation 2 configurations in metadata server that provide deterministic resource detection method
- ensure correct capture of the resource metadata

### LINKED ISSUE #5830: logging: auto-populate source location
tag: structured logging

Add information about filename and code line to all ingested logs with severity `DEBUG` unless this data is not populated by a user.
Allow passing a custom call depth to support use of the library in layered designs. Use call depth = 2 (callers of the `log()` function) by default.

### LINKED ISSUE #5831: logging: validate parent parameter when creating a new client
The `NewClient()` function accepts `parent` parameter which is used to compose the log name for the ingested logs.
Currently, this parameter is not validated. Adding validation will improve developer's experience and prevent unexpected behaviors at early step of the logging client creation instead of failing in each call to log ingestion API.

### LINKED ISSUE #5855: logging: detectResource() is slow when not in Google Env
**Client**

Cloud Logging

**Environment**

Local development

**Code**

```go
package main

import (
	"context"
	"log"
	"os"

	"cloud.google.com/go/logging"
)

func main() {
	ctx := context.Background()

	// Creates a client.
	projectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
	file := "my-credentials-file.json"
	client, err := logging.NewClient(ctx, projectID, option.WithCredentialsFile(file))
	if err != nil {
	}
	defer client.Close()

	logger := client.Logger("my-log") //  <<<----- This hangs for 5 seconds
	_ = logger
}
```

**Expected behavior**

Creating the `logger` from the `client` should be fast.

**Actual behavior**

the code hangs for about 5 seconds

**Additional context**

I run into this issue when I am developing locally and pushing logs to Cloud Logging. But this would also be an issue for anyone using Cloud Logging from a local environment.

The issue is with [isKubernetesEngine()](https://github.com/googleapis/google-cloud-go/blob/db75317f07e054ba3e552abd472b679d91cbf312/logging/resource.go#L142) function. This function starts off by making a call to `metadata.InstanceAttributeValue("cluster-name")`. The call fails and the error is classified as a retry. It then enters a retry loop of max 5 tries with back-off. 

All other detection functions failed quite quickly. Not sure if there is a way to detect GKE without using the metadata server...

This issue was introduced [here](https://github.com/googleapis/google-cloud-go/commit/a2538e16123c21da62036b56df8c104360f1c2d6)

This little issue really makes iterative coding/testing slow.

### LINKED ISSUE #6086: logging: integrate jsonlog-preview into structured logging version
Review the printing logs in the format of the Json strings to stdout/stderr implemented in the [jsonlog-preview](https://github.com/googleapis/google-cloud-go/tree/jsonlog-preview) branch and integrate it into the main branch for the next release of the library. The branch was previously released as the [preview](https://pkg.go.dev/cloud.google.com/go/logging@v1.5.0-jsonlog-preview/jsonlog) version.

### LINKED ISSUE #6087: logging: support W3C traceparent header
Add support for capturing tracing info from [traceparent](https://www.w3.org/TR/trace-context/) W3C trace context.
If both Google trace context and W3C trace context headers present, enforce capturing tracing info from the W3C header.
Due to differences in the definition of the sampling flag, the implementation should not handle the sampling flag of the W3C header.

### LINKED ISSUE #6131: logging: update documentation to include structured logging features
Update README and other relative documentation and code snippets to provide more information about new structured logging features that were added into the 1.5 version of the library.

### LINKED ISSUE #6132: logging: add sending minimal telemetry info to the ingestion mechanism
Need to capture telemetry about the package usage. The info includes the package version and language.
The info will be send once per process that uses the package to ingest application logs.
The telemetry will be sent on first application log ingestion event. The info will be packed as a small instrumentation log entry.

minherz:
Align resource detection with heuristics in logging libraries for other languages (PR #6022)
Adds automatic population of the source location for ingested log entries with severity = Debug (PR #6043)
Adds benchmark for detectResource() to validate the long delays of the logger creation in non-GCP environments (PR #6084)
Adds PartialSuccess logger option to set the partialSuccess flag in all write log entries calls to true (PR #6092)
Adds support for W3C trace context header: `traceparent` in addition to X-Cloud-Trace-Context (PR #6101)
Adds RedirectToStderr and RedirectToStdout logger options to enable printing Jsonified log entries to STDOUT or STDERR instead of ingesting them directly to Cloud Logging (PR #6130)
Adds sending instrumentation info about the package on each use (one time per process) (PR #6210)

Fixes #5829
Fixes #5831
Fixes #5830
Fixes #3824
Fixes #5855
Fixes #6087
Fixes #6086
Fixes #6131
Fixes #6132

### DIFF
--- a/logging/internal/environment.go
+++ b/logging/internal/environment.go
@@ -1,0 +1,75 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+import (
+	"io/ioutil"
+	"net"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"cloud.google.com/go/compute/metadata"
+)
+
+// ResourceAtttributesGetter abstracts environment lookup methods to query for environment variables, metadata attributes and file content.
+type ResourceAtttributesGetter interface {
+	EnvVar(name string) string
+	Metadata(path string) string
+	ReadAll(path string) string
+}
+
+var getter ResourceAtttributesGetter = &defaultResourceGetter{
+	metaClient: metadata.NewClient(&http.Client{
+		Transport: &http.Transport{
+			Dial: (&net.Dialer{
+				Timeout:   1 * time.Second,
+				KeepAlive: 10 * time.Second,
+			}).Dial,
+		},
+	})}
+
+// ResourceAttributes provides read-only access to the ResourceAtttributesGetter interface implementation.
+func ResourceAttributes() ResourceAtttributesGetter {
+	return getter
+}
+
+type defaultResourceGetter struct {
+	metaClient *metadata.Client
+}
+
+// EnvVar uses os.LookupEnv() to lookup for environment variable by name.
+func (g *defaultResourceGetter) EnvVar(name string) string {
+	return os.Getenv(name)
+}
+
+// Metadata uses metadata package Client.Get() to lookup for metadata attributes by path.
+func (g *defaultResourceGetter) Metadata(path string) string {
+	val, err := g.metaClient.Get(path)
+	if err != nil {
+		return ""
+	}
+	return strings.TrimSpace(val)
+}
+
+// ReadAll reads all content of the file as a string.
+func (g *defaultResourceGetter) ReadAll(path string) string {
+	bytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return ""
+	}
+	return string(bytes)
+}
--- a/logging/internal/environment.go
+++ b/logging/internal/environment.go
@@ -25,14 +25,14 @@ import (
 	"cloud.google.com/go/compute/metadata"
 )
 
-// ResourceAtttributesGetter abstracts environment lookup methods to query for environment variables, metadata attributes and file content.
-type ResourceAtttributesGetter interface {
+// ResourceAttributesGetter abstracts environment lookup methods to query for environment variables, metadata attributes and file content.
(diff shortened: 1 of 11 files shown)


losalex (approved):

logging/internal/environment.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"cloud.google.com/go/compute/metadata"
+)
+
+// ResourceAtttributesGetter abstracts environment lookup methods to query for environment variables, metadata attributes and file content.
+type ResourceAtttributesGetter interface {
```
quartzmo, line 29: Typo: Extra `t` in `ResourceAtttributesGetter`.

quartzmo: Also, I'm wondering if maybe this interface isn't really needed, and maybe you can let `ResourceAttributes` return  defaultResourceGetter instead, and just mock `metadata.Client` in the tests instead? Maybe you're already tried that and it didn't work out, but it looks like it might work to me.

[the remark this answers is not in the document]
minherz: the purpose is to allow safer path for testing the behavior by substituting dummy getter in tests.

PR #6034, opened 2022-05-16
chore(storage): switch user_project param to header

noahdietz:
Send user project for billing via system parameter instead of `CommonRequestParams.user_project` as this message will be removed in favor of the header.

cc: @cojenco FYI


codyoss (commented):
Does storage do something really special with user project? There are well defined ways for setting this value with ADC that the client should already be appending: https://github.com/googleapis/google-api-go-client/blob/a96b0372cf1164aa518fa516037df2aa73ca5ff6/transport/grpc/dial.go#L213-L237

noahdietz:
> Does storage do something really special with user project?

Kind of, yes. The Bucket and Object handles all have a `userProject` attribute that can be set by a user and is added to every request in the Apiary via the query-param version of this system parameter `userProject`. So it is already part of the public surface for users to modify the user project used for billing of their requests.

storage/grpc_client.go
```diff
@@ -591,3 +586,10 @@ func (c *grpcStorageClient) CreateHMACKey(ctx context.Context, desc *hmacKeyDesc
 func (c *grpcStorageClient) DeleteHMACKey(ctx context.Context, desc *hmacKeyDesc, opts ...storageOption) error {
 	return errMethodNotSupported
 }
+
+// setUserProjectMetadata appends a project ID to the outgoing Context metadata
```
tritone, line 590: Worth noting that this is for billing purposes, and generally optional except for requester-pays buckets.

tritone (approved):
One docs note, otherwise LGTM. Thanks for making this fix!

storage/grpc_client.go
noahdietz: Update! Thanks.

PR #6042, opened 2022-05-17
chore(internal/apidiff): reenable & refactor for submodule world

### LINKED ISSUE #6038: internal/apidiff: apidiff is suddenly failing broadly on all PRs
Suddenly, the `internal/apidiff` tool is failing (when run via internal/kokoro/check_incompat_changes.sh). We don't know why but since it is blocking every PR, we must disable it until we can figure it out.

[Example job result](https://source.cloud.google.com/results/invocations/f6800f4c-f577-4c39-9af1-b4bb17c014cf/targets/cloud-devrel%2Fclient-libraries%2Fgo%2Fgoogle-cloud-go%2Fpresubmit%2Flatest-version/log).

noahdietz:
This refactors the `internal/apidiff` to work better in the submodule-first state of this repo. Instead of looking at the repo manifest, the script finds all submodules, then diffs each subpackage (that contains a `doc.go`, which every generated surface has) within that submodule and reports the errors. Furthermore, it uses the `os.Command.Dir` property to execute commands in the proper directory rather than `cd` into each directly.

Fixes #6038.


codyoss (approved):
Thanks

PR #6043, opened 2022-05-17
feat(logging): implement auto-population of the calling source location

minherz:
auto-populate the ingest log entry with the source location of the calling code including file name, function name and code line number.
provide designated logger option: `SourceLocationPopulation` to control source population logic. The option can be initiated with the following values:
* `SourceLocationPopulation(DoNotPopulateSourceLocation)` - continue with existing behavior when the source location field of the log entry is not populated automatically (default)
* `SourceLocationPopulation(PopulateSourceLocationForDebugEntries)` - the source location is auto-populated for log entries with Debug severity
* `SourceLocationPopulation(DoNotPopulateSourceLocation)` - the source location is auto-populated for all log entries

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -33,6 +33,7 @@ import (
 	"log"
 	"net/http"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


minherz:
@codyoss can you have a look into the test implementation? maybe you will have a good idea how i can test it without patching the wanted result

PR #6047, opened 2022-05-17
chore(storage): implement GetObject and DeleteObject

cojenco:
This adds GetObject and DeleteObject
- [x] gRPC implementation 
- [x] HTTP implementation
- [x] basic emulator test
- [x] adds a helper `toProtoCommonObjectRequestParams` that sets client-side encryption to the proto library's CommonObjectRequestParams. The equivalent done in HTTP is [`setEncryptionHeaders`](https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L1984-L2002)


storage/client_test.go
```diff
@@ -213,6 +213,63 @@ func TestGetSetTestIamPolicyEmulated(t *testing.T) {
...
+			Name:   fmt.Sprintf("testObject-%d", time.Now().Nanosecond()),
+		}
+		w := veneerClient.Bucket(bucket).Object(want.Name).NewWriter(context.Background())
+		if _, err := w.Write(randomBytesToWrite); err != nil {
+			t.Fatalf("failed to populate test object: %v", err)
+		}
+		if err := w.Close(); err != nil {
+			t.Fatalf("closing object: %v", err)
+		}
+		err = client.DeleteObject(context.Background(), bucket, want.Name, -1, nil)
```
noahdietz, line 236: Nit: Does it make sense to add a `const` for the generation `-1` like `defaultGen`? IDK might make the code easier to read.

storage/grpc_client.go
```diff
@@ -409,12 +409,59 @@ func (c *grpcStorageClient) ListObjects(ctx context.Context, bucket string, q *Q
...
+		Bucket: bucketResourceName(globalProjectAlias, bucket),
+		Object: object,
+	}
+	if err := applyCondsProto("grpcStorageClient.DeleteObject", gen, conds, req); err != nil {
+		return err
+	}
+	if s.userProject != "" {
+		ctx = setUserProjectMetadata(ctx, s.userProject)
+	}
+	// Encryption doesn't apply to Delete.
```
noahdietz, line 424: I don't think this comment is necessary anymore as there isn't mention of encryption anywhere else in this method. Thanks for the note though, so I didn't have to ask if it was relevant give the context of the entire change :)

storage/storage.go
```diff
@@ -2001,6 +2001,19 @@ func setEncryptionHeaders(headers http.Header, key []byte, copySource bool) erro
...
 }
 
+// toProtoCommonObjectRequestParams sets client-side encryption to the proto library's CommonObjectRequestParams.
+func toProtoCommonObjectRequestParams(key []byte) *storagepb.CommonObjectRequestParams {
+	if key == nil {
+		return nil
+	}
+	keyHash := sha256.Sum256(key)
+	return &storagepb.CommonObjectRequestParams{
+		EncryptionAlgorithm:      "AES256",
```
noahdietz, line 2011: Nit: Does it make sense to refactor `"AES256"` into an unexported, global constant since it is used in two places? e.g. `const aes256Algorithm = "AES256"`. Not too opinionated on this though.

noahdietz (commented):
looks great, thanks Cathy

storage/http_client.go
```diff
@@ -378,12 +378,54 @@ func (c *httpStorageClient) ListObjects(ctx context.Context, bucket string, q *Q
...
+	}
+	return err
 }
-func (c *httpStorageClient) GetObject(ctx context.Context, bucket, object string, conds *Conditions, opts ...storageOption) (*ObjectAttrs, error) {
-	return nil, errMethodNotSupported
+
+func (c *httpStorageClient) GetObject(ctx context.Context, bucket, object string, gen int64, encryptionKey []byte, conds *Conditions, opts ...storageOption) (*ObjectAttrs, error) {
+	s := callSettings(c.settings, opts...)
+	req := c.raw.Objects.Get(bucket, object).Projection("full").Context(ctx)
+	err := applyConds("Attrs", gen, conds, req)
```
tritone, line 403: nit: can inline error as in the original.

storage/client_test.go
tritone: Seems like a reasonable idea, it's something of a magic number that comes up in a couple places and has confused me before.

storage/http_client.go
```diff
@@ -378,12 +378,54 @@ func (c *httpStorageClient) ListObjects(ctx context.Context, bucket string, q *Q
...
+	if err != nil {
+		return nil, err
+	}
+	if s.userProject != "" {
+		req.UserProject(s.userProject)
+	}
+	if err := setEncryptionHeaders(req.Header(), encryptionKey, false); err != nil {
+		return nil, err
+	}
+	setClientHeader(req.Header())
```
tritone, line 413: This will be ripped out following #6013 but it's fine to leave it for now

tritone (commented):
Looks really good, nice work!

storage/client_test.go
cojenco: Makes sense👍 added const `defaultGen`

storage/grpc_client.go
cojenco: done :)

storage/storage.go
cojenco: I think that sounds reasonable. Updated to make it a const.

storage/http_client.go
cojenco: done, thanks!

storage/client_test.go
```diff
@@ -260,7 +261,8 @@ func TestGetObjectEmulated(t *testing.T) {
 		if err := w.Close(); err != nil {
 			t.Fatalf("closing object: %v", err)
 		}
-		got, err := client.GetObject(context.Background(), bucket, want.Name, -1, nil, nil)
+		const defaultGen = int64(-1)
```
noahdietz, line 264: I wasn't very specific about this, but perhaps this should be a "global" constant at the top of a file? Perhaps in `storage.go` as well? I can see a few other files (`writer.go`, `storage_test.go` to name a few) where it might be useful, but no need to change any of those now.

storage/storage.go
```diff
@@ -1981,6 +1981,12 @@ func (c composeSourceObj) IfGenerationMatch(gen int64) {
 	}
 }
 
+const (
+	// The AES256 encryption algorithm used with the Customer-Supplied
+	// Encryption Keys feature.
+	aes256Algorithm = "AES256"
```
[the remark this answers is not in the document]
cojenco: I was looking in `storage.go` and didn't see any constants at the top of the file. It seems like constants are declared nearby where they are used. Not sure if I'm reading this correctly.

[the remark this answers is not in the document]
noahdietz: It is defined as a block at the top of the file: https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L75-L87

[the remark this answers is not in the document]
cojenco: Ah okay, thanks! I've added the 2 constants with comments in the top block in `storage.go`. Wasn't entirely sure since I saw multiple const blocks across the file, e.g. [L619 ](https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L619-L622) and [L233](https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L619-L622)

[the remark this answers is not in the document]
noahdietz: Ah well @tritone might have opinions but I usually keep constants in one spot per-file.

noahdietz (approved):

PR #6053, opened 2022-05-18
chore(storage): implement OpenReader, refactor gRPC read

noahdietz:
Implements the `storageClient.OpenReader` interface for both HTTP and gRPC using the existing implementations in `reader.go`. This actually refactors the existing gRPC Reader implementation into the gRPC implementation of the interface. Testing locally showed that the integration tests still pass.


noahdietz:
@tritone @cojenco PTAL when you have a chance!

storage/grpc_client.go
```diff
@@ -607,3 +735,119 @@ func (c *grpcStorageClient) DeleteHMACKey(ctx context.Context, desc *hmacKeyDesc
...
+
+// reopenStream "closes" the existing stream and attempts to reopen a stream and
+// sets the Reader's stream and cancelStream properties in the process.
+//
+// This is an experimental API and not intended for public use.
+func (r *gRPCReader) reopenStream(seen int64) (*storagepb.ReadObjectResponse, error) {
+	// Close existing stream and initialize new stream with updated offset.
+	r.Close()
+
+	res, cancel, err := r.reopen(r.seen)
```
cojenco, line 846: Is param `seen` intended to be passed in here instead?

storage/client_test.go
```diff
@@ -528,6 +528,51 @@ func TestDeleteDefaultObjectACLEmulated(t *testing.T) {
...
+			t.Fatalf("failed to populate test data: %v", err)
+		}
+		if err := w.Close(); err != nil {
+			t.Fatalf("closing object: %v", err)
+		}
+
+		params := &openReaderParams{
+			bucket: bucket,
+			object: want.Name,
+			gen:    -1,
```
cojenco, line 556: [`defaultGen`](https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L94) is now availabe :)

storage/grpc_client.go
noahdietz: Actually, I don't need the param anymore! `r.seen` is correct :) Fixed.

storage/client_test.go
noahdietz: Updated. ty ty!

storage/client.go
```diff
@@ -83,7 +83,7 @@ type storageClient interface {
 	ComposeObject(ctx context.Context, req *composeObjectRequest, opts ...storageOption) (*ObjectAttrs, error)
 	RewriteObject(ctx context.Context, req *rewriteObjectRequest, opts ...storageOption) (*rewriteObjectResponse, error)
 
-	OpenReader(ctx context.Context, r *Reader, opts ...storageOption) error
+	OpenReader(ctx context.Context, params *openReaderParams, opts ...storageOption) (*Reader, error)
```
tritone, line 86: This is really `NewRangeReader` rather than `OpenReader` right? It seems to correspond to the constructor method and you wouldn't call it again if the reader needs to be reopened.

tritone (commented):
Couple minor naming things but overall looks great, very clean refactor!

storage/client.go
```diff
@@ -211,6 +216,16 @@ type userProjectOption struct {
 
 func (o *userProjectOption) Apply(s *settings) { s.userProject = o.project }
 
+type openReaderParams struct {
```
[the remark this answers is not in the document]
noahdietz: I think I prefer the `params` style, as the proto request messages are named "{RPC}Request". I will rename the others afterwards I think.

storage/client.go
noahdietz: Yeah, I think you are right. I will rename it accordingly.

noahdietz: Done!

storage/client.go
```diff
@@ -101,6 +101,11 @@ type storageClient interface {
 	DeleteHMACKey(ctx context.Context, desc *hmacKeyDesc, opts ...storageOption) error
 }
 
+type transportReader interface {
```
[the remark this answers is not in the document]
noahdietz: Good point, switched to `io.ReadCloser`.

tritone (approved):

PR #6073, opened 2022-05-20
feat(retail): start generating apiv2alpha and apiv2beta

quartzmo:


internal/gapicgen/generator/config.go
```diff
@@ -1472,6 +1472,22 @@ var MicrogenGapicConfigs = []*MicrogenConfig{
 		// GA after 2022/06/16
 		ReleaseLevel: "beta",
 	},
+	{
+		InputDirectoryPath:    "google/cloud/retail/v2alpha",
+		Pkg:                   "retail",
+		ImportPath:            "cloud.google.com/go/retail/apiv2alpha",
+		GRPCServiceConfigPath: "retail_grpc_service_config.json",
+		ApiServiceConfigPath:  "retail_v2alpha.yaml",
+		ReleaseLevel:          "ga",
```
noahdietz, line 1481: alpha?

internal/gapicgen/generator/config.go
```diff
@@ -1472,6 +1472,22 @@ var MicrogenGapicConfigs = []*MicrogenConfig{
...
+	},
+	{
+		InputDirectoryPath:    "google/cloud/retail/v2beta",
+		Pkg:                   "retail",
+		ImportPath:            "cloud.google.com/go/retail/apiv2beta",
+		GRPCServiceConfigPath: "retail_grpc_service_config.json",
+		ApiServiceConfigPath:  "retail_v2beta.yaml",
+		ReleaseLevel:          "ga",
```
noahdietz, line 1489: beta?

internal/gapicgen/generator/config.go
quartzmo: @noahdietz @codyoss I was wondering: Is this release level for the API or for the client? And is the release level of a client always that of its module? We update this field to "ga" for a client when we bump its module version to `1.0.0`. The module containing this client is currently `1.2.0`.

noahdietz: This release level applies to the _package_, not necessarily the module. So the alpha and beta APIs should be in packages labeled "beta" and "alpha". This shows up in the package level documentation and thus on the godoc site. While everything is versioned as a module with a single version, documenting the packages as Alpha or Beta means that that packages themselves might not be stable and we reserve the right to break them within a non-major module version.

It's similar to how many Go projects label some _Go APIs_ that they export as "experimental", ([example](https://pkg.go.dev/google.golang.org/grpc#ForceCodec)). Our use of "alpha" and "beta" is for the same purpose, but at the package level.

quartzmo: Got it, thank you! I'll update the PR.

quartzmo: Fixed, thanks!

internal/gapicgen/generator/config.go
quartzmo: Fixed, thanks!

noahdietz (approved):

codyoss (approved):

PR #6074, opened 2022-05-20
doc(storage): clarify metadata restrictions

### LINKED ISSUE #5098: storage: files with a metadata field with / in its key returns error when reading
**Client**

GCS Client

**Environment**

Debian Buster Slim on GKE

**Code**

e.g.
```go
package main

func main() {
  // ...
  obj := bkt.Object(strings.TrimPrefix(req.URL.Path, "/"))
  reader, err := obj.NewReader(context.Background())

  if err != nil {
    return nil, err
  }
}
```

Seems to fail with the following error:

```
Get "https://storage.googleapis.com/gcs-bucket/file.jpg": stream error: stream ID 5; PROTOCOL_ERROR; invalid header field name "x-goog-meta-v1.namespace/width"
```

for a file with the following metadata entry: `v1.namespace/width: 1260`.

**Expected behavior**

To properly read the file.

**Actual behavior**

The error above.

**Additional context**

`gsutil` and other clients seem to read it fine, not sure what's going on. The code itself comes from [imgproxy](https://github.com/imgproxy/imgproxy/blob/78111b22730a1dd4c43bd8398561820de081e694/gcs_transport.go).

tritone:
This is a tricky bit related to the fact that the Go client uses
the XML API for downloads. If the user defined metadata fields
with these characters, downloads fail in a confusing way.

Fixes #5098


storage/storage.go
```diff
@@ -1312,6 +1312,10 @@ type ObjectAttrs struct {
 
 	// Metadata represents user-provided metadata, in key/value pairs.
 	// It can be nil if no metadata is provided.
+	//
+	// For object downloads using Reader, metadata keys are sent as headers.
+	// Therefore, avoid setting metadata keys using characters that are not valid
+	// for headers. See https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6
```
noahdietz, line 1318: Nit: Period at the end of this?

noahdietz (approved):

PR #6094, opened 2022-05-25
chore(bigquery): switch existing formatters to enable error wrapping

shollyman:
Did a quick audit to check for error formatters where we could potentially
wrap errors and were not currently.

This PR also added an explicit unwrapper to the custom error type exposed in bigquery/storage/managedwriter/adapt

Explicitly did not audit/update test code.


codyoss (approved):
Thank you for this!

PR #6113, opened 2022-06-01
fix(pubsub): use MaxInt instead of MaxInt64 for BufferedByteLimit

### LINKED ISSUE #6112: pubsub: v1.22.0 no longer compiles on 32-bit architectures
**Client**

pubsub

**Environment**

Any environment where GOARCH=386 is supported.

**Code**

e.g.
```go
package main

import "cloud.google.com/go/pubsub"

func main() {
	var _ = pubsub.ErrOversizedMessage
}
```

**Expected behavior**

Code compiles.

**Actual behavior**

Code doesn't compile. Error message:

```
Error: ../../../go/pkg/mod/cloud.google.com/go/pubsub@v1.22.0/topic.go:642:39: constant 9223372036854775807 overflows int
```

**Additional context**

This is probably caused by #6009. First noticed in a dependabot PR in fake-gcs-server: https://github.com/fsouza/fake-gcs-server/runs/6695112595

fsouza:
Should fix compatibility with 32-bit architectures.

Closes #6112.

### DIFF
--- a/pubsub/topic.go
+++ b/pubsub/topic.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"math"
 	"runtime"
 	"strings"
 	"sync"
@@ -51,6 +50,13 @@ const (
 	// MaxPublishRequestBytes is the maximum size of a single publish request
 	// in bytes, as defined by the PubSub service.
 	MaxPublishRequestBytes = 1e7
+)
+
+const (
+	// TODO: math.MaxInt was added in Go 1.17. We should use that once 1.17
+	// becomes the minimum supported version of Go.
+	intSize = 32 << (^uint(0) >> 63)
+	maxInt  = 1<<(intSize-1) - 1
 )
 
 // ErrOversizedMessage indicates that a message's size exceeds MaxPublishRequestBytes.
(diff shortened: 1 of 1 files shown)


hongalex (approved):

hongalex:
Thanks for catching this and creating this PR. Will try and get this out later today.

pubsub/topic.go
```diff
@@ -639,7 +639,7 @@ func (t *Topic) initBundler() {
 		// This is because there's no way to set "unlimited" for BufferedByteLimit,
 		// and simply setting it to MaxOutstandingBytes occasionally leads to issues where
 		// BufferedByteLimit is reached even though there are resources available.
-		t.PublishSettings.BufferedByteLimit = math.MaxInt64
+		t.PublishSettings.BufferedByteLimit = math.MaxInt
```
hongalex, line 642: Ah unfortunately this doesn't work since our min supported version is go1.15, `math.MaxInt` was added in 1.17. I believe the solution here is to detect whether the system uses 32 or 64 bits for ints using `strconv.IntSize` and then setting this value to either MaxInt32 or MaxInt64.

fsouza: Oh, that's a bummer. I went with option 2. Let me know if the comment is satisfactory.

PR #6129, opened 2022-06-06
docs(pubsub): add instructions for stream management

### LINKED ISSUE #5677: pubsub: improve documentation with large number of streams
**Client**

Pub/Sub

**Environment**

all / low CPU environments

** Description **

When using Streaming Pull (default) and a large number of streams (NumGoroutines), users may be capped by the number of streams by grpc. This can be mitigated by increasing the grpc connection pool size by doing the following:

```go
opts := []option.ClientOption{
	option.WithGRPCConnectionPool(8),
}
client, err := pubsub.NewClient(ctx, projID, opts...)
```

By default, the number of connections in the pool is the max of 4 or `GOMAXPROCS`/NumCPU. Assuming > 4 CPU cores, default max is 400 streams which is still a good default for most users. However, if they want to use more than this, they will have to pass in the grpc options as described above.

Follow up to #5513 and https://github.com/GoogleCloudPlatform/golang-samples/issues/2416

hongalex:
Fixes #5677


feywind (approved):

maroux:
Is the documentation correct in saying `GOMAXPROCS/NumCPU`. I'm thinking that was meant as `GOMAXPROCS _or_ NumCPU`, but I read it as `GOMAXPROCS divided by NumCPU`. Additionally, the [actual code](https://github.com/googleapis/google-cloud-go/blob/main/pubsub/pubsub.go#L134-L137) seems to be doing a `max(4, GOMAXPROCS)` so is the documentation just wrong, or did I miss something?

hongalex:
Yeah it's meant to be an OR, sorry that's not very clear I'll update that. And yes, max(4, GOMAXPROCS) is correct.

maroux:
Actually, wait, that code says `min(4, GOMAXPROCS)`.

hongalex:
You're right. Thanks for bringing it up! I'll get this addressed in the same PR

PR #6130, opened 2022-06-06
feat(logging): enable redirecting log ingestion to stdout/stderr

### LINKED ISSUE #6086: logging: integrate jsonlog-preview into structured logging version

minherz:
Add `LoggerOption` method to write log entries to the provided `io.Writer` instead of ingesting them to Logging Service. Before writing the log entries are Jsonified as a one line string following the [structured logging format](https://cloud.google.com/logging/docs/structured-logging#special-payload-fields).

Fixes issues 6086, 6131


logging/loggeroption.go
```diff
@@ -0,0 +1,193 @@
...
+type partialSuccessOption struct{}
+
+func (o *partialSuccessOption) set(l *Logger) {
+	l.partialSuccess = true
+}
+
+// RedirectToStdout configures the logger to print log entries to stdout as Jsonified strings instead of ingesting them
+// to Cloud Logging. See https://cloud.google.com/logging/docs/structured-logging#special-payload-fields for Json format
+// of the printed strings. Use this option when need to delegate log entries ingestion to out-of-process logging agent.
+func RedirectToStdout() LoggerOption {
```
codyoss, line 172: I think these two options should be combined to one. Something like `LogJSON(io.Writer)`. If the writer writter is nil default it to Stderr or something. That way there is more flexibility for users that might want to send the logs to a file for instance.

minherz: There were a few considerations to make it this way. I thought that making `RedirectToStdout` is more verbose than `Redirect(os.Stdout)`. Allowing any _writer_ opens a door for misuse and confusion because the format that the library uses to print the logs is supported **only** by GCP logging agents. Other uses will lead to logs not to be ingested to Cloud Logging which in turn defeats the purpose of the library.
The name such as "LogJSON" is misleading due the same considerations. While the format is JSON, it uses special fields that guarantee a specific behavior and only when logs are printed to standard output.

PR #6133, opened 2022-06-07
chore(firestore): Log an error message when an array filter is passed a non-array-value type.

xwkuang5:
This change logs an error message when the operand passed to `ARRAY_CONTAINS_ANY`, `IN` and `NOT_IN` filters are not of array types.


firestore/query.go
```diff
@@ -60,6 +60,13 @@ type Query struct {
 // in queries.
 const DocumentID = "__name__"
 
+// ArrayOperators contains the set of FieldFilter opterators that require an ArrayValue operand.
+var ArrayOperators = map[pb.StructuredQuery_FieldFilter_Operator]bool{
```
telpirion, line 64: Don't export a `var` unless it's okay for customers to overwrite the values. Either declare as `const` or hide (rename to camelCase).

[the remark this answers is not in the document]
xwkuang5: Done

telpirion:
@xwkuang5 Would you please describe how this is a breaking change in the PR description? Thank you!

codyoss:
@telpirion I am not sure if you actually want to label this as a breaking change? If you are doing this there is module renaming and a bunch of other stuff to consider. If this is fixing a case that has always errored server side it may be better to just take it as a fix and have a nice clear description in the commit message of what/why.

telpirion:
@codyoss Couldn't have said it better myself! @xwkuang5 please provide description of how this will break customers. Otherwise, I think we should follow @codyoss 's advice and just label this as a fix.

codyoss:
If it is determined this change should be considered breaking perhaps it is better, for now at least, to implement it in a non-breaking way. Maybe this "feature" can be feature flagged and off by default to change the API behavior for existing users, but allowing them to opt-in to the different behavior.

xwkuang5:
I think I'm missing some context around how we release breaking changes in the SDK. What is the reason for the qualifier "for now at least"? Is it because we don't want to cut a breaking release too often? Allowing them to opt-in the different behavior is likely not very useful: 

1. For users who are using the APIs correctly, the change is not breaking for them
2. For users who have code paths hitting the issues mentioned above, they can just update their codes to supply the right value type to the operator

I'm also curious how we deal with such bugs in general. Is there any past examples I can reference from?

xwkuang5:
PTAL, thanks

firestore/query.go
```diff
@@ -687,6 +695,15 @@ func (f filter) toProto() (*pb.StructuredQuery_Filter, error) {
 	if err != nil {
 		return nil, err
 	}
+	if _, ok := arrayOperators[op]; ok {
```
telpirion, line 698: Put in a code comment here to explain why this check is needed.

firestore/query.go
```diff
@@ -635,6 +636,13 @@ type filter struct {
 	value     interface{}
 }
 
+// arrayOperators contains the set of FieldFilter operators that require an ArrayValue operand.
```
telpirion, line 639: Since this is only used in one place, then move it inside of `toProto()`.

firestore/query.go
```diff
@@ -687,6 +695,15 @@ func (f filter) toProto() (*pb.StructuredQuery_Filter, error) {
 	if err != nil {
 		return nil, err
 	}
+	if _, ok := arrayOperators[op]; ok {
+		switch val.GetValueType().(type) {
+		case *pb.Value_ArrayValue:
+			break
+		default:
+			log.Printf("WARNING firestore: operator %q requires an array value", f.op)
```
telpirion, line 703: I'm not sure about this error message. I agree with the sentiment--let's tell developers that there's an issue here--but I'm not sure whether an all-caps "WARNING" is the best way to go.

@codyoss WDYT?

firestore/query.go
xwkuang5: Done

firestore/query.go
xwkuang5: Done

firestore/query.go
xwkuang5: Switched to direct comparison instead.

firestore/query.go
codyoss: In general we have refrained from any logging in our client libraries. Without proper wholistic support there is not a great way for users to turn off logs if something like this is done except with some global flags that could affect there code as well.

xwkuang5: Given we already have the issue logged and scheduled for fix in the next major breaking change https://github.com/googleapis/google-cloud-go/issues/6330, I think that not logging this warning is also fine. So it seems we can abandon this PR? 

What other information or stakeholders do we need to make a decision here?

PR #6138, opened 2022-06-08
chore(storage): implement Notification methods

cojenco:
This adds interface Notification methods
Tests pending on  testbench new release with fix -https://github.com/googleapis/storage-testbench/pull/360

- [x] `CreateNotification`
- [x] `GetNotification`
- [x] `DeleteNotification`
- [x] `ListNotifications`


storage/grpc_client.go
```diff
@@ -840,6 +841,100 @@ func (c *grpcStorageClient) DeleteHMACKey(ctx context.Context, desc *hmacKeyDesc
...
+		for {
+			n, err := gitr.Next()
+			if err == iterator.Done {
+				break
+			}
+			if err != nil {
+				return err
+			}
+			notifications = append(notifications, n)
+		}
```
noahdietz, line 869: How does the JSON API behave for List Notifications? Does it get every single Notification and return it all in one blob? I think we should work around using the iterator since it complicates things relative to HTTP, WDYT?

The way we'd do that is by "iterating" ourselves via the `NotificationIterator.InteralFetch` method [here](https://github.com/googleapis/google-cloud-go/blob/main/storage/internal/apiv2/storage_client.go#L1140). We can call that until there are no more results, collecting entire pages at a time.

This could of course be even more complicated...

cojenco: IIUC, the JSON API  returns a response with `items`, which is a list of all the Notifications. 

Currently, the library returns a map of notifications instead of an iterator when we call `bucket.Notifications`, and we don't have a `type NotificationIterator`.

I tried using `NotificationIterator.InteralFetch` [here](https://github.com/googleapis/google-cloud-go/commit/09eb7a670adaf595c15c5ba60c8a411a430f8aba#diff-eec5db1002a6674758e3a557d89495986c8450de01512d561e2a7962584c08f2R861) with no success unfortunately.  Could you help point me to how to correctly use `InternalFetch` in this case?

noahdietz: Ah, I would've just created an iterator via the GAPIC ListNotifications, then use _that_ iterator's InternalFetch.

cojenco: ooooh okay I think I understand, updated. what is a good value for pageSize?

noahdietz: The proto documents that the max is 100. You can just omit it and it should default to something reasonable.

storage/grpc_client.go
```diff
@@ -855,17 +855,12 @@ func (c *grpcStorageClient) ListNotifications(ctx context.Context, bucket string
...
-		for {
-			n, err := gitr.Next()
-			if err == iterator.Done {
-				break
-			}
-			if err != nil {
-				return err
-			}
-			notifications = append(notifications, n)
+		notifications, next, err = gitr.InternalFetch(20, "")
```
noahdietz, line 861: Do we _want_ to return all Notifications? If so, we still need to loop until next == "", but instead of add just 1 notification at a time, you can append entire pages.

storage/grpc_client.go
```diff
@@ -840,6 +841,100 @@ func (c *grpcStorageClient) DeleteHMACKey(ctx context.Context, desc *hmacKeyDesc
...
+		ctx = setUserProjectMetadata(ctx, s.userProject)
+	}
+	req := &storagepb.ListNotificationsRequest{
+		Parent: bucketResourceName(globalProjectAlias, bucket),
+	}
+	var notifications []*storagepb.Notification
+	err = run(ctx, func() error {
+		gitr := c.raw.ListNotifications(ctx, req, s.gax...)
+		for {
+			items, nextPageToken, err := gitr.InternalFetch(int(req.PageSize), req.PageToken)
```
tritone, line 861: I was initially confused about how the page token was being preserved between loop iterations but looks like `InternalFetch` modifies the request to handle this. Might be worth adding a comment about this since it does seem confusing!

Also @noahdietz can you just verify that this will preserve the page token if there is a failure from internalFetch mid-way through? Otherwise we might end up with duplicate results.

storage/http_client.go
```diff
@@ -935,6 +935,80 @@ func (c *httpStorageClient) DeleteHMACKey(ctx context.Context, desc *hmacKeyDesc
...
+	}
+	var res *raw.Notifications
+	err = run(ctx, func() error {
+		res, err = call.Context(ctx).Do()
+		return err
+	}, s.retry, true, setRetryHeaderHTTP(call))
+	if err != nil {
+		return nil, err
+	}
+	return notificationsToMap(res.Items), nil
```
tritone, line 957: Interestingly it seems like this implementation doesn't use page tokens-- so if there are more than 1000 notifications you won't get all of them. (I think this would be very uncommon though)

storage/http_client.go
```diff
@@ -935,6 +935,80 @@ func (c *httpStorageClient) DeleteHMACKey(ctx context.Context, desc *hmacKeyDesc
...
+	call := c.raw.Notifications.Delete(bucket, id)
+	if s.userProject != "" {
+		call.UserProject(s.userProject)
+	}
+	return run(ctx, func() error {
+		return call.Context(ctx).Do()
+	}, s.retry, s.idempotent, setRetryHeaderHTTP(call))
+}
+
+func (c *httpStorageClient) GetNotification(ctx context.Context, bucket string, id string, opts ...storageOption) (*Notification, error) {
```
tritone, line 1004: I'm not sure why we have this method in the interface-- it doesn't look to be in the public surface for the library. There is only AddNotification, DeleteNotification, and Notifications (which lists everything), all of which are on BucketHandle. I think we should remove this from the interface and from both transports.

storage/notifications.go
```diff
@@ -91,6 +92,30 @@ func toNotification(rn *raw.Notification) *Notification {
 	return n
 }
 
+func toNotificationFromProto(pbn *storagepb.Notification) *Notification {
+	n := &Notification{
+		ID:               pbn.Name,
```
tritone, line 97: It seems like elsewhere we have used getter functions for this (e.g. pbn.GetName()) rather than accessing attrs directly. @noahdietz do you have a preference? It'd be good to be consistent probably.

tritone (commented):
thanks Cathy!

storage/http_client.go
noahdietz: Might be worth adding a Note or TODO for later mentioning this. I too was surprised to see an un-paginated list api, even if it predates the design standards.

storage/grpc_client.go
noahdietz: >  looks like InternalFetch modifies the request to handle this

I'm not sure what you mean by this. `InternalFetch` will return the nextPageToken, but not modify the next request to use it. That's how I [read its code](https://github.com/googleapis/google-cloud-go/blob/main/storage/internal/apiv2/storage_client.go#L704-L725) at least. I think we need to set `req.PageToken = nextPageToken` right after line 869 before the next iteration.

storage/notifications.go
noahdietz: Yes! Thanks for catching this. We should use the protobuf accessors everywhere rather than direct access. Please update all such instances.

storage/grpc_client.go
noahdietz, line 861:
```suggestion
			items, nextPageToken, err := gitr.InternalFetch(int(req.GetPageSize()), req.GetPageToken())
```

cojenco: Ahh I was not preserving the nextPageToken between iterations >"< updated. Thank you both for the clarification!

storage/http_client.go
cojenco: @tritone Are we considering adding the GetNotification method since it's both supported by JSON and gRPC? I can remove it if not within scope.

storage/notifications.go
cojenco: good catch, updated

storage/http_client.go
```diff
@@ -935,6 +935,83 @@ func (c *httpStorageClient) DeleteHMACKey(ctx context.Context, desc *hmacKeyDesc
...
+
+	if n.ID != "" {
+		return nil, errors.New("storage: AddNotification: ID must not be set")
+	}
+	if n.TopicProjectID == "" {
+		return nil, errors.New("storage: AddNotification: missing TopicProjectID")
+	}
+	if n.TopicID == "" {
+		return nil, errors.New("storage: AddNotification: missing TopicID")
+	}
```
cojenco, line 975: Another question is do we want to remove client side validation here?

storage/grpc_client.go
tritone: Hah @noahdietz that's what I thought originally looking through the code and I talked myself out of it for some reason. Yup that sounds like the right answer.

storage/http_client.go
tritone: Ah so I did a bit of digging and found the following:
* The JSON API [doesn't actually support page tokens](https://cloud.google.com/storage/docs/json_api/v1/notifications/list), probably because...
* There is a [quota](https://cloud.google.com/storage/quotas#buckets) of 100 notifications per bucket.

I'm not sure how this would effect gRPC, if at all.

storage/http_client.go
tritone: I think let's remove it from the interface and remove the implementations as well. It's not in scope to add library surface for this project and we haven't had anyone asking for this functionality.

storage/http_client.go
noahdietz: I think we should keep page size unspecified. The server will "do the right thing" and if for some reason that quota is raised to 500, we won't have to change our code or be worried about what might happen - it will "just work".

Can we add a comment that the server will use a default page_size (just sent an email to inquire what the default actually is), and that our code will not set a page_size? We could leave the param just in case we want to set it later. WDYT?

storage/http_client.go
noahdietz: Perhaps that should go above the interface? So we don't need it here, but we should keep it in level above. That way, both transports benefit, right?

storage/http_client.go
tritone: Sounds fine to me! Agreed that we don't want to depend of behavior that may change going forward.

storage/http_client.go
tritone: Ah good catch. I think this should stay above the interface for now. In the future, we may want to remove it entirely in order to make the API validate this rather than the client.

storage/http_client.go
cojenco: Gotcha... IIUC, based on the [same block of code](https://github.com/googleapis/google-cloud-go/blob/main/storage/internal/apiv2/storage_client.go#L704-L725), `req.GetPageSize()` is not set here and the zero value fallbacks to the API default pageSize of 100. 

So I've added a comment but no change in our code. PTAL and let me know if my understanding is correct.

storage/notifications.go
```diff
@@ -151,6 +176,8 @@ func (b *BucketHandle) AddNotification(ctx context.Context, n *Notification) (re
 
 // Notifications returns all the Notifications configured for this bucket, as a map
 // indexed by notification ID.
+//
+// Note: Pagination is not currently supported.
```
tritone, line 180: Let's remove this; it's not actionable for library users and they are guaranteed to receive all notifications due to entity limits.

storage/grpc_client.go
```diff
@@ -840,6 +841,84 @@ func (c *grpcStorageClient) DeleteHMACKey(ctx context.Context, desc *hmacKeyDesc
...
+	}
+
+	return notificationsToMapFromProto(notifications), nil
+}
+
+func (c *grpcStorageClient) CreateNotification(ctx context.Context, bucket string, n *Notification, opts ...storageOption) (ret *Notification, err error) {
+	ctx = trace.StartSpan(ctx, "cloud.google.com/go/storage.grpcStorageClient.CreateNotification")
+	defer func() { trace.EndSpan(ctx, err) }()
+
+	if n.ID != "" {
```
tritone, line 885: Remove client-side validation here as well (to be left above the interface).

tritone (commented):
Couple more minor things, otherwise LGTM

storage/http_client.go
```diff
@@ -935,6 +935,66 @@ func (c *httpStorageClient) DeleteHMACKey(ctx context.Context, desc *hmacKeyDesc
 	return errMethodNotSupported
 }
 
+// Notification methods.
+
+// ListNotifications returns all the Notifications configured for this bucket, as a map indexed by notification ID.
+//
+// Note: Pagination is not currently supported.
```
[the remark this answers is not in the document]
cojenco: updated, thanks!

tritone (approved):

PR #6150, opened 2022-06-10
chore(internal/gapicgen): support stub generation in google-cloud-go

noahdietz:
Enables generation of protobuf/gRPC stubs in google-cloud-go GAPIC subdirectory. The constructed importpath is mapped using the protobuf/gapic option `-M` to remap the `go_package` in all proto files designated as input to be generated.

Important: Setting `StubsDir` in the microgen configuration **does not** stop generation in genproto.


internal/gapicgen/generator/config.go
```diff
@@ -1524,5 +1530,8 @@ var MicrogenGapicConfigs = []*MicrogenConfig{
 		// GAPIC-level retries and allow the veneer layer to handle retries.
 		ApiServiceConfigPath: "storage_v2.yaml",
 		ReleaseLevel:         "alpha",
+		// Results in protobuf/grpc code generated into the following package:
+		// cloud.google.com/go/storage/internal/apiv2/stubs.
+		StubsDir: "stubs",
```
codyoss, line 1535: Should this have internal in here?

internal/gapicgen/generator/gapics.go
```diff
@@ -308,6 +308,22 @@ func (g *GapicGenerator) microgen(conf *MicrogenConfig) error {
...
 	}
+	if conf.StubsDir != "" {
+		// Enable protobuf/gRPC generation in the google-cloud-go directory.
+		args = append(args, "--go_out=plugins=grpc:"+g.googleCloudDir)
+
+		// For each file to be generated i.e. each file in the proto package,
+		// override the go_package option. Applied to both the protobuf/gRPC
+		// generated code, and to notify the GAPIC generator of the new
+		// import path used to reference those stubs.
+		stubPkg := conf.ImportPath + "/" + conf.StubsDir
```
codyoss, line 319: nit: filepath.Join

internal/gapicgen/generator/config.go
noahdietz: The implementation actually uses this as a suffix on the ImportPath, which in this case is `cloud.google.com/go/storage/internal/apiv2` - it already has internal in it.

internal/gapicgen/generator/gapics.go
noahdietz: Done!

PR #6181, opened 2022-06-15
fix(pubsublite): fixed version in stream headers

tmdiep:
The major and minor versions should be sent as struct number values, rather than string values.


hongalex (approved):

tmdiep:
Thanks @hongalex!

PR #6182, opened 2022-06-15
chore(pubsublite): mark pubsublite library as GA

hongalex:
The Lite library has already been GA for a while, just needed to update this json file

Internal tracking: b/235876797


tmdiep (approved):

noahdietz (approved):

codyoss (changes requested):
Please update the value here also: https://github.com/googleapis/google-cloud-go/blob/main/internal/gapicgen/generator/gapics.go#L527

That is the source of truth that generated the metadata file.

meredithslota:
Relevant value is now https://github.com/googleapis/google-cloud-go/blob/main/internal/gapicgen/generator/gapics.go#L576

codyoss (approved):

Issue #6180, opened 2022-06-15
pubsub: Emit outstanding messages metrics for publishers


schallert:
**Is your feature request related to a problem? Please describe.**

The PubSub package emits [two metrics](https://github.com/googleapis/google-cloud-go/blob/5a2ed6b2cd1c304e0f59daa29959863bff9b5c29/pubsub/trace.go#L163-L164) related to outstanding messages: `OutstandingMessages` and `OutstandingBytes`. The docs mention

> OutstandingMessages is a measure of the number of outstanding messages held by the client before they are processed.

As a user, "messages held by the client" is something I care about in two contexts:

1. In the context of publishing, I want to know how many messages the pubsub client has buffered that have not yet been published and persisted by the PubSub service.
2. In the context of subscribing, I want to know how many messages the pubsub client has buffered that my application has not yet processed.

From the docs, I would have expected `OutstandingMessages` to emit metrics for both of those contexts. However, in practice, it seems like the client only increments `OutstandingMessages` in the context of (2) ([code link](https://github.com/googleapis/google-cloud-go/blob/5a2ed6b2cd1c304e0f59daa29959863bff9b5c29/pubsub/trace.go#L181)). At first this was surprising, as it seems like the flow controller (which is used for both publishing and subscribing) [is responsible](https://github.com/googleapis/google-cloud-go/blob/3f41531579b7a55acea66fec8362e9134125c8a0/pubsub/flow_controller.go#L140) for recording this metric. However, I saw that the metric is only registered with the subscription tag, which makes sense as to why it doesn't record for publishing to a topic.

This is frustrating, because I cannot tell from the pubsub client's internal metrics how many messages have been buffered but not yet persisted to PubSub. This is helpful for debugging cases where the PubSub service may be experiencing latency on Publish RPCs, I am experiencing network problems, etc.

**Describe the solution you'd like**

I would like the pubsub package to emit metrics on how many messages are outstanding in the context of publishing. I don't mind if this requires a separate `View` for producer outstanding metrics and subscriber outstanding metrics.

**Describe alternatives you've considered**

We have implemented our own metrics of how many messages have not yet been fully published by keeping track of `PublishResult`s and decrementing counters when `.Get()` returns. It's not as clean as having this natively in the client.

hongalex:
[...]

PR #6187, opened 2022-06-15
feat(pubsub): report publisher outstanding metrics

schallert:
Fixes https://github.com/googleapis/google-cloud-go/issues/6180.


schallert:
Marking as draft for a bit. Based on internal testing, there's one more path where we need to upsert the topic tag on the context.

pubsub/flow_controller.go
```diff
@@ -176,3 +201,19 @@ func (f *flowController) bound(size int) int64 {
...
+		recordStat(ctx, PublisherOutstandingMessages, n)
+	} else {
+		recordStat(ctx, OutstandingMessages, n)
+	}
+}
+
+func (f *flowController) recordOutstandingBytes(ctx context.Context, n int64) {
+	if f.purpose == flowControllerPurposeTopic {
+		recordStat(ctx, PublisherOutstandingBytes, n)
+	} else {
```
hongalex, line 216: If we're not going to check the unknown value here, we should either remove that value above or explicitly check `f.purpose == flowControllerPurposeSubscription` here.

pubsub/flow_controller.go
```diff
@@ -176,3 +201,25 @@ func (f *flowController) bound(size int) int64 {
 func (f *flowController) count() int {
 	return int(atomic.LoadInt64(&f.countRemaining))
 }
+
+func (f *flowController) recordOutstandingMessages(ctx context.Context, n int64) {
+	switch f.purpose {
+	case flowControllerPurposeTopic:
+		recordStat(ctx, PublisherOutstandingMessages, n)
+	case flowControllerPurposeSubscription:
+		fallthrough
```
schallert, line 210: Made this fall through to maintain backwards compatibility with the old behavior. However, as written there shouldn't be any cases where a "real" client has an unknown purpose, so I'm happy to make `default` a noop. Or I can remove the unknown value entirely and default to `flowControllerPurposeSubscription`.

hongalex: I think we should remove the unknown value since flowcontroller is only used for publisher/subscribers here. Thanks!

PR #6173, opened 2022-06-15
feat(internal/pubsub): prepare internal/message for exactly once

hongalex:
Add new methods to ack handler and introduces `AckWithResult` and `NackWithResult` to the message struct. This change is necessary for exactly once.

This PR is being merged into a separate branch, which will eventually be merged into main once the exactly once feature is ready to merge.

PR #6201, opened 2022-06-16
feat(pubsub): add AckWithResult and NackWithResult to message

hongalex:
This PR adds new versions of ack/nack methods that can return failures or successes via AckResult, which was implemented and merged into a branch in #6173. This PR points the dependency of the root module (cloud.google.com/go) to that branch, which should be removed after exactly once is ready to merge.


pubsub/message.go
```diff
@@ -90,6 +91,32 @@ func toMessage(resp *pb.ReceivedMessage, receiveTime time.Time, doneFunc iterDon
...
+//   if err != nil {
+//       // TODO: Handle error.
+//   }
+type AckResult = ipubsub.AckResult
+
+// AcknowledgeStatus represents the status of an Ack or Nack request.
+type AcknowledgeStatus = ipubsub.AcknowledgeStatus
+
+const (
+	// AcknowledgeStatusSuccess indicates the request was a success.
```
pradn, line 108: Not necessary to repeat the name of the enum in the comment.

pubsub/message.go
```diff
@@ -112,12 +147,37 @@ func (ah *psAckHandler) OnNack() {
 	ah.done(false)
 }
 
+func (ah *psAckHandler) OnAckWithResult() *AckResult {
+	if !ah.exactlyOnceDelivery {
+		return newSuccessAckResult()
+	}
+	ah.ackResult = ipubsub.NewAckResult()
```
pradn, line 154: If this library supports clean shutdown semantics, you might have to store the futures somewhere so you can cleanly close them all at shutdown. Just a thought for the future - this is pt 9 of the "Exactly-once Client Library PR Sequencing" doc.

pubsub/message.go
```diff
@@ -112,12 +147,37 @@ func (ah *psAckHandler) OnNack() {
...
+	ah.done(true)
+	return ah.ackResult
+}
+
+func (ah *psAckHandler) OnNackWithResult() *AckResult {
+	if !ah.exactlyOnceDelivery {
+		return newSuccessAckResult()
+	}
+	ah.ackResult = ipubsub.NewAckResult()
+	ah.done(false)
```
pradn, line 164: Does this `false` mean "done with nack? Perhaps add an inline parameter comment to clarify?

pubsub/streaming_pull_test.go
```diff
@@ -90,14 +90,14 @@ func testStreamingPullIteration(t *testing.T, client *Client, server *mockServer
...
-		wantAckh, _ := msgAckHandler(want)
+		wantAckh, _ := msgAckHandler(want, false)
 		wantAckh.calledDone = true
 		got := gotMap[wantAckh.ackID]
 		if got == nil {
 			t.Errorf("%d: no message for ackID %q", i, wantAckh.ackID)
 			continue
 		}
-		if !testutil.Equal(got, want, cmp.AllowUnexported(Message{}, psAckHandler{}), cmpopts.IgnoreTypes(time.Time{}, func(string, bool, time.Time) {})) {
+		if !testutil.Equal(got, want, cmp.AllowUnexported(Message{}, psAckHandler{}), cmpopts.IgnoreTypes(time.Time{}, func(string, bool, *AckResult, time.Time) {})) {
```
pradn, line 100: Break into a few lines for readability?

pradn (changes requested):

pubsub/message.go
hongalex: This is the Go linter style. Running `golint ./...` gives the following error otherwise:

`pubsub/message.go:108:2: comment on exported const AcknowledgeStatusSuccess should be of the form "AcknowledgeStatusSuccess ..."`

pubsub/message.go
hongalex: Yeah this was inherited from `OnAck()` and `OnNack()`, which shows up slightly before these lines in the same file. I added comments for clarity anyway since these methods are more complicated than the aforementioned ones.

pubsub/streaming_pull_test.go
hongalex: Done

pubsub/message.go
hongalex: Got it, yeah started keeping track of this now.

PR #6202, opened 2022-06-16
test(bigquery): address minor test issues

shollyman:
in the first, the integration tests running too long came into conflict
with a expiration time that was too short.  Lengthened the expiration used
in the test to something more accomodating.

In the second, cleanup some usages of xerrors in testing that can be moved to
errors as we're now no longer supporting older go versions.

Fixes: #6198
Fixes: #6199


codyoss (approved):
LGTM

PR #6204, opened 2022-06-16
fix(storage): allow to use age=0 in OLM conditions

frankyn:
Storage Go doesn't allow setting `AgeInDays=0` after discussing with @tritone. This PR is the agreed upon path to support `Age=0` in OLM conditions. It also notes that other int fields in OLM Conditions do not support `0` values.

Fixes: #6539, #6240


storage/bucket.go
```diff
@@ -698,7 +698,12 @@ const (
 //
 // All configured conditions must be met for the associated action to be taken.
 type LifecycleCondition struct {
+	// AllObject is used to select all objects in a bucket by
+	// setting Age=0.
```
tritone, line 702: ... setting AgeInDays to 0.

storage/bucket.go
```diff
@@ -1617,6 +1634,10 @@ func toLifecycle(rl *raw.BucketLifecycle) Lifecycle {
 			},
 		}
 
+		if rr.Condition.Age == 0 {
+			r.Condition.AllObjects = true
+		}
```
tritone, line 1639: Update this in `toLifecycleFromProto` below as well.

tritone (commented):
Thanks Frank!

storage/bucket.go
```diff
@@ -716,10 +721,12 @@ type LifecycleCondition struct {
 
 	// DaysSinceCustomTime is the days elapsed since the CustomTime date of the
 	// object. This condition can only be satisfied if CustomTime has been set.
+	// Note: Using `0` as the value will not be ignored by the library and not sent to the API.
```
tritone, line 724: "will be ignored", not "will not be ignored". Same in L729 below.

Also, add this note to NumNewerVersions as well.

tritone (commented):
Some comment verbiage stuff. Also, would you mind adding a case to integration tests as well? Could go here: https://github.com/googleapis/google-cloud-go/blob/main/storage/integration_test.go#L255

frankyn:
Thanks @tritone, I'm blocked at the moment. I'm not sure how I can detect if Age is actually returned by the API as `0` or if it's not. 

Do you have tips or guidance here? I tried using NullFields but that's only for Requests and not Responses.

tritone:
Update: conferred with @frankyn and we see that LifecycleRule.Age is implemented as `int64` in google-api-go-client as well. Therefore, we can't see from the response whether the server sent the field or not if it was a zero. This means that we could end up corrupting users' lifecycle rules if they use this library to do a get/modify/update loop on their set of rules for a bucket.

Really this field should have been special-cased in the Apiary library as a pointer, but we can't make that change now in Apiary as it would be a breaking change for Apiary users. We're looking into workarounds and should have an update soon.

storage/bucket.go
```diff
@@ -1562,7 +1554,12 @@ func toLifecycle(rl *raw.BucketLifecycle) Lifecycle {
 				NumNewerVersions:        rr.Condition.NumNewerVersions,
 			},
 		}
-		r.Condition.AgeInDays = getAgeCondition(rr.Condition.Age)
+		if rr.Condition.Age != nil {
+			r.Condition.AgeInDays = *rr.Condition.Age
+			if rr.Condition.Age == googleapi.Int64(0) {
```
tritone, line 1559: I would say `if *rr.Condition.Age == 0`, no need to create another pointer. (Actually I don't know if this works as written, since the pointers will not actually be equal.)

storage/bucket.go
```diff
@@ -1608,6 +1605,11 @@ func toLifecycleFromProto(rl *storagepb.Bucket_Lifecycle) Lifecycle {
 			},
 		}
 
+		// TODO(#6205): This may not be needed for gRPC
```
tritone, line 1608: We probably need to do some deeper digging into how this works/doesn't for gRPC.

tritone (commented):
As I mentioned before, I'd like to have a case added to the integration tests for this as well, showing that the roundtrip of having allObjects set to/from the server works as expected. Can go here: https://github.com/googleapis/google-cloud-go/blob/main/storage/integration_test.go#L255

storage/bucket.go
frankyn: Is there a place where metadata integration tests exist, maybe I'm not looking hard enough? Also this could wait until after this PR is merged unless there's a reason to wait to sus out gRPC first?

tritone (approved):

storage/bucket.go
```diff
@@ -1455,7 +1450,14 @@ func toRawLifecycle(l Lifecycle) *raw.BucketLifecycle {
 			},
 		}
 
-		setAgeCondition(r.Condition.AgeInDays, &rr.Condition.Age)
+		if r.Condition.AgeInDays > 0 {
+			rr.Condition.Age = googleapi.Int64(r.Condition.AgeInDays)
+		}
+
+		if r.Condition.AllObjects {
```
frankyn, line 1457: Question: we may want to return an error here when both `AgeInDays` and `AllObjects` are set at the same time, what do y'all think?

codyoss (approved):
Overall LGTM, just a small nit.

storage/bucket.go
tritone: Discussed offline; AllObjects should take precedence if both are set.

tritone (approved):

storage/bucket.go
tritone: Yes, this can happen later

PR #6210, opened 2022-06-17
chore(logging): add instrumentation

### LINKED ISSUE #6132: logging: add sending minimal telemetry info to the ingestion mechanism

minherz:
Capture instrumentation info about the package by ingesting a single instrumentation log entry per process on the event the process sends an application log first time (i.e. calls Logger.LogSync or Logger.Log).
The instrumentation info capturing reflects the ingestion method:
- direct call to Logging API
- redirecting as JSON to standard output

When ingesting directly, the Log ID is set to `diagnostic-log`. When redirecting to standard output the entry will be ingested to the same Log ID as all other entries.

Fixes #6132


logging/apiv2/config_client.go
```diff
@@ -572,7 +572,7 @@ func (c *configGRPCClient) Connection() *grpc.ClientConn {
 // the `x-goog-api-client` header passed on each request. Intended for
 // use by Google-written clients.
 func (c *configGRPCClient) setGoogleClientInfo(keyval ...string) {
```
codyoss, line 574: Please revert files in this directory, they can only be updated by our regeneration job.

logging/instrumentation.go
```diff
@@ -0,0 +1,54 @@
...
+
+// telemetryPayload
+type telemetryPayload struct {
+	InstrumentationSource []map[string]string `json:"instrumentation_source"`
+	Runtime               string              `json:"runtime,omitempty"`
+}
+
+type instrumentation struct {
+	payload *telemetryPayload
+	once    *sync.Once
```
codyoss, line 32: nit: make a non-pointer type. sync.Once can be used without the need to be initalized at all. Example: https://pkg.go.dev/sync#example-Once

logging/instrumentation.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+var collectedInstrumentation = &instrumentation{
```
codyoss, line 35: nit: this could be a non-pointer an just be written `var collectedInstrumentation instrumentation`

logging/instrumentation.go
```diff
@@ -0,0 +1,54 @@
...
+	once: new(sync.Once),
+}
+
+func detectedInstrumentation() *telemetryPayload {
+	collectedInstrumentation.once.Do(func() {
+		if collectedInstrumentation.payload == nil {
+			collectedInstrumentation.payload = &telemetryPayload{
+				InstrumentationSource: []map[string]string{
+					{
+						"name":    "go",
```
codyoss, line 45: Are these keys standardized anywhere? I wonder if this should be `language`?

logging/instrumentation.go
```diff
@@ -0,0 +1,54 @@
...
+						"version": internal.Version,
```
codyoss, line 46: maybe libraryVersion?

logging/logging_test.go
```diff
@@ -1358,9 +1356,104 @@ func TestRedirectOutputFormats(t *testing.T) {
...
+			got := strings.TrimSpace(buffer.String())
+			if got != tc.want {
+				t.Errorf("got(%v),want(%v):\n", got, tc.want)
+			}
+		})
+	}
+	internal.IngestInstrumentation = iiStatus
+}
+
 func ExampleRedirectAsJson_withStdout() {
```
codyoss, line 1456: Should be `ExampleRedirectAsJSON_ExampleRedirectAsJson_withStdout`

logging/logging.go
```diff
@@ -923,3 +928,41 @@ func printEntryToWriter(entry *logpb.LogEntry, w io.Writer) error {
 	}
 	return err
 }
+
+func (l *Logger) redirectAsJSON(pbe *logpb.LogEntry) (err error) {
+	err = serializeEntryToWriter(pbe, l.redirectOutputWriter)
+	if err == nil {
+		if internal.IngestInstrumentation {
+			internal.IngestInstrumentation = false
```
losalex, line 936: Correct me if I wrong, but if the library is used in multithreaded environment, you might end up adding instrumentation records multiple times. Also this code is repeated below, so I would move it into separate function like `setInstrumentationStatus()` which could update the `internal.IngestInstrumentation` to `false` and return it's previous value - this way you also would be able to make a lock if it is needed.

logging/logging.go
```diff
@@ -923,3 +928,41 @@ func printEntryToWriter(entry *logpb.LogEntry, w io.Writer) error {
 	}
 	return err
 }
+
+func (l *Logger) redirectAsJSON(pbe *logpb.LogEntry) (err error) {
+	err = serializeEntryToWriter(pbe, l.redirectOutputWriter)
+	if err == nil {
```
codyoss, line 934: Flip this logic, keep happy path left justified.

logging/logging.go
```diff
@@ -923,3 +928,41 @@ func printEntryToWriter(entry *logpb.LogEntry, w io.Writer) error {
 	}
 	return err
 }
+
+func (l *Logger) redirectAsJSON(pbe *logpb.LogEntry) (err error) {
```
codyoss, line 932: nit: avoid named returns when possible.

logging/logging.go
```diff
@@ -923,3 +928,41 @@ func printEntryToWriter(entry *logpb.LogEntry, w io.Writer) error {
...
+			instrumentation.LogName = internal.LogPath(l.client.parent, "diagnostic-log")
+			entries = append(entries, instrumentation)
+		}
+	}
+	return entries, partialSuccess
+}
+
+func (l *Logger) instrumentationEntry() (*logpb.LogEntry, error) {
+	ent := Entry{
+		Payload: map[string]*telemetryPayload{
```
codyoss, line 962: This whole map could be cached.

logging/instrumentation.go
minherz: yes, it follows a format defined in the instrumentation PRD.

logging/instrumentation.go
minherz: same as the [previous](https://github.com/googleapis/google-cloud-go/pull/6210#pullrequestreview-1011005733) comment.

logging/logging.go
codyoss: calling a sync.Once multiple times is low cost. I am not sure this extra logic is needed, or I am confused why it might be?

logging/logging.go
codyoss: Applies below too.

logging/logging_test.go
minherz: Did you mean `ExampleRedirectAsJSON_withStdout`?

codyoss (changes requested):
Ran out of time but here is some initial feedback to go off of. Will re-review in full after. Thanks!

logging/logging.go
minherz: The goal was to have a single return. Will flip :+1:

logging/logging.go
minherz: According to the current design this is sent only once. Although the object is small, keeping it in memory is not required.

logging/apiv2/config_client.go
minherz: would it be OK to replicate `versionGo()` function?

logging/logging.go
losalex: The goal is to print instrumentation record only once per process

minherz: I will add the locking mechanism

logging/internal/instrumentation.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+)
+
+// IngestInstrumentation returns status of sending instrumentation info
+func IngestInstrumentation() bool {
+	return ingestInstrumentation
+}
+
+// SetIngestInstrumentation updates status of sending instrumentation info.
+// Returns true if the status is updated and false otherwise.
```
losalex, line 55: nit: Comment should mention something like "Returns true if the status was updated by new value and false if new value is not different than existing"

logging/internal/instrumentation.go
```diff
@@ -0,0 +1,71 @@
...
+
+// telemetryPayload
+type InstrumentationPayload struct {
+	InstrumentationSource []map[string]string `json:"instrumentation_source"`
+	Runtime               string              `json:"runtime,omitempty"`
+}
+
+type instrumentation struct {
+	payload *InstrumentationPayload
+	once    sync.Once
```
codyoss, line 29: nit: like mutex move one above the fields it gaurds

logging/internal/instrumentation.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// IngestInstrumentation returns status of sending instrumentation info
+func IngestInstrumentation() bool {
+	return ingestInstrumentation
+}
+
+// SetIngestInstrumentation updates status of sending instrumentation info.
+// Returns true if the status is changed and false otherwise.
+func SetIngestInstrumentation(f bool) bool {
```
codyoss, line 56: If false is the only value ever sent should this just disable the ingestion and not take a bool at all?

logging/internal/instrumentation.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+)
+
+// IngestInstrumentation returns status of sending instrumentation info
+func IngestInstrumentation() bool {
+	return ingestInstrumentation
+}
+
+// SetIngestInstrumentation updates status of sending instrumentation info.
+// Returns true if the status is changed and false otherwise.
```
codyoss, line 55: nit: remove "and false otherwise"

logging/logging_test.go
```diff
@@ -1358,9 +1355,90 @@ func TestRedirectOutputFormats(t *testing.T) {
...
+	defer client.Close()
+	iiStatus := internal.IngestInstrumentation()
+
+	entry := &logging.Entry{Severity: logging.Info, Payload: "test string"}
+	logger := client.Logger("test-instrumentation")
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got = nil
+			internal.SetIngestInstrumentation(tc.ingestionFlag)
+			err := logger.LogSync(context.TODO(), *entry)
```
codyoss, line 1393: nit: TODO implys there will be eventual code changes here. I would just make a context.Background at the top of this test and pass that to all calls.

logging/logging_test.go
```diff
@@ -1358,9 +1355,90 @@ func TestRedirectOutputFormats(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			if len(got) != tc.want {
+				t.Errorf("got(%v),want(%v):\n", got, tc.want)
```
codyoss, line 1398: nit: space after comma, no need for trailing \n

logging/logging_test.go
```diff
@@ -1358,9 +1355,90 @@ func TestRedirectOutputFormats(t *testing.T) {
...
+			}
+			diagnosticEntry := false
+			for _, ent := range got {
+				if internal.LogIDFromPath("projects/test", ent.LogName) == "diagnostic-log" {
+					diagnosticEntry = true
+					break
+				}
+			}
+			if tc.ingestionFlag != diagnosticEntry {
+				t.Errorf("instrumentation entry misplaced: got(%v),want(%v):\n", diagnosticEntry, tc.ingestionFlag)
```
codyoss, line 1408: nit: space after comma, no need for trailing \n

logging/logging_test.go
```diff
@@ -1358,9 +1355,90 @@ func TestRedirectOutputFormats(t *testing.T) {
...
+		`{"message":"test string","severity":"INFO","timestamp":"seconds:1000"}
+{"message":{"logging.googleapis.com/diagnostic":{"instrumentation_source":[{"name":"go","version":"` + internal.Version + `"}],"runtime":"` + internal.VersionGo() + `"}},"severity":"DEFAULT","timestamp":"seconds:1000"}`,
+		`{"message":"test string","severity":"INFO","timestamp":"seconds:1000"}`,
+	}
+	entry := &logging.Entry{Severity: logging.Info, Payload: "test string"}
+	buffer := &strings.Builder{}
+	logger := client.Logger("test-redirect-output", logging.RedirectAsJSON(buffer))
+	iiStatus := internal.IngestInstrumentation()
+	internal.SetIngestInstrumentation(true)
+	for i, _ := range want {
```
codyoss, line 1427: nit: can remove second return value completly from range instead of using underscore.

logging/logging_test.go
```diff
@@ -1358,9 +1355,90 @@ func TestRedirectOutputFormats(t *testing.T) {
...
+		buffer.Reset()
+		err := logger.LogSync(context.TODO(), *entry)
```
codyoss, line 1429: nit: same as above, prefer context.Background

logging/logging_test.go
```diff
@@ -1358,9 +1355,90 @@ func TestRedirectOutputFormats(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		got := strings.TrimSpace(buffer.String())
+		if got != want[i] {
+			t.Errorf("got(%v),want(%v):\n", got, want[i])
```
codyoss, line 1435: nit: space after comma, remove trailing \n

logging/internal/instrumentation.go
```diff
@@ -0,0 +1,71 @@
...
+func SetIngestInstrumentation(f bool) bool {
+	ok := false
+	if f != ingestInstrumentation {
+		mu.Lock()
```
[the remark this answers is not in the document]
minherz: Good catch, it will not. I will move to use `*sync.Once` so I can manipulate the value in the tests.

logging/internal/instrumentation.go
```diff
@@ -0,0 +1,71 @@
...
+type instrumentation struct {
+	payload *InstrumentationPayload
+	once    sync.Once
+}
+
+var (
+	// ingestInstrumentation keeps tracks of sending instrumentation lib info
+	ingestInstrumentation = true
+	// mu guards ingestInstrumentation
+	mu sync.Mutex
```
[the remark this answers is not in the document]
minherz: will be replaced with `sync.Once` following another comment.

logging/internal/instrumentation.go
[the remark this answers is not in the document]
minherz: I've made relatively large refactoring to address your suggestion. I also used this opportunity to separate "instrumentation" of the log entry being ingested from the main logic of the logging.
To support testing I had to keep the sync primitive in the `internal` package but the rest of the code resides in `logging` package but in the separate file.
LMK if you have further comments.

logging/internal/common.go
```diff
@@ -16,14 +16,20 @@ package internal
...
+	"unicode"
 )
 
 const (
 	// ProdAddr is the production address.
 	ProdAddr = "logging.googleapis.com:443"
 )
 
+// InstrumentOnce guards instrumenting logs one time
+var InstrumentOnce = new(sync.Once)
```
codyoss, line 31: Why is this declared in internal? Feels like it should live next to the code it intends to gaurd and/or be wrapped in a struct. Either this should move or the code that is being guarded should move here.

This can just be `var InstrumentOnce sync.Once`

logging/instrumentation.go
```diff
@@ -0,0 +1,81 @@
...
+			},
+		},
+		Runtime: internal.VersionGo(),
+	}
+)
+
+// instrumentLogs appends log entry with library instrumentation info to the
+// list of log entries on the first function's call.
+func (l *Logger) instrumentLogs(entries []*logpb.LogEntry) ([]*logpb.LogEntry, bool) {
+	instrumentationAdded := false
```
codyoss, line 47:
```suggestion
	var instrumentationAdded bool
```

logging/instrumentation.go
```diff
@@ -0,0 +1,81 @@
...
+	internal.InstrumentOnce.Do(func() {
+		ie, err := l.instrumentationEntry()
+		if err != nil {
+			return
```
codyoss, line 51: Could you add a comment here since the error is being ignored.

codyoss (commented):
Thanks the code is looking great, just a couple of small things.

PR #6230, opened 2022-06-21
fix(storage): allow for  Age int64* type and int64 type

frankyn:
Hi @codyoss and @tritone,

I added a helper func to handle different types (int64 and int64*) when reading data from Storage Apiary client. 
I didn't add from Manual -> Apiary client because I need some help typecasting without Go yelling at me. 

Towards unblocking: #6204


tritone (commented):
Very nice, thanks for adding this Frank!

codyoss:
Some of my comments might not make sense if you refactor the getter/setter into working directly on fields FYI. Feel free to dismiss those

codyoss:
Also forgot to ask, how much of a hot path is this code on for normal operations? There is be a penalty for temporarily using reflection.

frankyn:
> Also forgot to ask, how much of a hot path is this code on for normal operations? There is be a penalty for temporarily using reflection.

Anytime a customer gets Bucket.Attrs(), it will go through this path iff OLM policies are set on the bucket.
It will be a common path for customers but not as common as working with uploading or downloading objects.

storage/bucket.go
```diff
@@ -1503,6 +1503,19 @@ func toCORSFromProto(rc []*storagepb.Bucket_Cors) []CORS {
 	return out
 }
 
+// Used to handle breaking change in Autogen Storage client OLM Age field
+// from int64 to *int64 gracefully in the manual client
+// Method should be removed once breaking change is made and introduced to this client
+func setAgeCondition(age *int64, ageField interface{}) {
+	c := reflect.Indirect(reflect.ValueOf(ageField))
```
codyoss, line 1510: This works, but I usually see this written:

```suggestion
	c := reflect.ValueOf(ageField).Elem()
```

storage/bucket.go
```diff
@@ -1503,6 +1503,19 @@ func toCORSFromProto(rc []*storagepb.Bucket_Cors) []CORS {
 	return out
 }
 
+// Used to handle breaking change in Autogen Storage client OLM Age field
+// from int64 to *int64 gracefully in the manual client
+// Method should be removed once breaking change is made and introduced to this client
+func setAgeCondition(age *int64, ageField interface{}) {
```
codyoss, line 1509: Can stick with this, but I would probably deal with an int64 and for the pointer use-case write: `c.Set(reflect.ValueOf(&age))`

storage/bucket.go
```diff
@@ -1595,6 +1609,21 @@ func toProtoLifecycle(l Lifecycle) *storagepb.Bucket_Lifecycle {
...
 
+// Used to handle breaking change in Autogen Storage client OLM Age field
+// from int64 to *int64 gracefully in the manual client
+// Method should be removed once breaking change is made and introduced to this client
+func getAgeCondition(ageField interface{}) int64 {
+	v := reflect.ValueOf(ageField)
+	if v.Kind() == reflect.Int64 {
+		return v.Interface().(int64)
+	} else if v.Kind() == reflect.Ptr {
+		if v.Interface().(*int64) != nil {
```
codyoss, line 1620:
```
                if val, ok := v.Interface().(*int64); ok {
			return *val
		}
```

codyoss (approved):
LGTM, thanks for all the changes!

tritone (approved):

Albert-Gao:
interesting, this PR "introduced" a breaking change... which breaks `firebase.google.com/go/v4 v4.8.0`
but I do not know why Golang installed the 1.22.1 of the this lib rather than 1.21 that firebase-admin-go depends on :D

tritone:
> interesting, this PR "introduced" a breaking change... which breaks `firebase.google.com/go/v4 v4.8.0` but I do not know why Golang installed the 1.22.1 of the this lib rather than 1.21 that firebase-admin-go depends on :D
> 
> image

@Albert-Gao I believe the error you are seeing is from having incompatible versions of cloud.google.com/go/storage and google.golang.org/api, rather than from this PR. Also, I think it is an issue with your project's build rather than with firebase.google.com/go/v4 because the deps for the firebase package predate this change.

I would try upgrading both cloud.google.com/go/storage and google.golang.org/api to the most recent version, or downgrading google.golang.org/api to <= 0.93.0 if you need to stay on an older version of storage.

If this doesn't work, please file an issue rather than adding a PR comment here.

Albert-Gao:
thanks, @tritone  :)

PR #6256, opened 2022-06-24
fix(bigquery): populate user_email field when using JobFromProject()

seongwon77:
The field itself was added in the commit 7b18bf67abfec72beb23fc10741b46f6f9d61a9f, but it was not added in the field allowlist.  Due to this missing, the function `JobFromProject` has only returned an empty string for email which is wrong.

API reference: https://cloud.google.com/bigquery/docs/reference/rest/v2/Job


shollyman (approved):
Thanks for the contribution!  Apologies it took longer than expected to get to this.  I'll follow this up with https://github.com/googleapis/google-cloud-go/pull/6470 which will augment testing for this.

PR #6257, opened 2022-06-24
feat(pubsub/pstest): subscription message ordering

### LINKED ISSUE #6255: pubsub/pstest: support message ordering
As of right now, fake pubsub subscriptions completely ignore ordering key (it forwards it to the client which handles the invariant of not processing two messages with the same ordering key at once but the order is random). We have a distributed processing algorithm that depends on order of messages with the same key (we need to write last version to the DB) and right now testing it requires using real pubsub/emulator which are a pain for the CI system.

pstest fake server should handle message ordering with the same guarantees the real instance has.

Using emulator or real gcp pubsub for CI but it bloats the costs unnecessarily.

**Additional context**
The issue is caused by the fact that ordering keys are completely ignored by the fake subscription logic and the messages are delivered in a completely random order regardless of the settings (see [code](https://github.com/googleapis/google-cloud-go/blob/3db07733dfc4347796e50254faa91d93b19bd82a/pubsub/pstest/fake.go#L1039), golang map is not ordered by keys but even if it were IDs are compared lexographically as strings, meaning 10 < 2).

arriven:
Fix #6255 

It isn't the most effective solution (something like storing `map[orderingKey || id]struct{id, *message}` would be more effective) but considering that the package is intended to use for unit tests I think that lower risk outweighs the performance penalty. I would be happy to update it to a more effective solution if needed


pubsub/pstest/fake.go
```diff
@@ -1028,6 +1028,32 @@ func (s *subscription) pull(max int) []*pb.ReceivedMessage {
...
+		m  *message
+	}
+	orderingKeyMap := make(map[string]msg)
+	for id, m := range msgs {
+		orderingKey := m.proto.Message.OrderingKey
+		if orderingKey == "" {
+			orderingKey = id
+		}
+		if val, ok := orderingKeyMap[orderingKey]; !ok || m.proto.Message.PublishTime.AsTime().Before(val.m.proto.Message.PublishTime.AsTime()) {
+			orderingKeyMap[orderingKey] = msg{m: m, id: id}
```
hongalex, line 1048: I'm a little confused as to how this works. If there are multiple messages with the same ordering key, doesn't this just override the map at that ordering key? Or is this intention only to deliver 1 message per ordering key at a time?

arriven: Yes, it was intended. Alternative option would be to sort the collection accordingly but I'm not sure if that would work properly with multiple clients reusing the same subscription (AFAIU it's a valid use case and messages with the same ordering key should still not be processed in parallel) so I went ahead with a safer option as my first choice

arriven: Regardless of that I'd be happy to change the approach if you think something else works better

hongalex: > I'm assuming that real pubsub just makes sure that any consecutive messages with the same ordering key get delivered to the same client until all the messages with that key get acknowledged/deleted by retention

Yes, there is subscriber affinity in Pub/Sub with ordering keys but also agree this is probably out of the scope of the fake. I also think that testing subscriber affinity is something that is out of scope for testing in the client library so I think this approach works well. Thanks for authoring this change!

hongalex (approved):

Issue #6278, opened 2022-06-29
bigquery/storage/managedwriter: deadlocks due to locking changes


shollyman:
https://github.com/googleapis/google-cloud-go/pull/6243 changed the surface area of the critical sections, and while addressing one problem introduced new unlocking problems, particularly for error returns.

Need to address this.

PR #6279, opened 2022-06-29
fix(bigquery/storage/managedwriter): address potential deadlocks

shollyman:
TODO: testing

Fixes: #6278


fsaintjacques:
Usage of defer for unlocking would reduce this problem occurring again due to refactor. The for loop in the `append` method could be wrapped in a closure and there the defer would suffer from block-scope issues.

shollyman:
Closing this in favor of PR 6304.

PR #6283, opened 2022-06-30
chore(storage): gRPC ComposeObject implementation

tritone:
Adds implementation in gRPC and HTTP for client.ComposeObject
as well as an emulator test.
    
Required a bit of refactoring around method signature and params
but hopefully this makes it clearer what is going on.

### DIFF
--- a/storage/client.go
+++ b/storage/client.go
@@ -278,11 +278,24 @@ type newRangeReaderParams struct {
 
 type composeObjectRequest struct {
 	dstBucket     string
-	dstObject     string
-	srcs          []string
-	gen           int64
-	conds         *Conditions
+	dstObject     composeDstObject
+	srcs          []composeSrcObject
 	predefinedACL string
+	encryptionKey []byte
+	sendCRC32C    bool
+}
+
+// composeObject manages per-object params for compose sources and destination.
+type composeSrcObject struct {
+	name  string
+	gen   int64
+	conds *Conditions
+}
+
+type composeDstObject struct {
+	name  string
+	conds *Conditions
+	attrs *ObjectAttrs // attrs to set on the destination object.
 }
 
 type rewriteObjectRequest struct {
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -624,7 +624,38 @@ func (c *grpcStorageClient) UpdateObjectACL(ctx context.Context, bucket, object 
 // Media operations.
 
 func (c *grpcStorageClient) ComposeObject(ctx context.Context, req *composeObjectRequest, opts ...storageOption) (*ObjectAttrs, error) {
-	return nil, errMethodNotSupported
+	s := callSettings(c.settings, opts...)
+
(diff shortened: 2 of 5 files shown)


storage/grpc_client.go
```diff
@@ -625,7 +625,51 @@ func (c *grpcStorageClient) UpdateObjectACL(ctx context.Context, bucket, object
...
+		return nil, err
+	}
+	if req.sendCRC32C {
+		dstObjPb.Checksums.Crc32C = &req.dstObject.attrs.CRC32C
+	}
+
+	srcs := []*storagepb.ComposeObjectRequest_SourceObject{}
+	for _, src := range req.srcs {
+		srcObjPb := &storagepb.ComposeObjectRequest_SourceObject{Name: src.name}
+		srcObjPb.Generation = src.gen // what if default value?
```
noahdietz, line 645: Is this question about "What if the src object generation is actually just the default value (e.g. `0`)"?

storage/grpc_client.go
```diff
@@ -625,7 +625,51 @@ func (c *grpcStorageClient) UpdateObjectACL(ctx context.Context, bucket, object
...
+		ctx = setUserProjectMetadata(ctx, s.userProject)
+	}
+
+	dstObjPb := req.dstObject.attrs.toProtoObject(req.dstBucket)
+	dstObjPb.Name = req.dstObject.name
+	if err := applyCondsProto("ComposeObject destination", -1, req.dstObject.conds, dstObjPb); err != nil {
+		return nil, err
+	}
+	if req.sendCRC32C {
+		dstObjPb.Checksums.Crc32C = &req.dstObject.attrs.CRC32C
```
frankyn, line 639: This is a no-op for compose operations; (doesn't do anything).

storage/grpc_client.go
frankyn: If gen = 0 it might have unintended behavior because gen = 0 usually represents object doesn't exist.

storage/client.go
```diff
@@ -278,11 +278,23 @@ type newRangeReaderParams struct {
...
 	dstBucket     string
-	dstObject     string
-	srcs          []string
-	gen           int64
-	conds         *Conditions
+	dstObject     composeDstObject
+	srcs          []composeSrcObject
 	predefinedACL string
+	encryptionKey []byte
+	sendCRC32C    bool
```
frankyn, line 285: sending crc32c doesn't do anything, noted below.

frankyn (changes requested):

storage/grpc_client.go
tritone: I think it's handled here: https://github.com/googleapis/google-cloud-go/blob/0d124aea0fd34510d58167a2b64a17c70c07e43e/storage/storage.go#L2123

storage/client.go
frankyn: I was incorrect, looks like crc32c should be checked if passed in the request. Approving PR, thanks for your patience @tritone

frankyn (approved):

PR #6304, opened 2022-07-05
fix(bigquery/storage/managedstorage): improve internal locking

### LINKED ISSUE #6278: bigquery/storage/managedwriter: deadlocks due to locking changes

shollyman:
This PR addresses some potential deadlocks introduced as part of a refactor
of critical section code.  Based on a suggestion from @[fsaintjacques](https://github.com/fsaintjacques) this
PR switches to closures/funcs where unlocking is handled via defer, rather than explicit unlocking.

This PR adds testing to help ensure further refactors don't reintroduce deadlocks in either
the append or stream closure methods.

This PR also includes small change to the retry predicate to disallow retries for 
context-based errors (cancelations/expirations).

Fixes #6278


bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -351,28 +350,33 @@ func (ms *ManagedStream) Close() error {
...
-	arc, ch, err := ms.getStream(arc, false)
-	if err != nil {
-		return err
-	}
-	if ms.arc == nil {
-		return fmt.Errorf("no stream exists")
-	}
-	err = (*arc).CloseSend()
-	if err == nil {
+	closeErr := func() error {
```
fsaintjacques, line 353: Since this part is not run in a loop, I don't see the need for a closure; could you just `defer ms.mu.Unlock()`? That would also protect the following block from races

```
	if ms.cancel != nil {
		ms.cancel()
		ms.cancel = nil
	}
```

shollyman: Simplified the Close().

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -255,94 +255,93 @@ func (ms *ManagedStream) append(requestCtx context.Context, pw *pendingWrite, op
...
-	var err error
-
 	for {
-		// critical section:  Things that need to happen inside the critical section:
-		//
-		// * Getting the stream connection (in case of reconnects)
-		// * Issuing the append request
-		// * Adding the pending write to the channel to keep ordering correct on response
-		ms.mu.Lock()
+		appendErr, numRows := func() (error, int64) {
```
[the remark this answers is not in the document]
shollyman: Went a bit further and split it out into it's own method, and renamed the internal append() to appendWithRetry() to clarify it's purpose.

steffnay (approved):

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -242,10 +242,75 @@ func (ms *ManagedStream) openWithRetry() (storagepb.BigQueryWrite_AppendRowsClie
...
+		ms.schemaDescriptor = proto.Clone(pw.newSchema).(*descriptorpb.DescriptorProto)
+	}
+	arc, ch, err = ms.getStream(arc, reconnect)
+	if err != nil {
+		return 0, err
+	}
+
+	// Resolve the special work for the first append on a stream.
+	var req *storagepb.AppendRowsRequest
+	ms.streamSetup.Do(func() {
```
codyoss, line 279: question: in the case of reconnects does the special value need to be sent again?

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -242,10 +242,75 @@ func (ms *ManagedStream) openWithRetry() (storagepb.BigQueryWrite_AppendRowsClie
...
-// append handles the details of adding sending an append request on a stream.  Appends are sent on a long
+// lockingAppend handles a single append attempt.  When successful, it returns the number of rows
+// in the request for metrics tracking.
+func (ms *ManagedStream) lockingAppend(requestCtx context.Context, pw *pendingWrite) (int64, error) {
+	// critical section:  Things that need to happen inside the critical section:
+	//
+	// * Getting the stream connection (in case of reconnects)
+	// * Issuing the append request
+	// * Adding the pending write to the channel to keep ordering correct on response
+	ms.mu.Lock()
```
codyoss, line 253: nit: this can move to be under the context err check

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -255,104 +320,43 @@ func (ms *ManagedStream) append(requestCtx context.Context, pw *pendingWrite, op
...
-	var arc *storagepb.BigQueryWrite_AppendRowsClient
-
 	// Critical section: get connection, close, mark closed.
 	ms.mu.Lock()
+	defer ms.mu.Unlock()
+
+	var arc *storagepb.BigQueryWrite_AppendRowsClient
 	arc, ch, err := ms.getStream(arc, false)
 	if err != nil {
 		return err
```
codyoss, line 362: maybe defer the channel close here so it is done in all cases?

codyoss (commented):
LGTM me overall, just a nit and a question

bigquery/storage/managedwriter/managed_stream.go
shollyman: The channel lifecycle shouldn't be bound like this, since it's responsible for the ordering of the asynchronous responses.  We already have existing testing for goroutine leaks to detect these kind of issues.

bigquery/storage/managedwriter/managed_stream.go
codyoss: Ah I missed the reset, that is what I was wondering about!

bigquery/storage/managedwriter/managed_stream.go
codyoss: Is it purposefully not closed if CloseSend errors? That is why I made the suggestion

PR #6313, opened 2022-07-07
chore(storage): RewriteObject implementation

noahdietz:
Implements the RewriteObject transport-agnostic interface methods for both HTTP and gRPC. This API is used to power the Object Copy functionality of the client.


storage/http_client.go
```diff
@@ -684,7 +684,56 @@ func (c *httpStorageClient) ComposeObject(ctx context.Context, req *composeObjec
...
+	if req.predefinedACL != "" {
+		call.DestinationPredefinedAcl(req.predefinedACL)
+	}
+	if err := applyConds("Copy destination", req.gen, req.conds, call); err != nil {
+		return nil, err
+	}
+	if err := applySourceConds(req.srcGen, req.srcConds, call); err != nil {
+		return nil, err
+	}
+	// The userProject, whether source or destination project, is decided by the code calling the interface.
```
tritone, line 707: UserProject is just about billing for the call; doesn't have anything to do with source/destination. (Though in the current code it's set via the source or destination object handles, which is confusing).

storage/http_client.go
```diff
@@ -684,7 +684,56 @@ func (c *httpStorageClient) ComposeObject(ctx context.Context, req *composeObjec
...
+	var err error
+	setClientHeader(call.Header())
+
+	retryCall := func() error { res, err = call.Do(); return err }
+
+	if err := run(ctx, retryCall, s.retry, s.idempotent, setRetryHeaderHTTP(call)); err != nil {
+		return nil, err
+	}
+
+	r := &rewriteObjectResponse{
```
tritone, line 729: https://github.com/googleapis/google-cloud-go/blob/0c820891a987f4a1c6563106e4e54b8e7b46cb30/storage/copy.go#L100 

... requires ObjectSize as well; is there a way we can pass this back also? On the other hand, it doesn't look like `token` is used.

storage/storage.go
```diff
@@ -1838,6 +1841,33 @@ func applySourceConds(gen int64, conds *Conditions, call *raw.ObjectsRewriteCall
 	return nil
 }
 
+func applySourceCondsProto(gen int64, conds *Conditions, call *storagepb.RewriteObjectRequest) error {
```
tritone, line 1844: So odd that this requires a whole separate method when compose sources do not (seems like this is already the case in existing code though).

storage/client_test.go
```diff
@@ -276,6 +276,48 @@ func TestGetObjectEmulated(t *testing.T) {
...
+			srcGen:    defaultGen,
+		}
+		got, err := client.RewriteObject(context.Background(), req)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !got.done {
+			t.Fatal("didn't finish writing!")
+		}
+		if want := int64(len(randomBytesToWrite)); got.written != want {
```
tritone, line 315: Just noting that we're not guaranteed for the entire copy to complete in one call to RewriteObject. It would be nice to have a test with a larger file which will take multiple calls, but I think it's fine to let integration tests cover that case.

tritone (commented):
Sorry for the delay, generally looks great! A few minor things.

storage/http_client.go
noahdietz: I'll just remove the comment then.

storage/http_client.go
noahdietz: That is reachable by doing `res.resource.ObjectSize` of a rewriteObjectResponse. 

Token is necessary because currently `callRewrite` sets `c.Token`. We will need to do the same thing with the code that handles the `rewriteObjectResponse` in copy.go

storage/storage.go
noahdietz: Right there was already a source-specific conds helper for HTTP, so had to copy that for gRPC. Weird!

storage/client_test.go
noahdietz: Makes sense to me, let's leave it for the integration tests for now, but definitely something we need to test.

storage/client.go
```diff
@@ -286,16 +286,20 @@ type composeObjectRequest struct {
...
-	dstBucket     string
-	dstObject     string
-	dstKeyName    string
-	attrs         *ObjectAttrs
-	gen           int64
-	conds         *Conditions
-	predefinedACL string
-	token         string
+	srcBucket        string
+	srcObject        string
```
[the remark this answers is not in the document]
noahdietz: I refactored the existing one to make them "generic". Good idea! Done.

storage/http_client.go
tritone: Makes sense, I wasn't looking at the response closely enough.

storage/client.go
```diff
@@ -278,31 +278,33 @@ type newRangeReaderParams struct {
...
 
 type rewriteObjectRequest struct {
-	srcBucket     string
-	srcObject     string
-	dstBucket     string
-	dstObject     string
-	dstKeyName    string
+	srcObject     sourceObject
+	dstObject     destinationObject
 	attrs         *ObjectAttrs
```
tritone, line 308: I believe now you can get of `attrs` and `conds` in favor of the relevant `destinationObject` fields, and `gen` in favor of `sourceObject.gen`.

tritone (approved):
One more note from the refactoring; otherwise LGTM

storage/client.go
noahdietz: Ah good call, I will do that.

storage/grpc_client.go
```diff
@@ -783,7 +783,7 @@ func (c *grpcStorageClient) RewriteObject(ctx context.Context, req *rewriteObjec
 	if s.userProject != "" {
 		ctx = setUserProjectMetadata(ctx, s.userProject)
 	}
-	if err := applyCondsProto("Copy destination", req.gen, req.conds, call); err != nil {
+	if err := applyCondsProto("Copy destination", req.srcObject.gen, req.dstObject.conds, call); err != nil {
```
tritone, line 786: This is still not correct -- you should just pass -1 for generation to indicate the latest. Passing a specific generation for a destination object is not possible.

tritone (approved):

storage/grpc_client.go
tritone: Yup looks good now! Thanks!

PR #6314, opened 2022-07-08
feat(bigtable): add PolicyType for GCPolicy and expose public methods for different types of GC policies

hoangpham95:
We want to expose the capability of checking types of the GCPolicy. Since we're actively avoids exposing `proto()` method, this PR is adding the GC Policy types for GCPolicy and several methods to access: 
- `version` for `MaxVersionGCPolicy`
- `duration` for `MaxAgeGCPolicy`


bigtable/gc.go
```diff
@@ -35,7 +35,7 @@ const (
 // A GCPolicy represents a rule that determines which cells are eligible for garbage collection.
 type GCPolicy interface {
 	String() string
-	proto() *bttdpb.GcRule
+	Proto() *bttdpb.GcRule
```
codyoss, line 38: In general we try not to expose protos directly out from the manual layers. Also adding a new method to an interface is a breaking change in Go

codyoss (changes requested):

bigtable/gc.go
telpirion: Agreed with @codyoss . IIRC, we discussed not exposing protos directly in #5993 . Why this breaking change?

telpirion (changes requested):

bigtable/gc.go
codyoss: Pretend a user implemented your that interface for their type:

```go
type Foo struct {}
```

telpirion (approved):

bigtable/gc.go
```diff
@@ -90,21 +108,21 @@ func (up unionPolicy) proto() *bttdpb.GcRule {
...
 // MaxVersionsPolicy returns a GC policy that applies to all versions of a cell
 // except for the most recent n.
-func MaxVersionsPolicy(n int) GCPolicy { return maxVersionsPolicy(n) }
+func MaxVersionsPolicy(n int) GCPolicy { return MaxVersionsGCPolicy(n) }
 
-type maxVersionsPolicy int
+type MaxVersionsGCPolicy int
 
-func (mvp maxVersionsPolicy) String() string { return fmt.Sprintf("versions() > %d", int(mvp)) }
+func (mvp MaxVersionsGCPolicy) String() string { return fmt.Sprintf("versions() > %d", int(mvp)) }
```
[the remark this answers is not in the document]
hoangpham95: Done.

codyoss (dismissed):
I'll defer to Eric for the rest of the review but I would ask:

1. Add documentation to everything being made public by these changes.
2. Evaluate if all the methods on these previously private type really should be made public
3. It seems like a large increase in surface, I would make sure it is all really need for this feature.
4. Update body message in this PR to be more descriptive of the current changes

hoangpham95:
> I'll defer to Eric for the rest of the review but I would ask:
> 
> 1. Add documentation to everything being made public by these changes.
> 2. Evaluate if all the methods on these previously private type really should be made public
> 3. It seems like a large increase in surface, I would make sure it is all really need for this feature.
> 4. Update body message in this PR to be more descriptive of the current changes

1 and 4 is done. For 2 and 3, the following snippets is going to be added to the Terraform package:
```go
func gcPolicyToGCRuleString(gc bigtable.GCPolicy) map[string]interface{} {
    result := make(map[string]interface{})
    switch (bigtable.GetGCPolicyType(fi.FullGCPolicy)) {
    case PolicyMaxAge:
        age := gc.(MaxAgeGCPolicy).GetDurationString()
        result["max_age"] = age
    case PolicyMaxVersion:
        version := gc.(MaxVersionGCPolicy).(int)
        result["version"] = version
    case PolicyUnion:
        result["mode"] = "union"
        result["rules"] = []interface{}{}
        for _, c := range gc.(UnionGCPolicy).Children {
            result["rules"] = append(result["rules"], gcPolicyToGCRuleString(c))
        }
    case PolicyIntersection:
        result["mode"] = "intersection"
        result["rules"] = []interface{}{}
        for _, c := range gc.(IntersectionGCPolicy).Children {
            result["rules"] = append(result["rules"], gcPolicyToGCRuleString(c))
        }
    default:
        fallthrough
    }
    return result
}
```

Issue #6321, opened 2022-07-11
bigquery: storage/managedwriter: wrong error at the wrong place for large batch inserts


maffka123:
**Client**

go/bigquery/storage/managedwriter

**Environment**

MacOS Monterey 12.3
Windows 10

**Go Environment**

$ go version: go1.18 darwin/amd64  or go1.17.5 windows/amd64

**Code**

e.g.
```go
package main

import (
	"cloud.google.com/go/bigquery/storage/managedwriter"
	"context"
	storagepb "google.golang.org/genproto/googleapis/cloud/bigquery/storage/v1"
	"google.golang.org/protobuf/proto"
	"google.golang.org/protobuf/reflect/protodesc"
	"log"
)

func main() {
	ctx := context.Background()
	ms, client := prepBQStream()
	p := prepRows()
	res, err := ms.AppendRows(ctx, p)
	if err != nil {
		log.Fatalf("AppendRows call error: %v", err)
	}
	_, err = res.GetResult(ctx)
	if err != nil {
		log.Fatalf("append returned error: %v", err)
	}
	_, err = ms.Finalize(ctx)
	if err != nil {
		log.Fatalf("error during Finalize: %v", err)
	}
	req := &storagepb.BatchCommitWriteStreamsRequest{
		Parent:       managedwriter.TableParentFromStreamName(ms.StreamName()),
		WriteStreams: []string{ms.StreamName()},
	}

	resp, err := client.BatchCommitWriteStreams(ctx, req)
	if err != nil {
		log.Fatalf("client.BatchCommit: %v", err)
	}
	if len(resp.GetStreamErrors()) > 0 {
		log.Fatalf("stream errors present: %v", resp.GetStreamErrors())
	}
	client.Close()

}

func prepRows() [][]byte {
	rows := [][]byte{}
	a := "akkkkkdkd"
	b := "bwefdffdgfg"
	g := "gsrertrthh"
	datetime := proto.String(time.Now().Format("2006-01-02 15:04:05"))
	q := 567865432.74

	row := Row{Col1: &a, Col2: &b, Col3: &g, Date: datetime, Qty: &q}
	l, err := proto.Marshal(&row)
	if err != nil {
		log.Fatalf("could not marshal proto %v", err)
	}

	for i := 0; i < 1000000; i++ {
		rows = append(rows, l)
	}
	return rows
}

func prepBQStream() (*managedwriter.ManagedStream, *managedwriter.Client) {
	ctx := context.Background()
[...]

PR #6338, opened 2022-07-14
fix(bigquery/storage/managedwriter): improve network reconnection

shollyman:
Issuing a sufficiently large single append request is enough to trigger
the server backend to close an existing grpc stream.  This PR addresses
the problem by allowing a failed request to signal that subsequent
requests should request a new grpc stream connection.

This PR also adds an integration test that induces the failure by
issuing a large request, and ensures subsequent requests succeed.

Towards: https://github.com/googleapis/google-cloud-go/issues/6321


bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -184,6 +185,11 @@ func (ms *ManagedStream) getStream(arc *storagepb.BigQueryWrite_AppendRowsClient
 		return nil, nil, ms.err
 	}
 
+	// Previous activity on the stream indicated it is not healthy, so propagate that as a reconnect.
```
yirutang, line 188: What are you seeing here? I thought passing in a 11MB message should be a in stream failure from server point of view. Or maybe some stubby layer failure is there before this?

shollyman: Discussed out of band.  Internal issue 239050626 filed internally with more details.

codyoss (approved):
Nice usability change!

bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -297,6 +306,11 @@ func (ms *ManagedStream) lockingAppend(requestCtx context.Context, pw *pendingWr
 		err = (*arc).Send(pw.request)
 	}
 	if err != nil {
+		// Transient connection loss.  If we got io.EOF from a send, we want subsequent appends to
+		// reconnect the network connection for the stream.
+		if err == io.EOF {
```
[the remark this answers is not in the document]
shollyman: That's deep enough down the stack it's not clear if it will stay consistent, I'll switch over to errors.Is.

PR #6344, opened 2022-07-15
fix(spanner): fix pool.numInUse exceeding MaxOpened

hengfengli:
Fix a race condition issue that occasionally, `pool.numInUse` can exceed `MaxOpened`.


rahul2393 (approved):
LGTM

PR #6346, opened 2022-07-15
feat(logging): add logger minimal severity

### LINKED ISSUE #6345: logging: add logger minimal severity
**Problem:**
All logs are sent to the Stackdriver, including debug logs. This is a problem for the security and privacy of logged data. Debug logs may contain sensitive data (for example: credit cards numbers or incoming request auth token).

**Solution:**
I suggest to add logger option `MinimalSeverity` (default value `Default`). `Logger.Log` and `Logger.LogSync` methods compare log entry severity with minimal severity. If entry severity lesser than minimal severity then log entry is ignored.

**Considered alternatives:**
I considered the option of creating my own wrapper over the logger, which would implement the above logic. But this logic is implemented in many well-known logging libraries (for example: zap or logrus), I decided to try to create this issue.

kimmelserj:
Fixes #6345


logging/loggeroption.go
```diff
@@ -187,3 +187,11 @@ type redirectOutputOption struct {
 func (o *redirectOutputOption) set(l *Logger) {
 	l.redirectOutputWriter = o.writer
 }
+
+// MinimalSeverity is the allowed severity of log entries.
+// If the log entry has the severity lesser than the minimal severity, then the log entry will be ignored.
+func MinimalSeverity(s Severity) LoggerOption { return minimalSeverity(s) }
```
minherz, line 193: nit: I'd suggest to make the name more verbose. "MinimalSeverity" requires additional clarification. Consider "IngestedSeverityThreshold" or "MinimalLoggedSeverity" or something similar.

logging/logging.go
```diff
@@ -628,6 +630,9 @@ func jsonValueToStructValue(v interface{}) *structpb.Value {
 // and will block, it is intended primarily for debugging or critical errors.
 // Prefer Log for most uses.
 func (l *Logger) LogSync(ctx context.Context, e Entry) error {
+	if e.Severity < l.minimalSeverity {
```
minherz, line 633: nit: why not `<=`? Inclusive threshold is implied by the name: "minimal severity" -> "severity below which nothing is ingested"
nit: should it return a specialized error indicating that the ingestion did not take place? returning `nil` means everything is OK but the job still was not done.

logging/logging_test.go
```diff
@@ -242,6 +325,87 @@ func TestLogAndEntries(t *testing.T) {
 	}
 }
 
+func TestLogWithMinimalSeverity(t *testing.T) {
```
minherz, line 328: nit: it looks like a great opportunity for refactoring since both `TestLogSyncWithMinimalSeverity()` and `TestLogWithMinimalSeverity()` have very similar code lines.

minherz (commented):
Please, have a look at provided comments.

minherz:
@daniel-sanche please have a look into this new contribution.

logging/loggeroption.go
kimmelserj: Ok. I will rename `MinimalSeverity` to `MinimalLoggedSeverity`.

logging/logging.go
kimmelserj: > nit: why not <=? Inclusive threshold is implied by the name: "minimal severity" -> "severity below which nothing is ingested"

I came from the definition of a minimum: https://en.wikipedia.org/wiki/Maxima_and_minima. The minimal severity is the minimum allowed severity value. That's why I used `<` in the expression.

> nit: should it return a specialized error indicating that the ingestion did not take place? returning nil means everything is OK but the job still was not done.

An error is defined as a difference between the desired and actual behavior. The minimal severity has desired behavior, that's why nil is returned.

logging/logging_test.go
```diff
@@ -208,6 +208,89 @@ func TestLogSync(t *testing.T) {
...
+			mustBeLogged: true,
+		},
+		{
+			severity:     logging.Alert,
+			mustBeLogged: true,
+		},
+		{
+			severity:     logging.Emergency,
+			mustBeLogged: true,
+		},
```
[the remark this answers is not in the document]
kimmelserj: Got it. I will rewrite the tests to check edge cases.

logging/logging_test.go
kimmelserj: Ok. I will try to do something with it.

logging/loggeroption.go
kimmelserj: Done

logging/logging_test.go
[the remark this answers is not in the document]
kimmelserj: Done

logging/logging_test.go
kimmelserj: Done

kimmelserj:
@daniel-sanche could you take a look on the PR? I have already done all the suggested changes.

kimmelserj:
@daniel-sanche 🙏

daniel-sanche:
I definitely agree that this is an issue that should be addressed, thanks for putting this solution together!

That said, I want to be careful to make sure that this is the right solution before merging it. I'd like to discuss this with some of the go experts within GCP libraries to make sure I'm not missing any other solutions that should be considered. Since we're getting near the release freeze for the holidays, these discussions will likely have to wait for next quarter.

Sorry for the late response. Just wanted to give a quick update that this is on our radar

logging/logging.go
```diff
@@ -254,6 +254,7 @@ type Logger struct {
 	populateSourceLocation int
 	partialSuccess         bool
 	redirectOutputWriter   io.Writer
+	minimalLoggedSeverity  Severity
```
daniel-sanche, line 257: Is this a breaking change? Could this have any affects on existing users?

logging/logging.go
```diff
@@ -274,6 +275,7 @@ func (c *Client) Logger(logID string, opts ...LoggerOption) *Logger {
 		populateSourceLocation: DoNotPopulateSourceLocation,
 		partialSuccess:         false,
 		redirectOutputWriter:   nil,
+		minimalLoggedSeverity:  Default,
```
daniel-sanche, line 278: This should be mentioned with the [other LoggerOptions in the README](https://github.com/googleapis/google-cloud-go/blob/56f4cdd066cc9eaeece2c6fb466d58c3e7c41563/logging/README.md#logger-configuration-options)

daniel-sanche (changes requested):
Sorry about the wait. I left a couple comments, but overall this looks like a great addition.

If you don't have the bandwidth to finish this up anymore, I can take it on too

Thanks!

logging/logging.go
```diff
@@ -630,6 +632,9 @@ func jsonValueToStructValue(v interface{}) *structpb.Value {
 // and will block, it is intended primarily for debugging or critical errors.
 // Prefer Log for most uses.
 func (l *Logger) LogSync(ctx context.Context, e Entry) error {
+	if e.Severity < l.minimalLoggedSeverity {
```
losalex, line 635: Not sure if this is valid concern, but since `Default` the lowest severity, we should also introduce `None` to be the lowest severity, thus anyone could also log with `Default` severity which correct me if I wrong a valid cloud Logging severity.

logging/logging.go
kimmelserj: No. Severity zero value is zero = LogSeverity_DEFAULT. That's why [this if](https://github.com/googleapis/google-cloud-go/pull/6346/files#diff-6bb8ce03ad8192446d441a3d85bf0fb8307e3d7829bc027785b35ce95868da70R664) will be skipped if minimalLoggedSeverity is not set. It's backward compatible change.

logging/logging.go
kimmelserj: Sure. I will add this to README.

logging/logging.go
kimmelserj: I got you. The `Default` severity level is a very strange thing. [Here](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logseverity), the `Default` severity is described as "The log entry has no assigned severity level." We cannot recognize the severity level for "default" logs. Apparently, "default" logs will always have to be logged, regardless of the minimum severity level. I will prepare the changes in the code. Thank you for insight.)

logging/logging.go
kimmelserj: Done.

logging/logging.go
kimmelserj: Done.

losalex: Thanks @kimmelserj! 
@daniel-sanche, thinking about it twice I wonder if we could add Severity.None (-1) same way as we did in Java. This way we could make a logic to be straightforward - the only addition would be to make sure that this severity cannot be used in log entries

logging/logging.go
daniel-sanche, line 278: nit: A comment explaining that default means no filtering would be helpful

logging/logging.go
```diff
@@ -632,7 +632,7 @@ func jsonValueToStructValue(v interface{}) *structpb.Value {
 // and will block, it is intended primarily for debugging or critical errors.
 // Prefer Log for most uses.
 func (l *Logger) LogSync(ctx context.Context, e Entry) error {
-	if e.Severity < l.minimalLoggedSeverity {
+	if e.Severity != Default && e.Severity < l.minimalLoggedSeverity {
```
daniel-sanche, line 635: nit: a comment here would also be useful. Something like `Ignore logs with severity less than minimalLoggedSeverity. Default severity logs are always written`

logging/logging.go
daniel-sanche: > Not sure if this is valid concern, but since Default the lowest severity, we should also introduce All to be the "include-all" severity, thus anyone could also log with Default severity which correct me if I wrong a valid cloud Logging severity.

My comment is out of date now since I'm late, but I believe the original code with `minimalLoggedSeverity` was already the "include-all" option, since it was filtering logs out using `<` and not `<=`

> Apparently, "default" logs will always have to be logged, regardless of the minimum severity level. I will prepare the changes in the code. Thank you for insight.)

Hmm I'm not sure if I agree with this behaviour. I'd probably say we should treat Default as as a filterable level below debugging with severity 0 like [in the spec](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logseverity). But there's definitely an argument to this change too, so I'm fine either way

> @daniel-sanche, thinking about it twice I wonder if we could add Severity.None (-1) same way as we did in Java. This way we could make a logic to be straightforward - the only addition would be to make sure that this severity cannot be used in log entries

Yeah, it may be something worth considering following Java for the next step if people ask for it, but I'd vote we should keep this PR simple for now and merge this without it.

logging/logging.go
```diff
@@ -661,7 +661,7 @@ func (l *Logger) LogSync(ctx context.Context, e Entry) error {
 
 // Log buffers the Entry for output to the logging service. It never blocks.
 func (l *Logger) Log(e Entry) {
-	if e.Severity < l.minimalLoggedSeverity {
+	if e.Severity != Default && e.Severity < l.minimalLoggedSeverity {
```
daniel-sanche, line 664: nit: same comment here too

daniel-sanche (approved):
LGTM with small comments, but we should wait for Alex's feedback before merging

logging/logging.go
losalex: > > Not sure if this is valid concern, but since Default the lowest severity, we should also introduce All to be the "include-all" severity, thus anyone could also log with Default severity which correct me if I wrong a valid cloud Logging severity.
> 
> My comment is out of date now since I'm late, but I believe the original code with `minimalLoggedSeverity` was already the "include-all" option, since it was filtering logs out using `<` and not `<=`
> 
> > Apparently, "default" logs will always have to be logged, regardless of the minimum severity level. I will prepare the changes in the code. Thank you for insight.)
> 
> Hmm I'm not sure if I agree with this behaviour. I'd probably say we should treat Default as as a filterable level below debugging with severity 0 like [in the spec](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logseverity). But there's definitely an argument to this change too, so I'm fine either way
> 
> > @daniel-sanche, thinking about it twice I wonder if we could add Severity.None (-1) same way as we did in Java. This way we could make a logic to be straightforward - the only addition would be to make sure that this severity cannot be used in log entries
> 
> Yeah, it may be something worth considering following Java for the next step if people ask for it, but I'd vote we should keep this PR simple for now and merge this without it.

@kimmelserj , seems adding Severity.None is not a big of a fix - can you please let us know if you can make it?

logging/logging_test.go
```diff
@@ -209,6 +209,55 @@ func TestLogSync(t *testing.T) {
 	}
 }
 
+func TestLogSyncWithMinimalLoggedSeverity(t *testing.T) {
```
minherz, line 212: nit: consider refactoring instead of creating just another copy of the same code.

logging/logging_test.go
```diff
@@ -243,6 +292,50 @@ func TestLogAndEntries(t *testing.T) {
 	}
 }
 
+func TestLogWithMinimalLoggedSeverity(t *testing.T) {
```
minherz, line 295: nit: same comment. the current implementation increases technical debt for tests instead of decreasing it.

codyoss:
Closing this PR as it is outdated. Feel free to reopen if you still want to address the open feedback.

PR #6350, opened 2022-07-18
chore(storage): remove adapters dependency

tritone:
The go-type-adapters package is used only minimally in storage,
and nowhere else in google-cloud-go. Seems easiest to just drop
this dependency.


codyoss (approved):

tritone:
> IDK why I didn't use this in the first place: https://pkg.go.dev/google.golang.org/protobuf@v1.28.0/types/known/timestamppb
> 
> Do you want to switch your PR to use that? Or keep it as is? doesn't matter to me.

Maybe let's just leave it like this for now for consistency? We can always make that update later.

PR #6370, opened 2022-07-19
feat(storage): export ShouldRetry

### LINKED ISSUE #6362: storage: export default shouldRetry function
I've recently stumbled across similar errors to https://github.com/googleapis/google-cloud-go/issues/3735 and applied the proposed fix there - created a new `ShouldRetry` function for the storage client. However, it seems very odd that I have to copy the code from the default retry function - which also means that any additions/fixes to it will not be reflected in my code automatically. I think it would be nice to have that function exported from the `storage` package, so it can be used like that:
```
gcsClient.SetRetry(storage.WithErrorFunc(func(err error) bool {
	if storage.DefaultShouldRetry(err) { // this function comes from the storage package
		return true
	}
	return customShouldRetry(err) // my custom retry logic
}))
```

tritone:
Export the default func to determine whether an error is retryable.
This makes it easier for users to use the WithErrorFunc option
without copying a lot of code.

Fixes #6362


storage/grpc_client.go
```diff
@@ -1229,7 +1229,7 @@ func (r *gRPCReader) Close() error {
 // an attempt to reopen the stream.
 func (r *gRPCReader) recv() (*storagepb.ReadObjectResponse, error) {
 	msg, err := r.stream.Recv()
-	if err != nil && shouldRetry(err) {
+	if err != nil && ShouldRetry(err) {
```
tritone, line 1232: Note: ShouldRetry should not be referenced directly here; instead we should be using the configured value from the retry configuration passed through the interface. Will file a separate bug to fix this up. FYI @noahdietz

noahdietz (approved):

PR #6375, opened 2022-07-19
test(profiler): compile busybench before running backoff test

amchiclet:
The backoff test is timing out because all the 45 processes that are spawned use the command `go run`, which compiles busybench to a temp directory and run it. This means 45 processes compile the same code and have their own copy of the compiled code in their own temp directory. On the VM, this requires a lot of resources and causes delays (~10 minutes) when launching busybench, causing it to timeout.

This PR compiles busybench with `go build` first so all the processes can use the single binary. As a result the tests finish within 1 hour.

Also cleaned up the script to install go.

Tested on my GCP project and the backoff test finishes within the time limit.


wyk9787 (approved):
LGTM pending CI results.

PR #6382, opened 2022-07-20
chore(internal/gapicgen): update spanner db admin

noahdietz:
Target new, more up-to-date service config for use in generation of spanner database admin.


codyoss (approved):

codyoss (approved):
Thanks!

PR #6387, opened 2022-07-21
feat(pubsub): complete AckResult for exactly once

hongalex:
Pipe the output of Ack/Modack operations to AckResult that are returned when users call Ack/NackWithResult. This involves removing a callback and makes `sendAck` and `sendModAck` have repeated code. However, removing the callback also makes it more readable. In addition, we don't infinitely retry these methods until they fail anymore, since that will be handled by a later PR.

To support this, this PR also includes
1. Switching `processResults` back to using a map[string]string for ackID->error strings (since that's what apiError metadata field returns)
2. Changing the output of `processResults` and `pending` maps to use a new struct that contains both ackIDs and AckResult (for retrying later on)
3. Capping the max number of ackIDs per ack/modack to 2500 (which is in line with Python) instead of calculating it on the fly

### DIFF
--- a/pubsub/iterator.go
+++ b/pubsub/iterator.go
@@ -57,6 +57,19 @@ var (
 	minDurationPerLeaseExtensionExactlyOnce = 1 * time.Minute
 )
 
+// ackResultWithID stores an ackResult with its corresponding ackID
+type ackResultWithID struct {
+	r     *AckResult
+	ackID string
+}
+
+func newAckResultWithID(r *AckResult, ackID string) *ackResultWithID {
+	return &ackResultWithID{
+		r:     r,
+		ackID: ackID,
+	}
+}
+
 type messageIterator struct {
 	ctx        context.Context
 	cancel     func() // the function that will cancel ctx; called in stop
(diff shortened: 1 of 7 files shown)


pubsub/iterator.go
```diff
@@ -427,127 +437,111 @@ func (it *messageIterator) handleKeepAlives() {
...
+			apiErr, ok := apierror.FromError(err)
+			if ok {
+				md = apiErr.Metadata()
+				st = apiErr.GRPCStatus()
 			}
+			resultsByAckID := make(map[string]*AckResult)
+			for _, ackID := range toSend {
+				resultsByAckID[ackID] = m[ackID]
+			}
+			// TODO(hongalex): retry the ackIDs with transient retriable errors.
```
feywind, line 481: I'm assuming this is for the mentioned next PR?

pubsub/iterator.go
```diff
@@ -427,127 +437,111 @@ func (it *messageIterator) handleKeepAlives() {
...
+		// want to cancel this RPC when the iterator is stopped.
+		cctx, cancel2 := context.WithTimeout(context.Background(), 60*time.Second)
+		defer cancel2()
+		err := it.subc.ModifyAckDeadline(cctx, &pb.ModifyAckDeadlineRequest{
+			Subscription:       it.subName,
+			AckDeadlineSeconds: deadlineSec,
+			AckIds:             toSend,
+		})
+		it.eoMu.RLock()
+		exactlyOnceDelivery := it.enableExactlyOnceDelivery
```
feywind, line 517: This actually brings up an interesting point for me - if the EO semantics are enabled, and we receive messages and they're queued up for acks, and then EO is disabled... do we do the existing messages using the EO semantics, or switch immediately?

hongalex: Perhaps @maheshgattani can comment on this. I don't think switching exactly once delivery on/off is a common use case, so either behavior should be ok here. It's significantly easier to keep track of this through the actual on/off status of exactly once (through StreamingPullResponse) rather than keeping track of exactly once for every message.

pubsub/iterator.go
hongalex: Yeah it is, I just left it as a todo for now.

gnagel:
For my own curiosity, what's left in order to support exactly once delivery for pubsub messages with the go client?

hongalex:
@gnagel This PR needs to be merged, followed by retrying of temporary failures. After that we can start merging everything.

gnagel:
> @gnagel This PR needs to be merged, followed by retrying of temporary failures. After that we can start merging everything.

Fantastic, thank you for working on this.

I'm very excited to see this rollout and use it in production 😎

pradn (approved):

PR #6403, opened 2022-07-22
feat(artifactregistry): start generating apiv1

quartzmo:


internal/gapicgen/generator/config.go
```diff
@@ -1624,6 +1624,15 @@ var MicrogenGapicConfigs = []*MicrogenConfig{
...
 		ReleaseLevel: "beta",
 	},
+	{
+		InputDirectoryPath:    "google/devtools/artifactregistry/v1",
+		Pkg:                   "artifactregistry",
+		ImportPath:            "cloud.google.com/go/artifactregistry/apiv1",
+		GRPCServiceConfigPath: "artifactregistry_grpc_service_config.json",
+		ApiServiceConfigPath:  "artifactregistry_v1.yaml",
+		// GA after 2022/08/22
+		ReleaseLevel: "beta",
```
quartzmo, line 1634: @codyoss Do we still put the beta label on this package, even though the module is already 1.x and the service is GA?

codyoss: Yes, because this changes some of the generated docs at a package level. Looks like this module may have been marked GA too soon. But our package docs state for the other client it is really a beta.

codyoss (approved):

PR #6412, opened 2022-07-25
ci: use header-checker for license headers

enocom:


enocom:
Configuration details are here: https://github.com/googleapis/repo-automation-bots/tree/main/packages/header-checker-lint.

The test deleted in this commit ignores `.pb.go` files, but since the headers should be identical, I've left that change off.

codyoss (approved):
Thanks you for this!

PR #6414, opened 2022-07-25
chore(storage): integrate Compose in new interface

tritone:
Change production code paths for Composer to use the new transport
agnostic interface.

Required a few setup steps to make tests pass, which will be needed for
other integrations too:
* Add initialization of tc to NewClient
* Add new useGRPC flag to remove switches on tc in Reader/Writer

Also a few cleanups:
* Use EncryptionKey from dstObject
* Use defaultGen instead of hard coded int

Also, a unit test checks that setting Generation with composer causes
a failure. Currently this is enforced by applyConds. I added
a separate validation for this instead, but I could do a pass-
through of the Generation through the interface -- it's just
confusing looking because setting Generation on a new object
is never valid.


storage/copy.go
```diff
@@ -204,45 +207,39 @@ func (c *Composer) Run(ctx context.Context) (attrs *ObjectAttrs, err error) {
...
+			conds: src.conds,
 		}
-		req.SourceObjects = append(req.SourceObjects, srcObj)
+		req.srcs = append(req.srcs, s)
 	}
 
-	call := c.dst.c.raw.Objects.Compose(c.dst.bucket, c.dst.object, req).Context(ctx)
-	if err := applyConds("ComposeFrom destination", c.dst.gen, c.dst.conds, call); err != nil {
-		return nil, err
+	// TODO: factor this out to a function?
```
noahdietz, line 232: Is the definition of idempotency _here_ a generalized one? Or is it specific to Copy/Compose?

noahdietz (approved):
LGTM, thanks for doing that bit of leg work to get ancillary things set up. Looking into why presubmits didn't run.

noahdietz:
> Looking into why presubmits didn't run.

I reran kokoro and it actually executed this time. Just need to fix a [lint error](https://source.cloud.google.com/results/invocations/59049204-a019-4283-b616-ae71b4602a62/targets/cloud-devrel%2Fclient-libraries%2Fgo%2Fgoogle-cloud-go%2Fpresubmit%2Flatest-version/log) `gofmt -w storage/copy.go`.

storage/copy.go
tritone: It's specific to copy/compose. However was thinking I could add a helper func to create the storage opts just to eliminate a few of these lines of boilerplate which will be in every func:

```
func makeStorageOpts(idempotent bool, retry *retryConfig, userProject string) []storageOption {...}
```

Maybe this defeats the point of the options interface though?

noahdietz: > Maybe this defeats the point of the options interface though?

Perhaps...let's just keep the TODO and evaluate as we go?

tritone: I'll be adding this in a separate PR.

PR #6426, opened 2022-07-26
chore(storage): add makeStorageOpts helper

tritone:
This takes out a few lines of boilerplate that will be
present in all top-level methods.


cojenco (approved):
Thanks Chris for adding this!

noahdietz (approved):
Tight!

PR #6430, opened 2022-07-26
chore(storage): migrate HMAC operations

noahdietz:
Migrate HMAC operations to transport-agnostic interface.

Integration tests pass locally for HMAC operations.


tritone (approved):
Looks great!

PR #6433, opened 2022-07-26
chore(storage): migrate top-level methods

cojenco:
Migrate few top-level methods to transport-agnostic interface.
- GetServiceAccount
- ListBuckets
- CreateBucket*

*Interface refactoring
Shown in the [current implementation](https://github.com/googleapis/google-cloud-go/blob/main/storage/bucket.go#L91), bucketName is referenced from `BucketHandle`, so proposing to include `bucket string` as an arg for method `CreateBucket()`

Integration tests pass locally

### DIFF
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -2125,7 +2125,6 @@ type BucketIterator struct {
 	buckets   []*BucketAttrs
 	pageInfo  *iterator.PageInfo
 	nextFunc  func() error
-	retry     *retryConfig
 }
 
 // Next returns the next result. Its second return value is iterator.Done if
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -2125,6 +2125,7 @@ type BucketIterator struct {
 	buckets   []*BucketAttrs
 	pageInfo  *iterator.PageInfo
 	nextFunc  func() error
+	retry     *retryConfig
 }
 
 // Next returns the next result. Its second return value is iterator.Done if
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -82,27 +82,11 @@ func (b *BucketHandle) Create(ctx context.Context, projectID string, attrs *Buck
 	ctx = trace.StartSpan(ctx, "cloud.google.com/go/storage.Bucket.Create")
 	defer func() { trace.EndSpan(ctx, err) }()
 
-	var bkt *raw.Bucket
-	if attrs != nil {
-		bkt = attrs.toRawBucket()
-	} else {
-		bkt = &raw.Bucket{}
+	o := makeStorageOpts(true, b.retry, b.userProject)
+	if _, err := b.c.tc.CreateBucket(ctx, projectID, b.name, attrs, o...); err != nil {
+		return err
 	}
-	bkt.Name = b.name
-	// If there is lifecycle information but no location, explicitly set
-	// the location. This is a GCS quirk/bug.
-	if bkt.Location == "" && bkt.Lifecycle != nil {
-		bkt.Location = "US"
-	}
-	req := b.c.raw.Buckets.Insert(projectID, bkt)
-	setClientHeader(req.Header())
-	if attrs != nil && attrs.PredefinedACL != "" {
-		req.PredefinedAcl(attrs.PredefinedACL)
-	}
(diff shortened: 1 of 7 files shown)


storage/bucket.go
```diff
@@ -2124,17 +2108,8 @@ func (it *ObjectIterator) fetch(pageSize int, pageToken string) (string, error)
...
-		client:    c,
-		projectID: projectID,
-	}
-	it.pageInfo, it.nextFunc = iterator.NewPageInfo(
-		it.fetch,
-		func() int { return len(it.buckets) },
-		func() interface{} { b := it.buckets; it.buckets = nil; return b })
-
-	return it
+	o := makeStorageOpts(true, c.retry, "")
```
noahdietz, line 2111: Why `""` for userProject? Is that not in scope for this RPC?

storage/storage.go
```diff
@@ -2106,17 +2106,9 @@ func toProtoCommonObjectRequestParams(key []byte) *storagepb.CommonObjectRequest
...
-	var err error
-	err = run(ctx, func() error {
-		res, err = r.Context(ctx).Do()
-		return err
-	}, c.retry, true, setRetryHeaderHTTP(r))
-	if err != nil {
-		return "", err
-	}
-	return res.EmailAddress, nil
+	o := makeStorageOpts(true, c.retry, "")
```
noahdietz, line 2109: Same question here WRT `""` for userProject.

storage/bucket.go
tritone: UserProject is a method on BucketHandle, so there's no way for library users to actually set UserProject on these top level calls: https://pkg.go.dev/cloud.google.com/go/storage#BucketHandle.UserProject

Maybe we should add a method on client as well but that's outside the scope here.

storage/client.go
```diff
@@ -44,7 +44,7 @@ type storageClient interface {
 	// Top-level methods.
 
 	GetServiceAccount(ctx context.Context, project string, opts ...storageOption) (string, error)
-	CreateBucket(ctx context.Context, project string, attrs *BucketAttrs, opts ...storageOption) (*BucketAttrs, error)
+	CreateBucket(ctx context.Context, project, bucket string, attrs *BucketAttrs, opts ...storageOption) (*BucketAttrs, error)
```
tritone, line 47: Nice catch on this

tritone (approved):

cojenco:
@tritone @noahdietz Any thoughts on adding `retryConfig` to `BucketIterator`... this is what I've done for now

Moving to the new interface detaches referencing `Client` (and client.retry) in `BucketIterator`, [like so](https://github.com/googleapis/google-cloud-go/blob/main/storage/bucket.go#L2129)

This stems from a [failed test](https://github.com/googleapis/google-cloud-go/blob/main/storage/storage_test.go#L1358-L1360) unable to find retryConfig from Client

storage/bucket.go
noahdietz: Makes sense. Thanks!

noahdietz:
> @tritone @noahdietz Any thoughts on adding `retryConfig` to `BucketIterator`... this is what I've done for now

I think that is fine. Did you have a particular concern? I did notice though that the actual API calls [still reference](https://github.com/googleapis/google-cloud-go/pull/6433/commits/b7662b41451bc90fd122738c5e7dacafa0d0a0be#diff-eec5db1002a6674758e3a557d89495986c8450de01512d561e2a7962584c08f2L194) `s.retry` - should that be `it.retry` instead? Can't comment on something not in the scope of the diff I guess :)

> Moving to the new interface detaches referencing Client (and client.retry) in BucketIterator

Are we able to/should we remove the BucketIterator `client *Client` field as a result? Haven't looked to see if it is referenced anywhere else.

cojenco:
> Are we able to/should we remove the BucketIterator `client *Client` field as a result? Haven't looked to see if it is referenced anywhere else.

Great callout, Noah,  `client *Client` field should be removed from BucketIterator. That helped me discover a todo to remove the no-longer used `fetch` method, which is the other only place referencing client field :)

Also spoke to Chris and realized my original question is more of a test issue. The call is retrieving retryConfig with no concerns, so removing the now-outdated test.

noahdietz (approved):

PR #6435, opened 2022-07-26
chore(storage): migrate Copier

tritone:
Migrate Copier to use the new transport-agnostic interface.
I found one additional piece of data that needs to be passed back
through the interface in the process (object size).
    
I removed one integration test. This test is old and currently
calls rewriteObject in a way that an external library user could
not accomplish. I don't think this test adds any value.

### DIFF
--- a/storage/copy.go
+++ b/storage/copy.go
@@ -86,21 +86,53 @@ func (c *Copier) Run(ctx context.Context) (attrs *ObjectAttrs, err error) {
 	if c.DestinationKMSKeyName != "" && c.dst.encryptionKey != nil {
 		return nil, errors.New("storage: cannot use DestinationKMSKeyName with a customer-supplied encryption key")
 	}
+	if c.dst.gen != defaultGen {
+		return nil, fmt.Errorf("storage: generation cannot be specified on copy destination, got %v", c.dst.gen)
+	}
 	// Convert destination attributes to raw form, omitting the bucket.
 	// If the bucket is included but name or content-type aren't, the service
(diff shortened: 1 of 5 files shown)


storage/copy.go
```diff
@@ -86,69 +85,57 @@ func (c *Copier) Run(ctx context.Context) (attrs *ObjectAttrs, err error) {
...
+		token:         c.RewriteToken,
+	}
+
+	isIdempotent := c.dst.conds != nil && (c.dst.conds.GenerationMatch != 0 || c.dst.conds.DoesNotExist)
+	var userProject string
+	if c.dst.userProject != "" {
+		userProject = c.dst.userProject
+	} else if c.src.userProject != "" {
+		userProject = c.src.userProject
+	}
```
noahdietz, line 121: Nit, matter of preference, not important: Could just default to setting one and just check the other

```
userProject := c.dst.userProject
if c.src.UserProject != "" {
  userProject = c.src.userProject
}
```

noahdietz (approved):

storage/copy.go
tritone: I thought about this, but: on one hand, with the existing logic, the `dst` one takes precedence. On the other, it reads a bit strangely to use src by default because it's actually very unlikely for someone to set it via src because of how the options are normally chained (would be kind of a strange construction). Soooo I figured i'd just leave it looking like this.

noahdietz: Ok that makes total sense! Thanks.

cojenco (approved):

PR #6438, opened 2022-07-27
chore(storage): migrate bucket methods

cojenco:
Migrate Bucket operations to transport-agnostic interface. Integration tests pass locally
- DeleteBucket
- GetBucket
- UpdateBucket
- LockBucketRetentionPolicy
- ListObjects

Part of the migration
- removed Bucket helpers `new{Get/Delete/Patch}Call`
- similar to ListBuckets, removed internal fetch method that is no longer used by the iterator

### DIFF
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -20,7 +20,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"net/http"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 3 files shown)


tritone (approved):

cojenco:
@tritone thanks for the review. 

Wanted to confirm with you that [this note](https://github.com/googleapis/google-cloud-go/blob/main/storage/bucket.go#L1363-L1365) is out of date. Looking at our external docs, this should be available to the general public? If so, I'll remove this part of the doc.

tritone:
> @tritone thanks for the review.
> 
> Wanted to confirm with you that [this note](https://github.com/googleapis/google-cloud-go/blob/main/storage/bucket.go#L1363-L1365) is out of date. Looking at our external docs, this should be available to the general public? If so, I'll remove this part of the doc.

Yup that's correct, feel free to remove. We should go through at some point and audit these, I think there are some other ones we can get rid of also (e.g. HMAC)

PR #6440, opened 2022-07-27
chore(storage): migrate ACL operations

tritone:
Migrate ACL ops to the new transport agnostic interface. No
interface changes were necessary for this one and tests pass.


noahdietz (approved):

noahdietz:
From golint:
> storage/acl.go:156:9: if block ends with a return statement, so drop this else and outdent its block

Wish we had "please fix" style reporting. Using similar tools in GitHub Actions will enable such PR interaction.

Issue #6439, opened 2022-07-27
presubmit masks build failures


codyoss:
We recently had some build failures that should have been caught by our CI. I believe they were not because we transform our tests logs into junit format that just passes test output and likely ignores the messages about the build failing. We need to make sure we capture build failures in the presubmit. Could be as simple as adding a tests to run go build for every module with changes.

Related: https://github.com/googleapis/google-cloud-go/issues/6419 https://github.com/googleapis/google-cloud-go/pull/6436

PR #6441, opened 2022-07-28
test: start building all code

codyoss:
Code is built when testing it, but some type of build failures are
not properly translated by the junit converter we use. Adding this
extra build will catch this previously missed failures and should
affect test time minimally as builds will have been cached.

Fixes: #6439

### DIFF
--- a/internal/kokoro/presubmit.sh
+++ b/internal/kokoro/presubmit.sh
@@ -67,6 +67,11 @@ runPresubmitTests() {
     | go-junit-report -set-exit-code > sponge_log.xml
   # Add the exit codes together so we exit non-zero if any module fails.
   exit_code=$(($exit_code + $?))
+  # There are some types of build failures that go-junit-report does not
+  # properly handle. We can be safe here and run go build as well to catch such
+  # failures.
+  go build ./...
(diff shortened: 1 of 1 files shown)


quartzmo (approved):

quartzmo:
Kokoro:

```
+ cat sponge_log.log
+ go-junit-report -set-exit-code
+ exit_code=2
+ go build ./...
/tmpfs/src/google-cloud-go
+ exit_code=2
```

codyoss:
Haha OOMed the machine for trying to build the snippet directory. I will add a check for skipping the building of that one

noahdietz:
> 

Would it make sense to add a build tag to snippet files?

codyoss:
@noahdietz My only fear there is that the code is less copy pastable. I think tags are pretty advanced and users that drill into the sample and copy the whole file may be confused why `go run/go bulid` is not just working without removing the magic comment

noahdietz:
> @noahdietz My only fear there is that the code is less copy pastable. I think tags are pretty advanced and users that drill into the sample and copy the whole file may be confused why `go run/go bulid` is not just working without removing the magic comment

That's fair! No need to then.

PR #6442, opened 2022-07-28
chore(storage): migrate Notification methods

noahdietz:
Migrates the Notification APIs to the transport agnostic surface. Integration tests passed and added a check to ensure that AddNotification returns a Notification with a non-empty ID.

### DIFF
--- a/storage/notifications.go
+++ b/storage/notifications.go
@@ -157,21 +157,10 @@ func (b *BucketHandle) AddNotification(ctx context.Context, n *Notification) (re
 	if n.TopicID == "" {
 		return nil, errors.New("storage: AddNotification: missing TopicID")
 	}
-	call := b.c.raw.Notifications.Insert(b.name, toRawNotification(n))
-	setClientHeader(call.Header())
-	if b.userProject != "" {
-		call.UserProject(b.userProject)
-	}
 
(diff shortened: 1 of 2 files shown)


storage/notifications.go
```diff
@@ -217,12 +195,6 @@ func (b *BucketHandle) DeleteNotification(ctx context.Context, id string) (err e
...
 
-	call := b.c.raw.Notifications.Delete(b.name, id)
-	setClientHeader(call.Header())
-	if b.userProject != "" {
-		call.UserProject(b.userProject)
-	}
-	return run(ctx, func() error {
-		return call.Context(ctx).Do()
-	}, b.retry, true, setRetryHeaderHTTP(call))
+	opts := makeStorageOpts(true, b.retry, b.userProject)
```
tritone, line 198: Hmmmm so the retry conformance tests are flagging Delete; however I can't see any difference in retry behavior either here or in the http_client.go implementation. Any idea what's going on?

noahdietz: I just saw that too and was going to ask the same question haha

noahdietz: Fixed.

tritone (approved):

PR #6447, opened 2022-07-29
feat(godocfx): add friendlyAPIName

### LINKED ISSUE #5949: feat(godocfx): add friendly api name to yaml
```doc-templates``` will display the friendly api name in the ```<h1>```([pr](https://github.com/googleapis/doc-templates/pull/367)) and first paragraph summary ([pr](https://github.com/googleapis/doc-templates/pull/373)) of generated pages. ```godocfx``` should include a ```friendlyApiName``` as an item key in the yaml which describes the simplified name of the Go library along with its API version.  For example, the Cloud Vision apiv1p1beta1 library should set the friendly api name to be something like "Cloud Vision V1p1beta1 API"

This change will help discern documentation across API versions.

### PR DESCRIPTION
**Pull request by tbpg**
cc @dansaadati 

Fixes #5949.

### CODE REVIEW

File: internal/godocfx/testdata/golden/index.yml [hunk 1]
```diff
@@ -288,6 +288,7 @@ items:
   - cloud.google.com/go/storage.Writer.Write
   - cloud.google.com/go/storage.SignedURL
   alt_link: https://pkg.go.dev/cloud.google.com/go/storage
+  friendlyApiName: Storage API
```
dansaadati, line 291: This looks good! This should also be generated for child pages of the library as well.

tbpg: Which pages are you referring to? For Go, there is only one page per package.

dansaadati: Actually, on closer look of Go API reference pages, it looks like there's one page per library 🤯. I don't think this applies to Go then.

codyoss (approved):
LGTM

PR #6452, opened 2022-08-01
chore(storage): migrate Writer

tritone:
Migrate Writer to use the transport-agnostic interface. A few
minor fixes:

* Pass in retry, idempotency and user project via settings rather
than determining below interface.
* Use default generation since this can't be sent for this call.


noahdietz (approved):
Wow a remarkably clean PR! That went smoother than I thought tbh considering complexities of Writer, great stuff Chris.

PR #6457, opened 2022-08-02
chore(storage): migrate Reader

tritone:
Migrate the Reader to use the new transport-agnostic interface.
A few minor fixes were required:
* Move validations above interface
* Add settings for call and fix up userProject for gRPC
* Add param for sending with accept-encoding: gzip (HTTP only)
* Fix up reader resumption test


noahdietz (approved):
LGTM just need a gofmt on storage/client.go per kokoro.

PR #6464, opened 2022-08-02
feat(bigquery): add preview support for default values

shollyman:


Neenu1995 (approved):

bigquery/table_integration_test.go
```diff
@@ -78,7 +78,54 @@ func TestIntegration_TableCreateWithConstraints(t *testing.T) {
...
+	if client == nil {
+		t.Skip("Integration tests skipped")
+	}
+	ctx := context.Background()
+	table := dataset.Table("defaultvalues")
+	schema := Schema{
+		{Name: "str_col", Type: StringFieldType, DefaultValueExpression: "'FOO'"},
+		{Name: "timestamp_col", Type: TimestampFieldType, DefaultValueExpression: "CURRENT_TIMESTAMP()"},
+	}
+	err := table.Create(ctx, &TableMetadata{
```
alvarowolfx, line 93: If this runs twice before the table expires, the test is going to fail here right ? Maybe we need to clean up the table after the test finishes.

shollyman: Cleanups are actually done by removing the enclosing dataset and its contents.  Each run gets a unique dataset.

init stuff can be found over here: https://github.com/googleapis/google-cloud-go/blob/main/bigquery/integration_test.go#L226

alvarowolfx (approved):
LGTM

PR #6465, opened 2022-08-02
feat(storage): refactor to use transport-agnostic interface

noahdietz:


internal/kokoro/continuous.sh
```diff
@@ -56,7 +56,8 @@ set -eo pipefail
 set -x
 
 # cd to project dir on Kokoro instance
-cd github/google-cloud-go
+# TODO(#6444): change this back to github/google-cloud-go before merging to main
```
noahdietz, line 59: Need to remove this.

tritone (approved):

cojenco (approved):
🎉

internal/kokoro/continuous.sh
noahdietz: Done in [9757864](https://github.com/googleapis/google-cloud-go/pull/6465/commits/9757864c9a0704e5f33257cefef95ce06de6f10f)

PR #6480, opened 2022-08-05
chore(bigquery/storage/managedwriter): add client instrumentation

shollyman:
adds instrumentation for this client to aid in debugging scenarios with
the write backend


bigquery/storage/managedwriter/options.go
```diff
@@ -70,7 +70,7 @@ func WithMaxInflightBytes(n int) WriterOption {
 // This is generally for diagnostic purposes only.
 func WithTraceID(traceID string) WriterOption {
 	return func(ms *ManagedStream) {
-		ms.streamSettings.TraceID = traceID
+		ms.streamSettings.TraceID = buildTraceID(traceID)
```
yirutang, line 73: Do you want to allow users to add in their own trace id?

shollyman: This changes from just setting the traceID in total to building a prefix for the library, and the user-specific trace ID.  Same as jsonwriter etc.

shollyman: take a look at options_test.go for the expected output

codyoss (approved):

PR #6485, opened 2022-08-08
feat(pubsub): retry temporary failures for ack/modacks

hongalex:
This PR adds the logic for retrying temporary ack/modack (including nack) failures for exactly once delivery.

Some other changes captured in this PR
1. Updating `subscription_test.go` and `integration_test.go` to support exactly once delivery for publish/subscribing.
   (Git diff is not great here, but a lot of changes just  involves moving a lot of the publish/receive test code into a sub-test function so they can be invoked in parallel. Not much logic change.)
3. Reverted all `ackResultWithID` to just be `AckResult`. Passing around the map of ackID->AckResult works best and we just needed to be consistent.

### DIFF
--- a/pubsub/iterator.go
+++ b/pubsub/iterator.go
@@ -447,6 +447,8 @@ func (it *messageIterator) handleKeepAlives() {
 	it.checkDrained()
 }
 
+// sendAck is used to confirm acknowledgement of a message. If exactly once delivery is
+// enabled, we'll retry these messages for a short duration in a goroutine.
 func (it *messageIterator) sendAck(m map[string]*AckResult) {
 	ackIDs := make([]string, 0, len(m))
 	for k := range m {
@@ -486,10 +488,12 @@ func (it *messageIterator) sendAck(m map[string]*AckResult) {
 	}
 }
 
+// sendModAck is used to extend the lease of messages or nack them.
 // The receipt mod-ack amount is derived from a percentile distribution based
 // on the time it takes to process messages. The percentile chosen is the 99%th
 // percentile in order to capture the highest amount of time necessary without
-// considering 1% outliers.
+// considering 1% outliers. If the ModAck RPC fails and exactly once delivery is
+// enabled, we retry it in a separate goroutine for a short duration.
 func (it *messageIterator) sendModAck(m map[string]*AckResult, deadline time.Duration) {
 	deadlineSec := int32(deadline / time.Second)
 	ackIDs := make([]string, 0, len(m))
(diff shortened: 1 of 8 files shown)


pubsub/iterator.go
```diff
@@ -479,19 +470,18 @@ func (it *messageIterator) sendAck(m map[string]*ackResultWithID) {
...
 			for _, ackID := range toSend {
 				resultsByAckID[ackID] = m[ackID]
 			}
-			// TODO(hongalex): retry the ackIDs with transient retriable errors.
-			_, _ = processResults(st, resultsByAckID, md)
+			st, md := extractMetadata(err)
+			_, toRetry := processResults(st, resultsByAckID, md)
+			if len(toRetry) > 0 {
+				// Retry acks in a separate goroutine.
+				go func() {
```
mmicatka, line 481: For the retries of transient errors, do we want to add  back-off of some sort?

pubsub/iterator.go
```diff
@@ -528,39 +518,103 @@ func (it *messageIterator) sendModAck(m map[string]*ackResultWithID, deadline ti
...
 
-func (it *messageIterator) sendAckIDRPC(ackResultsByAckID map[string]*AckResult, maxSize int, call func([]string) error) bool {
-	ackIDs := make([]string, 0, len(ackResultsByAckID))
-	for k := range ackResultsByAckID {
-		ackIDs = append(ackIDs, k)
+func (it *messageIterator) retryAcks(m map[string]*AckResult) {
+	ctx, cancel := context.WithTimeout(context.Background(), exactlyOnceDeliveryRetryDeadline)
+	defer cancel()
+	bo := newExactlyOnceBackoff()
+	for {
```
mmicatka, line 542: I don't think we do this in the Java client library, but thoughts on adding a similar retry count that exists in the `retryModAcks` here?

hongalex: I can see adding this retry count for metrics but it's out of the scope of Prad's PR sequencing doc I think. IIRC, we cap modacks retries since after a while messages will just be expired.

pubsub/iterator.go
```diff
@@ -528,39 +518,103 @@ func (it *messageIterator) sendModAck(m map[string]*ackResultWithID, deadline ti
...
+			}
+			return
+		}
+		// Only retry modack requests up to 3 times.
+		if deadlineSec != 0 && retryCount > 3 {
+			ackIDs := make([]string, 0, len(m))
+			for k := range m {
+				ackIDs = append(ackIDs, k)
+			}
+			log.Printf("automatic lease modack retry failed for following IDs: %v", ackIDs)
```
codyoss, line 593: nit: does pubsub log today? I don't think most(any?) of our clients do right now

codyoss (approved):
Overall looked good, just one nit/question.

mmicatka (approved):
LGTM

pubsub/iterator.go
hongalex: Summarizing our direct conversation, we don't log right now and most Go clients don't. There should be a better way to do logging and we should expose hooks for users to do this.

PR #6486, opened 2022-08-08
feat(spanner): set client wide ReadOptions, ApplyOptions, and TransactionOptions

shuheiktgw:
Currently, if we want to set ReadOptions, ApplyOptions, or TransactionOptions, we need to pass those options as a parameter each time we call the related functions. That is tedious and hard to maintain. Instead, I'd like to specify those options as part of the ClientConfig and use them as default values as the QueryOptions already does. Thanks for your review 🙂


spanner/transaction.go
```diff
@@ -1106,11 +1143,19 @@ type CommitResponse struct {
...
+// CommitOptions provides options for committing a transaction in a database.
 type CommitOptions struct {
 	ReturnCommitStats bool
 }
 
+// merge combines two CommitOptions that the input parameter will have higher
+// order of precedence.
+func (co CommitOptions) merge(opts CommitOptions) CommitOptions {
+	return CommitOptions{
+		ReturnCommitStats: co.ReturnCommitStats || opts.ReturnCommitStats,
```
shuheiktgw, line 1155: I could not figure out a good way to merge the ReturnCommitStats...

shuheiktgw:
@rahul2393 Would you review this PR when you have time? 🙂

rahul2393:
@shuheiktgw Can you also please add relevant tests for this?

shuheiktgw:
Thanks for your comment, @rahul2393! I've added unit tests, so would you review the PR? 🙏

rahul2393:
Please resolve conflicts @shuheiktgw

shuheiktgw:
Sure, I've resolved the conflict so would you take a look? 🙂 @rahul2393

rahul2393 (approved):

Issue #6487, opened 2022-08-09
bigquery: managedwriter.WithAppendRowsCallOption does not pass CallOptions when creating the raw AppendRows client


acleone:
**Client**

cloud.google.com/go/bigquery@v1.37.0

#5078 added `managedwriter.WithAppendRowsCallOption`, but the CallOptions are not passed to `streamFunc` when creating the underlying `AppendRows` client, so `WithAppendRowsCallOption` does nothing. This seems to be an oversight when calling `streamFunc` in `ms.open`.

**Code**

 https://github.com/googleapis/google-cloud-go/blob/bigquery/v1.37.0/bigquery/storage/managedwriter/client.go#L104 :

```go
		open: func(streamID string, opts ...gax.CallOption) (storagepb.BigQueryWrite_AppendRowsClient, error) {
			arc, err := streamFunc(
				// Bidi Streaming doesn't append stream ID as request metadata, so we must inject it manually.
-				metadata.AppendToOutgoingContext(ctx, "x-goog-request-params", fmt.Sprintf("write_stream=%s", streamID)))
+				metadata.AppendToOutgoingContext(ctx, "x-goog-request-params", fmt.Sprintf("write_stream=%s", streamID)),
+				opts...)
			if err != nil {
				return nil, err
			}
			return arc, nil
		},
```

Repro:

```go
package main

import (
	"cloud.google.com/go/bigquery/storage/managedwriter"
	"github.com/googleapis/gax-go/v2"
	"google.golang.org/grpc"
)

func main() {
	managedClient, _ := managedwriter.NewClient(...)
	gaxOpt := gax.WithGRPCOptions(grpc.MaxCallSendMsgSize(10 * 1024 * 1024))
	writerOpts := []managedwriter.WriterOption{
		managedwriter.WithDestinationTable(tablePath),
		managedwriter.WithType(managedwriter.DefaultStream),
		managedwriter.WithSchemaDescriptor(...),
		managedwriter.WithAppendRowsCallOption(gaxOpt),
	}
	managedStream, _ := managedClient.NewManagedStream(
		context.Background(),
		writerOpts...)
	// managedStream.open does not correctly pass opts to AppendRows
	// func (c *BigQueryWriteClient) AppendRows(ctx context.Context, opts ...gax.CallOption) (storagepb.BigQueryWrite_AppendRowsClient, error)
	
}
```

**Expected behavior**

[...]

PR #6488, opened 2022-08-09
fix(bigquery/storage/managedwriter): propagate calloptions to append

shollyman:
Reporter identified that we're not properly propagating user-specified
call options when we open the AppendRows bidi stream.

Fixes: https://github.com/googleapis/google-cloud-go/issues/6487


shollyman:
The actual fix here is to pass the variadic `opts...` when `streamFunc` is invoked after attaching metadata.  The rest of the changes in this PR are to make it testable.

quartzmo (approved):

PR #6493, opened 2022-08-09
feat(bigquery): add trace instrumentation support for individual rpcs

shollyman:
This PR adds further opentelemetry trace instrumentation.  Specifically, it adds span information for individual rpc-level retries.

This PR also adds two new callsites where we add retry support:
* dataset deletion
* job listing


alvarowolfx (approved):
LGTM

shollyman:
> Out of curiosity, how do you usually test those changes related to tracing ? Do you run a local app pointing to Stackdriver, use some local stack for tracing like Jaeger or other options ?

Good point, and thanks for keeping me honest.  Added a new integration test for tracing.  And good news, it highlighted other issues which have been addressed.

PR #6499, opened 2022-08-10
feat(compute/metadata): add http timeout to metadata client

### LINKED ISSUE #6498: metadata: need http timeout
**Client**

metadata

**Code**

https://github.com/googleapis/google-cloud-go/blob/main/compute/metadata/metadata.go#L66

tcp dial timeout is set, but http timeout is not.  If http server connects, but is slow to send response, client will hang indefinitely

**Expected behavior**

http timeout occurs

**Actual behavior**

client hangs indefinitely

therealmitchconnors:
Fixes #6498

### DIFF
--- a/compute/metadata/metadata.go
+++ b/compute/metadata/metadata.go
@@ -71,7 +71,7 @@ func newDefaultHTTPClient() *http.Client {
 				KeepAlive: 30 * time.Second,
 			}).Dial,
 		},
-		Timeout: 2 * time.Second,
+		Timeout: 5 * time.Second,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


therealmitchconnors:
> I think this is a reasonable request but that is a rather short timeout. Could be make this 5 seconds at least to start. We have similar timeouts elsewhere in this package.

I have updated to 5 seconds.  Please note that the dial timeout is still set to 2 seconds, so the timeout behavior will vary depending on the specifics.

codyoss (approved):
Thanks

PR #6500, opened 2022-08-10
feat: add new aliasgen tool

codyoss:
This tool is for an upcoming migration effort. We can use it to
make the migration quicker and less error prone.


internal/aliasgen/aliasgen.go
```diff
@@ -0,0 +1,421 @@
...
+		ti.name = v.Name()
+		obj, isPtr, err := getTypeNameForFn(v.Type(), false)
+		if err != nil {
+			return nil, err
+		}
+		ti.typeName = obj.Name()
+		ti.pkg = obj.Pkg().Name()
+		ti.isPtr = isPtr
+		fn.returns = append(fn.returns, ti)
+	}
```
noahdietz, line 150: These could go in a function that takes a `*types.Tuple` and then called twice, once with `sig.Params()` and once with `sig.Results()` for the `fn.params` and `fn.returns` respectively. Thoughts?

noahdietz (commented):
This is coooooool

internal/aliasgen/aliasgen_test.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		t.Fatalf("ReadFile: %v", err)
+	}
+	wantBytes, err := os.ReadFile(filepath.Join(goldenDir, "fake", "alias.go"))
+	if err != nil {
+		t.Fatalf("ReadFile: %v", err)
+	}
+	gotBytes = gotBytes[bytes.IndexRune(gotBytes, '\n')+1:]
+	wantBytes = wantBytes[bytes.IndexRune(wantBytes, '\n')+1:]
+	if ok := bytes.Equal(gotBytes, wantBytes); !ok {
```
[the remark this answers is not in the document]
codyoss: Good idea.

internal/aliasgen/cmd/main.go
```diff
@@ -0,0 +1,41 @@
...
+	"cloud.google.com/go/internal/aliasgen"
+)
+
+var (
+	srcDir  = flag.String("source", "", "the source directory to scan to make aliases")
+	destDir = flag.String("destination", "", "the destination directory where the aliases will be generated")
+)
+
+const (
+	softLineBreak = 77
```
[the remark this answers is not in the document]
codyoss: Oops you are right!

internal/aliasgen/aliasgen.go
codyoss: great idea.

internal/aliasgen/aliasgen.go
```diff
@@ -0,0 +1,423 @@
...
+		importPath: pkg.PkgPath,
+		pkg:        strings.TrimSuffix(pkg.Name, "pb"),
+	}
+
+	// Load corresponding documentation.
+	docPkg, err := doc.NewFromFiles(pkg.Fset, pkg.Syntax, pkg.PkgPath)
+	if err != nil {
+		return nil, err
+	}
+	identToDoc := make(map[string]string, len(docPkg.Types))
```
quartzmo, line 82: Would it also be possible to load docs from `Consts` and `Vars` in order to add similar `Deprecated` labels, such as the deprecation on these consts, for example? https://go.dev/src/os/file.go?h=Deprecated%3A#L87

codyoss: I could add this. The reason I did not is these are not treated special in editor or godoc like methods/types/packages. See the godoc for that example: https://pkg.go.dev/os#pkg-constants. It is just a comment so it would be surfaced by tools.

codyoss: While I was at it just did this for the funcs too since they did not have the warning.

internal/aliasgen/aliasgen.go
```diff
@@ -0,0 +1,432 @@
...
+		}
+
+		// build return info
+		for i, r := range f.returns {
+			if i == 0 {
+				if _, err := fmt.Fprintf(w, " "); err != nil {
+					return err
+				}
+				if len(f.returns) > 1 {
+					if _, err := fmt.Fprintf(w, "("); err != nil {
```
quartzmo, line 336: I believe this is untested, maybe add a func with multiple return values to `fakepb`?

codyoss: This actually never happens so I will just simpify code and return an error if it does.

quartzmo (approved):

PR #6506, opened 2022-08-11
feat(pubsub): support exactly once delivery

hongalex:
The `pubsub-exactly-once` branch has code that has been reviewed in individual PRS already.


pubsub/iterator.go
```diff
@@ -232,17 +259,32 @@ func (it *messageIterator) receive(maxToPull int32) ([]*Message, error) {
...
+		if _, ok := it.pendingNacks[ackID]; !ok {
+			// Don't use the message's AckResult here since these are only for receipt modacks.
+			// ModAckResults are transparent to the user anyway so these can automatically succeed.
+			// We can't use an empty AckResult here either since SetAckResult will try to
+			// close the channel without checking if it exists.
+			ackIDs[ackID] = newSuccessAckResult()
+		}
+		// If exactly once is enabled, keep track of all pending AckResults
+		// so we can cleanly close them all at shutdown.
+		it.eoMu.Lock()
```
maheshgattani, line 279: Any concern about performance impact to non exactly once delivery subs due to these locks, here and below?

can the critical section be reduced if we read the value of `it.enableExactlyOnceDelivery` into a local variable inside a reader lock, and then execute the rest of the logic using that variable? Would that impact the logical correctness?

pubsub/iterator.go
```diff
@@ -624,3 +723,66 @@ func maxDuration(x, y time.Duration) time.Duration {
 	}
 	return y
 }
+
+const (
+	transientErrStringPrefix     = "TRANSIENT_"
+	transientInvalidAckErrString = transientErrStringPrefix + "FAILURE_INVALID_ACK_ID"
```
maheshgattani, line 729: This should never happen. Invalid ack id is a permanent failure, always. I do not see this variable used anywhere apart from tests so maybe the logic is right, just that we don't need this variable.

hongalex: Ah, I wasn't aware of this. I was mostly copying [Python's error strings](https://github.com/googleapis/python-pubsub/blob/e600ad8228930445765ffa0c45500a7779e25817/tests/unit/pubsub_v1/subscriber/test_streaming_pull_manager.py#L2187) though I removed the variable anyway and made it a generic "TRANSIENT_FAILURE" error string for tests.

pubsub/iterator.go
hongalex: Yeah the lock acquisition should've been a `RLock() + RUnlock()` here and outside of the loop. I made a similar optimization (moving the mutex locking) in other places as well. The only place where the mutex is locked for writing is when receiving on the stream.

maheshgattani (approved):

pubsublite/pscompat/subscriber.go
```diff
@@ -75,6 +75,22 @@ func (ah *pslAckHandler) OnNack() {
 	ah.subInstance = nil
 }
 
+// OnAckWithResult is required implementation for the ack handler
+// for Cloud Pub/Sub's exactly once delivery feature. This will
+// ack the message and return nil.
+func (ah *pslAckHandler) OnAckWithResult() *pubsub.AckResult {
+	ah.OnAck()
+	return nil
```
tmdiep, line 83: There seems to be a discrepancy between the comments in internal/pubsub/message.go ("if using Pub/Sub Lite, AckResult readies immediately with a AcknowledgeStatus.Success") and the return value here.

Should this return a non-nil `pubsub.AckResult`? Same below for `OnNackWithResult`.

Please correct me, if I've misread.

hongalex: Hm yeah in hindsight we should probably just return an AckResult that returns success instead. I'll update this and the comments.

tmdiep: LGTM for pubsublite. Thanks!

tmdiep (approved):

quartzmo (approved):

PR #6511, opened 2022-08-12
chore(logging): added extra variables on environment test runs

daniel-sanche:
This PR adds extra labels to the environment test kokoro script to make filtering easier for our test dashboards. It sets the following variables:

area: observability
product: logging
language: go

Also switched the Python env var for functions tests to match the version in the test container


internal/kokoro/environment.sh
```diff
@@ -29,6 +29,11 @@ if [[ -z "${PROJECT_ROOT:-}"  ]]; then
     PROJECT_ROOT="github/google-cloud-go"
 fi
 
+# add kokoro labels for testgrid filtering
+export PRODUCT_AREA_LABEL=observability
```
codyoss, line 33: Are these envvars already allow-listed in the internal kokoro config?

daniel-sanche: yes

codyoss (approved):

PR #6515, opened 2022-08-15
feat(spanner/spansql): Add tests and parser support for THEN RETURN

c2nes:
This change adds THEN RETURN support for DML statements. The existing "Query" methods already support executing such statements, so this change is limited to updating documentation, adding integration tests, and updating the spansql package. Support has not been added to the spannertest in-memory implementation of Spanner, but the feature absence has been noted in the corresponding README.


c2nes:
Note: Integration tests are not expected to pass until `THEN RETURN` support is launched.

spanner/doc.go
```diff
@@ -338,10 +338,14 @@ NULL-able STRUCT values.
 # DML and Partitioned DML
 
 Spanner supports DML statements like INSERT, UPDATE and DELETE. Use
-ReadWriteTransaction.Update to run DML statements. It returns the number of rows
-affected. (You can call use ReadWriteTransaction.Query with a DML statement. The
-first call to Next on the resulting RowIterator will return iterator.Done, and
-the RowCount field of the iterator will hold the number of affected rows.)
+ReadWriteTransaction.Update to run simple DML statements and
+ReadWriteTransaction.Query to run DML statements containing "THEN RETURN"
```
rajatbhatta, line 342: I think mentioning "THEN RETURN" here will make it specific to GoogleStandardSQL dialect (PostgreSql dialect uses "RETURNING"). Should we change the documentation to reflect both the dialects?

spanner/spannertest/README.md
```diff
@@ -24,6 +24,7 @@ by ascending esotericism:
 - SELECT HAVING
 - more literal types
 - DEFAULT
+- THEN RETURN
```
rahul2393, line 27: We are supporting it through this PR, right? I guess we want to say PG.Returning is not supported.

c2nes: No, I did not intend to add this support myself. We will be adding support to the standard Spanner Emulator however.

spanner/doc.go
c2nes: Done.

rajatbhatta: @c2nes: Can you check if we missed the documentation update here?

spanner/integration_test.go
```diff
@@ -2943,6 +2943,130 @@ func TestIntegration_DML(t *testing.T) {
 	}
 }
 
+func TestIntegration_DMLReturning(t *testing.T) {
```
rajatbhatta, line 2946: @c2nes: Can we add end-to-end tests for running DML statement with returning clause using `tx.Update` and `tx.BatchUpdate` as well.

spanner/doc.go
c2nes: Whoops, sorry about that. Forgot to stage the doc update before amending. Should be updated now.

spanner/integration_test.go
c2nes: Done.

rahul2393 (approved):

codyoss:
@rahul2393 are there still plans to accecpt this PR?

codyoss:
@rahul2393 see comment above ^

PR #6537, opened 2022-08-17
testing(bigquery/storage/managedwriter):  instrumentation retries

shollyman:
We've caught more than one instance of a race failure.  Add retries
for the opencensus metrics test to reduce the likelihood of this
failures.

Fixes: #6535


codyoss (approved):
Thanks!

PR #6538, opened 2022-08-17
test(storage): skip tests using OLM Age condition

frankyn:
When https://github.com/googleapis/google-api-go-client/pull/1598 is released; these tests will break due to the breaking change. This is a preemptive PR to skip these tests which will be reenabled once the break is released.

### DIFF
--- a/storage/bucket_test.go
+++ b/storage/bucket_test.go
@@ -25,9 +25,9 @@ import (
 	raw "google.golang.org/api/storage/v1"
 )
 
-// TODO: re-enable tests after breaking change is released and AgeInDays is a int64*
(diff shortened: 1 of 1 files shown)


storage/bucket_test.go
```diff
@@ -25,7 +25,9 @@ import (
 	raw "google.golang.org/api/storage/v1"
 )
 
+// TODO: re-enable tests after breaking change is released and AgeInDays is a int64*
 func TestBucketAttrsToRawBucket(t *testing.T) {
+	t.Skip()
```
tritone, line 30: Make a bug for the todo and put the link as an argument to the skip.

tritone (approved):

PR #6554, opened 2022-08-23
chore(logging): update environment tests to use python3.9

daniel-sanche:
A [recent change](https://github.com/googleapis/google-cloud-go/commit/57c2a9ebfc2467974c488750da6b51fa3f81009b) removed the Python installation used by the logging environment tests, in favor of baking Python into the image. This caused the tests to fail, as `python` and `python3.7` were no longer present in the image

This PR addresses this by using `python3` instead, and changing the default Python runtime variable to 3.9


codyoss (approved):

codyoss:
Sorry about that!

PR #6556, opened 2022-08-23
test(storage): add scenario 8 to retry conformance tests

cojenco:
This adds scenario 8 to test downloads can handle complex retries, per internal docs go/retry-conf-test-media-ops

- add data integrity checks for downloads in retry conf tests
- introduce new field `method.Group` to test proto
- update test schema
- construct 9MiB larger test files to enable S8

**Update:** testbench new tag released on 08/29 with https://github.com/googleapis/storage-testbench/pull/397

Testcase as the following
1. client makes a request
2. request returns with 2xx
3. client starts reading the request body
4. the body stream is broken a few bytes into reading

Note: comma in the license headers had to be removed per the newly enabled [header-checker-lint](https://github.com/googleapis/repo-automation-bots/blob/main/packages/header-checker-lint/src/header-parser.ts#L24)


storage/retry_conformance_test.go
```diff
@@ -199,10 +202,39 @@ var methods = map[string][]retryFunc{
...
+			wr, err := io.Copy(ioutil.Discard, r)
+			if got, want := wr, len(randomBytesToWrite); got != int64(want) {
+				return fmt.Errorf("body length mismatch\ngot:\n%v\n\nwant:\n%v", got, want)
+			}
 			return err
 		},
 	},
+	"storage.objects.download": {
+		func(ctx context.Context, c *Client, fs *resources, _ bool) error {
+			// Before running the test method, populate a large test object of 9 MiB.
```
tritone, line 214: Since this happens before the test, I think this should be another fixture in `*resources` rather than something happening in the download method. We could call it LARGE_OBJECT. Does that seem reasonable? I understand that this would be a proto change also...

storage/retry_conformance_test.go
```diff
@@ -199,10 +202,39 @@ var methods = map[string][]retryFunc{
...
+			if err := uploadTestObject(fs.bucket.Name, objName, randomBytes9MB); err != nil {
+				return fmt.Errorf("failed to create 9 MiB large object pre test, err: %v", err)
+			}
+			// Download the large test object for the S8 download method group.
+			r, err := c.Bucket(fs.bucket.Name).Object(objName).NewReader(ctx)
+			if err != nil {
+				return err
+			}
+			defer r.Close()
+			body, err := ioutil.ReadAll(r)
```
tritone, line 225: nit: might just call this data rather than body

tritone (commented):
Overall looks good, one suggestion about how this should be factored

storage/retry_conformance_test.go
cojenco: done

storage/retry_conformance_test.go
cojenco: Thanks Chris for the review. Your proposal is reasonable. I have reservations at this moment as it calls for changes in the json test schema, test proto, and moreover the implementations completed in other languages. 

Currently, Java uses a switch case bytes supplier that accommodates S8 large objects, Node.js has a similar helper function that generates given size objects, and Python uses a pytest fixture. My bad for not clarifying the construction details in the design doc earlier. How about we add this in the future improvements section?

tritone: Discussed offline; we think this is a good idea but will file an issue for the work across repos rather than block this PR on the change.

tritone (approved):

Issue #6492, opened 2022-08-09
bigquery: schema field of type NUMERIC is not assignable to struct field


ejstembler:
**Client**

BigQuery

**Environment**

golang:1.18

**Go Environment**

go version go1.18.5 darwin/amd64

```
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/my-user/Library/Caches/go-build"
GOENV="/Users/my-user/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/my-user/.asdf/installs/golang/1.18.5/packages/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/my-user/.asdf/installs/golang/1.18.5/packages"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/Users/my-user/.asdf/installs/golang/1.18.5/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/Users/my-user/.asdf/installs/golang/1.18.5/go/pkg/tool/darwin_amd64"
GOVCS=""
GOVERSION="go1.18.5"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/my-user/Projects/my-project/go.mod"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/j3/zqjvh17x5hz0d87m0wtkq6cm0000gn/T/go-build792818999=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

**Code**

```go
package main

type MyRecord struct {
    Quantity float64 `bigquery:"Quantity,nullable"` // NUMERIC BigQuery type
    // other fields...
}
```

```go
package main

[...]

PR #6560, opened 2022-08-24
docs(bigquery): add numeric and bignumeric to RowIterator docs

alvarowolfx:
Document field type conversions for `NUMERIC` and `BIGNUMERIC` types when using the `RowIterator` for reading data from BigQuery. 

Solves issue #6492


shollyman (approved):
Thanks for this!

PR #6566, opened 2022-08-25
feat(bigquery): improve error when reading null values

### LINKED ISSUE #2612: bigquery: Logging field name on error in null value setting to struct field
An error occurs when the null value of bigquery set to the primitive type field of struct of go, and the following log appears.
https://github.com/googleapis/google-cloud-go/blob/c4d15341e71f04c0ff1e32ddb664efb79da28377/bigquery/value.go#L111

this error log is abstract and it is difficult to debug.Especially when the number of fields in the struct is large, it takes time to find the offending field.

Specifically output to the log which field name(or tag name) the error occurred in mapping.

For example, the following log.
`NULL Value cannot assignable to struct field any_field_name of type int.`

Add logs to the following parts of the module
https://github.com/googleapis/google-cloud-go/blob/c4d15341e71f04c0ff1e32ddb664efb79da28377/bigquery/value.go#L430

```
log.Printf("set value to %s", vstruct.Type().FieldByIndex(op.fieldIndex).Name)
err = op.setFunc(field, values[op.valueIndex])
```

alvarowolfx:
Improve error message when reading NULL values from a query and trying to fit into a struct field.

Resolves #2612


bigquery/value.go
```diff
@@ -455,6 +455,10 @@ func runOps(ops []structLoaderOp, vstruct reflect.Value, values []Value) error {
 			err = setRepeated(field, values[op.valueIndex].([]Value), op.setFunc)
 		} else {
 			err = op.setFunc(field, values[op.valueIndex])
+			if errors.Is(err, errNoNulls) {
+				f := vstruct.Type().FieldByIndex(op.fieldIndex)
+				err = fmt.Errorf("bigquery: NULL cannot be assigned to field `%s` of type %s", f.Name, f.Type.Name())
```
shollyman, line 460: A thing to keep an eye on for the future is whether we should have a custom error type.  My suspicion is this isn't a frequent enough failure to warrant custom errors at this point, nor is it super actionable unless you're dynamically rewriting things.

shollyman (approved):
Thanks for this!

PR #6568, opened 2022-08-25
test(bigtable): adds test proxy for Bigtable test suite

telpirion:
For use with the [Bigtable test suite](https://github.com/googleapis/cloud-bigtable-clients-test).


bigtable/internal/testproxy/proxy.go
```diff
@@ -102,10 +102,10 @@ func (c credentialsBundle) NewWithMode(mode string) (credentials.Bundle, error)
 //
 // There are three base cases to address:
 //  1. CreateClientRequest specifies no unique credentials; so ADC will be used.
-//     This method returns an empty slice.
-//  2. CreateClientRequest specifies both call and channel credentials. In
-//     this case, we need to create a combined credential.
-//  3. CreateClientRequest specifies only a channel credential.
+//     This method returns an empty slice..
+//  2. CreateClientRequest specifies only a channel credential.
+//  3. CreateClientRequest specifies both call and channel credentials. In
```
liujiongxin, line 401: Does the ordering matter? Also the docstring doesn't agree with the code in terms of order.

bigtable/internal/testproxy/proxy.go
```diff
@@ -202,30 +213,71 @@ func (s *goTestProxyServer) CreateClient(ctx context.Context, req *pb.CreateClie
...
+		cancel()
+		return nil, stat.Error(codes.AlreadyExists,
+			fmt.Sprintf("%s: ClientID already exists %v", logLabel, err))
+	}
+
+	if s.clientIDs == nil {
+		s.clientIDs = make(map[string]testClient)
+	}
+
+	s.clientIDs[req.ClientId] = testClient{
```
liujiongxin, line 550: Should lock be used to have exclusive update of the map?

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,448 @@
...
+	"google.golang.org/api/option"
+	btpb "google.golang.org/genproto/googleapis/bigtable/v2"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/test/bufconn"
+)
+
+const (
+	buf = 1024 * 1024
```
enocom, line 37: These are good local variable names, but not great const names because they're used well beyond their declaration here. Cf. "The greater the distance between a name's declaration and its uses, he longer the name should be." (https://go.dev/talks/2014/names.slide#4)

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+	err = populateTable(bts)
+	if err != nil {
+		log.Fatalf("testproxy setup: can't populate mock table")
+	}
+
+	// 3) Start the test proxy server
+	lis = bufconn.Listen(buf)
+	s := newProxyServer(lis)
+	go func() {
```
enocom, line 138: How does this goroutine exit?

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,448 @@
...
+		if err := s.Serve(lis); err != nil {
+			log.Fatalf("failed to serve: %v", err)
+		}
+	}()
+
+	// 2) Create the test proxy client
+	conn2, err := grpc.DialContext(ctx, lis.Addr().String(), grpc.WithContextDialer(bufDialer), grpc.WithTransportCredentials(insecure.NewCredentials()))
```
enocom, line 145: nit: line length

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,448 @@
...
+	if err != nil {
+		log.Fatalf("testproxy setup: failed to dial testproxy: %v", err)
+	}
+	defer conn2.Close()
+	c := pb.NewCloudBigtableV2TestProxyClient(conn2)
+	client = &c
```
enocom, line 151: Why not assign directly into client?

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,448 @@
...
+	// However, we need to have the test proxy server running in order to create
+	// a reusable client stored in its memory
+	req := &pb.CreateClientRequest{
+		ClientId:   tpc,
+		ProjectId:  "client",
+		DataTarget: bta,
+		InstanceId: "instance",
+	}
+
+	_, err = (*client).CreateClient(ctx, req)
```
enocom, line 164: The indirection is unnecessary here. Go will do that for you.

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,448 @@
...
+	if err != nil {
+		log.Fatalf("testproxy setup:  CreateClient() failed: %v", err)
+	}
+
+	m.Run()
```
enocom, line 169: os.Exit(m.Run())

bigtable/internal/testproxy/proxy_test.go
enocom: Ditto in the cases below.

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,448 @@
...
+	ctx := context.Background()
+
+	req := &pb.CreateClientRequest{
+		ClientId:   cid,
+		ProjectId:  "client",
+		DataTarget: bta,
+		InstanceId: "instance",
+	}
+
+	_, err := (*client).CreateClient(ctx, req)
```
enocom, line 184: What is the significance of error being nil? Did the client initiate a connection to a server? A comment here would help.

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,448 @@
...
+		ClientId:  cid,
+		CancelAll: true,
+	})
+
+	if err != nil {
+		t.Errorf("testproxy test: CreateClient() teardown failed: %v", err)
+	}
+}
+
+func TestRemoveClient(t *testing.T) {
```
enocom, line 200: Is this a duplicate of the test above? Shall it be rolled into the test above?

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,448 @@
...
+	}
+
+	resp, err := (*client).MutateRow(ctx, req)
+	if err != nil {
+		t.Fatalf("testproxy test: MutateRow() returned error: %v", err)
+	}
+
+	if resp.Status.Code != int32(codes.OK) {
+		t.Errorf("testproxy test: MutateRow() didn't return OK; got %v", resp.Status.Code)
+	}
```
enocom, line 326: Should this check that the new row has in fact mutated?

enocom: Ditto below in BulkMutate

enocom (changes requested):
Just looking at the test. I'll make another pass on the proxy.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+
+		creds = credentials.NewClientTLSFromCert(pool, sslTargetName)
+		if err != nil {
+			return nil, err
+		}
+	case *pb.ChannelCredential_None:
+		creds = insecure.NewCredentials()
+	default:
+		ctx := context.Background()
+		c, err := gauth.FindDefaultCredentials(ctx, []string{"https://www.googleapis.com/auth/cloud-platform"}...)
```
enocom, line 491: gauth.FindDefaultCredentials(ctx, "https://www.googleapis.com/auth/cloud-platform")

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		// TODO(developer): Determine how to pass this call option back to caller
+		option.WithTokenSource(c.TokenSource)
+
+		return nil, nil
+	}
+	return creds, nil
```
enocom, line 501: You might be able to remove this since you have a default case with a return in it.

bigtable/internal/testproxy/proxy.go
enocom: +1 I'd guess that CreateClient should be thread safe and so a lock is required here.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+	s.clientIDs[req.ClientId] = testClient{
+		c: c,
+		cancels: []context.CancelFunc{
+			cancel,
+		},
+		appProfileID:        req.AppProfileId,
+		perOperationTimeout: req.PerOperationTimeout,
+	}
+
+	res := &pb.CreateClientResponse{}
```
enocom, line 559: return &pb.CreateClientResponse{}, nil

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+
+	localCtx, cancel := context.WithCancel(context.Background())
+	c, err := bigtable.NewClient(localCtx, req.ProjectId, req.InstanceId, option.WithGRPCConn(conn))
+	if err != nil {
+		cancel()
+		return nil, stat.Error(codes.Internal,
+			fmt.Sprintf("%s: failed to create client: %v", logLabel, err))
+	}
+
+	if s.clientIDs == nil {
```
enocom, line 546: Does this need to be threadsafe?

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+func (s *goTestProxyServer) CreateClient(ctx context.Context, req *pb.CreateClientRequest) (*pb.CreateClientResponse, error) {
+	if req.ClientId == "" ||
+		req.DataTarget == "" ||
+		req.ProjectId == "" ||
+		req.InstanceId == "" {
+		return nil, stat.Error(codes.InvalidArgument,
+			fmt.Sprintf("%s must provide ClientId, DataTarget, ProjectId, and InstanceId", logLabel))
+	}
+
+	if _, exists := s.clientIDs[req.ClientId]; exists {
```
enocom, line 522: Does this need to be thread safe?

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+		return nil, nil
+	}
+	return creds, nil
+}
+
+// goTestProxyServer represents an instance of the test proxy server. It keeps
+// a reference to individual clients instances (stored in a testClient object).
+type goTestProxyServer struct {
+	pb.UnimplementedCloudBigtableV2TestProxyServer
+	clientIDs map[string]testClient // clientIDs has all of the bigtable.Client objects under test
```
enocom, line 508: Every time you access `clientIDs`, you'll need to synchronize access since all these methods below are called concurrently.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+	t := btc.c.Open(rrq.TableName)
+
+	ctx = btc.addCancelFunction(ctx)
+
+	rowPbs := rrq.Rows
+	var rs bigtable.RowSet
+
+	// Bigtable client doesn't have a Table.GetAll() function--RowSet must be
+	// provided for ReadRows. Use InfiniteRange() to get the full table.
+	if len(rowPbs.GetRowKeys()) == 0 && len(rowPbs.GetRowRanges()) == 0 {
```
enocom, line 649: Whenever you have an if/else that assigns into the same variable in both branches, consider initializing the variable to a default value, and overriding it under a if statement. The else can be removed in that case.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+		return nil, stat.Error(codes.InvalidArgument, "request to CheckAndMutateRow() is missing inner request")
+	}
+
+	trueMuts := mutationFromProto(rrq.TrueMutations)
+	falseMuts := mutationFromProto(rrq.FalseMutations)
+
+	rfPb := rrq.PredicateFilter
+	var f bigtable.Filter
+
+	if rfPb != nil {
```
enocom, line 809: Ditto here on the if/else. You can initialize `f` to `bigtablePassAllFilter()` and then override it if need be.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+	s := grpc.NewServer()
+	pb.RegisterCloudBigtableV2TestProxyServer(s, &goTestProxyServer{})
+	log.Printf("server listening at %v", lis.Addr())
+	return s
+}
+
+func main() {
+	flag.Parse()
+
+	log.Println(*port)
```
enocom, line 952: A nicer log message here might be nice.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+
+	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", *port))
+	if err != nil {
+		log.Fatalf("failed to listen: %v", err)
+	}
```
enocom, line 957: this needs a `defer lis.Close()` after it

enocom: ditto below for the proxy server.

enocom (changes requested):

bigtable/internal/testproxy/README.md
```diff
@@ -1,13 +1,26 @@
...
+for the Test Frameworks to see usage steps.
+
+## Running the test proxy
 
 1. Open a command terminal at the root of the Bigtable client library.
 
 1. Run the test proxy (this)
 
 ```
 go run internal/testproxy/main.go
```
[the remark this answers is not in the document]
telpirion: Done.

bigtable/internal/testproxy/proxy.go
```diff
@@ -757,18 +760,22 @@ func (s *goTestProxyServer) BulkMutateRows(ctx context.Context, req *pb.MutateRo
 	// Iterate over any errors returned, matching indices with errors. If
 	// errs is nil, this block is skipped.
 	for i, e := range errs {
+		var me *btpb.MutateRowsResponse_Entry
 		if e != nil {
-			me := &btpb.MutateRowsResponse_Entry{
+			me = &btpb.MutateRowsResponse_Entry{
 				Index: int64(i),
 				Status: &status.Status{
 					Code:    int32(codes.Internal),
```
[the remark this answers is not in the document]
telpirion: No, because the type of `e` is `error` rather than a `status.Status` with a code.

bigtable/internal/testproxy/main.go
```diff
@@ -40,6 +41,32 @@ func (s *goTestProxyServer) RemoveClient(ctx context.Context, req *pb.RemoveClie
...
+	t := s.btClient.Open(tName)
+
+	r, err := t.ReadRow(ctx, req.RowKey)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if r != nil {
+		return nil, fmt.Errorf("no error or row returned from ReadRow()")
```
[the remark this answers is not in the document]
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: No, the ordering doesn't matter. This comment provides guidance for future maintainers of this code, by giving explicit numbers to each branch/technique for getting credentials.

bigtable/internal/testproxy/proxy.go
telpirion: [facepalm] Yep, I completely forgot about making this application threadsafe. I'll update the code accordingly.

bigtable/internal/testproxy/proxy_test.go
telpirion: I believe that it exits when the tests are done running. My thinking was that when the main routine (the test process) ends, all child goroutines (like this one) are closed.

I'm open to suggestions for a better approach!

bigtable/internal/testproxy/proxy_test.go
telpirion: Yeah, `TestCreateClient()` and `TestRemoveClient()` are essentially two halves of the same coin. They should be combined.

bigtable/internal/testproxy/proxy_test.go
telpirion: Ensuring that the row was in fact mutated is delegated to the larger test suite operation. The code under test (the test proxy) doesn't directly mutate the row (that behavior happens in the Bigtable service, which is not under test).

bigtable/internal/testproxy/proxy_test.go
telpirion: Done.

bigtable/internal/testproxy/proxy_test.go
telpirion: Done.

bigtable/internal/testproxy/proxy_test.go
telpirion: Done.

bigtable/internal/testproxy/proxy_test.go
telpirion: Done.

bigtable/internal/testproxy/proxy_test.go
telpirion: Done.

bigtable/internal/testproxy/proxy_test.go
telpirion: Done.

bigtable/internal/testproxy/proxy_test.go
telpirion: Merged / remove duplicate

bigtable/internal/testproxy/proxy_test.go
telpirion: .... and if I remove this dereference as well :(

bigtable/internal/testproxy/proxy.go
telpirion: The compiler doesn't like it if I remove that return statement 😬

bigtable/internal/testproxy/proxy_test.go
enocom: Yes, that will work probably 90% of the time. Note that the grpc Server has a Stop method, which will cause Serve to finally return. I'd suggest something like this:

``` golang
s := newProxyServer(lis)
defer s.Stop()
```

Stop will also close your listener, so that's another benefit.


https://pkg.go.dev/google.golang.org/grpc#Server.Stop

bigtable/internal/testproxy/proxy.go
telpirion: I typically try to do that. However, in this case there's some nasty type casting that I'm trying to work around.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+	pb "github.com/googleapis/cloud-bigtable-clients-test/testproxypb"
+	"google.golang.org/api/option"
+	btpb "google.golang.org/genproto/googleapis/bigtable/v2"
+	"google.golang.org/genproto/googleapis/rpc/status"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+	oauth "google.golang.org/grpc/credentials/oauth"
+
```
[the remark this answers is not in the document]
telpirion: Done.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,963 @@
...
+	c                   *bigtable.Client     // c stores the Bigtable client under test
+	cancels             []context.CancelFunc // cancels stores a cancel() for each call to this client
+	appProfileID        string               // appProfileID is currently unused
+	perOperationTimeout *duration.Duration   // perOperationTimeout sets a custom timeout for methods calls on this client
+}
+
+// addCancelFunction appends a context.CancelFunc to testClient.cancels slice.
+// It returns a new context object composed from the original.
+func (tc *testClient) addCancelFunction(ctx context.Context) context.Context {
+	var cancel context.CancelFunc
```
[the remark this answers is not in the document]
telpirion: Done

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+
+// mutationFromProto translates a slice of Bigtable v2.Mutation objects into
+// a single Bigtable.Mutation object.
+func mutationFromProto(mPbs []*btpb.Mutation) *bigtable.Mutation {
+	m := bigtable.NewMutation()
+	for _, mpb := range mPbs {
+
+		switch mut := mpb.Mutation; mut.(type) {
+		case *btpb.Mutation_DeleteFromColumn_:
+			del := mut.(*btpb.Mutation_DeleteFromColumn_)
```
enocom, line 148: In the context of the case statement, mut will be the type that you matched on. So doing yet another type assertion is unnecessary.

enocom: Ditto below.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+}
+
+// filterFromProto translates a Bigtable v2.RowFilter object into a Bigtable
+// Filter object.
+func filterFromProto(rfPb *btpb.RowFilter) (f *bigtable.Filter) {
+
+	switch fpb := rfPb.Filter; fpb.(type) {
+	case *btpb.RowFilter_Chain_:
+		c := fpb.(*btpb.RowFilter_Chain_)
+		fs := make([]bigtable.Filter, 0)
```
enocom, line 189: var fs []bigtable.Filter

enocom: Ditto below

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+			m.Set(fam, string(col), bts, val)
+
+		}
+	}
+	return m
+}
+
+// filterFromProto translates a Bigtable v2.RowFilter object into a Bigtable
+// Filter object.
+func filterFromProto(rfPb *btpb.RowFilter) (f *bigtable.Filter) {
```
enocom, line 184: This named return is fine, but generally discouraged in anything but short functions/methods. It would also work to just declare f up-front:

``` golang
var f *bigtable.Filter
```

enocom: That then means you have a:

``` golang
default:
  return nil
}
```

at the end

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+//  2. CreateClientRequest specifies only a channel credential.
+//  3. CreateClientRequest specifies both call and channel credentials. In
+//     this case, we need to create a combined credential (Bundle).
+//
+// Discussed [here](https://github.com/grpc/grpc-go/tree/master/examples/features/authentication).
+// Note that the Go client libraries don't explicitly have the concept of
+// channel credentials, call credentials, or composite call credentials per
+// [gRPC documentation](https://grpc.io/docs/guides/auth/).
+func getCredentialsOptions(req *pb.CreateClientRequest) ([]grpc.DialOption, error) {
+	opts := make([]grpc.DialOption, 0)
```
enocom, line 406: var opts []grpc.DialOption

bigtable/internal/testproxy/proxy.go
enocom: Oh I see some of the branches are missing returns. OK. Not a big deal.

bigtable/internal/testproxy/proxy.go
enocom: Unless you're doing some time consuming IO in this method, you could just do this in all these methods:

``` golang
s.clientsLock.Lock()
defer s.clientsLock.Unlock()
```

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+		return nil, stat.Error(codes.InvalidArgument,
+			fmt.Sprintf("%s: ClientID does not exist", logLabel))
+	}
+	defer s.clientsLock.Unlock()
+
+	if doCancelAll {
+		btc.cancelAll()
+	}
+	btc.c.Close()
+	delete(s.clientIDs, clientID)
```
enocom, line 582: This also needs to be protected.

bigtable/internal/testproxy/proxy.go
enocom: This nil check isn't theadsafe still. See the comment above about wrapping the method in a lock.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+		Row: pbRow,
+	}
+
+	return res, nil
+}
+
+// ReadRows responds to the ReadRows RPC. This method gets all of the column
+// data for a set of rows, a range of rows, or the entire table.
+func (s *goTestProxyServer) ReadRows(ctx context.Context, req *pb.ReadRowsRequest) (*pb.RowsResult, error) {
+	s.clientsLock.Lock()
```
enocom, line 636: There's a missing unlock here that will leave the lock checked out on line 639. Best to unlock with defer.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+		Row: rowsPb,
+	}
+
+	return res, nil
+}
+
+// MutateRow responds to the MutateRow RPC. This methods applies a series of
+// changes (or deletions) to a single row in a table.
+func (s *goTestProxyServer) MutateRow(ctx context.Context, req *pb.MutateRowRequest) (*pb.MutateRowResult, error) {
+	s.clientsLock.Lock()
```
enocom, line 696: Ditto here -- there's a missing unlock in some cases.

enocom: And below.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+
+	ctx = btc.addCancelFunction(ctx)
+
+	errs, err := t.ApplyBulk(ctx, keys, muts)
+	if err != nil {
+		return nil, err
+	}
+	s.clientsLock.Unlock()
+
+	entries := make([]*btpb.MutateRowsResponse_Entry, 0)
```
enocom, line 773: var entries []*btpb.MutateRowsResponse_Entry

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+			me = &btpb.MutateRowsResponse_Entry{
+				Index: int64(i),
+				Status: &status.Status{
+					Code:    int32(codes.Internal),
+					Message: e.Error(),
+				},
+			}
+
+		}
+		entries = append(entries, me)
```
enocom, line 789: This will sometimes add nil to entries. Is that what you intend?

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+	t := btc.c.Open(rrq.TableName)
+	k := string(rrq.RowKey)
+	rpb := rrq.Rules
+
+	rmw := bigtable.NewReadModifyWrite()
+
+	for _, rp := range rpb {
+		switch r := rp.Rule; r.(type) {
+		case *btpb.ReadModifyWriteRule_AppendValue:
+			av := r.(*btpb.ReadModifyWriteRule_AppendValue)
```
enocom, line 922: Don't need this for the reason noted above.

enocom (changes requested):

bigtable/internal/testproxy/proxy.go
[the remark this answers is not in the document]
liujiongxin: Will this help? (https://jbrandhorst.com/post/grpc-errors/)

There are functions like status.FromError() and status.Convert().

bigtable/internal/testproxy/proxy.go
telpirion: I'm afraid the compiler doesn't like it when I remove this type cast, specifically when I attempt to call the methods below.

bigtable/internal/testproxy/proxy.go
telpirion: Is it covered by the deferred `Unlock()` above?

bigtable/internal/testproxy/proxy.go
telpirion: The unlock is missing, it's down below. I was trying to be mindful of the Big O time in these functions (but everything after the `Unlock()` happens in constant time).

bigtable/internal/testproxy/proxy.go
telpirion: Nope! Nice catch.

bigtable/internal/testproxy/proxy.go
telpirion: The compiler gets unhappy if I remove the type cast.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+	logLabel = "cbt-go-proxy"
+)
+
+// rowToProto converts a Bigtable Go client Row struct into a
+// Bigtable protobuf Row struct. It iterates over all of the column families
+// (keys) and ReadItem slices (values) in the client Row struct
+func rowToProto(btRow bigtable.Row) (*btpb.Row, error) {
+	pbRow := &btpb.Row{
+		Key:      []byte(btRow.Key()),
+		Families: make([]*btpb.Family, 0),
```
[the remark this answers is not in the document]
telpirion: Done (and good to know!).

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,976 @@
...
+			var start, end string
+			var rr bigtable.RowRange
+
+			switch rrs.StartKey.(type) {
+			case *btpb.RowRange_StartKeyClosed:
+				start = string(rrs.GetStartKeyClosed())
+			case *btpb.RowRange_StartKeyOpen:
+				start = string(rrs.GetStartKeyOpen())
+			default:
+				start = ""
```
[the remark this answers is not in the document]
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
telpirion: Oh, gotcha. Done.

bigtable/internal/testproxy/proxy.go
telpirion: Done (I hope!)

bigtable/internal/testproxy/proxy_test.go
telpirion: Done.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,626 @@
...
+	case *btpb.RowFilter_StripValueTransformer:
+		sv := bigtable.StripValueFilter()
+		f = &sv
+
+	case *btpb.RowFilter_ApplyLabelTransformer:
+		alf := fpb.(*btpb.RowFilter_ApplyLabelTransformer)
+		l := alf.ApplyLabelTransformer
+		al := bigtable.LabelFilter(l)
+		f = &al
+	default:
```
enocom, line 331: Since you're returning `f` below which would be `nil` if any of these case statements don't match, you can remove this branch OR just put a return in each case statement and remove the `return f` below. I personally find returning out of case statements easier to read when it gets this long.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,626 @@
...
+// credentialsBundle implements credentials.Bundle interface
+// [See documentation for usage](https://pkg.go.dev/google.golang.org/grpc/credentials#Bundle).
+type credentialsBundle struct {
+	channel credentials.TransportCredentials
+	call    credentials.PerRPCCredentials
+}
+
+// TransportCredentials gets the channel credentials as TransportCredentials
+func (c credentialsBundle) TransportCredentials() credentials.TransportCredentials {
+	return c.channel
```
enocom, line 356: Should this be a pointer receiver? As written all the methods here will make a copy of channel and call, which might be OK, but just want to confirm.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,626 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		// TODO(developer): Determine how to pass this call option back to caller
+		option.WithTokenSource(c.TokenSource)
+
+		return nil, nil
+	}
+	return creds, nil
```
enocom, line 476: You might consider returning out of the case statements above just because this switch statement is a little long. Up to you.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,626 @@
...
+		return nil, stat.Error(codes.InvalidArgument,
+			fmt.Sprintf("%s: ClientID does not exist", logLabel))
+	}
+
+	// this closes every ClientConn in the pool.
+	btc.c.Close()
+	delete(s.clientIDs, clientID)
+
+	resp := &pb.RemoveClientResponse{}
+	return resp, nil
```
enocom, line 551:
``` golang
return &pb.RemoveClientResponse{}, nil
```

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,201 @@
...
+	lis    *bufconn.Listener
+	client *pb.CloudBigtableV2TestProxyClient
+)
+
+func bufDialer(context.Context, string) (net.Conn, error) {
+	return lis.Dial()
+}
+
+// helper function to populate the in-memory BT table.
+// TODO(developer): Expose this (and bttest) as suite of test tools
```
enocom, line 55: Does this mean the test proxy maintainers should do this? Or someone else? TODO(developer) usually means non-google user of the package. Perhaps just `TODO:`?

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,201 @@
...
+func populateTable(bts *bttest.Server) error {
+	ctx := context.Background()
+
+	conn, err := grpc.Dial(bts.Addr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
+	if err != nil {
+		return fmt.Errorf("testproxy setup: can't dial inmem Bigtable server: %v", err)
+	}
+	defer conn.Close()
+
+	adminClient, err := bigtable.NewAdminClient(ctx, "client", "instance", option.WithGRPCConn(conn), option.WithGRPCDialOption(grpc.WithBlock()))
```
enocom, line 65: Wrapping this under 100 lines would be nice. Ditto below.

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,201 @@
...
+		return fmt.Errorf("testproxy setup: can't create AdminClient: %v", err)
+	}
+	defer adminClient.Close()
+
+	if err := adminClient.CreateTable(ctx, tableName); err != nil {
+		return fmt.Errorf("testproxy setup: can't create table: %v", err)
+	}
+
+	// Create column families
+	count := 3
```
enocom, line 76: Why three? A comment would be nice.

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// 2) Create the test proxy client
+	conn2, err := grpc.DialContext(ctx, lis.Addr().String(),
+		grpc.WithContextDialer(bufDialer),
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		log.Fatalf("testproxy setup: failed to dial testproxy: %v", err)
+	}
+	defer conn2.Close()
+	c := pb.NewCloudBigtableV2TestProxyClient(conn2)
```
enocom, line 152:
``` golang
c := pb.NewCloudBigtableV2TestProxyClient(conn2) 
_, err = c.CreateClient(ctx, req)
```

enocom (changes requested):
Some minor comments and then I think this is ready to merge.

bigtable/internal/testproxy/proxy.go
telpirion: The compiler doesn't like the missing return statement, so I just deleted the `default` branch.

bigtable/internal/testproxy/proxy.go
telpirion: No, the interface requires a value receiver.

bigtable/internal/testproxy/proxy.go
telpirion: The compiler doesn't like it if I remove the final return statement.

bigtable/internal/testproxy/proxy_test.go
telpirion: I had intended to leave this comment for future maintainers ... but this is an inappropriate place for such a reminder.

Removed.

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy_test.go
telpirion: Done.

bigtable/internal/testproxy/proxy_test.go
telpirion: Comment added.

bigtable/internal/testproxy/proxy_test.go
telpirion: I simplified some of this code, but I still require the assignment (since `client` is declared in the outer scope and I can't assign directly to `client`).

bigtable/internal/testproxy/proxy.go
enocom: Change the switch statement above to `switch mut := mpb.Mutation.(type) {` and you'll be able to remove these type assertions.

bigtable/internal/testproxy/proxy.go
```diff
@@ -0,0 +1,622 @@
...
+		}
+	}
+	return m
+}
+
+// filterFromProto translates a Bigtable v2.RowFilter object into a Bigtable
+// Filter object.
+func filterFromProto(rfPb *btpb.RowFilter) *bigtable.Filter {
+	var f *bigtable.Filter
+	switch fpb := rfPb.Filter; fpb.(type) {
```
enocom, line 181:
``` golang
switch fpb := rfPb.Filter.(type) {
```

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	_, err := (*client).CreateClient(ctx, req)
+	if err != nil {
+		t.Fatalf("testproxy test: CreateClient() failed: %v", err)
+	}
+
+	t.Log("testproxy test: client created successfully in test proxy")
+
+	_, err = (*client).RemoveClient(ctx, &pb.RemoveClientRequest{
```
enocom, line 194: client.RemoveClient

bigtable/internal/testproxy/proxy_test.go
```diff
@@ -0,0 +1,202 @@
...
+	ctx := context.Background()
+
+	req := &pb.CreateClientRequest{
+		ClientId:   cid,
+		ProjectId:  "client",
+		DataTarget: bigtableAddress,
+		InstanceId: "instance",
+	}
+
+	_, err := (*client).CreateClient(ctx, req)
```
enocom, line 187: client.CreateClient

enocom (approved):

bigtable/internal/testproxy/proxy.go
telpirion: Done.

bigtable/internal/testproxy/proxy_test.go
telpirion: Done.

bigtable/internal/testproxy/proxy_test.go
telpirion: Done.

PR #6570, opened 2022-08-25
feat(bigquery): add CreateIfNotExists helpers and related errors

### LINKED ISSUE #2808: bigquery: unpredictable or poorly documented errors for checking if a table exists

When using to bigquery package it is important to check if a table exists or not before starting to insert rows into it. To do this, the doc string for dataset.Table() recommends:

```
Table creates a handle to a BigQuery table in the dataset. To determine if a table exists, call Table.Metadata. If the table does not already exist, use Table.Create to create it.
```

If you call Metadata() on a table pointer that is returned from dataset.Table() you get a Metadata pointer and an error. If that table doesn't exist the metadata pointer is nil and the error reads: `googleapi: Error 404: Not found: Table security-263001:osquery.macos_kextstat, notFound`

However, how can I verify that the error is because of the table not being found, and not something like a dropped network connection? 

 There should be an error defined by the bigquery client library such as `bigquery.TableNotFound`

The only option I see here is to string search for "404" or "Not found"

**Additional context**

alvarowolfx:
This adds the helpers methods `CreateIfNotExists` to the `Dataset` and `Table` structs, to replicate the same semantics of `CREATE IF NOT EXISTS` from SQL. Also adds `ErrTableNotFound` and `ErrDatasetNotFound` that can be returned from the `Metadata` call, to avoid having to look underneath the library for `googleapi.Error`s. 

Resolves #2808

### DIFF
--- a/bigquery/dataset.go
+++ b/bigquery/dataset.go
@@ -18,12 +18,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 	"time"
 
 	"cloud.google.com/go/internal/optional"
 	"cloud.google.com/go/internal/trace"
 	bq "google.golang.org/api/bigquery/v2"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 )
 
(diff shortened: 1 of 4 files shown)


codyoss:
@alvarowolfx can this be closed?

PR #6572, opened 2022-08-26
chore(internal/gapicgen): incorporate aliasfix config into gapicgen 

codyoss:
As we migrate types into google-cloud-go it will be important to keep track of what has been migrated. Aliasfix already needs to keep track of this too so it knows which imports to re-write. Because of this I exported this config mapping so the same config can be used across the two tools.

The migration process will have three stages per library:
1. No migrated: Everything as it is today.
2. In progress: New types have been generated in google-cloud-go.
3. Migrated: Aliases have been created referencing the new types and the GAPIC layer has been re-written in terms of the types in google-cloud-go.

These changes in gapicgen handle everything on the google-cloud-go side for use as we flip config flags. The process for creating aliases and maintaining them for a period of time is out-of-band for this flow. This is because aliases need to reference released google-cloud-go code. I will have another small tool that we can run as a cron to keep aliases up-to-date once a week soon.

As written these code has no changes until we start flipping the migration statues.


codyoss:
The following PRs were generated by running the following command with different config settings: `go run cloud.google.com/go/internal/gapicgen/cmd/genbot -local -gocloud-dir /path/to/google-cloud-go -only-gapics -gapic=cloud.google.com/go/secretmanager/apiv1beta1`

In progress example: https://github.com/codyoss/google-cloud-go/pull/2
Migrated example: https://github.com/codyoss/google-cloud-go/pull/3

internal/gapicgen/generator/genproto.go
```diff
@@ -133,10 +133,13 @@ func (g *GenprotoGenerator) Regen(ctx context.Context) error {
 		grpc := !noGRPC[pkg]
 		pk := pkg
 		fn := fileNames
-		grp.Go(func() error {
-			log.Println("running protoc on", pk)
-			return g.protoc(fn, grpc)
-		})
+
+		if !isMigrated(pkg) {
```
quartzmo, line 137: What do you think about also logging a statement for the `isMigrated` `true` case, to indicate that `protoc` is not running due to aliasgen.

internal/gapicgen/generator/gapics.go
```diff
@@ -311,19 +311,30 @@ func (g *GapicGenerator) microgen(conf *MicrogenConfig) error {
...
 		// override the go_package option. Applied to both the protobuf/gRPC
 		// generated code, and to notify the GAPIC generator of the new
 		// import path used to reference those stubs.
-		stubPkg := filepath.Join(conf.ImportPath, conf.StubsDir)
+		stubPkg := filepath.Join(conf.ImportPath, stubsDir)
 		for _, f := range protoFiles {
 			f = strings.TrimPrefix(f, g.googleapisDir+"/")
-			rerouteGoPkg := fmt.Sprintf("M%s=%s;%s", f, stubPkg, conf.Pkg)
-			args = append(args, "--go_opt="+rerouteGoPkg, "--go_gapic_opt="+rerouteGoPkg)
+			if conf.ImportPath == "cloud.google.com/go/storage/internal/apiv2" {
```
quartzmo, line 325: Maybe a comment to explain what's special about storage?

internal/gapicgen/generator/genproto.go
codyoss: done.

internal/gapicgen/generator/gapics.go
codyoss: done

internal/gapicgen/generator/genproto.go
```diff
@@ -139,6 +139,8 @@ func (g *GenprotoGenerator) Regen(ctx context.Context) error {
 				log.Println("running protoc on", pk)
 				return g.protoc(fn, grpc)
 			})
+		} else {
+			log.Panicf("skipping, %q has been migrated", pkg)
```
quartzmo, line 143: Why panic here?

codyoss: whoa, that was a mistake. Auto-complete failed me there. I will update rn

PR #6573, opened 2022-08-26
docs: add client options endpoint to doc.go

quartzmo:
This doc addition is required for an upcoming PR for snippets comments that:

> should include a link to documentation that describes how to set a regional endpoint for that language's service clients. For instance, the .NET link is: https://cloud.google.com/dotnet/docs/reference/help/client-configuration#endpoint.

See b/240118976

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -22,19 +22,22 @@ of sub-packages.
 All clients in sub-packages are configurable via client options. These options are
 described here: https://godoc.org/google.golang.org/api/option.
 
-## Endpoint
+## Endpoint Override
 
-Endpoint configuration is used to specify the API endpoint to which requests are
-sent. It is primarily used for services that support or require regional endpoints.
-For example, the Vertex AI service recommends that you set the location with the
-features you want that is closest to your physical location or the physical location
-of your users. There is no global endpoint for Vertex AI. See
+Endpoint configuration is used to specify the URL to which requests are
+sent. It is used for services that support or require regional endpoints, as well
+as for other use cases such as [testing against fake
+servers](https://github.com/googleapis/google-cloud-go/blob/main/testing.md#testing-grpc-services-using-fakes).
+
+For example, the Vertex AI service recommends that you configure the endpoint to the
+location with the features you want that is closest to your physical location or the
+location of your users. There is no global endpoint for Vertex AI. See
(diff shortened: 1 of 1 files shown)


doc.go
```diff
@@ -22,6 +22,21 @@ of sub-packages.
 All clients in sub-packages are configurable via client options. These options are
 described here: https://godoc.org/google.golang.org/api/option.
 
+## Endpoint
```
codyoss, line 25: nit: I think we should call this section Endpoint Override.

doc.go
```diff
@@ -22,6 +22,21 @@ of sub-packages.
...
+sent. It is primarily used for services that support or require regional endpoints.
+For example, the Vertex AI service recommends that you set the location with the
+features you want that is closest to your physical location or the physical location
+of your users. There is no global endpoint for Vertex AI. See
+[Vertex AI - Locations](https://cloud.google.com/vertex-ai/docs/general/locations)
+for more details. The following example demonstrates configuring a Vertex AI client
+with a regional endpoint:
+
+	ctx := context.Background()
+	endpoint := "us-central1-aiplatform.googleapis.com"
```
codyoss, line 37: I believe gRPC clients require the port.

doc.go
```diff
@@ -22,6 +22,21 @@ of sub-packages.
...
+	client, err := aiplatform.NewDatasetClient(ctx, option.WithEndpoint(endpoint))
+
```
codyoss, line 39: The other use case where I see people using endpoint overrides is when they are testing against fake servers. Do you think we should mention this at all here as another potential use-case? There is an example of how this is done here: https://github.com/googleapis/google-cloud-go/blob/main/testing.md#testing-grpc-services-using-fakes

quartzmo: Added, thanks!

doc.go
quartzmo: Updated, thanks!

doc.go
quartzmo: Fixed. Ooooops! Should I add this to the [WithEndpoint](https://pkg.go.dev/google.golang.org/api/option#WithEndpoint) docs??? Right now they say nothing about the `url` param.

> // WithEndpoint returns a ClientOption that overrides the default endpoint
> // to be used for a service.

codyoss (approved):
Thanks!

PR #6575, opened 2022-08-26
chore: exclude 1.15 from go build in kokoro/presubmit.sh

quartzmo:
Exclude go1.15 due to the go1.16 build constraint in `third_party` to fix the recurring kokoro build error:

```sh
package cloud.google.com/go/internal/godocfx
	imports cloud.google.com/go/third_party/go/doc: build constraints exclude all Go files in /tmpfs/src/google-cloud-go/third_party/go/doc
+ '[' -f emulator_test.sh ']'
+ cat sponge_log.log
+ go-junit-report -set-exit-code
+ exit_code=1
```


quartzmo:
After adding the empty file to package `doc`, `go1.15 build` then encountered the following error:

```

```

Therefore, I increased the build constraint in `third_party/pkgsite/print_type.go` to `1.16` as well, since that file has dependencies on `doc` files that were already constrained to `1.16`.

codyoss (approved):
LGTM, just update squashed message when merging

PR #6596, opened 2022-08-31
feat(bigquery): add support for explicit query parameter type

### LINKED ISSUE #4704: bigquery: big.Rat limited to 9 digits
**Client**

BigQuery

**Environment**

N/A

**Go Environment**

```
go version go1.16.6 darwin/arm64

GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/Users/thomas/Library/Caches/go-build"
GOENV="/Users/thomas/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/thomas/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/thomas/go"
GOPRIVATE=""
GOROOT="/opt/homebrew/Cellar/go/1.16.6/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/opt/homebrew/Cellar/go/1.16.6/libexec/pkg/tool/darwin_arm64"
GOVCS=""
GOVERSION="go1.16.6"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/dev/null"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/lv/qbz3lh7j6gdc43dd1p22bcc40000gn/T/go-build541126118=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

**Code**

https://pkg.go.dev/cloud.google.com/go/bigquery#InferSchema documents that `big.Rat` types are limited to 9 digits after then decimal, this also seems to be the case for parameters to query. However BigQuery supports much more precision (up to 76, https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#numeric-type). It would be nice to able to query using these high precision numbers.

I guess that the parameters are send as strings so you have to make a decision, but could the limit be increased? Alternatively can we send our own string (with our own defined precision) as number?

Also related: https://github.com/golang/go/issues/47911: Support precision of -1 in Rat.FloatString (that will use the required number of digits to represent the number).

**Expected behavior**

I expected large precision numbers to be support.

**Actual behavior**

Large numbers are limited to 9 digits after the decimal.

alvarowolfx:
Allow users to pass query parameter with an explicit data types instead of relying just on reflection of Go types. Right now the data type that is affect by this change is `*big.Rat`, which before was only converted to `NUMERIC` with 9 digits, and now can be send as a `BIGNUMERIC`.

This adds two methods:
* `NewScalarQueryParameter(name string, paramType FieldType, value interface{}) QueryParameter`
* `NewArrayQueryParameter(name string, paramType FieldType, value interface{}) QueryParameter`

Example usage:
```
ctx := context.Background()
client, err := bigquery.NewClient(ctx, projectID)
if err != nil {
        return fmt.Errorf("bigquery.NewClient: %v", err)
}
defer client.Close()

rat := big.NewRat(12345, 10e10)
q := client.Query("SELECT @bigval, @array")
q.Parameters = []bigquery.QueryParameter{
        {
                Value: NewScalarQueryParameter("bigval", "BIGNUMERIC", rat)
        },
        {
                Value: NewArrayQueryParameter("array", "STRING", []string{"a","b","c"}
        },
}
```

I'm still evaluating if we need something like `NewStructQueryParameter` to pass a list of QueryParameters with explicit data types for each field. 

Most of this was inspired by the Python implementation that has the same kind of methods for creating query parameters.
> see: https://cloud.google.com/bigquery/docs/parameterized-queries#python_1

Resolves #4704


alvarowolfx:
I changed to a slight different approach, having some specific struct and interface that can be used explicitly or with the current QueryParameter struct.


```
This adds two methods:

NewScalarQueryParameter(name string, paramType FieldType, value interface{}) QueryParameter
NewArrayQueryParameter(name string, paramType FieldType, value interface{}) QueryParameter
Example usage:

ctx := context.Background()
client, err := bigquery.NewClient(ctx, projectID)
if err != nil {
        return fmt.Errorf("bigquery.NewClient: %v", err)
}
defer client.Close()

rat := big.NewRat(12345, 10e10)
q := client.Query("SELECT @bigval, @array, @struct")
q.Parameters = []bigquery.QueryParameter{
        {
                Value: ScalarQueryParameter{Name: "bigval", Type: "BIGNUMERIC", Value: rat}
        },
        ArrayQueryParameter{Name: "array", Value: []string{"a","b","c"}, Type: ScalarQueryParameter{Type: "STRING"}}.QueryParameter(),
        StructQueryParameter{
	        Name: "val",
	        Value: map[string]ParameterValue{
		        "Timestamp": ScalarQueryParameter{
			        Value: ts,
		        },
		        "BigNumericArray": ArrayQueryParameter{
			        Value: []string{
				        BigNumericString(bigRat),
				        BigNumericString(rat),
			        },
		        },
		        "SubStruct": StructQueryParameter{
			        Value: map[string]ParameterValue{
				        "String": ScalarQueryParameter{
					        Value: "c",
				        },
			        },
		        },
	        },
	        Schema: map[string]ParameterType{
		        "Timestamp": ScalarQueryParameter{
			        Type: "TIMESTAMP",
		        },
		        "BigNumericArray": ArrayQueryParameter{
			        Type: ScalarQueryParameter{
				        Type: "BIGNUMERIC",
			        },
		        },
		        "SubStruct": StructQueryParameter{
			        Schema: map[string]ParameterType{
				        "String": ScalarQueryParameter{
					        Type: "STRING",
				        },
			        },
		        },
	        },
        }.QueryParameter()
}
```

bigquery/params.go
```diff
@@ -127,14 +127,179 @@ type QueryParameter struct {
...
+// ParameterType represents an explicit typed parameter to a query.
+// When using those interfaces, the priority is to avoid type inference.
+type ParameterType interface {
+	parameterType() *bq.QueryParameterType
+}
+
+// ParameterValue represents an explicit parameter value to a query
+// When using those interfaces, the priority is to avoid type inference.
+type ParameterValue interface {
+	parameterValue() *bq.QueryParameterValue
```
shollyman, line 143: same here re: unexported methods

bigquery/params.go
```diff
@@ -127,14 +127,179 @@ type QueryParameter struct {
...
 	// string.
 	Value interface{}
+
+	explicitParameter Parameter
 }
 
-func (p QueryParameter) toBQ() (*bq.QueryParameter, error) {
+// ParameterType represents an explicit typed parameter to a query.
+// When using those interfaces, the priority is to avoid type inference.
+type ParameterType interface {
```
shollyman, line 136: It seems a bit odd to have an exported interface with no exported methods.

Historically we've tried to avoid leaking discovery types through as we don't control their implementation, and thus our guarantees about breaking changes etc.

bigquery/params.go
```diff
@@ -127,14 +127,179 @@ type QueryParameter struct {
...
+}
+
+// StructQueryParameter represents an explicit complex type parameter with a given schema.
+type StructQueryParameter struct {
+	// Name is used for named parameter mode.
+	// It must match the name in the query case-insensitively.
+	Name string
+
+	// Schema complex struct value passed as a map of field names and ParameterTypes.
+	Schema map[string]ParameterType
```
shollyman, line 203: leaf fields of a struct don't have to have names, so using a map will cause issues in those cases.

https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#struct_type

bigquery/params.go
alvarowolfx: Yeah, I didn't like that too, but the ParameterType and ParameterValue interfaces both have methods that rely on the internal `bq.QueryParameterType` and `bq.QueryParameterValue`, which I was trying to avoid letting users know about it too much.

bigquery/params.go
alvarowolfx: I tried to add support for leaf fields without a name, but I could not find how to do that properly. I'm getting this error from the backend:

```
```

I was looking here on the API docs and they don't provide an example on how that can be done.
https://cloud.google.com/bigquery/docs/reference/rest/v2/QueryParameter#queryparametervalue

I also noticed that the Python API is based on the equivalent to a Go Map, that's why I started with that idea here. 
https://cloud.google.com/bigquery/docs/samples/bigquery-query-params-structs#bigquery_query_params_structs-python

bigquery/params.go
```diff
@@ -129,12 +139,57 @@ type QueryParameter struct {
 	Value interface{}
 }
 
-func (p QueryParameter) toBQ() (*bq.QueryParameter, error) {
+// QueryParameterValue is a go type for representing a explicit typed BigQuery Query Parameter.
+type QueryParameterValue struct {
+	Type        StandardSQLDataType
```
shollyman, line 144: Document these fields as appropriate.

bigquery/params.go
```diff
@@ -129,12 +139,57 @@ type QueryParameter struct {
 	Value interface{}
 }
 
-func (p QueryParameter) toBQ() (*bq.QueryParameter, error) {
+// QueryParameterValue is a go type for representing a explicit typed BigQuery Query Parameter.
```
shollyman, line 142: I'm still on the fence about the name due to the collision with the API, but it's a reasonable name otherwise.  

Should we document the behavior for nested types as part of the type description?  Specifically that you need to fully specify types and values for complex params?

bigquery/standardsql.go
```diff
@@ -63,6 +63,23 @@ func (ssdt *StandardSQLDataType) toBQ() (*bq.StandardSqlDataType, error) {
 	return bqdt, nil
 }
 
+func (ssdt StandardSQLDataType) toBQQueryType() *bq.QueryParameterType {
```
shollyman, line 66: ToBQParamType maybe?

bigquery/params.go
```diff
@@ -306,14 +363,13 @@ func paramValue(v reflect.Value) (*bq.QueryParameterValue, error) {
 		return res, nil
 
 	case typeOfRat:
-		// big.Rat types don't communicate scale or precision, so we cannot
```
shollyman, line 309: HOORAY

bigquery/integration_test.go
```diff
@@ -1916,6 +1917,147 @@ func TestIntegration_QueryParameters(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+			[]Value{[]Value{"a", "b"}},
+			[]interface{}{"a", "b"},
+		},
+		{
+			"SELECT @val",
```
shollyman, line 1958: thanks for exercising a complex representation!

bigquery/params.go
```diff
@@ -116,6 +117,15 @@ type QueryParameter struct {
 	// For scalar values, you can supply the Null types within this library
 	// to send the appropriate NULL values (e.g. NullInt64, NullString, etc).
 	//
+	// For values with a explicit data type, *QueryParameterValue can be used.
```
shollyman, line 120: maybe "to specify query parameters explicitly rather by inference" instead?

bigquery/params.go
alvarowolfx: Humm, I need to bring back this comment, because still if you pass a `*big.Rat`, it is going to be converted to a NUMERIC type. For BIGNUMERIC you have to use the new `QueryParameterValue`

shollyman (changes requested):
I think the only work remaining is to flesh out the documentation for QueryParameterValue (and it's fields) and ship it.  Approving contingent on that.

shollyman (approved):

shollyman (approved):
Looks good, and I like the explicit examples of usage.  Thanks for putting this together!

PR #6598, opened 2022-09-01
fix(spanner): pass userAgent to cloud spanner requests

rahul2393:


spanner/client.go
```diff
@@ -134,6 +134,11 @@ type ClientConfig struct {
 	// override the default values.
 	CallOptions *vkit.CallOptions
 
+	// UserAgent is the prefix to the user agent header. This is used to supply information
+	// such as application name or partner tool.
+	// Recommended format: ``application-or-tool-ID/major.minor.version``.
+	UserAgent string
```
olavloite, line 140: This is now an exported field that could be used by anyone to set any random value. Would it be possible to either:
1. Somehow make this unexported? I know that it is called from a different package, so literally making it unexported without any other changes won't work. But would it be possible to in some way or another make this less accessible?
2. Or otherwise: At least mark this as an internal option that users should not set, and have a fixed list of values that are allowed to be set. Now anyone could set any value, and it does not in any way tell you that users should normally not do this.

rahul2393: What will be the issue if user set custom values? since this will just append to the existing header 
example if userAgent is "go-sql-spanner/0.1" the header will look like 

```
gl-go/1.19.0 go-sql-spanner/0.1 gapic/1.36.0 gax/2.4.0 grpc/1.48.0
```

So no breaking changes to existing metrics and tracking, from driver and ORMs we will not expose this field for sure.

rahul2393: In python also we are not adding any validation/check.
Example for Django we pass like this: https://github.com/googleapis/python-spanner-django/pull/140
And in the client lib there are no validations.

olavloite: Are you sure that this does not change anything? The handwritten Spanner client will normally also call `setClientInfo` with `gccl/<version>`. That seems to be overwritten if the application has set a user agent. At least, in your example above the `gccl/<version>` part seems to be missing, which means that it is not just appending.

And even then I would suggest that we at least document `UserAgent` as an internal field. There's nothing that users can do with it, and it might confuse them if they can set it as if it is the same as any other field, but they won't see it turn up anywhere.

hengfengli:
The userAgent will be passed as a ClientOption in: https://github.com/googleapis/go-sql-spanner/blob/6d164d7dc7ac109735dacd1e146d5c16b53b5922/driver.go#L227-L235

The user ClientOption will override the default one `fmt.Sprintf("spanner-go/v%s", internal.Version)`: https://github.com/googleapis/google-cloud-go/blob/66f36d49290969f4b9a59b097271ab9eb0851b02/spanner/client.go#L238-L247

I wonder why this is not working?

spanner/client.go
rahul2393: Make sense, please take a look

rahul2393:
@hengfengli Yes the override is working fine, but x-goog-api-client is not using the ClientOptions, and internally we use the header value for this key for tracking requests.

olavloite (approved):

Issue #6595, opened 2022-08-31
bigquery/storage/write: getting error stream.Recv(): rpc error: code = Unavailable desc = the connection is draining


Av1shay:
**Client**

BigQueryWriteClient (cloud.google.com/go/bigquery v1.32.0)

**Environment**

Compute VM instances in GCP

**Go Environment**

1.18

**Code**

```go
package main

func main() {
    // ...

    lenRows := len(rows)

    // create chunks of chunkSize size to not send too many rows at once
    chunkSize := 15000
    chunks := make([]interface{}, 0, int(math.Ceil(float64(lenRows)/float64(chunkSize))))
    for i := 0; i < lenRows; i += chunkSize {
        end := i + chunkSize
	if end > lenRows {
	    end = lenRows
	}
	chunks = append(chunks, rows[i:end])
    }

     stream, err := bigqueryWriteClient.AppendRows(ctx)
     if err != nil {
         log.Fatalf("AppendRows: %s", err)
    }
    rowProto, ok := rows[0].(protoreflect.ProtoMessage)
    if !ok {
        log.Fatalf("row type suppose to be <protoreflect.ProtoMessage>, got: %T", rows[0])
    }
    descriptor, err := adapt.NormalizeDescriptor(rowProto.ProtoReflect().Descriptor())
    if err != nil {
        log.Fatalf("NormalizeDescriptor(): %s", err)
    }
    streamName := fmt.Sprintf("projects/%s/datasets/%s/tables/%s/streams/_default", projectID, datasetID, tableID)

    go func() {
        err := sendToStream(stream, streamName, descriptor, chunks)
        if err != nil {
             log.Fatalf(ctx, "sendToStream(): %s", err)
        }
    }()

    for {
        _, err := stream.Recv()
	if err == io.EOF {
	    break
        }
        if err != nil {
            log.Fatalf("stream.Recv(): %s", err)
        }
    }
}

func sendToStream(stream storagepb.BigQueryWrite_AppendRowsClient, streamName string, descriptor *descriptorpb.DescriptorProto, dataChunks []interface{}) error {
    streamReqs, err := buildStreamRequests(streamName, descriptor, dataChunks)
    if err != nil {
        return err
    }

    for _, req := range streamReqs {
        if err := stream.Send(req); err != nil {
	    return fmt.Errorf("stream.Send(): %s", err)
	}
    }

    return stream.CloseSend()
}

[...]

PR #6609, opened 2022-09-02
feat(bigquery/storage/managedwriter): augment reconnection logic

shollyman:
This PR augments the reconnection logic to include the grpc transport
stream drain error as a condition where we should force reconnect,
rather the waiting for the io.EOF of the connection fully closing.

Related: https://github.com/googleapis/google-cloud-go/issues/6595


alvarowolfx (approved):
LGTM

PR #6611, opened 2022-09-02
build(release-please): sentence case CHANGELOG

bcoe:
This plugin will automatically sentence case the first word in the generated CHANGELOG. This makes the CHANGELOG look more consistent.

Here's an example where case is varied:

Screen Shot 2022-09-02 at 3 51 16 PM


quartzmo:
What do you think about downcasing instead of capitalizing these sentences? I feel like the majority of the commit messages that are used to build changelogs are lower case (per [conventionalcommits.org](https://www.conventionalcommits.org/)), therefore that seems to me to be the original form.

quartzmo (approved):

quartzmo:
I get it now, the current PR activates sentence case for every changelog... in every repo? As long as it's consistent, SGTM

quartzmo:
@codyoss PTAL

codyoss (approved):

PR #6616, opened 2022-09-06
feat(secretmanager): start generating beta stubs

codyoss:
Generated by running: 
`go run cloud.google.com/go/internal/gapicgen/cmd/genbot -local -only-gapics -gocloud-dir=/Users/codyoss/oss/google-cloud-go -gapic=cloud.google.com/go/secretmanager/apiv1beta1`


codyoss:
It looks like this one might be a bad one to start with. The protos no longer exist. We should likely remove this client I think

PR #6625, opened 2022-09-07
chore(internal/aliasgen): clean dir and fix import bug

codyoss:


quartzmo (approved):
Cool to see how to remove the contents of a directory without deleting the directory itself.

PR #6627, opened 2022-09-07
chore: remove mock_test files

codyoss:
These files are no longer generated and they would need to be updated during our alias migration if they were still around. I don't think they provide too much value so they are being removed.


tritone (commented):
I see that storage is still used and you restored it in a later commit. Should we be working to remove this dep?

noahdietz:
> I see that storage is still used and you restored it in a later commit. Should we be working to remove this dep?

That was a hand-written file, whereas the ones @codyoss is trying to delete are *generated* files, previously created by the monolith generator. The move to microgenerator did not carry this feature over, so the files in this PR haven't been updated in a while.

tl;dr: Storage mock_test.go is fine as it is.

tritone (approved):

PR #6637, opened 2022-09-08
chore(internal/aliasfix): properly truncate and backup files

codyoss:
- Use os.Create to truncate existing file if needed
- Use named imports to capture error
- Create a bak incase anything goes wrong. We don't want to delete peoples code :)
- Bubble up error if any from WalkDir


noahdietz (approved):
Cool, good idea on the bak

PR #6639, opened 2022-09-08
chore(storage): add multi-transport test runner

noahdietz:
Introduces a `multiTransportTest` runner that executes a given test function with each transport-specific client library, HTTP/JSON and gRPC. This also migrates a few tests that worked OOTB as PoC. Each test should be migrated individually so that any fixes necessary can be made in the same PR (there will be some, I already experienced a few). 

Removes an unnecessary `-` suffix from the gRPC test bucket name prefix used to create the `uid` generator for gRPC.

Switches `uid` generators into `Short` mode to keep names under the 63 character bucket name limit.


storage/integration_test.go
```diff
@@ -238,6 +240,33 @@ func testConfigGRPC(ctx context.Context, t *testing.T) (gc *Client) {
...
+// test function with the sub-test instance, the context it was given, the bucket
+// name to use directly, or as a prefix, and the client to use.
+func multiTransportTest(ctx context.Context, t *testing.T, test func(*testing.T, context.Context, string, *Client)) {
+	for transport, client := range initTransportClients(ctx, t) {
+		t.Run(transport, func(t *testing.T) {
+			defer client.Close()
+
+			bucket := bucketName
+			if transport == "grpc" {
+				bucket = grpcBucketName
```
tritone, line 262: I don't think this is going to work universally. Some tests use the global var bucketName but others need to be able to create/delete multiple buckets inside the test. Perhaps we can have the test supply a slice of bucket names and then we prepend the gRPC prefix when running on gRPC.

tritone (commented):
Generally seems like a good approach!

storage/integration_test.go
noahdietz: The comment I added for this method indicates we can just use this bucket name as a prefix for bucket names to be created. Would that work? Makes the names kind of long, but we don't actually need to read them.

tritone: Ah yeah, probably fine -- let's call it bucketPrefix or something then for clarity.

noahdietz: We can name the parameter "bucketPrefix" in the individual tests that are provided as the test function and need to use it as such, but there are also tests that just want to use it for the bucket it is, so in those it would be named "bucket". Perhaps two parameters would be better and tests can decide which they'd like to use haha

noahdietz: I changed it so that the test function must accept both `bucket string, toCreate string`, but in practice they can just do `bucket string, _ string` or `_ string, toCreate string`. The test runner will generate a new `toCreate` bucket name using the proper transport UID space for each test.

tritone: This still seems kind of needlessly confusing to me -- I think all the tests need to know is whether to add the gRPC prefix to their bucket names or not, right? We can just provide all tests with a prefix (would be empty for HTTP) and they can prepend it to the name of any bucket they are using or creating.

noahdietz: The point is to avoid doing any prefixing or extra logic in the tests - just accept a bucket name and do the test with it. We'd need to write that prefixing logic in every test otherwise. Since bucket names must differ between gRPC and HTTP for the time being, this is the simplest way to abstract that requirement from the test logic.

Another alternative is to have the tests pass in a `newBucket bool` which indicates that the test will create a new bucket and it wants a fresh bucket name in the `bucket string` param rather than the pre-populated "global test" bucket.

noahdietz: I switched it to `bucket string, prefix string` so we can do as you suggest - either the test uses the prepopulated `bucket`, or they use the `prefix` to create their new bucket. It's six in one hand and half a dozen in the other.

tritone (approved):

PR #6643, opened 2022-09-09
refactor(bigquery/storage/managedwriter): internal context refactor

shollyman:
This PR lays the groundwork for enabling retry of append requests on
the underlying stream.  It moves the request context into pending write,
and simplifies the signature of internal methods by using the retained
context rather than propagating it via arguments.

We also add an internal attemptCount property to the pendingWrite, but
do not use it meaningfully in the PR.  It will be used to evaluate retry
logic in future PRs.

Towards internal b/245963141


bigquery/storage/managedwriter/appendresult.go
```diff
@@ -174,7 +180,9 @@ func newPendingWrite(appends [][]byte) *pendingWrite {
 				},
 			},
 		},
-		result: newAppendResult(appends),
+		result:       newAppendResult(appends),
+		attemptCount: 1,
```
alvarowolfx, line 184: This `attemptCount` attribute is not being used anywhere yet. Is this for the future `AppendRetryer` interface ?

alvarowolfx (commented):
In general it looks good to me, just added a minor comment about the `attemptCount` attribute.

bigquery/storage/managedwriter/appendresult.go
shollyman: Yes, the top description notes it's being added but not used yet.

alvarowolfx (approved):

Issue #6651, opened 2022-09-12
bigquery: Timestamp is not parsed correctly


whs:
**Client**

BigQuery

**Environment**

Local

**Go Environment**

```
$ go version go1.19 linux/amd64
go version go1.19 linux/amd64
$ go env
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/whs/.cache/go-build"
GOENV="/home/whs/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/whs/go/pkg/mod"
GONOPROXY="git.wndv.co"
GONOSUMDB="git.wndv.co"
GOOS="linux"
GOPATH="/home/whs/go"
GOPRIVATE="<private>"
GOPROXY="<private>,direct"
GOROOT="/usr/lib/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.19"
GCCGO="gccgo"
GOAMD64="v3"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/whs/<private>/go.mod"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build3667506472=/tmp/go-build -gno-record-gcc-switches"
```

**Code**

e.g.
```go
package main

func getQuery(job *bigquery.Job) string {
	cfg, err := job.Config()
	if err != nil {
		panic("this block should never hit")
	}
}
```

**Expected behavior**

The error block is not hit

**Actual behavior**

time.ParseError

- Layout = {string} "2006-01-02 15:04:05.999999-07:00"
- Value = {string} "2020-10-15T00:00:00Z"

[...]

PR #6653, opened 2022-09-12
fix(bigquery): parse timestamp query parameter with RFC3339

alvarowolfx:
Timestamps on BigQuery have a [format similar to RFC3339](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#canonical_format_for_timestamp_literals). 

When creating a Query Job via the API, users can pass a timestamp with string `2020-10-15T00:00:00Z`, which is a valid parameter.  But if a given Job was created with a `TIMESTAMP` following that format and is read back using this library, it fails (as reported on https://github.com/googleapis/google-cloud-go/issues/6651 ) because we expect `TIMESTAMPS` to have only [this format](https://github.com/googleapis/google-cloud-go/blob/5de75ddd406c0f1506a3f5ab835c0cae1f5f4b7a/bigquery/params.go#L33): `"2006-01-02 15:04:05.999999-07:00"`

This PR also tries to parse `TIMESTAMPS` using Go `RFC3339Nano`, similarly on how the `civil.ParseDateTime` function tries multiple formats.

Resolves https://github.com/googleapis/google-cloud-go/issues/6651

### DIFF
--- a/bigquery/params.go
+++ b/bigquery/params.go
@@ -416,7 +416,15 @@ func convertParamValue(qval *bq.QueryParameterValue, qtype *bq.QueryParameterTyp
 		if isNullScalar(qval) {
 			return NullTimestamp{Valid: false}, nil
 		}
-		return time.Parse(timestampFormat, qval.Value)
+		t, err := time.Parse(timestampFormat, qval.Value)
+		if err != nil {
+			t, err = time.Parse(time.RFC3339Nano, qval.Value)
+			if err != nil {
+				return nil, err
+			}
+		}
+		return t, nil
+
 	case "DATETIME":
 		if isNullScalar(qval) {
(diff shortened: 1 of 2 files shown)


bigquery/params.go
```diff
@@ -416,7 +416,15 @@ func convertParamValue(qval *bq.QueryParameterValue, qtype *bq.QueryParameterTyp
 		if isNullScalar(qval) {
 			return NullTimestamp{Valid: false}, nil
 		}
-		return time.Parse(timestampFormat, qval.Value)
+		t, err := time.Parse(timestampFormat, qval.Value)
```
shollyman, line 419: Fallback for the win!

shollyman (approved):

PR #6656, opened 2022-09-12
fix(pubsub): remove unused AckResult map

hongalex:
`iterator.pendingAckResults` was supposed to store the pending ack results for exactly once delivery and shut them down on context cancellation. However, we don't do this since we wait for all messages to be acked/nacked/expired before shutting down the subscriber. I'm also removing this since the map currently has the possibility of creating a memory leak if messages are not properly acked/nacked.


pubsub/iterator.go
```diff
@@ -261,9 +256,6 @@ func (it *messageIterator) receive(maxToPull int32) ([]*Message, error) {
 	maxExt := time.Now().Add(it.po.maxExtension)
 	ackIDs := map[string]*AckResult{}
 	it.mu.Lock()
-	it.eoMu.RLock()
-	enableExactlyOnceDelivery := it.enableExactlyOnceDelivery
```
shollyman, line 265: This is the only bit that bears scrutiny.  Will not setting this affect any other behaviors?

hongalex: This was only for reading the field from the iterator object into a local variable. It was used for this function only.

shollyman (approved):

Issue #6657, opened 2022-09-13
bigquery/storage: ManagedStream should not hold context.Canceled and context.DeadlineExceeded error


tamayika:
**Client**

BigQuery/Storage

**Environment**

Any

**Go Environment**

Any

**Code**

```go
package main

func main() {
	/* initialization skipped */
	// this repro uses context.WithCancel(), but the same happens on context.WithTimeout() and context.DeadlineExceeded
	ctx, cancel := context.WithCancel(context.Background())
	go func() {
		// wait to ManagedStream.lockingAppend()
		time.Sleep(100 * time.Millisecond)
		cancel()
	}()
	_, err := stream.AppendRows(ctx, bytes)                 // context.Canceled is returned: expected
	_, err = stream.AppendRows(context.Background(), bytes) // context.Canceled is returned: unexpected
}
```

This problem is race condition, so please add `time.Sleep(1 * time.Second)` to https://github.com/googleapis/google-cloud-go/blob/main/bigquery/storage/managedwriter/managed_stream.go#L266

**Expected behavior**

ManagedStream does not hold context error.

**Actual behavior**

ManagedStream holds context error.

**Screenshots**

**Additional context**

shollyman:
Thanks for the report.  Yes, its an error to retain the append-level error as the stream error.  PR forthcoming.

PR #6659, opened 2022-09-13
fix(bigquery/storage/managedwriter): fix incorrect error retention

shollyman:
Failures for an individual append can get retained as the terminal error for the stream.  Redact the logic responsible, and add a test that validates the correct behavior using context-based errors.

Fixes: https://github.com/googleapis/google-cloud-go/issues/6657


alvarowolfx (approved):
LGTM

Issue #6612, opened 2022-09-02
Field marked as deprecated is being used by firebase trigger for cloud function


mogli91:
https://github.com/googleapis/google-cloud-go/blob/989cd0e8da8f50ab5d2353b93732b4b997f4c868/functions/metadata/metadata.go#L39

The docs going with the above structs say that the field `RawPath` is only being used by storage in combination with pubsub. Please mention that it is also used by Firebase Realtime Database Triggers (a trigger for cloud functions) to identify the resource (document) that triggered the call.

https://cloud.google.com/functions/docs/calling/realtime-database#functions_firebase_rtdb-go

quartzmo:
You wrote `RawType`, did you mean `RawPath`?

mogli91:
Indeed, thanks for pointing this out. I have updated the issue with `RawPath` instead of `RawType`. Apologies for the typo.

quartzmo:
@mogli91 PTAL at my edit in #6660, and feel free to suggest changes.

PR #6660, opened 2022-09-13
docs(functions): update metadata.Resource docs

quartzmo:
closes: #6612


mogli91:
I suggest the following edit:
It is also used when triggering Cloud Functions by [Firebase Realtime Database Triggers](https://cloud.google.com/functions/docs/calling/realtime-data).

The link points to an actual feature (Triggers for Cloud Functions) and not only an example. This is why I would change the description.

quartzmo:
@mogli91 Updated.

codyoss (approved):

PR #6662, opened 2022-09-13
chore(storage): use context value to skip transport tests

noahdietz:
Introduce the `skipTransportTestKey` `context.Value` key in order to skip specific integration tests for one of the two supported transports. The test uses one of the two helpers to create the proper context, either `skipGRPC` or `skipHTTP`, combined with the reason for skipping as the value, and the test runner will skip that transport if it is present in the context supplied to it, logging the reason with the Skip.

Refactor and skip two tests, BucketLifecycle which is broken for gRPC atm, and MultiMessageWriteGRPC which is just meant to exercise edge cases in the gRPC Writer implementation.


tritone (approved):
Thanks Noah!

PR #6679, opened 2022-09-15
fix(spanner): spanner sessions will be cleaned up from the backend

rahul2393:
This will remove the API call to explicitly delete the session from backend, expactation is that sessions will be garbage collected in the backend and to align with other languages Java etc.


spanner/integration_test.go
```diff
@@ -454,9 +454,6 @@ loop:
 		if s == nil {
 			break
 		}
-		// This will delete the session on the backend without removing it
-		// from the pool.
-		s.Value.(*session).delete(context.Background())
```
olavloite, line 459: This should not be removed. The intention behind this is to verify that the session pool is able to recover from a situation where the backend garbage collects sessions that are still in the pool without the pool knowing about it (or expecting it).

rahul2393: Thanks for pointing this out.

olavloite (approved):

PR #6686, opened 2022-09-16
refactor(bigquery/storage/managedwriter): simplify lockingAppend

shollyman:
This PR is another preparatory change towards implementing append
retries.  Here, we move responsibility for recording stats into the
lockingAppend() method, which allows us to simplify the return to
emitting only the error.

This will in turn allow us to use lockingAppend within the receiver
to re-enqueue appends.

There's a slight bit of cleverness in moving stats collection until after
we've unlocked.  This is already tested by
`TestIntegration_ManagedWriter/group/Instrumentation` so this PR 
adds no additional tests.

Towards internal b/245963141


alvarowolfx (approved):
LGTM

PR #6692, opened 2022-09-18
feat(firestore): adds COUNT aggregation query

telpirion:
This PR implements the StructuredAggregationQuery feature for Firestore.


firestore/mock_test.go
```diff
@@ -187,6 +187,27 @@ func (s *mockServer) RunQuery(req *pb.RunQueryRequest, qs pb.Firestore_RunQueryS
...
+	for _, res := range responses {
+		switch res := res.(type) {
+		case *pb.RunAggregationQueryResponse:
+			if err := qs.Send(res); err != nil {
+				return err
+			}
+		case error:
+			return res
+		default:
+			panic(fmt.Sprintf("bad response type in RunAggregationQuery: %+v", res))
```
enocom, line 205: Does this need to be a panic? Wouldn't an error return get the job done?

firestore/query.go
```diff
@@ -304,6 +304,15 @@ func (q Query) Deserialize(bytes []byte) (Query, error) {
 	return q.fromProto(&runQueryRequest)
 }
 
+// NewAggregationQuery returns an AggregationQuery with this query as its
+// base query.
+func (q *Query) NewAggregationQuery() *AggregationQuery {
+	return &AggregationQuery{
+		query:            q,
+		aggregateQueries: make([]*pb.StructuredAggregationQuery_Aggregation, 0),
```
enocom, line 312: Since you're not allocating a certain length, is there a need to make a slice at all? The first append below would do this for you.

firestore/query.go
```diff
@@ -1036,3 +1045,72 @@ func (it *btreeDocumentIterator) next() (*DocumentSnapshot, error) {
 }
 
 func (*btreeDocumentIterator) stop() {}
+
+// AggregationQuery allows for generating aggregation results of an underlying
+// basic query. A single AggregationQuery can contain multiple aggregations.
+type AggregationQuery struct {
+	aggregateQueries []*pb.StructuredAggregationQuery_Aggregation // aggregateQueries contains all of the queries for this request.
```
enocom, line 1052: nit: comments above fields makes for better line lengths and easier-to-parse code

firestore/query.go
```diff
@@ -1036,3 +1045,72 @@ func (it *btreeDocumentIterator) next() (*DocumentSnapshot, error) {
...
+		Operator: &pb.StructuredAggregationQuery_Aggregation_Count_{},
+	}
+
+	a.aggregateQueries = append(a.aggregateQueries, aq)
+
+	return a
+}
+
+// Get retrieves the aggregation query results from the service.
+func (a *AggregationQuery) Get(ctx context.Context) (AggregationResult, error) {
```
enocom, line 1070: I see we use `map[string]interface{}` elsewhere. It's probably best to match that an not introduce a type alias for a map?

https://pkg.go.dev/cloud.google.com/go/firestore#DocumentSnapshot.Data

enocom (changes requested):

firestore/query.go
telpirion: Ack. Resolved over chat.

firestore/query.go
telpirion: Done.

firestore/query.go
telpirion: Done.

firestore/mock_test.go
telpirion: Done. Bad copy pasta

firestore/mock_test.go
```diff
@@ -187,6 +187,27 @@ func (s *mockServer) RunQuery(req *pb.RunQueryRequest, qs pb.Firestore_RunQueryS
...
+func (s *mockServer) RunAggregationQuery(req *pb.RunAggregationQueryRequest, qs pb.Firestore_RunAggregationQueryServer) error {
+	res, err := s.popRPC(req)
+	if err != nil {
+		return err
+	}
+	responses := res.([]interface{})
+	for _, res := range responses {
+		switch res := res.(type) {
+		case *pb.RunAggregationQueryResponse:
+			if err := qs.Send(res); err != nil {
```
enocom, line 199: `return qs.Send(res)`

With a return in each branch, you can also remove the `return nil` outside of the statement.

enocom (approved):

firestore/mock_test.go
telpirion: Compiler doesn't like it when I remove the `return` statement, even when I add a `return nil` after the `if` in the first branch.

PR #6695, opened 2022-09-19
feat(bigquery/storage/managedwriter): support append retries

shollyman:
This PR adds support for automatic retry of failed appends.  Failures are evaluated both from the perspective of any receive errors getting the response, as well as any response that may be embedded in response from the service.  Previous behavior was to simply re-attempt failures when issuing the append.

This PR also adds a new WriterOption (`DisableWriteRetries(disable bool)`) to control this behavior (default is to have retries enabled).

For users sensitive to write duplication, this PR also exposes a new TotalAttempts method on the AppendResult, which will indicate the total number of times this write was attempted.

This also tries to clean up retries in general a bit more.  The generated client will already retry unary RPCs, subject to the [service config](https://github.com/googleapis/googleapis/blob/master/google/cloud/bigquery/storage/v1/bigquerystorage_grpc_service_config.json) present when generating the storage API.

We specifically clarify/introduce two additional kinds of retries above that: a unary retry and a stateless retry.

The unary retry is used to (re)open the underlying bidi network connection which appends are sent upon, as we want to be resilient to reconnection.  The unary retry is effectively stateful for the operation of reopening the connection, and thus uses a gax-based backoff that backs off with increasing intervals.

The stateless retry is used when processing the responses returning from the backend on the bidi stream, where backing off incrementally doesn't make sense.  Instead, we use a base backoff and jitter, and for cases where a more severe backoff is warranted (throughput exhaustion) we use a multiplication factor.  The intent here is to provide backpressure which will eventually saturate the append queue and cause blocking/rejection of writes until the backlog recovers.


bigquery/storage/managedwriter/managed_stream.go
```diff
@@ -324,6 +328,8 @@ func (ms *ManagedStream) lockingAppend(pw *pendingWrite) error {
 	}
 	if err != nil {
 		if shouldReconnect(err) {
```
yirutang, line 330: We should reconnect if we experience over x times of in stream retries. And also there are some known internal errors that needs to be treated as unavailable:
https://screenshot.googleplex.com/9xtULSCaXLMJSJT

shollyman: aborted/cancelled/unavailable are all valid codes for retry scenarios already.

The notion of in-connection errors is not encoded in this PR.  Can you clarify?  Do you mean X sequential errors, or X total on a given connection? Is this error restricted to certain status codes?  Are the errors from the bidi receive, or are they embedded in the response message?

yirutang: I mean if you have already retried X times on the same connection without succeeding, maybe it is worth to reconnect. We had some issues when the whole connection and not able to recover, it should be a bug on our side but it would be nice to cover it up in the client. (I failed to find the bug... the bug was on internal error).

By establishing a new connection, we found a new task and retry everything. So it is worth doing that.

shollyman: I'm inclined to put this logic in a subsequent PR to prevent this one becoming too unwieldy.  Added the scenario to the existing design doc.

yirutang: Sure.

yirutang (approved):

bigquery/storage/managedwriter/retry.go
```diff
@@ -27,64 +27,97 @@ import (
...
-		return r.bo.Pause(), true
+		// All other non-status errors are treated as non-retryable (including context errors).
+		return
 	}
 	switch s.Code() {
 	case codes.Aborted,
 		codes.Canceled,
 		codes.DeadlineExceeded,
 		codes.Internal,
 		codes.Unavailable:
```
yirutang, line 56: Sorry, missed this, please retry on failed_precondition.

PR #6698, opened 2022-09-19
chore: fix v2 beta edge case

codyoss:
- There was a bug in the code used to generate the mappings files
where the location for type found in v2 libraries were placed in
the correct directory.
- Also start generating betas of old client in v2 modules.
- Flipped these "new" libs to migrated since they have not been generated in v2 before

Changes:

feat(recaptchaenterprise/v2): start generating apiv1beta1

feat(vision/v2): start generating apiv1p1beta1


codyoss:
cc @adrianajg

quartzmo (approved):

PR #6699, opened 2022-09-20
feat(spanner): retry spanner transactions and mutations when RST_STREAM error

rahul2393:
* Clients need to retry the Spanner transactions in case of RST_STREAM error returned from the backend

### DIFF
--- a/spanner/retry.go
+++ b/spanner/retry.go
@@ -80,13 +80,13 @@ func (r *spannerRetryer) Retry(err error) (time.Duration, bool) {
 }
 
 // runWithRetryOnAbortedOrSessionNotFound executes the given function and
-// retries it if it returns an Aborted or Session not found error. The retry
-// is delayed if the error was Aborted. The delay between retries is the delay
+// retries it if it returns an Aborted, Session not found error or certain Internal errors. The retry
+// is delayed if the error was Aborted or Internal error. The delay between retries is the delay
 // returned by Cloud Spanner, or if none is returned, the calculated delay with
 // a minimum of 10ms and maximum of 32s. There is no delay before the retry if
 // the error was Session not found.
 func runWithRetryOnAbortedOrSessionNotFound(ctx context.Context, f func(context.Context) error) error {
-	retryer := onCodes(DefaultRetryBackoff, codes.Aborted)
+	retryer := onCodes(DefaultRetryBackoff, codes.Aborted, codes.Internal)
 	funcWithRetry := func(ctx context.Context) error {
 		for {
 			err := f(ctx)
--- a/spanner/transaction.go
+++ b/spanner/transaction.go
@@ -1387,51 +1387,64 @@ func (t *writeOnlyTransaction) applyAtLeastOnce(ctx context.Context, ms ...*Muta
 		ts time.Time
 		sh *sessionHandle
 	)
+	defer func() {
+		if sh != nil {
+			sh.recycle()
+		}
+	}()
 	mPb, err := mutationsProto(ms)
 	if err != nil {
 		// Malformed mutation found, just return the error.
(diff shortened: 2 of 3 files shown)


spanner/retry.go
```diff
@@ -80,13 +80,13 @@ func (r *spannerRetryer) Retry(err error) (time.Duration, bool) {
...
-// retries it if it returns an Aborted or Session not found error. The retry
-// is delayed if the error was Aborted. The delay between retries is the delay
+// retries it if it returns an Aborted, Session not found error or certain Internal errors. The retry
+// is delayed if the error was Aborted or Internal error. The delay between retries is the delay
 // returned by Cloud Spanner, or if none is returned, the calculated delay with
 // a minimum of 10ms and maximum of 32s. There is no delay before the retry if
 // the error was Session not found.
 func runWithRetryOnAbortedOrSessionNotFound(ctx context.Context, f func(context.Context) error) error {
-	retryer := onCodes(DefaultRetryBackoff, codes.Aborted)
+	retryer := onCodes(DefaultRetryBackoff, codes.Aborted, codes.Internal)
```
olavloite, line 89: This will make it retry on any `Internal` error, also those that are not retryable. This could cause an infinite (or at least very long-lasting) loop if someone tries to execute a transaction that happens to hit a feature that returns an internal error.

spanner/transaction.go
```diff
@@ -1387,51 +1387,64 @@ func (t *writeOnlyTransaction) applyAtLeastOnce(ctx context.Context, ms ...*Muta
...
-	for {
-		if sh == nil || sh.getID() == "" || sh.getClient() == nil {
-			// No usable session for doing the commit, take one from pool.
-			sh, err = t.sp.take(ctx)
-			if err != nil {
-				// sessionPool.Take already retries for session
-				// creations/retrivals.
-				return ts, err
+	// Make a retryer for Aborted and certain Internal errors.
+	retryer := onCodes(DefaultRetryBackoff, codes.Aborted, codes.Internal)
```
olavloite, line 1402: This also could cause an infinite loop.

spanner/retry.go
```diff
@@ -80,13 +80,13 @@ func (r *spannerRetryer) Retry(err error) (time.Duration, bool) {
...
 
 // runWithRetryOnAbortedOrSessionNotFound executes the given function and
-// retries it if it returns an Aborted or Session not found error. The retry
-// is delayed if the error was Aborted. The delay between retries is the delay
+// retries it if it returns an Aborted, Session not found error or certain Internal errors. The retry
+// is delayed if the error was Aborted or Internal error. The delay between retries is the delay
 // returned by Cloud Spanner, or if none is returned, the calculated delay with
 // a minimum of 10ms and maximum of 32s. There is no delay before the retry if
 // the error was Session not found.
 func runWithRetryOnAbortedOrSessionNotFound(ctx context.Context, f func(context.Context) error) error {
```
olavloite, line 88: Could we have a couple of tests for the changes in this PR that verify that transactions are retried if the `RST_STREAM` error is returned, but not for any other random internal error?

rahul2393: @olavloite Test "TestClient_ApplyAtLeastOnce_NonRetryableInternalErrors" added, 
 Please take a look at the code again.
 `This could cause an infinite ` => This is not true. The following code only says that:

if it is an internal error and its error message is not in the list, then return `false`.
If it is an internal error and it error message in the list, then let it pass to the next check `r.Retryer.Retry(err)`.

It early returns `false` for non-retryable internal errors before the real retry check.

If the user has specified to not retry `internal errors`, then `r.Retryer.Retry(err)` will return `false`.

olavloite (approved):

PR #6704, opened 2022-09-20
chore: update how we handle mod replacements

codyoss:
- Add a tidy after replace. The genproto dependency is getting more complex as it now associates with many aliases. A tidy is now requried.
- Since we are tidying we can no longer just simply drop the replace as there would have been dependency updates because of the tidy. Instead we can achive the same effect by using git restore.


quartzmo (approved):
Great fix, thank you!!

PR #6707, opened 2022-09-20
feat(bigquery/analyticshub): start generating apiv1

quartzmo:


codyoss (dismissed):

internal/gapicgen/generator/config.go
```diff
@@ -1763,6 +1763,16 @@ var MicrogenGapicConfigs = []*MicrogenConfig{
 		Transports:            []string{"grpc", "rest"},
 		ReleaseLevel:          "beta",
 	},
+	{
```
shollyman, line 1766: mostly a nit that can be cleaned up in a subsequent PR, but I thought we kept the configs ordered by input path?

shollyman (approved):

internal/gapicgen/generator/config.go
quartzmo: I don't think this has ever been the case. The first three are:

```
		ImportPath:           "cloud.google.com/go/compute/apiv1",
...
		ImportPath:            "cloud.google.com/go/texttospeech/apiv1",
...
		ImportPath:            "cloud.google.com/go/asset/apiv1",
```

codyoss (approved):

internal/gapicgen/generator/config.go
shollyman: there's a partially ordering joke to be made here, clearly.  Disregard my comment entirely. :)

PR #6712, opened 2022-09-20
chore(datastore): enable Go 1.13 error wrapping

quartzmo:


datastore/save.go
```diff
@@ -166,7 +166,7 @@ func reflectFieldSave(props *[]Property, p Property, name string, opts saveOpts,
 		if pSaver, ok := vi.(PropertyLoadSaver); ok {
 			val, err := pSaver.Save()
 			if err != nil {
-				return fmt.Errorf("field save error: %v", err)
+				return fmt.Errorf("field save error: %w", err)
```
quartzmo, line 169: I don't think `Save()` actually raises any complex error that is worth wrapping, but I changed this usage to `%w` just to future-proof it in case a rich error is added in the future. Let me know if you disagree and would rather leave this usage as `%v`.

tritone: Drive by on this -- PropertyLoadSaver is an interface that would be implemented on their data types by the library user. We would definitely want to wrap their error rather than swallowing it.

quartzmo: I suspected something like that, thank you @tritone !

telpirion (approved):

PR #6721, opened 2022-09-22
chore: restrict gapic proto find func to a dir

codyoss:
Some apis, mainly aiplatform, have nested helper types in subdir under the serivce proto defs. We only want to look at the files in the service dir to generate the gapic service.

This unblocks us generating aliases for aiplatform.


internal/gapicgen/generator/gapics.go
```diff
@@ -279,20 +279,16 @@ func (g *GapicGenerator) microgen(conf *MicrogenConfig) error {
 	log.Println("microgen generating", conf.Pkg)
 
 	var protoFiles []string
-	if err := filepath.Walk(g.googleapisDir+"/"+conf.InputDirectoryPath, func(path string, info os.FileInfo, err error) error {
-		if err != nil {
-			return err
-		}
-		// Ignore compute_small.proto which is just for testing and would cause a collision if used in generation.
-		//
-		// TODO(noahdietz): Remove this when it is no longer needed.
```
quartzmo, line 288: Do you want to preserve these comments?

codyoss: good, point, done.

quartzmo (approved):
Great fix, probably should have been like this from the start!

PR #6736, opened 2022-09-26
chore(internal/gapicgen): add alias mode to gapicgen

codyoss:
This mode will be run as a cron to keep alias files found in go-genproto up-to-date with the most current release of the google-cloud-go service libray. We need to make sure to refresh aliases only on released versions of google-cloud-go clients. Genproto will naturally lag behind google-cloud-go a little be, but this is fine and the lag may act as an insentive to migrate to the new stubs dir.

### DIFF
--- a/internal/gapicgen/cmd/genbot/alias.go
+++ b/internal/gapicgen/cmd/genbot/alias.go
@@ -1,0 +1,85 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"path/filepath"
+	"strings"
+
+	"cloud.google.com/go/internal/aliasfix"
+	"cloud.google.com/go/internal/aliasgen"
+	"cloud.google.com/go/internal/gapicgen/execv/gocmd"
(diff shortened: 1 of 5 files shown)


internal/gapicgen/cmd/genbot/alias.go
```diff
@@ -0,0 +1,176 @@
...
+	grp.Go(func() error {
+		if c.gocloudDir != "" {
+			return nil
+		}
+		return git.DeepClone("https://github.com/googleapis/google-cloud-go", filepath.Join(tmpDir, "gocloud"))
+	})
+	if err := grp.Wait(); err != nil {
+		return err
+	}
+	genprotoDir := deafultDir(filepath.Join(tmpDir, "genproto"), c.genprotoDir)
```
quartzmo, line 102: Typo: `deafultDir`. Sorry I missed this when the function was initially reviewed.

codyoss: done.

quartzmo (approved):
Very nice!

PR #6743, opened 2022-09-26
chore(storage): enable Go 1.13 error wrapping

quartzmo:


tritone (commented):
All this looks good. Would you like to add wrapping for the sentinel errors in this PR as well (as we discussed)? Or should I do that separately?

I'd also like to update the errors section in doc.go to discuss wrapping

quartzmo:
Let's do that separately. Will the sentinel error conversion be a breaking change?

PR #6746, opened 2022-09-26
feat(bigtable): add update table metadata support

kimihrr:
- Add `DeletionProtection` field to `TableConf` and `TableInfo`
- Add `UpdateTable` function to change the value of `DeletionProtection`
- Make sure that only `DeletionProtection` can be updated through `UpdateTable`
- Add `mockTableAdminClock` and `setupTableClient` to support tests for Table admin APIs
- Add tests for `updateTable`
- `createTable` changes to support `DeletionProtection` will be added in the next pull request


bigtable/admin.go
```diff
@@ -275,6 +278,35 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
...
 }
 
+// UpdateTable updates a table in the instance from the given configuration.
+// Only deletion protection can be updated at this moment
+func (ac *AdminClient) UpdateTable(ctx context.Context, conf *TableConf) (updated bool, err error) {
+	if conf.TableID == "" {
+		return false, errors.New("TableID is required")
+	}
+	if conf.Families != nil || conf.SplitKeys != nil {
+		return false, errors.New("Only deletion protection field can be updated")
```
yuhanpenggoogle, line 288: deletion protection bit

bigtable/admin_test.go
```diff
@@ -25,6 +26,83 @@ import (
...
 
+type mockTableAdminClock struct {
+	btapb.BigtableTableAdminClient
+
+	getTableReq    *btapb.GetTableRequest
+	updateTableReq *btapb.UpdateTableRequest
+	getTableResp   *btapb.Table
+}
+
+func (c *mockTableAdminClock) GetTable(
```
yuhanpenggoogle, line 37: Is this function ever used?

bigtable/admin_test.go
```diff
@@ -25,6 +26,83 @@ import (
...
+	c := setupTableClient(t, mock)
+
+	// Check if the deletion protection updates correctly
+	deletion_protection := true
+	_, err := c.UpdateTable(context.Background(), &TableConf{TableID: "My-table", DeletionProtection: &deletion_protection})
+	if err != nil {
+		t.Fatalf("UpdateTable failed: %v", err)
+	}
+	updateTableReq := mock.updateTableReq
+	if updateTableReq.Table.Name != "My-table" {
```
yuhanpenggoogle, line 77: I think for this case, err would always be nil.
You want to print the updateTableReq.Table.Name instead.
(same for below)

bigtable/admin.go
```diff
@@ -275,6 +278,35 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
 	return err
 }
 
+// UpdateTable updates a table in the instance from the given configuration.
+// Only deletion protection can be updated at this moment
```
kevinsi4508, line 282: Nit: Only deletion protection can be updated at this moment.

End a sentence with a period.

bigtable/admin.go
```diff
@@ -275,6 +278,35 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
...
+func (ac *AdminClient) UpdateTable(ctx context.Context, conf *TableConf) (updated bool, err error) {
```
kevinsi4508, line 283: Why are we using `TableConf` here? Are all the fields in `TableConf` relevant to UpdateTable?

Maybe use a new struct: UpdateTableConf
{
        TableID   string
	// Set to true to make the table protected against data loss
	// i.e. deleting the table, the column families in the table and the instance containing the table would be prohibited
	DeletionProtection *bool
}

bigtable/admin.go
```diff
@@ -275,6 +278,35 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
...
+		updateMask.Paths = append(updateMask.Paths, "deletion_protection")
+
+		req := &btapb.UpdateTableRequest{
+			Table: &btapb.Table{
+				Name:               conf.TableID,
+				DeletionProtection: *conf.DeletionProtection,
+			},
+			UpdateMask: updateMask,
+		}
+		_, err := ac.tClient.UpdateTable(ctx, req)
```
kevinsi4508, line 304: Why are we ignoring the response here?

bigtable/admin.go
```diff
@@ -306,6 +338,8 @@ type TableInfo struct {
 	// DEPRECATED - This field is deprecated. Please use FamilyInfos instead.
 	Families    []string
 	FamilyInfos []FamilyInfo
+	// When true: deleting the table, the column families in the table and the instance containing the table is prohibited
```
kevinsi4508, line 341: nit: When set to true, deleting the table ....

kevinsi4508 (commented):
I have not looked at the test yet.

bigtable/admin.go
yuhanpenggoogle: Update: use "Only deletion protection can be enabled"

bigtable/admin.go
```diff
@@ -275,6 +278,35 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
...
+			Table: &btapb.Table{
+				Name:               conf.TableID,
+				DeletionProtection: *conf.DeletionProtection,
+			},
+			UpdateMask: updateMask,
+		}
+		_, err := ac.tClient.UpdateTable(ctx, req)
+		return true, err
+	}
+	return false, errors.New("Deletion protection is required")
```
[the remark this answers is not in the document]
yuhanpenggoogle: Update: no need to change this.

bigtable/admin.go
```diff
@@ -217,6 +217,9 @@ type TableConf struct {
 	SplitKeys []string
 	// Families is a map from family name to GCPolicy
 	Families map[string]GCPolicy
+	// Set to true to make the table protected against data loss
+	// i.e. deleting the table, the column families in the table and the instance containing the table would be prohibited
+	DeletionProtection *bool
```
telpirion, line 222: Why is this a pointer to a bool rather than just a bool?

bigtable/admin.go
telpirion, line 288: NIT: Go style is to write messages as all lower-case.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

bigtable/admin.go
```diff
@@ -275,6 +278,35 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
...
+// Only deletion protection can be updated at this moment
+func (ac *AdminClient) UpdateTable(ctx context.Context, conf *TableConf) (updated bool, err error) {
+	if conf.TableID == "" {
+		return false, errors.New("TableID is required")
+	}
+	if conf.Families != nil || conf.SplitKeys != nil {
+		return false, errors.New("Only deletion protection field can be updated")
+	}
+
+	if conf.DeletionProtection != nil {
```
telpirion, line 291: If you change `DeletionProtection` to a bool rather than a pointer, you can just write:

```
if conf.DeletionProtection { ... }
```

bigtable/admin.go
telpirion: +1

Also, if we get an error from the call to the service, should we return true to the caller?

bigtable/admin.go
telpirion, line 307: NIT: write error messages in lower case.

bigtable/admin.go
telpirion, line 341: It's common style for Godoc to begin a comment on an exported field with the name of the field.

bigtable/admin_test.go
telpirion: If it isn't used (right now), it's probably best to send an UNIMPLEMENTED response to the client. That helps future maintainers to know that the logic isn't complete.

bigtable/admin_test.go
```diff
@@ -25,6 +26,83 @@ import (
...
+		t.Fatalf("UpdateTableRequest does not match: %v", err)
```
telpirion, line 78: NIT: I would use `t.Errorf()` here instead of `t.Fatalf()`. `Fatalf()` exits the test, which means that we don't get any results from the other "asserts."

bigtable/admin_test.go
```diff
@@ -25,6 +26,83 @@ import (
...
+	}
+	if updateTableReq.Table.DeletionProtection != true {
+		t.Fatalf("UpdateTableRequest does not match: %v", err)
```
telpirion, line 81: See above about `Fatalf()`

bigtable/admin_test.go
```diff
@@ -25,6 +26,83 @@ import (
...
+	}
+
+	// Check if the update fails when TableID is not provided
+	_, err = c.UpdateTable(context.Background(), &TableConf{DeletionProtection: &deletion_protection})
+	if fmt.Sprint(err) != "TableID is required" {
+		t.Fatalf("UpdateTable failed: %v", err)
```
telpirion, line 87: See above about `Fatalf()`.

bigtable/admin_test.go
```diff
@@ -25,6 +26,83 @@ import (
...
+	}
+
+	// Check if the update fails when deletion protection is not provided
```
telpirion, line 90: If this is a separate test case, I think it makes sense logically to abstract it into a new test method.

bigtable/admin_test.go
```diff
@@ -25,6 +26,83 @@ import (
...
+	_, err = c.UpdateTable(context.Background(), &TableConf{TableID: "My-table"})
+
+	if fmt.Sprint(err) != "Deletion protection is required" {
+		t.Fatalf("UpdateTable failed: %v", err)
+	}
+
+	// Check if the update fails when split keys or column families are provided
```
telpirion, line 97: Same as previous comment: if this should be a separate test case, abstract it into a new method.

telpirion (changes requested):

bigtable/admin.go
kevinsi4508: I believe this field is optional. Again, I think we should use a different struct update.

bigtable/admin.go
```diff
@@ -275,6 +278,47 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
 	return err
 }
 
+// TableConf contains all of the information necessary to create a table with column families.
```
kevinsi4508: ... to update a table.

bigquery/storage/managedwriter/client.go
```diff
@@ -117,6 +117,8 @@ func (c *Client) buildManagedStream(ctx context.Context, streamFunc streamClient
 			gax.WithGRPCOptions(grpc.MaxCallRecvMsgSize(10 * 1024 * 1024)),
 		},
 		open: createOpenF(ctx, streamFunc),
+		// We add the new retryer by default, and add a new option to disable it.
```
kevinsi4508: ??? revert?

bigtable/admin.go
```diff
@@ -275,6 +278,47 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
...
+type UpdateTableConf struct {
+	TableID string
+	// Set to true to make the table protected against data loss
+	// i.e. deleting the table, the column families in the table and the instance containing the table would be prohibited
+	DeletionProtection *bool
+}
+
+// UpdateTable updates a table in the instance from the given configuration.
+// Only deletion protection can be updated at this period.
+func (ac *AdminClient) UpdateTable(ctx context.Context, conf *UpdateTableConf) (updated bool, err error) {
```
kevinsi4508, line 291: Do we really need a bool? no error means the table is updated successfully, right?

bigtable/admin.go
```diff
@@ -306,6 +350,8 @@ type TableInfo struct {
 	// DEPRECATED - This field is deprecated. Please use FamilyInfos instead.
 	Families    []string
 	FamilyInfos []FamilyInfo
+	// When set to true deleting the table, the column families in the table and the instance containing the table is prohibited
```
kevinsi4508, line 353: Is this used? when we read the table back?

bigtable/admin.go
```diff
@@ -275,6 +278,47 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
...
+				Name:               conf.TableID,
+				DeletionProtection: *conf.DeletionProtection,
+			},
+			UpdateMask: updateMask,
+		}
+		lro, err := ac.tClient.UpdateTable(ctx, req)
+		if err != nil {
+			return false, err
+		}
+		err = longrunning.InternalNewOperation(ac.lroClient, lro).Wait(ctx, nil)
```
kevinsi4508, line 313: Do we usually wait here or return the operation back?

bigtable/admin.go
```diff
@@ -217,6 +217,9 @@ type TableConf struct {
 	SplitKeys []string
 	// Families is a map from family name to GCPolicy
 	Families map[string]GCPolicy
+	// Set to true to make the table protected against data loss
+	// i.e. deleting the table, the column families in the table and the instance containing the table would be prohibited
```
kevinsi4508, line 221: Shouldn't we remove this? We don't use it in UpdateTable, right?

bigtable/admin.go
```diff
@@ -275,6 +278,47 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
...
+			},
+			UpdateMask: updateMask,
+		}
+		lro, err := ac.tClient.UpdateTable(ctx, req)
+		if err != nil {
+			return err
+		}
+		err = longrunning.InternalNewOperation(ac.lroClient, lro).Wait(ctx, nil)
+		if err != nil {
+			return err
```
kevinsi4508, line 315: Just "return err"

bigtable/admin.go
kevinsi4508: Why this is resolved?

bigtable/admin.go
kevinsi4508: nvm, we are doing this in file. It is as not flexible as return the long running operation itself, but I okay with this.

bigtable/admin.go
kevinsi4508: Maybe remove this for now.

bigtable/admin_test.go
```diff
@@ -25,6 +26,77 @@ import (
...
+	deletion_protection := true
+	err := c.UpdateTable(context.Background(), &UpdateTableConf{TableID: "My-table", DeletionProtection: &deletion_protection})
+	if err != nil {
+		t.Errorf("UpdateTable failed: %v", err)
+	}
+	updateTableReq := mock.updateTableReq
+	if updateTableReq.Table.Name != "My-table" {
+		t.Errorf("UpdateTableRequest does not match, TableID: %v", updateTableReq.Table.Name)
+	}
+	if updateTableReq.Table.DeletionProtection != true {
```
kevinsi4508, line 71: compare protos? 

Example:
                       if !cmp.Equal(gotConfig, tc.wantConfig) {
				t.Fatalf("want autoscaling config = %v, got = %v", tc.wantConfig, gotConfig)
			}

bigtable/admin_test.go
```diff
@@ -25,6 +26,77 @@ import (
...
+	c := setupTableClient(t, mock)
+
+	// Check if the update fails when deletion protection is not provided
+	err := c.UpdateTable(context.Background(), &UpdateTableConf{TableID: "My-table"})
+
+	if fmt.Sprint(err) != "deletion protection is required" {
+		t.Errorf("UpdateTable failed: %v", err)
+	}
+}
+
```
kevinsi4508, line 99: Maybe add a test case where ac.tClient.UpdateTable(ctx, req) returns an error.

kevinsi4508 (approved):
LGTM

bigtable/admin.go
```diff
@@ -275,6 +275,44 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
...
+				DeletionProtection: *conf.DeletionProtection,
+			},
+			UpdateMask: updateMask,
+		}
+		lro, err := ac.tClient.UpdateTable(ctx, req)
+		if err != nil {
+			return err
+		}
+		err = longrunning.InternalNewOperation(ac.lroClient, lro).Wait(ctx, nil)
+		return err
```
kevinsi4508, line 311: return longrunning.InternalNewOperation(ac.lroClient, lro).Wait(ctx, nil)

bigtable/admin.go
telpirion: Please don't resolve comments without explaining why you haven't made the requested changes.

bigtable/admin.go
```diff
@@ -275,6 +275,44 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
 	return err
 }
 
+// UpdateTableConf contains all of the information necessary to update a table with column families.
+type UpdateTableConf struct {
+	TableID string
+	// Set to true to make the table protected against data loss
+	// i.e. deleting the table, the column families in the table and the instance containing the table would be prohibited
```
telpirion, line 282: NIT: Insert a line break in this line.

bigtable/admin.go
```diff
@@ -275,6 +275,44 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
...
+	DeletionProtection *bool
```
telpirion, line 283: Per a previous comment, I don't think this needs to be a pointer to a bool--it can just be a bool.

telpirion (changes requested):

bigtable/admin_test.go
```diff
@@ -25,6 +27,96 @@ import (
...
+	c, err := NewAdminClient(ctx, "my-cool-project", "my-cool-instance")
+	if err != nil {
+		t.Fatalf("NewAdminClient failed: %v", err)
+	}
+	c.tClient = ac
+	return c
+}
+
+func TestTableAdmin_UpdateTable(t *testing.T) {
+	mock := &mockTableAdminClock{updateTableError: nil}
```
kevinsi4508, line 60: We don't need to set updateTableError: nil, right?

bigtable/admin.go
```diff
@@ -275,6 +275,67 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
 	return err
 }
 
+// set to true to make the table protected against data loss
+// i.e. deleting the table, the column families in the table,
+// and the instance containing the table would be prohibited
+type DeletionProtection int
+
+const (
+	UnSet DeletionProtection = 0
```
telpirion, line 284: NIT: "Unset" rather than "UnSet"

bigtable/admin.go
```diff
@@ -275,6 +275,67 @@ func (ac *AdminClient) CreateColumnFamily(ctx context.Context, table, family str
 	return err
 }
 
+// set to true to make the table protected against data loss
```
telpirion, line 278: Type comments should start with the name of the type. 

Suggestion: "DeletionProtection indicates whether the table is protected ..."

bigtable/admin.go
telpirion: I also think "True" and "False" could be confusing here. I would recommend the set "None, Protected, Unprotected".

bigtable/admin_test.go
```diff
@@ -25,6 +27,96 @@ import (
...
+		t.Fatalf("NewAdminClient failed: %v", err)
+	}
+	c.tClient = ac
+	return c
+}
+
+func TestTableAdmin_UpdateTable(t *testing.T) {
+	mock := &mockTableAdminClock{}
+	c := setupTableClient(t, mock)
+	var deletion_protection DeletionProtection
```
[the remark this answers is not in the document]
telpirion: Use camelCase instead of snake case, e.g. "deletionProtection" not "deletion_protection".

telpirion (changes requested):
I think this feature probably needs an integration test, too.

telpirion (approved):

PR #6753, opened 2022-09-27
feat(spanner): Proto Columns Changes

harshachinta:
This is draft PR containing changes to add support for proto columns


harshachinta:
Support already added in https://github.com/googleapis/google-cloud-go/pull/9315

PR #6754, opened 2022-09-27
chore(storage): retry configs for gRPC media ops

tritone:
Adds retry config to grpcReader and grpcWriter types so that they can be propagated to operations as needed. Ensure that the configured values are used for all retry calls, and that UserProject is consistently passed through as well.


noahdietz (approved):
Well done, makes perfect sense.

PR #6755, opened 2022-09-27
feat(datastore): adds snapshot reads

telpirion:
This PR adds snapshot reads to the client and transactions. Snapshot reads allow for a time-specific read-only version of the database to be accessed.

Note: This PR does not include integration tests, as read results do not indicate a read time.


datastore/datastore.go
```diff
@@ -679,3 +701,16 @@ func (c *Client) Mutate(ctx context.Context, muts ...*Mutation) (ret []*Key, err
 	}
 	return ret, nil
 }
+
+// ReadOptions provides specific instructions for how to access documents in the database.
+// Currently, only ReadTime is supported.
+type ReadOptions struct {
```
enocom, line 707: Without any context on what the future might hold, I might add a `WithReadTime` below and otherwise not export this struct. In the future you could add a `WithReadOptions` perhaps. In either case, I'm inclined to expand the API only as much as is needed and then leave the door open for a more generic API in the future. WDYT?

datastore/datastore.go
```diff
@@ -679,3 +701,16 @@ func (c *Client) Mutate(ctx context.Context, muts ...*Mutation) (ret []*Key, err
...
+	ReadTime time.Time
+}
+
+// WithReadOptions specifies constraints for accessing documents from the database,
+// e.g. at what time snapshot to read the documents.
+// Once set, all subsequent reads from the client use these options.
```
enocom, line 713: What if a client wants to a read, and then in the future do another read? This comment makes me think this method cannot be called twice when in reality, it probably could be adjusted as needed. Perhaps rephrase as, "The client uses this value for subsequent reads"

datastore/transaction.go
```diff
@@ -67,6 +70,22 @@ func (w maxAttempts) apply(s *transactionSettings) {
 	}
 }
 
+// WithReadTime returns a TransactionOption that specifies a snapshot of the
+// database to view.
+func WithReadTime(t time.Time) TransactionOption {
```
enocom, line 75: This pattern looks better to me and is more user friendly compared to requiring users to configure a struct correctly. Could we use this above?

telpirion: Done.

datastore/datastore.go
telpirion: Done.

datastore/datastore.go
telpirion: Done.

datastore/datastore.go
```diff
@@ -679,3 +701,40 @@ func (c *Client) Mutate(ctx context.Context, muts ...*Mutation) (ret []*Key, err
...
+	readTime time.Time
+}
+
+// WithReadOptions specifies constraints for accessing documents from the database,
+// e.g. at what time snapshot to read the documents.
+// The client uses this value for subsequent reads, unless additional ReadOptions
+// are provided.
+func (c *Client) WithReadOptions(ro ...ReadOption) *Client {
+	for _, r := range ro {
+		switch o := r.(type) {
```
enocom, line 734: This type switch is not necessary since all your `ro` have the same signature.

datastore/datastore_test.go
```diff
@@ -3484,15 +3584,15 @@ func TestDeferredMissing(t *testing.T) {
 }
 
 func TestGetWithNilKey(t *testing.T) {
-	client := &Client{}
+	client := &Client{readSettings: &readSettings{}}
```
enocom, line 3587: This should be done in the `NewClient` initializer.

datastore/datastore.go
```diff
@@ -53,9 +55,10 @@ const resourcePrefixHeader = "google-cloud-resource-prefix"
 
 // Client is a client for reading and writing data in a datastore dataset.
 type Client struct {
-	connPool gtransport.ConnPool
-	client   pb.DatastoreClient
-	dataset  string // Called dataset by the datastore API, synonym for project ID.
+	connPool     gtransport.ConnPool
+	client       pb.DatastoreClient
+	dataset      string // Called dataset by the datastore API, synonym for project ID.
+	readSettings *readSettings
```
enocom, line 61: We'll need to initialize this in `NewClient` to avoid nil checks below.

datastore/datastore.go
```diff
@@ -679,3 +701,40 @@ func (c *Client) Mutate(ctx context.Context, muts ...*Mutation) (ret []*Key, err
 	}
 	return ret, nil
 }
+
+// ReadTime specifies a snapshot (time) of the database to read.
+func ReadTime(t time.Time) ReadOption {
+	return docReadTime{t}
+}
+
+type docReadTime struct {
```
enocom, line 710: If you don't intend to add any additional fields to this struct you could also do this:

``` golang
type docReadTime time.Time
```

datastore/datastore_test.go
```diff
@@ -3081,6 +3081,103 @@ func TestPutMultiTypes(t *testing.T) {
...
+		},
+	}
+
+	client := &Client{
+		client:       fakeClient,
+		readSettings: &readSettings{},
+	}
+
+	ctx := context.Background()
+	client.WithReadOptions(ReadTime(time.Now()))
```
enocom, line 3117: How are we checking that the read time is actually sent to the backend? Maybe I'm missing something, but I don't see how this test exercises that logic.

enocom (changes requested):

datastore/datastore_test.go
telpirion: This is just a unit test. Any check where we read the response's read time would be merely a result of how we set up the mock (and hence not useful as a test case).

telpirion: This test is really just to ensure that using `Client.WithReadOptions()` doesn't result in an error.

datastore/datastore.go
telpirion: Done.

datastore/datastore.go
telpirion: Done.

datastore/datastore_test.go
telpirion: The `readSettings` are now set in `NewClient()`. However, refactoring all of the unit tests to use `NewClient()` is beyond the scope of this PR.

datastore/datastore.go
telpirion: Done.

datastore/datastore_test.go
```diff
@@ -3081,6 +3081,103 @@ func TestPutMultiTypes(t *testing.T) {
 	}
 }
 
+func TestGetWithReadTime(t *testing.T) {
+	type Ent struct {
```
enocom, line 3085: Le's not export the type, especially since it's test only. Also, would just one field be enough? The second doesn't really alter the meaning of the test and seems unnecssary.

datastore/datastore_test.go
```diff
@@ -3081,6 +3081,103 @@ func TestPutMultiTypes(t *testing.T) {
...
+	e := &pb.Entity{
+		Key: keyToProto(k),
+		Properties: map[string]*pb.Value{
+			"A": {ValueType: &pb.Value_IntegerValue{IntegerValue: 1}},
+			"B": {ValueType: &pb.Value_StringValue{StringValue: "one"}},
+		},
+	}
+	fakeClient := &fakeDatastoreClient{
+		lookup: func(*pb.LookupRequest) (*pb.LookupResponse, error) {
+			return &pb.LookupResponse{
```
enocom, line 3100: Could you capture the lookup request's read time here to validate that the client is passing it through correctly?

datastore/datastore_test.go
enocom: Ditto below.

datastore/datastore_test.go
enocom: Ditto below.

enocom (changes requested):

datastore/datastore_test.go
telpirion: Done.

PR #6768, opened 2022-09-28
test(storage): add tokensource to GRPC client creation

BrennaEpp:
This is a step in the HTTP client creation for tests that was removed (and never added to GRPC client config). It enables the use of `GCLOUD_TESTS_GOLANG_KEY` and project env variable set to a different project than the default app creds.

Also adds dual transport to `TestIntegration_UpdateDefaultEventBasedHold`


tritone (approved):
Thanks for the fix!

PR #6786, opened 2022-09-29
feat(internal/testutil): make it easier to test integration tests locally

codyoss:
Added a new environment variable that is checked by the code used to source credentials for integration tests. Now if a maintainer sets `GCLOUD_TESTS_IMPERSONATE_CREDENTIALS=true` they will enter a flow where they can impersonate our testing service account using their local ADC creds.

Doing something like this was recently quite useful for debugging a BigQuery test failure locally. I imagine this will also make writing/testing integration tests locally much easier.


quartzmo (approved):
Very cool! I really appreciate this.

PR #6789, opened 2022-09-29
chore(storage): fix sending extraneous crc3c

noahdietz:
Enabling the `HashesOnUpload` test discovered a bug in the client library that may actually be a bug in the service (filed something separately for that). The WriteObjectSpec.resource.checksums.crc32c was being set unconditionally on WriteObject and even though it is supposed to be ignored as OUTPUT_ONLY the service was validating it. This caused `HashesOnUpload` to fail with the fix in this PR.

Also enables `NonexistentBucket`, `NoUnicodeNormalization`, and refactors `PerObjectStorageClass` but skipping gRPC b.c of a bug.


storage/grpc_client.go
```diff
@@ -1687,6 +1687,11 @@ func (w *gRPCWriter) writeObjectSpec() (*storagepb.WriteObjectSpec, error) {
 	spec := &storagepb.WriteObjectSpec{
 		Resource: attrs.toProtoObject(w.bucket),
 	}
+	// Object.checksums is OUTPUT_ONLY and should never be sent.
+	// WriteObjectRequest.object_checksums should be used instead.
+	if checksums := spec.GetResource().GetChecksums(); checksums != nil {
```
tritone, line 1692: How would this be set in the first place? If library users can set it, I would say we should give them a validation error rather than silently unsetting it.

noahdietz: If the Object attrs that the user provides include a CRC32C checksum, then it will be included on the request, because on line 1688, the `attrs.toProtoObject` just blindly converts from attrs to proto Object.

So I don't think a validation error should be used. The CRC32C in question comes from the Writer's embedded ObjectAttrs that is actually used to transmit the CRC32C when SendCRC32C is `true`. The documentation for these fields and on the test indicates that if SendCRC32C is `false` at time of send, the CRC32C value is ignored ([recently added docs](https://github.com/googleapis/google-cloud-go/commit/7c8e2cd5d0a07204d81e2d86634c25332438c43f)). So I don't think we can throw a validation exception. I also don't want to modify the ObjectAttrs from the Writer as it would wipe out whatever the user set there, so doing this clear on the resulting proto object pre-Write is safer imo.

Also, this is basically working around the bug in the service behavior (http://b/249845565). Setting this field should actually be ignored by the service.

noahdietz: Ok so as discussed offline, we shouldn't be copying the ObjectChecksums into a proto Object b.c it is no editable, and we only need to convert from ObjectAttrs to proto Object for mutation RPCs. So I've removed that bit. I also needed to fix how the MD5 hash was sent as it was erroneously being sent only when SendCRC32C was `true`.

tritone (commented):
Ah, looks like the unit test TestObjectAttrsToProtoObject needs to be fixed up as well as per the presubmit.

tritone (approved):

PR #6797, opened 2022-09-30
feat(bigtable): adds readrow grpc to test proxy

telpirion:
This PR is a follow-up to #6568 .


bigtable/internal/testproxy/proxy.go
```diff
@@ -554,7 +555,52 @@ func (s *goTestProxyServer) RemoveClient(ctx context.Context, req *pb.RemoveClie
 // ReadRow responds to the ReadRow RPC. This method gets all of the column
 // data for a single row in the Table.
 func (s *goTestProxyServer) ReadRow(ctx context.Context, req *pb.ReadRowRequest) (*pb.RowResult, error) {
-	return nil, stat.Error(codes.Unimplemented, "ReadRow not implemented")
+	s.clientsLock.RLock()
```
enocom, line 558: In this case, since you're only ready the `clientIDs` map to get a client, you don't need to hold the lock for the duration of the method.

Instead you can do this:

``` golang
s.clientsLock.RLock()
btc, exists := s.clientIDs[req.ClientId]
s.clientsLock.RUnlock()
```

enocom (changes requested):

bigtable/internal/testproxy/proxy.go
telpirion: Done.

enocom (approved):

PR #6802, opened 2022-10-03
fix(internal/godocfx): Use default html rendering

tbpg:
The previous setup was rendering doc headings as H1s. Now that the default GoDoc supports links and lists, we can use it directly.

As a result, we can delete the third_party copy of the go/doc package and the custom goldmark plugin.

`ToMarkdown` was added in #3816.


codyoss (approved):
🥳

PR #6806, opened 2022-10-03
test(storage): convert some bucket update int tests to multitransport

BrennaEpp:
also updates CORS test to use t.Run

### DIFF
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -91,6 +91,7 @@ var (
 )
 
 func TestMain(m *testing.M) {
+	os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", os.Getenv("GCLOUD_TESTS_GOLANG_KEY"))
 	cleanup := initIntegrationTest()
 	cleanupEmulatorClients := initEmulatorClients()
 	exit := m.Run()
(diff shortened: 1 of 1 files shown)


storage/integration_test.go
```diff
@@ -3428,199 +3428,195 @@ func TestIntegration_CancelWrite(t *testing.T) {
...
+		h := testHelper{t}
 
-	bkt := client.Bucket(uidSpace.New())
-	h.mustCreate(bkt, testutil.ProjID(), &BucketAttrs{})
-	defer h.mustDeleteBucket(bkt)
-	attrs := h.mustBucketAttrs(bkt)
-	if attrs.DefaultEventBasedHold != false {
-		t.Errorf("got=%v, want=%v", attrs.DefaultEventBasedHold, false)
-	}
+		bkt := client.Bucket(prefix + uidSpace.New())
```
tritone, line 3504: Do we need to use prefix here if we are only using one bucket?

storage/integration_test.go
```diff
@@ -3428,199 +3428,195 @@ func TestIntegration_CancelWrite(t *testing.T) {
...
-		if diff := testutil.Diff(attrs.CORS, test.want); diff != "" {
-			t.Errorf("input: %v\ngot=-, want=+:\n%s", test.input, diff)
+		} {
+			t.Run(test.desc, func(t *testing.T) {
+				h := testHelper{t}
+
+				bkt := client.Bucket(prefix + uidSpace.New())
+				h.mustCreate(bkt, testutil.ProjID(), &BucketAttrs{CORS: initialSettings})
+				defer h.mustDeleteBucket(bkt)
+				// Set VersioningEnabled attr so we don't fail when setting CORS to nil
```
tritone, line 3489: Maybe clarify that this is so we don't send an empty update/patch request, which is invalid for gRPC

tritone (approved):
nice! two small comments

storage/integration_test.go
BrennaEpp: This allows a diff prefix between http and grpc and is needed right now for grpc.

noahdietz: While the allowlist is in place, we should always create our test buckets with the prefix. It is empty for `http` tests fwiw.

tritone: I was asking why we are using the `prefix` arg to `multiTransportTest` rather than the `bucket` arg. But I think I was confused anyway, we need to create a bucket for this test rather than using the shared one because we want to be modifying the CORS config.

PR #6807, opened 2022-10-04
chore(storage): ignore 0 time when converting to bucket

BrennaEpp:
also enables TestIntegration_CustomTime, TestIntegration_UpdateRetentionPolicy, TestIntegration_DeleteObjectInBucketWithRetentionPolicy, TestIntegration_LockBucket and TestIntegration_LockBucket_MetagenerationRequired


storage/bucket.go
```diff
@@ -1373,7 +1373,7 @@ func toRetentionPolicy(rp *raw.BucketRetentionPolicy) (*RetentionPolicy, error)
 }
 
 func toRetentionPolicyFromProto(rp *storagepb.Bucket_RetentionPolicy) *RetentionPolicy {
-	if rp == nil {
+	if rp == nil || rp.EffectiveTime.AsTime().Unix() == 0 {
```
noahdietz, line 1376:
```suggestion
	if rp == nil || rp.GetEffectiveTime().AsTime().Unix() == 0 {
```

Always use the accessor methods when reading a proto field. It protects us from _some_ breaking changes and also handles the case of the receiver message being `nil` to avoid panics.

BrennaEpp: done

tritone (approved):

PR #6808, opened 2022-10-04
chore(storage): implement gRPC ListObjects read_mask support

noahdietz:
This adds `Query`-to-`FieldMask` conversion for `Object` attributes and incorporates that into `ListObjects` for gRPC.

This also carves out the `ListObjects`-specific tests from `TestIntegration_Objects` and enables those tests for multi-transport testing.


storage/integration_test.go
```diff
@@ -2067,7 +2102,9 @@ func testObjectsIterateAllSelectedAttrs(t *testing.T, bkt *BucketHandle, objects
 		EndOffset:   "obj2",
 	}
 	var selectedAttrs []string
-	for k := range attrToFieldMap {
+	// Iterate over proto fields instead because it doesn't have MediaLink.
+	// TODO(noahdietz): figure out how to handle this better.
+	for k := range attrToProtoFieldMap {
```
noahdietz, line 2107: Is this OK @tritone ? They JSON and gRPC support the same set of fields in their respective Object schema, except gRPC doesn't have a MediaLink field...

storage/storage.go
```diff
@@ -1555,13 +1594,17 @@ var attrToFieldMap = map[string]string{
 // https://cloud.google.com/storage/docs/json_api/v1/how-tos/performance
 func (q *Query) SetAttrSelection(attrs []string) error {
```
tritone, line 1595: I think it might be cleaner to push all the gRPC logic down into toFieldMask, and push the JSON logic down into a new toFieldSelection() func. Then this func can just save the attrs passed in as `q.attrSelection []string` and we don't end up doing any wrong-transport work. What do you think?

tritone (commented):
Thanks for working through this!

storage/storage.go
```diff
@@ -1546,6 +1552,39 @@ var attrToFieldMap = map[string]string{
...
+	"Updated":                 "update_time",
+	"KMSKeyName":              "kms_key",
+	"TemporaryHold":           "temporary_hold",
+	"RetentionExpirationTime": "retention_expire_time",
+	"Metadata":                "metadata",
+	"EventBasedHold":          "event_based_hold",
+	"Owner":                   "owner",
+	"CustomerKeySHA256":       "customer_encryption",
+	"CustomTime":              "custom_time",
+	// TODO(noahdietz): No such field on the Object proto, but is it an HTTP-specific thing?
```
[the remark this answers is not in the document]
noahdietz: Sounds good, I will ask around.

storage/grpc_client.go
```diff
@@ -411,6 +411,12 @@ func (c *grpcStorageClient) ListObjects(ctx context.Context, bucket string, q *Q
 			it.items = append(it.items, b)
 		}
 
+		// Response is always non-nil after a successful request.
+		res := gitr.Response.(*storagepb.ListObjectsResponse)
```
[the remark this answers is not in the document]
noahdietz: Right, this RPC is non-standard in that it has two repeated fields in a paginated method. Automatic pagination is only designed to work on one paginated field, the repeated resource field.

This "Response" accessor was put in for this exact reason, when the response contained non-standard fields that users *might* want access to.

Unfortunately, this is the way it must be done.

storage/storage.go
noahdietz: I agree, it felt dirty doing both in the same spot. Good plan, I'll do that.

storage/grpc_client.go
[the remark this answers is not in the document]
tritone: Sounds good, thanks for clarifying!

storage/storage.go
noahdietz: Ok. Done!

storage/integration_test.go
noahdietz: I might just revert this given the new logic for building a fieldmask...

tritone (approved):
Looks good, thanks Noah!

PR #6809, opened 2022-10-04
chore(storage): multi-transport test for RangeReader

noahdietz:
Enable multi-transport testing for `TestIntegration_ObjectsRangeReader`.

We must temporarily skip gRPC due to a bug in how the StartOffset is reported in the response.

### DIFF
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -893,51 +893,54 @@ func TestIntegration_ConditionalDelete(t *testing.T) {
 }
 
 func TestIntegration_ObjectsRangeReader(t *testing.T) {
-	ctx := context.Background()
-	client := testConfig(ctx, t)
-	defer client.Close()
-	bkt := client.Bucket(bucketName)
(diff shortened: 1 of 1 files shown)


storage/integration_test.go
```diff
@@ -893,51 +893,54 @@ func TestIntegration_ConditionalDelete(t *testing.T) {
...
+				if err != nil {
+					t.Fatalf("Failed to make range read: %v", err)
+				}
+				defer r.Close()
 
-			if got, want := r.Attrs.StartOffset, int64(len(contents))-5; got != want {
-				t.Fatalf("StartOffset mismatch, got %d want %d", got, want)
-			}
+				if got, want := r.Attrs.StartOffset, int64(len(contents))-5; got != want {
+					t.Fatalf("StartOffset mismatch, got %d want %d", got, want)
```
tritone, line 931: I would say that this and subsequent checks should be `Errorf` rather than `Fatalf`, they are not blocking the test from proceeding and make it easier to debug the issue you had with start offset in gRPC.

tritone (approved):
Thanks!

storage/integration_test.go
noahdietz: Ok, changed!

PR #6813, opened 2022-10-06
feat(bigtable): add create table metadata support

kimihrr:
- Add `DeletionProtection` field to `TableConf` and `TableInfo`
- Change `CreateTable` and `CreatePresplitTable` functions to pass `DeletionProtection` equals to `None` to `CreateTableFromConf` which can accept `DeletionProtection` equals to `None`, `Protected` or `Unprotected`
- Add unit tests for `CreateTableFromConf`
- Add integration tests for `DeletionProtection`


bigtable/admin.go
```diff
@@ -373,6 +384,8 @@ type TableInfo struct {
 	// DEPRECATED - This field is deprecated. Please use FamilyInfos instead.
 	Families    []string
 	FamilyInfos []FamilyInfo
+	// DeletionProtection indicates whether the table is protected against data loss
+	DeletionProtection bool
```
kevinsi4508, line 388: Be careful, for NAME_ONLY table view, this field will not be populated in GetTable and ListTables.

We probably should just use DeletionProtection DeletionProtection to be consistent.

We probably should document this a bit here.

bigtable/admin.go
```diff
@@ -421,6 +434,7 @@ func (ac *AdminClient) TableInfo(ctx context.Context, table string) (*TableInfo,
 			FullGCPolicy: gcRuleToPolicy(fam.GcRule),
 		})
 	}
+	ti.DeletionProtection = res.DeletionProtection
```
kevinsi4508, line 437: Add a comment: We expect DeletionProtection in the response because Table_SCHEMA_VIEW is used.

kevinsi4508 (commented):
I will look at your tests tomorrow. Thanks!

bigtable/admin_test.go
```diff
@@ -56,15 +66,34 @@ func setupTableClient(t *testing.T, ac btapb.BigtableTableAdminClient) *AdminCli
 	return c
 }
 
-func TestTableAdmin_UpdateTable(t *testing.T) {
+func TestTableAdmin_CreateTableFromConf(t *testing.T) {
```
kevinsi4508, line 69: I usually use a parameterized test in C++ to test both Protected and Unprotected cases. I am not sure how this can be done in go.

bigtable/integration_test.go
```diff
@@ -1148,6 +1148,73 @@ func TestIntegration_SampleRowKeys(t *testing.T) {
 	}
 }
 
+func TestIntegration_TableDeletionProtection(t *testing.T) {
```
kevinsi4508, line 1151: Add some comments here to explain exactly what we are testing. Thanks!

bigtable/integration_test.go
```diff
@@ -1148,6 +1148,73 @@ func TestIntegration_SampleRowKeys(t *testing.T) {
...
+			"fam2": MaxVersionsPolicy(2),
+		},
+		DeletionProtection: Protected,
+	}
+
+	if err := adminClient.CreateTableFromConf(ctx, &tblConf); err != nil {
+		t.Errorf("create table from config: %v", err)
+	}
+
+	table, err := adminClient.getTable(ctx, myTableName, btapb.Table_FULL)
```
kevinsi4508, line 1183: Please check err. It will give us more details if some is wrong with getTable.

Why can't we use TableInfo instead here?

bigtable/integration_test.go
```diff
@@ -1148,6 +1148,73 @@ func TestIntegration_SampleRowKeys(t *testing.T) {
...
+
+	// Check if the deletion protection works properly
+	if err = adminClient.DeleteColumnFamily(ctx, tblConf.TableID, "fam1"); err == nil {
+		t.Errorf("deletetion protection does not work properly while deleting the column family: %v", err)
+	}
+	if err = adminClient.DeleteTable(ctx, tblConf.TableID); err == nil {
+		t.Errorf("deletetion protection does not work properly while deleting the table: %v", err)
+	}
+
+	updateTblConf := UpdateTableConf{
```
kevinsi4508, line 1197: nit: updateTableConf

bigtable/integration_test.go
```diff
@@ -1148,6 +1148,73 @@ func TestIntegration_SampleRowKeys(t *testing.T) {
...
+	if table.DeletionProtection != true {
+		t.Errorf("Table Deletion Protection is wrong: %v", err)
+	}
+
+	// Check if the deletion protection works properly
+	if err = adminClient.DeleteColumnFamily(ctx, tblConf.TableID, "fam1"); err == nil {
+		t.Errorf("deletetion protection does not work properly while deleting the column family: %v", err)
+	}
+	if err = adminClient.DeleteTable(ctx, tblConf.TableID); err == nil {
+		t.Errorf("deletetion protection does not work properly while deleting the table: %v", err)
```
kevinsi4508, line 1194: ditto

bigtable/admin.go
```diff
@@ -211,18 +211,35 @@ func (ac *AdminClient) Tables(ctx context.Context) ([]string, error) {
 	return names, nil
 }
 
+// DeletionProtection indicates whether the table is protected against data loss
+// i.e. when set to protected, deleting the table, the column families in the table,
+// and the instance containing the table would be prohibited.
+type DeletionProtection int
```
kevinsi4508, line 217: A newbie Go question; is the default value of DeletionProtection None? If not, should we make sure it is?

bigtable/admin_test.go
```diff
@@ -56,18 +65,54 @@ func setupTableClient(t *testing.T, ac btapb.BigtableTableAdminClient) *AdminCli
...
+	c := setupTableClient(t, mock)
+
+	deletionProtection := Protected
+	err := c.CreateTableFromConf(context.Background(), &TableConf{TableID: "My-table", DeletionProtection: deletionProtection})
+	if err != nil {
+		t.Errorf("CreateTableFromConf failed: %v", err)
+	}
+	createTableReq := mock.createTableReq
+	if !cmp.Equal(createTableReq.TableId, "My-table") {
+		t.Errorf("CreateTableRequest does not match, TableID: %v", createTableReq.Table.Name)
```
kevinsi4508, line 79: do you mean: createTableReq.TableId?

Maybe mention what is not matching, in this case the table ID.

Maybe something like this: "t.Errorf("Unexpected table ID in the request: %v, expected %v", createTableReq.TableId, "My-table").

bigtable/admin_test.go
```diff
@@ -56,18 +65,54 @@ func setupTableClient(t *testing.T, ac btapb.BigtableTableAdminClient) *AdminCli
...
+	}
+	if !cmp.Equal(createTableReq.Table.DeletionProtection, true) {
+		t.Errorf("CreateTableRequest does not match, TableID: %v", createTableReq.Table.Name)
```
kevinsi4508, line 82: ditto

kevinsi4508 (approved):
LGTM

bigtable/admin.go
kimihrr: I believe yes. The default is the one that has number equals to 0 which is None in our case.

telpirion: +1 to what Kimia said. Default is 0 (None).

bigtable/admin_test.go
```diff
@@ -56,18 +65,54 @@ func setupTableClient(t *testing.T, ac btapb.BigtableTableAdminClient) *AdminCli
...
+func TestTableAdmin_UpdateTableWithDeletionProtection(t *testing.T) {
 	mock := &mockTableAdminClock{}
 	c := setupTableClient(t, mock)
 	deletionProtection := Protected
 
 	// Check if the deletion protection updates correctly
 	err := c.UpdateTableWithDeletionProtection(context.Background(), "My-table", deletionProtection)
 	if err != nil {
-		t.Errorf("UpdateTable failed: %v", err)
+		t.Errorf("UpdateTableWithDeletionProtection failed: %v", err)
```
telpirion, line 112: Consider using `t.Fatalf()` here.

If `err` here is not `nil`, will the can the next if clause (`cmp.Equal`) return true?

telpirion (approved):
Approved with a single nit.

bigtable/admin_test.go
kevinsi4508: Do you mean if we should end the test with t.Fatalf() because it is meaningless to continue from here if there is an error updating the table?

If yes, then yeah, we should end the test there if err is not nil.

kevinsi4508:
Thanks Eric! appreciate the review!

bigtable/admin_test.go
kimihrr: I thought the same and I changed some of the `Errorf` to `Fatalf`. Thank you so much Eric for the review!

Issue #6816, opened 2022-10-06
storage: storage.(*Writer).Write() with cancelled context causes write race


dragonsinth:
**Client**

Google Cloud Storage

**Code**

run the following test with `-race` enabled

```go
package gcs_test

import (
	"cloud.google.com/go/storage"
	"context"
	"testing"
)

func TestWriteRace(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background())
	cancel()
	cl, err := storage.NewClient(ctx)
	t.Log(err)
	writer := cl.Bucket("foo").Object("bar").NewWriter(ctx)
	_, err = writer.Write([]byte("data"))
	t.Log(err)
	err = writer.Close()
	t.Log(err)
}
```

**Expected behavior**

Either `Write` or `Close` fails with `context.Cancelled`

**Actual behavior**
```
==================
WARNING: DATA RACE
Write at 0x00c0035d9880 by goroutine 58:
  cloud.google.com/go/storage.(*Writer).openWriter()
      go/pkg/mod/cloud.google.com/go/storage@v1.26.0/writer.go:198 +0xa14
  cloud.google.com/go/storage.(*Writer).Write()
      go/pkg/mod/cloud.google.com/go/storage@v1.26.0/writer.go:129 +0x94
  gcs_test.TestWriteRace()
      ./race_test.go:15 +0x534
  testing.tRunner()
      GOROOT/src/testing/testing.go:1446 +0x188
  testing.(*T).Run.func1()
      GOROOT/src/testing/testing.go:1493 +0x40

Previous read at 0x00c0035d9880 by goroutine 59:
  cloud.google.com/go/storage.(*Writer).CloseWithError()
      go/pkg/mod/cloud.google.com/go/storage@v1.26.0/writer.go:225 +0x148
  cloud.google.com/go/storage.(*Writer).monitorCancel()
      go/pkg/mod/cloud.google.com/go/storage@v1.26.0/writer.go:215 +0x13c
  cloud.google.com/go/storage.(*Writer).openWriter.func2()
      go/pkg/mod/cloud.google.com/go/storage@v1.26.0/writer.go:179 +0x34

Goroutine 58 (running) created at:
  testing.(*T).Run()
      GOROOT/src/testing/testing.go:1493 +0x55c
  testing.runTests.func1()
      GOROOT/src/testing/testing.go:1846 +0x90
  testing.tRunner()
      GOROOT/src/testing/testing.go:1446 +0x188
[...]

PR #6817, opened 2022-10-06
fix(storage): fix read-write race in Writer.Write

dragonsinth:
https://github.com/googleapis/google-cloud-go/issues/6816

`w.monitorCancel()` -> `w.CloseWithError()` reads the values of `w.opened` and `w.pw` concurrently with the calling routine initializing them; if the incoming context is cancelled before or during initialization.

The fix is simply to deploy starting the monitor until all fields are initialized.  The older `open()` code from previous releases did this in the right order:

https://github.com/googleapis/google-cloud-go/blob/storage/v1.24.0/storage/writer.go#L130-L134

### DIFF
--- a/storage/writer.go
+++ b/storage/writer.go
@@ -190,6 +190,9 @@ func (w *Writer) openWriter() (err error) {
 		progress:           w.progress,
 		setObj:             func(o *ObjectAttrs) { w.obj = o },
 	}
+	if err := w.ctx.Err(); err != nil {
+		return err // short-circuit
+	}
 	w.pw, err = w.o.c.tc.OpenWriter(params, opts...)
 	if err != nil {
 		return err
--- a/storage/writer.go
+++ b/storage/writer.go
@@ -176,7 +176,6 @@ func (w *Writer) openWriter() (err error) {
 
 	isIdempotent := w.o.conds != nil && (w.o.conds.GenerationMatch >= 0 || w.o.conds.DoesNotExist == true)
 	opts := makeStorageOpts(isIdempotent, w.o.retry, w.o.userProject)
-	go w.monitorCancel()
 	params := &openWriterParams{
 		ctx:                w.ctx,
 		chunkSize:          w.ChunkSize,
@@ -196,6 +195,7 @@ func (w *Writer) openWriter() (err error) {
 		return err
 	}
 	w.opened = true
+	go w.monitorCancel()
 
 	return nil
 }


dragonsinth:
Note that the `w.mu.Lock()`/`w.mu.Unlock()` in `monitorCancel()` doesn't help because
a) it only guards `w.err`
b) `openWriter()` isn't holding the lock anyway

tritone (commented):
Thanks for catching this and for the proposed fix!

I think the spot you have put go `monitorCancel()` is too late. In the old code you linked, we launch `monitorCancel` after opening the pipe but before launching the go routine that actually sends a request. This reduces the likelihood that we would send a request after the context has already been cancelled. In this PR that logic happens inside the `OpenWriter` call on L193.

I think the answer is to pass `monitorCancel` in with the `openWriterParams` and then launch it from within the transport-specific clients and then have a callback to set `Writer.pw`. I can provide more direction on how to do this if you like, or implement it myself. I'd also like to add your repro as a test -- we have another test that checks for a similar condition [here](https://github.com/googleapis/google-cloud-go/blob/64bf56d9779b91b95e861ab96daeca44afbbfc48/storage/writer_test.go#L122) but this one passes because the cancellation happens after the first write I believe.

dragonsinth:
> I think the answer is to pass `monitorCancel` in with the `openWriterParams` and then launch it from within the transport-specific clients and then have a callback to set `Writer.pw`. I can provide more direction on how to do this if you like, or implement it myself. I'd also like to add your repro as a test -- we have another test that checks for a similar condition [here](https://github.com/googleapis/google-cloud-go/blob/64bf56d9779b91b95e861ab96daeca44afbbfc48/storage/writer_test.go#L122) but this one passes because the cancellation happens after the first write I believe.

Huh, I'm unfamiliar with the code underneath `tc.OpenWriter` as that jumps to an interface, but on the surface I wonder if it's really worth breaking through all those layers of abstraction?  And I'm not even sure that actually solves the data race... the race is with `w.pw` and `w.opened`, which don't get assigned until after `OpenWriter` returns, so if `OpenWriter` started up the monitor loop, you'd still run into the same race.

I can think of at least two less-invasive ideas that have the effect of avoiding initiating the call:

1) Explicitly check context immediately before calling OpenWriter, e.g.:

```go
	if err := w.ctx.Err(); err != nil {
		return err // short-circuit
	}
	w.pw, err = w.o.c.tc.OpenWriter(params, opts...)
	if err != nil {
		return err
	}
	w.opened = true
	go w.monitorCancel()
```

This will prevent the extra call in almost all cases.

2) Pass `w.ctx` down to `OpenWriter` and have it respect context internally; this would be the more standard way of doing things, but a more invasive change.

tritone:
@dragonsinth Thanks for the suggestions, agreed that those would both be less complicated. I think your option (1) seems like a good idea; option (2) seems like a little more complexity if the pipe has already been opened and we have to do cleanup that currently occurs in `monitorCancel`. If you want to make that change to this PR I would be happy to approve.

dragonsinth:
> @dragonsinth Thanks for the suggestions, agreed that those would both be less complicated. I think your option (1) seems like a good idea; option (2) seems like a little more complexity if the pipe has already been opened and we have to do cleanup that currently occurs in `monitorCancel`. If you want to make that change to this PR I would be happy to approve.

Sweet, updated this PR.

tritone (approved):
Looks good to me. I'll add a test based on your example in the issue in a separate PR.

dragonsinth:
Awesome, thanks @tritone

Issue #3880, opened 2021-04-01
bigquery: enable transparent storage api usage


derekperkins:
**Is your feature request related to a problem? Please describe.**
Retrieving data from larger queries is painfully slow. In the trace below, the query completed in < 1s, but then it took over 2 minutes to retrieve 180 MB. Typically transferring that inside GCP should take no more than a few seconds, if not sub-second.

**Describe the solution you'd like**
The Python libraries have already enabled usage by default. Not sure if there are any downsides to that, but it seems like it could be an easy win without exposing any more knobs. (just noticed that @shollyman approved it)
https://github.com/googleapis/python-bigquery/commit/e75ff8297c65981545b097f75a17cf9e78ac6772

Benchmark copied from the Python PR https://github.com/googleapis/python-bigquery/pull/55#issuecomment-610352614

![image](https://user-images.githubusercontent.com/3588778/113360674-7d3aed00-9307-11eb-8116-3c7a0c17a7ec.png)

If for some reason it shouldn't be enabled automatically, adding an option to QueryConfig would also be a good option.

**Describe alternatives you've considered**
Using the storage api manually, but that's a completely different paradigm with Avro/Arrow schemas and has a significant mental overhead compared to just using the standard BigQuery package.

**Additional context**

<img width="1187" alt="trace" src="https://user-images.githubusercontent.com/3588778/113360339-d0607000-9306-11eb-9cf6-d4abce647581.png">

derekperkins:
Thinking about it a little more, here's some proposed logic / pseudocode for when to use the storage api. Open to suggestions or clarifications of my assumptions.

```
var useStorageAPI bool
var tableRef *Table
var wg sync.WaitGroup

// the jobs.query response doesn't return the table reference we'll need for the storage api,
// so we should be able to fetch it concurrently while we wait for the query to complete (not tested)
// we may not need it, but it should be a simple call that will often reduce latency 
if jobsQuery {
    wg.Add(1)
    go func() {
      j, err := getJob()
      tableRef = j.QueryConfig.Dst
      wg.Done()
    }()
}

switch {
  // in this case, we either have all the rows or it's one more RPC to get the remainder, so no need
  // for added concurrency / complexity
  case jobsQuery && queryResp.TotalRows <= len(queryResp.Rows) * 2:
    useStorageAPI = false

  // for all other response sizes on jobs.query, use storage api to parallelize it
  case jobsQuery:
    useStorageAPI = true

[...]

PR #6822, opened 2022-10-07
feat(bigquery): use storage api for query jobs

alvarowolfx:
Initial work on using the Storage API for fetching results of a query. This is more efficient because it can download data in parallel by splitting the read session and using Arrow as a more efficient format. The API surface for users stay the same, with them being able to transform query results into user defined structs. Under the hood the library will take care of converting data represented in Arrow to the user defined struct.

One thing to note is that this introduces the first external dependency on the Apache Arrow Go library.

Initially we are gonna use it as an experimental feature and explicit ask users to create a `bqStorage.BigQueryReadClient`. 

Proposed by issue https://github.com/googleapis/google-cloud-go/issues/3880 and work on the Python library https://medium.com/google-cloud/announcing-google-cloud-bigquery-version-1-17-0-1fc428512171


bigquery/storage/reader/client.go
```diff
@@ -0,0 +1,101 @@
...
+	if c.rawClient == nil {
+		return fmt.Errorf("already closed")
+	}
+	c.rawClient.Close()
+	c.rawClient = nil
+	return nil
+}
+
+// NewReader establishes a new reader to fetch data.
+func (c *Client) NewReader(opts ...ReadOption) (*Reader, error) {
```
shollyman, line 77: I somewhat expected a reader to accept a query or table as part of the constructor.  Are you anticipating doing everything through ReadOption passing?

shollyman (commented):
Left a few initial comments

bigquery/storage/reader/arrow.go
```diff
@@ -0,0 +1,195 @@
...
+		arrowSchema:    r.Schema(),
+	}
+	return p, nil
+}
+
+// convertArrowRows converts an Arrow Record Batch into a series of Value slices.
+// schema is used to interpret the data from rows
+func (ap *arrowParser) convertArrowRows(recordBatch *bqStoragepb.ArrowRecordBatch) ([][]bigquery.Value, error) {
+	var rs [][]bigquery.Value
+	buf := bytes.NewBuffer(ap.rawArrowSchema)
```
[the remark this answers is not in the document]
alvarowolfx: Unfortunately that's how the IPC Arrow parser currently works. If the byte buffer that is being parsed doesn't have the schema in the header, it throws this error:
```
arrow/ipc: invalid message type (got=RecordBatch, want=Schema)
```

bigquery/storage/reader/client.go
alvarowolfx: The `Reader` can be reused to read multiple queries, a table or a job. I wasn't sure if was necessary to introduce this as we can make this available through the `Client` struct directly, but by having the `Reader`, it can be configured with the different `ReadOption`, which right now is only limited to the `MaxStreamCount`, but later we can add Row restrictions and Selected columns, that are available when creating a `ReadSession` with TableReadOptions attribute: 
thehttps://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#google.cloud.bigquery.storage.v1.ReadSession.TableReadOptions

bigquery/arrow.go
```diff
@@ -0,0 +1,196 @@
...
+
+// decodeArrowRecords decodes BQ ArrowRecordBatch into a list of arrow.Record.
+func (ap *arrowDecoder) decodeArrowRecords(serializedArrowRecordBatch []byte) ([]arrow.Record, error) {
+	buf := bytes.NewBuffer(ap.rawArrowSchema)
+	buf.Write(serializedArrowRecordBatch)
+	r, err := ipc.NewReader(buf, ipc.WithAllocator(ap.mem), ipc.WithSchema(ap.arrowSchema))
+	if err != nil {
+		return nil, err
+	}
+	records := []arrow.Record{}
```
shollyman, line 63: If we're not exposing the arrow iterator to users, does it make sense to convert to bigquery values in the decoder here? It looks like you manage the Retain / Release lifecycle of the arrow resources in a variety of places.

Does immediate conversion obviate that complexity?

bigquery/bigquery.go
```diff
@@ -97,6 +98,17 @@ func NewClient(ctx context.Context, projectID string, opts ...option.ClientOptio
 	return c, nil
 }
 
+// ConfigureStorageReadClient sets up Storage API connection to be used when fetching
+// large datasets from tables, jobs or queries.
+func (c *Client) ConfigureStorageReadClient(ctx context.Context, opts ...option.ClientOption) error {
```
shollyman, line 103: Should we just accept an instantiated ReadClient?  You've already exposed NewReadClient().

Additionally, should we block re-configuration?  e.g. once c.rc != nil we emit error on subsequent attempts?  As presented, we'd potentially leak resources from the previous client.

bigquery/integration_test.go
```diff
@@ -54,6 +54,7 @@ var record = flag.Bool("record", false, "record RPCs")
 
 var (
 	client                 *Client
+	sClient                *Client
```
shollyman, line 57: maybe make this name less ambiguous, since we already track a storage (GCS) client in here as well?

bigquery/query.go
```diff
@@ -149,6 +149,9 @@ type QueryConfig struct {
 	// Experimental: this option is experimental and may be modified or removed in future versions,
 	// regardless of any other documented package stability guarantees.
 	JobTimeout time.Duration
+
+	// Force usage of Storage API if client is available.
+	ForceStorageAPI bool
```
shollyman, line 154: Should this be in QueryConfig, or should it be on Client?  Alternatively, should we add this as a variadic option on our Read() func in tables/jobs/queries?

bigquery/storage_session.go
```diff
@@ -0,0 +1,190 @@
...
+	settings *settings
+	rc       *ReadClient
+
+	ctx     context.Context
+	table   *Table
+	tableID string
+
+	bqSession *bqStoragepb.ReadSession
+
+	// SerializedArrowSchema is an IPC-serialized Arrow Schema.
```
shollyman, line 41: Rather than exposing all these mutable things on the ReadSession, should it instead just have a method which returns a proto.Clone if the retained ReadSession?

bigquery/storage_session.go
```diff
@@ -0,0 +1,190 @@
...
+	MaxStreamCount int
+}
+
+func defaultSettings() *settings {
+	return &settings{
+		MaxStreamCount: 0,
+	}
+}
+
+// Run initiates a read session
```
shollyman, line 81: Should we instead expose a `NewReadSession()` that handles all this setup as part of instantiation?

bigquery/storage_session.go
alvarowolfx: I like the idea of returning a clone to avoid users messing with the attributes as they are used locally, but should we return our own `ReadSession` or the `bqStoragepb.ReadSession`  ? The second option would give access to internal proto stubs from the Storage API.
```
bqStoragepb "google.golang.org/genproto/googleapis/cloud/bigquery/storage/v1"
```

bigquery/storage_session.go
alvarowolfx: I added this `Run` method so if users want to lazy initialize the session they can, similarly on how our Query struct works, that users can call `Run` and get a `Job` and `Read` it later, or just call `query.Read` directly. In this case here, it users do not call run, we initialise it on the first `Next` call

bigquery/integration_test.go
alvarowolfx: yeah, I didn't liked the name too. Will make it more explicit.

bigquery/bigquery.go
alvarowolfx: Good point, I added a check to see if the client is already initialized here. On passing a `ReadClient` or not, I wasn't sure how much to expose users to that. I was trying to address two groups of users: 

1. Users that just want to enable the Storage API to get faster downloads and don't worry that much about details ( they don't need to know that the `ReadClient` exists for example )
2. Users that know about the Storage API and want to use the ReadClient, created a ReadSession and then read rows from it.

bigquery/arrow.go
alvarowolfx: Initially when the `ArrowIterator` and the `RowIterator` were one thing, the `arrow.Record` to `[][]bigquery.Value` conversion was happening on the same struct, but as this was divided, it was required to call Retain/Release ( which right now each one is only called once ).

Changing. this back will change some concepts, as the ArrowIterator maintains a `chan arrow.Record` and it's `Next` method blocks/waits for new items to arrive on that channel, while the `RowIterator` works by dequeueing a row from the `rows`  attribute. Having that separating makes the changes required on the `RowIterator` less invasive, at the cost of having to keep an eye on those memory reference counts.

bigquery/storage_session.go
shollyman: Yeah, we could certainly return a custom type with the exported fields as well.

bigquery/query.go
```diff
@@ -374,19 +377,31 @@ func (q *Query) Read(ctx context.Context) (it *RowIterator, err error) {
...
 	}
+
 	// we have a config, run on fastPath.
 	resp, err := q.client.runQuery(ctx, queryRequest)
 	if err != nil {
 		return nil, err
 	}
+
+	// If more pages are available, use the Storage API instead
+	if resp.PageToken != "" && q.client.rc != nil {
```
shollyman, line 396: This looks incorrect.    As written, this appears to force the user to re-run the job a second time if its a multi page result and complete.  inside the jobComplete check, we should resolve which iterator to use.

I don't think we get dest table in the query response, so we may need to pay the cost of a jobs.get call here.

bigquery/storage_iterator.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	it.decoder = decoder
+
+	go func() {
+		it.wg.Wait()
+		close(it.records)
+		it.done = true
+	}()
+
+	for _, readStream := range streams {
```
shollyman, line 170: You may want a semaphore here rather than spinning up N goroutines immediately for each reported stream, so you only have so many workers in flight at once.  Work through the list with a smaller number of active processStream instances

bigquery/query.go
```diff
@@ -374,19 +377,31 @@ func (q *Query) Read(ctx context.Context) (it *RowIterator, err error) {
 	ctx = trace.StartSpan(ctx, "cloud.google.com/go/bigquery.Query.Run")
 	defer func() { trace.EndSpan(ctx, err) }()
 	queryRequest, err := q.probeFastPath()
-	if err != nil {
+	if err != nil || q.forceStorageAPI {
```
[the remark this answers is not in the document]
alvarowolfx: Indeed was misplaced. For testing purposes, this flag is used to make Queries take a shortcut and use the Job.Read path and use the Storage API to fetch data.

bigquery/storage_client.go
```diff
@@ -0,0 +1,156 @@
...
+		rawClient:      rawClient,
+		projectID:      projectID,
+		maxWorkerCount: maxWorkerCount,
+		maxStreamCount: 0,
+	}
+
+	return rc, nil
+}
+
+// close releases resources held by the client.
```
shollyman, line 76: nit: case mismatch between method and comment

bigquery/storage_client.go
```diff
@@ -0,0 +1,156 @@
...
+func (c *readClient) Close() error {
+	if c.rawClient == nil {
+		return fmt.Errorf("already closed")
+	}
+	c.rawClient.Close()
+	c.rawClient = nil
+	return nil
+}
+
+// SessionForTable establishes a new session to fetch from a table using the Storage API
```
shollyman, line 86: same nit re: case consistency between comment/method

bigquery/storage_integration_test.go
```diff
@@ -0,0 +1,370 @@
...
+		err := it.Next(&s)
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			t.Fatalf("failed to fetch via storage API: %v", err)
+		}
+		i++
+	}
+	t.Logf("%d lines read", i)
```
shollyman, line 243: assuming all the t.Logf is temp debugging?

bigquery/storage_integration_test.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+	if !qRowIt.IsAccelerated() {
+		t.Fatalf("reading query should use Storage API")
+	}
+	if err = checkRowsRead(qRowIt, expectedRows); err != nil {
+		t.Fatalf("checkRowsRead(query): %v", err)
+	}
+}
+
+func TestIntegration_StorageReadQueryOrdering(t *testing.T) {
```
shollyman, line 210: I feel like this may end up more concise as a table-driven test, testcases have a test query and expectations (expected num streams, rows, etc) and you run through each of them.

bigquery/storage_integration_test.go
```diff
@@ -0,0 +1,370 @@
...
+
+	for {
+		var dst []Value
+		err := it.Next(&dst)
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			if errors.Is(err, context.DeadlineExceeded) {
+				continue
```
shollyman, line 362: This seems odd, but I may be missing something here.  You keep looping if context expires?

bigquery/storage_client.go
```diff
@@ -0,0 +1,156 @@
...
+func (rs *readSession) start() error {
+	tableReadOptions := &bqStoragepb.ReadSession_TableReadOptions{
+		SelectedFields: []string{},
+	}
+	createReadSessionRequest := &bqStoragepb.CreateReadSessionRequest{
+		Parent: fmt.Sprintf("projects/%s", rs.table.ProjectID),
+		ReadSession: &bqStoragepb.ReadSession{
+			Table:       rs.tableID,
+			DataFormat:  bqStoragepb.DataFormat_ARROW,
+			ReadOptions: tableReadOptions,
```
shollyman, line 131: Shouldn't need to specify/populate ReadOptions, you're not doing partial projection.

bigquery/storage_iterator.go
```diff
@@ -0,0 +1,267 @@
...
+
+	decoder, err := newArrowDecoderFromSession(it.session, it.schema)
+	if err != nil {
+		return err
+	}
+	it.decoder = decoder
+
+	wg := sync.WaitGroup{}
+	sem := semaphore.NewWeighted(int64(it.session.rc.maxWorkerCount))
+	go func() {
```
shollyman, line 166: I think this may race?  wg.Wait() before the first wg.Add will proceed, since there's nothing outstanding?  You could potentially do something like wg.Add(len(streams)) before starting up this and your reader goroutines.

bigquery/storage_iterator.go
```diff
@@ -0,0 +1,267 @@
...
+				it.processStream(readStreamName)
+				sem.Release(1)
+				wg.Done()
+			}(readStream.Name)
+		}
+	}()
+	return nil
+}
+
+func (it *arrowIterator) processStream(readStream string) {
```
shollyman, line 193: Should this (or consumeRowStream) allow for retries before propagating error?

bigquery/storage_iterator.go
alvarowolfx: good point, I changed it here.

bigquery/storage_iterator.go
```diff
@@ -0,0 +1,265 @@
...
+	wg := sync.WaitGroup{}
+	sem := semaphore.NewWeighted(int64(it.session.rc.maxWorkerCount))
+	go func() {
+		wg.Wait()
+		close(it.records)
+		close(it.errs)
+		it.done = true
+	}()
+
+	go func() {
```
shollyman, line 173: Dispatching in a goroutine is likely unnecessary, but the inner goroutine per processor is.

bigquery/storage_client.go
```diff
@@ -0,0 +1,159 @@
...
+	createReadSessionRequest := &storagepb.CreateReadSessionRequest{
+		Parent: fmt.Sprintf("projects/%s", rs.table.ProjectID),
+		ReadSession: &storagepb.ReadSession{
+			Table:      rs.tableID,
+			DataFormat: storagepb.DataFormat_ARROW,
+		},
+		MaxStreamCount: int32(rs.settings.maxStreamCount),
+	}
+	rpcOpts := gax.WithGRPCOptions(
+		grpc.MaxCallRecvMsgSize(1024 * 1024 * 129), // TODO: why needs to be of this size
```
shollyman, line 139: Default is 4MB: https://github.com/grpc/grpc-go/blob/07ac97c355970d6210cfa5341f2ba0850ec2757b/clientconn.go#L96

Read API can send batches up to 128MB.

bigquery/storage_iterator.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func newStorageRowIteratorFromJob(ctx context.Context, job *Job, totalRows uint64) (*RowIterator, error) {
+	cfg, err := job.Config()
+	if err != nil {
+		return nil, err
+	}
+	qcfg := cfg.(*QueryConfig)
+	if qcfg.Dst == nil {
+		// TODO: script job ?
```
shollyman, line 72: This still needs addressed.  Easy to exercise with a test case using multiple statement execution e.g `SELECT 1 as foo; SELECT foo, bar, baz from sometable;`

bigquery/storage_iterator.go
shollyman: Current implementation is appropriate for a unary call, but could be refined to something more stream oriented (e.g. fail after N consecutive attempts, reset attempts when we successfully request).

bigquery/storage_iterator_test.go
```diff
@@ -0,0 +1,128 @@
...
+		},
+		{
+			ctx:  cancelledCtx,
+			desc: "context cancelled/deadline exceeded",
+			errors: []error{
+				fmt.Errorf("random error"),
+				fmt.Errorf("another random error"),
+				fmt.Errorf("yet another random error"),
+			},
+			wantFail: false,
```
shollyman, line 71: Should this not be a failure?  If you want to validate which error the user sees between context and explicit errors, your testcase struct probably should capture a final error that can be compared.

bigquery/storage_iterator.go
```diff
@@ -0,0 +1,309 @@
...
+			}
+			if err != nil {
+				return nil, fmt.Errorf("failed to resolve table for script job: %w", err)
+			}
+			childJobs = append(childJobs, job)
+		}
+		if len(childJobs) == 0 {
+			return nil, fmt.Errorf("failed to resolve table for script job: no child jobs found")
+		}
+		lastJob := childJobs[0]
```
shollyman, line 87: Per our chat, the correct thing may be to look for the last job with a destination table.  Consider the case of multiple statements where the last statement isn't a SELECT.

shollyman (approved):
I think this is ready to ship, once you address the remaining items.  Left one more comment as a holdover from our chat today about selecting the correct statement from a multi-statement execution (script) where there's different types of statements.

bigquery/internal/query/order.go
```diff
@@ -0,0 +1,67 @@
...
+type qnode struct {
+	query    string
+	children []*qnode
+}
+
+// HasOrderedResults checks if a given SQL query returns ordered results.
+// This function uses a naive approach of checking the root level query
+// ( ignoring subqueries, functions calls, etc ) and checking
+// if it contains an ORDER BY clause.
+func HasOrderedResults(sql string) bool {
```
shollyman, line 30: I think this is a reasonable approach, though something like unmatched parens in an inline comment might muck this up.  To get more right requires more SQL parsing than we'd want to do locally, and the resolution is fairly simple in these instances.

shollyman (approved):

shollyman (approved):

PR #6827, opened 2022-10-07
chore(storage): enable several tests for grpc + related fixes

BrennaEpp:
Enables tests: BucketPolicyOnly, ACL, PredefinedACLs

Fixes: gets all fields in GetBucket and GetObject (were missing ACLs); makes sure ACL is sent when ACL is not sent but PredefinedACL is; initializes Bucket_IamConfig (was causing segfault); sends UBLA field when disabled in an update

Additionally cleans up acl checking in tests to use the same method


storage/bucket.go
```diff
@@ -907,23 +907,27 @@ func (ua *BucketAttrsToUpdate) toProtoBucket() *storagepb.Bucket {
...
-			Enabled: true,
+	if ua.UniformBucketLevelAccess != nil || ua.BucketPolicyOnly != nil || ua.PublicAccessPrevention != PublicAccessPreventionUnknown {
+		bktIAM = &storagepb.Bucket_IamConfig{}
+
+		if ua.BucketPolicyOnly != nil {
+			bktIAM.UniformBucketLevelAccess = &storagepb.Bucket_IamConfig_UniformBucketLevelAccess{
+				Enabled: optional.ToBool(ua.BucketPolicyOnly.Enabled),
+			}
+		}
+		if ua.UniformBucketLevelAccess != nil { //takes precedence over BucketPolicyOnly
```
tritone, line 920: Nit: move the comment above the code.

storage/grpc_client.go
```diff
@@ -522,7 +521,7 @@ func (c *grpcStorageClient) UpdateObject(ctx context.Context, bucket, object str
 	}
 	// Note: This API currently does not support entites using project ID.
 	// Use project numbers in ACL entities. Pending b/233617896.
-	if uattrs.ACL != nil {
+	if uattrs.ACL != nil || len(uattrs.PredefinedACL) > 0 {
```
tritone, line 524: I'm surprised that this is necessary for PredefinedACL given how it is set in L485

tritone (approved):
Nice fixes!

storage/grpc_client.go
BrennaEpp: This line is to make sure the update mask contains something so we can send the request

PR #6828, opened 2022-10-07
feat(storage): add Autoclass support

### LINKED ISSUE #6592: storage: add support for Autoclass
Autoclass feature at the bucket level, allows for the automatic selection of the best storage class based on object metadata and access patterns. Currently, Autoclass supports  **new buckets** only - a bucket can be enabled for Autoclass only at creation, once disabled it cannot be enabled again. See [go/gcs-dpe-autoclass](http://go/gcs-dpe-autoclass) (internal) for more details.

**System tests are only required in languages with code changes other than documentation:**
- Autoclass can be enabled via Bucket `Create`
- Autoclass can be disabled via Bucket `Patch` and `Update` operations
- Bucket `Get` operations should return the correct state of Autoclass on the bucket which includes the enabled boolean field and toggleTime timestamp field

Please wait for approval from myself before merging/releasing so we can coordinate around release dates.

cojenco:
Add support for Autoclass, which allows for automatic selection of the best storage class based on object access patterns

Fixes #6592


storage/bucket.go
```diff
@@ -1892,6 +1925,52 @@ func customPlacementFromProto(c *storagepb.Bucket_CustomPlacementConfig) *Custom
...
+		Enabled: a.Enabled,
+	}
+}
+
+func toAutoclass(a *raw.BucketAutoclass) *Autoclass {
+	if a == nil {
+		return nil
+	}
+	t, err := time.Parse(time.RFC3339, a.ToggleTime)
+	if err != nil {
```
tritone, line 1953: This is a bit confusing, so maybe add a comment that we are only parsing `ToggleTime` if available. We should also check in integration tests that it is parsed successfully (ie that it is non-zero).

storage/bucket.go
```diff
@@ -1079,6 +1103,10 @@ type BucketAttrsToUpdate struct {
 	// more information.
 	RPO RPO
 
+	// If set, updates the autoclass configuration of the bucket.
+	// See <TBD> for more information.
```
tritone, line 1107: Remember to fill this in before we merge.

storage/bucket.go
```diff
@@ -1892,6 +1925,52 @@ func customPlacementFromProto(c *storagepb.Bucket_CustomPlacementConfig) *Custom
 	return &CustomPlacementConfig{DataLocations: c.GetDataLocations()}
 }
 
+func (a *Autoclass) toRawBucketAutoclass() *raw.BucketAutoclass {
```
tritone, line 1928: Can leave "bucket" out of these method names (so `toRawAutoclass` and `toProtoAutoclass`)

storage/bucket.go
```diff
@@ -1892,6 +1925,52 @@ func customPlacementFromProto(c *storagepb.Bucket_CustomPlacementConfig) *Custom
...
+	if a == nil {
+		return nil
+	}
+	// Excluding read only field ToggleTime.
+	return &storagepb.Bucket_Autoclass{
+		Enabled: a.Enabled,
+	}
+}
+
+func toAutoclass(a *raw.BucketAutoclass) *Autoclass {
```
tritone, line 1948: Let's say `toAutoclassFromRaw` (mirroring customPlacement)

storage/bucket_test.go
```diff
@@ -60,6 +60,7 @@ func TestBucketAttrsToRawBucket(t *testing.T) {
 		Encryption: &BucketEncryption{DefaultKMSKeyName: "key"},
```
tritone, line 60: Let's make sure we have test coverage for the proto converters as well. Also, you should make an integration test to cover the roundtrip with the service (both create and update).

tritone (commented):
Nice start on this Cathy!

storage/bucket.go
cojenco: Comment added. I also updated the logic to match how RetentionPolicy handles valid/non-zero timestamps, per https://github.com/googleapis/google-cloud-go/pull/6807

storage/bucket.go
cojenco: done ;)

storage/bucket.go
cojenco: done

storage/bucket_test.go
cojenco: Yep was meaning to write an integration test... also added 2 tests for the proto converters. @tritone PTAL, thanks!

tritone: Thanks for adding these additional tests; didn't realize we were missing these unit tests for proto converter methods. We should consider auditing and filling others in if they are missing. @noahdietz @BrennaEpp

storage/bucket.go
tritone: Hmm, I actually think we should avoid returning an error if `ToggleTime` is not present. In general we want to avoid returning errors when the response from the server doesn't look exactly like we would expect; IMO it is to ignore the field rather than have this client-side validation on the response, in case the API is later updated in a way we don't expect.

storage/bucket_test.go
```diff
@@ -710,6 +712,322 @@ func TestNewBucket(t *testing.T) {
 	}
 }
 
+func TestNewBucketFromProto(t *testing.T) {
```
tritone, line 715: Thanks very much for adding this! We should have done it sooner.

storage/integration_test.go
```diff
@@ -855,6 +855,47 @@ func TestIntegration_PublicAccessPrevention(t *testing.T) {
...
+
+	// Get Autoclass configuration from bucket attrs.
+	attrs, err := bkt.Attrs(ctx)
+	if err != nil {
+		t.Fatalf("get bucket attrs failed: %v", err)
+	}
+	var toggleTime time.Time
+	if attrs != nil && attrs.Autoclass != nil {
+		if got, want := attrs.Autoclass.Enabled, true; got != want {
+			t.Fatalf("attr.Autoclass.Enabled = %v, want %v", got, want)
```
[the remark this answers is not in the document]
cojenco: Done, thanks!

storage/bucket.go
cojenco: Agreed that we want to avoid client-side validations on the response. I've updated the converter so that the field is ignored in those cases.

We'd want to revisit how we're handling the read only timestamp for RetentionPolicy as well. Looking at our code base, that requires a bigger scope of change, so I'm separating that into an independent tracking issue.

tritone (approved):
One test comment, otherwise LGTM pending feature release.

PR #6835, opened 2022-10-11
feat(pubsub): add support for snapshot labels

### LINKED ISSUE #6838: pubsub: support labels for snapshots
Cloud Labels are available for topics/subscriptions but not snapshots 

Add a way to set labels for snapshots

Provide a `CreateSnapshotWithConfig` and `UpdateSnapshot` method to match topic/subscription structs but since we can't meaningfully update snapshots, it's simpler to just have a single method to update snapshot labels.

**Additional context**

hongalex:
The way to update labels for snapshots will be a bit different than for topic/subscriptions. You need to call `snapshot.SetLabels` after one has been created. 

I decided against adding labels through a new `UpdateSnapshot` method since snapshots cannot be meaningfully modified after creation. In addition, `subscription.CreateSnapshot` doesn't support SnapshotConfigs so  supporting that would expand the API surface. Not adding this is consistent with the UI in Cloud Console where snapshots can be labelled only after creation.

Closes #6838


pubsub/snapshot.go
```diff
@@ -42,11 +43,32 @@ func (s *Snapshot) ID() string {
 	return s.name[slash+1:]
 }
 
+// SetLabels replaces the current set of labels completely with the new set.
```
shollyman, line 46: Maybe "SetLabels set or replaces the labels on a given snapshot."

The existing docstring seems overly focused on the replace case.

pubsub/snapshot.go
```diff
@@ -42,11 +43,32 @@ func (s *Snapshot) ID() string {
...
+func (s *Snapshot) SetLabels(ctx context.Context, label map[string]string) (*SnapshotConfig, error) {
+	var fields []string
+	fields = append(fields, "labels")
+	sc, err := s.c.subc.UpdateSnapshot(ctx, &pb.UpdateSnapshotRequest{
+		Snapshot: &pb.Snapshot{
+			Name:   s.name,
+			Labels: label,
+		},
+		UpdateMask: &fmpb.FieldMask{
+			Paths: fields,
```
shollyman, line 56: you don't appear to manipulate the fields in the mask, maybe just declare the paths using a literal ex:

`Paths: []string{"labels"},`

shollyman (approved):

pubsub/snapshot.go
```diff
@@ -43,17 +43,15 @@ func (s *Snapshot) ID() string {
 	return s.name[slash+1:]
 }
 
-// SetLabels replaces the current set of labels completely with the new set.
+// SetLabels set or replaces the labels on a given snapshot.
```
shollyman, line 46: set -> sets?  Clearly I am a dubious provider of docstring advice.

PR #6853, opened 2022-10-12
chore(storage): delete and migrate grpc specific tests

noahdietz:
Some previously added gRPC media operation integration tests are now covered by the well established JSON via the `multiTransportTest` helper. This removes several of those now redundant gRPC-specific tests.

It also "migrates" a few gRPC-specific tests that we'd like to keep and updates them to use more regular testing helpers. These are necessary to exercise transport-specific specific scenarios with regards to stream management.


storage/integration_test.go
```diff
@@ -1372,66 +1037,53 @@ func TestIntegration_MultiMessageWriteGRPC(t *testing.T) {
 }
 
 func TestIntegration_MultiChunkWriteGRPC(t *testing.T) {
-	ctx := context.Background()
+	multiTransportTest(skipHTTP("gRPC implementation specific test"), t, func(t *testing.T, ctx context.Context, bucket string, _ string, client *Client) {
```
tritone, line 1040: IMO this test should be dual-transport probably. It will also be multi-chunk for HTTP via the resumable upload API, and I don't see a similar test for HTTP currently.

tritone (commented):
Thanks Noah!

storage/integration_test.go
noahdietz: Ok! Great, done.

tritone (approved):

PR #6854, opened 2022-10-12
chore(logging): update blunderbuss assignments

daniel-sanche:
update blunderbuss config to send logging and error-reporting reviews to the more limited `googleapis/api-logging-reviewers`, rather than `googleapis/api-logging`

### DIFF
--- a/.github/blunderbuss.yml
+++ b/.github/blunderbuss.yml
@@ -25,4 +25,4 @@ assign_issues_by:
   - 'api: logging'
   - 'api: clouderrorreporting'
   to:
-  - googleapis/api-logging
+  - googleapis/api-logging-reviewers


.github/blunderbuss.yml
```diff
@@ -25,4 +25,4 @@ assign_issues_by:
   - 'api: logging'
   - 'api: clouderrorreporting'
   to:
-  - googleapis/api-logging
+  - googleapis/api-logging-reviewers
```
codyoss, line 28: Feels a little odd that this is different than what is in CODEOWNERS, should that be updated too?

codyoss (approved):
LGMT, just a small question

daniel-sanche:
closing this - we ended up changing the membership of `googleapis/api-logging` rather than creating a new reviewers group

PR #6858, opened 2022-10-13
test(storage): migrate SignedURL tests

tritone:
Migrate Signed URL integration tests to use multiTransportTest.

Similarly to Post Policy, I had to break up one test that uses two clients. Also, the _WithCreds test currently is skipped for gRPC, pending a fix to the constructor to capture credentials as we do for HTTP.


noahdietz (approved):

storage/integration_test.go
```diff
@@ -4404,74 +4389,103 @@ func TestIntegration_Scopes(t *testing.T) {
...
-			},
-			client: clientWithCredentials,
-		},
-		{
-			desc: "signing with the default sign bytes func",
-			opts: SignedURLOptions{
-				Method:         "GET",
-				Expires:        time.Now().Add(30 * time.Second),
-				GoogleAccessID: jwt.Email,
+		for _, test := range []struct {
```
BrennaEpp, line 4460: nit: this can probably be removed as we can only really have one client per test now, right? So we would never have more than one case here (same for the previous test).

storage/integration_test.go
```diff
@@ -4404,74 +4389,103 @@ func TestIntegration_Scopes(t *testing.T) {
...
+}
 
-	scopes := []string{ScopeFullControl, "https://www.googleapis.com/auth/cloud-platform"}
-	ts := testutil.TokenSource(ctx, scopes...)
-	if ts == nil {
-		t.Fatalf("Cannot get token source to create client")
-	}
-	clientWithoutPrivateKey := testConfig(ctx, t, option.WithTokenSource(ts))
-	defer clientWithoutPrivateKey.Close()
+func TestIntegration_SignedURL_Default(t *testing.T) {
```
BrennaEpp, line 4443: This only tests the "default" under certain auth conditions - I would update this name to be more specific, maybe something like:

```suggestion
func TestIntegration_SignedURL_DefaultSignBytes(t *testing.T) {
```

storage/integration_test.go
```diff
@@ -4404,74 +4389,103 @@ func TestIntegration_Scopes(t *testing.T) {
 
 }
 
-func TestIntegration_SignedURL_Bucket(t *testing.T) {
+func TestIntegration_SignedURL_WithCreds(t *testing.T) {
 	ctx := context.Background()
 
-	if testing.Short() && !replaying {
-		t.Skip("Integration tests skipped in short mode")
+	// By default we are authed with a token source, so don't have the context to
```
BrennaEpp, line 4395: I think this is no longer true, but we should send in explicit credentials still in case the way we authenticate by default changes.

storage/integration_test.go
```diff
@@ -4404,74 +4389,103 @@ func TestIntegration_Scopes(t *testing.T) {
...
+				if err != nil {
+					t.Fatalf("unable to create signed URL: %v", err)
+				}
+
+				if err := verifySignedURL(url, nil, contents); err != nil {
+					t.Fatalf("problem with the signed URL: %v", err)
+				}
+			})
+		}
+	})
```
BrennaEpp, line 4487: I think we need to pass in an option here to authenticate with a token source to test the sign bytes func here, since we removed that as the default authentication method for these tests (see comment above).

BrennaEpp (changes requested):
Thanks Chris! I think there is an issue with making sure we are authenticated the way we want to ensure we are testing the right thing, but otherwise LGTM

meredithslota:
@danielduhh — can you find someone to finish up this PR or close if it's no longer needed?

tritone:
> @danielduhh — can you find someone to finish up this PR or close if it's no longer needed?

I'll finish this up and merge it this week, sorry for the delay.

tritone:
Actually, I'm going to close and re-open due to merge conflicts.

PR #6861, opened 2022-10-13
chore(bigtable): adds CloseClient() to test proxy

telpirion:


bigtable/internal/testproxy/proxy.go
```diff
@@ -493,7 +494,17 @@ type goTestProxyServer struct {
 	pb.UnimplementedCloudBigtableV2TestProxyServer
 	clientsLock sync.RWMutex           // clientsLock prevents simultaneous mutation of the clientIDs map
 	clientIDs   map[string]*testClient // clientIDs has all of the bigtable.Client objects under test
+}
 
+func (s *goTestProxyServer) client(clientID string) (*testClient, bool) {
```
enocom, line 499: A little comment that this should always be called after locking `clientsLock` would be good.

enocom (approved):

bigtable/internal/testproxy/proxy.go
telpirion: Done.

PR #6862, opened 2022-10-13
test(storage): add test for Writer race

tritone:
Adds a test to detect a potential race condition that can arise
if the context is cancelled before writing to GCS. We already
have a test that detects a similar condition after we have
already started writing the object, but this test looks at the context
being cancelled before any writes occur.

When run with -race, this test fails in storage/v1.27.0, but it
passes when run with the fix in #6817.

Updates #6816

FYI @dragonsinth


dragonsinth (approved):
Nice!

noahdietz (approved):

PR #6863, opened 2022-10-14
fix(storage): fix request token passing for Copier.Run

### LINKED ISSUE #6857: storage: CopierFrom large files, progress resets and copy never finishes
We are copying a large file from a regional bucket to a multi-regional bucket, and seeing some strange behaviors. 

- If you do not specify the context timeout, the copier goes into an endless error/retry. Before we noticed this issue, we went through a Petabyte of egress in 1-2 weeks.
- We then added a timeout, copying continues to retry until it reaches the timeout period. 
- We also added a retry strategy of "never" but it still to continues to retry.

**Client**
GCS

**Environment**
golang:1.19.1-alpine 

**Go Environment**
1.19.1

**Code**

```go

func newGCPClient(ctx context.Context) (*storage.Client, error) {
	options := []option.ClientOption{option.WithCredentialsJSON([]byte(config.Env.GoogleCredentials))}
	return storage.NewClient(ctx, options...)
}

func CopyFile(fromBucket string, fromKey string, toBucket string, toKey string) error {
	ctx := context.Background()
	client, err := newGCPClient(ctx)
	if err != nil {
		return err
	}
	defer client.Close()

	src := client.Bucket(fromBucket).Object(fromKey)
	dst := client.Bucket(toBucket).Object(toKey)

	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
	defer cancel()

	copier := dst.CopierFrom(src)
	copier.ProgressFunc = func(copiedBytes, totalBytes uint64) {
		fmt.Println(fmt.Sprintf("progress: %d / %d", copiedBytes, totalBytes))
	}
	if _, err := copier.Run(ctx); err != nil {
		return err
	}
	return nil
}

func main() {
	config.LoadVariables()
	fromBucket := ""
	fromKey := ""
	toBucket := ""
	toKey := ""
	if err := CopyFile(fromBucket, fromKey, toBucket, toKey); err != nil {
		fmt.Println("ERROR: ", err)
		return
	}
	fmt.Println("SUCCESS")
}

```

**Expected behavior**
Progress increments and eventually finishes

**Actual behavior**
Progress seems to reset and it never finishes

Here are logs of the progress:
```
progress: 1862270976 / 8209196233
progress: 1904214016 / 8209196233
progress: 1929379840 / 8209196233
progress: 2231369728 / 8209196233
progress: 2038431744 / 8209196233
progress: 1551892480 / 8209196233
progress: 2298478592 / 8209196233
progress: 1929379840 / 8209196233
progress: 2449473536 / 8209196233
progress: 2021654528 / 8209196233
ERROR:  context deadline exceeded
Exiting.

```

Also, if we run the same code on two GCS buckets that are both multi-regional. The copy happens almost immediately.

Here are the logs
```
progress: 8209196233 / 8209196233
SUCCESS
Exiting.
```

BrennaEpp:
This fixes infinite retry issues.

Will follow up with tests to catch this behaviour.

Fixes #6857

### DIFF
--- a/storage/copy.go
+++ b/storage/copy.go
@@ -126,6 +126,7 @@ func (c *Copier) Run(ctx context.Context) (attrs *ObjectAttrs, err error) {
 		if err != nil {
 			return nil, err
 		}
+		c.RewriteToken = res.token
 		req.token = res.token
 		if c.ProgressFunc != nil {
 			c.ProgressFunc(uint64(res.written), uint64(res.size))
--- a/storage/copy.go
+++ b/storage/copy.go
@@ -126,7 +126,7 @@ func (c *Copier) Run(ctx context.Context) (attrs *ObjectAttrs, err error) {
 		if err != nil {
 			return nil, err
 		}
-		c.RewriteToken = res.token
+		req.token = res.token
 		if c.ProgressFunc != nil {
 			c.ProgressFunc(uint64(res.written), uint64(res.size))
 		}


storage/copy.go
```diff
@@ -126,7 +126,7 @@ func (c *Copier) Run(ctx context.Context) (attrs *ObjectAttrs, err error) {
 		if err != nil {
 			return nil, err
 		}
-		c.RewriteToken = res.token
+		req.token = res.token
```
tritone, line 129: You actually need to leave the update of `c.RewriteToken` as well -- this is an exported field of `Copier` and the user might use this to resume the copy if it's interrupted.

tritone (commented):
Nice fix!

tritone (approved):

PR #6880, opened 2022-10-17
chore(storage): add content type sniffing to gRPC writer

noahdietz:
Adds content type sniffing to the gRPC Writer implementation. This uses the newly copied to gax `DetermineContentType` API, originally from google.golang.org/api/internal/gensupport.

This upgrades some dependencies to do so.

### DIFF
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -25,6 +25,7 @@ import (
 	"cloud.google.com/go/internal/trace"
 	gapic "cloud.google.com/go/storage/internal/apiv2"
 	storagepb "cloud.google.com/go/storage/internal/apiv2/stubs"
+	"github.com/googleapis/gax-go/v2"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
 	"google.golang.org/api/option/internaloption"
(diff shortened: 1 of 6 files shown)


storage/grpc_client.go
```diff
@@ -1710,7 +1711,12 @@ func (w *gRPCWriter) writeObjectSpec() (*storagepb.WriteObjectSpec, error) {
 
 // read copies the data in the reader to the given buffer and reports how much
 // data was read into the buffer and if there is no more data to read (EOF).
+// Furthermore, if the attrs.ContentType is unset, the first bytes of content
+// will be sniffed for a matching content type.
 func (w *gRPCWriter) read() (int, bool, error) {
+	if w.attrs.ContentType == "" {
```
tritone, line 1717: So this happens for each call to `read()`, which happens once per chunk (if the media is larger than ChunkSize). I dug around a bit in apiary and IIUC the content sniffing happens once for the whole media input there. I'm wondering if this makes a difference and/or if content sniffing will even work when starting from the middle of a file. Any thoughts?

Perhaps we can make a test case for this, or add a check for content type to one of our tests for multi-chunk uploads.

tritone (approved):

storage/grpc_client.go
noahdietz: Once the `w.attrs.ContentType` is set (on the first `read()`), the sniffer wont be called again. The `w.attrs.ContentType` carry through each chunk. Right?

tritone (approved):

quartzmo (approved):

PR #6882, opened 2022-10-17
test(storage): test Copier.Run's token propagation + refactor int test

BrennaEpp:


storage/copy_test.go
```diff
@@ -69,3 +70,72 @@ func TestCopyBothEncryptionKeys(t *testing.T) {
...
+	}
+}
+
+type mockRewriteObject struct {
+	storageClient
+	expectedTokens chan string
+	sendTokens     chan string
+}
+
+func (m mockRewriteObject) RewriteObject(ctx context.Context, req *rewriteObjectRequest, opts ...storageOption) (*rewriteObjectResponse, error) {
```
tritone, line 117: As discussed offline, I believe you can use maxBytesRewrittenPerCall from https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite#parameters to force a multi-request copy in an integration test, rather than relying on mocks. See https://github.com/googleapis/google-api-go-client/blob/main/storage/v1/storage-gen.go#L10961

storage/integration_test.go
```diff
@@ -1609,6 +1571,132 @@ func TestIntegration_Objects(t *testing.T) {
 	})
 }
 
+func TestIntegration_Copy(t *testing.T) {
+	multiTransportTest(skipGRPC("allowlist issue potentially related to b/246634709"), t, func(t *testing.T, ctx context.Context, bucket string, prefix string, client *Client) {
```
tritone, line 1575: Let's discuss the allowlist issue offline, looks like the bug is closed.

storage/integration_test.go
```diff
@@ -1609,6 +1571,132 @@ func TestIntegration_Objects(t *testing.T) {
...
+				desc:     "copy to new region",
+				toObj:    "copy-new-region",
+				toBucket: bucketInDifferentRegion,
+			},
+		} {
+			t.Run(test.desc, func(t *testing.T) {
+				copyObj := test.toBucket.Object(test.toObj)
+				copier := copyObj.CopierFrom(obj)
+
+				copier.maxBytesRewrittenPerCall = 1048576
```
tritone, line 1661: Can you make this a per-testcase field? I think we should probably have at least one case where the object is in fact copied in a single operation.

storage/integration_test.go
```diff
@@ -1609,6 +1571,132 @@ func TestIntegration_Objects(t *testing.T) {
...
+
+				if attrs := test.copierAttrs; attrs != nil {
+					if attrs.contentEncoding != "" {
+						copier.ContentEncoding = attrs.contentEncoding
+					}
+				}
+
```
tritone, line 1668: Let's also specify Copier.ProgressFunc and check that the number of requests made is as expected (so we can ensure we are actually testing a multi-request copy).

tritone (commented):
Nice test! a few small suggestions.

storage/integration_test.go
tritone: Ah nice, missed that!

tritone (approved):
Nice test!

PR #6885, opened 2022-10-18
feat(pubsublite): create/update export subscriptions

tmdiep:
* Added support for creating and updating export subscriptions, which write messages to a configured destination, for example a Pub/Sub topic.  More destination types may be supported in the future.
* Added support for creating subscriptions initialized to a target location within the message backlog. A seek is performed for publish and event timestamps.


pubsublite/admin.go
```diff
@@ -155,35 +156,47 @@ func (ac *AdminClient) Topics(ctx context.Context, parent string) *TopicIterator
...
-	return startingOffset{location}
+	return targetLocation{location}
+}
+
+// AtTargetLocation specifies the target location within the message backlog
+// that a new subscription should be initialized to.
+//
+// An additional seek request is initiated if the target location is a publish
+// or event time. If the seek fails, the created subscription is not deleted.
+func AtTargetLocation(target SeekTarget) CreateSubscriptionOption {
```
codyoss, line 186: Is StartingOffset still needed after this? Should it be deprecated?

pubsublite/config.go
```diff
@@ -211,6 +212,121 @@ func (tc *TopicConfigToUpdate) toUpdateRequest() *pb.UpdateTopicRequest {
...
+
+	// ExportActive specifies that export processing should be enabled.
+	ExportActive
+
+	// ExportPaused specifies that export processing should be suspended.
+	ExportPaused
+)
+
+// Status references google.golang.org/genproto/googleapis/rpc/status.
+type Status = status.Status
```
codyoss, line 251: Do you actually care about the whole status or just the code from the status? If it is just the code you could alias that, or maybe even just use an in and map it?

pubsublite/config.go
```diff
@@ -211,6 +212,121 @@ func (tc *TopicConfigToUpdate) toUpdateRequest() *pb.UpdateTopicRequest {
...
 
+// ExportDestinationConfig is the configuration for exporting to a destination.
+// Implemented by *PubSubConfig.
+type ExportDestinationConfig interface {
+	setExportConfig(ec *pb.ExportConfig) string
+}
+
+// PubSubConfig configures messages to be exported to a Pub/Sub topic.
+// Implements the ExportDestinationConfig interface.
+type PubSubConfig struct {
```
codyoss, line 223: Maybe align the naming for this to closer match the interface it is going to fullfill PubSubExportDestinationConfig?

codyoss (commented):
Just a couple comments on the surface -- don't consider any of them blocking.

pubsublite/config.go
tmdiep: The entire status proto may be useful for debugging, as the message can help differentiate between errors with the same code. The error details may also be populated with metadata.

I used the Status proto, for consistency with bigtable: https://pkg.go.dev/cloud.google.com/go/bigtable#EncryptionInfo

However, I was unsure whether it would be better to convert to https://pkg.go.dev/google.golang.org/grpc/status#Status.

pubsublite/config.go
```diff
@@ -211,6 +212,121 @@ func (tc *TopicConfigToUpdate) toUpdateRequest() *pb.UpdateTopicRequest {
 	}
 }
 
+// ExportDestinationConfig is the configuration for exporting to a destination.
+// Implemented by *PubSubConfig.
+type ExportDestinationConfig interface {
```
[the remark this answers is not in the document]
tmdiep: We can't predict how many, but there may be several in the future, hence the oneof-like representation.

pubsublite/admin.go
tmdiep: I added a comment stating that StartingOffset will be removed in the next major revision.

pubsublite/config.go
tmdiep: Renamed to `PubSubDestinationConfig` to compromise on length and clarity :)

pubsublite/config.go
tmdiep: Obsolete. We changed the API to return enum values.

hongalex (approved):

PR #6886, opened 2022-10-18
feat(spanner): add support for Proto Columns

harshachinta:
This PR contains implementation changes to add support for Proto Messages and Proto Enum.


spanner/value.go
```diff
@@ -1844,6 +1850,47 @@ func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...decodeO
...
+		if err != nil {
+			return err
+		}
+		y, err := strconv.ParseInt(x, 10, 64)
+		if err != nil {
+			return errBadEncoding(v, err)
+		}
+		if reflect.ValueOf(p).Kind() != reflect.Ptr {
+			return errNotAPointer(p)
+		}
```
olavloite, line 1873: I think it would make sense to move this up a little to just after the `if p == nil` check. Then we do all the `is the destination variable correct` checks before other checks.

spanner/value.go
```diff
@@ -3564,6 +3611,21 @@ func encodeValue(v interface{}) (*proto3.Value, *sppb.Type, error) {
 		pt = proto.Clone(v.Type).(*sppb.Type)
 	case []GenericColumnValue:
 		return nil, nil, errEncoderUnsupportedType(v)
+	case protoreflect.Enum:
+		// TODO: here nil check not needed because a enum variable always has a default value. Recheck
```
olavloite, line 3615: Is this TODO being looked into?

olavloite (approved):
Looks generally good to me, with a couple of minor nits.

spanner/value.go
harshachinta: Thanks for the suggestion Knut. This check should ideally be above.
I have added the change in the recent commit.

spanner/value.go
harshachinta: I have removed the TODO comment. As we are fetching `.Number()`, it would be good to have the nil check.

spanner/go.mod
```diff
@@ -31,3 +31,6 @@ require (
 	golang.org/x/text v0.3.7 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
 )
+
+// To prevent failing builds as proto changes are not available in go-genproto until Public GA
+replace google.golang.org/genproto v0.0.0-20221014173430-6e2ab493f96b => github.com/harshachinta/go-genproto v0.0.0-20221020104338-f731337b715d
```
gauravpurohit06, line 36: - Please check if we can create branch on the [googleapis/google-cloud-go](https://github.com/googleapis/google-cloud-go) and do the Proto schema changes. 
- If we are able to create, use branch from [googleapis/google-cloud-go](https://github.com/googleapis/google-cloud-go) here.

spanner/testdata/book.pb.go
```diff
@@ -0,0 +1,229 @@
...
+		if ms.LoadMessageInfo() == nil {
+			ms.StoreMessageInfo(mi)
+		}
+		return ms
+	}
+	return mi.MessageOf(x)
+}
+
+// Deprecated: Use Book.ProtoReflect.Descriptor instead.
+func (*Book) Descriptor() ([]byte, []int) {
```
gauravpurohit06, line 114: Please replace book example with the standard Singers/Albums one.

gauravpurohit06 (changes requested):

spanner/go.mod
harshachinta: In addition to `go.mod` file, the `go-genproto` dependency needs to be updated in `internal/kokoro/build_samples.sh` file as well for the kokoro CI build to pass successfully.  This is because the `build_samples.sh` file builds the `golang-samples` file which has `go-genproto` pointing to the release version, and thus failing the build.

But the `build_samples.sh` file change is not needed for our proto column feature and hence reverted those changes.

spanner/testdata/book.pb.go
harshachinta: Replaced `Book` example with standard `Singers` proto.

rahul2393 (approved):

spanner/testdata/protos/singer.proto
```diff
@@ -0,0 +1,34 @@
...
+limitations under the License.
+*/
+
+syntax = "proto2";
+
+package protos;
+option go_package = "./protos";
+
+message SingerInfo {
+  optional int64 singerId = 1;
```
gauravpurohit06, line 23: Change it to `snake_case`.

spanner/testdata/protos/singer.proto
```diff
@@ -0,0 +1,34 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+syntax = "proto2";
+
+package protos;
```
gauravpurohit06, line 19: Package Name will be used later for parameterised queries,  let's better keep it consistent across the clients.  We can use something like `spanner.examples.music`.

gauravpurohit06 (approved):
LGTM with minor comments

spanner/testdata/protos/singer.proto
harshachinta: Done

spanner/testdata/protos/singer.proto
harshachinta: Updated package name

spanner/protoutils.go
```diff
@@ -128,3 +130,20 @@ func structType(fields ...*sppb.StructType_Field) *sppb.Type {
...
+
+func protoType() *sppb.Type {
+	return &sppb.Type{Code: sppb.TypeCode_PROTO}
+}
+
+func enumType() *sppb.Type {
+	return &sppb.Type{Code: sppb.TypeCode_ENUM}
+}
+
+func messageProto(m proto.Message) *proto3.Value {
```
charvisingla, line 142: We are going to support only proto2 for Verily Private GA. Not sure if there is any implication to use proto3.Value here instead of proto2.Value. Can you verify?

charvisingla (approved):
LGTM

spanner/value.go
```diff
@@ -1844,6 +1850,47 @@ func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...decodeO
...
+		}
+		x, err := getStringValue(v)
+		if err != nil {
+			return err
+		}
+		y, err := strconv.ParseInt(x, 10, 64)
+		if err != nil {
+			return errBadEncoding(v, err)
+		}
+		reflect.ValueOf(p).Elem().SetInt(y)
```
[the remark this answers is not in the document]
harshachinta: At client library end, as we do not have the schema of the enum it is difficult to validate the valid range of enum value.
I guess validation should be done from the backend side while inserting enum. If the enum is validated during insertion, then we can assume the value `y` is always valid.
However, if `y` is not a valid enum value, then there will not be any exception in above code. The integer value gets assigned to the enum and when user tries to print the enum it shows an integer.
For ex: Consider the following schema of Enum:
```
enum Genre {
  POP = 0;
  JAZZ = 1;
  FOLK = 2;
  ROCK = 3;
}
```
Here valid range is 0 to 3. 
If `y` is `2`, then `fmt.Println(p)` prints `FOLK`.
If `y` is `6` which is `invalid`, then `fmt.Println(p)` prints `6`

spanner/value.go
```diff
@@ -1844,6 +1850,47 @@ func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...decodeO
...
+			return errNilDst(p)
+		}
+		if reflect.ValueOf(p).Kind() != reflect.Ptr {
+			return errNotAPointer(p)
+		}
+		if code != sppb.TypeCode_ENUM {
+			return errTypeMismatch(code, acode, ptr)
+		}
+		if isNull {
+			return errDstNotForNull(ptr)
```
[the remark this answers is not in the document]
harshachinta: We will add a new type `NullEnum` to read null enum values. These features will be added in subsequent PR.

spanner/protoutils.go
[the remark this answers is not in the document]
harshachinta: The prefix cannot be Value as we have followed a common convention in the file.
However, we can rename it to `protoMessageProto` and `protoEnumProto`.

spanner/protoutils.go
```diff
@@ -128,3 +130,20 @@ func structType(fields ...*sppb.StructType_Field) *sppb.Type {
 func nullProto() *proto3.Value {
 	return &proto3.Value{Kind: &proto3.Value_NullValue{NullValue: proto3.NullValue_NULL_VALUE}}
 }
+
+func protoType() *sppb.Type {
+	return &sppb.Type{Code: sppb.TypeCode_PROTO}
```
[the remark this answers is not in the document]
harshachinta: I missed to add this in the PR. Thanks for pointing it out @Shua1 
I have committed the changes for adding `proto_type_fqn`

spanner/protoutils.go
```diff
@@ -128,3 +130,20 @@ func structType(fields ...*sppb.StructType_Field) *sppb.Type {
...
+}
+
+func enumType() *sppb.Type {
+	return &sppb.Type{Code: sppb.TypeCode_ENUM}
```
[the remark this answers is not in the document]
harshachinta: Added the changes for `proto_type_fqn`

spanner/protoutils.go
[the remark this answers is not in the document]
harshachinta: Added changes

spanner/value.go
```diff
@@ -3564,6 +3614,24 @@ func encodeValue(v interface{}) (*proto3.Value, *sppb.Type, error) {
 		pt = proto.Clone(v.Type).(*sppb.Type)
 	case []GenericColumnValue:
 		return nil, nil, errEncoderUnsupportedType(v)
+	case protoreflect.Enum:
+		var protoEnumfqn string
+		if v != nil {
+			pb.Kind = stringKind(strconv.FormatInt(int64(v.Number()), 10))
+			protoEnumfqn = string(v.Descriptor().FullName())
+		}
+		pt = enumType(protoEnumfqn)
```
Shua1, line 3623: If v is nil, it seems protoEnumfqn is not set. Is this expected?

Do we have a test for this? If so, which line. Thanks!

spanner/value.go
```diff
@@ -3564,6 +3614,24 @@ func encodeValue(v interface{}) (*proto3.Value, *sppb.Type, error) {
...
+		var protoMessagefqn string
+		if v != nil && proto.MessageReflect(v).IsValid() {
+			bytes, err := proto.Marshal(v)
+			if err != nil {
+				return nil, nil, err
+			}
+			pb.Kind = stringKind(base64.StdEncoding.EncodeToString(bytes))
+			protoMessagefqn = string(proto.MessageReflect(v).Descriptor().FullName())
+		}
+		pt = protoType(protoMessagefqn)
```
Shua1, line 3634: Same as above.

spanner/value.go
```diff
@@ -1844,6 +1850,50 @@ func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...decodeO
...
 	case *GenericColumnValue:
 		*p = GenericColumnValue{Type: t, Value: v}
+	case protoreflect.Enum:
+		if p == nil {
+			return errNilDst(p)
+		}
+		if reflect.ValueOf(p).Kind() != reflect.Ptr {
+			return errNotAPointer(p)
+		}
+		if code != sppb.TypeCode_ENUM {
```
Shua1, line 1860: This is fine for now. But we need to keep in mind that some user might want to decode a INT64 value into enum.

spanner/value.go
```diff
@@ -1844,6 +1850,50 @@ func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...decodeO
...
+		}
+		reflect.ValueOf(p).Elem().SetInt(y)
+	case proto.Message:
+		if p == nil {
+			return errNilDst(p)
+		}
+		if reflect.ValueOf(p).Kind() != reflect.Ptr {
+			return errNotAPointer(p)
+		}
+		if code != sppb.TypeCode_PROTO {
```
Shua1, line 1882: Different from the ENUM case, we do need to support decoding bytes typed result into PROTO. BYTES values and PROTO values are compatible with each other.

Please also add tests for both direction.

Shua1 (commented):
Thanks!

spanner/value.go
harshachinta: If `v` is nil, then `protoMessagefqn` by default gets assigned with default value of string which is `""`.

spanner/value.go
harshachinta: @Shua1 
Supporting compatibility between BYTES and PROTO values involves code changes in the decoding function of BYTES also. This needs a bit more testing and handling corner cases.
I will make a note of this and will support this case in my subsequent PR. Does that sound good?

spanner/value.go
Shua1: Maybe, we should make it happen for now. To be consistent with
1. the write path, where integer can be written into ENUM column.
2. the proto and bytes compatible behavior.

What do you think?

harshachinta: @Shua1 
I will add the above compatibilities as a separate PR. This will make the code change clear of why the changes were added. Otherwise it might create confusion of why Bytes were added into Proto code and vice versa.

harshachinta: @Shua1 
I have added the code changes and integration tests to validate the following compatibility behaviours between **Int64 and Enum**.

1. Write Enum into Int64 column
2. Write Int64 into Enum column
3. Read Int64 column value using Enum
4. Read Enum column value using Int64

spanner/value.go
harshachinta: @Shua1 
I have added the code changes and integration tests to validate the following compatibility behaviours between Bytes and Proto.

1. Write Bytes into Proto column
2. Write Proto into Bytes column
3. Read Proto column value using Bytes
4. Read Bytes column value using Proto

spanner/integration_test.go
```diff
@@ -2013,6 +2019,119 @@ func TestIntegration_BasicTypes(t *testing.T) {
...
+	}
+	bytesSingerProtoMessage, _ := proto.Marshal(&singerProtoMessage)
+
+	tests := []struct {
+		col  string
+		val  interface{}
+		want interface{}
+	}{
+		{col: "ProtoMessage", val: &singerProtoMessage, want: singerProtoMessage},
+		{col: "ProtoEnum", val: pb.Genre_ROCK, want: singerProtoEnum},
```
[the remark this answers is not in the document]
harshachinta: This is to differentiate the data passed in `val` and `want`. `pb.Genre_ROCK` is passed in `val`, and `singerProtoEnum` is passed in `want` in the tests. Also in some places we need to pass a pointer instead of just the constant. That is why we need both `pb.Genre_ROCK` and `singerProtoEnum`

harshachinta:
https://github.com/googleapis/google-cloud-go/pull/6886#discussion_r1005038127
Is it expected? We need a test for this.
Can we get the descriptor even if the value is nil as the descriptor just depends on the type?

**Answer**: The only way we can pass nil here is to directly pass `nil` while inserting data. We cannot create a nil instance of an `enum` (`pb.Genre_Rock` or `pb.Genre`) or `proto` (`pb.Book{}`) as they create default objects in every possibility of declaration. We cannot get descriptor fn for `nil`. So assigning `""` in that case.

spanner/value.go
harshachinta: The only way we can pass nil here is to directly pass `nil` while inserting data. We cannot create a nil instance of an `enum` (`pb.Genre_Rock` or `pb.Genre`) or `proto` (`pb.Book{}`) as they create default objects in every possibility of declaration. We cannot get descriptor fn for `nil`. So assigning `""` in that case.

harshachinta: Also this line will not be called as there is no way to send a nil instance of an enum or proto. I have followed the coding convention of client library and also added unit tests for the same.

spanner/value.go
harshachinta: Yes @Shua1. This behaviour is valid on our side also.

spanner/protoutils.go
```diff
@@ -128,3 +130,20 @@ func structType(fields ...*sppb.StructType_Field) *sppb.Type {
 func nullProto() *proto3.Value {
 	return &proto3.Value{Kind: &proto3.Value_NullValue{NullValue: proto3.NullValue_NULL_VALUE}}
 }
+
+func protoType(fqn string) *sppb.Type {
```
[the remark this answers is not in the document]
harshachinta: Modified the names @Shua1

spanner/protoutils.go
```diff
@@ -128,3 +130,20 @@ func structType(fields ...*sppb.StructType_Field) *sppb.Type {
...
+	return &sppb.Type{Code: sppb.TypeCode_PROTO, ProtoTypeFqn: fqn}
+}
+
+func enumType(fqn string) *sppb.Type {
```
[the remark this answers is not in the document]
harshachinta: Modified

spanner/value.go
harshachinta: The only way to pass a nil type of proto enum is by type casting and passing like `(*pb.Genre)(nil)`. As it is a nil type, calling `v.Descriptor()` raises a panic in golang with exception as follows:

`panic: value method Genre.Descriptor called using nil *Genre pointer`

So, the only option is to set `protoEnumfqn` as `""`.
I have added a unit test for the same [Test](https://github.com/googleapis/google-cloud-go/pull/6886/files#diff-d66cf51d595f7a3cfd7349436181f0575be539d660acefae20b4d22fcfb96a6fR480)

spanner/value.go
harshachinta: Added unit test for this also. 
As per protoreflect documentation, `(*pb.SingerInfo)(nil)` is an invalid message type. So setting `protoMessagefqn` as `""` is valid.

Text from Protoreflect document:
`An invalid message often corresponds to a nil pointer of the concrete message type`

Shua1 (commented):
Almost there.

spanner/value.go
harshachinta: The following code will give the same type `(*pb.Genre)(nil)` as pointed above
```
v := pb.Genre_ROCK
vp := &v
vp = nil
```
golang throws an exception when any method is called using nil pointer. The only reasonable option would be to throw an error to user indicating that nil types cannot be used.
I have added the code changes @Shua1. Can you please have a look?

Shua1 (approved):
LGTM

PR #6893, opened 2022-10-18
feat(dataform): start generating apiv1beta1

quartzmo:


internal/gapicgen/generator/config.go
```diff
@@ -1773,6 +1773,16 @@ var MicrogenGapicConfigs = []*MicrogenConfig{
...
 		ReleaseLevel: "beta",
 	},
+	{
+		InputDirectoryPath:    "google/cloud/dataform/v1beta1",
+		Pkg:                   "dataform",
+		ImportPath:            "cloud.google.com/go/dataform/apiv1beta1",
+		GRPCServiceConfigPath: "dataform_grpc_service_config.json",
+		ApiServiceConfigPath:  "dataform_v1beta1.yaml",
+		Transports:            []string{"grpc", "rest"},
+		// GA after 2022/11/18
```
noahdietz, line 1783: Nit: I don't think we need to ever GA a Beta API package :)

noahdietz (approved):

internal/gapicgen/generator/config.go
quartzmo: Good catch, thanks!

Issue #6135, opened 2022-06-07
chore(ci): move linting presubmit from kokoro into GitHub Actions


noahdietz:
We should move the Go linting tasks from kokoro into a GitHub Action for easier debugging, speed of execution (and thus quicker feedback), and ease of maintenance.

PR #6901, opened 2022-10-19
chore(ci): move apidiff to GitHub Action

noahdietz:
This moves the `apidiff` checks to GitHub Actions, based on the success we've had with the same thing in go-genproto.

It also removes the `apidiff` check from the kokoro presubmit.

Updates #6135.


.github/workflows/apidiff.yml
```diff
@@ -0,0 +1,83 @@
...
+      if: ${{ failure() && !github.event.pull_request.head.repo.fork }}
+      uses: actions/github-script@v6
+      with:
+        script: |
+          github.rest.issues.addLabels({
+            issue_number: ${{ github.event.number }},
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            labels: ['breaking change']
+          })
```
codyoss, line 83: nit: newlinee

codyoss (approved):
Thanks!

.github/workflows/apidiff.yml
noahdietz: Done!

PR #6911, opened 2022-10-24
chore(ci): move vet to GitHub Action

### LINKED ISSUE #6135: chore(ci): move linting presubmit from kokoro into GitHub Actions

noahdietz:
Moves the existing `vet.sh` script and execution (presubmit and continuous) to GitHub Actions (execute on Pull Requests and Pushes on `main`).

Note: The GitHub Action dictates which version of Go the script is executed with. Currently it is our "latest" which is Go `1.19`. I removed the Go version guard in `vet.sh` since it will only ever be executed on the version we dictate, rather than both earliest a& latest like with Kokoro.

I think that we should consider executing `vet` only on changed directories instead of the entire repository, similar to how `apidiff` works. That is an improvement we can make later though. This is just a start.

Fixes #6135


.github/workflows/vet.yml
```diff
@@ -0,0 +1,23 @@
...
+    - uses: actions/setup-go@v3
+      with:
+        go-version: '1.19'
+    - name: Install tools
+      run: |
+        go install golang.org/x/lint/golint@latest && \
+        go install golang.org/x/tools/cmd/goimports@latest && \
+        go install honnef.co/go/tools/cmd/staticcheck@latest
+    - name: Execute vet.sh
+      run: ./.github/workflows/vet.sh
```
codyoss, line 23: Should we just call the existing script instead and not make a copy or move it?

noahdietz: I meant to delete the `internal/kokoro/vet.sh` file, moving it to `.github/workflows`, because it was going to be executed there instead of in Kokoro. I think I accidentally added it back in!

Where would like the file to exist, I don't really care.

noahdietz: I've fixed it, so removed `internal/kokoro/vet.sh` in favor of `.github/workflows/vet.sh`.

codyoss (approved):
LGTM

PR #6912, opened 2022-10-24
chore(ci): fix apidiff handling of empty set

noahdietz:
`apidiff` was silently failing when there were no directories worth diffing. This didn't show up on PR status checks, but shows up in the Action history.


codyoss (approved):
LGTM

PR #6913, opened 2022-10-24
chore(bigtable): update bttest for table deletion

### LINKED ISSUE #6905: bigtable: add deletion protection support in the Bigtable emulator
Deletion protection is a new feature we need to support. Currently the emulator doesn't support it, hence pull/6813 can only pass in when read production environment is used, but not then the emulator is used.

A clear and concise description of what you want to happen. What should the API be? How would you use it?

Add deletion protection support (update table) in the emulator so that the integration test can pass.

**Additional context**

telpirion:
This PR should help to unblock #6813

Fixes #6905


bigtable/bttest/inmem.go
```diff
@@ -197,10 +201,60 @@ func (s *server) DeleteTable(ctx context.Context, req *btapb.DeleteTableRequest)
...
+		return nil, status.Errorf(codes.FailedPrecondition, "table %q is protected from deletion", req.Name)
+	}
 	delete(s.tables, req.Name)
 	return &emptypb.Empty{}, nil
 }
 
+func (s *server) UpdateTable(ctx context.Context, req *btapb.UpdateTableRequest) (*longrunning.Operation, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if _, ok := s.tables[req.GetTable().GetName()]; !ok {
```
enocom, line 214:
``` golang
tbl, ok := s.tables[req.GetTable().GetName()]
if !ok { /* ... */
...
```

enocom (approved):
One small Go nit.

kolea2 (approved):

bigtable/bttest/inmem.go
```diff
@@ -225,6 +278,9 @@ func (s *server) ModifyColumnFamilies(ctx context.Context, req *btapb.ModifyColu
 			tbl.counter++
 			tbl.families[mod.Id] = newcf
 		} else if mod.GetDrop() {
+			if tbl.isProtected {
+				return nil, status.Errorf(codes.FailedPrecondition, "table %q is protected from deletion", req.Name)
```
kevinsi4508, line 282: This is not consistent with ours server implementation I believe.

It should be like this:

if tbl.isProtected && request has at least one drop cf
    return FailedPrecondition

// Then start updating column families
for _, mod := range req.Modifications {

bigtable/bttest/inmem.go
```diff
@@ -197,10 +201,59 @@ func (s *server) DeleteTable(ctx context.Context, req *btapb.DeleteTableRequest)
...
 }
 
+func (s *server) UpdateTable(ctx context.Context, req *btapb.UpdateTableRequest) (*longrunning.Operation, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	tbl, ok := s.tables[req.GetTable().GetName()]
+	if !ok {
+		return nil, status.Errorf(codes.NotFound, "table %q not found", req.GetTable().GetName())
+	}
+	tbl.isProtected = req.GetTable().GetDeletionProtection()
```
kevinsi4508, line 218: We use [update_mask](https://source.corp.google.com/piper///depot/google3/google/bigtable/admin/v2/bigtable_table_admin.proto;bpv=1;bpt=1;l=801?gsn=update_mask&gs=kythe%3A%2F%2Fgoogle3%3Flang%3Dprotobuf%3Fpath%3Dgoogle%2Fbigtable%2Fadmin%2Fv2%2Fbigtable_table_admin.proto%234.9.2.1)

bigtable/bttest/inmem.go
telpirion: Ack. My reading of the protos implies that the two statements are (functionally) equivalent.

The `else if mod.GetDrop()` will evaluate as true if the request has at least one drop. The `if tlb.isProtected`, branch of the next statement thus returns an error if BOTH the conditions are true.

PR #6919, opened 2022-10-25
perf(storage): pre-allocate slices

estensen:
When we know the final size of the slice and map we can pre-allocate memory to avoid re-allocations.

It's a tiny bit faster to set the `len` and use the index to access the position, but for maps you need to add a counter so the decreased readability is not worth it IMO.


storage/acl.go
```diff
@@ -172,33 +172,33 @@ func (a *ACLHandle) configureCall(ctx context.Context, call interface{ Header()
 }
 
 func toObjectACLRules(items []*raw.ObjectAccessControl) []ACLRule {
-	var rs []ACLRule
-	for _, item := range items {
-		rs = append(rs, toObjectACLRule(item))
+	rs := make([]ACLRule, len(items))
```
tritone, line 175: This is not really equivalent; if `len(items) == 0`, we get an empty slice rather than a nil. Same elsewhere. This can make a difference in what is sent to the API.

tritone (commented):
Thanks for your contribution.

I'm not totally sold that this will be a large improvement given the many other sources of latency and delay for API calls. However, I'm open to merging it if you would fix the issue I note around empty vs. nil slices.

codyoss:
Closing due to lack of response.

PR #6921, opened 2022-10-25
chore(ci): ignore go.mod and go.sum changes in service dirs

noahdietz:
Discovered in #6918 that changes in "service root" directories, like updating `go.mod` and `go.sum`, where there are no Go files might cause false positives in apidiff. Those changes shouldn't produce an apidiff run.

### DIFF
--- a/.github/workflows/apidiff.yml
+++ b/.github/workflows/apidiff.yml
@@ -18,7 +18,7 @@ jobs:
       id: changed_dirs
       # Ignore changes to the internal and root directories.
       run: |
-        dirs=$(git diff-tree --no-commit-id --name-only --diff-filter=DMR -r ${{ steps.main.outputs.hash }}..HEAD | xargs -r -L1 dirname | uniq | grep -v -e 'internal' -e '\.' | cat)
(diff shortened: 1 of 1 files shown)


.github/workflows/apidiff.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       id: changed_dirs
       # Ignore changes to the internal and root directories.
       run: |
-        dirs=$(git diff-tree --no-commit-id --name-only --diff-filter=DMR -r ${{ steps.main.outputs.hash }}..HEAD | xargs -r -L1 dirname | uniq | grep -v -e 'internal' -e '\.' | cat)
+        dirs=$(git diff-tree --no-commit-id --name-only --diff-filter=DMR -r ${{ steps.main.outputs.hash }}..HEAD | grep -v -e 'go\.[mod|sum]' | xargs -r -L1 dirname | uniq | grep -v -e 'internal' -e '\.' | cat)
```
codyoss, line 21: Will this still work for modules like pubsub or storage that have sources at the module root?

codyoss (approved):

quartzmo (approved):

Issue #6311, opened 2022-07-07
compute/metadata: make its own module


codyoss:
Now that we generate a compute library we should separate out the metadata package into its own module. This will potentially have a nice benefit of lightening our dependency graph once things rely on it. 

Also in the future it would be nice to have a new version of this package/module that is context aware, see #4483. In order to do that though I think we first need to have a stable module to give people time to update should they choose to do so.

liggitt:
excited to get a tagged version of the dependency-free compute/metadata module so we can help consumers thin out their dependency graphs

PR #6940, opened 2022-10-26
chore: carve out compute/metadata module

codyoss:
This commit will be manually tagged
- compute/v1.12.0
- compute/metadata/v0.1.0

Updates: #6311

### DIFF
--- a/compute/metadata/go.mod
+++ b/compute/metadata/go.mod
@@ -1,0 +1,5 @@
+module cloud.google.com/go/compute/metadata
+
+go 1.19
+
+require cloud.google.com/go/compute v1.12.0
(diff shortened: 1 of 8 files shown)


codyoss:
This will require the typical fixup of sums post merge and another release following those.

codyoss:
The failing vet check is expected as it can't compute sums for something that does not exist yet.

noahdietz (approved):

compute/metadata/go.mod
```diff
@@ -0,0 +1,5 @@
+module cloud.google.com/go/compute/metadata
+
+go 1.19
+
+require cloud.google.com/go/compute v1.12.0
```
liggitt, line 5: is the intent to drop this in a subsequent version of compute/metadata?

codyoss: Yes, this is just required to make the transition some and for the modules to resolve. This process of carving out an existing module is documented on the go wiki if your curious: https://github.com/golang/go/wiki/Modules#is-it-possible-to-add-a-module-to-a-multi-module-repository

PR #6945, opened 2022-10-26
fix(datastore): attempt rollback after panic in transaction

### LINKED ISSUE #5969: datastore: rollback when panic occurs in RunInTransaction.
**Is your functional requirement relevant to the issue? Please describe.
Please describe.

We would like to be `rollback` in the unlikely event of a `panic` in a `RunInTransaction`.

**Describe the solution you want**.

We want rollback using `defer` and `recover` for `f(tx)`.

https://github.com/googleapis/google-cloud-go/blob/dd8973b7f72e4b153f7f135e4af8e89c66d3666d/datastore/transaction.go#L173

**Describe alternatives you have considered**.

none

**Additional Context**.

none

telpirion:
Fixes #5969


datastore/transaction.go
```diff
@@ -196,6 +196,17 @@ func (c *Client) RunInTransaction(ctx context.Context, f func(tx *Transaction) e
 	settings := newTransactionSettings(opts)
 	for n := 0; n < settings.attempts; n++ {
 		tx, err := c.newTransaction(ctx, settings)
+
+		// Ensure that any panic has a corresponding rollback
+		defer func(t *Transaction) {
```
codyoss, line 201: This seems likely like fixing a symptom and not the problem. When does this code panic? We should make it so it does not and return an error if needed.

telpirion: This library has a lot of places where it has a `panic()`. Cleaning up all of the panics will be a work item in itself.

telpirion:
Closing for now

PR #6954, opened 2022-10-27
feat(spanner): add support for handling null value in Proto columns

harshachinta:
This PR contains implementation changes to add support for handling null values in Proto Messages and Proto Enum type columns. Also the PR has changes to get `protoTypeFqn` for nil types.

### DIFF
--- a/spanner/integration_test.go
+++ b/spanner/integration_test.go
@@ -2048,14 +2048,12 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
 		{col: "ProtoMessage", val: &singerProtoMessage, want: NullProto{&singerProtoMessage, true}},
 		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, true}, want: singerProtoMessage},
 		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, true}, want: NullProto{&singerProtoMessage, true}},
-		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, false}, want: NullProto{}},
 		{col: "ProtoMessage", val: nil, want: NullProto{}},
 		// Proto Enum
 		{col: "ProtoEnum", val: pb.Genre_ROCK, want: singerProtoEnum},
 		{col: "ProtoEnum", val: pb.Genre_ROCK, want: NullEnum{&singerProtoEnum, true}},
 		{col: "ProtoEnum", val: NullEnum{pb.Genre_ROCK, true}, want: singerProtoEnum},
 		{col: "ProtoEnum", val: NullEnum{pb.Genre_ROCK, true}, want: NullEnum{&singerProtoEnum, true}},
-		{col: "ProtoEnum", val: NullEnum{pb.Genre_ROCK, false}, want: NullEnum{}},
 		{col: "ProtoEnum", val: nil, want: NullEnum{}},
 		// Test Compatibility between Int64 and ProtoEnum
 		{col: "Int64a", val: pb.Genre_ROCK, want: int64(3)},
@@ -2080,23 +2078,19 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
 		{col: "Int64a", val: NullEnum{pb.Genre_ROCK, true}, want: NullEnum{&singerProtoEnum, true}},
 		{col: "Int64a", val: NullInt64{3, true}, want: NullEnum{&singerProtoEnum, true}},
 		{col: "Int64a", val: NullInt64{3, false}, want: NullEnum{}},
-		{col: "Int64a", val: NullEnum{}, want: NullInt64{}},
 		{col: "ProtoEnum", val: NullInt64{3, true}, want: singerProtoEnum},
 		{col: "ProtoEnum", val: NullInt64{3, true}, want: NullEnum{&singerProtoEnum, true}},
 		{col: "ProtoEnum", val: NullInt64{3, true}, want: NullInt64{3, true}},
 		{col: "ProtoEnum", val: NullEnum{pb.Genre_ROCK, true}, want: NullInt64{3, true}},
-		{col: "ProtoEnum", val: NullEnum{pb.Genre_ROCK, false}, want: NullInt64{}},
 		// Test Compatibility between Bytes and NullProto
 		{col: "Bytes", val: NullProto{&singerProtoMessage, true}, want: bytesSingerProtoMessage},
 		{col: "Bytes", val: NullProto{&singerProtoMessage, true}, want: NullProto{&singerProtoMessage, true}},
 		{col: "Bytes", val: bytesSingerProtoMessage, want: NullProto{&singerProtoMessage, true}},
 		{col: "Bytes", val: bytesSingerProtoMessage},
-		{col: "Bytes", val: NullProto{}, want: []byte(nil)},
 		{col: "ProtoMessage", val: bytesSingerProtoMessage, want: NullProto{&singerProtoMessage, true}},
 		{col: "ProtoMessage", val: []byte(nil), want: NullProto{}},
 		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, true}, want: bytesSingerProtoMessage},
 		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, true}, want: NullProto{&singerProtoMessage, true}},
-		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, false}, want: []byte(nil)},
 	}
 
 	// Write rows into table first using DML.
--- a/spanner/integration_test.go
+++ b/spanner/integration_test.go
@@ -2043,8 +2043,20 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
 		val  interface{}
 		want interface{}
 	}{
+		// Proto Message
 		{col: "ProtoMessage", val: &singerProtoMessage, want: singerProtoMessage},
+		{col: "ProtoMessage", val: &singerProtoMessage, want: NullProto{&singerProtoMessage, true}},
+		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, true}, want: singerProtoMessage},
+		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, true}, want: NullProto{&singerProtoMessage, true}},
+		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, false}, want: NullProto{}},
+		{col: "ProtoMessage", val: nil, want: NullProto{}},
+		// Proto Enum
 		{col: "ProtoEnum", val: pb.Genre_ROCK, want: singerProtoEnum},
+		{col: "ProtoEnum", val: pb.Genre_ROCK, want: NullEnum{&singerProtoEnum, true}},
+		{col: "ProtoEnum", val: NullEnum{pb.Genre_ROCK, true}, want: singerProtoEnum},
+		{col: "ProtoEnum", val: NullEnum{pb.Genre_ROCK, true}, want: NullEnum{&singerProtoEnum, true}},
+		{col: "ProtoEnum", val: NullEnum{pb.Genre_ROCK, false}, want: NullEnum{}},
+		{col: "ProtoEnum", val: nil, want: NullEnum{}},
 		// Test Compatibility between Int64 and ProtoEnum
 		{col: "Int64a", val: pb.Genre_ROCK, want: int64(3)},
 		{col: "Int64a", val: pb.Genre_ROCK, want: singerProtoEnum},
@@ -2063,6 +2075,28 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
 		{col: "ProtoMessage", val: bytesSingerProtoMessage, want: bytesSingerProtoMessage},
 		{col: "ProtoMessage", val: &singerProtoMessage, want: bytesSingerProtoMessage},
 		{col: "ProtoMessage", val: &singerProtoMessage, want: singerProtoMessage},
+		// Test Compatibility between NullInt64 and NullEnum
+		{col: "Int64a", val: NullEnum{pb.Genre_ROCK, true}, want: NullInt64{3, true}},
+		{col: "Int64a", val: NullEnum{pb.Genre_ROCK, true}, want: NullEnum{&singerProtoEnum, true}},
+		{col: "Int64a", val: NullInt64{3, true}, want: NullEnum{&singerProtoEnum, true}},
+		{col: "Int64a", val: NullInt64{3, false}, want: NullEnum{}},
+		{col: "Int64a", val: NullEnum{}, want: NullInt64{}},
+		{col: "ProtoEnum", val: NullInt64{3, true}, want: singerProtoEnum},
+		{col: "ProtoEnum", val: NullInt64{3, true}, want: NullEnum{&singerProtoEnum, true}},
+		{col: "ProtoEnum", val: NullInt64{3, true}, want: NullInt64{3, true}},
+		{col: "ProtoEnum", val: NullEnum{pb.Genre_ROCK, true}, want: NullInt64{3, true}},
+		{col: "ProtoEnum", val: NullEnum{pb.Genre_ROCK, false}, want: NullInt64{}},
+		// Test Compatibility between Bytes and NullProto
+		{col: "Bytes", val: NullProto{&singerProtoMessage, true}, want: bytesSingerProtoMessage},
+		{col: "Bytes", val: NullProto{&singerProtoMessage, true}, want: NullProto{&singerProtoMessage, true}},
+		{col: "Bytes", val: bytesSingerProtoMessage, want: NullProto{&singerProtoMessage, true}},
+		{col: "Bytes", val: bytesSingerProtoMessage},
+		{col: "Bytes", val: NullProto{}, want: []byte(nil)},
+		{col: "ProtoMessage", val: bytesSingerProtoMessage, want: NullProto{&singerProtoMessage, true}},
+		{col: "ProtoMessage", val: []byte(nil), want: NullProto{}},
+		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, true}, want: bytesSingerProtoMessage},
+		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, true}, want: NullProto{&singerProtoMessage, true}},
+		{col: "ProtoMessage", val: NullProto{&singerProtoMessage, false}, want: []byte(nil)},
 	}
 
 	// Write rows into table first using DML.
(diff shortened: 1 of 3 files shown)


spanner/value_test.go
```diff
@@ -476,6 +476,10 @@ func TestEncodeValue(t *testing.T) {
 		// PROTO MESSAGE AND PROTO ENUM
 		{singer1ProtoMsg, protoMessageProto(singer1ProtoMsg), tProtoMessage, "Proto Message"},
 		{singer1ProtoEnum, protoEnumProto(singer1ProtoEnum), tProtoEnum, "Proto Enum"},
+		{NullProto{singer1ProtoMsg, true}, protoMessageProto(singer1ProtoMsg), tProtoMessage, "NullProto with value"},
+		{NullProto{}, nullProto(), protoMessageType(""), "NullProto with null"},
```
Shua1, line 480: For protoMessageType(""), we are using empty string. Is it true that the string content doesn't matter? If so, what about put something like: does.not.matter? :D

spanner/value.go
```diff
@@ -877,6 +877,96 @@ func (n *PGNumeric) UnmarshalJSON(payload []byte) error {
...
+// String implements Stringer.String for NullProto.
+func (n NullProto) String() string {
+	if !n.Valid {
+		return nullString
+	}
+	return n.ProtoVal.String()
+}
+
+// MarshalJSON implements json.Marshaler.MarshalJSON for NullProto.
+func (n NullProto) MarshalJSON() ([]byte, error) {
```
Shua1, line 900: Could you please help me understand when and how MarshalJSON is used?

spanner/value.go
```diff
@@ -3631,6 +3775,11 @@ func encodeValue(v interface{}) (*proto3.Value, *sppb.Type, error) {
 			}
 		}
 		pt = protoEnumType(protoEnumfqn)
+	case NullEnum:
+		if v.Valid {
+			return encodeValue(v.EnumVal)
+		}
+		pt = protoEnumType("")
```
Shua1, line 3782: We need to double check in DML, after parameter binding is available, if a parameter of type "" can be written into a column with type, "test.UserInfo", which essentially is a type mismatch.

You can also verify that internally.

spanner/value_test.go
gauravpurohit06: I tried it through Java Client and it doesn't matter what `protoTypeFqn` we are providing with null value it always get executed successfully.

harshachinta: As mentioned in below [comment](https://github.com/googleapis/google-cloud-go/pull/6954#discussion_r1008092202), the current changes do not pass a `""` string and instead fetches the correct `protoTypeFqn` even for nil types.

spanner/value.go
```diff
@@ -877,6 +877,96 @@ func (n *PGNumeric) UnmarshalJSON(payload []byte) error {
...
+	return jsonNullBytes, nil
+}
+
+// UnmarshalJSON implements json.Unmarshaler.UnmarshalJSON for NullProto.
+func (n *NullProto) UnmarshalJSON(payload []byte) error {
+	if payload == nil {
+		return fmt.Errorf("payload should not be nil")
+	}
+	if bytes.Equal(payload, jsonNullBytes) {
+		n.Valid = false
```
olavloite, line 913: Should we explicitly also set the `n.Value=nil`?

spanner/value.go
```diff
@@ -877,6 +877,96 @@ func (n *PGNumeric) UnmarshalJSON(payload []byte) error {
 	return nil
 }
 
+// NullProto represents a Cloud Spanner PROTO that may be NULL.
+type NullProto struct {
+	ProtoVal proto.Message // ProtoVal contains the value when it is non-NULL, and nil when NULL.
```
olavloite, line 882: When NULL: Is it really `nil`? Or is the zero value of `proto.Message`? (is `proto.Message` an interface or a struct?)

spanner/value.go
```diff
@@ -877,6 +877,96 @@ func (n *PGNumeric) UnmarshalJSON(payload []byte) error {
...
+	if err != nil {
+		return fmt.Errorf("payload cannot be converted to a proto message: got %v, err: %s", string(payload), err)
+	}
+	n.Valid = true
+	return nil
+}
+
+// NullEnum represents a Cloud Spanner ENUM that may be NULL.
+type NullEnum struct {
+	EnumVal protoreflect.Enum // EnumVal contains the value when it is non-NULL, and nil when NULL.
```
olavloite, line 926: Same here: Is it `nil` or `protoreflect.Enum{}` when the value is NULL?

spanner/value.go
```diff
@@ -877,6 +877,96 @@ func (n *PGNumeric) UnmarshalJSON(payload []byte) error {
...
+	return jsonNullBytes, nil
+}
+
+// UnmarshalJSON implements json.Unmarshaler.UnmarshalJSON for NullEnum.
+func (n *NullEnum) UnmarshalJSON(payload []byte) error {
+	if payload == nil {
+		return fmt.Errorf("payload should not be nil")
+	}
+	if bytes.Equal(payload, jsonNullBytes) {
+		n.Valid = false
```
olavloite, line 957: Here also: Should be explicitly set `n.Value` to the value it should have when the value is NULL?

spanner/value.go
```diff
@@ -944,9 +1034,9 @@ func errNilArrElemType(t *sppb.Type) error {
 	return spannerErrorf(codes.FailedPrecondition, "array type %v is with nil array element type", t)
 }
 
-// errNilDst returns error for decoding into nil interface{}.
-func errNilSrcPtr(dst interface{}) error {
-	return spannerErrorf(codes.InvalidArgument, "cannot use nil type %T", dst)
+// errNotValidSrc returns error is valid field is false
+func errNotValidSrc(dst interface{}) error {
+	return spannerErrorf(codes.InvalidArgument, "cannot have valid field as false in %T", dst)
```
olavloite, line 1039: I don't quite understand why we are changing this generic error? And also, the new error message does not really make sense to me. `cannot use nil type ...` would make me as a user understand that I used a nil where I should not. `cannot have valid field as false in ...` would not make me understand that I for example used a `nil` reference where I should not.

olavloite (commented):
I'm not familiar enough with proto columns to know whether all of this makes sense or not, so I will have to defer to others for that. I've left a couple of more generic comments.

spanner/value.go
harshachinta: `proto.Message` is an interface. The zero value or default value of an interface is `nil`.

spanner/value.go
harshachinta: `protoreflect.Enum` is an interface. The zero value or default value of an interface is `nil`.

spanner/value.go
harshachinta: @olavloite 
Previously we were restricting users from passing a custom nil type of type proto (ex: `(*pb.SingerInfo)(nil)`) due to the issue pointed out in [PR](https://github.com/googleapis/google-cloud-go/pull/6886#discussion_r1006305470). However, I found out a solution to allow nil types and therefore this error `errNilSrcPtr` is no longer needed. 
Note: Removing this error `errNilSrcPtr` will not affect existing code as no other methods use this except the Proto changes.

Now coming to error `errNotValidSrc`, custom types like `NullProto` and `NullEnum` have a field called `valid`. Users should explicitly set this field as `true` to indicate that the `value` in the struct is valid. However, if the `encodeValue` method receives this struct with `valid` as `false`, then the above error gets thrown to user.

A sample error would look like below
`cannot have valid field as false in spanner.NullProto`

spanner/value.go
harshachinta: Setting `Valid` field as `false` indicates user that value is `null`. However setting `value` to its zero value will be appropriate. Added code changes

spanner/value.go
harshachinta: Setting `Valid` field as `false` indicates user that value is `null`. However setting `value` to its zero value will be appropriate. Added code changes

spanner/value.go
olavloite: @harshachinta Thanks for the explanation.

I would still not know what I'm doing wrong if I get the error message `cannot have valid field as false in spanner.NullProto`. Is this field never allowed to be false? If so, why is it there? Or is it only disallowed in some cases? If so, the error message should indicate that.
Or is this some internal error that should never be returned, and is only there as a safety measure?

harshachinta: > Is this field never allowed to be false?

yes @olavloite . This field is used explicitly during `read` operation. This field `valid` is set by client library to `false`, whenever the data from db is `null`. 
During `write` operation, especially for protos we do not allow users to set `valid` field to be false. Other types like `NullString`, `NullInt64` considers this as a `null` value. However, for Protos we need to derive `proto_type_fqn` from the input being passed, and so does not allow users to set it to `false`. Hence the above exception gets thrown `cannot have valid field as false in spanner.NullProto`

Shua1: If that's the case, consider making the error msg more explicit that users are not expected to use this for writing.

But it does make it inconsistent with other Null{Type} structs.

For the case of not being able to infer the fqn, does go client library have the concept of untyped null? Are we able to make use of untyped null?

spanner/value.go
```diff
@@ -877,6 +877,98 @@ func (n *PGNumeric) UnmarshalJSON(payload []byte) error {
 	return nil
 }
 
+// NullProto represents a Cloud Spanner PROTO that may be NULL.
+type NullProto struct {
```
Shua1, line 881: We have been using ProtoMessage for proto in previous PR. Should we be consistent here?

spanner/value.go
```diff
@@ -1899,6 +2018,33 @@ func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...decodeO
 			return errBadEncoding(v, err)
 		}
 		proto.Unmarshal(y, p)
+	case *NullProto:
+		if p == nil {
+			return errNilDst(p)
+		}
+		if p.ProtoVal == nil {
+			return errNilDst(p.ProtoVal)
```
Shua1, line 2026: Does the users sees an error msg like this:
cannot decode into nil type proto.Message.

The original context of "NullProto.ProtoVal is nil" is lost. Is it possible to improve this?

Shua1: It this also true for other error msgs we give?

spanner/value.go
```diff
@@ -1899,6 +2018,33 @@ func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...decodeO
...
+			return errNotAPointer(p.ProtoVal)
+		}
+		if code != sppb.TypeCode_PROTO && code != sppb.TypeCode_BYTES {
+			return errTypeMismatch(code, acode, ptr)
+		}
+		if isNull {
+			*p = NullProto{}
+			break
+		}
+		x, err := getStringValue(v)
```
Shua1, line 2038: Line 2038 to 2045 is repeated multiple times in this file. Should we extract a method for it?

spanner/value.go
```diff
@@ -1877,6 +1969,33 @@ func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...decodeO
...
+			return errNotAPointer(p)
+		}
+		if code != sppb.TypeCode_ENUM && code != sppb.TypeCode_INT64 {
+			return errTypeMismatch(code, acode, ptr)
+		}
+		if isNull {
+			*p = NullEnum{}
+			break
+		}
+		x, err := getStringValue(v)
```
Shua1, line 1989: Line 1989 to 1996 is repeated multiple times. Should we extract method for them?

spanner/value.go
```diff
@@ -3620,30 +3766,42 @@ func encodeValue(v interface{}) (*proto3.Value, *sppb.Type, error) {
...
+				if err != nil {
+					return nil, nil, err
+				}
+				pb.Kind = stringKind(base64.StdEncoding.EncodeToString(bytes))
 			}
-			pb.Kind = stringKind(base64.StdEncoding.EncodeToString(bytes))
-			protoMessagefqn = string(proto.MessageReflect(v).Descriptor().FullName())
+			protoMessagefqn := string(proto.MessageReflect(v).Descriptor().FullName())
+			pt = protoMessageType(protoMessagefqn)
+		}
```
Shua1, line 3798: When would v be nil? Is doing nothing expected?

If it's expected, please add a comment.

spanner/integration_test.go
```diff
@@ -2043,8 +2043,18 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
 		val  interface{}
 		want interface{}
 	}{
+		// Proto Message
```
Shua1, line 2046: In addition to adding tests to this giant test, can we create an integration test that could also act like a sample?

So that we can clear see from a users' perspective, how these new interface looks like.

spanner/value.go
olavloite: I would expect this error message in that case to:
1. Say something like `field "valid" of spanner.NullProto may not be set to false when writing data to Cloud Spanner`. I feel that that explains better to the user exactly what they are doing wrong.
2. Include a hint that tells the user how they should write a null value, as this is inconsistent with other `Null{Type}` structs.

spanner/value.go
harshachinta: > When would v be nil? Is doing nothing expected?

v is `nil` only when a untyped null value is passed by user. As pb and pt are assigned by default in the beginning of switch case there is nothing to be done.

spanner/value.go
harshachinta: @olavloite 
As suggested I have modified the error to be more readable. Can you please review it?

spanner/integration_test.go
harshachinta: @Shua1 
Golang has a separate repository for samples and its associated integration tests that shows how users can use the functionality.
I am working on the samples currently and will be raising a separate PR for those changes.

spanner/value.go
harshachinta: Updated the comment to "when Valid is true"

spanner/value.go
harshachinta: @Shua1 
I have updated `NullProto` to be `NullProtoMessage` and `NullEnum` to be `NullProtoEnum`

spanner/value.go
harshachinta: I have created a method for the same.

spanner/value.go
harshachinta: Updated the error message to below for better readability
`field ProtoMessageVal in spanner.NullProtoMessage cannot be nil`

spanner/value.go
harshachinta: @Shua1 
We implement the `MarshalJSON()` and `UnmarshalJSON()` methods of `Marshaler` and `Unmarshaler` interfaces that belongs to the `json` package in golang. This is to provide the functionality of encoding custom types `NullProtoMessage` and `NullProtoEnum` into a `JSON` string and decoding `JSON` string into `NullProtoMessage` and `NullProtoEnum`.

**Usage:**
`json.Marshal` method can be used to encode a `NullProtoMessage` and `NullProtoEnum` into a JSON string.
`json.Unmarshal` method can be used to decode a JSON string into `NullProtoMessage` and `NullProtoEnum`.

As `NullProtoMessage` and `NullProtoEnum` are custom types we have to implement the above interfaces for the encode and decode to work as expected.

spanner/value.go
```diff
@@ -877,6 +877,98 @@ func (n *PGNumeric) UnmarshalJSON(payload []byte) error {
...
+// String implements Stringer.String for NullProto.
+func (n NullProto) String() string {
+	if !n.Valid {
+		return nullString
+	}
+	return n.ProtoVal.String()
+}
+
+// MarshalJSON implements json.Marshaler.MarshalJSON for NullProto.
+func (n NullProto) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
harshachinta: Tests for these methods are added.

rahul2393 (approved):

spanner/value.go
Shua1: Why do we need to encode and decode json?

Shua1 (approved):
It's a bit hard to imagine the impact of this change on how users would use it. Let's proceed to see the dedicated integration tests and samples. Do we have issue tracking for them?

spanner/value.go
```diff
@@ -877,6 +877,100 @@ func (n *PGNumeric) UnmarshalJSON(payload []byte) error {
 	return nil
 }
 
+// NullProtoMessage represents a Cloud Spanner PROTO that may be NULL.
+type NullProtoMessage struct {
+	ProtoMessageVal proto.Message // ProtoMessageVal contains the value when Valid is true, and nil when NULL.
+	Valid           bool          // Valid is true if ProtoMessageVal is not NULL.
```
olavloite, line 883: We should add an extra comment here to explain to users what they need to do to write a null value. Normally, they would be able to set `Valid: false` to create a null value, but if I understand it correctly, that is not the case for proto columns.

spanner/value.go
```diff
@@ -877,6 +877,100 @@ func (n *PGNumeric) UnmarshalJSON(payload []byte) error {
...
+		return fmt.Errorf("payload should not be nil")
+	}
+	if bytes.Equal(payload, jsonNullBytes) {
+		n.ProtoMessageVal = nil
+		n.Valid = false
+		return nil
+	}
+	err := json.Unmarshal(payload, n.ProtoMessageVal)
+	if err != nil {
+		return fmt.Errorf("payload cannot be converted to a proto message: got %v, err: %s", string(payload), err)
```
olavloite, line 919: I don't think we should include `string(payload)` in the error message. It could potentially be a __very__ long string.

spanner/value.go
```diff
@@ -877,6 +877,100 @@ func (n *PGNumeric) UnmarshalJSON(payload []byte) error {
...
+	}
+	n.Valid = true
+	return nil
+}
+
+// NullProtoEnum represents a Cloud Spanner ENUM that may be NULL.
+type NullProtoEnum struct {
+	ProtoEnumVal protoreflect.Enum // ProtoEnumVal contains the value when Valid is true, and nil when NULL.
+	Valid        bool              // Valid is true if ProtoEnumVal is not NULL.
```
olavloite, line 928: Same here regarding comments: We should explain to users how they would use this to write a null value, if they are not allowed to set this value to `false` when writing.

olavloite (approved):
LGTM, with some minor nits on comments and an error message

spanner/value.go
harshachinta: Removed the payload from the error message

harshachinta:
> It's a bit hard to imagine the impact of this change on how users would use it. Let's proceed to see the dedicated integration tests and samples. Do we have issue tracking for them?

harshachinta:
> It's a bit hard to imagine the impact of this change on how users would use it. Let's proceed to see the dedicated integration tests and samples. Do we have issue tracking for them?

As golang maintains a seperate repository for samples and its integration tests, I will be raising a PR for the dedicated samples and integration tests for proto columns. The draft PR for these changes can be found here - [PR-2754](https://github.com/GoogleCloudPlatform/golang-samples/pull/2754)

PR #6960, opened 2022-10-28
fix(bigquery/storage/managedwriter): remove old header routing

shollyman:
Previously, we needed to inject a routing header when opening the bidi AppendRows connection to successfully route traffic to the correct region.  Routing no longer needs this explicit hint, and works by examining the requests on the stream.

This change removes the header injection, and also simplifies the open contract we use to no longer include a stream ID, which was in place solely for the injection.

Related internal issue: 185842996

### DIFF
--- a/bigquery/storage/managedwriter/client.go
+++ b/bigquery/storage/managedwriter/client.go
@@ -27,7 +27,6 @@ import (
 	"google.golang.org/api/option"
 	storagepb "google.golang.org/genproto/googleapis/cloud/bigquery/storage/v1"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/metadata"
 )
 
 // DetectProjectID is a sentinel value that instructs NewClient to detect the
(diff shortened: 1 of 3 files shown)


alvarowolfx (approved):
LGTM

bigquery/storage/managedwriter/client.go
```diff
@@ -93,11 +92,11 @@ func (c *Client) NewManagedStream(ctx context.Context, opts ...WriterOption) (*M
...
 
 // createOpenF builds the opener function we need to access the AppendRows bidi stream.
-func createOpenF(ctx context.Context, streamFunc streamClientFunc) func(streamID string, opts ...gax.CallOption) (storagepb.BigQueryWrite_AppendRowsClient, error) {
-	return func(streamID string, opts ...gax.CallOption) (storagepb.BigQueryWrite_AppendRowsClient, error) {
+func createOpenF(ctx context.Context, streamFunc streamClientFunc) func(opts ...gax.CallOption) (storagepb.BigQueryWrite_AppendRowsClient, error) {
+	return func(opts ...gax.CallOption) (storagepb.BigQueryWrite_AppendRowsClient, error) {
 		arc, err := streamFunc(
-			// Bidi Streaming doesn't append stream ID as request metadata, so we must inject it manually.
-			metadata.AppendToOutgoingContext(ctx, "x-goog-request-params", fmt.Sprintf("write_stream=%s", streamID)), opts...)
+			// Previously, we needed to add metadata routing headers here, but this is no longer necessary.
```
yirutang, line 98: do you need to mention this?

bigquery/storage/managedwriter/client.go
```diff
@@ -27,7 +27,6 @@ import (
 	"google.golang.org/api/option"
 	storagepb "google.golang.org/genproto/googleapis/cloud/bigquery/storage/v1"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/metadata"
```
yirutang, line 30: add a e2e just to be sure?

shollyman: the existing e2es already test in non-default regions.

bigquery/storage/managedwriter/client.go
shollyman: removed

yirutang (approved):

PR #6970, opened 2022-10-31
fix(storage): use regex in v4SanitizeHeaders

rotemlinik:
The `v4SanitizeHeaders()` logic used `split(":")` func to separate the headers `key:val`, but in cases where the value itself contains `:` (such as dates in `RFC3339 format` as required), the value is partly lost.
I replaced the logic with a regex logic- same as the one used in `v2SanitizeHeaders()` and added a unit test for this case.


storage/storage.go
```diff
@@ -535,13 +535,12 @@ func v4SanitizeHeaders(hdrs []string) []string {
 		sanitizedHeader := strings.TrimSpace(hdr)
 
 		var key, value string
-		headerMatches := strings.Split(sanitizedHeader, ":")
-		if len(headerMatches) < 2 {
+		headerMatches := canonicalHeaderRegexp.FindStringSubmatch(sanitizedHeader)
```
tritone, line 538: I think not using the canonicalHeaderRegexp is actually an intentional difference between v2 and v4; see comments in L526-527.

However, maybe we should use [strings.SplitN](https://pkg.go.dev/strings#SplitN) instead so that only the first `:` is split on. Does that sound like a good solution to you?

BrennaEpp:
Closing this as #7603 contains the intended changes.

As of release v1.30.1, storage allows the use of values containing `:` in v4 headers

PR #6973, opened 2022-11-01
chore(bigtable): add timeouts to proxy methods

telpirion:
This PR applies the perOperationTimeout setting to the context object for each client call.

### DIFF
--- a/bigtable/internal/testproxy/proxy.go
+++ b/bigtable/internal/testproxy/proxy.go
@@ -356,6 +356,15 @@ type testClient struct {
 	isOpen              bool               // isOpen indicates whether this client is open for new requests
 }
 
+// timeout adds a timeout setting to a context if perOperationTimeout is set on
+// the testClient object.
+func (tc *testClient) timeout(ctx context.Context) (context.Context, context.CancelFunc) {
+	if tc.perOperationTimeout != nil {
+		return context.WithTimeout(ctx, tc.perOperationTimeout.AsDuration())
+	}
+	return ctx, nil
+}
+
 // credentialsBundle implements credentials.Bundle interface
 // [See documentation for usage](https://pkg.go.dev/google.golang.org/grpc/credentials#Bundle).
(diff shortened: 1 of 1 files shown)


bigtable/internal/testproxy/proxy.go
```diff
@@ -617,10 +626,7 @@ func (s *goTestProxyServer) ReadRow(ctx context.Context, req *pb.ReadRowRequest)
 		Row: &btpb.Row{},
 	}
 
-	if btc.perOperationTimeout != nil {
-		ct, _ := context.WithTimeout(ctx, btc.perOperationTimeout.AsDuration())
-		ctx = ct
-	}
+	ctx, _ = btc.timeout(ctx)
```
enocom, line 629: Let's add a `defer cancel()` on all of these.

See https://stackoverflow.com/a/44394873/927514.

enocom (changes requested):

bigtable/internal/testproxy/proxy.go
```diff
@@ -356,6 +356,15 @@ type testClient struct {
 	isOpen              bool               // isOpen indicates whether this client is open for new requests
 }
 
+// timeout adds a timeout setting to a context if perOperationTimeout is set on
+// the testClient object.
+func (tc *testClient) timeout(ctx context.Context) (context.Context, context.CancelFunc) {
+	if tc.perOperationTimeout != nil {
+		return context.WithTimeout(ctx, tc.perOperationTimeout.AsDuration())
+	}
+	return ctx, nil
```
enocom, line 365:
``` golang
return context.WithCancel(ctx)
```

enocom (changes requested):

bigtable/internal/testproxy/proxy.go
telpirion: Done.

enocom (approved):

Issue #5248, opened 2021-12-22
bigquery: retry on error code:500


careylam:
**Client**

BigQuery

**Environment**

MacOS

**Go Environment**

go version go1.17.2 darwin/amd64

$ go env

**Code**

e.g.
```go
func retryableError(err error) bool {
	if err == nil {
		return false
	}
	if err == io.ErrUnexpectedEOF {
		return true
	}
	// Special case due to http2: https://github.com/googleapis/google-cloud-go/issues/1793
	// Due to Go's default being higher for streams-per-connection than is accepted by the
	// BQ backend, it's possible to get streams refused immediately after a connection is
	// started but before we receive SETTINGS frame from the backend.  This generally only
	// happens when we try to enqueue > 100 requests onto a newly initiated connection.
	if err.Error() == "http2: stream closed" {
		return true
	}

	switch e := err.(type) {
	case *googleapi.Error:
		// We received a structured error from backend.
		var reason string
		if len(e.Errors) > 0 {
			reason = e.Errors[0].Reason
		}
		if e.Code == http.StatusServiceUnavailable || e.Code == http.StatusBadGateway || reason == "backendError" || reason == "rateLimitExceeded" {
			return true
		}
	case *url.Error:
		retryable := []string{"connection refused", "connection reset"}
		for _, s := range retryable {
			if strings.Contains(e.Error(), s) {
				return true
			}
		}
	case interface{ Temporary() bool }:
		if e.Temporary() {
			return true
		}
	}
	// Unwrap is only supported in go1.13.x+
	if e, ok := err.(interface{ Unwrap() error }); ok {
		return retryableError(e.Unwrap())
	}
	return false
}

```

**Expected behavior**
When e.Code == http.StatusInternalServerError, according to the error message from BigQuery,  it should retry. 

**Actual behavior**
The error is not retried and return directly. Here's a sample message from our production env. 
```
"error":"googleapi: Error 500: An internal error occurred and the request could not be completed. This is usually caused by a transient issue. Retrying the job with back-off as described in the BigQuery SLA should solve the problem: https://cloud.google.com/bigquery/sla. If the error continues to occur please contact support at https://cloud.google.com/support., internalError"
```

shollyman:
The `retryableError` predicate is used for retrying API method interactions (e.g. starting a query, getting table metadata, etc).

[...]

PR #6976, opened 2022-11-01
feat(bigquery): widen retry predicate

shollyman:
This PR adds 500,504 http response codes for the default retry predicate on unary retries.

This doesn't introduce job-level retries (jobs must be recreated wholly, they can't be effectively restarted), so the primary risk of this change is terminal job state propagating into the HTTP response code of one of the polling methods (e.g. job.getQueryResults, jobs.get, etc).  In this situation, job execution may appear to hang indefinitely or until context cancellation/expiry.

Related: https://github.com/googleapis/google-cloud-go/issues/5248


alvarowolfx (approved):
LGTM. But I think we might need to start some design doc or discussion around making the retry logic more configurable. Do we know of other client libraries that allow that to take some inspiration ?

shollyman:
> LGTM. But I think we might need to start some design doc or discussion around making the retry logic more configurable. Do we know of other client libraries that allow that to take some inspiration ?

There's some hooks we can consider such as https://pkg.go.dev/github.com/googleapis/gax-go/v2#WithRetry

Storage rolled their own client-level settings: https://pkg.go.dev/cloud.google.com/go/storage#Client.SetRetry

PR #6977, opened 2022-11-01
chore(bigtable): more text proxy tune-up

telpirion:
This PR:

* Changes how table IDs are passed to `Client.Open()`
* Changes the error returned to the caller for a closed client.

### DIFF
--- a/bigtable/internal/testproxy/proxy.go
+++ b/bigtable/internal/testproxy/proxy.go
@@ -514,15 +514,15 @@ type goTestProxyServer struct {
 
 // client retrieves a testClient from the clientIDs map. You must lock clientsLock before calling
 // this method.
-func (s *goTestProxyServer) client(clientID string) (*testClient, bool) {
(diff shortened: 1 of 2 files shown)


bigtable/internal/testproxy/proxy.go
```diff
@@ -507,15 +529,15 @@ type goTestProxyServer struct {
 
 // client retrieves a testClient from the clientIDs map. You must lock clientsLock before calling
 // this method.
```
quartzmo, line 531: Maybe move this comment into an error message?

quartzmo: > You must lock clientsLock before calling this method.

quartzmo (approved):
LGTM, but with question.

bigtable/internal/testproxy/proxy.go
telpirion: Ack. That makes sense, but is out-of-scope for this review. I'll consider your suggestion for future improvements.

PR #6991, opened 2022-11-03
feat(bigtable): add support for request stats

humbertowastaken:
feat(bigtable): add support for request stats


bigtable/bigtable.go
```diff
@@ -467,13 +479,30 @@ func RowFilter(f Filter) ReadOption { return rowFilter{f} }
 type rowFilter struct{ f Filter }
 
 func (rf rowFilter) set(req *btpb.ReadRowsRequest) { req.Filter = rf.f.proto() }
+func (rf rowFilter) getRequestStatsCallback() RequestStatsCallback {
```
mutianf, line 482: Hmm I'm unfamiliar with go syntax but why do we need to getRequestStatsCallback for rowFilter or Row limit options? And what does getRequestStatsCallback do?

mutianf (commented):
Is there a way to add a test for this?

bigtable/bigtable.go
alexoneill: Since the `ReadOption` interface includes the new method `getRequestStatsCallback` signature, for `rowFilter` and `limitRows` to adhere to this interface, they need to include an implementation of that method.

Go's interface system is not totally clear to readers IMO, but Go will automatically mark `rowFilter` and `limitRows` as implementing this interface at compile time as long as the methods match up.

Without these changes, `RowFilter` above breaks since it indicates that it returns a `ReadOption`, which means that `rowFilter` must have a method that matches the `getRequestStatsCallback` spec, else it can't be considered as part of this interface.

bigtable/bigtable.go
```diff
@@ -453,9 +461,13 @@ func prefixSuccessor(prefix string) string {
 	return string(ans)
 }
 
+// RequestStatsCallback describes a callback that receives a RequestStats for evaluation.
+type RequestStatsCallback func(*btpb.RequestStats)
```
telpirion, line 465: Please do not expose raw proto types in the manual client layers.

bigtable/bigtable.go
```diff
@@ -453,9 +461,13 @@ func prefixSuccessor(prefix string) string {
...
+
 // A ReadOption is an optional argument to ReadRows.
 type ReadOption interface {
 	set(req *btpb.ReadRowsRequest)
+	getRequestStatsCallback() RequestStatsCallback
```
telpirion, line 470: This is very wordy for Go. I would shorten this to `RequestStatsFunc`.

telpirion, line 470: If I read this correctly, every distinct type that implements the ReadOption interface must now have the `getRequestStatsCallback()` function. As this is a private method, it won't cause a breaking change in the API surface, but it will require refactoring all existing ReadOption types.

bigtable/bigtable.go
```diff
@@ -467,13 +479,30 @@ func RowFilter(f Filter) ReadOption { return rowFilter{f} }
...
 type limitRows struct{ limit int64 }
 
 func (lr limitRows) set(req *btpb.ReadRowsRequest) { req.RowsLimit = lr.limit }
+func (lr limitRows) getRequestStatsCallback() RequestStatsCallback {
+	return func(*btpb.RequestStats) {}
+}
+
+// WithRequestStatsCallback returns a ReadOption that will request RequestStats
+// and invoke the given callback on the resulting RequestStats.
+func WithRequestStatsCallback(f RequestStatsCallback) ReadOption { return withRequestStatsCallback{f} }
```
telpirion, line 498: This is also a bit wordy. I would shorten to `WithRequestStats`. The type of the parameter tells the developer that it's a function.

telpirion (changes requested):

telpirion:
This PR also requires a unit test and an integration test.

bigtable/bigtable.go
humbertowastaken: Fixed.

humbertowastaken:
Added integration tests.

bigtable/bigtable.go
```diff
@@ -453,9 +461,53 @@ func prefixSuccessor(prefix string) string {
 	return string(ans)
 }
 
+// ReadIterationStats captures information about the iteration of rows or cells over the course of
+// a read, e.g. how many results were scanned in a read operation versus the results returned.
+type ReadIterationStats struct {
```
telpirion, line 466: Need comments for fields in the `ReadIterationStats` struct.

bigtable/bigtable.go
```diff
@@ -453,9 +461,53 @@ func prefixSuccessor(prefix string) string {
...
+	CellsReturnedCount int64
+	CellsSeenCount     int64
+	RowsReturnedCount  int64
+	RowsSeenCount      int64
+}
+
+// RequestLatencyStats provides a measurement of the latency of the request as it interacts with
+// different systems over its lifetime, e.g. how long the request took to execute within a frontend
+// server.
+type RequestLatencyStats struct {
```
telpirion, line 476: Need comments for fields in the `RequestLatencyStats` struct.

bigtable/bigtable.go
```diff
@@ -453,9 +461,53 @@ func prefixSuccessor(prefix string) string {
...
+	FrontendServerLatency time.Duration
+}
+
+// FullReadStats captures all known information about a read.
+type FullReadStats struct {
```
telpirion, line 481: Need comments for fields in the `FullReadStats` struct.

bigtable/bigtable.go
```diff
@@ -453,9 +461,53 @@ func prefixSuccessor(prefix string) string {
...
+	ReadIterationStats  ReadIterationStats
+	RequestLatencyStats RequestLatencyStats
+}
+
+func makeFullReadStats(reqStats *btpb.RequestStats) FullReadStats {
```
telpirion, line 486: NIT: a small comment before this function will be useful for future Golang client maintainers.

bigtable/bigtable.go
```diff
@@ -453,9 +461,53 @@ func prefixSuccessor(prefix string) string {
...
+	statsView := reqStats.GetFullReadStatsView()
```
telpirion, line 487: Although this is guarded by an if statement above, I recommend a nil check here.

bigtable/bigtable.go
```diff
@@ -453,9 +461,53 @@ func prefixSuccessor(prefix string) string {
...
+	readStats := statsView.ReadIterationStats
+	latencyStats := statsView.RequestLatencyStats
+	return FullReadStats{
+		ReadIterationStats{readStats.CellsReturnedCount, readStats.CellsSeenCount, readStats.RowsReturnedCount, readStats.RowsSeenCount},
```
telpirion, line 491: NIT: Preferred style for Go is to instantiate an object by providing explicit field assignments.

E.g.

```
return FullReadStats {
    CellsReturnedCount: readStats.CellsReturnedCount,
    // etc etc
}
```

telpirion (changes requested):
I think this still needs a unit test in `bigtable_test.go`. The Go monorepo doesn't run integration tests for each PR until it has been merged. As a result, we need to ensure that unit tests exist to help us see breakages before a PR is merged.

In sum, bare minimum required changes:

+ unit test(s) for this callback feature in `bigtable_test.go`
+ comments for all publicly exposed fields to populate the reference docs

bigtable/bigtable.go
humbertowastaken: Done.

bigtable/bigtable.go
humbertowastaken: Done.

bigtable/bigtable.go
igorbernstein2: What would you recommend to do in the nil case? Since this is an internal impl detail, we can update this in a follow up PR

igorbernstein2:
At this point the only thing that missing is a unit test. This will require some refactoring in the emulator code, which we can do in a follow up PR. The functionality is covered by the integration test. Given the timing, I'm going to go ahead and merge this.

igorbernstein2 (approved):
lgtm.

PR #6994, opened 2022-11-03
chore(storage): gRPC support customer encryption

noahdietz:
Enables use of customer provided encryption for Object media operations (Read, Rewrite, Write (including resumable), etc.).

Enables both `TestIntegration_Encryption` and `TestIntegration_KMS` for gRPC.

### DIFF
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -1512,8 +1512,7 @@ func (w *gRPCWriter) queryProgress() (int64, error) {
 	var persistedSize int64
 	err := run(w.ctx, func() error {
 		q, err := w.c.raw.QueryWriteStatus(w.ctx, &storagepb.QueryWriteStatusRequest{
-			UploadId:                  w.upid,
-			CommonObjectRequestParams: toProtoCommonObjectRequestParams(w.encryptionKey),
+			UploadId: w.upid,
 		})
 		persistedSize = q.GetPersistedSize()
 		return err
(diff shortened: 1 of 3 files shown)


storage/grpc_client.go
```diff
@@ -1511,7 +1511,10 @@ func (w *gRPCWriter) startResumableUpload() error {
 func (w *gRPCWriter) queryProgress() (int64, error) {
 	var persistedSize int64
 	err := run(w.ctx, func() error {
-		q, err := w.c.raw.QueryWriteStatus(w.ctx, &storagepb.QueryWriteStatusRequest{UploadId: w.upid})
+		q, err := w.c.raw.QueryWriteStatus(w.ctx, &storagepb.QueryWriteStatusRequest{
+			UploadId:                  w.upid,
+			CommonObjectRequestParams: toProtoCommonObjectRequestParams(w.encryptionKey),
```
tritone, line 1516: Why does this RPC need encryptionKey?

tritone (commented):
One small question, otherwise LGTM! Thanks Noah.

storage/grpc_client.go
noahdietz: TBH IDK but it has the parameters. I can exclude it, but I figured I'd err on the side of caution. I can remove it though, wdyt?

noahdietz: I've removed it, its just silly atm.

tritone (approved):

PR #6999, opened 2022-11-07
feat(all): enable REGAPIC and REST numeric enums

noahdietz:
Enables REGAPIC and REST numeric enum serialization for all API clients, except for those that are known to only support one transport / not support REST numeric enum serialization.

`["grpc", "rest"]` is now the default set of transports generated for a package. REST numeric enum serialization is enabled by default. As such, this removes all explicit `Transports` instances that match the new default.

Note: There are three beta-level packages that will take breaking changes in a follow-up PR that remove REST support.


noahdietz:
I regenerated everything, pushed it to this PR, and `kokoro` was appeased. There was an issue with `vet.sh` which was my lack of `gofmt` prior to pushing. I've since reverted the generated changes to allow our job to do it.

noahdietz:
Similarly, I saw in a PR `apidiff` run on all affected packages we should not have any breaking changes in the next regen using this new config. There will be some follow up PRs to break a few clients that must be broken.

PR #7006, opened 2022-11-09
feat(storage): add storage buckets SelfLink field

amanenk:
My project needs self link of storage bucket to be displayed. 
The PR adds self link field  to `BucketAttrs` structure.

### DIFF
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -445,6 +445,10 @@ type BucketAttrs struct {
 	// more information.
 	RPO RPO
 
+	// SelfLink: The URI of this bucket.
+	// This field is read-only.
+	SelfLink string
+
 	// Autoclass holds the bucket's autoclass configuration. If enabled,
 	// allows for the automatic selection of the best storage class
 	// based on object access patterns.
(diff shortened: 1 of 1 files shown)


tritone (commented):
Can you speak more to why this is necessary for your use case?

In general all of the GCS resource types (bucket, object, ACL, etc) have a `selfLink` field, but we don't expose it via the library because it's not really relevant to users who use the SDK -- it's more of a construct for using the API directly.

In addition, it's easy to reconstruct a selfLink for a bucket; it's just going to be the standard API path for the bucket name in the JSON API (`https://www.googleapis.com/storage/v1/b/` + your bucket name).

amanenk:
Thank you for your fast response. We develop the tool that fetches resources from the cloud and checks fetched data if the configuration matches security requirements. It turned out that self-link (the field exists in other resources) is a part of the standard output of our tool.  Adding a `https://www.googleapis.com/storage/v1/b/` looks like a workaround and might break something in the future. Also, it feels a bit wrong to build a self-link when it is already returned from the API.

tritone:
> Thank you for your fast response. We develop the tool that fetches resources from the cloud and checks fetched data if the configuration matches security requirements. It turned out that self-link (the field exists in other resources) is a part of the standard output of our tool. Adding a `https://www.googleapis.com/storage/v1/b/` looks like a workaround and might break something in the future. Also, it feels a bit wrong to build a self-link when it is already returned from the API.

Makes sense. As I said, the selfLink field is really a construct of the JSON API for navigating resources, rather than an actual property of the bucket resource. I checked and see that selfLink is not exposed via the XML API, the GCP Cloud Console, or gsutil. Therefore I don't think it makes sense for us to expose it via the client library either.

amanenk:
Thank you for explanation. I might be wrong but it seems that `selfLink` is exposed in some cases. when I run `gcloud storage buckets describe` it gives selfLink field

PR #7011, opened 2022-11-09
feat(routing): start generating apiv2

quartzmo:


internal/gapicgen/generator/config.go
```diff
@@ -280,8 +280,7 @@ var MicrogenGapicConfigs = []*MicrogenConfig{
 		ApiServiceConfigPath:  "bigqueryreservation_v1beta1.yaml",
 		Transports:            []string{"grpc", "rest"},
 		ReleaseLevel:          "beta",
-	},
-	{
+	}, {
```
codyoss, line 283: nit: this looks unintentional

routing/go.mod
```diff
@@ -0,0 +1,26 @@
+module cloud.google.com/go/routing
```
codyoss, line 1: I bet we will eventually be generating more maps APIs. I think we should make the module under `cloud.google.com/go/maps` and have routing be a subdir

codyoss (changes requested):

routing/go.mod
quartzmo: Great, will do.

quartzmo: Fixed, thanks!

internal/gapicgen/generator/config.go
quartzmo: Fixed, thanks!

codyoss (approved):

PR #7012, opened 2022-11-10
feat(addressvalidation): start generating apiv1

quartzmo:


addressvalidation/go.mod
```diff
@@ -0,0 +1,26 @@
+module cloud.google.com/go/addressvalidation
```
codyoss, line 1: Same as the other PR, I think we should group these under a maps module.

codyoss (changes requested):

addressvalidation/go.mod
quartzmo: Agreed, will do.

maps/README.md
```diff
@@ -0,0 +1,44 @@
+# Address Validation API
+
+[![Go Reference](https://pkg.go.dev/badge/cloud.google.com/go/maps.svg)](https://pkg.go.dev/cloud.google.com/go/maps)
+
+Go Client Library for Address Validation API.
```
codyoss, line 5: This should be more generic for maps

codyoss (changes requested):
Just one small thing.

maps/README.md
quartzmo: Will do.

quartzmo: Fixed, thanks!

codyoss (approved):

PR #7014, opened 2022-11-10
feat(spanner): add support for Array of Proto columns

harshachinta:
This PR contains implementation changes to add support for handling array of Proto Messages and Proto Enum type columns.

To run integration tests, set the below environment variables:
```
GCLOUD_TESTS_GOLANG_PROJECT_ID=span-cloud-testing
GCLOUD_TESTS_GOLANG_SPANNER_HOST=staging-wrenchworks.sandbox.googleapis.com:443
```

### DIFF
--- a/spanner/integration_test.go
+++ b/spanner/integration_test.go
@@ -2036,6 +2036,13 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
 		Nationality: proto.String("Country1"),
 		Genre:       &singerProtoEnum,
 	}
+	singer2ProtoEnum := pb.Genre_FOLK
+	singer2ProtoMessage := pb.SingerInfo{
+		SingerId:    proto.Int64(1),
+		BirthDate:   proto.String("January"),
+		Nationality: proto.String("Country1"),
+		Genre:       &singer2ProtoEnum,
+	}
 	bytesSingerProtoMessage, _ := proto.Marshal(&singerProtoMessage)
 
 	tests := []struct {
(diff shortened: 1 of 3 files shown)


Shua1 (commented):
Was the previous conversations lost?

spanner/integration_test.go
```diff
@@ -2091,6 +2098,50 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
 		{col: "ProtoMessage", val: []byte(nil), want: NullProtoMessage{}},
 		{col: "ProtoMessage", val: NullProtoMessage{&singerProtoMessage, true}, want: bytesSingerProtoMessage},
 		{col: "ProtoMessage", val: NullProtoMessage{&singerProtoMessage, true}, want: NullProtoMessage{&singerProtoMessage, true}},
+		// Array of Proto Messages : Tests insert and read operations on ARRAY<PROTO> type column
+		{col: "ProtoMessageArray", val: []*pb.SingerInfo{&singerProtoMessage, &singer2ProtoMessage}},
+		{col: "ProtoMessageArray", val: []*pb.SingerInfo(nil)},
+		{col: "ProtoMessageArray", val: []*pb.SingerInfo{}},
+		// Array of Proto Enum : Tests insert and read operations on ARRAY<ENUM> type column
+		{col: "ProtoEnumArray", val: []pb.Genre{pb.Genre_ROCK, pb.Genre_FOLK}, want: []*pb.Genre{&singerProtoEnum, &singer2ProtoEnum}},
```
Shua1, line 2106: Add comment:
Enum value array.

spanner/integration_test.go
```diff
@@ -2091,6 +2098,50 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
...
+		{col: "ProtoEnumArray", val: []pb.Genre{pb.Genre_ROCK, pb.Genre_FOLK}, want: []pb.Genre{singerProtoEnum, singer2ProtoEnum}},
+		{col: "ProtoEnumArray", val: []*pb.Genre{&singerProtoEnum, &singer2ProtoEnum}},
```
Shua1, line 2108: Add comment:
Enum ptr array.

spanner/integration_test.go
```diff
@@ -2091,6 +2098,50 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
...
+		{col: "ProtoEnumArray", val: []*pb.Genre{&singerProtoEnum, &singer2ProtoEnum}, want: []pb.Genre{singerProtoEnum, singer2ProtoEnum}},
+		{col: "ProtoEnumArray", val: []pb.Genre{}, want: []*pb.Genre{}},
+		{col: "ProtoEnumArray", val: []pb.Genre(nil), want: []*pb.Genre(nil)},
```
Shua1, line 2111: Please comment what this case is.

Maybe it's obivous for a go expert.

spanner/integration_test.go
```diff
@@ -2091,6 +2098,50 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
...
+		{col: "ProtoEnumArray", val: []*pb.Genre{}, want: []pb.Genre{}},
+		{col: "ProtoEnumArray", val: []*pb.Genre(nil), want: []pb.Genre(nil)},
+		{col: "ProtoEnumArray", val: []pb.Genre{}},
+		{col: "ProtoEnumArray", val: []pb.Genre(nil)},
+		{col: "ProtoEnumArray", val: []*pb.Genre{}},
+		{col: "ProtoEnumArray", val: []*pb.Genre(nil)},
```
Shua1, line 2117: Consider adding test cases (for both enum and proto)
with multiple nil values
with mixed nil and non-nil values.

Need to check if this is possible in go and accepted by the server.

Shua1 (approved):
Generally LGTM, especially the test case. I will left the code style for the go client library experts, please double check with them. Thanks!

rahul2393 (approved):

spanner/value.go
```diff
@@ -2069,6 +2072,47 @@ func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...decodeO
...
+			typ := rv.Type()
+			if typ.Kind() != reflect.Ptr {
+				return errNotAPointer(ptr)
+			}
+			if isNull {
+				break
+			}
+			etyp := typ.Elem().Elem()
+			switch acode {
+			case sppb.TypeCode_PROTO, sppb.TypeCode_BYTES:
```
gauravpurohit06, line 2092: Please break it into smaller functions for better readability.  One for PROTO/BYTES and other for ENUM/INT64.

spanner/value.go
```diff
@@ -4119,6 +4230,42 @@ func encodeStructArray(v interface{}) (*proto3.Value, *sppb.Type, error) {
...
+// Encodes a slice of proto messages or enum in v to the spanner Value and Type
+// protos.
+func encodeProtoArray(v interface{}) (*proto3.Value, *sppb.Type, error) {
+	pb := nullProto()
+	var pt *sppb.Type
+	var err error
+	sliceval := reflect.ValueOf(v)
+	etyp := reflect.TypeOf(v).Elem()
+
+	if etyp.Implements(protoMsgReflectType) {
```
gauravpurohit06, line 4242: same as above. 
Please break it into smaller functions for better readability,  one for `protoMsgReflectType` and other for `protoEnumReflectType`.

gauravpurohit06 (approved):
Approve with minor comments.

spanner/value.go
```diff
@@ -2069,6 +2072,47 @@ func decodeValue(v *proto3.Value, t *sppb.Type, ptr interface{}, opts ...decodeO
 			return decodableType.decodeValueToCustomType(v, t, acode, atypeAnnotation, ptr)
 		}
 
+		// Check if the proto encoding is for an array of proto message or array of proto enum.
+		// The below scenario cannot be handled in a switch case statement because the pointer in this case
+		// would be an array of proto message or array of proto enum which are user defined types.
+		// Adding a generic switch case for these user defined types is not possible, hence we use proto encoding.
+		if code == sppb.TypeCode_ARRAY && (acode == sppb.TypeCode_PROTO ||
+			acode == sppb.TypeCode_ENUM || acode == sppb.TypeCode_INT64 ||
+			acode == sppb.TypeCode_BYTES) {
```
olavloite, line 2081: I don't think we should go into this branch for `INT64` and `BYTES`. As far as I can see, it does work in this case (that is; it does not interfere with any other cases), but it feels dangerous. We are basically saying that the 'catch-all' case for `ARRAY<INT64>` is `ProtoEnum` and for `ARRAY<BYTES>` is `Proto`. That is a choice that we cannot undo in the future.

spanner/integration_test.go
```diff
@@ -2169,6 +2224,95 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
...
+	protoEnumNilError := "*protos.Genre cannot support NULL SQL values"
+
+	tests := []struct {
+		col       string
+		val       interface{}
+		want      interface{}
+		wantCode  codes.Code
+		errString string
+	}{
+		// Array of Proto Messages : Tests read operation on ARRAY<PROTO> type column that has untyped/typed nil elements in array
```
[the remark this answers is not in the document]
harshachinta: We will be supporting `null` elements in the array of protos or enum using `ARRAY<NullProtoMessage>` and `ARRAY<NullProtoEnum>` (This is similar to existing `ARRAY<NullTime>`, `ARRAY<NullDate>` etc in client library). `NullProtoMessage` and `NullProtoEnum` was added in [PR-6954](https://github.com/googleapis/google-cloud-go/pull/6954). Will raise a separate PR for handling null proto and enum values in array.

spanner/integration_test.go
```diff
@@ -2169,6 +2224,95 @@ func TestIntegration_BasicTypes_ProtoColumns(t *testing.T) {
...
+		{col: "ProtoEnumArray", val: []*pb.Genre{(*pb.Genre)(nil), (*pb.Genre)(nil)}, wantCode: codes.InvalidArgument, errString: protoEnumNilError},
+	}
+
+	// Delete all the rows.
+	_, err := client.Apply(ctx, []*Mutation{Delete("Types", AllKeys())})
+	if err != nil {
+		t.Fatalf("failed to delete all rows: %v", err)
+	}
+
+	// Insert the rows using mutations.
```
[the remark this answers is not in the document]
harshachinta: @olavloite
The tests for that are already there in [integration_test.go](https://github.com/googleapis/google-cloud-go/blob/33419f9f5267aa67e8b240134251c13d074cfd58/spanner/integration_test.go#L2151) file. The remaining tests if any will be raised as a separate PR

spanner/integration_test.go
```diff
@@ -3882,7 +4026,7 @@ func prepareDBAndClientForProtoColumns(ctx context.Context, t *testing.T, spc Se
 	dbName := dbNameSpace.New()
 	// TODO: Remove this
 	testInstanceID = "go-int-test-proto-column"
-	dbName = "go_int_test_proto_column"
+	dbName = "go_int_test_proto_column_db"
```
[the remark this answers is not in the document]
harshachinta: This was to differentiate between `testInstanceID` and `dbName`.

spanner/value.go
harshachinta: @olavloite Here we do not 'catch-all' case for `ARRAY<INT64>` as `ProtoEnum` and for `ARRAY<BYTES>` as `ProtoMessage`. It depends on the interface that is passed by the user for decoding.

Let me try to explain this scenario for `ARRAY<INT64>`. Let us assume that spanner backend returns `code` as  `sppb.TypeCode_ARRAY`and `acode` as `sppb.TypeCode_INT64`.

**Scenario 1:**
If the user passes `[]int64` as the interface to read data then it gets caught in `decodeValue` function way before in the `switch` case as pointed [here](https://github.com/googleapis/google-cloud-go/blob/33419f9f5267aa67e8b240134251c13d074cfd58/spanner/value.go#L1363).
**Scenario 2:**
Similarly, if the user passes an `[]ProtoEnum` as the interface to read data then it executes the above [code](https://github.com/googleapis/google-cloud-go/blob/33419f9f5267aa67e8b240134251c13d074cfd58/spanner/value.go#L2079) as the interface passed does not satisfy any case in `switch` condition and comes down here.

I have added integration tests for these scenarios in [Scenario 1](https://github.com/googleapis/google-cloud-go/blob/33419f9f5267aa67e8b240134251c13d074cfd58/spanner/integration_test.go#L1854) and [Scenario 2](https://github.com/googleapis/google-cloud-go/blob/33419f9f5267aa67e8b240134251c13d074cfd58/spanner/integration_test.go#L2123)

spanner/integration_test.go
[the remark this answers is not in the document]
olavloite: OK

spanner/integration_test.go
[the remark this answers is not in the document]
olavloite: Thanks!

spanner/integration_test.go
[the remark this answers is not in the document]
olavloite: I meant the `TODO: Remove this`. Is that something that should be removed now, or in a later PR.

spanner/value.go
olavloite: Yes, I understand that, so it's not a full catch-all case for `ARRAY<INT64>`, but it is for everything __except__ `[]int64`. That means that if we in the future have some other type that we want to map to `ARRAY<INT64>` with a specific Go type, then we have a problem. Because that would then introduce a breaking change, as it would then no longer go into this branch, but into the branch for the specific type.

spanner/integration_test.go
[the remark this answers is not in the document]
harshachinta: The `TODO` will be removed in a later PR which supports DDL operations through client library.

spanner/value.go
harshachinta: @olavloite 
I have modified the decoding logic for proto array by checking the interface passed instead of checking the proto encoding passed from spanner backend - [Link](https://github.com/googleapis/google-cloud-go/blob/0ea08c1b557aff2aa32add463422c340f0c2349e/spanner/value.go#L2078)
This will not cause a breaking change as we go into the branch only if interface passed is of type array of proto columns. Can you please verify this?

spanner/integration_test.go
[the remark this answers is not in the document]
olavloite: ACK

spanner/value.go
olavloite: Thanks, LGTM

olavloite (approved):

PR #7025, opened 2022-11-11
chore(datastore): Add trace span for RunAggregationQuery.

xwkuang5:
This change add a trace span for a RunAggregationQuery call.


datastore/client.go
```diff
@@ -73,6 +73,17 @@ func (dc *datastoreClient) RunQuery(ctx context.Context, in *pb.RunQueryRequest,
 	return res, err
 }
 
+func (dc *datastoreClient) RunAggregationQuery(ctx context.Context, in *pb.RunAggregationQueryRequest, opts ...grpc.CallOption) (res *pb.RunAggregationQueryResponse, err error) {
```
telpirion, line 76: Do not add this code here. `RunAggregationQuery` is already in this package. If you'd like to add the trace in the `Client.RunAggregationQuery()` method, add that code here:

https://github.com/googleapis/google-cloud-go/blob/main/firestore/query.go#L1091

telpirion, line 76: We don't use raw protobuf types in the public API surface.

telpirion (changes requested):

datastore/client.go
xwkuang5: Does `AggregationQuery.Get` exist for datastore? I only saw the firestore's one.

xwkuang5: Not sure if I follow. It seems the `RunQuery` above also uses raw proto type?

telpirion: Oops, that was the wrong link:
https://github.com/googleapis/google-cloud-go/blob/main/datastore/query.go#L629

telpirion: The types exposed in that file aren't used directly by the client library. (I believe the file that you're adding code to is basically unused, but hadn't been able to prove that conclusively.)

xwkuang5: Wouldn't the file I'm adding code to get called from here [1] since it overrides the `RunAggregationQuery` call and similarly for `RunQuery` and others? 

[1] https://github.com/googleapis/google-cloud-go/blob/feb6b0a8b771875a55c75fa959e19f6f12079030/datastore/query.go#L663

telpirion: Nope, check out the [import](https://github.com/googleapis/google-cloud-go/blob/main/datastore/query.go#L30) and the [type `Client.client`](https://github.com/googleapis/google-cloud-go/blob/main/datastore/datastore.go#L59).

telpirion: Additional: Golang doesn't have the concept of overrides or shadowing in the same way that other languages do. If you have two methods which collide, the compiler simply throws an error.

xwkuang5: Got it, thanks for explaining.

datastore/query.go
```diff
@@ -745,6 +745,9 @@ func (c *Client) Run(ctx context.Context, q *Query) *Iterator {
 
 // RunAggregationQuery gets aggregation query (e.g. COUNT) results from the service.
 func (c *Client) RunAggregationQuery(ctx context.Context, aq *AggregationQuery) (AggregationResult, error) {
```
bhshkh, line 747: The build fails with err undefined on line 749
To define err, add named return values.

bhshkh (changes requested):

datastore/query.go
xwkuang5: Done.

bhshkh (approved):

PR #7042, opened 2022-11-15
feat(spanner): add support for handling null values in array of proto columns

harshachinta:
This PR adds support for handling null values in array of proto columns.

### DIFF
--- a/spanner/value.go
+++ b/spanner/value.go
@@ -1080,6 +1080,10 @@ func errNotAPointerField(dst interface{}, dstField interface{}) error {
 	return spannerErrorf(codes.InvalidArgument, "destination %T in %T must be a pointer", dstField, dst)
 }
 
+func errNilNotAllowed(dst interface{}, name string) error {
+	return spannerErrorf(codes.InvalidArgument, "destination %T do not support Null values. Use %s, an array with pointer type elements to read Null values", dst, name)
+}
+
 func parseNullTime(v *proto3.Value, p *NullTime, code sppb.TypeCode, isNull bool) error {
 	if p == nil {
 		return errNilDst(p)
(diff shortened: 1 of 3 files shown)


rahul2393 (approved):

spanner/value.go
```diff
@@ -3210,6 +3222,12 @@ func decodeProtoEnumArray(pb *proto3.ListValue, t *sppb.Type, rv reflect.Value)
 	// As the ENUM element in the Array is not a pointer type we cannot use decodeValue method
 	// and hence handle it separately.
 	for i, v := range pb.Values {
+		_, isNull := v.Kind.(*proto3.Value_NullValue)
+		// As the ENUM elements in the array are value type and not pointer type,
+		// we cannot support NULL values in the array
+		if isNull {
+			return errNilNotAllowed(ptr, "*[]*protoreflect.Enum")
```
Shua1, line 3229: The error msg is:

... Use *[]*protoreflect.Enum, an array with pointer type elements to read Null values.

Is this expected?

spanner/value.go
```diff
@@ -1080,6 +1080,10 @@ func errNotAPointerField(dst interface{}, dstField interface{}) error {
 	return spannerErrorf(codes.InvalidArgument, "destination %T in %T must be a pointer", dstField, dst)
 }
 
+func errNilNotAllowed(dst interface{}, name string) error {
+	return spannerErrorf(codes.InvalidArgument, "destination %T do not support Null values. Use %s, an array with pointer type elements to read Null values", dst, name)
```
olavloite, line 1084: nit:
```suggestion
	return spannerErrorf(codes.InvalidArgument, "destination %T does not support Null values. Use %s, an array with pointer type elements to read Null values", dst, name)
```

olavloite (approved):
LGTM, with a small nit on the error message.

PR #7062, opened 2022-11-17
feat(spanner): update go-genproto dependency

harshachinta:
This PR is created to update the go-genproto dependency to point to cloudspannerecosystem repository until the changes are public released.


harshachinta:
This PR is no longer needed as the Proto files are moved with in the google-cloud-go repository and the dependency to go-genproto for proto column feature changes [go-genproto proto columns](https://github.com/cloudspannerecosystem/temp-resources/tree/go-gen-proto-column-enhancement-alpha) repository is no longer required.

Issue #6391, opened 2022-07-21
logging: httpRequest RedirectAsJSON camel vs snake casing


nwadams:
**Client**

GCP logging [client](https://pkg.go.dev/cloud.google.com/go/logging) version 1.5.0

**Environment**

Alpine docker image on GKE

**Go Environment**
latest `golang:1.16-alpine`

**Expected behavior**

GCP logging request grouping behavior to show parent logs with the request method, url and parameters should work correctly as it does when logging to the GCP logging API directly without `logging.RedirectAsJSON(os.Stdout)`

<img width="224" alt="Screen Shot 2022-07-21 at 13 55 56" src="https://user-images.githubusercontent.com/879170/180313810-a0b1e62f-2401-4219-aca3-fcefd09e4d00.png">

**Actual behavior**

I am seeing some of the `httpRequest` params get logged in Stdout as `snake_case` instead of `camelCase` such as `request_method` instead of `requestMethod`. I believe this is causing the request grouping and parent/child request structure to not work in GCP logging UI because the [LogEntry JSON](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest) appears to expect camelCase values. 

<img width="370" alt="Screen Shot 2022-07-21 at 13 56 12" src="https://user-images.githubusercontent.com/879170/180313973-743c9e66-e33d-4ff5-8824-5407064101e2.png">

json Payload from console
```
{"message":"http log","severity":"INFO","httpRequest":{"request_method":"GET","request_url":"/health","status":200,"response_size":3,"user_agent":"GoogleHC/1.0","latency":{"nanos":96080},"protocol":"HTTP/1.1"},"timestamp":"seconds:1658435554 nanos:230493584","logging.googleapis.com/trace":"ommitted"}
```

A fix would be to ensure that these fields are correctly camel cased, or to have the GKE/logger ingestion correctly handle both snakecase and camelcase values for these fields to ensure log grouping is handled correctly. 

**Additional context**

Only related to using the new `logging.RedirectAsJSON(os.Stdout)` in v1..5.0
The [tests](https://github.com/googleapis/google-cloud-go/pull/6130/files#diff-fefad6d168f828ae5e5df7b8b3b5de4c6aa4fe08892980298153839ab56de4a8R1285) in the PR to enable this feature seem to suggest that the `snake_case` behavior is the expected/intended behavior but I think that is causing other problems with log grouping.

nwadams:
[...]

PR #7083, opened 2022-11-18
fix(logging): fix stdout log http request format

daniel-sanche:
We had a bug when using standard out logs with http request objects, where the http objects were being output using snake_case names instead of mixedCase. This caused the GCP Logging parser to not detect the objects as http_request types, and dump the data in the main JSON payload instead.

After this fix, these logs should appropriately parse http request objects, and put the data in the httpRequest metadata field as expected.

Related environment test changes: https://github.com/googleapis/env-tests-logging/pull/95

Fixes: https://github.com/googleapis/google-cloud-go/issues/6391


nwadams:
@daniel-sanche thanks for fixing this bug.

Issue #7087, opened 2022-11-21
httpreplay: enable to specify proxy listen address


nktks:
**Is your feature request related to a problem? Please describe.**
In my macOS Monterey(12.6.1), if execute like below code, os display popup like this(sorry popup lang is japanese).
```
package main

import (
	"encoding/json"
	"log"
	"time"

	"cloud.google.com/go/httpreplay"
)

func main() {
	now := time.Now()
	b, err := json.Marshal(now)
	if err != nil {
		log.Fatal(err)
	}
	_, err = httpreplay.NewRecorder("test", b)
	if err != nil {
		log.Fatal(err)
	}
	time.Sleep(10 * time.Second)
}
```
```
go run main.go
```
<img width="247" alt="image" src="https://user-images.githubusercontent.com/7553415/203012800-ab0ff312-bbb0-4550-9e98-a8ba3e609167.png">

This is because the httpreplay proxy listens on anycast IP 0.0.0.0 at [here](https://github.com/googleapis/google-cloud-go/blob/v0.107.0/httpreplay/internal/proxy/record.go#L134)
Listening to anycast IP is not a problem itself, but it's a little annoying to get a popup every time we run a test on macOS.

**Describe the solution you'd like**
When we want to use proxy only from locahost, enabling to use 127.0.0.1 is little useful.
How do you think about this?

**Describe alternatives you've considered**
I think fixing to 127.0.0.1 may not work if there is a case to access proxy from other host

nktks:
Thank you!

PR #7088, opened 2022-11-21
chore(httpreplay): specify localhost for proxy listener

quartzmo:
closes: #7087

@codyoss Is `httpreplay` ever used remotely? Will this change cause issues for anyone?

(Note that the package docs for `httpreplay` state: "This package is EXPERIMENTAL and is subject to change or removal without notice.")

### DIFF
--- a/httpreplay/internal/proxy/record.go
+++ b/httpreplay/internal/proxy/record.go
@@ -131,7 +131,7 @@ func newProxy(filename string) (*Proxy, error) {
 }
 
 func (p *Proxy) start(port int) error {
-	l, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+	l, err := net.Listen("tcp", fmt.Sprintf("127.0.0.1:%d", port))
 	if err != nil {
 		return err
 	}


codyoss:
@quartzmo I am not sure. We just use this package as a test utility. The default of listening on all unicast IPs seems fine imo -- but not an expert in this space. I think the user report can also be solved by playing with firewall perms on their machine perhaps as well if my translation was accurate. If you are feeling pretty confident that this will not break anything I am fine moving forward, or else I would say either just leave it. Thoughts?

quartzmo:
>  If you are feeling pretty confident that this will not break anything I am fine moving forward, or else I would say either just leave it.

It doesn't break anything when used locally (meaning, when the `httpreplay` client and listener are running on the same host), which is my impression of how the tool is supposed to be used. I would like to go forward with this change for the benefit (at least) of the #7087 poster. On the slight chance that breakage is reported for a use case we are not aware of, then I think we can change it back again. As stated: "EXPERIMENTAL and is subject to change or removal without notice."

codyoss (approved):

PR #7089, opened 2022-11-21
feat(iam): start generating and refresh some libraries

codyoss:
- Refresh the admin/apiv1 library. Keeping it out of auto generation due to conflicts in handcrafted files living alongside the client.
- Start generating a gapic for apiv1
- Start generating stubs dirs for both


internal/aliasfix/mappings.go
```diff
@@ -750,6 +750,14 @@ var GenprotoPkgMigration map[string]Pkg = map[string]Pkg{
 		ImportPath: "cloud.google.com/go/firestore/apiv1/firestorepb",
 		Status:     StatusInProgress,
 	},
+	"google.golang.org/genproto/googleapis/iam/v1": {
+		ImportPath: "cloud.google.com/go/iam/apiv1/iampb",
+		Status:     StatusInProgress,
```
quartzmo, line 755: Should this be `StatusMigrated`, since `iampb` is included in this PR?

internal/aliasfix/mappings.go
```diff
@@ -750,6 +750,14 @@ var GenprotoPkgMigration map[string]Pkg = map[string]Pkg{
...
+	},
+	"google.golang.org/genproto/googleapis/iam/admin/v1": {
+		ImportPath: "cloud.google.com/go/iam/admin/apiv1/adminpb",
+		Status:     StatusInProgress,
```
quartzmo, line 759: Should this be `StatusMigrated`, since `adminpb` is included in this PR?

quartzmo (approved):

internal/aliasfix/mappings.go
codyoss: nope, we still need to alias over to the new stubs since the current stubs are under use.

internal/aliasfix/mappings.go
codyoss: nope, we still need to alias over to the new stubs since the current stubs are under use.

codyoss:
Breaking changes here are unfortunate. I think because of this we will not be updating the admin client for now. It appears that client was generated with a version of the generator that did not have proper iterators yet. Eventually we should do a breaking changes for IAM and get these updated as current code is not maintainable/updatable with current tooling.

cc @noahdietz

noahdietz:
> Eventually we should do a breaking changes for IAM and get these updated as current code is not maintainable/updatable with current tooling.

Ack. Makes sense. That's a bummer ☹️

PR #7091, opened 2022-11-23
feat(spanner): support for enum columns as keys, index and integration tests

harshachinta:
This PR has the following changes,
1. Adds support for Enum columns in primary key and indexes.
2. Integration tests for Parameter Binding.
3. Integration tests for Proto fields and Enum columns as Primary keys.
4. Integration tests for Proto fields and Enum columns as Indexes.

### DIFF
--- a/spanner/integration_test.go
+++ b/spanner/integration_test.go
@@ -2326,7 +2326,7 @@ CREATE TABLE Singers (
  SingerInfo spanner.examples.music.SingerInfo,
  SingerGenre spanner.examples.music.Genre,
  SingerNationality STRING(1024) AS (SingerInfo.nationality) STORED,
-) PRIMARY KEY (SingerGenre);
+) PRIMARY KEY (SingerNationality, SingerGenre);
 
 CREATE INDEX SingerByNationalityAndGenre ON Singers(SingerNationality, SingerGenre) STORING (SingerId, FirstName, LastName);
 */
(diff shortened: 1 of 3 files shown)


rahul2393 (approved):

gauravpurohit06 (approved):

gauravpurohit06 (approved):

spanner/integration_test.go
```diff
@@ -2317,6 +2317,165 @@ func TestIntegration_BasicTypes_ProtoColumns_Errors(t *testing.T) {
...
+		},
+		// Filter rows using Proto Message Field Parameterized query
+		{sql: "SELECT SingerId, FirstName, LastName, SingerInfo, SingerGenre FROM Singers WHERE SingerInfo.Nationality=@country",
+			params: map[string]interface{}{
+				"country": "Country3",
+			},
+			want: [][]interface{}{{int64(3), "Singer3", "Singer3"}},
+		},
+	} {
+		ro := client.ReadOnlyTransaction()
```
olavloite, line 2434: This (probably) causes a session leak, as this read-only transaction is never closed.

spanner/integration_test.go
```diff
@@ -2317,6 +2317,165 @@ func TestIntegration_BasicTypes_ProtoColumns_Errors(t *testing.T) {
...
+		if err != nil {
+			t.Errorf("%d: Parameterized query returns error %v, want nil", i, err)
+		}
+		if !testEqual(got, test.want) {
+			t.Errorf("%d: got unexpected result from Parameterized query: %v, want %v", i, got, test.want)
+		}
+	}
+
+	// Read all rows based on Proto Enum Primary key column values
+	ro := client.ReadOnlyTransaction()
```
olavloite, line 2450: Same here, not closing a read-only transaction will cause a session leak.

olavloite (approved):

Shua1 (commented):
1. Please split the parameter binding part.
2. It seems more accurate, in the descriptor of the PR, to say "support for enum columns..." instead of "support for proto columns".

harshachinta:
> 1. Please split the parameter binding part.
> 2. It seems more accurate, in the descriptor of the PR, to say "support for enum columns..." instead of "support for proto columns".

1. Updated the PR description to make the changes more clearer. As the test for parameter binding is small, raising a separate PR for that would not be useful.
2. Updated the title of the PR.

Shua1 (commented):
The PR description says: "support for enum columns as keys, index, parameters".

But we are not adding support for parameters. Right? Please update.

harshachinta:
> The PR description says: "support for enum columns as keys, index, parameters".
> 
> But we are not adding support for parameters. Right? Please update.

Updated the description. Thanks for the review comment, @Shua1

spanner/integration_test.go
Shua1: Are these comments addressed?

harshachinta: Yes @Shua1. The changes have already been committed.

Issue #7103, opened 2022-11-29
bigquery/storage/managedwriter: multiplexed writes


shollyman:
This issue tracks the PRs related to supporting multiplex connections in managedwriter.

shollyman:
At this point, experimental multiplexing is available at head, but not baked into a release version.  Next release is 1.51.0, but it is likely release will be deferred until week of April 17.

shollyman:
While there continues to be smaller features related to multiplexing, going forward we'll track those individually rather than via this umbrella issue.

PR #7104, opened 2022-11-29
chore(bigquery/storage/managedwriter): internal refactor (flow controller, ids)

shollyman:
This PR adds a new internal mechanism to simplify duplicating flow controllers, and does some preliminary work to wire in a UUID-based ID for managed stream instances.  Neither is used elsewhere at this time, but lay the foundation for further refactoring.


bigquery/storage/managedwriter/flow_controller.go
```diff
@@ -50,6 +50,17 @@ func newFlowController(maxInserts, maxInsertBytes int) *flowController {
 	return fc
 }
 
+// copyFlowController is for creating a new flow controller based on
+// settings from another.  It does not copy flow state.
+func copyFlowController(in *flowController) *flowController {
```
alvarowolfx, line 55: should this function go to the test file as is only being used there or there are plans to use it later ?

alvarowolfx (approved):

bigquery/storage/managedwriter/flow_controller.go
shollyman: This is in prep of future refactors, it will be used for replicating the flow controller as we spin up new connections.

PR #7109, opened 2022-11-30
feat(pubsublite): set finalizer for PublisherClient

tmdiep:
Mitigates `PublisherClient.Stop` not being called when the publisher is no longer used. Users must still call `Stop` to promptly shut down the publisher, as finalizers run after an arbitrary amount of time.


pubsublite/pscompat/publisher.go
```diff
@@ -99,7 +106,15 @@ func NewPublisherClientWithSettings(ctx context.Context, topic string, settings
...
 		return nil, err
 	}
-	return &PublisherClient{settings: settings, wirePub: wirePub}, nil
+	publisher := &PublisherClient{settings: settings, wirePub: wirePub}
+
+	// Mitigation for Stop not being called when the publisher client is no longer
+	// used. Users must still call Stop to promptly shut down the publisher, as
+	// finalizers run after an arbitrary amount of time.
+	runtime.SetFinalizer(publisher, func(p *PublisherClient) {
+		p.wirePub.Stop()
```
codyoss, line 115: I am not sure I would do it now, but it might be good to drop a TODO here about adding a log warning eventually. In the not too distant future our libraries will have better access to a logger that can be used and configured by users. It would eventually be nice to be able to nudge users in the write direction and call Stop

dpcollins-google: +1

dpcollins-google, line 115: I would do "go p.wirePub.Stop()" to get it out of the garbage collector thread

dpcollins-google (approved):

pubsublite/pscompat/integration_test.go
```diff
@@ -590,6 +591,32 @@ func TestIntegration_PublishSubscribeSinglePartition(t *testing.T) {
...
+		if _, err := result.Get(ctx); err != nil {
+			t.Errorf("Publish() got err: %v", err)
+		}
+
+		result = publisher.Publish(ctx, &pubsub.Message{Data: []byte("finalizer2")})
+		// The finalizer runs during the next GC. Publish should still succeed
+		// because Stop flushes outstanding messages and waits for publish responses
+		// before closing connections.
+		runtime.GC()
+		if _, err := result.Get(ctx); err != nil {
```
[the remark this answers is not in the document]
tmdiep: Resolved offline: this is not necessary, even if stop is called in a goroutine, as Get is blocking.

pubsublite/pscompat/publisher.go
tmdiep: Glad to hear that a logger will be available soon!
Added a TODO.

tmdiep: Done

PR #7146, opened 2022-12-07
chore(errorreporting): Refactor error in error reporting

k-yomo:
When we don't register error handling function, error will be logged with `log.Println`.
However, since we don't add any context, it's unclear where the error is logged and what is the error for.
This PR fixes it by updating error message.
```
"2022/11/23 15:46:14 rpc error: code = PermissionDenied desc = User not authorized."
```


codyoss:
I think we will keep this as is for now. Consider opening an issue for discussion on how something like this could be valuable to you/others. Thanks.

PR #7149, opened 2022-12-08
feat(spanner): inline begin transaction for ReadWriteTransactions

rahul2393:
## Description

The Spanner client library will normally prepare a fraction of the sessions in the session pool with a read/write transaction. These sessions are used when an application requests a read/write transaction. This eliminates the need for a BeginTransaction RPC to be executed as part of a transaction, as the call has already been executed in advance by the session pool.

This can however be less efficient in some specific cases:
* When the number of read/write transactions that are needed is higher than the number of transactions the session pool is able to prepare in advance.
* When the fraction of read/write transactions is higher than what has been configured by the user (the default is 0.2).

Cloud Spanner also supports starting a new transaction as part of a query or DML statement. This eliminates the need for a separate round-trip to the server for starting a transaction. This PR adds the possibility to instruct the Spanner client to start read/write transactions by including this option in the first query/update statement of a read/write transaction, instead of preparing the read/write transactions in the session pool. This can improve the overall read/write transaction execution performance.

Including the `BeginTransaction` option with the first query or update statement of a transaction does require the client to block any other statement on the same transaction until the first statement has finished. This means that this strategy could perform less well in case of a read/write transaction that includes multiple parallel reads and/or writes.

With this PR we are going to make Inline begin transaction as default for ReadWriteTransactions


spanner/transaction.go
```diff
@@ -1056,6 +1076,10 @@ func (t *ReadWriteTransaction) batchUpdateWithOptions(ctx context.Context, stmts
 		return nil, ToSpannerError(err)
 	}
 
+	if t.tx == nil && resp.ResultSets != nil && len(resp.ResultSets) > 0 && resp.ResultSets[0].Metadata != nil {
```
olavloite, line 1079: Shouldn't this also include a `&& resp.ResultSets[0].Metadata.Transaction.GetId() != nil` (or similar)? I think this will clear the current transaction ID if you execute two batch DML operations in one transaction.

spanner/transaction.go
```diff
@@ -1072,20 +1096,72 @@ func (t *ReadWriteTransaction) batchUpdateWithOptions(ctx context.Context, stmts
...
+		return
 	}
 	t.mu.Lock()
 	defer t.mu.Unlock()
-	switch t.state {
-	case txClosed:
-		return nil, nil, errTxClosed()
-	case txActive:
-		return t.sh, ts, nil
+	if t.state == txInit {
```
olavloite, line 1159: When is this function called with `t.state != txInit`? And is it then reasonable that it is effectively a no-op?

spanner/transaction.go
```diff
@@ -1072,20 +1096,72 @@ func (t *ReadWriteTransaction) batchUpdateWithOptions(ctx context.Context, stmts
...
+			return sh, ts, nil
+		}
+		state := t.state
+		t.mu.Unlock()
+		return nil, nil, errUnexpectedTxState(state)
+	}
+}
+
+func (t *ReadWriteTransaction) setTransactionID(tx transactionID) {
+	if tx == nil {
```
olavloite, line 1154: When would this happen? And is it reasonable that it is a no-op?

spanner/transaction_test.go
```diff
@@ -268,7 +268,6 @@ func TestReadWriteTransaction_ErrorReturned(t *testing.T) {
 	requests := drainRequestsFromServer(server.TestSpanner)
```
olavloite, line 268: We should add test cases for known corner cases with InlineBeginTransaction, so at least for the following cases:
1. Verify that everything works as expected if a read/write transaction executes two queries in parallel as the first statement. Only one should include a `Begin`, and the other should block until the first has returned a transaction.
2. What happens if the first statement in a transaction returns an error?
3. What happens if the first statement in a transaction is a BatchDML, and the first N statements succeed, but statement N+1 fails?
4. What happens if the first statement in a transaction is a Read (not Query)?
5. What happens if the first statement in a transaction is a query or read operation that initially succeeds, but then later fails with an `Unavailable` error while consuming the stream of `PartialResultSet`s? The request will be retried, but it should not include a `BeginTransaction` option, as a transaction was already returned by the initial request.

olavloite (changes requested):

spanner/client_test.go
```diff
@@ -1420,6 +1421,66 @@ func TestClient_ReadWriteTransactionCommitAlreadyExists(t *testing.T) {
...
+		wg.Add(2)
+		go query(&firstTransactionID)
+		go query(&secondTransactionID)
+		wg.Wait()
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if firstTransactionID == nil || secondTransactionID == nil || string(firstTransactionID) != string(secondTransactionID) {
```
rahul2393, line 1471: >Verify that everything works as expected if a read/write transaction executes two queries in parallel as the first statement. Only one should include a Begin, and the other should block until the first has returned a transaction.

@olavloite is it sufficient to test this case?

spanner/stats.go
```diff
@@ -38,9 +38,7 @@ var (
 	tagCommonKeys    = []tag.Key{tagKeyClientID, tagKeyDatabase, tagKeyInstance, tagKeyLibVersion}
 
 	tagNumInUseSessions = tag.Tag{Key: tagKeyType, Value: "num_in_use_sessions"}
-	tagNumBeingPrepared = tag.Tag{Key: tagKeyType, Value: "num_sessions_being_prepared"}
-	tagNumReadSessions  = tag.Tag{Key: tagKeyType, Value: "num_read_sessions"}
-	tagNumWriteSessions = tag.Tag{Key: tagKeyType, Value: "num_write_prepared_sessions"}
+	tagNumSessions      = tag.Tag{Key: tagKeyType, Value: "num_sessions"}
```
hengfengli, line 41: Would changing metrics be a breaking change? Should we give a deprecation notice?

spanner/transaction.go
```diff
@@ -980,9 +996,13 @@ func (t *ReadWriteTransaction) update(ctx context.Context, stmt Statement, opts
 	if err != nil {
 		return 0, ToSpannerError(err)
 	}
+	if resultSet.Metadata != nil && resultSet.Metadata.Transaction != nil {
```
hengfengli, line 999: Do we also need to check if `t.setTransactionID` is nil here? I saw we checked this in another place.

spanner/client_test.go
olavloite: I would suggest that we also add a verification to the requests that we are seeing on the mock server, so verify that the first `ExecuteSqlRequest` on the server includes the `BeginTransaction` option, and that the second one includes a `TransactionOptions.Id`.

hengfengli (commented):
This is a big change. Generally, it looks good to me. My major concern is that do we understand what is the performance impact with this change? Do we have any plan to run benchmarks to compare with or without this change?

spanner/transaction.go
rahul2393: Updated to call this only when TransactionSelector::Begin is set. we are making sure that only one req will be executed with TransactionSelector::Begin in a read/write transaction.

spanner/client_test.go
rahul2393: Done

spanner/transaction.go
rahul2393: Updated to call this only when TransactionSelector::Begin is set. we are making sure that only one req will be executed with TransactionSelector::Begin in a read/write transaction.

spanner/stats.go
rahul2393: Done

spanner/transaction.go
```diff
@@ -1072,20 +1096,72 @@ func (t *ReadWriteTransaction) batchUpdateWithOptions(ctx context.Context, stmts
 
 // acquire implements txReadEnv.acquire.
 func (t *ReadWriteTransaction) acquire(ctx context.Context) (*sessionHandle, *sppb.TransactionSelector, error) {
-	ts := &sppb.TransactionSelector{
-		Selector: &sppb.TransactionSelector_Id{
-			Id: t.tx,
-		},
+	for {
+		t.mu.Lock()
+		switch t.state {
```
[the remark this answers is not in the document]
rahul2393: Added comment

spanner/client_test.go
```diff
@@ -1435,6 +1421,225 @@ func TestClient_ReadWriteTransactionCommitAlreadyExists(t *testing.T) {
...
+		t.Fatal("expected streaming query to use TransactionSelector::Begin")
+	}
+	if _, ok := requests[3].(*sppb.ExecuteSqlRequest).Transaction.GetSelector().(*sppb.TransactionSelector_Id); !ok {
+		t.Fatal("expected streaming query to use transactionID from explicit begin transaction")
+	}
+}
+
+// Given a transaction, When the StreamingRead fails halfway and stream is restarted with a resume token,
+// Then the transaction ID should be used from the first PartialResultSet.
+func TestClient_ReadWriteTransaction_FirstStatementAsReadFailsHalfway(t *testing.T) {
```
rahul2393, line 1553: >What happens if the first statement in a transaction is a query or read operation that initially succeeds, but then later fails with an Unavailable error while consuming the stream of PartialResultSets? The request will be retried, but it should not include a BeginTransaction option, as a transaction was already returned by the initial request.

@olavloite Is it sufficient for this test case.

spanner/client_test.go
```diff
@@ -1435,6 +1421,225 @@ func TestClient_ReadWriteTransactionCommitAlreadyExists(t *testing.T) {
...
+		&sppb.ExecuteSqlRequest{},
+		&sppb.CommitRequest{}}, requests); err != nil {
+		t.Fatal(err)
+	}
+	// Although the batch DML returned an error, that error was for the second statement. That
+	// means that the transaction was started by the first statement.
+	if _, ok := requests[1].(*sppb.ExecuteBatchDmlRequest).Transaction.GetSelector().(*sppb.TransactionSelector_Begin); !ok {
+		t.Fatal("expected streaming read to use TransactionSelector::Begin")
+	}
+	if _, ok := requests[2].(*sppb.ExecuteSqlRequest).Transaction.GetSelector().(*sppb.TransactionSelector_Id); !ok {
```
rahul2393, line 1638: >What happens if the first statement in a transaction is a BatchDML, and the first N statements succeed, but statement N+1 fails?

@olavloite is it sufficient for this case.

spanner/client_test.go
```diff
@@ -1435,6 +1421,225 @@ func TestClient_ReadWriteTransactionCommitAlreadyExists(t *testing.T) {
...
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	requests := drainRequestsFromServer(server.TestSpanner)
+	if err := compareRequests([]interface{}{
+		&sppb.BatchCreateSessionsRequest{},
+		&sppb.ExecuteSqlRequest{},
+		&sppb.BeginTransactionRequest{},
```
rahul2393, line 1538: >What happens if the first statement in a transaction returns an error?

@olavloite so is error from first statement is retryable(Aborted/Session not found) we will do explicit begin transaction and retry.

spanner/client_test.go
```diff
@@ -1435,6 +1421,225 @@ func TestClient_ReadWriteTransactionCommitAlreadyExists(t *testing.T) {
...
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	requests := drainRequestsFromServer(server.TestSpanner)
+	if err := compareRequests([]interface{}{
+		&sppb.BatchCreateSessionsRequest{},
+		&sppb.ReadRequest{},
+		&sppb.ReadRequest{},
```
rahul2393, line 1585: >What happens if the first statement in a transaction is a Read (not Query)?

@olavloite flow will be same if error is unavailable we will use the same selector as it was in initial request, once we get partial result having transactionID we will use it on other operations/statements.

hengfengli (approved):
LGTM. Please resolve Knut's comments before merging.

rahul2393:
Screenshot 2022-12-13 at 9 11 01 PM
Benchmark results from client_benchmarks_test.go

spanner/spannertest/inmem.go
```diff
@@ -533,11 +548,15 @@ func (s *server) ExecuteSql(ctx context.Context, req *spannerpb.ExecuteSqlReques
 	if err != nil {
 		return nil, err
 	}
-	return &spannerpb.ResultSet{
+	rs := &spannerpb.ResultSet{
```
codyoss, line 551: nit: below this line an unkeyed struct is being used which can lead to bugs. Please consisder added a stuct field key for the value.

spanner/client_benchmarks_test.go
```diff
@@ -177,8 +177,8 @@ func benchmarkClientBurstRead(b *testing.B, incStep uint64) {
 	for n := 0; n < b.N; n++ {
 		server, client, teardown := createBenchmarkServer(incStep)
 		sp := client.idleSessions
-		if uint64(sp.idleList.Len()+sp.idleWriteList.Len()) != sp.MinOpened {
-			b.Fatalf("session count mismatch\nGot: %d\nWant: %d", sp.idleList.Len()+sp.idleWriteList.Len(), sp.MinOpened)
+		if uint64(sp.idleList.Len()) != sp.MinOpened {
```
codyoss, line 180: nit: below in this test a goroutine is launched with the arg b. Calling fatal in a launched goroutine from a test can cause the whole suite to fail. Should be converted to an error or errors collected and considered fatal back in the main goroutine. I see this in several tests.

spanner/transaction.go
```diff
@@ -1071,21 +1127,94 @@ func (t *ReadWriteTransaction) batchUpdateWithOptions(ctx context.Context, stmts
...
+				case <-ctx.Done():
+					// The waiting for initialization is timeout, return error
+					// directly.
+					return nil, nil, errTxInitTimeout()
+				}
+			}
+			// If first statement with TransactionSelector::begin succeeded, t.state should have been changed to
+			// txActive, so we can just continue here.
+			continue
+		case txActive:
```
codyoss, line 1174: Should there be a default for this that returns an error for an invalid state?

spanner/transaction.go
```diff
@@ -1071,21 +1127,94 @@ func (t *ReadWriteTransaction) batchUpdateWithOptions(ctx context.Context, stmts
...
+			ts := &sppb.TransactionSelector{
+				Selector: &sppb.TransactionSelector_Id{
+					Id: t.tx,
+				},
+			}
+			t.mu.Unlock()
+			return sh, ts, nil
+		}
+		state := t.state
+		t.mu.Unlock()
```
codyoss, line 1185: Is this unlock needed?

spanner/transaction.go
```diff
@@ -1071,21 +1127,94 @@ func (t *ReadWriteTransaction) batchUpdateWithOptions(ctx context.Context, stmts
...
 	}
-	return nil, nil, errUnexpectedTxState(t.state)
+}
+
+func (t *ReadWriteTransaction) setTransactionID(tx transactionID) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	t.tx = tx
+	t.state = txActive
+	close(t.txReadyOrClosed)
```
codyoss, line 1216: This bit seems odd to me. Why is the channel being closed only to be reinitialized in the next line?

spanner/transaction.go
```diff
@@ -1123,14 +1252,64 @@ func (t *ReadWriteTransaction) begin(ctx context.Context) error {
 		t.state = txActive
 		return nil
 	}
-	tx, err := beginTransaction(contextWithOutgoingMetadata(ctx, t.sh.getMetadata()), t.sh.getID(), t.sh.getClient())
+	var (
+		locked bool
```
codyoss, line 1256: Why is this variable needed when you have a mutex?

codyoss (commented):
Just some general feedback and questions. I mostly just reviewed Go things as I don't have the full context of this code base. I would also recommend running your tests with the race detecor to make sure all of this new locking code does not have any races in it.

spanner/client_test.go
```diff
@@ -1435,6 +1421,225 @@ func TestClient_ReadWriteTransactionCommitAlreadyExists(t *testing.T) {
...
+	}
+	if _, ok := requests[1].(*sppb.ReadRequest).Transaction.GetSelector().(*sppb.TransactionSelector_Begin); !ok {
+		t.Fatal("expected streaming read to use TransactionSelector::Begin")
+	}
+	if _, ok := requests[2].(*sppb.ReadRequest).Transaction.GetSelector().(*sppb.TransactionSelector_Id); !ok {
+		t.Fatal("expected streaming read to use transactionID from previous success request")
+	}
+}
+
+func TestClient_ReadWriteTransaction_BatchDmlWithErrorOnSecondStatement(t *testing.T) {
```
olavloite, line 1597: We should also verify the correct behavior if the __first__ statement is invalid and the Batch DML request is the first request in the transaction. In that case the transaction should also be restarted with an explicit BeginTransaction RPC. Otherwise, the Batch DML request won't be part of the transaction.

spanner/client_test.go
```diff
@@ -2143,12 +2344,10 @@ func TestFailedUpdate_ShouldRollback(t *testing.T) {
 	if got, want := status.Convert(err).Code(), codes.InvalidArgument; got != want {
 		t.Fatalf("Error mismatch\nGot: %v\nWant: %v", got, want)
 	}
-	// The failed update should trigger a rollback.
+	// No rollback request will be initiated because the client does not receive any transactionId.
```
olavloite, line 2347: This is not the correct behavior (although in this specific case it does not matter). If the first statement in a transaction fails, then we should restart the transaction and use an explicit BeginTransaction RPC. Otherwise, if this transaction would contain more statements, the following statement in the transaction would include the BeginTransaction option. That again would mean that the first statement seems to have been executed on the transaction, while it actually hasn't.

See also the other comments.

olavloite:
@rahul2393 It seems that we are not adding any tests for `StmtBasedTransaction`. Will those transactions also use inline-begin-transaction? If so, we should add tests for those as well to verify the correct behavior.

spanner/client_test.go
```diff
@@ -1435,6 +1421,225 @@ func TestClient_ReadWriteTransactionCommitAlreadyExists(t *testing.T) {
...
+		&sppb.BatchCreateSessionsRequest{},
+		&sppb.ReadRequest{},
+		&sppb.ReadRequest{},
+		&sppb.CommitRequest{}}, requests); err != nil {
+		t.Fatal(err)
+	}
+	if _, ok := requests[1].(*sppb.ReadRequest).Transaction.GetSelector().(*sppb.TransactionSelector_Begin); !ok {
+		t.Fatal("expected streaming read to use TransactionSelector::Begin")
+	}
+	if _, ok := requests[2].(*sppb.ReadRequest).Transaction.GetSelector().(*sppb.TransactionSelector_Id); !ok {
```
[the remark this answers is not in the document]
harshachinta: @olavloite Done. Please check if this is enough.
https://github.com/googleapis/google-cloud-go/pull/7149/commits/136af306c01e2fe2ebd072bb367b7cc7717004a2#diff-f1597c036ee4eafc4b130e2a22ba1a80bee46d24a1cf228dbc4ac4f114ae9fafR1595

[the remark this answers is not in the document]
olavloite: Thanks

spanner/spannertest/inmem.go
harshachinta: Done

spanner/transaction.go
harshachinta: The invalid transaction state case is handled outside the switch case.

spanner/transaction.go
harshachinta: Yes. It is need to unlock the transaction when the transaction state is not valid (as none of the switch cases will be executed).

spanner/client_benchmarks_test.go
rahul2393: Thanks @codyoss for suggestion, will be doing benchmark changes in another PR, since we are not changing benchmark tests itself as part of this feature wanted to have improvements separate.

spanner/transaction.go
rahul2393: Channel is closed to un-block other parallel routines waiting for transaction state to be Closed/Active, we are re-initialising is to new channel for the case of Closed so that when doing retry we don't access the closed channel from previous run.

spanner/client_test.go
rahul2393: Ack, fixed

spanner/client_test.go
```diff
@@ -1435,6 +1421,293 @@ func TestClient_ReadWriteTransactionCommitAlreadyExists(t *testing.T) {
...
+		t.Fatal("expected streaming read to use transactionID from previous success request")
+	}
+	if requests[2].(*sppb.ReadRequest).ResumeToken == nil {
+		t.Fatal("expected streaming read to include resume token")
+	}
+}
+
+func TestClient_ReadWriteTransaction_BatchDmlWithErrorOnFirstStatement(t *testing.T) {
```
rahul2393, line 1600: >That is good (although an explicit BeginTransaction RPC is actually not needed in these cases). My question is actually if the error is not retryable?

>So you have the following (naive) transaction (pseudo code):

>_, err := tx.update("insert into foo (id, value) values (1, 'one')");
if err.Code == Code.AlreadyExists {
  tx.update("update foo set value='one' where id=1");
}
>In this case the initial insert statement will not return a transaction id if it fails, but it is important that the statement is part of the entire transaction that we execute, as the returned error code actually tells something about the state of the database. So in this case we also need to restart the transaction with an explicit BeginTransaction.

@olavloite added the test case for this case.
Thanks

spanner/client_test.go
rahul2393: Ack, added the test case.

rahul2393:
@olavloite We are not changing the StmtBasedTransaction with this change, because we expect customers to do error handling and retry // management, so if we do inline begin & retries it will be confusing to them.
Updated the PR description.
Thanks

rahul2393:
> Just some general feedback and questions. I mostly just reviewed Go things as I don't have the full context of this code base. I would also recommend running your tests with the race detecor to make sure all of this new locking code does not have any races in it.

We run tests using -race flag in Kokoro already https://github.com/googleapis/google-cloud-go/blob/main/internal/kokoro/continuous.sh#L87

olavloite (commented):
This looks mostly good to me. I still have a couple of small nits/questions. Please take a look at those, but I think that those will be my last comments on this PR.

spanner/read.go
```diff
@@ -126,6 +130,19 @@ func (r *RowIterator) Next() (*Row, error) {
...
 	for len(r.rows) == 0 && r.streamd.next() {
 		prs := r.streamd.get()
+		if r.setTransactionID != nil {
+			// this is when Read/Query is executed using ReadWriteTransaction
+			// and server returned the first stream response.
+			if prs.Metadata != nil && prs.Metadata.Transaction != nil {
+				r.setTransactionID(prs.Metadata.Transaction.GetId())
+			} else {
+				// retry with explicit BeginTransaction
+				r.setTransactionID(nil)
```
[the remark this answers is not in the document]
olavloite: Thanks!

olavloite (approved):
LGTM

PR #7158, opened 2022-12-12
feat(storage): json downloads

BrennaEpp:


storage/integration_test.go
```diff
@@ -253,9 +253,11 @@ func testConfigGRPC(ctx context.Context, t *testing.T, opts ...option.ClientOpti
 
 // initTransportClients initializes Storage clients for each supported transport.
 func initTransportClients(ctx context.Context, t *testing.T, opts ...option.ClientOption) map[string]*Client {
+	withJSON := append(opts, WithJSONReads())
 	return map[string]*Client{
-		"http": testConfig(ctx, t, opts...),
-		"grpc": testConfigGRPC(ctx, t, opts...),
+		"http":      testConfig(ctx, t, opts...),
+		"grpc":      testConfigGRPC(ctx, t, opts...),
+		"jsonReads": testConfig(ctx, t, withJSON...),
```
tritone, line 260: How much time does this add to the test run? Wondering if it's worth limiting this to reader tests rather than running all transports again.

storage/integration_test.go
```diff
@@ -1334,19 +1337,20 @@ func TestIntegration_Objects(t *testing.T) {
 			if got, want := rc.ContentType(), "text/plain"; got != want {
 				t.Errorf("ContentType (%q) = %q; want %q", obj, got, want)
 			}
-			if got, want := rc.CacheControl(), "public, max-age=60"; got != want {
-				t.Errorf("CacheControl (%q) = %q; want %q", obj, got, want)
-			}
+			// differs
```
tritone, line 1340: Thoughts on what's going on here? Should we block off time to pair/discuss?

storage/storage.go
```diff
@@ -1704,6 +1710,8 @@ type Conditions struct {
 	// GenerationNotMatch specifies that the object must not have the given
 	// generation for the operation to occur.
 	// If GenerationNotMatch is zero, it has no effect.
+	// This condition only works for object reads if the WithJSONReads client
```
tritone, line 1713: `[WithJSONReads]` to make a link, I think

storage/storage.go
```diff
@@ -1722,10 +1730,13 @@ type Conditions struct {
 	// MetagenerationNotMatch specifies that the object must not have the given
 	// metageneration for the operation to occur.
 	// If MetagenerationNotMatch is zero, it has no effect.
+	// This condition only works for object reads if the WithJSONReads client
+	// option is set.
 	MetagenerationNotMatch int64
 }
 
 func (c *Conditions) validate(method string) error {
+
```
tritone, line 1739: extra line

storage/reader_test.go
```diff
@@ -387,6 +393,8 @@ func TestContentEncodingGzipWithReader(t *testing.T) {
 			gz := gzip.NewWriter(w)
 			gz.Write(original)
 			gz.Close()
+		default:
+			fmt.Fprintf(w, "unrecognized URL %s", r.URL)
```
tritone, line 397: This should be a failure?

storage/reader.go
```diff
@@ -142,12 +142,17 @@ func uncompressedByServer(res *http.Response) bool {
 func parseCRC32c(res *http.Response) (uint32, bool) {
 	const prefix = "crc32c="
 	for _, spec := range res.Header["X-Goog-Hash"] {
-		if strings.HasPrefix(spec, prefix) {
-			c, err := decodeUint32(spec[len(prefix):])
-			if err == nil {
-				return c, true
+		values := strings.Split(spec, ",")
```
tritone, line 145: Can you speak about why this change is necessary? How does the header format differ between APIs?

storage/option.go
```diff
@@ -0,0 +1,75 @@
...
+
+// A storageClientOption is an option for a Google Storage client.
+type storageClientOption interface {
+	option.ClientOption
+	ApplyStorageOpt(*storageConfig)
+}
+
+// WithJSONReads is an option that may be passed to a Storage Client on creation.
+// It sets the client to use the JSON API for object reads.
+// Set this option to obtain new features such as the GenerationNotMatch and
```
tritone, line 49: These aren't really "new" features. You can just say that it is required to use GenerationNotMatch. (MetagenerationMatch shouldn't be relevant for downloads).

storage/option.go
```diff
@@ -0,0 +1,75 @@
...
+// MetagenerationNotMatch conditions.
+//
+// Note that when this option is set, reads will return a zero date for
+// [ReaderObjectAttrs].LastModified and may return a different value for
```
tritone, line 53: Can you do `[ReaderObjectAttrs.LastModified]` to link to the specific field?

storage/option.go
```diff
@@ -0,0 +1,75 @@
...
+// [ReaderObjectAttrs].CacheControl.
+func WithJSONReads() option.ClientOption {
+	return &withReadAPI{useJSON: true}
+}
+
+// WithXMLReads is an option that may be passed to a Storage Client on creation.
+// It sets the client to use the JSON API for object reads.
+//
+// This is the current default, but will be deprecated in the future.
```
tritone, line 62: I don't think we have to discuss deprecation now.

storage/option.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// A storageClientOption is an option for a Google Storage client.
+type storageClientOption interface {
+	option.ClientOption
+	ApplyStorageOpt(*storageConfig)
+}
+
+// WithJSONReads is an option that may be passed to a Storage Client on creation.
+// It sets the client to use the JSON API for object reads.
```
tritone, line 48: Note here that XML is used by default.

storage/http_client.go
```diff
@@ -777,6 +780,13 @@ func (c *httpStorageClient) NewRangeReader(ctx context.Context, params *newRange
...
 	defer func() { trace.EndSpan(ctx, err) }()
 
+	if c.config.useJSONforReads {
+		return c.newRangeReaderJSON(ctx, params, opts...)
+	}
+	return c.newRangeReaderXML(ctx, params, opts...)
+}
+
+func (c *httpStorageClient) newRangeReaderXML(ctx context.Context, params *newRangeReaderParams, opts ...storageOption) (r *Reader, err error) {
 	s := callSettings(c.settings, opts...)
```
tritone, line 790: Let's make the settings in `NewRangeReader` and just pass it through.

storage/http_client.go
```diff
@@ -975,6 +986,195 @@ func (c *httpStorageClient) NewRangeReader(ctx context.Context, params *newRange
...
+		if params.length < 0 && start < 0 {
+			call.Header().Set("Range", fmt.Sprintf("bytes=%d", start))
+		} else if params.length < 0 && start > 0 {
+			call.Header().Set("Range", fmt.Sprintf("bytes=%d-", start))
+		} else if params.length > 0 {
+			// The end character isn't affected by how many bytes we've seen.
+			call.Header().Set("Range", fmt.Sprintf("bytes=%d-%d", start, params.offset+params.length-1))
+		}
+		// We wait to assign conditions here because the generation number can change in between reopen() runs.
+		if err := applyConds("NewReader", params.gen, params.conds, call); err != nil {
```
tritone, line 1026: We should probably add a test to verify that preconditions including ifGenerationNotMatch are passed through as expected

storage/http_client.go
```diff
@@ -975,6 +986,195 @@ func (c *httpStorageClient) NewRangeReader(ctx context.Context, params *newRange
...
+				var e *googleapi.Error
+				if errors.As(err, &e) {
+					if e.Code == http.StatusNotFound {
+						return ErrObjectNotExist
+					}
+				}
+				return err
+			}
+
+			if res.StatusCode == http.StatusNotFound {
```
tritone, line 1047: Is this not duplicating what's above?

storage/http_client.go
```diff
@@ -975,6 +986,195 @@ func (c *httpStorageClient) NewRangeReader(ctx context.Context, params *newRange
...
+			// back the whole file regardless of the range count passed in as per:
+			//      https://cloud.google.com/storage/docs/transcoding#range,
+			// thus we have to manually move the body forward by seen bytes.
+			if decompressiveTranscoding(res) && seen > 0 {
+				_, _ = io.CopyN(ioutil.Discard, res.Body, seen)
+			}
+
+			// If a generation hasn't been specified, and this is the first response we get, let's record the
+			// generation. In future requests we'll use this generation as a precondition to avoid data races.
+			if params.gen < 0 && res.Header.Get("X-Goog-Generation") != "" {
```
tritone, line 1081: Do we need to get this from the header, or is it populated in the apiary response somewhere?

storage/http_client.go
```diff
@@ -975,6 +986,195 @@ func (c *httpStorageClient) NewRangeReader(ctx context.Context, params *newRange
...
+		Attrs:    attrs,
+		size:     size,
+		remain:   remain,
+		wantCRC:  crc,
+		checkCRC: checkCRC,
+		reader: &httpReader{
+			reopen: reopen,
+			body:   body,
+		},
+	}, nil
```
tritone, line 1175: I'm wondering if it's possible to deduplicate some of this code with what you have for XML. I don't have a great idea of how to do this off the top of my head, but worth thinking about maybe.

storage/reader.go
BrennaEpp: I added a comment to the function. 
The XML header looks something like: [crc32c=xxx md5=xxx] 
while the JSON header looks like:         [crc32c=xxx,md5=xxx]

So I added on a split on comma

storage/storage.go
BrennaEpp: I double checked this, it does not create a link in VSCode nor on pkg.go.dev docs. See attachment.

image

storage/option.go
BrennaEpp: Double-checked this one too, looks like the answer is no:
image

storage/option.go
BrennaEpp: Good point, I updated the comment.

storage/integration_test.go
BrennaEpp: Tests are limited to reader tests using the `skipJSONReads` context.

storage/integration_test.go
BrennaEpp: I separated the reader portion out of this test, and tested last modified and cache control separately.

storage/reader_test.go
BrennaEpp: As discussed offline, the mock just returns this and the test decides how to use it and whether it is a failure state - in this case, the test will fail upon getting such a response.

tritone (commented):
Generally looking good, a few more minor things

storage/retry_conformance_test.go
```diff
@@ -247,13 +257,22 @@ var methods = map[string][]retryFunc{
...
-			c.config.useJSONforReads = true
+			client, ok := c.tc.(*httpStorageClient)
+			if ok {
+				client.config.readAPIWasSet = true
+				client.config.useJSONforReads = true
+				defer func() {
+					client.config.readAPIWasSet = false
+					client.config.useJSONforReads = false
+				}()
+			}
```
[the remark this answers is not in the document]
BrennaEpp: hm maybe, did we have a plan for running these tests with GRPC too?

storage/reader_test.go
```diff
@@ -35,42 +35,40 @@ import (
...
-			continue
-		}
-		gotb, err := ioutil.ReadAll(r)
-		if err != nil {
-			t.Errorf("%d/%d: %v", test.offset, test.length, err)
-			continue
-		}
-		if got := string(gotb); got != test.want {
-			t.Errorf("%d/%d: got %q, want %q", test.offset, test.length, got, test.want)
+	multiReaderTest(ctx, t, func(t *testing.T, c *Client) {
```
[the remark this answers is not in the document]
BrennaEpp: Just that, and adding the JSON URL to the mock GCS service used in TestContentEncodingGzipWithReader

storage/integration_test.go
```diff
@@ -4053,25 +4098,96 @@ func TestIntegration_ReaderAttrs(t *testing.T) {
...
+			t.Fatalf("cache control; got: %s, want: %s", got, want)
+		}
 	})
 }
 
 // Test that context cancellation correctly stops a download before completion.
 func TestIntegration_ReaderCancel(t *testing.T) {
 	multiTransportTest(context.Background(), t, func(t *testing.T, ctx context.Context, bucket, _ string, client *Client) {
-		ctx, close := context.WithDeadline(ctx, time.Now().Add(time.Second*30))
+		ctx, close := context.WithDeadline(ctx, time.Now().Add(time.Minute))
```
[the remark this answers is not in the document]
BrennaEpp: It was failing locally, I assume due to a very slow connection at the time. Changed it back as it has no longer been an issue.

storage/http_client.go
BrennaEpp: As we discussed the other day, it indeed must be from the header as  apiary doesn't add it explicitly to the response anywhere.

storage/http_client.go
BrennaEpp: Gave it a go at splitting it up, PTAL

storage/http_client.go
BrennaEpp: Added an integration test for the conditions on reader

storage/grpc_client.go
```diff
@@ -861,13 +861,6 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
 		ctx = setUserProjectMetadata(ctx, s.userProject)
 	}
 
-	// A negative length means "read to the end of the object", but the
```
tritone, line 858: Is there a reason to make this gRPC changes in the same PR? it might make more sense to do them separately unless necessary.

storage/integration_test.go
```diff
@@ -4203,10 +4206,96 @@ func TestIntegration_ReaderCacheControl(t *testing.T) {
...
+		_, err := o.NewReader(ctx)
+		if !errors.Is(err, ErrObjectNotExist) {
+			t.Fatalf("expected ErrObjectNotExist, got %v", err)
+		}
+	})
+}
+
+// TestIntegration_JSONReaderConditions tests only JSON reads as some conditions
+// do not work with XML.
+func TestIntegration_JSONReaderConditions(t *testing.T) {
```
tritone, line 4222: Thanks for adding this!

storage/http_client.go
```diff
@@ -804,191 +803,28 @@ func (c *httpStorageClient) newRangeReaderXML(ctx context.Context, params *newRa
...
-		}, s.retry, s.idempotent, setRetryHeaderHTTP(nil))
-		if err != nil {
-			return nil, err
-		}
-		return res, nil
-	}
+	reopen := readerReopen(ctx, req.Header, params, s,
+		func() (*http.Response, error) { return c.hc.Do(req) },
+		func() error { return setConditionsHeaders(req.Header, params.conds) },
+		func(gen int64) { req.URL.RawQuery = fmt.Sprintf("generation=%d", params.gen) })
```
tritone, line 818: Looks like the function `gen int64` arg isn't actually needed/used, you can just remove it for clarity (and use `params.gen` directly for JSON as well).

tritone (commented):
Couple more small things, in general the refactor looks good and I'm comfortable going ahead and merging this

storage/grpc_client.go
BrennaEpp: Good point. Originally it was to fix part of the test for GRPC, but we are skipping it anyway. Separate PR opened.

tritone (approved):

PR #7159, opened 2022-12-12
fix(logging): fix kuberentes environment test

daniel-sanche:
A logging test is currently failing due to use of the deprecated auth plugin. This PR fixes it by installing and using the replacement plugin


losalex (approved):

internal/kokoro/environment.sh
```diff
@@ -86,6 +86,9 @@ if [[ $ENVIRONMENT == *"kubernetes"* ]]; then
   mkdir -p ~/.local/bin
   mv ./kubectl ~/.local/bin
   export PATH=$PATH:~/.local/bin/
+  # install auth plugin
+  apt-get install google-cloud-sdk-gke-gcloud-auth-plugin
```
codyoss, line 90: This is fine for now, but likely this should be baked into our base image if needed instead of reinstalling on every rerun. Would you file an issue for this?

internal/kokoro/environment.sh
```diff
@@ -86,6 +86,9 @@ if [[ $ENVIRONMENT == *"kubernetes"* ]]; then
...
+  export USE_GKE_GCLOUD_AUTH_PLUGIN=True
```
codyoss, line 91: What does this control -- will it affect other tests? If so would you mind unsetting it at the end of the script?

codyoss (approved):
Approved but please see comments

internal/kokoro/environment.sh
daniel-sanche: This ensures that kubectl uses the gcloud auth plugin, now that the previous method is deprecated

I could disable it at the end of the tests, but keep in mind that this `environment.sh` file is only used for the logging library-specific environment test entrypoint, so this shouldn't impact other tests

PR #7174, opened 2022-12-16
refactor(bigquery/storage/managedwriter): augment connection pool

shollyman:
This PR adds two references to live code:
* a pool reference on ManagedStream
* a reference to ManagedStream on pendingWrite

The ManagedStream->pool reference is to allow a writer to resolve where to find its associated connection, retries, lookups, etc.

The reference on the pendingWrite is primarily in service of retries, particularly when we need to re-enqueue and thus potentially re-resolve what connection is associated with the writer.  This PR also moves some of the retry processing code onto the connectionPool in service to that goal.  As before, this is new code that isn't yet referenced from existing functionality.

This PR also more substantially starts to carve out connection management in the pool, providing a basic connection resolver and eviction capabilities.  This initial implementation is primitive, and aligns with our current behavior (single unshared connection per writer).

The existing lockingAppend behavior on ManagedStream is added to the connection abstraction.  We do not move appendWithRetries as that should stay with the writer (we may want to retry using a different connection).

We also add some testing of the mapping behavior to ensure we're consistently updating the map for resolution and eviction.


bigquery/storage/managedwriter/connection.go
```diff
@@ -148,6 +298,68 @@ func (co *connection) getStream(arc *storagepb.BigQueryWrite_AppendRowsClient, f
 	return co.arc, co.pending, co.err
 }
 
+// lockingAppend handles a single append request on a connection.
+func (c *connection) lockingAppend(pw *pendingWrite) error {
```
alvarowolfx, line 302: are you planning to remove the `lockingAppend` method on both `connection` and `ManagedStream` on this PR or you are going to remove from `ManagedStream` already ?

bigquery/storage/managedwriter/connection.go
```diff
@@ -96,6 +160,69 @@ func (cp *connectionPool) openWithRetry(co *connection) (storagepb.BigQueryWrite
...
+func (cp *connectionPool) processRetry(pw *pendingWrite, srcConn *connection, appendResp *storagepb.AppendRowsResponse, initialErr error) {
+	err := initialErr
+	for {
+		pause, shouldRetry := pw.writer.statelessRetryer().Retry(err, pw.attemptCount)
+		if !shouldRetry {
+			// Should not attempt to re-append.
+			pw.markDone(appendResp, err, srcConn.fc)
+			return
+		}
+		time.Sleep(pause)
```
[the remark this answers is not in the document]
shollyman: Good point, was avoiding gax because it's not really a gax-based backoff but the fast exit it desirable.

bigquery/storage/managedwriter/connection.go
shollyman: When I reach the refactor point where I modify ManagedStream most of the functionality goes away, lockingAppend and other connection-oriented funcs will be removed.

shollyman:
Closing this, taking a slightly different approach now.

PR #7181, opened 2022-12-20
chore(internal/godocfx): go mod tidy

tbpg:
The docs build is failing because of a missing `go.sum` entry.


codyoss (approved):

codyoss:
Looks like upgrading the deps caused a test golden test to fail.

PR #7202, opened 2023-01-03
feat(internal/postprocessor): make OwlBot postprocessor

adrianajg:
OwlBot will be configured to copy generated files directly into their corresponding client directories. When triggered by the OwlBot automation robot, the postprocessor performs the following tasks:
- run `$ go mod tidy` in all repo directories
- run `$ gofmt -s -d -w -l .` in all repo directories
- regenerate snippets for all clients
- write manifest file with info about all gapic configurations
- update pull request title and description, amending commit titles to match Conventional Commit format
    - The amended PR title and description are written to a file in a well-known-location, which OwlBot uses to amend the open PR
    - The `.github.OwlBot.yaml` file must include an `api-name` field with the value `REPLACEME` in order to label all commit messages with the placeholder scope '[REPLACEME]'.


internal/postprocessor/main.go
```diff
@@ -0,0 +1,403 @@
...
+			modules = append(modules, filepath.Join(*clientRoot, dir))
+		}
+		log.Println("Postprocessor running on", modules)
+	} else {
+		log.Println("Postprocessor running on all modules.")
+	}
+
+	if *googleapisDir == "" {
+		log.Println("creating temp dir")
+		tmpDir, err := ioutil.TempDir("", "update-postprocessor")
```
codyoss, line 82: nit: use `os.MkdirTemp` instead

internal/postprocessor/main_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+var googleapisDir string
+
+func TestMain(m *testing.M) {
+	flag.StringVar(&googleapisDir, "googleapis-dir", "", "Enter local googleapisDir to avoid cloning")
+	flag.Parse()
+
+	if googleapisDir == "" {
+		log.Println("creating temp dir")
+		tmpDir, err := ioutil.TempDir("", "update-postprocessor")
```
codyoss, line 37: nit: use os.MkdirTemp instead

codyoss (changes requested):
Two small nits and the PR needs to be updated and resolve a merge conflict.

internal/postprocessor/main.go
adrianajg: Done.

internal/postprocessor/main_test.go
adrianajg: Done.

internal/postprocessor/execv/command.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		if strings.Contains(path, "internal") {
+			return filepath.SkipDir
+		}
+		if d.Name() == "go.mod" {
+			if err := fn(filepath.Dir(path)); err != nil {
+				return err
+			}
+		}
+		return nil
```
noahdietz, line 77: Nit/alternative layout:
```suggestion
		if d.Name() != "go.mod" {
			return nil
		}

                // process Go module directory.
                return fn(filepath.Dir(path))
```

internal/postprocessor/execv/gocmd/gocmd.go
```diff
@@ -0,0 +1,143 @@
...
+// Build attempts to build all packages recursively from the given directory.
+func Build(dir string) error {
+	log.Println("building generated code")
+	c := execv.Command("go", "build", "./...")
+	c.Dir = dir
+	if _, err := c.Output(); err != nil {
+		if ee, ok := err.(*exec.ExitError); ok {
+			log.Printf("Error Output: %s", ee.Stderr)
+		}
+		return err
```
noahdietz, line 113: I would do a similar restructuring here.

internal/postprocessor/execv/gocmd/gocmd.go
```diff
@@ -0,0 +1,143 @@
...
+	c := execv.Command("go", "list", "-f", "'{{.Module.Dir}}'")
+	c.Dir = dir
+	if out, err = c.Output(); err != nil {
+		if ee, ok := err.(*exec.ExitError); ok {
+			if strings.Contains(string(ee.Stderr), "build constraints exclude all Go files") {
+				return "", ErrBuildConstraint
+			}
+		}
+		return "", err
+	}
```
noahdietz, line 100: Nit/[style](https://google.github.io/styleguide/go/decisions#indent-error-flow), the primary goal is to reduce indentation and clutter (`"If you are using a variable for more than a few lines of code, it is generally not worth using the if-with-initializer style."`):
```suggestion
	out, err = c.Output()
	var ee *exec.ExitError
	if errors.As(err, &ee)) && strings.Contains(string(ee.Stderr), "build constraints exclude all Go files") {
		return "", ErrBuildConstraint
	} else if err != nil {
		return "", err
	}
```

prehaps the `Contains` check could be shortened/simplified too, idk.

noahdietz (commented):
Cool stuff @adrianajg ! If any of my comments refer to code that was strictly moved directories, feel free to ignore them. They are all style nits anyways :)

internal/postprocessor/execv/gocmd/gocmd.go
adrianajg: Restructured, but put inside if block checking if error is not nil as per Cody's recommendation.

adrianajg (commented):
Thanks for the review! I addressed your comments, sometimes with additional tweaks. I also updated the README.md as I noticed there was some outdated info there. Let me know if you see anything else.

codyoss (approved):
LGTM, thanks

noahdietz (approved):

quartzmo:
🎉

PR #7220, opened 2023-01-05
docs: add a debug.md

codyoss:
This is a doc that can grow overtime. It is meant to be a resource we can share with our users when they open issues and are trying to track down bugs.


debug.md
```diff
@@ -0,0 +1,150 @@
...
+## gRPC based clients
+
+### Try setting grpc-go's debug variables
+
+Try setting the following environment variables for grpc-go:
+`GRPC_GO_LOG_VERBOSITY_LEVEL=99` `GRPC_GO_LOG_SEVERITY_LEVEL=info`. These are
+good for diagnosing connection level failures. For more information please see
+[grpc-go's debug documentation](https://pkg.go.dev/google.golang.org/grpc/examples/features/debugging#section-readme).
+
+### Add in your own logging with a gRPC interceptors
```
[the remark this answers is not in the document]
codyoss: I think that is a good call out. I will add a disclaimer to the sections the log request payloads

debug.md
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func main() {
+    ctx := context.Background()
+
+    // Create a transport with authentication built-in detected with
+    // [ADC](https://google.aip.dev/auth/4110). Note you will have to pass any
+    // required scoped for the client you are using.
+    trans, err := htransport.NewTransport(ctx,
+        http.DefaultTransport,
```
tritone, line 70: Note that this leaves out the perf tuning on the transport that we do in google-api-go-client. I've struggled to communicate this with customers in the past, any thoughts?

debug.md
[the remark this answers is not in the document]
tritone: +1, this should be a top level thing.

noahdietz (approved):

debug.md
codyoss: I think that is fine for the scope of this document. I would say these are one-off type of things users should add to their code when required. I would not expect this to be something that is long-lived. So I think it is fine that we gloss over those details here. Thoughts?

tritone: If that's the case I would want to clarify that these are intended for local debugging and not for production use. I have seen users put this kind of logic into production to try to catch transient issues for example. It's problematic if they accidentally hurt their performance while doing that

codyoss: Added a general top-level warning about perf and such

debug.md
```diff
@@ -0,0 +1,162 @@
+# Debugging tips and tricks
+
+While working with the Go Client libraries you may run into some situations
+where you need a deeper level of understanding about what is going on in order
+to solve your problem. Here are some tips and tricks that you can use in these
+cases. Note that many of the tips in this document will have a performance
```
tritone, line 6: Maybe bold *Note* as well :)

tritone (approved):

debug.md
codyoss: ack

PR #7230, opened 2023-01-10
feat(storage): add ComponentCount as part of ObjectAttrs

Tulsishah:


storage/storage.go
```diff
@@ -1315,6 +1315,27 @@ type ObjectAttrs struct {
 	// later value but not to an earlier one. For more information see
 	// https://cloud.google.com/storage/docs/metadata#custom-time .
 	CustomTime time.Time
+
+	// As of 2015-06-03, the official GCS documentation for this
```
tritone, line 1319: This is too much detail for public docs, and we are not faking a component count of 1 in this library. You can just give a short summary of what component count is instead.

I would say something like: `ComponentCount is the number of objects contained within a composite object. For non-composite objects, the value will be zero. This field is read-only.`

storage/integration_test.go
```diff
@@ -1662,6 +1662,9 @@ func TestIntegration_Compose(t *testing.T) {
 		if _, err := c.Run(ctx); err != nil {
 			t.Fatalf("ComposeFrom error: %v", err)
 		}
+		if c.ComponentCount != int64(len(objects)) {
+			t.Fatalf("ComponentCount of dst Object: %v is incorrect", c.ComponentCount)
```
tritone, line 1666: This should be `Errorf` rather than `Fatalf` since it does not block continuing the test. Same for the other error below.

Also, you should specify "got" and "want" values in the errors.

tritone (commented):
Two minor comments, otherwise looks good.

storage/integration_test.go
Tulsishah: fixed

tritone (approved):
LGTM pending presubmit run

tritone (commented):
LGTM after fixing `attrToProto` method

storage/storage.go
```diff
@@ -1585,6 +1586,7 @@ var attrToProtoFieldMap = map[string]string{
 	"Owner":                   "owner",
 	"CustomerKeySHA256":       "customer_encryption",
 	"CustomTime":              "custom_time",
+	"ComponentCount":          "componentCount",
```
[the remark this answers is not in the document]
Tulsishah: fixed

tritone (approved):

Issue #7233, opened 2023-01-10
longrunning: cannot build longrunning since v0.3.0


darist:
**Client**

longrunning

**Environment**

linux (using bazel)

**Go Environment**

$ go version

go version go1.19.4 linux/amd64

$ go env

GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/darist/.cache/go-build"
GOENV="/home/darist/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/darist/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/darist/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/lib/go-1.19"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go-1.19/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.19.4"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/darist/tmp/datainfra-beagle-fusion/go.mod"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build3973921892=/tmp/go-build -gno-record-gcc-switches"

**Code**

```shell
$ bazel build @com_google_cloud_go_longrunning//:longrunning
```

**Expected behavior**

Expect build to succeed.

**Actual behavior**

```shell
$ bazel build @com_google_cloud_go_longrunning//:longrunning
INFO: Analyzed target @com_google_cloud_go_longrunning//:longrunning (3 packages loaded, 9 targets configured).
INFO: Found 1 target...
[...]

PR #7234, opened 2023-01-10
chore: update import paths in manual layers

codyoss:
Updates: #7233


robfig:
I was also running into this issue - your commit fixed some of it, but I'm also encountering other usages of the old pb path in /logging, e.g.
```
external/com_google_cloud_go_logging/apiv2/config_client.go:1158:55: cannot use resp (variable of type *"google.golang.org/genproto/googleapis/longrunning".Operation) as type *longrunningpb.Operation in argument to longrunning.InternalNewOperation
```

Looks like the import is still the old one at HEAD:
```
	longrunningpb "google.golang.org/genproto/googleapis/longrunning"
```
https://github.com/googleapis/google-cloud-go/blob/main/logging/apiv2/config_client.go#L34

Should there be a repo-wide find-replace, or am I misunderstanding? 
Thanks!

codyoss:
@robfig As long as you are using a newer version of the genproto library this should not be an issue. This is mostly a cosmetic change in the code.

codyoss:
But yes, more locations likely can be updated as well and will be.

quartzmo (approved):

quartzmo:
@robfig Here is the alias for `Operation` in the latest `go-genproto` from `google.golang.org/genproto/googleapis/longrunning` to `cloud.google.com/go/longrunning/autogen/longrunningpb`: https://github.com/googleapis/go-genproto/blob/main/googleapis/longrunning/alias.go#L70

PR #7235, opened 2023-01-10
feat(bigquery): add dataset/table collation

alvarowolfx:


bigquery/dataset.go
```diff
@@ -34,6 +34,19 @@ type Dataset struct {
 	c         *Client
 }
 
+// Collation describes the collation type of a given table or dataset
+type Collation string
+
+// Constants describing available collations
+var (
```
shollyman, line 41: The main thing I'm worried about here is whether we should be striving to present an enum-like thing, or if we should treat this more like a string.  Will we end up with a bunch of locale codes and suffixes for behavior?

bigquery/dataset.go
```diff
@@ -104,6 +125,10 @@ type DatasetMetadataToUpdate struct {
 	// in the dataset.
 	DefaultEncryptionConfig *EncryptionConfig
 
+	// Defines the default collation specification of future tables
+	// created in the dataset.
+	DefaultCollation *Collation
```
shollyman, line 130: This gets back to the "enum or string" point.  If our values are more open-ended, maybe this should be optional.String?

bigquery/table_integration_test.go
```diff
@@ -592,3 +592,70 @@ func TestIntegration_TableUseLegacySQL(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer table.Delete(ctx)
+	md, err := table.Metadata(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if md.DefaultCollation != CaseInsensitiveCollation {
+		t.Fatalf("expected default collation to be `%v`, but found `%v`", CaseInsensitiveCollation, md.DefaultCollation)
```
shollyman, line 617: Given the type, %q is probably more appropriate for the formatter.

bigquery/dataset.go
alvarowolfx: that's a good point. I ended up guiding myself up just from the proto files. Reading the docs around the collation feature it make sense to be more open as a string.

bigquery/table_integration_test.go
alvarowolfx: that's a good one, I didn't know about `%q`

bigquery/table_integration_test.go
```diff
@@ -592,3 +592,72 @@ func TestIntegration_TableUseLegacySQL(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if md.DefaultCollation != caseInsensitiveCollation {
+		t.Fatalf("expected default collation to be `%v`, but found `%v`", caseInsensitiveCollation, md.DefaultCollation)
+	}
+	for _, field := range md.Schema {
+		if field.Type == StringFieldType {
+			if field.Collation != caseInsensitiveCollation {
+				t.Fatalf("expected all columns to have collation `%v`, but found `%v` on field :%v", caseInsensitiveCollation, field.Collation, field.Name)
```
shollyman, line 624: more candidates for %q here and elsewhere in the tests?

shollyman (approved):
Generally looks good.  Do we have a link to BQ docs we can add to the new fields to help users understand the behaviors in more detail?  If so, let's reference them in the field comments.

PR #7256, opened 2023-01-13
fix(bigquery): create/update an isolated dataset for collation feature

alvarowolfx:
Resolves flaky tests from https://github.com/googleapis/google-cloud-go/issues/7254


shollyman (approved):
This addresses the manipulation of the main test dataset. Do we still have the issue of the feature enablement in the pool projects or was all the noise just this?

PR #7257, opened 2023-01-13
refactor: simplify options, instrumentation for multiplexing

shollyman:
This PR consolidates where in managedwriter that options apply by moving the destinationTable inside of streamSettings.

This PR also make changes to how instrumentation works.  Currently, we retain a long-lived context on the the ManagedStream and that's used for both connection management and to attach tag context to instrumentation reporting (e.g. to tag metrics with the appropriate keys).  With this change, connection-oriented metrics no longer tag with the writer tags (stream ID, data origin).  We also remove the stream's keys from the default views the package exposes for these metrics.

When we cut over to connections and pools, the context retained on the ManagedStream will only used for metrics tagging, and thus will come from the parent context setup elsewhere (namely the pool context).


bigquery/storage/managedwriter/instrumentation.go
```diff
@@ -173,24 +171,41 @@ func createSumView(m stats.Measure, keys ...tag.Key) *view.View {
...
+	return kCtx
+}
+
+// recordWriterStat records a measure which may optionally contain writer-related tags like stream ID
+// or data origin.
+func recordWriterStat(ms *ManagedStream, m *stats.Int64Measure, n int64) {
+	stats.Record(ms.ctx, m.M(n))
+}
+
+func recordConnStat(ms *ManagedStream, m *stats.Int64Measure, n int64) {
```
alvarowolfx, line 207: unused empty function

bigquery/storage/managedwriter/integration_test.go
```diff
@@ -894,8 +894,16 @@ func testInstrumentation(ctx context.Context, t *testing.T, mwClient *Client, bq
...
-		if len(metricData[0].Tags) != 1 {
-			t.Errorf("%q: only expected 1 tag, got %d", tv.Name, len(metricData[0].Tags))
+		var wantTags int
+		switch tv.Name {
+		case "cloud.google.com/go/bigquery/storage/managedwriter/stream_open_count",
+			"cloud.google.com/go/bigquery/storage/managedwriter/stream_open_retry_count":
+			wantTags = 0
+		default:
+			wantTags = 1
+		}
```
alvarowolfx, line 904: I feel like this check for the amount of tags ended up a bit confusing. Maybe instead of iterating on `testedViews`, a custom struct can be created that have a `View` and expected tags, following a table driven pattern.

alvarowolfx (approved):

bigquery/storage/managedwriter/instrumentation.go
shollyman: removed

bigquery/storage/managedwriter/integration_test.go
shollyman: added a wantTags map for validating this more explicityly.

Issue #7207, opened 2023-01-04
logging: make it easier to emit Request-scoped logs using standard golang logging APIs


tang-fh:
**Is your feature request related to a problem? Please describe.**

In Cloud Run, if the emitted log entry has its trace field populated with the request's trace, then those logs are grouped as children of the (Platform-emitted) request log.

The mechanism is explained in https://github.com/googleapis/google-cloud-go/blob/02c04a7c6c5e96026ce0710748a54580065a8102/logging/doc.go#L131-L148

In the case of Cloud Run, the request log has the necessary fields populated to be interpreted as a parent entry.

**Describe the solution you'd like**

I would like a method
```
func (l *Logger) RequestLogger(s Severity, r *http.Request) *log.Logger
```
analogous to 
```
func (l *Logger) StandardLogger(s Severity) *log.Logger
```
which would enable code of the form
```
type myHandler struct {
    logger *logging.Logger
}

func (h *myHandler) ServeHttp(w http.ResponseWriter, r *http.Request) {
    slg := h.logger.RequestLogger(logging.Info, r)

    slg.Println("an informative message")
}
```

**Describe alternatives you've considered**
The current workaround would be to explicitly set field `logging.Entry.Trace` with the trace value extracted from the `http.Request` object. Extracting the trace is non-trivial. See an example implementation in https://github.com/googleapis/google-cloud-go/blob/02c04a7c6c5e96026ce0710748a54580065a8102/logging/logging.go#L732-L753

**Additional context**
Logs Explorer in Cloud Console can render child log entries in tree-like format underneath the parent (request) log entries.

daniel-sanche:
Hey Francis, thanks for putting this together!

> The current workaround would be to explicitly set field logging.Entry.Trace ...

Note that an HTTPRequest object can be attached to a log if you use the client.Logger object directly, rather than the StandardLogger. Example here: https://pkg.go.dev/cloud.google.com/go/logging#example-HTTPRequest. But after a quick look I think you're correct that there's no easy way to do this with the Standard Logger

> I would like a method
> func (l *Logger) RequestLogger(s Severity, r *http.Request) *log.Logger
> 
> analogous to
> func (l *Logger) StandardLogger(s Severity) *log.Logger

[...]

PR #7260, opened 2023-01-16
fix(logging): Prefer X-Cloud-Trace-Context over Traceparent

tang-fh:
When both X-Cloud-Trace-Context and Traceparent headers are present in the request, extract trace context from the former.

Addresses a bug identified in issue #7207.

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -729,18 +729,7 @@ func populateTraceInfo(e *Entry, req *http.Request) bool {
 			return false
 		}
 	}
-	header := req.Header.Get("Traceparent")
-	if header != "" {
-		// do not use traceSampled flag defined by traceparent because
-		// flag's definition differs from expected by Cloud Tracing
-		traceID, spanID, _ := deconstructTraceParent(header)
-		if traceID != "" {
-			e.Trace = traceID
-			e.SpanID = spanID
-			return true
-		}
-	}
-	header = req.Header.Get("X-Cloud-Trace-Context")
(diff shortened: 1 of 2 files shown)


logging/logging.go
```diff
@@ -729,25 +729,25 @@ func populateTraceInfo(e *Entry, req *http.Request) bool {
 			return false
 		}
 	}
-	header := req.Header.Get("Traceparent")
+	header := req.Header.Get("X-Cloud-Trace-Context")
```
losalex, line 732: Not sure is it is valid concern, but I wonder if such a change can break existing clients which get used to previous order...

tang-fh: > Not sure is it is valid concern, but I wonder if such a change can break existing clients which get used to previous order...

I think this is a valid concern. A change like this for sure can break existing consumers of this library which depend on the old behaviour.

daniel-sanche: In that case, I think we should close this and [save it for v2.0.0](https://github.com/googleapis/google-cloud-go/milestone/15)

losalex: Makes sense. One additional option is to introduce config parameter controlling which header is taken first - by default such config should prefer existing method to keep backward compatibility. With v.2.0.0 release we could switch default to be a new behavior

tang-fh:
Thanks for filing https://github.com/googleapis/google-cloud-go/issues/7264 .

I think it makes sense not to proceed with this PR in its current form.

PR #7261, opened 2023-01-16
feat(logging): Add (*Logger). StandardLoggerFromTemplate() method.

tang-fh:
This method is a generalised but more expensive analog of (*Logger).StandardLogger() which takes a prototype Entry struct as argument.


daniel-sanche:
This PR would have breaking changes over the last release. We need to either change it to keep existing functionality stable, or save it for [v2.0.0](https://github.com/googleapis/google-cloud-go/milestone/15)

tang-fh:
> This PR would have breaking changes over the last release. We need to either change it to keep existing functionality stable, or save it for [v2.0.0](https://github.com/googleapis/google-cloud-go/milestone/15)

This wasn't intended to be a breaking change. See my issue comment https://github.com/googleapis/google-cloud-go/issues/7207#issuecomment-1387140558 .

logging/logging.go
```diff
@@ -301,16 +302,15 @@ func (c *Client) Logger(logID string, opts ...LoggerOption) *Logger {
...
-	s Severity
+	e *Entry // A prototype Entry object
 }
 
-func (w severityWriter) Write(p []byte) (n int, err error) {
-	w.l.Log(Entry{
-		Severity: w.s,
-		Payload:  string(p),
-	})
+func (w protoEntryWriter) Write(p []byte) (n int, err error) {
```
daniel-sanche, line 310: nit: The `proto` prefix usually is typically used to mean `protobuf` elsewhere in the codebase. I think this should be renamed to reduce potential confusion. Maybe call it a `template` instead?

logging/logging_test.go
```diff
@@ -717,6 +717,47 @@ func TestStandardLogger(t *testing.T) {
...
 }
 
+func TestStandardLoggerFromEntry(t *testing.T) {
+	initLogs() // Generate new testLogID
+	ctx := context.Background()
+	lg := client.Logger(testLogID)
+
+	slg := lg.StandardLoggerFromEntry(&logging.Entry{
+		Severity: logging.Info,
+		Trace:    "projects/P/traces/105445aa7843bc8bf206b120001000",
```
daniel-sanche, line 727: More tests with different metadata filled out would be useful. Especially with payloads in the prototype entry, we want to make sure that's well covered by tests

tang-fh: Done

logging/logging.go
tang-fh: Done.

Good idea. I also think of `proto` to mean proto message. `template` is better.

daniel-sanche (approved):
LGTM

PR #7262, opened 2023-01-17
fix(compute/metadata): Remove unused dependency - cloud.google.com/go/compute

dims:
This PR is intended to help [Untangle genproto / cloud.google.com/go dependencies](https://github.com/kubernetes/kubernetes/issues/113366). 

I see that the reference to the previous package was explicitly added following instructions in https://github.com/golang/go/wiki/Modules#is-it-possible-to-add-a-module-to-a-multi-module-repository, i am not sure we need to keep that reference going forward (i am sure it was needed to bootstrap this module). Please see the kubernetes issue above for the problems it causes.

cc @liggitt 

Signed-off-by: Davanum Srinivas <davanum@gmail.com>


liggitt:
I know this is on @codyoss's radar, but I'm not sure of the expected timeline for dropping this

liggitt:
with go1.20 released, is this unblocked now?

PR #7271, opened 2023-01-19
fix(storage): specify credentials with STORAGE_EMULATOR_HOST

shuheiktgw:
Encountered the same problem with https://github.com/googleapis/google-api-go-client/issues/558 when specifying `STORAGE_EMULATOR_HOST`. For our case, we'd like to pass `option.WithTokenSource`, and it raises the following error with `STORAGE_EMULATOR_HOST` during the test. To avoid the error, we should pass `internaloption.SkipDialSettingsValidation` as the Spanner client does.

```
options.WithoutAuthentication is incompatible with any option that provides credentials
```


storage/storage_test.go
```diff
@@ -1992,6 +1992,20 @@ func TestAttrToFieldMapCoverage(t *testing.T) {
...
 
+func TestEmulatorWithCredentialsFile(t *testing.T) {
+	originalStorageEmulatorHost := os.Getenv("STORAGE_EMULATOR_HOST")
+	defer os.Setenv("STORAGE_EMULATOR_HOST", originalStorageEmulatorHost)
+
+	os.Setenv("STORAGE_EMULATOR_HOST", "localhost:1234")
+
+	client, err := NewClient(context.Background(), option.WithCredentialsFile("/path/to/key.json"))
+	if err != nil {
+		t.Errorf("failed creating a client with credentials file when running agains an emulator: %v", err)
```
tritone, line 2003: you can just use t.Fatalf()

tritone (commented):
Thanks for the PR! I think this looks good to me. @BrennaEpp or @codyoss any thoughts?

storage/storage_test.go
shuheiktgw: Thank you. Fixed it 🙂

shuheiktgw:
@tritone @BrennaEpp @codyoss I've fixed the PR, so would you review it again? 🙏

storage/storage_test.go
```diff
@@ -1992,6 +1992,20 @@ func TestAttrToFieldMapCoverage(t *testing.T) {
 	}
 }
 
+func TestEmulatorWithCredentialsFile(t *testing.T) {
+	originalStorageEmulatorHost := os.Getenv("STORAGE_EMULATOR_HOST")
+	defer os.Setenv("STORAGE_EMULATOR_HOST", originalStorageEmulatorHost)
+
+	os.Setenv("STORAGE_EMULATOR_HOST", "localhost:1234")
```
BrennaEpp, line 1999:
```suggestion
	t.Setenv("STORAGE_EMULATOR_HOST", "localhost:1234")
```

BrennaEpp (commented):
Just one suggestion, otherwise looks good.

storage/storage_test.go
shuheiktgw: Didn't know t.Setenv, thank you!

shuheiktgw:
@tritone @BrennaEpp @codyoss Thank you for your review! Would you review the PR again?

tritone (approved):
thanks for the contribution!

PR #7294, opened 2023-01-24
fix(internal/postprocessor): include module and package in scope

adrianajg:
With this change the postprocessor includes the module in scopes for nested clients.

### DIFF
--- a/internal/postprocessor/main.go
+++ b/internal/postprocessor/main.go
@@ -363,7 +363,10 @@ func (c *config) getScopeFromGoogleapisCommitHash(commitHash string) (string, er
 	for _, filePath := range files {
 		for _, config := range generator.MicrogenGapicConfigs {
 			if config.InputDirectoryPath == filepath.Dir(filePath) {
-				scope := config.Pkg
+				// trim prefix
+				scope := strings.TrimPrefix(strings.TrimPrefix(config.InputDirectoryPath, "google/"), "cloud/")
+				// trim version
+				scope = filepath.Dir(scope)
 				if _, value := scopesMap[scope]; !value {
 					scopesMap[scope] = true
 					scopes = append(scopes, scope)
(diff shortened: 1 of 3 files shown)


internal/postprocessor/main.go
```diff
@@ -363,7 +363,10 @@ func (c *config) getScopeFromGoogleapisCommitHash(commitHash string) (string, er
 	for _, filePath := range files {
 		for _, config := range generator.MicrogenGapicConfigs {
 			if config.InputDirectoryPath == filepath.Dir(filePath) {
-				scope := config.Pkg
+				// trim prefix
+				scope := strings.TrimPrefix(strings.TrimPrefix(config.InputDirectoryPath, "google/"), "cloud/")
```
codyoss, line 367: I am not sure we should use the inputdirectorypath here. There are other clients that come from other roots. For example the cloudbuild client comes from devtools: https://github.com/googleapis/googleapis/tree/master/google/devtools/cloudbuild

adrianajg: In the example you are providing the InputDirectoryPath would be `"google/devtools/cloudbuild/v1"`, and after running this logic the scope would be `"devtools/cloudbuild"`. Is that not correct?

codyoss: nope, we trim off devtools and it becomes just cloudbuild: https://pkg.go.dev/cloud.google.com/go/cloudbuild.

I think there are a few other edge cases like this that make it hard to make assumptions off of the inputDir

adrianajg: Okay is it safe to use those assumptions with ImportPath instead and trip `cloud.google.com/go` and the version?

adrianajg: Used ImportPath

codyoss (approved):

PR #7295, opened 2023-01-24
feat(pubsub): add support for schema revisions

hongalex:
This PR adds support for committing, rolling back, listing, and deleting schema revisions.

There are also changes made to the fake to embed unimplemented versions of the publish, subscriber, and schema services rather than the interface. This is causing apidiff to complain since this is a breaking change, but in reality customers can still continue to use the package as intended. This will also reduce the amount of toil in the future as additional methods are added (avoiding the change described [here](https://github.com/googleapis/google-cloud-go/pull/7232#issuecomment-1377830786)).


pubsub/pstest/fake.go
```diff
@@ -1391,29 +1404,30 @@ func (s *GServer) CreateSchema(_ context.Context, req *pb.CreateSchemaRequest) (
...
+		RevisionId:         genRevID(),
+		RevisionCreateTime: timestamppb.Now(),
 	}
-	s.schemas[name] = sc
+	s.schemas[name] = append(s.schemas[name], sc)
 
 	return sc, nil
 }
 
 func (s *GServer) GetSchema(_ context.Context, req *pb.GetSchemaRequest) (*pb.Schema, error) {
```
kamalaboulhosn, line 1418: GetSchema also allows one to pass in a path with the revision ID, e.g., projects/myproject/schemas/myschema@myrevision

kamalaboulhosn (changes requested):
Do we need to make any changes to topic to support the fact that schema settings can now be updated or will that just work?

pubsub/pstest/fake.go
hongalex: Done

hongalex:
> Do we need to make any changes to topic to support the fact that schema settings can now be updated or will that just work?

Yeah this didn't work when I first drafted this. I've since added the proper piping for that and tests for this.

pubsub/integration_test.go
```diff
@@ -2011,3 +2014,49 @@ func TestExactlyOnceDelivery_PublishReceive(t *testing.T) {
 	// Tests for large messages (larger than the 4MB gRPC limit).
 	testPublishAndReceive(t, client, 0, false, true, 1, 5*1024*1024)
 }
+
+func TestIntegration_TopicUpdateSchema(t *testing.T) {
+	ctx := context.Background()
+	// TODO(hongalex): update these staging endpoints after schema evolution is GA.
+	c := integrationTestClient(ctx, t, option.WithEndpoint("staging-pubsub.sandbox.googleapis.com:443"))
```
kamalaboulhosn, line 2021: Let's just not even merge until we can remove this. That should hopefully be next week.

kamalaboulhosn (approved):

PR #7300, opened 2023-01-25
feat(internal/postprocessor): only regen snippets for changed modules

adrianajg:
This change identifies the changed modules in the PR using the commit URLs in the open OwlBot PR description. The scopes are used to limit snippet regeneration to only those modules.


internal/postprocessor/main.go
```diff
@@ -193,15 +197,72 @@ func (c *config) RegenSnippets() error {
 	return nil
 }
 
-func (c *config) replaceAllForSnippets(googleCloudDir, snippetDir string) error {
-	return execv.ForEachMod(googleCloudDir, func(dir string) error {
+func (c *config) parseAPIShortnames(confs []*generator.MicrogenConfig, manualEntries []generator.ManifestEntry) (map[string]string, error) {
+	shortnames := map[string]string{}
+	runConfs := confs
```
codyoss, line 202: If you need to copy a slice you almost always want to copy the elments explicitly or you can have unintended behavior creep in: https://go.dev/play/p/gkUJknLHYiU

internal/postprocessor/main.go
```diff
@@ -280,23 +341,24 @@ func docURL(cloudDir, importPath string) (string, error) {
 	return "https://cloud.google.com/go/docs/reference/" + mod + "/latest/" + pkgPath, nil
 }
 
-func (c *config) AmendPRDescription(ctx context.Context) error {
+func (c *config) AmendPRDescription(ctx context.Context) ([]string, error) {
```
codyoss, line 344: Can you add docs to this function. Without them it is unclear why or what this function is returning by reading the signature.

internal/postprocessor/main.go
```diff
@@ -280,23 +341,24 @@ func docURL(cloudDir, importPath string) (string, error) {
...
 	log.Println("Amending PR title and body")
 	pr, err := c.getPR(ctx)
 	if err != nil {
-		return err
+		return []string{}, err
```
codyoss, line 348: For slices, always return nil in cases like these.

codyoss: Happens more below.

internal/postprocessor/main.go
```diff
@@ -280,23 +341,24 @@ func docURL(cloudDir, importPath string) (string, error) {
...
 	}
-	return c.writePRCommitToFile(newPRTitle, newPRBody)
+	return modules, c.writePRCommitToFile(newPRTitle, newPRBody)
 }
 
-func (c *config) processCommit(title, body string) (string, string, error) {
+func (c *config) processCommit(title, body string) (string, string, []string, error) {
```
codyoss, line 357: I think this functions is starting to return too much. I would either refactor this work to be in multiple functions if possible or return a struct of data.

internal/postprocessor/main.go
```diff
@@ -150,6 +149,13 @@ func (c *config) run(ctx context.Context) error {
 		log.Println("exiting post processing early")
 		return nil
 	}
+	// TODO(codyoss): In the future we may want to make it possible to be able
+	// to run this locally with a user defined remote branch.
+	modules, err := c.AmendPRDescription(ctx)
+	if err != nil {
+		return err
+	}
+	c.modules = append(c.modules, modules...)
```
codyoss, line 158: This feels odd here. The method that returns this has access to c already and could maybe be manipulated deeper?

codyoss: It would also be great if we could feed module to the modTidy command run. Those can be expensive calls too so limiting the processing to affected module would speed things up even more.

codyoss (changes requested):

internal/postprocessor/main.go
adrianajg: I think I fixed this by changing the function to return only an error.

internal/postprocessor/main.go
adrianajg: Done

internal/postprocessor/main.go
adrianajg: I've changed the logic so that, rather than returning scopes, it simply modifies the config within the function. Do you think this return is okay now?

internal/postprocessor/main.go
```diff
@@ -150,6 +149,13 @@ func (c *config) run(ctx context.Context) error {
...
 	if err := gocmd.ModTidyAll(c.googleCloudDir); err != nil {
 		return err
 	}
```
adrianajg, line 159: Do we want to add a check to see if c.modules is empty, and if so, mod tidy all?

adrianajg (commented):
Addressed requested changes and have a question if you don't mind taking a look. Thank you!

internal/postprocessor/main.go
```diff
@@ -150,7 +149,12 @@ func (c *config) run(ctx context.Context) error {
 		log.Println("exiting post processing early")
 		return nil
 	}
-	if err := gocmd.ModTidyAll(c.googleCloudDir); err != nil {
+	// TODO(codyoss): In the future we may want to make it possible to be able
+	// to run this locally with a user defined remote branch.
+	if err := c.AmendPRDescription(ctx); err != nil {
```
codyoss, line 154: I think it may be best to keep this at the end of the logic if possible. That way we don't update the PR if other steps failed. Thoughts?

internal/postprocessor/main.go
```diff
@@ -72,13 +72,12 @@ func main() {
 	log.Println("googleapis-dir set to", *googleapisDir)
 	log.Println("branch set to", *branchOverride)
 	log.Println("prFilepath is", *prFilepath)
+	log.Println("directories are", *directories)
 
 	var modules []string
```
codyoss, line 77: Can this variable just go away and we pass directories to our struct?

internal/postprocessor/main.go
```diff
@@ -162,28 +166,41 @@ func (c *config) run(ctx context.Context) error {
...
+	for _, module := range c.modules {
+		dirs = append(dirs, filepath.Join(c.googleCloudDir, module))
+	}
+
+}
+
+func (c *config) TidyAffectedMods() error {
+	dirs := c.getDirs()
+	for _, dir := range dirs {
+		if err := gocmd.ModTidy(dir); err != nil {
```
codyoss, line 183: This works nice for running locally, but will dirs be populated in the case of owlbot PRs? Maybe we need to find a way to programitically figure this out with git commands.

internal/postprocessor/main.go
codyoss: I think if empty we should detect what code to run for. See other comments.

internal/postprocessor/main.go
```diff
@@ -193,15 +210,42 @@ func (c *config) RegenSnippets() error {
 	return nil
 }
 
-func (c *config) replaceAllForSnippets(googleCloudDir, snippetDir string) error {
-	return execv.ForEachMod(googleCloudDir, func(dir string) error {
+func (c *config) getConfs() []*generator.MicrogenConfig {
```
codyoss, line 213: This getter in more involved and selectively does things. Would you mind adding some godoc here.

internal/postprocessor/main.go
codyoss: I see now why you movded it up. I would say we should refactor out the bit we need to do earlier(module dection),

codyoss (commented):
A few comments

internal/postprocessor/main.go
adrianajg: Done, I also changed the name to be more descriptive.

internal/postprocessor/main.go
adrianajg: I am resolving this for now, and may look into the possibility of extracting the changed modules for this part of the postprocessing with git commands in another PR.

internal/postprocessor/main.go
adrianajg: Separated it so that the PR Title and Description are stored in the config struct and set at the beginning of running the postprocessor. Writing to file occurs separately as a last step in the postprocessing.

adrianajg (commented):
Addressed changes. Let me know if you see anything else. Thanks!

codyoss (approved):

PR #7319, opened 2023-01-27
fix(pubsub): fix bug with AckWithResult with exactly once disabled

### LINKED ISSUE #7301: pubsub: AckWithResult does not request ack when exactly-once delivery is disabled but should do
Hi team!

Pubsub.AckWithResult and NackWithResult methods does not call doneFunc(meaning not requesting ack to subscription), just return mocked success response when exactly-once delivery is disabled.

document and comments says like below

> If exactly-once delivery is NOT enabled on the subscription, or if using Pub/Sub Lite, AckResult readies immediately with a AcknowledgeStatus.Success. Since acks in Cloud Pub/Sub are best effort when exactly-once delivery is disabled, the message may be re-delivered. Because re-deliveries are possible, you should ensure that your processing code is idempotent, as you may receive any given message more than once.
https://pkg.go.dev/cloud.google.com/go/internal/pubsub#Message.AckWithResult

I expected AckWithResult request ack in background and always responds success response, but actually **not requesting ack in non-exactly-once delivery environment.**

Is this behaviour expected?

```go
func (ah *psAckHandler) OnAckWithResult() *AckResult {
	if !ah.exactlyOnceDelivery {
		return newSuccessAckResult()
	}
	// call done with true to indicate ack.
	ah.done(true)
	return ah.ackResult
}

func (ah *psAckHandler) OnNackWithResult() *AckResult {
	if !ah.exactlyOnceDelivery {
		return newSuccessAckResult()
	}
	// call done with false to indicate nack.
	ah.done(false)
	return ah.ackResult
}
```
https://github.com/googleapis/google-cloud-go/blob/82aa2ee9381f793bd731f1b6789fc18e4b671bd7/pubsub/message.go#L160

Adding ah.done call also in not exactly-once delivery environment.

Or changing code comments.

hongalex:
Fixes #7301

### DIFF
--- a/pubsub/streaming_pull_test.go
+++ b/pubsub/streaming_pull_test.go
@@ -513,8 +513,8 @@ func newMock(t *testing.T) (*Client, *mockServer) {
 }
 
 // pullN calls sub.Receive until at least n messages are received.
-// Wait a provided number of seconds before cancelling.
-func pullN(ctx context.Context, sub *Subscription, n, wait int, f func(context.Context, *Message)) ([]*Message, error) {
+// Wait a provided duration before cancelling.
+func pullN(ctx context.Context, sub *Subscription, n int, wait time.Duration, f func(context.Context, *Message)) ([]*Message, error) {
 	var (
 		mu   sync.Mutex
 		msgs []*Message
@@ -529,7 +529,7 @@ func pullN(ctx context.Context, sub *Subscription, n, wait int, f func(context.C
 		if nSeen >= n {
 			// Wait a specified amount of time so that for exactly once delivery,
 			// Acks aren't cancelled immediately.
-			time.Sleep(time.Duration(wait) * time.Second)
+			time.Sleep(wait)
 			cancel()
 		}
 	})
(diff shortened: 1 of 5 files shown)


pubsub/streaming_pull_test.go
```diff
@@ -513,7 +513,8 @@ func newMock(t *testing.T) (*Client, *mockServer) {
 }
 
 // pullN calls sub.Receive until at least n messages are received.
-func pullN(ctx context.Context, sub *Subscription, n int, f func(context.Context, *Message)) ([]*Message, error) {
+// Wait a provided number of seconds before cancelling.
+func pullN(ctx context.Context, sub *Subscription, n, wait int, f func(context.Context, *Message)) ([]*Message, error) {
```
shollyman, line 517: the wait as int seems like an odd choice (wrong type, no indication of the base interval).  Maybe just make it time.Duration?

shollyman (approved):
I'm not well versed in the ack handling, but this seems reasonable.  One quick piece of feedback regarding the testing `pullN` args.

PR #7320, opened 2023-01-27
fix(logging): correctly populate SourceLocation when logging via (*Logger).StandardLogger

### LINKED ISSUE #7289: logging: incorrect SourceLocation when logging via (*Logger).StandardLogger()
**Client**

Logging

**Environment**

Any

**Go Environment**

```
$ go version
go version go1.19.5 linux/amd64
$ go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/fhltang/.cache/go-build"
GOENV="/home/fhltang/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/fhltang/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/fhltang/go"
GOPRIVATE=""
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOVCS=""
GOVERSION="go1.19.5"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/fhltang/google-cloud-go/logging/go.mod"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build2749232686=/tmp/go-build -gno-record-gcc-switches"
```

**Code**

See commit 3396afd16cebf748c7fe3f111e2936a5506b9865 (from fork https://github.com/tang-fh/google-cloud-go/tree/sourcelocation) which adds a unit test to reproduce the problem.

```go
	lg := client.Logger(testLogID, logging.SourceLocationPopulation(logging.AlwaysPopulateSourceLocation))
	slg := lg.StandardLogger(logging.Info)

	slg.Print("info")
```

**Expected behavior**

Logs "info" and its source location should be the filename, function and line number of the `slg.Print()` call site.

**Actual behavior**

The source location filename, function and line number are `logging.go` and `cloud.google.com/go/logging.severityWriter.Write` (and whatever line number the call to `logging.Logger.Log()` is.

**Other**

I can prepare a PR to fix this.

tang-fh:
Fixes #7289


logging/logging.go
```diff
@@ -311,7 +311,12 @@ type templateEntryWriter struct {
 func (w templateEntryWriter) Write(p []byte) (n int, err error) {
 	e := *w.template
 	e.Payload = string(p)
-	w.l.Log(e)
+	// The second argument to logInternal() is how many frames to skip
+	// from the call stack when determining the source location. In the
+	// current implementation of log.Logger (i.e. Go's logging library)
```
daniel-sanche, line 316: The "current implementation" part worries me a bit. Is there a way to find this value dynamically?

Although I it looks like the tests will warn us if this value changes in the future?

daniel-sanche (approved):
LGTM

logging/logging.go
tang-fh: Yes, this was my concern too. I'm not sure there's an easy way to determine the call depth correction dynamically. The Go logging library implementation also has such an assumption baked in, see https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/log/log.go;l=165-170;drc=c7942f87a2587ee989f6e282d887b4652119133a and https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/log/log.go;l=213;drc=c7942f87a2587ee989f6e282d887b4652119133a (as an example caller).

I don't think there's a way to avoid this. The only place we can inject this `SourceLocation` is in [`templateWriter.Write()`](https://github.com/googleapis/google-cloud-go/blob/c71b090f59317d43a2a5f2d9dd81c493b14d8908/logging/logging.go#L311). We have to make an assumption for call depth adjustment in any case.

PR #7323, opened 2023-01-30
refactor(bigquery/storage/managedwriter): introduce send optimizers

shollyman:
This PR introduces a new optimizer abstraction for connection objects, but doesn't wire it in (happening in another open PR).

The purpose of the optimizers is to leverage awareness of previous requests to reduce transferred bytes.  We provide three optimizers, based on the expected usage:
* passthroughOptimizer - used for testing (no-op optimizer)
* simplexOptimizer - used for connections where only a single stream will be sending data (e.g. explicit streams)
* multiplexOptimizer - used for connections shared by multiple writers


alvarowolfx (approved):
LGTM

PR #7362, opened 2023-02-03
fix(pubsub): allow updating topic schema fields individually

hongalex:
Currently, the entirety of `SchemaSettings` must be updated when calling `UpdateTopic`. Instead, allow updating these fields individually (e.g. first/last revision) without needing to specify schema ID or encoding.


feywind (approved):

pubsub/pstest/fake.go
```diff
@@ -368,8 +368,26 @@ func (s *GServer) UpdateTopic(_ context.Context, req *pb.UpdateTopicRequest) (*p
...
+			if t.proto.SchemaSettings == nil {
+				t.proto.SchemaSettings = &pb.SchemaSettings{}
+			}
+			t.proto.SchemaSettings.FirstRevisionId = req.Topic.SchemaSettings.FirstRevisionId
+		case "schema_settings.last_revision_id":
+			if t.proto.SchemaSettings == nil {
+				t.proto.SchemaSettings = &pb.SchemaSettings{}
+			}
+			t.proto.SchemaSettings.LastRevisionId = req.Topic.SchemaSettings.LastRevisionId
 		default:
```
kamalaboulhosn, line 391: It is also possible to just have schema_settings, which will update all of them. Though I guess the Go library is setting each field explicitly?

pubsub/topic.go
```diff
@@ -407,7 +407,18 @@ func (t *Topic) updateRequest(cfg TopicConfigToUpdate) *pb.UpdateTopicRequest {
 	}
 	if cfg.SchemaSettings != nil {
```
kamalaboulhosn, line 408: How does one remove the schema config entirely?

kamalaboulhosn (changes requested):

pubsub/topic.go
hongalex: Yeah I missed this. I added a section that handles clearing SchemaSettings when it detects the zero value, similar to MessageStoragePolicy.

pubsub/pstest/fake.go
hongalex: Yeah each field is set explicitly so updating all of them at once is possible too.

hongalex (commented):
Addressed comments. Also, kokoro is failing because of a unrelated sample issue, not because of any tests related to pubsub

kamalaboulhosn (approved):

PR #7372, opened 2023-02-07
feat(spanner): add databoost enabled property for batch transactions

rahul2393:


rahul2393:
@codyoss with this PR we added an optional argument to existing function which will not break older versions still apidiff reports failure, is it ok to ignore?

gauravpurohit06 (approved):

PR #7377, opened 2023-02-08
pubsub/pstest: fix panic on undelivered message

adg:
If a running pstest Server has a subscription without a dead letter topic configured, and there's a pending message in the subscription that is not received after the retention duration (10 minutes), the deliver loop for the subscription will trigger a panic.

This change prevents the panic by checking not trying to deliver to the dead letter topic unless one is configured.

Also includes a refactor to the time handling to make testing with time less awkward.


pubsub/pstest/fake.go
```diff
@@ -1143,7 +1133,9 @@ func (s *subscription) maintainMessages(now time.Time) {
 		pubTime := m.proto.Message.PublishTime.AsTime()
 		// Remove messages that have been undelivered for a long time.
 		if !m.outstanding() && now.Sub(pubTime) > retentionDuration {
-			s.publishToDeadLetter(m)
+			if s.proto.DeadLetterPolicy != nil {
```
hongalex, line 1136: I think instead of putting this here, we should place this check inside `publishToDeadLetter` since it's used in other places.

hongalex: Actually on second thought, the other places we call `publishToDeadLetter` is guarded by a `deadLetterCandidate` call. I'm also not entirely sure if placing messages that have expired passed retention duration into the dead letter queue is the behavior the live service uses. Let me double check

hongalex: Yeah after some follow up, this behavior is not consistent with the live service and is a bug.

My recommendation is that we should remove this call. This will be a breaking change but a minor one. Since this package is experimental it should be ok as a one off change. If you aren't ok with doing this, you can just keep this PR as parts that refactor the time changes (thanks for doing that btw) and I'll create a separate PR to remove this call.

adg: Yeah that was my rationale for not originally putting the check in `deadLetterCandidate`.

adg: OK I amended my commit to just not deliver to dead letter on expiry.

hongalex (approved):

Issue #7387, opened 2023-02-08
longrunning: Poll should return full Operation.error


noahdietz:
The `longrunning` Poll method [has been overwriting](https://github.com/googleapis/google-cloud-go/blob/6ec787fb392cd3c82a3ce608489e4d6e358eccbc/longrunning/longrunning.go#L108-L109) the `Operation.error` with just the status code from the original. This is not acceptable as the google.rpc.Status contains valuable error details in the `details` field. 

Instead, the full `googl.rpc.Status` must be returned, unmodified OR an `apierror.APIError` ([ref](https://pkg.go.dev/github.com/googleapis/gax-go/v2/apierror#APIError)) that has the full parsed error details and is backwards compat.

PR #7392, opened 2023-02-09
fix(longrunning): properly parse errors with apierror

codyoss:
Fixes: #7387


noahdietz (approved):
Thank you for fixing this!

PR #7418, opened 2023-02-13
feat(bigquery): add support for session in load jobs

alvarowolfx:
Towards internal b/268630262


Neenu1995 (approved):

shollyman (approved):
PR looks good, but please take a quick pass through our various doc comments to ensure we're adequately capturing the new usage of sessions and connection properties.

For example, https://github.com/googleapis/google-cloud-go/blob/main/bigquery/query.go#L490 describes ConnectionProperty only in the context of a query request, and this changes it to be a load or query.

shollyman (approved):

bigquery/integration_test.go
```diff
@@ -1421,7 +1421,88 @@ func TestIntegration_Load(t *testing.T) {
 		t.Fatal(err)
 	}
 	checkReadAndTotalRows(t, "reader load", table.Read(ctx), wantRows)
+}
+
+func TestIngeration_LoadWithSessionSupport(t *testing.T) {
```
shollyman, line 1426: s/TestIngeration/TestIntegration/

alvarowolfx: oops, thanks for catching that

PR #7420, opened 2023-02-14
chore(storage): adjust GRPC reader remain - off by one

BrennaEpp:
Also moves the params.length code to be together.


storage/grpc_client.go
```diff
@@ -963,7 +963,7 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
 	cr := msg.GetContentRange()
 	if cr != nil {
 		r.Attrs.StartOffset = cr.GetStart()
-		r.remain = cr.GetEnd() - cr.GetStart() + 1
```
tritone, line 966: Should or do we have a test checking this behavior?

tritone (approved):

PR #7430, opened 2023-02-15
chore(storage): add range reads to benchmarks

BrennaEpp:


storage/internal/benchmarks/README.md
```diff
@@ -24,6 +24,9 @@ This runs 1000 iterations on 512kib to 2Gib files in the background, sending out
 | -gc_f | whether to force garbage collection <br> before every write or read benchmark |  `true` or `false` (present/not present) | `false` |
 | -min_size | minimum object size in bytes | any positive integer | `512` |
 | -max_size | maximum object size in bytes | any positive integer | `2 097 152` (2 GiB) |
+| -range_size | size of the range to read in bytes | any positive integer | `0` (<=0 reads the full object) |
```
tritone, line 27: Wondering if we should have a max/min range size rather than a fixed size in order to allow a mix of sizes, as we do for full objects. What do you think?

storage/internal/benchmarks/README.md
```diff
@@ -24,6 +24,9 @@ This runs 1000 iterations on 512kib to 2Gib files in the background, sending out
...
+| -min_read_offset | minimum read offset for range reads in bytes | any integer >0 | `0` |
```
tritone, line 28: Maybe clarify that the offset is for the start of the range to be read.

tritone (commented):
One API question, otherwise looks good!

storage/internal/benchmarks/README.md
BrennaEpp: Perhaps, but right now the workload we are looking at uses specific sizes. I'm wary of adding more parameters that won't be used. Is there a specific experiment you'd like to see with varying sizes?

tritone (approved):

PR #7455, opened 2023-02-17
chore(internal/gapicgen): fix dockerfile build

codyoss:
With the release of 1.20 golang no longer ships a bunch of stuff pre-compiled, see: https://tip.golang.org/doc/go1.20

This seems to mean non-glibc distros like alpine are harder to use Go with. From what I am seeing you basically need to build from source. That is a decent number of steps to get all the deps right so lets copy the go binary from the offical image for now. Was getting the error: Error loading shared library libresolv.so.2


quartzmo (approved):
Bravo.

Issue #7380, opened 2023-02-08
bigquery/storage/managedwriter: EnableWriteRetries seems to lead to panics after ManagedStream context cancellation


zapo:
**Client**

Bigquery

**Environment**

GKE (also reproducible locally)

**Go Environment**

$ go version
go version go1.19 linux/amd64

**Code**

This simulates how we use the managedwriter with concurrent streams for the same client.

```go
package main

import (
	"context"
	"log"
	"sync"
	"test/pb"

	"cloud.google.com/go/bigquery/storage/managedwriter"
	"google.golang.org/protobuf/proto"
)

func main() {
	ctx := context.Background()
	client, err := managedwriter.NewClient(ctx, "<redacted>")
	if err != nil {
		log.Fatal(err)
	}

	var wg sync.WaitGroup

	for i := 0; i < 2; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()

			ctx, cancel := context.WithCancel(ctx)
			defer cancel()

			stream, err := client.NewManagedStream(ctx,
				managedwriter.WithType(managedwriter.DefaultStream),
				managedwriter.WithDestinationTable("<redacted>"),
				managedwriter.EnableWriteRetries(true),
			)
			if err != nil {
				log.Fatal(err)
			}

			for i := 0; i < 10000; i++ {
				rec := &pb.TestRecord{}
				buf, _ := proto.Marshal(rec)

				if _, err := stream.AppendRows(ctx, [][]byte{buf}); err != nil {
					log.Fatalf("append row: %v", err)
				}
			}

			if err := stream.Close(); err != nil {
				log.Fatalf("stream close: %v", err)
			}
		}()
	}
	wg.Wait()

	if err := client.Close(); err != nil {
		log.Fatalf("client close: %v", err)
	}
}

```

**Expected behavior**

Does not panic

**Actual behavior**

Panics with: `panic: close of closed channel` when using `EnableWriteRetries(true)`. 
However this seems related to cancellation of context passed to NewManagedStream and concurrency.

**Additional context**

```
panic: close of closed channel
goroutine 10580 [running]:
[cloud.google.com/go/bigquery/storage/managedwriter.(*pendingWrite).markDone(0xc047e496c0](http://cloud.google.com/go/bigquery/storage/managedwriter.(*pendingWrite).markDone(0xc047e496c0), 0x2fd4a60?, {0x2fd4a60?, 0x45024e0?}, 0xc049f6af30)
[...]

PR #7467, opened 2023-02-21
fix(bigquery): avoid double-channel-close during context cancellation

shollyman:
EnableWriteRetries exposes an issue with finalizing a pending write.
The appendWithRetry function incorrectly marks a pending write done
in certain situations (context cancellations), when responsibility
is held by it's callers.

PR adds fix and a test that exercises the failure condition.

Fixes: https://github.com/googleapis/google-cloud-go/issues/7380


alvarowolfx (approved):
LGTM.

PR #7486, opened 2023-02-23
refactor(bigquery/storage/managedwriter): make send optimizers safer

shollyman:
Make the send optimizers safer to work with arbitrary requests that may not be fully populated.  This largely affects test code as the normal useage of the library prevents partial constructions of "real" requests, but is added to avoid possible panics.


bigquery/storage/managedwriter/send_optimizer.go
```diff
@@ -104,32 +107,47 @@ func (mo *multiplexOptimizer) optimizeSend(arc storagepb.BigQueryWrite_AppendRow
...
+				if pr := cp.GetProtoRows(); pr != nil {
+					pr.WriterSchema = nil
+				}
 			}
 		}
-		resp = arc.Send(cp)
-		if resp == nil && swapOnSuccess {
+		err = arc.Send(cp)
+		if err == nil && swapOnSuccess {
 			cp.GetProtoRows().Rows = nil
```
alvarowolfx, line 121: might be good to add a check to see if cp.GetProtoRows() returns nil like was added on other cases

shollyman: Good spot, updating.

alvarowolfx (approved):

PR #7490, opened 2023-02-23
refactor(bigquery/storage/managedwriter): add custom client options

shollyman:
This PR revisits the expected behavior for config knobs in the client. Previously, all configuration was done when instantiating a writer (aka a ManagedStream).  There are some chicken-and-egg problems related to multiplex settings, as connection options are decoupled from individual writers.

This PR adds the following unexported custom client options (but does not yet use them for anything):

* enableMultiplex
* defaultInflightRequests
* defaultInflightBytes
* defaultAppendRowsCallOption

This PR also removes the still-unexported enableMultiplex from the set of defined WriterOption options which can be passed when instantiating individual writes.

Additionally, this refactor includes a correctness fix for the traceID option that was causing the traceID to duplicate the initial token.

### DIFF
--- a/bigquery/storage/managedwriter/options.go
+++ b/bigquery/storage/managedwriter/options.go
@@ -162,7 +162,7 @@ func WithMaxInflightBytes(n int) WriterOption {
 // This is generally for diagnostic purposes only.
 func WithTraceID(traceID string) WriterOption {
 	return func(ms *ManagedStream) {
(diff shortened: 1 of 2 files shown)


shollyman:
build failure is unrelated, something about documentai samples appears be having issues?

alvarowolfx (approved):

PR #7500, opened 2023-03-01
feat(spanner): add x-goog-spanner-route-to-leader header to Spanner RPC contexts for RW/PDML transactions.

rahul2393:
The header is added to support leader-aware-routing feature, which aims at reducing cross-regional latency for RW/PDML transactions in a multi-region instance.


spanner/pdml.go
```diff
@@ -105,7 +105,7 @@ func (c *Client) partitionedUpdate(ctx context.Context, statement Statement, opt
 func executePdml(ctx context.Context, sh *sessionHandle, req *sppb.ExecuteSqlRequest) (count int64, err error) {
 	var md metadata.MD
 	// Begin transaction.
-	res, err := sh.getClient().BeginTransaction(contextWithOutgoingMetadata(ctx, sh.getMetadata()), &sppb.BeginTransactionRequest{
+	res, err := sh.getClient().BeginTransaction(ctx, &sppb.BeginTransactionRequest{
```
olavloite, line 108: Why are we no longer calling `contextWithOutgoingMetadata` here?

spanner/transaction.go
```diff
@@ -1676,7 +1681,7 @@ func (t *writeOnlyTransaction) applyAtLeastOnce(ctx context.Context, ms ...*Muta
 					return ToSpannerError(err)
 				}
 			}
-			res, err := sh.getClient().Commit(contextWithOutgoingMetadata(ctx, sh.getMetadata()), &sppb.CommitRequest{
+			res, err := sh.getClient().Commit(contextWithOutgoingMetadata(ctx, sh.getMetadata(), false), &sppb.CommitRequest{
```
olavloite, line 1684: Why is this never using routeToLeader? Would it not make sense to make this use the client setting, as it is a read/write transaction?

spanner/client.go
```diff
@@ -147,24 +152,33 @@ type ClientConfig struct {
 	// database by this client.
 	DatabaseRole string
 
+	// RouteToLeader specifies if all the requests of type read-write and PDML
+	// need to be routed to the leader region.
+	//
+	// Default: true
+	RouteToLeader bool
```
olavloite, line 159: I think that we need to invert this to `DisableRouteToLeader` with a default false value. The problem now is that anyone who is currently using `NewClientWithConfig(ClientConfig{ .... })` will create a config with `RouteToLeader=false`, unless they change their code after this change. We want this option to be enabled by default, but in the current setup, it won't be for anyone who does not use the config-less `NewClient(...)` method.

olavloite (changes requested):

spanner/pdml.go
rahul2393: executePdml is called from partitionedUpdate method where create and pass sessionHandle and request, so it makes sense to set the headers there only, after this change context with metadata is set the parent function itself, it will also save computing `contextWithOutgoingMetadata` multiple times.

olavloite: Ah, makes sense.

spanner/transaction.go
```diff
@@ -1676,7 +1681,7 @@ func (t *writeOnlyTransaction) applyAtLeastOnce(ctx context.Context, ms ...*Muta
 					return ToSpannerError(err)
 				}
 			}
-			res, err := sh.getClient().Commit(contextWithOutgoingMetadata(ctx, sh.getMetadata()), &sppb.CommitRequest{
+			res, err := sh.getClient().Commit(contextWithOutgoingMetadata(ctx, sh.getMetadata(), true), &sppb.CommitRequest{
```
olavloite, line 1684: Shouldn't this just use the client setting for `disableRouteToLeader`?

If not: It is a read/write transaction, so shouldn't the default then be `false`?

rahul2393: Thanks, fixed

olavloite (approved):

Issue #7496, opened 2023-02-27
spanner: Deadlock in Spanner client library


osterante:
**Client**

Spanner v1.44.0

**Environment**

The local environment on M1 MacBookPro

**Go Environment**
$ go version
go version go1.20 darwin/arm64

<details>
<summary>$ go env</summary>

```
GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/Users/UserName/Library/Caches/go-build"
GOENV="/Users/UserName/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/UserName/go/pkg/mod"
GOOS="darwin"
GOPATH="/Users/UserName/go"
GOPROXY="https://proxy.golang.org/,direct"
GOROOT="/Users/UserName/.anyenv/envs/goenv/versions/1.20.0"
GOSUMDB="[sum.golang.org](http://sum.golang.org/)"
GOTMPDIR=""
GOTOOLDIR="/Users/UserName/.anyenv/envs/goenv/versions/1.20.0/pkg/tool/darwin_arm64"
GOVCS=""
GOVERSION="go1.20"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/UserName/Projects/Project/go.mod"
GOWORK=""
CGO_CFLAGS="-O2 -g"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-O2 -g"
CGO_FFLAGS="-O2 -g"
CGO_LDFLAGS="-O2 -g"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/x3/0yyktc4d5_11gvj8gsjxsck5mc3vtb/T/go-build887811120=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

</details>

We found a deadlock in the Spanner library on our e2e test.
Seems to try to acquire multiple locks at https://github.com/googleapis/google-cloud-go/blob/main/spanner/session.go#L322 and https://github.com/googleapis/google-cloud-go/blob/main/spanner/session.go#L961.

The stack trace is following:

```
[...]

PR #7501, opened 2023-03-01
fix(spanner): possible deadlock when deleting sessions from different places

rahul2393:
fix: https://github.com/googleapis/google-cloud-go/issues/7496


harshachinta (approved):
LGTM

PR #7503, opened 2023-03-01
feat(alloydb): start generating apiv1, apiv1beta, apiv1alpha

quartzmo:


internal/gapicgen/generator/config.go
```diff
@@ -2006,6 +2006,34 @@ var MicrogenGapicConfigs = []*MicrogenConfig{
 		// GA after 2023/04/01
 		ReleaseLevel: "beta",
 	},
+	{
+		InputDirectoryPath:    "google/cloud/alloydb/v1",
+		stopGeneration:        true,
```
codyoss, line 2011: I think I missed this on your previous PR, but this should only be flipped once migrated to owlbot.

codyoss (approved):

PR #7508, opened 2023-03-02
refactor: augment poolRouter interface (and simpleRouter)

shollyman:
This PR extends the poolRouter interface to include more lifecycle signalling:
* when a router takes control of it's connectionPool
* when a connection pool is going away
* when a writer is attached to the pool
* when a writer is detached from the pool

The core problem this addresses is proper shutdown between the components.  We need to propagate this information to shutdown cleanly so that connections can go away in response to their writers.

This PR also augments our existing simpleRouter to adhere to the enlarged interface contract.  The pool attach/detach events are left unimplemented as we need to plumb this in when we wire up the abstractions end to end, which will happen when we rewire the internals to use the new abstractions.

A test has been added to assert simpleRouter state when writers are attached and removed.

### DIFF
--- a/bigquery/storage/managedwriter/connection.go
+++ b/bigquery/storage/managedwriter/connection.go
@@ -74,14 +74,14 @@ func (pool *connectionPool) processWrite(pw *pendingWrite) error {
 
 func (pool *connectionPool) addWriter(writer *ManagedStream) error {
 	if pool.router != nil {
-		return pool.router.attachWriter(writer)
+		return pool.router.writerAttach(writer)
 	}
 	return nil
 }
 
-func (pool *connectionPool) disconnectWriter(writer *ManagedStream) error {
+func (pool *connectionPool) removeWriter(writer *ManagedStream) error {
 	if pool.router != nil {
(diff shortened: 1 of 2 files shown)


bigquery/storage/managedwriter/connection.go
```diff
@@ -71,6 +72,20 @@ func (pool *connectionPool) processWrite(pw *pendingWrite) error {
 	return conn.appendWithRetry(pw)
 }
 
+func (pool *connectionPool) addWriter(writer *ManagedStream) error {
+	if pool.router != nil {
+		return pool.router.writerAttach(writer)
+	}
+	return nil
```
alvarowolfx, line 79: when there is no router available, shouldn't this return an error ?

alvarowolfx (approved):
Looks good to me, I added a minor comment on the `addWriter`. Not sure if there are cases where the router is not going to be set.

bigquery/storage/managedwriter/connection.go
shollyman: good point, updating

PR #7510, opened 2023-03-03
feat(bigquery): add support for storage billing model

### LINKED ISSUE #6978: bigquery: Dataset storage billing model

We want to create BQ datasets and set the [storage billing model](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#storagebillingmodel) property to `PHYSICAL`, but it doesn't look like this option is available using the latest release.

We would like a `StorageBillingModel` field to be added to the `bigquery.DatasetMetadata` struct so we can create datasets with Physical billing storage model.

We are currently using this SDK and would rather not resort to calling the [BQ REST API](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/insert) by ourselves...

alvarowolfx:
Resolves #6978


bigquery/dataset.go
```diff
@@ -117,6 +134,12 @@ type DatasetMetadataToUpdate struct {
 	// created in the dataset.
 	DefaultCollation optional.String
 
+	// Storage billing model to be used for all tables in the dataset.
+	// Can be set to PHYSICAL. Default is LOGICAL.
+	// Once you change a dataset's storage billing model to use physical bytes, you can't change it back to using logical bytes again.
+	// More details: https://cloud.google.com/bigquery/docs/datasets-intro#dataset_storage_billing_models
+	StorageBillingModel StorageBillingModel
```
shollyman, line 141: Should this be an optional.String?

bigquery/dataset.go
```diff
@@ -298,6 +323,13 @@ func bqToDatasetMetadata(d *bq.Dataset, c *Client) (*DatasetMetadata, error) {
 	return dm, nil
 }
 
+func bqToStorageBillingModel(storageBillingModel string) StorageBillingModel {
```
shollyman, line 326: I worry that this kind of thing is brittle if the service introduces another billing model.

shollyman (approved):

PR #7516, opened 2023-03-06
testing(bigquery/storage/managedwriter): add basic CDC example

shollyman:


bigquery/storage/managedwriter/integration_test.go
```diff
@@ -504,6 +509,144 @@ func testCommittedStream(ctx context.Context, t *testing.T, mwClient *Client, bq
...
 
+// testSimpleCDC demonstrates basic Change Data Capture (CDC) functionality.   We add an initial set of
+// rows to a table, then use CDC to apply updates.
+func testSimpleCDC(ctx context.Context, t *testing.T, mwClient *Client, bqClient *bigquery.Client, dataset *bigquery.Dataset) {
+	testTable := dataset.Table(tableIDs.New())
+
+	if err := testTable.Create(ctx, &bigquery.TableMetadata{
+		Schema: testdata.ExampleEmployeeSchema,
+		Clustering: &bigquery.Clustering{
+			Fields: []string{"id"},
```
alvarowolfx, line 520: Why is this needed for CDC ?

alvarowolfx (approved):
Added a minor question on the table clustering. The rest looks good to me, this is a great sample, I didn't know before joining here that BQ supported CDC and this is a good example on how to use it. I think we should add this as an actual sample to the website and have it on other languages.

bigquery/storage/managedwriter/integration_test.go
shollyman: CDC functional requires the table to have a primary key, and clustering setup for it to reduce scans.

Without it, you get an error of the flavor `The primary keys and clustering keys of the table project.dataset.table are required to create an upsert stream.`

PR #7532, opened 2023-03-09
refactor: switch from syntax to presence checks

shollyman:
Added in response to internal cl/514925348, which is trying to deprecate syntax version checking in favor of capability checks.


alvarowolfx (approved):
LGTM

PR #7543, opened 2023-03-13
fix(storage): SignedURL v4 allows headers with colons in value

BrennaEpp:
Previously, this:

```go
opts := &storage.SignedURLOptions{  
       Scheme:  storage.SigningSchemeV4,  
       Headers: []string{  
            "x-goog-meta-start-time:2023-02-10T03:",   
        },  
        ...
}
url, err := client.Bucket(bucket).SignedURL(object, opts)  
```
would return a URL that would return a 403 (signature mismatch) with the header `x-goog-meta-start-time:2023-02-10T03:`;
you would have to specify the header as `x-goog-meta-start-time:2023-02-10T03` to get a 200 with the provided URL.

Based on [our public docs](https://cloud.google.com/storage/docs/xml-api/reference-headers#xgoogmeta), values should be able to use any US-ASCII characters, which includes colons.

This PR fixes the splitting of the headers so that colons (and anything after them) are preserved in the header values.


tritone (approved):
Nice fix. Can you look into adding a test case for this to the cross-lang signed URL conformance tests as well? https://github.com/googleapis/conformance-tests/blob/main/storage/v1/v4_signatures.json

PR #7552, opened 2023-03-15
chore: go mod tidy

codyoss:
Follow up to #7542, it seems CI did not catch these changes were needed. Need to investigate this more.


codyoss:
Force merging to fix build, all CI passed

PR #7568, opened 2023-03-16
fix(pubsub): check response of receipt modacks for exactly once delivery

### LINKED ISSUE #7585: pubsub: exactly-once delivery should wait for the receipt modAck to succeed before proceeding
The current behavior is that the receipt modAck is "fire and forget", much like non-EOD receipt modAcks.

However, any delivery to the client logic should block on a successful receipt modAck. If the receipt modAck fails with a permanent error, we should not send the message to the client and not even log it. If it's a temporary failure, it should be retried.

hongalex:
This PR aligns the behavior of receipt modacks with the Python client. Specifically, if exactly once delivery is enabled, receipt modacks that fail should be retried up to 3 times (similar to lease modacks), but message processing should be blocked on the result of this modack.

This PR also removes a test with the fake since `DeadlineExceeded` errors will never happen in the `Nack` case if modacks are never successful.

Fixes #7585


pubsub/iterator.go
```diff
@@ -264,22 +274,52 @@ func (it *messageIterator) receive(maxToPull int32) ([]*Message, error) {
...
+		// If exactly once is enabled, we should wait until modack responses are successes
+		// before attempting to process messages.
+		it.sendModAck(ackIDs, deadline, false)
+		for ackID, ar := range ackIDs {
+			ctx := context.Background()
+			_, err := ar.Get(ctx)
+			if err != nil {
+				delete(pendingMessages, ackID)
+				it.mu.Lock()
+				defer it.mu.Unlock()
```
shollyman, line 310: This defer seems problematic for unlocking.  What happens when a second entry in this for loop has an error?

hongalex: good catch, removed the use of defer here

PR #7583, opened 2023-03-17
fix(storage): retract versions with Copier bug

tritone:
Retract storage releases affected by #6857. This will prevent the Go tool from selecting these versions using MVS. Builds where these releases are currently selected will still work, but the user will receive a warning.


codyoss (approved):

BrennaEpp (approved):
Thanks!

PR #7589, opened 2023-03-19
feat(datastore): EntityFilter for AND/OR queries

bhshkh:
Adding new method 'FilterEntity' to Query which can be used to construct composite queries.
Adding the following new structs to be used with FilterEntity
- AND
- OR
- PropertyFilter


datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
 	Direction sortDirection
 }
 
+// EntityFilter represents a datastore filter
+type EntityFilter interface {
+	toEntityFilter() (EntityFilter, error)
+	toProto() (*pb.Filter, error)
+}
+
+// PropertyFilter represent field based filter
```
telpirion, line 101: quibble: should be "represent_s_"

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+}
+
+func (pf PropertyFilter) toProto() (*pb.Filter, error) {
+
+	if pf.FieldName == "" {
+		return nil, errors.New("datastore: empty query filter field name")
+	}
+	v, err := interfaceToProto(reflect.ValueOf(pf.Value).Interface(), false)
+	if err != nil {
+		return nil, fmt.Errorf("datastore: bad query filter value type: %v", err)
```
telpirion, line 122: suggestion: use the %w format so as to expose the wrapped error message.

datastore/integration_test.go
```diff
@@ -484,6 +484,57 @@ func testSmallQueries(ctx context.Context, t *testing.T, client *Client, parent
 		x()
 	}
 }
+func TestIntegration_FilterEntity(t *testing.T) {
+	ctx := context.Background()
+	client := newTestClient(ctx, t)
+	defer client.Close()
+
+	parent := NameKey("SQParent", "TestIntegration_Filters"+suffix, nil)
```
telpirion, line 492: quibble: this being a test, you can be even more terse in your variable naming. Up to you!

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+	return pf, nil
+}
+
+// CompositeFilter represents datastore composite filters
+type CompositeFilter interface {
+	EntityFilter
+	isCompositeFilter()
+}
+
+// OR represents OR'ed filters
```
telpirion, line 160: quibble: I wouldn't use OR as a verb here. 

Suggestion:

```
// OR represents an union of two or more filters
```

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+type OR struct {
+	Filters []EntityFilter
+}
+
+func (OR) isCompositeFilter() {}
```
telpirion, line 165: suggestion: it looks like this method is only used to satisfy the CompositeFilter interface above. Same with AND below. Seems to me that you should be able to have both OR and AND structs fulfill the same interface without creating a dummy method. Try to remove `isCompositeFilter()`.

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+		if err != nil {
+			return nil, err
+		}
+		validFilters[i] = validFilter
+	}
+	or.Filters = validFilters
+	return or, nil
+}
+
+// AND represents AND'ed filters
```
telpirion, line 197: quibble: Don't use `AND` as a verb.

Suggestion:

```
// AND represents the intersection of two or more filters.
```

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+type AND struct {
+	Filters []EntityFilter
+}
+
+func (AND) isCompositeFilter() {}
+
+func (and AND) toProto() (*pb.Filter, error) {
+
+	pbFilters := make([]*pb.Filter, len(and.Filters))
```
telpirion, line 206: See previous about declaring slices.

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+// OR represents OR'ed filters
+type OR struct {
+	Filters []EntityFilter
+}
+
+func (OR) isCompositeFilter() {}
+
+func (or OR) toProto() (*pb.Filter, error) {
+
+	pbFilters := make([]*pb.Filter, len(or.Filters))
```
telpirion, line 169: suggestion: I think you can tighten up this code a bit, make it less dependent on getting the length of the filters.

```
var pbFilters []*pb.Filter
for _, filter := range or.Filters {
    pbFilter, err := filter.toProto()
   if err != nil {
        return nil, err
   }
   pbFilters = append(pbFilters, pbFilter)
}

```

telpirion: Here and below.

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+		pbFilters[i] = pbFilter
+	}
+	return &pb.Filter{FilterType: &pb.Filter_CompositeFilter{CompositeFilter: &pb.CompositeFilter{
+		Op:      pb.CompositeFilter_OR,
+		Filters: pbFilters,
+	}}}, nil
+}
+
+func (or OR) toEntityFilter() (EntityFilter, error) {
+	validFilters := make([]EntityFilter, len(or.Filters))
```
telpirion, line 185: See previous about declaring slices.

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+		pbFilters[i] = pbFilter
+	}
+	return &pb.Filter{FilterType: &pb.Filter_CompositeFilter{CompositeFilter: &pb.CompositeFilter{
+		Op:      pb.CompositeFilter_AND,
+		Filters: pbFilters,
+	}}}, nil
+}
+
+func (and AND) toEntityFilter() (EntityFilter, error) {
+	validFilters := make([]EntityFilter, len(and.Filters))
```
telpirion, line 222: See previous about declaring slices.

datastore/query.go
```diff
@@ -176,6 +326,22 @@ func (q *Query) Transaction(t *Transaction) *Query {
...
 }
 
+// FilterEntity returns a query with provided filter.
+//
+// Filter can be a single field comparison or a composite filter
+// AND and OR are supported composite filters
+// Filters in multiple calls are AND'ed together
+func (q *Query) FilterEntity(entityFilter EntityFilter) *Query {
+	q = q.clone()
+	validEntityFilter, err := entityFilter.toEntityFilter()
```
telpirion, line 336: quibble: I would shorten `validEntityFilter`. This is a pretty long name for a variable that isn't used outside of this method. I recommend using something like `vf`.

datastore/query.go
```diff
@@ -411,25 +546,11 @@ func (q *Query) toProto() (*pb.Query, error) {
 	}
 	var filters []*pb.Filter
 	for _, qf := range q.filter {
-		if qf.FieldName == "" {
-			return nil, errors.New("datastore: empty query filter field name")
-		}
-		v, err := interfaceToProto(reflect.ValueOf(qf.Value).Interface(), false)
+		pbFilter, err := qf.toProto()
```
telpirion, line 549: praise: Nice refactor here!

datastore/query_test.go
```diff
@@ -500,6 +500,77 @@ var (
 	}
 )
 
+func TestPropertyFilterToProto(t *testing.T) {
+	testCases := []struct {
```
telpirion, line 504: suggestion: It looks like there's only one valid test case that you're testing (all other testCases focus on error states). I might add a couple more test cases. For example, it looks like you allow developers to pass in field names that include whitespaces--you should probably add a testCase for that.

datastore/query_test.go
```diff
@@ -500,6 +500,77 @@ var (
...
+		}
+	}
+}
+
+func TestCompositeFilterToProto(t *testing.T) {
+	testCases := []struct {
+		cf         CompositeFilter
+		wantErrMsg string
+	}{
+		{&AND{[]EntityFilter{&PropertyFilter{"x", "=", 4}}}, ""},
```
telpirion, line 532: Same as above--consider whether there are other valid EntityFilters that you should have testCases for.

datastore/query.go
telpirion: Somehow my comments are out-of-sync ☹️ . See my comment on line 169.

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+
+	for i, filter := range or.Filters {
+		pbFilter, err := filter.toProto()
+		if err != nil {
+			return nil, err
+		}
+		pbFilters[i] = pbFilter
+	}
+	return &pb.Filter{FilterType: &pb.Filter_CompositeFilter{CompositeFilter: &pb.CompositeFilter{
+		Op:      pb.CompositeFilter_OR,
```
danieljbruce, line 179: Is this common naming convention in golang?

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+// EntityFilter represents a datastore filter
+type EntityFilter interface {
+	toEntityFilter() (EntityFilter, error)
+	toProto() (*pb.Filter, error)
+}
+
+// PropertyFilter represent field based filter
+//
+// The operator parameter takes the following strings: ">", "<", ">=", "<=",
+// "=", "!=", "in", and "not-in".
```
danieljbruce, line 104: Is it worth mentioning `operatorToString` in this comment here?

danieljbruce (commented):
Looks good!

datastore/integration_test.go
```diff
@@ -484,6 +484,57 @@ func testSmallQueries(ctx context.Context, t *testing.T, client *Client, parent
...
+	baseQuery := NewQuery("SQChild").Ancestor(parent)
+	testSmallQueries(ctx, t, client, parent, children, []SQTestCase{
+		{
+			"I>1",
+			baseQuery.Filter("T=", now).FilterEntity(PropertyFilter{"I", ">", 1}),
+			5,
+			3 + 4 + 5 + 6 + 7 + 96 + 95 + 94 + 93 + 92,
+		},
+		{
+			"(T = now) and (((J > 97) and (T = tomorrow)) or (J < 94))",
```
enocom, line 514: Should we be adding a test case or two where the logic is just a simple OR perhaps?

datastore/query.go
enocom: Not really -- but there are all sorts of unusual naming conventions in protobufs.

datastore/query.go
enocom: "a" union? ;-)

datastore/query_test.go
enocom, line 532: Let's always use field names in structs here and below.

datastore/query_test.go
```diff
@@ -500,6 +500,77 @@ var (
...
+		{&OR{[]EntityFilter{&PropertyFilter{"x", "==", 4}}}, "datastore: unknown query filter operator"},
+	}
+	for _, tc := range testCases {
+		_, gotErr := tc.cf.toProto()
+		gotErrMsg := ""
+		if gotErr != nil {
+			gotErrMsg = gotErr.Error()
+		}
+
+		if gotErrMsg != tc.wantErrMsg {
```
enocom, line 546: Do we want to compare error messages or just confirm an error occurred?

datastore/query.go
```diff
@@ -45,22 +45,33 @@ const (
...
+	equal.String():       pb.PropertyFilter_EQUAL,
+	greaterEq.String():   pb.PropertyFilter_GREATER_THAN_OR_EQUAL,
+	greaterThan.String(): pb.PropertyFilter_GREATER_THAN,
+	in.String():          pb.PropertyFilter_IN,
+	notIn.String():       pb.PropertyFilter_NOT_IN,
+	notEqual.String():    pb.PropertyFilter_NOT_EQUAL,
+}
+
+func (op operator) String() string {
+	var operatorToString = map[operator]string{
```
enocom, line 60: This map will be allocated on every call to string. Shall we move it up to package-level scope to avoid this?

datastore/query.go
```diff
@@ -81,6 +92,145 @@ type order struct {
...
+		return nil, fmt.Errorf("datastore: invalid operator %q in filter", pf.Operator)
+	}
+	pf.Operator = op
+
+	unquotedFieldName, err := unquote(pf.FieldName)
+	if err != nil {
+		return nil, fmt.Errorf("datastore: invalid syntax for quoted field name %q", pf.FieldName)
+	}
+	pf.FieldName = unquotedFieldName
+	return pf, nil
```
enocom, line 151: Since this method is constructing a new PropertyFilter, we should make that clear especially since this is a value receiver (and not a pointer receiver). Maybe:

``` golang
return PropertyFilter{Operator: op, FieldName: unquotedFieldName, Value: pf.Value}, nil
```

datastore/query.go
bhshkh: Done

datastore/query.go
bhshkh: Done

datastore/query.go
bhshkh: On removing, I see error:
`
cannot use &OR{…} (value of type *OR) as CompositeFilter value in struct literal: *OR does not implement CompositeFilter (missing method isCompositeFilter)
`

datastore/integration_test.go
bhshkh: Done

datastore/query_test.go
bhshkh: Done

datastore/query.go
bhshkh: Moved

datastore/query.go
bhshkh: Done

datastore/integration_test.go
```diff
@@ -484,6 +484,57 @@ func testSmallQueries(ctx context.Context, t *testing.T, client *Client, parent
...
+		{I: 3, J: 96, T: now, U: now},
+		{I: 4, J: 95, T: now, U: now},
+		{I: 5, J: 94, T: now, U: now},
+		{I: 6, J: 93, T: now, U: now},
+		{I: 7, J: 92, T: now, U: now},
+	}
+	baseQuery := NewQuery("SQChild").Ancestor(parent)
+	testSmallQueries(ctx, t, client, parent, children, []SQTestCase{
+		{
+			"I>1",
```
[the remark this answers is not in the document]
bhshkh: Done

datastore/integration_test.go
```diff
@@ -484,6 +484,57 @@ func testSmallQueries(ctx context.Context, t *testing.T, client *Client, parent
...
 }
+func TestIntegration_FilterEntity(t *testing.T) {
+	ctx := context.Background()
+	client := newTestClient(ctx, t)
+	defer client.Close()
+
+	parent := NameKey("SQParent", "TestIntegration_Filters"+suffix, nil)
+	now := timeNow.Truncate(time.Millisecond).Unix()
+	tomorrow := timeNow.Truncate(time.Millisecond).AddDate(0, 0, 1).Unix()
+	children := []*SQChild{
```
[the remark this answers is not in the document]
bhshkh: Not needed. Here, SQChild is the kind of entity being used for testing. Any such structs can be used to work with the client

datastore/integration_test.go
```diff
@@ -484,6 +484,57 @@ func testSmallQueries(ctx context.Context, t *testing.T, client *Client, parent
 		x()
 	}
 }
+func TestIntegration_FilterEntity(t *testing.T) {
```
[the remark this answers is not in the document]
bhshkh: Done. Had missed running gofmt

datastore/query_test.go
bhshkh: Done. Added all the testcases that were written for FilterField

datastore/query_test.go
bhshkh: Done.

datastore/query.go
bhshkh: Done

datastore/query.go
bhshkh: Done

datastore/query.go
```diff
@@ -176,6 +326,22 @@ func (q *Query) Transaction(t *Transaction) *Query {
...
+// AND and OR are supported composite filters
+// Filters in multiple calls are AND'ed together
+func (q *Query) FilterEntity(entityFilter EntityFilter) *Query {
+	q = q.clone()
+	validEntityFilter, err := entityFilter.toEntityFilter()
+	if err != nil {
+		q.err = err
+		return q
+	}
+	q.filter = append(q.filter, validEntityFilter)
```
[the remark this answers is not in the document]
bhshkh: Yes. It is not threadsafe, but it is a pre-existing functionality. So users should already be familiar with this behavior.

enocom:
@bhshkh When you're ready for another round of reviews, just "Re-Request Review" on the side panel above and I'll take a look.

datastore/query.go
```diff
@@ -45,22 +45,33 @@ const (
...
-	notEqual:    pb.PropertyFilter_NOT_EQUAL,
-}
-
-// filter is a conditional filter on query results.
-type filter struct {
-	FieldName string
-	Op        operator
-	Value     interface{}
+var operatorToProto = map[string]pb.PropertyFilter_Operator{
+	lessThan.String():    pb.PropertyFilter_LESS_THAN,
```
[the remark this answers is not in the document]
bhshkh: Done

datastore/integration_test.go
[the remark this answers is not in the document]
enocom: Not a blocking concern, but another thing to clean up would be to not export the type in that case. Since this is a test nothing will every use that type outside of this package. But that can be something to address in future passes.

datastore/query.go
```diff
@@ -176,6 +323,22 @@ func (q *Query) Transaction(t *Transaction) *Query {
 	return q
 }
 
+// FilterEntity returns a query with provided filter.
+//
+// Filter can be a single field comparison or a composite filter
+// AND and OR are supported composite filters
+// Filters in multiple calls are joined together by AND
+func (q *Query) FilterEntity(entityFilter EntityFilter) *Query {
```
enocom, line 331: nit: the type already tells us it's a "EntityFilter" so the argument name could just be `e` or `ef`.

datastore/query.go
```diff
@@ -176,6 +323,22 @@ func (q *Query) Transaction(t *Transaction) *Query {
...
+	q = q.clone()
+	vf, err := entityFilter.toEntityFilter()
```
enocom, line 333: nit: The naming seems a little circular here. Design looks good, but I wonder if there's a better name.

enocom (approved):
A couple of minor nits, but big picture looks great.

datastore/integration_test.go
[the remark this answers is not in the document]
bhshkh: Will do in subsequent PRs

datastore/query_test.go
```diff
@@ -482,24 +482,136 @@ var (
...
-		{"x ==", "x", "==", 0, ""},
-		{"x =<", "x", "=<", 0, ""},
-		{"x =>", "x", "=>", 0, ""},
-		{"x !", "x", "!", 0, ""},
-		{"x ", "x", "", 0, ""},
-		{"x", "x", "", 0, ""},
-		{`" x =`, `" x`, "=", 0, ""},
-		{`" x ="`, `" x `, `="`, 0, ""},
-		{"` x \" =", "` x \"", "=", 0, ""},
+		{"x IN", "x", "IN", "", ""},
```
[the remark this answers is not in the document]
bhshkh: Done

datastore/query.go
bhshkh: Changed the name to toValidFilter

datastore/query.go
bhshkh: Done

datastore/query.go
```diff
@@ -81,6 +84,150 @@ type order struct {
 	Direction sortDirection
 }
 
+// EntityFilter represents a datastore filter
```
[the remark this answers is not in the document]
codyoss: This comment applies to all doc comments for newly added code.

datastore/query.go
```diff
@@ -81,6 +84,150 @@ type order struct {
...
+}
+
+// CompositeFilter represents datastore composite filters
+type CompositeFilter interface {
+	EntityFilter
+	isCompositeFilter()
+}
+
+// OR represents an union of two or more filters
+type OR struct {
```
codyoss, line 158: nit: should be Or. But I might alo suggest making the name more expressive like OrFilter?

codyoss: also same for and

datastore/query.go
```diff
@@ -81,6 +84,150 @@ type order struct {
 	Direction sortDirection
 }
 
+// EntityFilter represents a datastore filter
+type EntityFilter interface {
```
codyoss, line 88: Instead of making an interface here I wonder if this should just a a builder type or if the Filter should be something that dynamically manipulates a query and returns a query rather than passing in filters to a method on query.

So instead of
```
datastore.NewQuery("foo").FilterEntity(datastore.AND{
    Filters: []EntityFilter{
				PropertyFilter{FieldName: "x", Operator: "==", Value: 4},
				PropertyFilter{FieldName: "y", Operator: "<", Value: 3},
			},
})
```

Something more like:
```
datastore.NewQuery("foo").AndFilter(
	PropertyFilter{FieldName: "x", Operator: "==", Value: 4},
	PropertyFilter{FieldName: "y", Operator: "<", Value: 3},
})
```

Thoughts? @enocom wdyt? I have not really used this client before so I don't know what the current exp is like. Maybe the current design aligns with patterns already in the client.

codyoss (changes requested):

datastore/query.go
enocom: The fluent interface is certainly nice. How would we support complex queries with a mix of And's and Or's?

Otherwise, I think the way @bhshkh has it here is clean and easy to understand (and write sophisticated queries).

codyoss: > The fluent interface is certainly nice. How would we support complex queries with a mix of And's and Or's?

Can you give an example with the current API, I am not sure I mean by complex. Or maybe a test should be added for something like this. I would think that all the filters though just keep returning the query. But like I said I have not used this client so I trust your judgment here if you think the current approach has better support for the features needing to be supported.

bhshkh: @codyoss 
Sample complex query : (T = now) and (  ( (J > 97) and (T = tomorrow) ) or (J < 94)  )
https://github.com/googleapis/google-cloud-go/blob/f775d45f95fa7a1ac2d89083c7193c990000abdd/datastore/integration_test.go#L528-L545

datastore/query.go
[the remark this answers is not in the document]
bhshkh: Done

datastore/query.go
codyoss: Ah missed this case sorry, this seems as good in that case.

codyoss (approved):

PR #7602, opened 2023-03-20
chore(storage): add idempotency header

tritone:
This additional header will allow idempotent retries on certain requests which are non-idempotent currently.


storage/invoke_test.go
```diff
@@ -186,22 +187,26 @@ func TestInvoke(t *testing.T) {
...
+			wantClientHeaderFormat := "gccl-invocation-id/.{36} gccl-attempt-count/[0-9]+ gl-go/.* gccl/"
+			match, err := regexp.MatchString(wantClientHeaderFormat, gotClientHeader)
 			if err != nil {
 				s.Fatalf("compiling regexp: %v", err)
 			}
 			if !match {
-				s.Errorf("X-Goog-Api-Client header has wrong format\ngot %v\nwant regex matching %v", gotHeader, wantHeaderFormat)
+				s.Errorf("X-Goog-Api-Client header has wrong format\ngot %v\nwant regex matching %v", gotClientHeader, wantClientHeaderFormat)
+			}
+			if gotIdempotencyHeader != initialIdempotencyHeader {
```
noahdietz, line 208: Nit: consider using [`cmp.Diff`](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff) and reporting the diff in the error message.

noahdietz (approved):

storage/invoke_test.go
tritone: Looked into this, but unfortunately it doesn't work because x-goog-api-client also contains library & Go version information. We search for a substring regex for the added components rather than looking for an exact match for the whole thing. I guess we could reconstruct those components as well but it will make the test brittle to non-breaking changes in apiary, which I'd rather not do.

PR #7615, opened 2023-03-22
feat(iam): promote to GA

quartzmo:
Release-As: 1.0.0


internal/gapicgen/generator/config.go
```diff
@@ -1852,8 +1852,7 @@ var MicrogenGapicConfigs = []*MicrogenConfig{
 		Pkg:                  "iam",
 		ImportPath:           "cloud.google.com/go/iam/apiv1",
 		ApiServiceConfigPath: "iam_meta_api.yaml",
-		// GA after 2022/12/21
-		ReleaseLevel:         "beta",
+		ReleaseLevel:         "ga",
```
noahdietz, line 1855: Not sure how necessary this is because `stopGeneration: true` and it is `ga` in the bazel target. But maybe it's good to keep things relatively in sync.

noahdietz (approved):

PR #7621, opened 2023-03-23
chore(storage): update client pool to be static and initialized on start (benchmarking)

BrennaEpp:
Changes the client pool used to be static and created on program start.

This permits the script to observe long-running clients, clients processing excessive bytes, and less variable runs.

* customClient param to be removed/cleanedup in a separate PR


storage/internal/benchmarks/client_pool.go
```diff
@@ -31,130 +30,125 @@ import (
...
+	return p, func() {
+		for _, c := range p.clients {
+			c.Close()
+		}
+	}
 }
 
-// we can share clients as long as the app buffer sizes are constant
-var httpClients, gRPCClients *clientPool
+// Get rotates through clients. This means the work may not be evenly distributed,
```
tritone, line 68: Did we consider tracking how many ops are currently running on a client and implementing a priority queue based on that? Not something we have to do now necessarily but might be interesting if we see evidence of a workload being distributed unevenly.

storage/internal/benchmarks/client_pool.go
```diff
@@ -31,130 +30,125 @@ import (
...
-	New: func() *storage.Client {
-		// For debuggability's sake, these are HTTP
-		clientMu.Lock()
-		client, err := storage.NewClient(context.Background())
-		clientMu.Unlock()
-		if err != nil {
-			log.Fatalf("storage.NewClient: %v", err)
-		}
+	// return client to queue so that it will be used again without blocking
+	p.clientQueue <- client
```
tritone, line 74: Might want to note in a comment that this code is not totally deterministic if called from multiple threads at the same time -- you could have clients swap places in the queue potentially. But, I don't think it's necessary to add a mutex since this is presumably acceptable.

storage/internal/benchmarks/client_pool.go
```diff
@@ -31,130 +30,125 @@ import (
...
-			client, err := initializeHTTPClient(context.Background(), opts.writeBufferSize, opts.readBufferSize, true)
-			if err != nil {
-				log.Fatalf("initializeHTTPClient: %v", err)
-			}
+var httpClients, gRPCClients, nonBenchmarkingClients *clientPool
 
-			return client
-		},
-	}
+func initializeClientPools(ctx context.Context, opts *benchmarkOptions) func() {
```
tritone, line 81: Maybe add a comment to explicitly note that this creates separate pools for HTTP and gRPC if need be

storage/internal/benchmarks/client_pool.go
```diff
@@ -175,8 +169,8 @@ func initializeHTTPClient(ctx context.Context, writeBufferSize, readBufferSize i
 		IdleConnTimeout:       90 * time.Second,
 		TLSHandshakeTimeout:   10 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
-		WriteBufferSize:       writeBufferSize,
-		ReadBufferSize:        readBufferSize,
+		WriteBufferSize:       p.writeBufferSize,
+		ReadBufferSize:        p.readBufferSize,
 	}
```
tritone, line 174: Unrelated to this PR but I realized that the setting of `MaxIdleConnsPerHost` that we do in google-api-go-client is missing here. We should discuss offline.

storage/internal/benchmarks/client_pool.go
```diff
@@ -197,8 +191,8 @@ func initializeHTTPClient(ctx context.Context, writeBufferSize, readBufferSize i
 	return client, err
 }
 
-func initializeGRPCClient(ctx context.Context, writeBufferSize, readBufferSize int, connPoolSize int, useDefaults bool) (*storage.Client, error) {
-	if useDefaults {
+func initializeGRPCClient(ctx context.Context, p clientParams) (*storage.Client, error) {
+	if !p.customClient {
```
tritone, line 195: Also a bit off topic, but the flow of the logic here is kind of confusing. How about something like:
```
var opts []option.ClientOption
if !p.CustomClient {
  // set custom client opts
}
clientMu.Lock()
// set env var and initialize client using opts...
clientMu.Unlock()
```

Same thing for HTTP.

storage/internal/benchmarks/main.go
```diff
@@ -320,9 +325,8 @@ func (br *benchmarkResult) selectParams(opts benchmarkOptions) {
 			if api == grpcAPI {
 				br.params.appBufferSize = 32000 // default for GRPC
 			}
+			return
```
tritone, line 328: What does this change do?

storage/internal/benchmarks/w1r3.go
```diff
@@ -180,15 +171,7 @@ func (r *w1r3) run(ctx context.Context) error {
...
-		client, close, err := getClient(ctx, opts, *r.readResults[i])
-		if err != nil {
-			return fmt.Errorf("getClient: %w", err)
-		}
-		defer func() {
-			if err := close(); err != nil {
-				log.Printf("close client: %v", err)
-			}
-		}()
+		client := getClient(ctx, *r.readResults[i])
```
tritone, line 174: IMO for w1r3 we should probably just use the same client for all steps rather than checking out another. what do you think?

tritone (commented):
Generally looks good, a few comments/questions

storage/internal/benchmarks/client_pool.go
BrennaEpp: Yes, it is something that could be done in the future. 

I considered keeping track of the number of bytes being processed currently on each client and assigning based on that, but I think it may over-complicate the code. 

I did do some basic tests and the clients processed similar (or equal) number of bytes.

storage/internal/benchmarks/client_pool.go
BrennaEpp: I did just this (I think) in my follow-up PR that adds JSON and removes the customClient var. I think we can leave this as-is for this PR and discuss the flow in that one.

storage/internal/benchmarks/client_pool.go
BrennaEpp: I added a comment. I think it is acceptable. If we are concerned with unbalanced workloads, we may implement a priority queue as mentioned in the other comment.

Though, adding a mutex here is probably low-cost?

storage/internal/benchmarks/client_pool.go
tritone: Sounds good, yeah I think we can leave as-is for now

storage/internal/benchmarks/client_pool.go
tritone: Agreed, I think it's fine, just good to have a comment for anyone looking at this code in the future.

storage/internal/benchmarks/client_pool.go
tritone: We should set the default to `100` for sure rather than the net/http default, even if it's not a param for now. But it would be a relevant param for a high throughput workload as well.

storage/internal/benchmarks/client_pool.go
tritone: Okay sounds good!

tritone (approved):
One thing to fix re: MaxIdleConnsPerHost; otherwise LGTM

storage/internal/benchmarks/client_pool.go
BrennaEpp: `100` is actually the http default and is what it's set to here.

tritone: I think you are confused with `MaxIdleConns` maybe?

See https://pkg.go.dev/net/http#DefaultMaxIdleConnsPerHost and https://pkg.go.dev/net/http#Transport.MaxIdleConnsPerHost . It must be set explicitly or else the value will be 2, which is not good for perf under high load

BrennaEpp: Yep, I got confused. Thanks for bringing this up! I set `MaxIdleConnsPerHost` to `100`

PR #7622, opened 2023-03-23
chore(storage): benchmarking:  allow range parameter for object size

BrennaEpp:


storage/internal/benchmarks/README.md
```diff
@@ -19,9 +19,7 @@ This runs 1000 iterations on 512kib to 2Gib files in the background, sending out
 | -samples | number of samples to report | any positive integer | `8000` |
 | -workers | number of goroutines to run at once; set to 1 for no concurrency | any positive integer | `16` |
 | -api | which API to use | `JSON`: use JSON to upload and XML to download <br> `XML`: use JSON to upload and XML to download <br> `GRPC`: use GRPC <br> `Mixed`: select an API at random for each upload/download  <br> `DirectPath`: use GRPC with direct path | `Mixed` |
-| -object_size | object size in bytes | any positive integer | `1 048 576` (1 MiB) |
-| -min_object_size | minimum object size in bytes <br> ignored if object_size is set | any positive integer | `512` |
-| -max_object_size | maximum object size in bytes <br> ignored if object_size is set | any positive integer | `2 147 483 648` (2 GiB) |
+| -object_size | object size in bytes; can be a range min..max | any positive integer | `1 048 576` (1 MiB) |
```
tritone, line 22: Why `..` for the separator, rather than a dash or a colon?

tritone (commented):
Code looks fine but can you explain reason for this change? it seems more confusing to overload this argument rather than multiple args as you have now.

storage/internal/benchmarks/README.md
BrennaEpp: This is what the team agreed upon. I think it makes sense to overload the parameter rather than have different cases (ignoring certain parameters).

PR #7650, opened 2023-03-29
fix(pubsub/pstest): Clear Subscription when calling `ClearMessages`.

mikeklaas:
Currently, `ClearMessages()` only clears `Server` messages, not touching any messages that have been published to a Subscription but not yet delivered.  This can cause odd behaviour if those messages are delivered as `pstest` assumes that any such messages exist in the Server's `msgsByID` map.  Attempting to ModAck such a message results in a NPE.


mikeklaas:
Happy to rebase or update the PR if helpful for reviewing.

hongalex (approved):

PR #7652, opened 2023-03-30
chore(storage): handle bucket labels for gRPC

BrennaEpp:
* Turns on BucketUpdate integration test for GRPC


storage/grpc_client.go
```diff
@@ -354,7 +354,15 @@ func (c *grpcStorageClient) UpdateBucket(ctx context.Context, bucket string, uat
 	if uattrs.Autoclass != nil {
 		fieldMask.Paths = append(fieldMask.Paths, "autoclass")
 	}
-	// TODO(cathyo): Handle labels. Pending b/230510191.
+
+	for label := range uattrs.setLabels {
+		fieldMask.Paths = append(fieldMask.Paths, fmt.Sprintf("labels.%s", label))
+	}
+
+	for label := range uattrs.deleteLabels {
```
tritone, line 362: Maybe add a comment to explain what is going on here since it has been a cause for confusion in the past.

`// Delete a label by not including it in Bucket.Labels but adding the key to the update mask.`

storage/integration_test.go
```diff
@@ -524,96 +524,87 @@ func TestIntegration_BucketLifecycle(t *testing.T) {
...
-		t.Fatal("bucket should not have versioning by default")
-	}
-	if len(attrs.Labels) > 0 {
-		t.Fatal("bucket should not have labels initially")
-	}
+		b := client.Bucket(prefix + uidSpace.New())
+		h.mustCreate(b, testutil.ProjID(), nil)
+		defer h.mustDeleteBucket(b)
 
-	// Using empty BucketAttrsToUpdate should be a no-nop.
```
tritone, line 544: It looks like you removed this check with an empty `BucketAttrsToUpdate`, any particular reason?

tritone (commented):
two small things, otherwise LGTM

storage/integration_test.go
BrennaEpp: Yes, I removed it because gRPC returns an error in that case, while it is a no-op for HTTP. Is there value in keeping this test, do you think? If there is, we'd have to either accept both cases here (could make for some not-nice looking code), or separate this check into 2 tests, one for HTTP and one for JSON

BrennaEpp: Discussed offline, opened #7685 to address the discrepancy issue.

tritone (approved):

PR #7656, opened 2023-03-30
test(storage): separate out public object test and enable gRPC

BrennaEpp:
Separated out the public object test from _Objects test into its own test


tritone (approved):
thanks!

PR #7658, opened 2023-03-30
test(spanner): Fine Grained Access Control for PG dialect

rajatbhatta:
This PR adds support for database roles for PG dialect and un-skips the integration tests which were earlier skipped for PG dialect. With this, Database roles support is now available in both GoogleSql and PostgreSql dialects.

Note: Database Roles support is still not available in Emulator.


codyoss:
@rajatbhatta can this be closed?

PR #7668, opened 2023-03-31
feat(spanner): add directed reads feature

harshachinta:
[DO NOT MERGE]
This PR adds code changes for Directed Reads.


spanner/apiv1/spannerpb/spanner.pb.go
rahul2393, line 1: This is auto-generate file please make sure not to merge it.

spanner/client_test.go
```diff
@@ -2971,66 +3067,68 @@ func TestBatchReadOnlyTransactionFromID_ReadOptions(t *testing.T) {
...
 type QueryOptionsTestCase struct {
-	name   string
-	client QueryOptions
-	env    QueryOptions
-	query  QueryOptions
-	want   QueryOptions
+	name      string
+	client    QueryOptions
+	clientDRO *sppb.DirectedReadOptions
+	env       QueryOptions
```
harshachinta, line 3073: @rahul2393 
Should we create a new struct for Directed Read Options instead of modifying the existing `QueryOptionsTestCase`? If we write a new struct then there would be new methods added which would look similar to  [testQueryOptions](https://github.com/googleapis/google-cloud-go/blob/1b0d41cb6b63daee3bc707515d8c301b5dc40320/spanner/client_test.go#L452) and [checkReqsForQueryOptions](https://github.com/googleapis/google-cloud-go/blob/1b0d41cb6b63daee3bc707515d8c301b5dc40320/spanner/client_test.go#L463).
Which one would be the right way to move with?

spanner/client_test.go
```diff
@@ -2971,66 +3067,68 @@ func TestBatchReadOnlyTransactionFromID_ReadOptions(t *testing.T) {
...
 }
 
 type ReadOptionsTestCase struct {
-	name   string
-	client *ReadOptions
-	read   *ReadOptions
-	want   *ReadOptions
+	name      string
+	client    *ReadOptions
+	clientDRO *sppb.DirectedReadOptions
```
harshachinta, line 3129: @rahul2393 
Similar to above [comment](https://github.com/googleapis/google-cloud-go/pull/7668/commits/1b0d41cb6b63daee3bc707515d8c301b5dc40320#r1159666644)
Should we create a new struct for Directed Read Options instead of modifying the existing `ReadOptionsTestCase`?

spanner/client.go
```diff
@@ -161,6 +177,8 @@ type ClientConfig struct {
 	// Logger is the logger to use for this client. If it is nil, all logging
 	// will be directed to the standard logger.
 	Logger *log.Logger
+
+	DirectedReadOptions *sppb.DirectedReadOptions
```
asthamohta, line 181: Add description here as well

spanner/transaction.go
```diff
@@ -853,6 +890,33 @@ func (t *ReadOnlyTransaction) release(err error) {
...
+func errInvalidLenDirectedReadOptions() error {
+	return spannerErrorf(codes.InvalidArgument, "Maximum length of replica selection allowed in DirectedReadOptions is 10")
+}
+
+func errInvalidDirectedReadOptions() error {
+	return spannerErrorf(codes.InvalidArgument, "Only one of IncludeReplicas or ExcludeReplicas can be set")
+}
+
+func (t *ReadOnlyTransaction) validateDirectedReadOptions(dro *sppb.DirectedReadOptions) (bool, error) {
+	if dro != nil {
```
asthamohta, line 906: Apologies since I am not that familiar with Golang and the correction isn't valid here but isn't it possible to use the function defined in client or to define an independent helper function rather than declaring the exact same function twice?

spanner/transaction.go
```diff
@@ -54,6 +54,8 @@ type txReadEnv interface {
 	// release should be called at the end of every transactional read to deal
 	// with session recycling.
 	release(error)
+
+	validateDirectedReadOptions(dro *sppb.DirectedReadOptions) (bool, error)
```
asthamohta, line 58: Add description

spanner/transaction.go
```diff
@@ -165,6 +167,8 @@ type ReadOptions struct {
 
 	// The request tag to use for this request.
 	RequestTag string
+
+	DirectedReadOptions *sppb.DirectedReadOptions
```
asthamohta, line 171: Add description

spanner/transaction.go
```diff
@@ -853,6 +890,33 @@ func (t *ReadOnlyTransaction) release(err error) {
 	}
 }
 
+func errCannotSetDirectedReadOptions() error {
+	return spannerErrorf(codes.InvalidArgument, "DirectedReadOptions cannot be set for ReadWriteTransaction or PDML")
```
asthamohta, line 894: This is a bad request error not invalid arguement

spanner/transaction.go
harshachinta: Refactored to use the same function in client for validation DRO.

spanner/transaction.go
harshachinta: `BadRequest` as a code is not available as part of grpc codes.
Can you please suggest which one would be more suitable from the available one's (https://github.com/grpc/grpc-go/blob/master/codes/codes.go) as this is what golang client uses.

spanner/client.go
harshachinta: Done

spanner/transaction.go
harshachinta: Done

spanner/transaction.go
harshachinta: Done

spanner/transaction.go
harshachinta: As discussed offline, we will be using `InvalidArgument` only

spanner/client_test.go
harshachinta: Using the same struct to avoid code duplication

spanner/client_test.go
harshachinta: Using the same struct to avoid code duplication

asthamohta:
LGTM.

rahul2393 (approved):

asthamohta (approved):

spanner/apiv1/spannerpb/spanner.pb.go
harshachinta: Done

rahul2393 (approved):

spanner/client.go
```diff
@@ -477,6 +477,8 @@ func (c *Client) BatchReadOnlyTransaction(ctx context.Context, tb TimestampBound
 	t.txReadOnly.qo = c.qo
 	t.txReadOnly.ro = c.ro
 	t.txReadOnly.disableRouteToLeader = true
+	t.txReadOnly.qo.DirectedReadOptions = c.dro
```
harshachinta, line 480: @rahul2393 
Need your help to validate this case. Lets discuss on Monday.

harshachinta: Directed read options can be set for Partitioned read and Partitioned query since they internally call read and execute_streaming_sql methods. Since the transaction type for `BatchReadOnlyTransaction` is read-only, we need to consider the client level option as fallback.

rahul2393 (approved):

PR #7671, opened 2023-03-31
fix(bigquery): RowIterator.Schema not filled when using Storage Read API

alvarowolfx:
Found an issue when reading a query and unmarshalling to a user defined struct. See code sample: 

```
type UserStruct struct {
    Name string
    Number  int
    State string
}

// Enable Storage API usage for fetching data
err = client.EnableStorageReadClient(ctx)
if err != nil {
    // TODO: Handle error.
}

sql := fmt.Sprintf(`SELECT name, number, state FROM %s WHERE state = "CA"`, `bigquery-public-data.usa_names.usa_1910_current`)

q := client.Query(sql)
it, err := q.Read(ctx)
if err != nil {
    // TODO: Handle error.
}
for {
    var s UserStruct
    err := it.Next(&s)
    if err == iterator.Done {
        break
    }
    if err != nil {
        // TODO: Handle error.
    }
    fmt.Println(s)
}
// Output 
// { 0 }
// { 0 }
// { 0 }
```

The target struct was returning with it's zero value, not being filled with data.


bigquery/storage_integration_test.go
```diff
@@ -287,10 +300,22 @@ func TestIntegration_StorageReadQueryMorePages(t *testing.T) {
 		Forks NullInt64
 	}
 
+	var firstValue S
+	err = it.Next(&firstValue)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if cmp.Equal(firstValue, S{}) {
```
shollyman, line 309: Is it worth amending the backing query to ensure at least one of the columns has a NOT NULL predicate and _some_ value is present?  Or is it sufficient as is because we expect the Null* types to be populated?

shollyman (approved):
Generally seems fine, one question around the test query driving the test.

shollyman (approved):
ship it!

PR #7673, opened 2023-03-31
feat(bigquery/storage/managedwriter): expose connection multiplexing as experimental

shollyman:
This PR exposes the necessary options to control the new experimental multiplexing features within the managedwriter package, and attempts to document them sufficiently for correct consumption.


yirutang (approved):
Description looks good to me.

bigquery/storage/managedwriter/options.go
```diff
@@ -48,31 +61,55 @@ type writerClientOption interface {
...
+// be in flight on a connection at a time.  This limit is enforced on all connections
+// created by the instantiated Client.
 //
-// TODO: export this as part of the multiplex feature launch.
-func defaultMaxInflightRequests(n int) option.ClientOption {
+// Note: the WithMaxInflightRequests WriterOption can still be used to control
+// the behavior for individual ManagedStream writers when not using multiplexing.
+//
+// This ClientOption is EXPERIMENTAL and subject to change.
+func DefaultMaxInflightRequests(n int) option.ClientOption {
```
codyoss, line 112: nit: most of our options are prefixed with With, could we carry that convention here as well?

bigquery/storage/managedwriter/options.go
```diff
@@ -61,14 +61,14 @@ type writerClientOption interface {
...
-// EnableMultiplexing is an EXPERIMENTAL option that enables connection sharing
+// WithMultiplexing is an EXPERIMENTAL option that controls connection sharing
 // when instantiating the Client.  Only writes to default streams can leverage the
 // multiplex pool.  Internally, the client maintains a pool of connections per BigQuery
 // destination region, and will grow the pool to it's maximum allowed size if there's
 // sufficient traffic on the shared connection(s).
 //
 // This ClientOption is EXPERIMENTAL and subject to change.
-func EnableMultiplexing(enable bool) option.ClientOption {
+func WithMultiplexing(enable bool) option.ClientOption {
```
codyoss, line 71: This one I think is fine to not use the With and instead pass nothing. The presence of the option seems like a logical true if passed. But I will leave it up to you

codyoss (approved):
Left one more consideration, but I am fine either way

bigquery/storage/managedwriter/options.go
shollyman: Decided to do `WithMultiplexing()` here without argument(s).

PR #7674, opened 2023-04-01
fix(datastore): fix NoIndex for array property

aaron0x:
When converting proto to Entity with array property, ExcludeFromIndexes appears in the values level, not the top level.


bhshkh (approved):

datastore/load_test.go
```diff
@@ -602,6 +602,62 @@ func TestTimezone(t *testing.T) {
...
+	}
+
+	dst, err := protoToEntity(src)
+	if err != nil {
+		t.Fatalf("protoToEntity: %v", err)
+	}
+
+	if !testutil.Equal(want, dst) {
+		t.Errorf("NoIndex should be correct: compare:\ngot:  %#v\nwant: %#v", dst, want)
+	}
```
bhshkh, line 658:
```suggestion
	cmpProperties := func(p1, p2 Property) bool {
		return p1.Name < p2.Name
	}
	if !testutil.Equal(want.Properties, dst.Properties, cmpopts.SortSlices(cmpProperties)) {
		t.Errorf("NoIndex should be correct: Property:\ngot:  %#v\nwant: %#v", dst, want)
	}
	if !testutil.Equal(want.Key, dst.Key) {
		t.Errorf("NoIndex should be correct: Key:\ngot:  %#v\nwant: %#v", dst, want)
	}
```

datastore/load_test.go
```diff
@@ -602,6 +602,62 @@ func TestTimezone(t *testing.T) {
 	}
 }
 
+func TestLoadArrayIndex(t *testing.T) {
```
bhshkh, line 605: This fails intermittently with error:

```
=== RUN   TestLoadArrayIndex
    load_test.go:657: NoIndex should be correct: compare:
        got:  &datastore.Entity{Key:(*datastore.Key)(0xc0005457c0), Properties:[]datastore.Property{datastore.Property{Name:"non-indexed", Value:[]interface {}{"3", "4"}, NoIndex:true}, datastore.Property{Name:"indexed", Value:[]interface {}{"1", "2"}, NoIndex:false}}}
        want: &datastore.Entity{Key:(*datastore.Key)(0xc0004b1b80), Properties:[]datastore.Property{datastore.Property{Name:"indexed", Value:[]interface {}{"1", "2"}, NoIndex:false}, datastore.Property{Name:"non-indexed", Value:[]interface {}{"3", "4"}, NoIndex:true}}}
--- FAIL: TestLoadArrayIndex (0.00s)
```

Please see suggested changes.

bhshkh (changes requested):

datastore/load_test.go
aaron0x: thanks for the suggestion.

bhshkh (approved):

PR #7682, opened 2023-04-03
chore: decouple post processor from gapicgen

codyoss:
- This will make it easier to onboard new clients by configuring
  only two files: OwlBot.yaml and config.yaml
- Remove all deps on gapicgen. This meant inlining more logic into
  the post processor
- Split out logic into domain files
- Add idea of a post processor config that reads file dynamically
  ensuring we can onboard new clients without re-deploying post
  processor
- Use a heuristic to determine a release-level


codyoss:
Note once review is done I will add the updated post-processor owlbot revision

internal/postprocessor/.gitignore
```diff
@@ -0,0 +1 @@
+postprocessor
```
quartzmo, line 1: Why is this needed?

codyoss: If you run `go build` in this directory it produces a binary called postprocessor, this makes sure that never gets uploaded to github

internal/postprocessor/config.yaml
```diff
@@ -124,3 +124,435 @@ modules:
   - webrisk
   - websecurityscanner
   - workflows
+
+service-configs:
```
quartzmo, line 128: Would it be possible to sort these alphabetically?

codyoss: They are, by input directory

internal/postprocessor/mainifest.go
```diff
@@ -0,0 +1,267 @@
...
+	for scanner.Scan() && lineCnt < 50 {
+		line := scanner.Text()
+		if strings.Contains(line, betaIndicator) {
+			return "beta", nil
+		}
+	}
+	return "ga", nil
+}
+
+var manualEntries = []ManifestEntry{
```
quartzmo, line 146: I know these are very stable, but just have to ask: Should this be extracted to a config file too?

internal/postprocessor/config.yaml
```diff
@@ -555,4 +559,111 @@ service-configs:
...
+ - input-directory: google/cloud/translate/v3
+   service-config: translate_v3.yaml
+   import-path: cloud.google.com/go/translate/apiv3
+ - input-directory: google/cloud/vision/v1
+   service-config: vision_v1.yaml
+   import-path: cloud.google.com/go/vision/v2/apiv1
+ - input-directory: google/cloud/vision/v1p1beta1
+   service-config: vision_v1p1beta1.yaml
+   import-path: cloud.google.com/go/vision/v2/apiv1p1beta1
+ - input-directory: google/cloud/workstations/v1beta
```
quartzmo, line 664: Unsorted, move `workstations` down?

internal/postprocessor/config.yaml
quartzmo: Regarding sort order, see comment below on `workstations`.

quartzmo: Can modules now be derived from service-configs?

codyoss: They can not, a module can have N clients at different levels and modules don't have to be directly off the root.

internal/postprocessor/config.yaml
codyoss: It is sorted, maybe this was in an older commit

internal/postprocessor/mainifest.go
codyoss: I like the idea, done

internal/postprocessor/config.yaml
```diff
@@ -0,0 +1,776 @@
...
+ - input-directory: google/spanner/admin/instance/v1
+   service-config: spanner_admin_instance.yaml
+ - input-directory: google/spanner/v1
+   service-config: spanner.yaml
+ - input-directory: google/storage/v2
+   service-config: storage_v2.yaml
+ - input-directory: google/storagetransfer/v1
+   service-config: storagetransfer_v1.yaml
+   # All entries below are unmigrated clients. We need to store extra
+   # info for them since they don't have an entry in owlbot.yaml yet.
```
quartzmo, line 671: Do some of the unmigrated clients keys shadow entries above this line? For example, 'google/devtools/cloudtrace/v1'?

codyoss: Oops you are right, I generated the initial set before I realized non-migrated clients needs more info. Updated

PR #7688, opened 2023-04-04
chore(storage): handle object metadata (grpc)

BrennaEpp:
* t.Cleanup does not work with gRPC (connection is closed before t.Cleanup runs - probably a question of using defer elsewhere)


storage/grpc_client.go
```diff
@@ -540,7 +540,10 @@ func (c *grpcStorageClient) UpdateObject(ctx context.Context, bucket, object str
 	if uattrs.ACL != nil || len(uattrs.PredefinedACL) > 0 {
 		fieldMask.Paths = append(fieldMask.Paths, "acl")
 	}
-	// TODO(cathyo): Handle metadata. Pending b/230510191.
+
+	if uattrs.Metadata != nil {
+		fieldMask.Paths = append(fieldMask.Paths, "metadata")
```
tritone, line 545: Should we specify which keys are being updated here using dot notation? I'm a bit confused so maybe we can discuss tomorrow.

BrennaEpp: Discussed offline, this was fine too but it'll be clearer to explicitly send added keys using dot notation, and we can do the same for deleted keys once/if we add support for deleting specific keys.

tritone (approved):

Issue #7604, opened 2023-03-21
bigquery: Timeout not working


jeancarlosgarcia:
**Client**

Bigquery

**Code**
```
package client

func executeQuery(ctx context.Context, q *bigquery.Query) (*bigquery.RowIterator, error) {
	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
	defer cancel()

	return q.Read(ctx)
}
```

**Expected behavior**

It should return a timeout error.

**Actual behavior**

Even assigning a timeout of 1 * time.Nanosecond, the client takes more than the configured time.

shollyman:
Thanks for the report, was able to confirm we're not respecting context properly here.  I'll look into adding a fix for this.

PR #7693, opened 2023-04-05
fix(bigquery): respect context during query execution

shollyman:
This PR updates query polling and insertion to better respect context status.  Key change is to push context to the discovery library, but there are some small refactors to make use of context passing more consistent at various callsites.

Fixes: https://github.com/googleapis/google-cloud-go/issues/7604


bigquery/bigquery.go
```diff
@@ -173,24 +173,40 @@ func (c *Client) insertJob(ctx context.Context, job *bq.Job, media io.Reader) (*
...
-		sCtx := trace.StartSpan(ctx, "bigquery.jobs.query")
-		res, err = call.Do()
-		trace.EndSpan(sCtx, err)
-		return err
+	type respS struct {
+		res *bq.QueryResponse
+		err error
 	}
+	ch := make(chan *respS, 1)
+	go func() {
```
codyoss, line 181: if we pass the context to the call obj, can all of this extra code be avoided?

shollyman: Good spot, thanks for calling that out.  removed the outer context patterns and just plumbed the context into the discovery rpcs.

codyoss (approved):
LGTM

PR #7695, opened 2023-04-05
refactor(logging): move testing util functions into own file

tang-fh:
Move functions used for tests into their own file.

This removes the build dependency on logging_unexported_test.go from package logging_test.

### DIFF
--- a/logging/testutils.go
+++ b/logging/testutils.go
@@ -1,0 +1,34 @@
+// Copyright 2016 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


quartzmo (approved):

logging/testutils.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2023 Google LLC
```
codyoss, line 1: This should be in a test file so it is not added to the public api of this package.

codyoss (changes requested):

PR #7704, opened 2023-04-06
test(storage): add 408 to upload conf tests

tritone:
Add cases with a 408 error to the conformance tests for uploads. Also updates one other test scenario that differed from the source json in googleapis/conformance-tests.


storage/internal/test/conformance/retry_tests.json
```diff
@@ -265,7 +268,7 @@
       "description": "downloads_handle_complex_retries",
       "cases": [
         {
-          "instructions": ["return-broken-stream"]
+          "instructions": ["return-broken-stream", "return-broken-stream"]
```
BrennaEpp, line 271: I don't see a matching change for this line in https://github.com/googleapis/conformance-tests/pull/78/files

tritone: Discussed offline, this was another upstream change that the go repo hadn't picked up

ddelgrosso1 (approved):

PR #7705, opened 2023-04-06
chore: onboard compute to owlbot

codyoss:


noahdietz:
There is a nasty hack to add an extra generator flag specifically for `compute`:
https://github.com/googleapis/google-cloud-go/blob/7a9e21121ca8debffe11fe16d1b96b0d969e11ac/internal/gapicgen/generator/gapics.go#L323-L326

There is a bazel equivalent I believe. I just remembered this, sorry _face palm_

noahdietz (approved):
LGTM once https://github.com/googleapis/googleapis/pull/793 is in

PR #7706, opened 2023-04-07
feat(bigquery): expose dataset MaxTimeTravelHours

shollyman:


bigquery/dataset.go
```diff
@@ -58,6 +58,10 @@ type DatasetMetadata struct {
 	// More information: https://cloud.google.com/bigquery/docs/reference/standard-sql/collation-concepts
 	DefaultCollation string
 
+	// MaxTimeTravelHours represents the number of hours for the max time travel for all tables
+	// in the dataset.  Durations are rounded towards zero for the nearest hourly value.
+	MaxTimeTravelHours time.Duration
```
codyoss, line 63: nit: I get the intention with the word Hours here, but it does feel a little off since this is just a duration. Maybe drop hours. Also maybe TraversalTime instead of TimeTravel -- although it does sound cool 🚀

codyoss (approved):

PR #7708, opened 2023-04-07
chore(internal): update postprocessor README instructions

quartzmo:


internal/postprocessor/README.md
```diff
@@ -13,41 +13,75 @@ fork instead of the repo.
...
-### Run post-processor on all clients
+### Docker container
 
-From the `google-cloud-go/internal/postprocessor` directory run: 
-```bash
-go run main.go -client-root="../.." -googleapis-dir="/path/to/local/googleapis"
-```
+*Note*: The docker container needs to be built with the context of the entire
+`google-cloud-go/internal` directory. If building the container directly with
+`docker`, rather than with Cloud Build as instructed below, be sure to do so
```
codyoss, line 39: This is fine for now, but would you file an issue about making this easier. I don't think we should have to have the context of the internal directory anymore since I decoupled the postprocessor from gapicgen earlier this week.

internal/postprocessor/README.md
```diff
@@ -13,41 +13,75 @@ fork instead of the repo.
...
+
+   *Note*: OwlBot will eventually open a pull request to update the SHA if it
+   discovers a new version of the container.
+10. After your PR is approved and CI is green, merge your changes.
 
-### Initializing new modules
-To initialize the `internal/version.go`, `go.mod`, `README.md`, and `CHANGES.md` files in a new module, add the module to the slice in `modconfig.go`. The entry should correspond to the location where the `go.mod` file should be initialized minus the prefix "google-cloud-go/"
+## Initializing new modules
+To initialize the `internal/version.go`, `go.mod`, `README.md`, and `CHANGES.md`
+files in a new module, add the module to the slice in `modconfig.go`. The entry
```
codyoss, line 85: This should be updated to reflect that this now happens in config.yaml

internal/postprocessor/README.md
quartzmo: Can we just delete this *note*? I think building the container with Cloud Build is easy and simple.

quartzmo: Deleted the note. Let me know if you want something added here, or if an issue needs to be filed.

internal/postprocessor/README.md
quartzmo: Fixed, thanks!

PR #7725, opened 2023-04-11
fix(pubsublite): refresh OAuth tokens 5 minutes before expiry

tmdiep:
Overrides the default of 10s in the OAuth library to mitigate failure to refresh auth tokens, causing the publisher and subscriber clients to permanently error.


codyoss (changes requested):
discussion offline

tmdiep:
This will break users who have set other auth mechanisms in the API options (e.g. API keys, creds file), as only one type can be set and is validated here: https://github.com/googleapis/google-api-go-client/blob/8019ef6850bbd9dacb6f42796f2d8ca9940b875b/internal/settings.go#L111

(thanks codyoss for the warning)

codyoss:
auth library by default refreshes sooner and asyncly. I don't think this is needed anymore.

PR #7728, opened 2023-04-11
feat(storage): add support for matchGlob parameter

### LINKED ISSUE #7727: storage: add support for matchGlob
Support for the `matchGlob` parameter when [listing objects](https://cloud.google.com/storage/docs/json_api/v1/objects/list#parameters) in a bucket is not available from the SDK. 

Add an attribute to the `storage.Query` struct to be able to provide a glob pattern to match.

The alternative is to use the JSON API directly, which works, but then you need to handle all the boilerplate with headers and authentication etc. by your self.

goober:
This PR aims to add support for the matchGlob parameter when listing objects.
It currently only adds support for the parameter when using the http API.

Fixes #7727


storage/http_client.go
```diff
@@ -347,6 +347,7 @@ func (c *httpStorageClient) ListObjects(ctx context.Context, bucket string, q *Q
 		req.EndOffset(it.query.EndOffset)
 		req.Versions(it.query.Versions)
 		req.IncludeTrailingDelimiter(it.query.IncludeTrailingDelimiter)
+		req.MatchGlob(it.query.MatchGlob)
```
tritone, line 350: We'll also need to add this to the other interface implementation in grpc_client.go

tritone (commented):
Thanks for the PR. We need to add this to the gRPC implementation as well, and add a case for this to the unit/integration tests. LMK if you'd like pointers on how to do this, or if you're okay with us taking care of it.

goober:
Thank you @tritone. It does not matter for me if you take it from here or if I update the PR with necessary changes. However, I have not looked into where to find any documentation around the gRPC api, and if the parameter is available there, so some pointer would be great.

Crystaleclipspawl (approved):

tritone:
Apologies for the delay. Closing this in favor of #8097 which has the necessary testing etc.

Issue #7740, opened 2023-04-12
spanner: Error wraped into Internal


osterante:
**Client**

Spanner v1.45.0

**Environment**

The local environment on M1 MacBookPro

**Go Environment**
$ go version
go version go1.20 darwin/arm64

<details>
<summary>$ go env</summary>

```
GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/Users/UserName/Library/Caches/go-build"
GOENV="/Users/UserName/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/UserName/go/pkg/mod"
GOOS="darwin"
GOPATH="/Users/UserName/go"
GOPROXY="https://proxy.golang.org/,direct"
GOROOT="/Users/UserName/.anyenv/envs/goenv/versions/1.20.0"
GOSUMDB="[sum.golang.org](http://sum.golang.org/)"
GOTMPDIR=""
GOTOOLDIR="/Users/UserName/.anyenv/envs/goenv/versions/1.20.0/pkg/tool/darwin_arm64"
GOVCS=""
GOVERSION="go1.20"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/UserName/Projects/Project/go.mod"
GOWORK=""
CGO_CFLAGS="-O2 -g"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-O2 -g"
CGO_FFLAGS="-O2 -g"
CGO_LDFLAGS="-O2 -g"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/x3/0yyktc4d5_11gvj8gsjxsck5mc3vtb/T/go-build887811120=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

</details>

**Code**

e.g.
```go
func main() {
        ctx, cancel := context.WithCancel(context.Background())
        client, _ := spanner.NewClient(ctx, "path to db")

        _, err = client.ReadWriteTransaction(ctx, func(ctx context.Context, tx *spanner.ReadWriteTransaction) error {
[...]

PR #7744, opened 2023-04-13
fix(spanner): context timeout should be wrapped correctly

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/7740


spanner/client.go
```diff
@@ -533,7 +533,8 @@ func (c *Client) rwTransaction(ctx context.Context, f func(context.Context, *Rea
 		}
 		if t.shouldExplicitBegin(attempt) {
 			if err = t.begin(ctx); err != nil {
-				return spannerErrorf(codes.Internal, "error while BeginTransaction during retrying a ReadWrite transaction: %v", err)
+				trace.TracePrintf(ctx, nil, "Error while BeginTransaction during retrying a ReadWrite transaction: %v", ToSpannerError(err))
+				return ToSpannerError(err)
```
olavloite, line 537: Could we have a test that shows that this actually fixes the problem? I'm a little surprised that the sample code given in the original issue gets into this branch of the code, as it indicates that the transaction should use an explicit `BeginTransaction` call. It's not clear to me why the example would do that, as it just uses a normal `SELECT` statement as the first statement in the transaction. That would then normally mean that the `BeginTransaction` is inlined with the execution of that statement.

rahul2393: Added tests, this happend because we do explicit begin transaction if first statement returns error, in this case call the iter.Next() returned cancelled and while doing explicit Begin we returned customer internal error. https://github.com/googleapis/google-cloud-go/blob/main/spanner/read.go#L143

olavloite (approved):

PR #7746, opened 2023-04-13
fix(pubsublite): enforce minimum publisher and subscriber timeout of 2 minutes

tmdiep:
Low timeout values are harmful, as they cause the publisher or subscriber client to terminate if exceeded, which then causes confusion. If a value less than 2 minutes is set, it will be clamped to 2 minutes and a warning will be logged. In the future, a value less than 2 minutes will result in an error.


hongalex (approved):

pubsublite/internal/wire/settings.go
```diff
@@ -98,8 +102,8 @@ func validatePublishSettings(settings PublishSettings) error {
 	if settings.DelayThreshold <= 0 {
 		return errors.New("pubsublite: invalid publish settings. DelayThreshold duration must be > 0")
 	}
-	if settings.Timeout <= 0 {
-		return errors.New("pubsublite: invalid publish settings. Timeout duration must be > 0")
+	if settings.Timeout < minTimeout {
+		return errors.New("pubsublite: invalid publish settings. Timeout duration must be >= 2 minutes")
```
codyoss, line 106: I wonder if instead of erroring we just default to two minutes in these "error" cases if that really should be the min acceptable value. Still a behaviour change, but it will allow all old code to keep working without need of a code change. Thoughts?

tmdiep: Thanks for the suggestion! I've clamped to a minimum of 2 minutes and logged a warning, in case users are relying on a low timeout. The warning says we will error in the future (say after 6 months).

codyoss (approved):
Thanks, LGTM

PR #7749, opened 2023-04-13
chore(storage): move stubs dir to storagepb

codyoss:
This is to align this the pattern we use with other apis and will clean up cutting this generation over to owlbot. This also temporarily stops the old generation pipeline for updating storage. A internal CL updating generated paths and another pr to onboard onto owlbot will followup this commit.


codyoss:
Will merge closer to the time the backend change is made for generation

noahdietz (approved):

storage/acl.go
```diff
@@ -20,7 +20,7 @@ import (
 	"reflect"
 
 	"cloud.google.com/go/internal/trace"
-	storagepb "cloud.google.com/go/storage/internal/apiv2/stubs"
+	storagepb "cloud.google.com/go/storage/internal/apiv2/storagepb"
```
tritone, line 23: I guess this no longer needs an alias import, is there some other reason you are keeping the alias?

tritone (approved):
Thanks Cody!

PR #7773, opened 2023-04-17
chore(storage): continue on failures & ensure all runs are printed [benchmarking]

BrennaEpp:
This PR restructures error and cleanup handling.

This PR should do the following:

- Ensure cleanup always happens.
- Cleanup errors do not immediately terminate the program nor stop future runs, but still cause the program to exit with an error code.
- Run errors  do not immediately terminate the program nor stop future runs, but still cause the program to exit with an error code. 
- Cleanup and run errors are logged.
- Ensure all run results are printed instead of exiting right away (return err or set bool so that we can wait for results to finish printing). Previously, results may not have gotten the chance to be written as we directly exited on error.
- Runs do not start setup if context is cancelled.
- Timeouts are properly noted for GRPC runs. Timeouts are considered a sub category of error.

* This also removes an option that was not being used and we don't have a need for (the prober can add labels itself)

Previously, objects were being created and downloaded directly into the tmp folder. To ease cleanup, they are now downloaded to a specific subfolder in tmp (cleanup just deletes that subfolder and all its contents).


tritone:
Could you resolve the merge conflict first before I review?

storage/internal/benchmarks/download_benchmark.go
```diff
@@ -25,12 +25,13 @@ import (
...
-	object      string
-	rangeStart  int64
-	rangeLength int64
+	client              *storage.Client
+	objectSize          int64
+	bucket              string
+	object              string
+	rangeStart          int64
+	rangeLength         int64
+	downloadToDirectory string
```
tritone, line 34: Can you explain how adding this opt is related? Is it something that's part of the public surface or just intended for testing?

storage/internal/benchmarks/main.go
```diff
@@ -249,28 +251,53 @@ func main() {
...
+				// Cleanup may be expected to fail if there is an issue with the run.
+				log.Printf("run cleanup failed: %v", err)
+				exitWithErrorCode = true
 			}
 			return nil
 		})
 	}
 
-	benchGroup.Wait()
+	err := benchGroup.Wait()
```
tritone, line 285: `Errgroup.Wait` seems to return a single error. Does it make sense to exit with just this error on L295? I guess other errors will be in the logs?

storage/internal/benchmarks/main.go
```diff
@@ -249,28 +251,53 @@ func main() {
...
 	if outputFile != "" {
 		// if sending output to a file, we can use stdout for informational logs
 		fmt.Printf("\nTotal time running: %s\n", time.Since(start).Round(time.Second))
 	}
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if exitWithErrorCode {
```
tritone, line 298: Under what circumstances would `exitWithErrorCode` be true but the script would not have already exited on L295 above?

storage/internal/benchmarks/utils.go
```diff
@@ -127,3 +130,18 @@ func populateDependencyVersions() error {
...
+
+// errorIsDeadLineExceeded functions like errors.Is(err, context.DeadlineExceeded)
+// Except, it unwraps the error to look for GRPC DeadlineExceeded errors.
+func errorIsDeadLineExceeded(err error) bool {
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
+
+	for err = errors.Unwrap(err); err != nil; {
+		if status.Code(err) == codes.DeadlineExceeded {
```
tritone, line 142: Nice

storage/internal/benchmarks/w1r3.go
```diff
@@ -51,20 +49,26 @@ import (
 //     This includes opening the file, writing the object, and verifying the hash
 //     (if applicable).
 //   - Take another snapshot of memory stats.
-//   - Delete the file from the OS.
```
tritone, line 54: Did you mean to remove these comments? Or was this a mistake in merge conflict resolution?

tritone (commented):
Few questions, perhaps we can meet about this on Monday

storage/internal/benchmarks/main.go
BrennaEpp: I think it makes sense. Run and cleanup errors will be in the logs. Setup errors are the only ones that would cause `Errgroup.Wait` to fail. Since the context is cancelled when the first setup error is seen, it should be the first error and the one we want to print. Additional errors would happen after the context is cancelled, at which point they are no longer that relevant.

storage/internal/benchmarks/main.go
BrennaEpp: So for runs and cleanup, if you see L270 and L278, we log the error and set `exitWithErrorCode`, but we do not return the error. This is because we want to continue benchmarking, but still log the error. If you return the error, the group's context is cancelled. 

I just added a commit (missed that earlier) to make sure the group's context (based on the original context) is the one being used in the setup and runs. So, any future runs would stop on seeing the context cancelled when an error is returned. We want this for setup since a setup error indicates in most cases that future setups will also fail.

So, if setup returns a failure, runs would be cancelled, cleanup done, and then `Errgroup.Wait` would return with an error and the script would exit on L295.

If the run or cleanup returns a failure, it would be logged, and the script would continue until it finishes `i` samples, then `Errgroup.Wait` would return without an error and we would reach these lines indicating to exit with an error code.

storage/internal/benchmarks/w1r3.go
BrennaEpp: I did - this is no longer done in the run, it's now done on the cleanup step (which has its own comment).

tritone (approved):

storage/internal/benchmarks/download_benchmark.go
tritone: Sounds good, maybe just note this in the commit msg as well.

PR #7775, opened 2023-04-17
chore(dataproc): add dataproc/v2 to OwlBot

quartzmo:


.github/.OwlBot.yaml
```diff
@@ -74,6 +74,7 @@ deep-remove-regex:
   - /datalabeling/apiv1beta1/
   - /dataplex/apiv1/
   - /dataproc/apiv1/
```
noahdietz, line 76: This should be removed I think because it will not longer be regenerated, so we don't want to remove it.

.github/.OwlBot.yaml
```diff
@@ -74,6 +74,7 @@ deep-remove-regex:
...
+  - /dataproc/v2/
```
noahdietz, line 77: I think to be consistent with the trailing segment of every other item, we should use the api version at the end:
```suggestion
  - /dataproc/v2/apiv1/
```

.github/.OwlBot.yaml
```diff
@@ -239,6 +240,7 @@ deep-preserve-regex:
   - /spanner/admin/instance/apiv1/init.go
   - /spanner/apiv1/spanner_client_options.go
   - /dataproc/apiv1/longrunning.go
```
noahdietz, line 242: Not sure if this line is necessary if the `deep-remove-regex` no longer includes `/dataproc/apiv1`.

.github/.OwlBot.yaml
```diff
@@ -389,6 +391,8 @@ deep-copy-regex:
     dest: /dataplex/apiv1
   - source: /google/cloud/dataproc/v1/cloud.google.com/go/dataproc/apiv1
     dest: /dataproc/apiv1
```
noahdietz, line 393: This could probably be removed since it will no longer be generated.

.github/.OwlBot.yaml
quartzmo: I think we should leave it in during the transition.

.github/.OwlBot.yaml
quartzmo: Removed, thanks.

.github/.OwlBot.yaml
quartzmo: Removed, thanks.

.github/.OwlBot.yaml
```diff
@@ -74,6 +74,7 @@ deep-remove-regex:
   - /datalabeling/apiv1beta1/
   - /dataplex/apiv1/
   - /dataproc/apiv1/
+  - /dataproc/v2/apiv1/
```
codyoss, line 77: remove v2

.github/.OwlBot.yaml
```diff
@@ -387,8 +388,8 @@ deep-copy-regex:
     dest: /datalabeling/apiv1beta1
   - source: /google/cloud/dataplex/v1/cloud.google.com/go/dataplex/apiv1
     dest: /dataplex/apiv1
-  - source: /google/cloud/dataproc/v1/cloud.google.com/go/dataproc/apiv1
-    dest: /dataproc/apiv1
+  - source: /google/cloud/dataproc/v1/cloud.google.com/go/dataproc/v2/apiv1
+    dest: /dataproc/v2/apiv1
```
codyoss, line 392: remove v2

codyoss (changes requested):
If we are planning to do this in place then the v2 needs to be dropped from the dest. That will only be apart of the import path

.github/.OwlBot.yaml
quartzmo: Done, thanks.

codyoss (approved):

PR #7776, opened 2023-04-18
chore(storage): migrate SignedURL tests

tritone:
Finishing off these tests for gRPC. One test has to be skipped for gRPC because credential capture logic in the client constructor is not present in the gRPC constructor. I'll file an issue for this if there isn't one already.

Addressed feedback from #6858


BrennaEpp (approved):
LGTM 👍

PR #7777, opened 2023-04-18
chore(storage): handle zero-byte reads in gRPC

### LINKED ISSUE #7756: storage: reader with length 0 returns full object [GRPC]
We should make a decision on what we are going to do wrt asking for 0 bytes with grpc, so that we can enable `TestIntegration_Reader` for gRPC.

Currently, if you ask for 0 bytes (ex: `Object.NewRangeReader(ctx, offset, length: 0)`):

- with HTTP: we make a call to GCS for the full object, close the body and replace it with an empty body.

- with gRPC: we make a call to GCP for the full object (it is not possible to make a call for 0 bytes) and return it

Options:
A. Do nothing. Document that asking for 0 bytes will be equivalent to asking for all bytes in gRPC.
B. Return an error in gRPC if user asks for 0 bytes.
C. Return an error in both gRPC and HTTP. This was originally discussed when implementing this in HTTP.  Probably not feasible now without a breaking change.
D. Implement something similar to HTTP for gRPC. This would maintain parity between HTTP and gRPC but would allow users to inadvertently ask for 0 bytes.

Note: see original discussion on what to do in this case: https://code-review.git.corp.google.com/c/gocloud/+/3430/9..20/storage/storage.go#b467

tritone:
Reads with a range of zero bytes should match the behavior in JSON/XML by requesting the whole object, but immediately closing the stream before fetching all the data.

I cleaned up some confusing discrepancies around the semantics of Reader.length as well.

Reader integration tests pass locally and can now be enabled for gRPC.

Fixes #7756

### DIFF
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -907,15 +907,8 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
 
 		req.ReadOffset = params.offset + seen
 
-		// A negative length means "read to the end of the object", but the
-		// read_limit field it corresponds to uses zero to mean the same thing. Thus
-		// we coerce the length to 0 to read to the end of the object.
-		if params.length < 0 {
-			params.length = 0
-		}
-
-		// Only set a ReadLimit if length is greater than zero, because zero
-		// means read it all.
+		// Only set a ReadLimit if length is greater than zero, because <= 0 means
+		// to read it all.
 		if params.length > 0 {
 			req.ReadLimit = params.length - seen
 		}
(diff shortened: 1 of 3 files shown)


storage/grpc_client.go
```diff
@@ -996,8 +990,15 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
 		r.remain = size
 	}
 
+	// For a zero-length request, explicitly close the stream and set remaining
+	// bytes to zero.
+	if params.length == 0 {
```
BrennaEpp, line 995: Do we want to consider improving perf for zero-length requests by requesting less bytes to start with as we are just going to discard them?

tritone: I don't think this is enough of a mainstream use case to optimize for; also, only the first 2MiB of the object will come through in the first `recv()`. This seems fine to me.

BrennaEpp (approved):

PR #7792, opened 2023-04-19
chore(internal/postprocessor): add a workaround for moving new snippets

codyoss:
This solution is sub-optimal, but I would like to get our current
owl-bot-pr in a mergable state as it is starting to pile up. Will
come back to this when I better understand the issue. Tested this
code locally though and it works.

Before this was getting a `no such file or directory` error. After
making the dir I got `file exists`. Adding remove after creating
seems to fix the issue. I assume this is some file permission
type of issue.


quartzmo (approved):
Thank you!

Issue #7798, opened 2023-04-20
storage: determine what to do on chunksize=0 [GRPC]


BrennaEpp:
Address TODOs around L1472 in grpc_client.go

Should we actually use the minimum of 256 KB here [chunksize=0] when the user indicates they want minimal memory usage? We cannot do a zero-copy, bufferless upload like HTTP/JSON can.

Must make a note to update documentation accordingly.

tritone:
Changed the chunkSize=0 logic to do the following:

1. Use a single-shot upload rather than resumable
2. Allocate 256k for the buffer (which is the min possible)

PR #7799, opened 2023-04-20
chore(storage): round up Chunksize to nearest multiple of 256kib [GRPC]

### LINKED ISSUE #7754: storage: round up Chunksize value to nearest multiple of 256kib [GRPC]
In HTTP we round the value for Writer.Chunksize up to the nearest multiple on 256kib. 

We should do the same for GRPC, so that setting the Chunksize to a value such as `96` does not return an error, as is the case for HTTP.

BrennaEpp:
Fixes #7754

GRPC skipped due to #7798 affecting the last test `chunksize 0 uploads everything`

Also cleans up a typo I noticed in ProgressFunc docs


storage/grpc_client.go
```diff
@@ -1471,6 +1472,12 @@ func (r *gRPCReader) reopenStream() (*storagepb.ReadObjectResponse, error) {
 
 func newGRPCWriter(c *grpcStorageClient, params *openWriterParams, r io.Reader) *gRPCWriter {
 	size := params.chunkSize
+
+	// Round up chunksize to nearest 256KiB
+	if size%googleapi.MinUploadChunkSize != 0 {
+		size += googleapi.MinUploadChunkSize - (size % googleapi.MinUploadChunkSize)
```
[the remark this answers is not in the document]
tritone: And, the if statement is superfluous.

storage/integration_test.go
```diff
@@ -2250,6 +2250,75 @@ func TestIntegration_WriterContentType(t *testing.T) {
...
+		obj := client.Bucket(bucket).Object("writer-chunksize-test" + uidSpaceObjects.New())
+		objSize := 17 << 10 << 10 // 17 Mib
+		contents := bytes.Repeat([]byte("a"), objSize)
+
+		for _, test := range []struct {
+			desc             string
+			chunksize        int
+			wantBytesPerCall int64
+		}{
+			{
```
tritone, line 2265: Let's also check that progressfunc is called the expected number of times for each case.

BrennaEpp: Good call, adding that uncovered a discrepancy we have between HTTP and GRPC - the progress function is only called in HTTP if the upload is resumable (so it is not called at all for a one-shot), in GRPC the function is called at least once even if the upload is not resumable. 

The docs for the progress func do specify it gets called only if the upload requires more than one call, so I assume we want to modify gRPC to comply with this?

storage/grpc_client.go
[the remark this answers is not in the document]
BrennaEpp: The if statement is needed when a user provides an exact multiple of 256kib. Without the if statement, the chunksize would be `chunksize+256kib` instead of just `chunksize`.

I am inclined to keep the code as-is since it's copied directly from [here](https://github.com/googleapis/google-api-go-client/blob/main/googleapi/googleapi.go#L248-L250); might as well keep it consistent.

storage/integration_test.go
tritone: Agreed, that sounds right.

storage/integration_test.go
```diff
@@ -2251,41 +2251,53 @@ func TestIntegration_WriterContentType(t *testing.T) {
...
 				desc:             "default chunksize",
 				chunksize:        16 << 10 << 10,
 				wantBytesPerCall: 16 << 10 << 10,
+				wantCallbacks:    0,
 			},
 			{
 				desc:             "small chunksize rounds up to 256kib",
 				chunksize:        1,
 				wantBytesPerCall: 256 << 10,
+				wantCallbacks:    int(math.Ceil(float64(objSize) / (256 << 10))),
```
tritone, line 2276: This is confusing and opaque, just put the literal number of expected calls here I think.

tritone (approved):
One other test thing, otherwise lgtm

storage/integration_test.go
BrennaEpp: Opened https://github.com/googleapis/google-cloud-go/issues/7839 to keep track

Issue #7800, opened 2023-04-20
datastore: *Client.RunAggregationQuery() panics


MarioGravel:
**Client**

datastore

**Environment**

Cloud Run

**Go Environment**

$ go version
go version go1.19.7 windows/amd64

$ go env
set GO111MODULE=on
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\xxxxxxxx\AppData\Local\go-build
set GOENV=C:\Users\xxxxxxxx\AppData\Roaming\go\env
set GOEXE=.exe
set GOEXPERIMENT=
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOINSECURE=
set GOMODCACHE=C:\Users\xxxxxxxx\go\pkg\mod
set GONOPROXY=
set GONOSUMDB=
set GOOS=windows
set GOPATH=C:\Users\xxxxxxxx\go
set GOPRIVATE=
set GOPROXY=https://proxy.golang.org,direct
set GOROOT=C:\Program Files\Go
set GOSUMDB=sum.golang.org
set GOTMPDIR=
set GOTOOLDIR=C:\Program Files\Go\pkg\tool\windows_amd64
set GOVCS=
set GOVERSION=go1.19.7
set GCCGO=gccgo
set GOAMD64=v1
set AR=ar
set CC=gcc
set CXX=g++
set CGO_ENABLED=1
set GOMOD=D:\xxxxxxxx\go.mod
set GOWORK=
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
set CGO_LDFLAGS=-g -O2
set PKG_CONFIG=pkg-config
set GOGCCFLAGS=-m64 -mthreads -fno-caret-diagnostics -Qunused-arguments -Wl,--no-gc-sections -fmessage-length=0 -fdebug-prefix-map=C:\Users\xxxxxxxx\AppData\Local\Temp\go-build961903078=/tmp/go-build -gno-record-gcc-switches

**Code**

```go
func GetCrowdSizeByRoom(dsClient *datastore.Client, room *Room) int {
	query := datastore.NewQuery("Human").FilterField("Room", "=", room.Key).NewAggregationQuery().WithCount("Size")
        // Next line panics
	size, err := dsClient.RunAggregationQuery(context.Background(), query)
	if err != nil {
[...]

PR #7803, opened 2023-04-20
fix(datastore): datastoreClient.RunAggregationQuery

MarioGravel:
fix(datastore): PKG:datastore TYPE:datastoreClient FUNC:RunAggregationQuery

Bug:
See bug https://github.com/googleapis/google-cloud-go/issues/7800 

Calling `func (c *datastore.Client) RunAggregationQuery` panics at line `789` when calling `c.client.RunAggregationQuery`. `c.client` is defined `type pb.DatastoreClient interface` and it should have `RunAggregationQuery(ctx context.Context, in *RunAggregationQueryRequest, opts ...grpc.CallOption) (*RunAggregationQueryResponse, error)` but, `c.client` is implemented as a `type datastoreClient struct` and is missing `func (dc *datastoreClient) RunAggregationQuery(ctx context.Context, in *pb.RunAggregationQueryRequest, opts ...grpc.CallOption) (res *pb.RunAggregationQueryResponse, err error)`.

This bug was covered by the fact that `datastoreClient` complies to `type pb.DatastoreClient interface` because the interface is part of it struct declaration:
```go
// From /datastore/client.go, line 36
...
type datastoreClient struct {
	// Embed so we still implement the DatastoreClient interface,
	// if the interface adds more methods.
	pb.DatastoreClient
...
```

Also(Bonus):
`func (c *datastore.Client) RunAggregationQuery(ctx context.Context, aq *AggregationQuery) (AggregationResult, error)` returning an `type AggregationResult map[string]interface{}` but the content of type `interface{}` are `*datastore.Value` from `google.golang.org/genproto/googleapis/datastore/v1/entity.pb.go`. Not very useful.

Solution:
- Add `func (dc *datastoreClient) RunAggregationQuery(ctx context.Context, in *pb.RunAggregationQueryRequest, opts ...grpc.CallOption) (res *pb.RunAggregationQueryResponse, err error)` 
- Add a `switch` statement to use concrete types(when possible) thru `GetXXXValue() YYY` of `google.golang.org/genproto/googleapis/datastore/v1 :: *datastore.Value`


alon-ne:
Hi, I'm also encountering this issue. Panic at line 789 just like @MarioGravel saw

trifox (approved):

bhshkh (approved):

bhshkh (commented):
Please add unit and integration tests.

Crystaleclipspawl (approved):

bhshkh (approved):

bhshkh:
Reverting the code changes to return concrete type in https://github.com/googleapis/google-cloud-go/pull/8167.
To get concrete type from aggregation result, cast the returned value to pb.Value and then call GetXXX
e.g. 

```
	count, _ := res[countAlias]
	pbVal := count.(*pb.Value)
	intVal := pbVal.GetIntegerValue()
	fmt.Println("Count is %v", intVal)
```

Returning concrete types will break existing code of other consumers of the library.

datastore/client.go
```diff
@@ -73,6 +73,17 @@ func (dc *datastoreClient) RunQuery(ctx context.Context, in *pb.RunQueryRequest,
...
 
+func (dc *datastoreClient) RunAggregationQuery(ctx context.Context, in *pb.RunAggregationQueryRequest, opts ...grpc.CallOption) (res *pb.RunAggregationQueryResponse, err error) {
+	ctx = trace.StartSpan(ctx, "cloud.google.com/go/datastore.datastoreClient.RunAggregationQuery")
+	defer func() { trace.EndSpan(ctx, err) }()
+
+	err = dc.invoke(ctx, func(ctx context.Context) error {
+		res, err = dc.c.RunAggregationQuery(ctx, in, opts...)
+		return err
+	})
+	return res, nil
```
rd-steffen-tellhelm, line 84: This code never returns an error, I think this shall be: `return res, err`

rd-steffen-tellhelm: For anyone facing the same issue: this is especially annoying when you use testcontainers for firestore in datastore mode that currently not support aggregated queries and get a nil panic instead of the proper error that the function is not (yet) implemented.

bhshkh: Created https://github.com/googleapis/google-cloud-go/pull/8222 to return error

PR #7807, opened 2023-04-20
chore: add go.work

codyoss:
This should allow us better control over dependencies and opening up multi-module workspaces in editors.

Ran the following commands to generate the changes:

- go work init
- go work use -r .
- go work sync
- go mod tidy (everywhere)

Fixes: #7750


quartzmo (approved):
Awesome, let's try it!

PR #7819, opened 2023-04-21
chore(internal/postprocessor): remove moving of snippets

quartzmo:
* Replace MoveSnippets with UpdateSnippetsMetadata

Depends on https://github.com/googleapis/gapic-generator-go/releases/tag/v0.35.7.


internal/postprocessor/main.go
```diff
@@ -309,70 +309,47 @@ func (p *postProcessor) getDirs() []string {
...
-					return err
-				}
-				if err := os.Rename(snpDir, toDir); err != nil {
-					return err
-				}
-			} else {
-				return err
-			}
-		}
-		if strings.Contains(clientRelPath, "debugger/apiv2") {
```
codyoss, line 349: can you add this hack back in. This is the one client we generate that is not in a module so it does not have version info to read today.

quartzmo: done

Issue #6033, opened 2022-05-15
bigquery: storage write api support for streaming to tens of thousands of tables easily (in a single dataset)


johnmck24:
**Is your feature request related to a problem? Please describe.**
We are inserting ~20,000 rows per second to ~200,000 different tables (spread relatively equally). Each table is used to store a user activity on our service (they are daily-partitioned using a custom timestamp column, and have specific settings like partition expiration that differs between users/tables).

We are experimenting migrating from the legacy `.insertAll` streaming insert API to the new storage write API, and the new API doesn't seem to be designed for mass-inserting to many different tables. In our case, this would mean opening tens of thousands of streams for a small write throughput on each, and spreading those over hundreds of connections. Aside from resulting in a very complex custom load-balanced implementation, this just doesn't seem to be a fit for a many-tables streaming insertion scenario, and won't scale as we grow.

**Describe the solution you'd like**
Ideally, we would want to use a stream to insert in different tables in a dataset (or any dataset), the same way the legacy `.insertAll` API works where an HTTP/2 stream is used to POST rows to any table.

**Describe alternatives you've considered**
Aside from sticking with the legacy `.insertAll` streaming insert API, we don't see any non-legacy alternatives.

**Additional context**
Our user base is growing rapidly, so any workaround that would work for 200,000 tables but not 500,000 tables is a no-go for us.

shollyman:
We've been looking into a multiplexing feature for this use case with the write API.  One of the interesting complexities is schema management.  You mention there's a variety of partitioning/expiration settings, but what's the cardinality on schema diversity?  For the 200k tables, are we dealing with 200k distinct schemas?  Does a single writer tend to write messages of the same schema to a set of tables, or would a single writer be responsible for a both many tables and schemas?

shollyman:
cc @yirutang

shollyman:
Marking this as a P3 feature request for now in the client.  We need support from the backend to move forward on this, hence the lower priority.

yirutang:
@GaoleMeng

derekperkins:
This is an issue for us too. We have identical schema datasets per customer and are still using legacy streaming

shollyman:
Happily closing this request.  With https://github.com/googleapis/google-cloud-go/releases/tag/bigquery/v1.51.0, you can leverage connection multiplexing to share a single network connection between multiple/many default stream writers.

See the package documentation for more information:
[...]

PR #7848, opened 2023-04-27
docs(bigquery/storage/managedwriter):  improve tuning guidance

shollyman:
Based on feedback from @derekperkins, revisit how we talk about connection pooling options.

### DIFF
--- a/bigquery/storage/managedwriter/doc.go
+++ b/bigquery/storage/managedwriter/doc.go
@@ -246,5 +246,8 @@ multiplexing or not) may also wish to consider further tuning of this behavior. 
 sets a reasonable default, but this can be tuned further by leveraging the WithGRPCConnectionPool ClientOption,
 documented here:
 https://pkg.go.dev/google.golang.org/api/option#WithGRPCConnectionPool
+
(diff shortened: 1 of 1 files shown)


derekperkins:
Are you opposed to listing some specific scenarios, or giving some more explicit guidance like you did for this?
[`WithGRPCConnectionPool = WithMultiplexPoolLimit + expected # of explicit streams`](https://github.com/googleapis/google-cloud-go/issues/6033#issuecomment-1516613382)

shollyman:
> Are you opposed to listing some specific scenarios, or giving some more explicit guidance like you did for this? [`WithGRPCConnectionPool = WithMultiplexPoolLimit + expected # of explicit streams`](https://github.com/googleapis/google-cloud-go/issues/6033#issuecomment-1516613382)

I'm not opposed to it, but I would like to see more concrete feedback from users with various workload shapes before recommending specific values.  The alternative is to do more profiling with synthetic workloads, but that can yield unrealistic results.  The implicit tradeoff we're approaching here is effectively sizing number of clients vs number of writers per client.

codyoss (approved):

PR #7856, opened 2023-04-28
chore(internal/postprocessor): add support for root dir OwlBot dest

quartzmo:
* Fixes post-processor for OwlBot.yaml paths to root dir / for snippets

refs: #7850

### DIFF
--- a/internal/postprocessor/config.go
+++ b/internal/postprocessor/config.go
@@ -102,11 +102,9 @@ func (p *postProcessor) loadConfig() error {
 		}
 		if li.ImportPath == "" {
 			li.ImportPath = v.Source[i+1:]
-			fmt.Printf("set li.ImportPath from v.Source: %s", v.Source)
 		}
 		if li.RelPath == "" {
 			li.RelPath = strings.TrimPrefix(li.ImportPath, "cloud.google.com/go")
-			fmt.Printf("set li.RelPath from li.ImportPath: %s", li.ImportPath)
 		}
 		c.ClientRelPaths = append(c.ClientRelPaths, li.RelPath)
 	}
--- a/internal/postprocessor/config.go
+++ b/internal/postprocessor/config.go
@@ -56,6 +56,7 @@ func (p *postProcessor) loadConfig() error {
 			InputDirectory string `yaml:"input-directory"`
 			ServiceConfig  string `yaml:"service-config"`
 			ImportPath     string `yaml:"import-path"`
+			RelPath        string `yaml:"rel-path"`
 		} `yaml:"service-configs"`
 		ManualClients []*ManifestEntry `yaml:"manual-clients"`
 	}
@@ -90,17 +91,24 @@ func (p *postProcessor) loadConfig() error {
 		c.GoogleapisToImportPath[v.InputDirectory] = &libraryInfo{
 			ServiceConfig: v.ServiceConfig,
 			ImportPath:    v.ImportPath,
+			RelPath:       v.RelPath,
 		}
 	}
 	for _, v := range owlBotConfig.DeepCopyRegex {
-		c.ClientRelPaths = append(c.ClientRelPaths, v.Dest)
 		i := strings.Index(v.Source, "/cloud.google.com/go")
 		li, ok := c.GoogleapisToImportPath[v.Source[1:i]]
 		if !ok {
 			return fmt.Errorf("unable to find value for %q, it may be missing a service config entry", v.Source[1:i])
 		}
-		li.ImportPath = v.Source[i+1:]
-		li.RelPath = v.Dest
+		if li.ImportPath == "" {
+			li.ImportPath = v.Source[i+1:]
+			fmt.Printf("set li.ImportPath from v.Source: %s", v.Source)
+		}
+		if li.RelPath == "" {
+			li.RelPath = strings.TrimPrefix(li.ImportPath, "cloud.google.com/go")
+			fmt.Printf("set li.RelPath from li.ImportPath: %s", li.ImportPath)
+		}
+		c.ClientRelPaths = append(c.ClientRelPaths, li.RelPath)
 	}
 	p.config = c
 	return nil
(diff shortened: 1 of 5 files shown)


internal/postprocessor/config.go
```diff
@@ -90,17 +91,24 @@ func (p *postProcessor) loadConfig() error {
...
 		i := strings.Index(v.Source, "/cloud.google.com/go")
 		li, ok := c.GoogleapisToImportPath[v.Source[1:i]]
 		if !ok {
 			return fmt.Errorf("unable to find value for %q, it may be missing a service config entry", v.Source[1:i])
 		}
-		li.ImportPath = v.Source[i+1:]
-		li.RelPath = v.Dest
+		if li.ImportPath == "" {
+			li.ImportPath = v.Source[i+1:]
+			fmt.Printf("set li.ImportPath from v.Source: %s", v.Source)
```
codyoss, line 105: This may be noisey, do we want these? If so, they should use log instead of fmt

internal/postprocessor/config.yaml
```diff
@@ -364,6 +364,8 @@ service-configs:
     service-config: dataplex_v1.yaml
   - input-directory: google/cloud/dataproc/v1
     service-config: dataproc_v1.yaml
+    import-path: cloud.google.com/go/dataproc/v2/apiv1
```
codyoss, line 367: Will we always need the import path now in this file? If so should we calculate it, or make it an explicit config?

internal/postprocessor/main.go
```diff
@@ -315,6 +315,10 @@ func (p *postProcessor) UpdateSnippetsMetadata() error {
 		// OwlBot dest relative paths in ClientRelPaths begin with /, so the
 		// first path segment is the second element.
 		moduleName := strings.Split(clientRelPath, "/")[1]
+		if moduleName == "" {
+			return errors.New("UpdateSnippetsMetadata: moduleName is empty")
```
codyoss, line 319: nit: the prefix colon is usual done with go package names. Also errors should not start with a capital letter. Maybe provide more context instead. `fmt.Errorf("unable to parse module name for %v", clientRelPath)`

internal/postprocessor/main.go
```diff
@@ -315,6 +315,10 @@ func (p *postProcessor) UpdateSnippetsMetadata() error {
...
+		}
+		log.Println("UpdateSnippetsMetadata: ", clientRelPath)
```
codyoss, line 321: nit: we already know the function we are in given the top level logging in the function. Maybe all of these log lines could be replaced by one at the top of the loop the dumps clientRelPath? Or errors from the function should return this context perhaps. Thoughts?

codyoss (changes requested):

internal/postprocessor/main.go
quartzmo: Fixed, thanks!

internal/postprocessor/main.go
quartzmo: This line is needed in order to debug configuration errors that break the preceding lines. Without it, maintainers will have no idea which client is responsible for the error from the line above. Removed the prefix, thanks!

internal/postprocessor/config.go
quartzmo: These were development debug statements that I forgot to remove. Fixed, thanks!

internal/postprocessor/config.yaml
quartzmo: When client paths in `.OwlBot.yaml` are changed to copy to the root directory of google-cloud-go, as here:

```yaml
  - source: /google/cloud/confidentialcomputing/v1/cloud.google.com/go
    dest: /
```

Then `import-path` must be added to `config.yaml`, since this information was lost when `source` and `dest` was truncated, above.

However, in this PR I am only changing a small number of these client configurations manually. I will change the remaining entries in a follow-up PR.

codyoss: Sure, in the follow up PR, we should consider making this import path inferred, I think we could read it from a proto file since we know the source dir. I think the less config we have the better and more likely that things will not break from typos 😄

internal/postprocessor/main.go
```diff
@@ -316,9 +316,9 @@ func (p *postProcessor) UpdateSnippetsMetadata() error {
 		// first path segment is the second element.
 		moduleName := strings.Split(clientRelPath, "/")[1]
 		if moduleName == "" {
-			return errors.New("UpdateSnippetsMetadata: moduleName is empty")
+			return fmt.Errorf("unable to parse module name for %v", clientRelPath)
 		}
-		log.Println("UpdateSnippetsMetadata: ", clientRelPath)
+		log.Println(clientRelPath)
```
codyoss, line 321: nit: maybe a little more context here, like add the word processing or something?

codyoss (approved):
Would you update the sha again too

internal/postprocessor/main.go
quartzmo: I moved the statement after the filtering for `debugger` etc, and changed the module name to the metadata file glob path. Sample output when run locally in `internal/postprocessor`:

```sh
2023/05/01 12:08:26 updating snippets metadata
2023/05/01 12:08:26 updating  ../../internal/generated/snippets/accessapproval/apiv1/snippet_metadata.*.json
2023/05/01 12:08:26 updating  ../../internal/generated/snippets/accesscontextmanager/apiv1/snippet_metadata.*.json
...
2023/05/01 12:08:26 updating  ../../internal/generated/snippets/compute/apiv1/snippet_metadata.*.json
2023/05/01 12:08:26 skipping, file not found with glob:  ../../internal/generated/snippets/confidentialcomputing/apiv1/snippet_metadata.*.json
2023/05/01 12:08:26 updating  ../../internal/generated/snippets/confidentialcomputing/apiv1alpha1/snippet_metadata.*.json
2023/05/01 12:08:26 updating  ../../internal/generated/snippets/contactcenterinsights/apiv1/snippet_metadata.*.json
2023/05/01 12:08:26 updating  ../../internal/generated/snippets/container/apiv1/snippet_metadata.*.json
```

PR #7860, opened 2023-05-01
chore(orgpolicy): remove broken version file

noahdietz:
Removes `version.go` file that should've been cleaned up when the `orgpolicy/apiv1` client was removed.


quartzmo (approved):

noahdietz:
There is no breaking change being reported by `apidiff / apidiff (orgpolicy/apiv1)`, it is just a broken package that can't be compiled without this change.

Issue #7636, opened 2023-03-28
firestore: Bulkwriter not doing what writebatch used to do. Can't write 2 fields to same doc if one is a value and other a Servertimestamp at same time, for ex.


diver2:
The straightforward code below also returns the same error, due to the same reason.
```go
func main() {
	ctx := context.Background()
	app, err := firebase.NewApp(ctx, nil)
	if err != nil {
		log.Fatalf("error initializing app: %v\n", err)
	}

	client, err12 := app.Firestore(ctx)
	if err12 != nil {
		log.Fatalf("app.Firestore: %v", err12)
	}
	map1 := map[string]interface{}{}
	map1["answer"] = 42
	map1["createdAt"] = firestore.ServerTimestamp
	client.Collection("AATEST").Doc("SETDOC").Set(ctx, map1) // THIS WORKS

	bulkWriter := client.BulkWriter(ctx)
	_, errBulkBUGS := bulkWriter.Set(
		client.Collection("AATEST").Doc("SETDOCVIABULK"),
		map1,
	)
	if errBulkBUGS != nil {
		fmt.Printf("errBulkBUGS = %v\n", errBulkBUGS)
	}
	bulkWriter.End()
	return
}
```
returns: errBulkBUGS = firestore: too many writes sent to bulkwriter

_Originally posted by @diver2 in https://github.com/googleapis/google-cloud-go/issues/7558#issuecomment-1472735595_

diver2:
See also - 
https://github.com/googleapis/google-cloud-go/issues/7558#issuecomment-1472723266

telpirion:
cc @bcoe and @codyoss 

Hello @diver2 

Please be sure to review our [CODE_OF_CONDUCT](https://github.com/googleapis/google-cloud-go/blob/main/CODE_OF_CONDUCT.md) before responding. I will attempt to answer your questions one-by-one.

1. The `Bulkwriter.Set()` method will evaluate the `map` that you've provided as two separate writes--one for the key "answer" and another for "createdAt". If you comment out the line `map["createdAt"] = firestore.ServerTimestamp`, then the error doesn't happen. 
**Action item**: This definitely could be clearer in the documentation.
[...]

PR #7864, opened 2023-05-02
fix(firestore): Replace usage of transform with update_transform in batch write 

bhshkh:
Replacing all usages of transforms with update_transforms will enable writes with field transforms to be performed in a single write. 

This is important for BulkWriter to ensure that writes with field transforms do not have to be split into separate batches.

This will resolve https://github.com/googleapis/google-cloud-go/issues/7636


firestore/integration_test.go
```diff
@@ -1721,6 +1721,19 @@ func TestIntegration_ColGroupRefPartitionsLarge(t *testing.T) {
 		t.Errorf("Unexpected number of documents across partitions: got %d, want %d", got, want)
 	}
 }
+func TestIntegration_BulkWriter_Set(t *testing.T) {
```
enocom, line 1724: A little comment above this test to explain what it's doing might be nice here.

enocom (approved):

kolea2 (approved):

PR #7872, opened 2023-05-02
chore(storage): fix progress call in gRPC Writer

### LINKED ISSUE #7798: storage: determine what to do on chunksize=0 [GRPC]

### LINKED ISSUE #7839: storage: Writer.ProgressFunc is always called at least once [GRPC]
The progress function is only called in HTTP if the upload is resumable (so it is not called at all for a one-shot), in gRPC the function is called at least once even if the upload is not resumable.

The docs for the progress func specify it gets called only if the upload requires more than one call, so we should probably modify gRPC to comply with this.

tritone:
Fixes #7839


BrennaEpp (approved):
Looks good, just note that this fixes #7839 not #7798

tritone:
> Looks good, just note that this fixes #7839 not #7798

Whoops, good catch

PR #7881, opened 2023-05-04
docs(storage): WithXMLReads should mention XML instead of JSON API

v0lkc:
I've spotted a copy/paste issue in the description of `storage.WithXMLReads`. It states it would use the JSON API instead of XML, which is obviously wrong.


BrennaEpp (approved):
Thank you for noticing and opening a PR!

tritone (approved):

PR #7883, opened 2023-05-04
chore(internal/postprocessor): migrate all clients to OwlBot snippets

quartzmo:
* Add /internal/generated/snippets to deep-remove-regex
* Remove relative path from both source and dest, to include snippets in scope
* Add import-path for all service-configs in config.yaml


.github/.OwlBot.yaml
```diff
@@ -113,9 +113,227 @@ deep-remove-regex:
...
+  - /internal/generated/snippets/bigquery/storage/apiv1beta1/
+  - /internal/generated/snippets/bigquery/storage/apiv1beta2/
+  - /internal/generated/snippets/billing/apiv1/
+  - /internal/generated/snippets/billing/budgets/apiv1/
+  - /internal/generated/snippets/billing/budgets/apiv1beta1/
+  - /internal/generated/snippets/binaryauthorization/apiv1/
+  - /internal/generated/snippets/binaryauthorization/apiv1beta1/
+  - /internal/generated/snippets/certificatemanager/apiv1/
+  - /internal/generated/snippets/channel/apiv1/
+  # - /internal/generated/snippets/cloudbuild/apiv1/v2/
```
noahdietz, line 168: Can you include a comment as to why this is commented out, please.

.github/.OwlBot.yaml
```diff
@@ -113,9 +113,227 @@ deep-remove-regex:
...
+  - /internal/generated/snippets/dataform/apiv1beta1/
+  - /internal/generated/snippets/datafusion/apiv1/
+  - /internal/generated/snippets/datalabeling/apiv1beta1/
+  - /internal/generated/snippets/dataplex/apiv1/
+  - /internal/generated/snippets/dataproc/apiv1/
+  - /internal/generated/snippets/dataqna/apiv1alpha/
+  - /internal/generated/snippets/datastore/admin/apiv1/
+  - /internal/generated/snippets/datastream/apiv1/
+  - /internal/generated/snippets/datastream/apiv1alpha1/
+  # - /internal/generated/snippets/debugger/apiv2/
```
noahdietz, line 194: Same

.github/.OwlBot.yaml
```diff
@@ -113,9 +113,227 @@ deep-remove-regex:
...
+  - /internal/generated/snippets/dialogflow/apiv2beta1/
+  - /internal/generated/snippets/dialogflow/cx/apiv3/
+  - /internal/generated/snippets/dialogflow/cx/apiv3beta1/
+  - /internal/generated/snippets/discoveryengine/apiv1beta/
+  - /internal/generated/snippets/dlp/apiv2/
+  - /internal/generated/snippets/documentai/apiv1/
+  - /internal/generated/snippets/documentai/apiv1beta3/
+  - /internal/generated/snippets/domains/apiv1beta1/
+  - /internal/generated/snippets/edgecontainer/apiv1/
+  # - /internal/generated/snippets/errorreporting/apiv1beta1/
```
noahdietz, line 206: Same

.github/.OwlBot.yaml
```diff
@@ -113,9 +113,227 @@ deep-remove-regex:
...
+  - /internal/generated/snippets/essentialcontacts/apiv1/
+  - /internal/generated/snippets/eventarc/apiv1/
+  - /internal/generated/snippets/eventarc/publishing/apiv1/
+  - /internal/generated/snippets/filestore/apiv1/
+  - /internal/generated/snippets/firestore/apiv1/
+  # - /internal/generated/snippets/firestore/apiv1/admin/
```
noahdietz, line 212: Same

.github/.OwlBot.yaml
```diff
@@ -113,9 +113,227 @@ deep-remove-regex:
...
+  - /internal/generated/snippets/support/apiv2/
+  - /internal/generated/snippets/talent/apiv4/
+  - /internal/generated/snippets/talent/apiv4beta1/
+  - /internal/generated/snippets/texttospeech/apiv1/
+  - /internal/generated/snippets/tpu/apiv1/
+  - /internal/generated/snippets/trace/apiv1/
+  - /internal/generated/snippets/trace/apiv2/
+  - /internal/generated/snippets/translate/apiv3/
+  - /internal/generated/snippets/video/livestream/apiv1/
+  # - /internal/generated/snippets/video/stitcher/apiv1/
```
noahdietz, line 317: Same

.github/.OwlBot.yaml
```diff
@@ -267,445 +485,447 @@ deep-preserve-regex:
...
-  - source: /google/cloud/tasks/v2/cloud.google.com/go/cloudtasks/apiv2
-    dest: /cloudtasks/apiv2
-  - source: /google/cloud/tasks/v2beta2/cloud.google.com/go/cloudtasks/apiv2beta2
-    dest: /cloudtasks/apiv2beta2
-  - source: /google/cloud/tasks/v2beta3/cloud.google.com/go/cloudtasks/apiv2beta3
-    dest: /cloudtasks/apiv2beta3
-  - source: /google/cloud/compute/v1/cloud.google.com/go/compute/apiv1
-    dest: /compute/apiv1
+  - source: /google/cloud/accessapproval/v1/cloud.google.com/go
+    dest: /
```
noahdietz, line 489: Would it be possible to make this the default value for `dest` for Go, perhaps? So we don't need this for every entry. If that is a bigger change elsewhere, we can come back to it

noahdietz (commented):
I see you are just copying some already commented out entries. If you have any idea why they are commented out, it would be great to include why in here, but no problem if you don't know.

.github/.OwlBot.yaml
quartzmo: I need to ask @codyoss exactly what the conditions are to uncomment these. Since he's not in today, let's just leave them undocumented in this PR.

.github/.OwlBot.yaml
quartzmo: I checked the [OwlBot README](https://github.com/googleapis/repo-automation-bots/blob/main/packages/owl-bot) as well as [config-files.ts](https://github.com/googleapis/repo-automation-bots/blob/main/packages/owl-bot/src/config-files.ts), but I didn't see any support for default values.

.github/.OwlBot.yaml
noahdietz: Then no need to block this PR on that.

.github/.OwlBot.yaml
noahdietz: Sounds like a feature request! Do you think it would be good to do? If so, can you file a feature request against owlbot?

Of course we shouldn't block this PR on completion of that.

noahdietz (approved):

PR #7884, opened 2023-05-05
chore(internal/actions): refactor changefinder

noahdietz:
Refactors `changefinder` to support nested submodules i.e. `vision/v2`.

Changes the default compare the current `HEAD` against `origin/main`.

Add a `-q` quiet mode, all logs will be buffered and dumped in the event of an error.

Add a `-dir` flag to specify a different root dir than the default of `cwd`.

Add a `-format` flag with values `plain` (stdout, new line delimited), and `github` (prints github action output). Defaults to `plain`.

Add a `-gh-var` flag that accepts the name of the GitHub variable the output should set,  defaults to `submodules`.

This will be able to replace the git-fu used in the `apidiff` workflow and it can power a "change description" action, basically the inverse of the `apidiff` workflow that reports the _added_ surfaces.

Try it out by running the following:
```
cd $GOOGLE_CLOUD_GO
git checkout origin owl-bot-copy
go run ./internal/actions/cmd/changefinder -q
```


quartzmo:
I wasn't aware of changefinder before this PR. It doesn't appear to have any documentation whatsoever. What do you think about adding a short README or even just a few sentences of function docs for `main`? The description for this PR has some useful content.

noahdietz:
> I wasn't aware of changefinder before this PR. It doesn't appear to have any documentation whatsoever. What do you think about adding a short README or even just a few sentences of function docs for `main`? The description for this PR has some useful content.

Great idea.

Honestly, I wasn't aware of it either until I started to consider writing the exact same thing and bumped into it while starting to do so :)

quartzmo:
I think this module and file are currently orphaned. It was added in #4537 and referenced only in `.github/workflows/release-submodule.yaml`, which no longer exists.

noahdietz:
> I think this module and file are currently orphaned. It was added in #4537 and referenced only in `.github/workflows/release-submodule.yaml`, which no longer exists.

Well I'm glad it wasn't removed, because it's quite useful!

quartzmo:
Can you add your new usage(s) to this PR?

noahdietz:
> Can you add your new usage(s) to this PR?

If it is OK, I'd prefer to make that change in a separate PR in case I need to revert it (i.e. i break the `apidiff` workflow).

quartzmo (approved):

noahdietz:
Add a README!

PR #7899, opened 2023-05-09
feat(spanner): enable client to server compression

rahul2393:
Enable compression of network traffic from client to server.


spanner/client.go
```diff
@@ -263,13 +276,17 @@ func NewClientWithConfig(ctx context.Context, database string, config ClientConf
...
-func allClientOpts(numChannels int, userOpts ...option.ClientOption) []option.ClientOption {
+func allClientOpts(numChannels int, compression string, userOpts ...option.ClientOption) []option.ClientOption {
 	generatedDefaultOpts := vkit.DefaultClientOptions()
 	clientDefaultOpts := []option.ClientOption{
 		option.WithGRPCConnectionPool(numChannels),
 		option.WithUserAgent(fmt.Sprintf("spanner-go/v%s", internal.Version)),
 		internaloption.EnableDirectPath(true),
 	}
+	if compression == "gzip" {
+		clientDefaultOpts = append(clientDefaultOpts, option.WithGRPCDialOption(grpc.WithDefaultCallOptions(
```
olavloite, line 287: Would it be possible to add a test with the mock Spanner server that verifies that we actually receive this header?

olavloite (approved):
LGTM (with a potential request for a test), assuming that it has been verified that the setting actually has an effect (e.g. by looking at the actual number of bytes sent/received)

spanner/client.go
rahul2393: Added the unit test for checking the request headers, also adding screenshot to verify all the grpc requests will have the call option for setting compression 
Screenshot 2023-05-19 at 11 55 13 AM

olavloite: Thanks for the extra info and adding the test. LGTM!

PR #7940, opened 2023-05-15
feat(pubsub): expose common errors for easier handling

### LINKED ISSUE #7925: pubsub: publish error that requires ResumePublish should wrap and return a specific exported error

When publishing with an ordering key you may end up in a situation where you're required to [`ResumePublish` because the scheduler was paused](https://github.com/googleapis/google-cloud-go/blob/74e68952dafa226b2c51f7f60c38637f05ebbc51/pubsub/topic.go#L719). This would be much easier to detect and handle if instead of:
```go
err = fmt.Errorf("pubsub: Publishing for ordering key, %s, paused due to previous error. Call topic.ResumePublish(orderingKey) before resuming publishing", orderingKey)
```
we had something like:
```go
// first
var ErrPublishingPaused = errors.New("Call topic.ResumePublish(orderingKey) before resuming publishing")
// and then
err = fmt.Errorf("pubsub: Publishing for ordering key, %s, paused due to previous error. %w", orderingKey, ErrPublishingPaused)
```

In the example above I believe there's a good enough, fully backwards compatible and idiomatic solution. Similar implementations could be proposed that don't rely on `errors.New` if needed.

hongalex:
Fixes #7925


pubsub/topic.go
```diff
@@ -535,7 +535,7 @@ func (t *Topic) Subscriptions(ctx context.Context) *SubscriptionIterator {
 	}
 }
 
-var errTopicStopped = errors.New("pubsub: Stop has been called for this topic")
+var ErrTopicStopped = errors.New("pubsub: Stop has been called for this topic")
```
noahdietz, line 538: Can you add a comment to this exported variable?

pubsub/topic.go
```diff
@@ -548,6 +548,8 @@ var errTopicStopped = errors.New("pubsub: Stop has been called for this topic")
 //	}
 type PublishResult = ipubsub.PublishResult
 
+var ErrTopicOrderingNotEnabled = errors.New("Topic.EnableMessageOrdering=false, but an OrderingKey was set in Message. Please remove the OrderingKey or turn on Topic.EnableMessageOrdering")
```
noahdietz, line 551: Ditto, please add a comment.

codyoss, line 551: nit: error message should start with a lower case letter. Also other errors in this package seem to prefix with pubsub. Should we do that here too?

codyoss (commented):
Are both of these errors we expect people to handle in their code or are they informational. I would say only expose these if you expect people to check for this error and take an action in code.

pubsub/topic.go
hongalex: As is, this is to maintain complete backwards compatibility (the error message is completely the same in this case). We didn't standardize on prefixing with pubsub.

I also struggled with fixing this for `ErrPublishingPaused`, which starts in the middle of the previous error message to allow for error wrapping.

hongalex:
> Are both of these errors we expect people to handle in their code or are they informational. I would say only expose these if you expect people to check for this error and take an action in code.

Good question. For `ErrTopicStopped`, that topic is fully stopped and needs to be re-initialized before calling `Publish` again. Otherwise, the application can't continue. I suspect that this could actually be used the code (need to record that the Publish failed in this way)

For `ErrTopicOrderingNotEnabled`, this is probably more informational and wouldn't require any action to be taken in the code. I'll remove this.

hongalex:
@codyoss @noahdietz I switched `ErrPublishingPaused` to a custom error that holds onto the ordering key. This leads to a slightly more elegant solution in that the string isn't awkwardly defined in two places, but now requires comparing with `errors.As` instead of `errors.Is` (since the ordering key can't be known beforehand). Thoughts on this approach instead?

codyoss (approved):
LGTM

Issue #7348, opened 2023-02-03
vision and recaptchaenterprise clients need v2 module moved to root post 1.17 upgrade


codyoss:
After this they can be onboarded to owlbot more easily.

PR #7941, opened 2023-05-15
chore: move vision v2 to in-place module strategy

codyoss:
This change moves the sources from the v2 folder up a dir and changes the name of the module in vision. The v2 client is now just up a directory and we are no longer storing the source for v1 of the module.

Updates: #7348


codyoss:
This will show up a a breaking change, but it is not really.

quartzmo (approved):

PR #7955, opened 2023-05-17
chore: move recaptcha to in-place versioning

codyoss:


codyoss:
This is actually a non-breaking change

quartzmo (approved):

PR #7962, opened 2023-05-18
chore(all): Use fieldmask types directly instead of field_mask genproto alias

liggitt:
related to https://github.com/kubernetes/kubernetes/issues/113366#issuecomment-1292400083

cc @codyoss


quartzmo:
Replaced by #8029, #8030, #8031, #8032

PR #7977, opened 2023-05-22
feat(pubsub): add support for cloud storage subscriptions

hongalex:
Add support for subscriptions that pushes messages to Cloud Storage.


pubsub/pstest/fake.go
```diff
@@ -504,6 +504,11 @@ func (s *GServer) CreateSubscription(_ context.Context, ps *pb.Subscription) (*p
 	} else if ps.BigqueryConfig.Table != "" {
 		ps.BigqueryConfig.State = pb.BigQueryConfig_ACTIVE
 	}
+	if ps.CloudStorageConfig == nil {
+		ps.CloudStorageConfig = &pb.CloudStorageConfig{}
```
shollyman, line 508: why the empty submessage?  Not populating the message would be my expectation.  This would change some of your expectations elsewhere (and require the typical nil check code for accessing inner fields).

pubsub/subscription.go
```diff
@@ -786,14 +936,21 @@ func (s *Subscription) Config(ctx context.Context) (SubscriptionConfig, error) {
...
 
-	// If non-nil, the bigquery config is changed. Cannot be set at the same time as PushConfig.
+	// If non-nil, the bigquery config is changed. At most one of PushConfig, BigQueryConfig, or CloudStorageConfig
+	// can be set.
 	// If currently in bigquery mode, set this value to the zero value to revert to a Pull based subscription,
 	BigQueryConfig *BigQueryConfig
 
+	// If non-nil, the Cloud Storage config is changed. At most one of PushConfig, BigQueryConfig, or CloudStorageConfig
+	// can be set.
+	// If currently in CloudStorage mode, set this value to the zero value to revert to a Pull based subscription,
```
lahuang4, line 951: Optional: "the zero value" sounds a bit odd, does it make sense to say "nil" instead? I realize we already have this wording in place for the PushConfig comment so feel free to push back on this.

pubsub/subscription.go
```diff
@@ -855,6 +1012,14 @@ func (s *Subscription) updateRequest(cfg *SubscriptionConfigToUpdate) *pb.Update
 		psub.BigqueryConfig = cfg.BigQueryConfig.toProto()
```
lahuang4, line 1012: Why don't we do the same empty-table check for the BQ config?

pubsub/subscription_test.go
```diff
@@ -497,6 +497,60 @@ func TestBigQuerySubscription(t *testing.T) {
...
+	defer client.Close()
+	defer srv.Close()
+
+	topic := mustCreateTopic(t, client, "t")
+	bucket := "fake-bucket"
+	csCfg := CloudStorageConfig{
+		Bucket:         bucket,
+		FilenamePrefix: "some-prefix",
+		FilenameSuffix: "some-suffix",
+		OutputFormat: &CloudStorageOutputFormatAvroConfig{
```
lahuang4, line 513: Can you also test out creating a Text output format subscription? You could update the subscription's output format to Text or create a new sub.

pubsub/subscription_test.go
```diff
@@ -497,6 +497,60 @@ func TestBigQuerySubscription(t *testing.T) {
...
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	got := cfg.CloudStorageConfig
+	want = CloudStorageConfig{}
+	if diff := testutil.Diff(got, want); diff != "" {
+		t.Fatalf("remove cloud storage subscription mismatch: \n%s", diff)
+	}
+
```
lahuang4, line 551: Nit: extra empty line

pubsub/pstest/fake.go
```diff
@@ -608,7 +611,6 @@ func (s *GServer) UpdateSubscription(_ context.Context, req *pb.UpdateSubscripti
 		case "bigquery_config":
 			// If bq config is nil here, it will be cleared.
 			// Otherwise, we'll consider the subscription active if any table is set.
-			sub.proto.BigqueryConfig = req.GetSubscription().GetBigqueryConfig()
```
lahuang4, line 611: Why remove this?

pubsub/subscription.go
hongalex: Zero value is correct here since nil means we're not trying to update he config. Setting a zero value (aka empty) CloudStorageConfig is how we're able to clear it. We translate this to a `nil` CloudStorageConfig before passing it into the update request.

pubsub/subscription.go
hongalex: We didn't before and that was what was causing a bug. We fixed it by checking for this properly in the `toProto()` method. I'll make sure CloudStorageConfig is updated in this way as well so we don't have to make these checks here.

pubsub/subscription_test.go
hongalex: Good point, I added an update test here.

pubsub/pstest/fake.go
hongalex: Good catch! reverted

lahuang4 (approved):

pubsub/pstest/fake.go
```diff
@@ -617,6 +620,14 @@ func (s *GServer) UpdateSubscription(_ context.Context, req *pb.UpdateSubscripti
 				}
 			}
 
+		case "cloud_storage_config":
+			sub.proto.CloudStorageConfig = req.GetSubscription().GetCloudStorageConfig()
```
shollyman, line 624: there's some invariant here that ensures req.GetSubscription() doesn't return nil?

pubsub/subscription.go
```diff
@@ -305,17 +419,23 @@ type SubscriptionConfig struct {
 	Topic *Topic
 
 	// If push delivery is used with this subscription, this field is
-	// used to configure it. Either `PushConfig` or `BigQueryConfig` can be set,
-	// but not both. If both are empty, then the subscriber will pull and ack
-	// messages using API methods.
+	// used to configure it. At most one of `PushConfig`, `BigQueryConfig`,
+	// and `CloudStorageConfig` can be set. If all are empty, then the
```
shollyman, line 423: s/and/or here and in the BigQueryConfig comments?

shollyman (approved):

pubsub/pstest/fake.go
hongalex: Correct, it's checked in the first lines above

PR #7980, opened 2023-05-23
fix(pubsub): allow clearing of topic schema

### LINKED ISSUE #7979: pubsub: Unable to update or clear SchemaSettings
The way `SchemaSettings` messages are added to `UpdateTopicRequest`s is incorrect. 

If you want to clear the `schema_settings` field on a topic, you pass a zero-valued `SchemaSettings` in the `TopicConfigToUpdate` struct per [this comment](https://github.com/GoogleCloudPlatform/golang-samples/blob/4996af5c9d2615b08abfe53f3176222134b697fd/pubsub/schemas/update_topic_schema.go#L38-L39C5)

However, in [this section](https://github.com/googleapis/google-cloud-go/blob/8d62d278a25effd0a0efb3977d94b6c4ee3636cf/pubsub/topic.go#L410C1-L433) of code which handles that field, it incorrectly passes the empty struct to the `UpdateTopicRequest` when it should be `nil` in the case where we wish to clear our schema. Also, when we wish to update a field in the `SchemaSettings`, we are neglecting to set the `schema_settings` update mask path. This section needs to be reworked to:

1) Have a `nil` `schema_settings` field and the `schema_settings` update mask path returned in the `UpdateTopicRequest` in the case where we pass a zero-valued `SchemaSettings` in the `TopicConfigToUpdate` struct.
2) Ensure we pass the `schema_settings` update mask path when we pass a non-`nil` `SchemaSettings` in the `TopicConfigToUpdate` struct.

hongalex:
Fix a bug that prevents clearing of topic schema because a zero valued SchemaSetting was being passed rather than a `nil` value one.

In addition, fix a bug with the fake that makes setting the `schema_settings` field mask mean changing the entire `SchemaSetting`, rather than just accounting for clearing. This makes it consistent with https://google.aip.dev/161#specifying-specific-fields

Fixes #7979


acocuzzo (approved):

saurav1991:
Can this please be merged and released as its impacting our production workflows?

hongalex:
@saurav1991 Thanks for your patience, released in pubsub 1.31.0

PR #7985, opened 2023-05-24
feat(logging): accept values of type error in Entry.Payload

ammario:
It's natural to pass in values of type error when logging at the error level. We were accidentally doing this and were left with a confusing, empty payload in the logging UI.

You can see [here](https://go.dev/play/p/xTNZmTvBfwC) how an `error` JSON marshals into `{}`.


codyoss:
Closing this for now. If you would still like to see this can you please start by opening a feature request issue. Thanks.

PR #7988, opened 2023-05-25
chore(discoveryengine): add config to generate apiv1

quartzmo:


internal/postprocessor/config.yaml
```diff
@@ -44,6 +44,7 @@ modules:
   - datastream
   - deploy
   - dialogflow
+  - discoveryengine
```
quartzmo, line 47: @codyoss Was the omission of the module configuration intentional for `discoveryengine/v1beta`? Assuming not, adding it now.

noahdietz (commented):
No need to change the release-please config like in https://github.com/googleapis/google-cloud-go/pull/7987/files ?

internal/postprocessor/config.yaml
codyoss: I don't think so.

quartzmo:
> No need to change the release-please config like in https://github.com/googleapis/google-cloud-go/pull/7987/files ? Or was that done already?

Correct, already added to `release-please-manifest-submodules.json` and `release-please-config-yoshi-submodules.json`. But it was never added to `modules` in `config.yaml`. This PR fixes that, in addition to adding the `v1` client.

noahdietz (approved):

PR #7996, opened 2023-05-26
feat(bigtable): allow custom timeout for admin operations and extended default timeout

ron-gal:
* extend admin operations default timeout to 180 seconds.
* allow change to default timeout.
* add async functions that return an operation, to be awaited separately by the client.

### DIFF
--- a/bigtable/admin.go
+++ b/bigtable/admin.go
@@ -328,7 +328,6 @@ func (ac *AdminClient) UpdateTableWithDeletionProtection(ctx context.Context, ta
 // only deletion protection can be updated at this period.
 // table ID is required.
 func (ac *AdminClient) updateTableWithConf(ctx context.Context, conf *UpdateTableConf) error {
-	op, err := ac.updateTableWithConfAsync(ctx, conf)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


bigtable/admin.go
```diff
@@ -502,6 +519,24 @@ func (ac *AdminClient) DropAllRows(ctx context.Context, table string) error {
...
+
+// CreateTableFromSnapshotAsync creates a table from snapshot.
+// The table will be created in the same cluster as the snapshot.
+// It returns an awaitable operation.
+//
+// This is a private alpha release of Cloud Bigtable snapshots. This feature
+// is not currently available to most Cloud Bigtable customers. This feature
+// might be changed in backward-incompatible ways and is not recommended for
+// production use. It is not subject to any SLA or deprecation policy.
+func (ac *AdminClient) CreateTableFromSnapshotAsync(ctx context.Context, table, cluster, snapshot string) (*longrunning.Operation, error) {
```
codyoss, line 539: PR does not mention adding these extra methods. Also seeing the work Async in a method name does not seem idiomatic. 

There have also been talks of generating a GAPIC admin surface for this API. I wonder if that is a better route forward for exposing these operations.

ron-gal: You are right, added to PR comment. These were discussed after initial version of the PR was created.
I think Async is a good name. It means that you can call this function and it will be done on it's own time, and you don't have to wait for it.

bhshkh (approved):

bigtable/admin.go
```diff
@@ -2068,3 +2223,12 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
 	_, err = ac.tClient.UpdateBackup(ctx, req)
 	return err
 }
+
+// AwaitOperation waits for an operation to be completed for interval duration.
+func AwaitOperation(ctx context.Context, op *longrunning.Operation, resp protoiface.MessageV1, interval time.Duration) error {
```
codyoss, line 2228: I don't think adding this function to the public api provides value. Using LRO is standard across all libraries and is well documented already.

igorbernstein2:
Chatted offline. There doesnt seem to be an immediate need for this and the doubling the LRO methods is not worth it.

PR #8000, opened 2023-05-29
chore(storage): add workload 6 (directories) [benchmarks]

BrennaEpp:
Open questions remaining:
- Should directories have a mix of file sizes, or is a single file size for every non-directory object okay for what we want to measure?
-   Should each subdirectory contain at least 1 file and/or subdirectory, or is an empty subdirectory acceptable? 
- What is the distribution of files to directories? Should it be a ratio of 1:1 (on average an equal amount of directories and files are created) or 2:1 (on average a file is likely to be created 2/3s of the time), or something more disparate like 10:1? Currently ratio is 1:1.
- What will the output look like? Should `object_size` == total directory size? Right now we output a new key `directory_size` and keep `object_size` to the size of objects in the directory, since it's uniform.
- The name for the input parameter to the script - I chose `-directory` but there might be something better?
- Is ignoring directories okay when uploading the directory to GCS? Or do we want to try uploading a 0 byte object or something like that?


storage/internal/benchmarks/directory_benchmark.go
```diff
@@ -0,0 +1,294 @@
...
+
+// directory benchmark is a benchmark that uploads a randomly generated
+// directory once and then downloads it once. The objects in the directory may
+// be uploaded/downloaded using parallelization.
+//
+// SET UP
+//   - Initialize structs to populate with the benchmark results.
+//   - Select a random API to use to upload/download the object, unless it is
+//     set in the command-line,
+//   - Select a random size for the objects that will be uploaded; this size will
```
tritone, line 40: IMO a mix of file sizes is good (for each one pick a random value between the lower and upper limit).

storage/internal/benchmarks/directory_benchmark.go
```diff
@@ -0,0 +1,294 @@
...
+//     be between two values configured in the command-line.
+//   - Create a directory with 1 to 1000 files of that size and subdirectories.
```
tritone, line 42: This seems like an arbitrary range, and will make some runs too short to give meaningful data. Can we add a command line flag for number of files? Or a flag for total size of the directory?

storage/internal/benchmarks/directory_benchmark.go
```diff
@@ -0,0 +1,294 @@
...
+//   - Select, for the upload and the download separately, the following parameters:
+//   - the application buffer size set in the command-line
+//   - the chunksize (only for uploads) set in the command-line,
+//   - if the client library will perform CRC32C and/or MD5 hashes on the data.
+//
+// BENCHMARK
+//   - Grab a storage client from the pool.
+//   - Take a snapshot of the current memory stats.
+//   - Upload the entire directory, capturing the time taken.
+//   - Take another snapshot of memory stats.
```
tritone, line 52: It would be more useful to be able to capture memory usage as the operation runs. Can this perhaps run in a separate Go routine on a timer and capture stats at fixed intervals?

storage/internal/benchmarks/directory_benchmark.go
```diff
@@ -0,0 +1,294 @@
...
+//   - Select a random size for the objects that will be uploaded; this size will
+//     be between two values configured in the command-line.
+//   - Create a directory with 1 to 1000 files of that size and subdirectories.
+//   - Select, for the upload and the download separately, the following parameters:
+//   - the application buffer size set in the command-line
+//   - the chunksize (only for uploads) set in the command-line,
+//   - if the client library will perform CRC32C and/or MD5 hashes on the data.
+//
+// BENCHMARK
+//   - Grab a storage client from the pool.
```
tritone, line 49: This should probably assign a separate client to each file if we are running with multiple files and threads.

storage/internal/benchmarks/utils.go
```diff
@@ -108,6 +109,44 @@ func generateRandomFile(path string, size int64) (string, error) {
...
+// subdirectories. Files may be located in any of the subdirectories.
+// The number of bytes in the created files is returned.
+func fillDirectoryRandomly(dirPath string, size int64) (int64, error) {
+	currDirectoryLevel := 0
+	currPath := dirPath
+	currNumBytes := int64(0)
+
+	// directory contains a random number of objects =< 1000
+	for i := randomInt(1, 1000); i > 0; i-- {
+		// choose at random whether to create a directory or a file
```
tritone, line 122: In response to your top-level question, I think a flatter structure (10:1 or more files to directories) seems potentially more realistic, and means that we won't spend as much time during the download phase creating directories).

storage/internal/benchmarks/utils.go
```diff
@@ -138,10 +177,13 @@ func errorIsDeadLineExceeded(err error) bool {
...
 	}
 
-	for err = errors.Unwrap(err); err != nil; {
+	err = errors.Unwrap(err)
+	for err != nil {
 		if status.Code(err) == codes.DeadlineExceeded {
 			return true
 		}
+
+		err = errors.Unwrap(err)
```
tritone, line 186: Why is this needed?

storage/internal/benchmarks/directory_benchmark.go
```diff
@@ -0,0 +1,294 @@
...
+func (r *directoryBenchmark) downloadDirectory(ctx context.Context, client *storage.Client, numWorkers int) (elapsedTime time.Duration, err error) {
+	benchGroup, ctx := errgroup.WithContext(ctx)
+	benchGroup.SetLimit(numWorkers)
+
+	// Set timer
+	start := time.Now()
+	defer func() { elapsedTime = time.Since(start) }()
+
+	// Get an iterator to list all objects under the directory
+	it := client.Bucket(r.bucketName).Objects(context.Background(), &storage.Query{
```
tritone, line 217: I think using multiple calls to list with a prefix and delimiter is potentially a more realistic way to walk the "directory" structure and limits how much you need to check whether your directories exist locally. However I'm not sure it makes any difference perf wise so possibly not worth it.

Also, you can use Query.SetAttrSelection to only get object names (since the other metadata is not relevant).

tritone (commented):
Nice start on this! A few comments.

storage/internal/benchmarks/directory_benchmark.go
BrennaEpp: Yes, I plan on adding a flag for size or file number so it can be more deterministic as well. Waiting for feedback on if a mix of file sizes is appropriate/deterministic enough.

storage/internal/benchmarks/utils.go
BrennaEpp: hm, this comment made me reconsider adding this here. Will revert this change and open a separate PR for this, as this change is due to a bug I found while writing this PR, but is not necessarily relevant here.

storage/internal/benchmarks/utils.go
BrennaEpp: I agree (thanks for the input!), I think 10:1 would be around the minimum ratio that would look more realistic as a directory structure - but I don't have data to back it up so let's see what Frank says

storage/internal/benchmarks/directory_benchmark.go
BrennaEpp: Good point, will change.

storage/internal/benchmarks/directory_benchmark.go
BrennaEpp: I agree, really I haven't put much focus on the capturing or analysis of memory stats. Do you have a proposal for how memory stats should be output? Maybe we could add more metrics (would need a bit of investigation to make sure it can be done and captured by the prober framework; although for current workloads running on the framework I don't know how useful the data would actually be)?

storage/internal/benchmarks/directory_benchmark.go
BrennaEpp: Leaving as-is since the directory will be a flat structure anyway, as discussed.

Set the query attr selection as suggested.

storage/internal/benchmarks/utils.go
BrennaEpp: As discussed offline, the directory will not contain any subdirectories.

storage/internal/benchmarks/directory_benchmark.go
BrennaEpp: Done

storage/internal/benchmarks/directory_benchmark.go
BrennaEpp: Done as discussed

storage/internal/benchmarks/directory_benchmark.go
tritone: I don't have a fully fleshed out proposal for a metric, but I think what would be useful to capture is average & max memory usage while the workload is running, based on snapshots taken at (say) a 60s interval. But, I think it's beyond the scope of this PR -- could you file a separate issue to add this?

storage/internal/benchmarks/README.md
```diff
@@ -31,6 +31,7 @@ This runs 1000 iterations on 512kib to 2Gib files in the background, sending out
 | -connection_pool_size | GRPC connection pool size | any positive integer | 4 |
 | -force_garbage_collection | whether to force garbage collection <br> before every write or read benchmark |  `true` or `false` (present/not present) | `false` |
 | -timeout | timeout (maximum time running benchmarks) <br> the program may run for longer while it finishes running processes | any [time.Duration](https://pkg.go.dev/time#Duration) | `1h` |
-| -directory | upload and download an entire directory | `true`/`false` | `false` |
+| -workload | `1` will run a w1r3 (write 1 read 3) benchmark <br> `6` will run a benchmark uploading and downloading a directory (once each) | `1` or `6` | `1` |
+| -objects_per_directory | total number of objects in a directory; <br> only applies to workload 6 | any positive integer | `1000` |
```
tritone, line 35: Maybe just `-num_objects`? It should be clear that there is one directory and this controls the total number of objects uploaded/downloads.

storage/internal/benchmarks/utils.go
```diff
@@ -109,39 +108,23 @@ func generateRandomFile(path string, size int64) (string, error) {
 	return f.Name(), err
 }
 
-// fillDirectoryRandomly fills the directory with up to 1000 different files and
-// subdirectories. Files may be located in any of the subdirectories.
-// The number of bytes in the created files is returned.
-func fillDirectoryRandomly(dirPath string, size int64) (int64, error) {
-	currDirectoryLevel := 0
-	currPath := dirPath
+// fillDirectoryRandomly fills the directory with the number of different files
```
tritone, line 111: Maybe make it explicit that there is only one directory created (no substructure), and change the name to something like `createTempDirectory` (or something else that doesn't include "random".

tritone (approved):
Couple minor comments, overall looks good! Feel free to resolve and merge when ready.

storage/internal/benchmarks/README.md
BrennaEpp: I wanted to make it clear that this parameter only applies to directories. `-num_objects` might be confused with `-samples`. If you think that isn't an issue, I'm fine with changing it to `num_objects`

PR #8017, opened 2023-05-31
fix(storage): fix reads with "./" in object name [XML]

### LINKED ISSUE #7236: storage: ErrObjectNotExist in NewRangeReader for "./" prefixed objects
**Client**

Storage

**Environment**

MacOS

**Go Environment**

$ go version

```
go version go1.18.7 darwin/arm64
```

**Code**

`NOTE` in the following examples, both `gs://bucket/object1.mcap` and `gs://bucket/./object1.mcap` objects exist (and are copies of one another).

```go
handle := gcs.Bucket(bucketName).Object("object1.mcap")
fmt.Println(handle.Attrs(ctx))
// Prints the attributes for `object1.mcap`
r, err := handle.NewRangeReader(ctx, 0, -1)
// Works as expected, no error.
```

Compare to:

```go
handle := gcs.Bucket(bucketName).Object("./object1.mcap")
fmt.Println(handle.Attrs(ctx))
// Prints the attributes for `./object1.mcap` as above
r, err := handle.NewRangeReader(ctx, 0, -1)
// Error: `storage: object doesn't exist`
```

**Expected behavior**

Running the same script for both existing objects, they should behave the same. The "./" object prefix shouldn't cause unexpected issues.

**Actual behavior**

* There's no difference in `handle.Attrs(ctx)` — this prints the same for both `./object1.mcap` and `object1.mcap`
* But `handle.NewRangeReader(ctx, 0, -1)` works for `object1.mcap` only, and displays an error for `./object1.mcap`:

```
storage: object doesn't exist
```

**Additional context**

`NOTE` that `x/object1.mcap` or other paths will work as well, the only exception so far being the `./` prefix. Objects with "./" prefix should be legitimate.

BrennaEpp:
Previously, XML reads failed when the object had "./" in the name. This fixes the encoding for such a case.

Fixes #7236


tritone (approved):
nice fix!

PR #8020, opened 2023-05-31
test: retry create remote udf function

alvarowolfx:
Fix flaky issue #7994


bigquery/routine_integration_test.go
```diff
@@ -121,7 +123,14 @@ func TestIntegration_RoutineRemoteUDF(t *testing.T) {
 			TypeKind: "STRING",
 		},
 	}
-	if err := routine.Create(ctx, meta); err != nil {
+
+	err = internal.Retry(ctx, gax.Backoff{}, func() (stop bool, err error) {
+		if err := routine.Create(ctx, meta); err != nil {
+			return false, err
```
shollyman, line 129: Do you want to evaluate this via `retryableError()` or similar?

shollyman (approved):

bigquery/routine_integration_test.go
alvarowolfx: do you feel like is needed as this is just for integration test ?

shollyman: We can leave as is and swap if it causes issues down the line.

PR #8055, opened 2023-06-05
feat(bigquery): support for tables primary and foreign keys

alvarowolfx:
Add support for creating/updating tables with primary/foreign keys on BigQuery. 

Rollout not fully complete, so is not gonna pass integration tests.

Towards internal b/285349974


bigquery/table.go
```diff
@@ -147,6 +147,99 @@ type TableMetadata struct {
...
+
+	// ForeignKeys represent a list of foreign keys constraints.
+	// The foreign key is not enforced.
+	ForeignKeys []*ForeignKey
+}
+
+// PrimaryKey represents the primary key constraint on a table's columns.
+type PrimaryKey struct {
+	// Columns that compose the primary key constraint.
+	Columns []string
```
shollyman, line 164: Question:  are there ordering constraints here (and in FK) we need to document?  Or does the order of columns not matter?

bigquery/table.go
```diff
@@ -1024,6 +1146,15 @@ type TableMetadataToUpdate struct {
 	// in the table.
 	DefaultCollation optional.String
 
+	// Optionally specifies a PrimaryKey constraint on a table's columns.
+	// Updated only if present.
+	// The primary key is not enforced.
+	PrimaryKey *PrimaryKey
+
+	// ForeignKeys represent a list of foreign keys constraints.
+	// The foreign key is not enforced.
```
shollyman, line 1155: pluralize the comment?  foreign keys not enforced?

bigquery/table.go
```diff
@@ -947,6 +1057,18 @@ func (tm *TableMetadataToUpdate) toBQ() (*bq.Table, error) {
 		t.DefaultCollation = optional.ToString(tm.DefaultCollation)
 		forceSend("DefaultCollation")
 	}
+	if tm.PrimaryKey != nil || len(tm.ForeignKeys) > 0 {
```
shollyman, line 1060: Is special logic needed for clearing the TableConstraints message?  ex: if both PK and FK are empty?  Otherwise, how does a user remove TableConstraints?

bigquery/table.go
alvarowolfx: AFAIK the ordering doesn't matter, but I'll double check internally

bigquery/table.go
alvarowolfx: ah good point, I forgot about allowing users to remove constraints. I just pushed an update to allow that.

bigquery/table_integration_test.go
```diff
@@ -661,3 +661,175 @@ func TestIntegration_TableDefaultCollation(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if md.PrimaryKey.Columns[0] != "name" {
+		t.Fatalf("expected table primary key to contain column `name`, but found %q", md.PrimaryKey.Columns)
+	}
+
+	md, err = table.Update(ctx, TableMetadataToUpdate{
+		PrimaryKey: &PrimaryKey{
+			Columns: []string{}, // clean primary keys
```
shollyman, line 693: is Columns even necessary, or is a default value &PrimaryKey{} sufficient?

shollyman (approved):

PR #8068, opened 2023-06-08
feat(firestore): add support for FirebaseEvent

yurifedoseev:
Reopen https://github.com/googleapis/google-cloud-go/pull/4335
Discussion https://github.com/googleapis/google-cloud-go/issues/1438

Cloud Functions supports Firestore triggers and when a document is updated an event is send to the function with this format. There is currently no way to convert pb.Document to a normal Go struct. [setFromProtoValue](https://github.com/googleapis/google-cloud-go/blob/61166ef305533a6c5b0b3275df2cb544d678339a/firestore/from_value.go#L27) could be exported to make it possible or a new helper function based on this [DataTo](https://github.com/googleapis/google-cloud-go/blob/27b5416a80e6a65ee9eebda55f06def51b9b7b88/firestore/document.go#L115) line.

We're using this patch more then year in production.  It works fine. We just want to move from fork to upstream to simplify google cloud deploy. 

For example, using fork forces us to upload all go dependencies on gcloud functions instead of just `go.mod` file


yurifedoseev:
Recreated PR with the only commit https://github.com/googleapis/google-cloud-go/pull/8069

PR #8069, opened 2023-06-08
feat(firestore): add support for parsing proto from FirebaseEvent

yurifedoseev:
Reopen https://github.com/googleapis/google-cloud-go/pull/4335
Discussion https://github.com/googleapis/google-cloud-go/issues/1438

Cloud Functions supports Firestore triggers and when a document is updated an event is send to the function with this format. There is currently no way to convert pb.Document to a normal Go struct. [setFromProtoValue](https://github.com/googleapis/google-cloud-go/blob/61166ef305533a6c5b0b3275df2cb544d678339a/firestore/from_value.go#L27) could be exported to make it possible or a new helper function based on this [DataTo](https://github.com/googleapis/google-cloud-go/blob/27b5416a80e6a65ee9eebda55f06def51b9b7b88/firestore/document.go#L115) line.

We're using this patch more then year in production.  It works fine. We just want to move from fork to upstream to simplify google cloud deploy. 

For example, using fork forces us to upload all go dependencies on gcloud functions instead of just `go.mod` file

### DIFF
--- a/firestore/event.go
+++ b/firestore/event.go
@@ -1,0 +1,108 @@
+// Copyright 2017 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package firestore
+
+import (
+	"context"
+	"encoding/json"
+	"time"
+
+	"cloud.google.com/go/functions/metadata"
+	pb "google.golang.org/genproto/googleapis/firestore/v1"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+// FirestoreEvent is the payload of a Firestore event.
+type FirestoreEvent struct {
+	OldValue   FirestoreValue `json:"oldValue"`
+	Value      FirestoreValue `json:"value"`
+	UpdateMask struct {
+		FieldPaths []string `json:"fieldPaths"`
+	} `json:"updateMask"`
+}
+
+// FirestoreValue holds Firestore fields.
+type FirestoreValue struct {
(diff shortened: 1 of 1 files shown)


codyoss:
I don't think we should be adding event types to this repo as they already exist elsewhere and are generated: https://github.com/googleapis/google-cloudevents-go/tree/main/cloud/firestoredata. I could see value in the helper function though. Let me do some asking around about if we have any plans to integrate with cloud events more.

yurifedoseev:
> I don't think we should be adding event types to this repo as they already exist elsewhere and are generated: https://github.com/googleapis/google-cloudevents-go/tree/main/cloud/firestoredata. I could see value in the helper function though. Let me do some asking around about if we have any plans to integrate with cloud events more.

Let me add an extra info about the problem. 
We use Firestore with a bunch of GCP function triggers. For example, we stream the data to analytics storages, aggregate some models or complete heavy logic asynchronously. 
We map firestore snapshots to objects on backend using our proto models. Something like that
```go
// load model
snap, err := docRef.Get(ctx)
model := &api.ProtoModel{}
err = snap.DataTo(model)
return model
```
All other BE codes operators with this proto models, all good.

We receive `firestore.FirestoreEvent` model in GCP functions, for example, in triggers for updating documents by path.
The struct is defined in google sdk as:
```go
// code in cloud.google.com/go/firestore
// FirestoreEvent is the payload of a Firestore event.
type FirestoreEvent struct {
	OldValue   FirestoreValue `json:"oldValue"`
	Value      FirestoreValue `json:"value"`
	UpdateMask struct {
		FieldPaths []string `json:"fieldPaths"`
	} `json:"updateMask"`
}

// FirestoreValue holds Firestore fields.
type FirestoreValue struct {
	CreateTime time.Time   `json:"createTime"`
	Fields     interface{} `json:"fields"`
	Name       string      `json:"name"`
	UpdateTime time.Time   `json:"updateTime"`
}
``` 

The problem that there is no convenient way to convert `FirestoreValue` to our `ProtoModel` as we do in backend. Which means that we can't reuse our code which works with `ProtoModels`.
That's why our fork was created which allows us to convert `FirestoreValue` to our model in the same way as we do in backend.
It could be just our add-on (not fork) but it requires access to non-exported method[ `pathToDoc` ](https://github.com/googleapis/google-cloud-go/blob/main/firestore/from_value.go#L399) from sdk which handles `DocumentRefs`.
If there is any way to handle this problem without a fork I'll be glad to try it.

PR #8104, opened 2023-06-15
feat(profiler): support configurable debug logging destination

### LINKED ISSUE #8089: profiler: Debug log output stream is not configurable
Currently when `DebugLogging` is set to `true` the profiler package always logs to stderr. 

```go
		err = profiler.Start(profiler.Config{
			ProjectID:      cfg.GoogleProjectID,
			Service:        "my-service",
			ServiceVersion: buildInfo.Main.Version,
			DebugLogging:   true,
		})
```

Produced error messages:

Screenshot 2023-06-07 at 13 04 53

Do you think it makes sense to expose the internally used logger by any means? This way one will able to configure a different output stream for the debug log messages.

noahdietz:
Adds the ability to configure the destination of debug logs coming from a profiler. The default is `os.Stderr`.

Incorporates it into an existing test to make sure the routing works.

Fixes #8089


profiler/profiler.go
```diff
@@ -130,6 +131,10 @@ type Config struct {
 	// defaults to false.
 	DebugLogging bool
 
+	// DebugLoggingOut is where the logger will write debug logs to, if enabled.
+	// It defaults to os.Stderr.
+	DebugLoggingOut io.Writer
```
codyoss, line 136: nit: Out seem short for Output which would mirror what is in stdlib: https://pkg.go.dev/log#SetOutput
Maybe be more explicit and use the full word

codyoss (approved):

profiler/profiler.go
noahdietz: Cool, sounds good to me! Done.

PR #8113, opened 2023-06-15
feat(storage): respect WithEndpoint for SignedURLs and PostPolicy

BrennaEpp:
Clients initiated with a non default endpoint will use that endpoint for SignedURLs and Post Policies.

Uses the endpoint set for `client.readHost` as we set that anyway to the correct endpoint when initializing a client. It could also be extracted from `client.raw.BasePath` but `client.readHost` is already in the form we need.


storage/storage.go
```diff
@@ -442,6 +456,8 @@ type SignedURLOptions struct {
 	// Scheme determines the version of URL signing to use. Default is
 	// SigningSchemeV2.
 	Scheme SigningScheme
+
+	endpoint string
```
tritone, line 460: Now I'm wondering if we should just make this exported to support users who are using the standalone func as well. I don't really see a big downside to doing so. WDYT?

storage/storage.go
```diff
@@ -284,23 +284,37 @@ type bucketBoundHostname struct {
 	hostname string
 }
 
-func (s pathStyle) host(bucket string) string {
+func (s pathStyle) host(endpoint, bucket string) string {
+	if endpoint != "" {
+		return stripScheme(endpoint)
+	}
+
+	// This check is needed for clientless calls to SignedURL
```
tritone, line 292: This comment is a bit confusing, I guess you are saying that endpoint will be set to STORAGE_EMULATOR_HOST already if called via the bucketHandle method? If so probably not necessary to mention.

storage/bucket.go
```diff
@@ -173,6 +173,9 @@ func (b *BucketHandle) Update(ctx context.Context, uattrs BucketAttrsToUpdate) (
 // [Overview of access control]: https://cloud.google.com/storage/docs/accesscontrol#signed_urls_query_string_authentication
 // [automatic detection of credentials]: https://pkg.go.dev/cloud.google.com/go/storage#hdr-Credential_requirements_for_signing
 func (b *BucketHandle) SignedURL(object string, opts *SignedURLOptions) (string, error) {
+	// Extract the correct endpoint from the readhost set on the client
+	opts.endpoint = b.c.readHost
```
tritone, line 177: I think calling this `endpoint` is a bit misleading. Usually I think of endpoint as being the full path for the JSON API, i.e. `https://storage.googleapis.com/storage/v1`. Just `storage.googleapis.com` would be a hostname.

Sidenote, `readHost` is also a bit misleading because now using XML reads is optional; maybe we should rename to `xmlHost`. (not required for this PR though).

storage/bucket_test.go
```diff
@@ -1210,3 +1212,301 @@ func TestDetectDefaultGoogleAccessID(t *testing.T) {
...
+		opts         *SignedURLOptions
+		// Note for future implementors: X-Goog-Signature generated by having
+		// the client run through its algorithm with pre-defined input and copy
+		// pasting the output. These tests are not great for testing whether
+		// the right signature is calculated - instead we rely on the backend
+		// and integration tests for that.
+		want string
+	}{
+		{
+			desc:         "SignURLV4 creates link to resources in emulator",
```
tritone, line 1240: These cases really seem like they belong in the conformance tests for signed URLs once the project is complete. Could you add a TODO to move them? (and I guess this would also require exporting the hostname option rather than having it be internal only)

storage/storage.go
```diff
@@ -268,7 +268,7 @@ const (
 type URLStyle interface {
```
tritone, line 268: Can you also update the docs for PathStyle to indicate that the provided hostname is taken into account?

storage/storage.go
```diff
@@ -284,23 +284,37 @@ type bucketBoundHostname struct {
...
 
+	// Fallback to default endpoint - clientless calls to SignURL/PostPolicy
+	// will not have an endpoint attached
 	return "storage.googleapis.com"
 }
 
-func (s virtualHostedStyle) host(bucket string) string {
+func (s virtualHostedStyle) host(endpoint, bucket string) string {
+	if endpoint != "" {
+		return bucket + "." + stripScheme(endpoint)
```
tritone, line 304: I'm not totally sure about this one and whether it is expected to be supported. Let's discuss offline next week.

tritone (commented):
Nice start on this, a  few comments.

storage/storage.go
BrennaEpp: Yes, that is what the comment implies - this check is only needed for using the standalone func, because the endpoint would be set already for the bucket method. But you're right - we have tests that will fail if the check is removed; will remove comment.

storage/storage.go
BrennaEpp: I think the main concern I'd have is that this field gets overwritten in the bucketHandle methods (so it can get set to option.withEndpoint without users needing to add that to the opts). Although, I guess we could check and only override this field if not provided in opts - how does that sound? That way this field is exported and takes precedence over everything else (emulator env var, WithEndpoint option...).

storage/storage.go
BrennaEpp: Done. To-do on virtual hosted style, if we decide to support it.

storage/bucket.go
BrennaEpp: You are correct. 

Here is a PR for renaming `readHost`: https://github.com/googleapis/google-cloud-go/pull/8160

storage/post_policy_v4.go
```diff
@@ -113,8 +113,13 @@ type PostPolicyV4Options struct {
 	// Optional.
 	Conditions []PostPolicyV4Condition
 
+	// Hostname sets the host of the signed post policy. This field overrides
+	// any endpoint set on a storage Client or through STORAGE_EMULATOR_HOST.
+	// Not compatible with BucketBoundHostname URLStyle.
```
tritone, line 118: I would say "only compatible with PathStyle" because we omitted VirtualHosted also. Same in the docs for SignedURLOptions below.

tritone (approved):
One more minor comment, otherwise LGTM

Issue #8127, opened 2023-06-16
internal/retry: wrapped errors have grpc status code OK when the wrapped error is not a gRPC error


atollena:
**Client**

[All clients that use retries](https://github.com/googleapis/google-cloud-go/blob/4eb8a58b1ca80239d046e0b759a292e24e85a772/internal/retry.go#L80-L85).

**Environment**

Any.

**Go Environment**

Any.

**Code**

Any code using internal.Retry().

**Expected behavior**

An error's `GRPCStatus()` method should never returns `nil`, which maps to a status with `Code.OK` (https://github.com/grpc/grpc-go/blob/642dd63a85275a96d172f446911fd04111e2c74c/internal/status/status.go#L71-L76). Instead it should return an actual error code, such as `Status.Unknown`.

**Actual behavior**

If `Status.FromError` returns !ok, which it does if the wrapped error is not a gRPC error, then the wrapped error has `GRPCStatus()` returning nil (Status.OK).

**Additional context**

We discovered this problem because it can cause panics in recent versions of grpc-go. See this issue: https://github.com/grpc/grpc-go/issues/6373.

noahdietz:
Thanks for reporting this. Let's take a look at your fix!

atollena:
Fixed by https://github.com/googleapis/google-cloud-go/commit/005d2dfb6b68bf5a35bfb8db449d3f0084b34d6e.

PR #8128, opened 2023-06-16
fix(internal/retry): never return nil from GRPCStatus()

atollena:
An error's GRPCStatus() method should never returns nil, which maps to a status with Code.OK (https://github.com/grpc/grpc-go/blob/642dd63a85275a96d172f446911fd04111e2c74c/internal/status/status.go#L71-L76). Instead it should return an actual error code, such as Status.Unknown.

This addresses https://github.com/googleapis/google-cloud-go/issues/8127.


internal/retry.go
```diff
@@ -17,6 +17,7 @@ package internal
 import (
 	"context"
 	"fmt"
+	"google.golang.org/grpc/codes"
```
noahdietz, line 20: This should be down with the other imports. Please run `gofmt -w internal/retry.go` that should fix it.

noahdietz (commented):
This is basically the same change we made to gax-go for a similar reason. Thanks for doing this. One nit.

noahdietz:
@tritone @codyoss I'm not as familiar with where this package is used. Just want to double check - any issues with this change?

noahdietz:
@atollena sorry out of habit I caught up your branch 😬 might want to pull before pushing a new commit

codyoss:
I believe this should be fine. This is used by a handful of our manual clients.

internal/retry.go
noahdietz: Fixed it myself, needed to use `goimports` instead, my bad.

noahdietz (approved):

noahdietz:
Thanks @codyoss ! I've updated the PR in place. Setting `automerge`.

tritone (approved):
I do feel like the contract is slightly ambiguous here, and it is a bit odd to have status.FromError return something for a non-gRPC error (even say if the call being wrapped for retry is not a gRPC call). However, I don't think this will cause any problems.

PR #8150, opened 2023-06-19
fix(storage): Same method name and Trace Span name

sinmetal:
I think the method name and the Trace Span Name are expected to be the same, so I fixed that.


tritone (approved):
Thanks for the fix and sorry to miss this one earlier.

PR #8152, opened 2023-06-19
feat(spanner): add databoost property for batch transactions

harshachinta:
Adding support for databoost enabled in PartitionedRead and PartitionedQuery

### DIFF
--- a/spanner/batch.go
+++ b/spanner/batch.go
@@ -150,13 +150,14 @@ func (t *BatchReadOnlyTransaction) PartitionReadUsingIndexWithOptions(ctx contex
 	}
 	// Prepare ReadRequest.
 	req := &sppb.ReadRequest{
-		Session:        sid,
-		Transaction:    ts,
-		Table:          table,
-		Index:          index,
-		Columns:        columns,
-		KeySet:         kset,
-		RequestOptions: createRequestOptions(readOptions.Priority, readOptions.RequestTag, ""),
+		Session:          sid,
+		Transaction:      ts,
+		Table:            table,
+		Index:            index,
+		Columns:          columns,
+		KeySet:           kset,
+		RequestOptions:   createRequestOptions(readOptions.Priority, readOptions.RequestTag, ""),
+		DataBoostEnabled: readOptions.DataBoostEnabled,
 	}
 	// Generate partitions.
 	for _, p := range resp.GetPartitions() {
@@ -212,13 +213,14 @@ func (t *BatchReadOnlyTransaction) partitionQuery(ctx context.Context, statement
 
 	// prepare ExecuteSqlRequest
 	r := &sppb.ExecuteSqlRequest{
-		Session:        sid,
-		Transaction:    ts,
-		Sql:            statement.SQL,
-		Params:         params,
-		ParamTypes:     paramTypes,
-		QueryOptions:   qOpts.Options,
-		RequestOptions: createRequestOptions(qOpts.Priority, qOpts.RequestTag, ""),
+		Session:          sid,
+		Transaction:      ts,
+		Sql:              statement.SQL,
+		Params:           params,
+		ParamTypes:       paramTypes,
+		QueryOptions:     qOpts.Options,
+		RequestOptions:   createRequestOptions(qOpts.Priority, qOpts.RequestTag, ""),
+		DataBoostEnabled: qOpts.DataBoostEnabled,
 	}
 
 	// generate Partitions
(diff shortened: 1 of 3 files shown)


gauravpurohit06 (approved):
LGTM, Can we just re-check why the integration tests are passing ?

spanner/batch.go
```diff
@@ -150,13 +150,14 @@ func (t *BatchReadOnlyTransaction) PartitionReadUsingIndexWithOptions(ctx contex
...
-		KeySet:         kset,
-		RequestOptions: createRequestOptions(readOptions.Priority, readOptions.RequestTag, ""),
+		Session:          sid,
+		Transaction:      ts,
+		Table:            table,
+		Index:            index,
+		Columns:          columns,
+		KeySet:           kset,
+		RequestOptions:   createRequestOptions(readOptions.Priority, readOptions.RequestTag, ""),
+		DataBoostEnabled: readOptions.DataBoostEnabled,
```
gauravpurohit06, line 160: Add a one line comment: 
`If this is for a partitioned read and this field is set to true, the request will be
executed via Spanner independent compute resources. `

spanner/batch.go
```diff
@@ -212,13 +213,14 @@ func (t *BatchReadOnlyTransaction) partitionQuery(ctx context.Context, statement
...
-		QueryOptions:   qOpts.Options,
-		RequestOptions: createRequestOptions(qOpts.Priority, qOpts.RequestTag, ""),
+		Session:          sid,
+		Transaction:      ts,
+		Sql:              statement.SQL,
+		Params:           params,
+		ParamTypes:       paramTypes,
+		QueryOptions:     qOpts.Options,
+		RequestOptions:   createRequestOptions(qOpts.Priority, qOpts.RequestTag, ""),
+		DataBoostEnabled: qOpts.DataBoostEnabled,
```
gauravpurohit06, line 223: Add a one line comment:
`If this is for a partitioned query and this field is set to true, the request will be executed via Spanner independent compute resources. `

gauravpurohit06 (changes requested):

spanner/batch.go
harshachinta: I have added it under ReadOptions
https://github.com/googleapis/google-cloud-go/pull/8152/files#diff-98c680ad0e9a6842d77932c6b37ee80a322d5b5e462a26a4f0154844ea4e8513R169

spanner/batch.go
harshachinta: i have added it under QueryOptions
https://github.com/googleapis/google-cloud-go/pull/8152/files#diff-98c680ad0e9a6842d77932c6b37ee80a322d5b5e462a26a4f0154844ea4e8513R169

gauravpurohit06 (approved):

spanner/transaction.go
```diff
@@ -165,16 +165,20 @@ type ReadOptions struct {
 
 	// The request tag to use for this request.
 	RequestTag string
+
+	// If this is for a partitioned read and DataBoostEnabled field is set to true, the request will be executed via Spanner independent compute resources.
```
olavloite, line 169: nit: split this comment over two lines

Also add the following to the end of the comment: `Setting this option for regular read operations has no effect.`

spanner/transaction.go
```diff
@@ -188,6 +192,9 @@ func (ro ReadOptions) merge(opts ReadOptions) ReadOptions {
 	if opts.RequestTag != "" {
 		merged.RequestTag = opts.RequestTag
 	}
+	if opts.DataBoostEnabled {
```
olavloite, line 195: Hmmm.... This means that you cannot override a `false` value in `ro`. I guess that is OK, as we don't allow 'unsetting' any of the other values through this method either. (No action needed, just thinking out loud....)

spanner/transaction.go
```diff
@@ -357,16 +369,20 @@ type QueryOptions struct {
 
 	// The request tag to use for this request.
 	RequestTag string
+
+	// If this is for a partitioned query and DataBoostEnabled field is set to true, the request will be executed via Spanner independent compute resources.
```
olavloite, line 373: Same as above: split over two lines and add an extra sentence that it has no effect on regular reads.

olavloite (approved):

spanner/transaction.go
harshachinta: Added Knut.

spanner/transaction.go
harshachinta: Added

PR #8164, opened 2023-06-21
fix(bigquery/storage/managedwriter): correct reconnection logic

shollyman:
Signalling for an AppendRows stream when schema changes is predicated on the backend's status for the connection.  For a simplex (non-multiplexed) connection, the expectation is the client closes and reconnects to signal there's a change in the schema.

For a connection in multiplex mode, no reconnection is necessary and the backend will look at the schema for changes.

In managedwriter, we allow a user to specify multiplex at the outset, but for connections that haven't actually sent writes for more than a single stream ID the backend doesn't recognize the multiplex status.  Essentially, client state is out of sync with server state until a second stream writes on the same connection.

This PR expands the interface for send optimizer to signal whether the optimizer has sent writes for multiple connections, and uses it when making the determination about schema-based reconnects.  It also augments the schema evolution test to validate using multiple combinations of writer and client options.


alvarowolfx (approved):
LGTM

PR #8166, opened 2023-06-22
test: fix brittle tests

codyoss:
These tests relied on the fact that the retry option was the first. It is likely in the future orders will change even more so these tests should eventually be written or removed. We should exercies that retries work more generally if we don't already.

Fixes: #8095
Fixes: #8146


harshachinta (approved):
Thanks for fixing this, Cody. I was about to raise a PR as continuous tests were failing because of this issue.

noahdietz (approved):

PR #8167, opened 2023-06-22
fix(datastore): Change aggregation result to return generic value

bhshkh:
Partially reverting the change made in https://github.com/googleapis/google-cloud-go/pull/7803 as it is a breaking  change. The code that uses AggregationResult expected interface datatype in the map value but now concrete data types are being returned and break the code.


codyoss (approved):

marwansalem:
hello, how can I use the result of an aggregation query? Can you point me towards the documentation?

bhshkh:
> hello, how can I use the result of an aggregation query? Can you point me towards the documentation?

https://github.com/googleapis/google-cloud-go/issues/7800#issuecomment-1622224523

PR #8168, opened 2023-06-23
feat(pubsub): enable project autodetection and detect empty project

### LINKED ISSUE #3828: pubsub: consider returning non-nil error from NewClient when projectID is invalid
This issue is motivated by my experience in https://github.com/golang/go/issues/44948#issuecomment-797227262.

We had code that was publishing to a pub/sub topic that looked roughly like this:

<details>

```Go
// publishToTopic publishes body to the given topic.
// It returns the ID of the message published.
func publishToTopic(ctx context.Context, topic string, body []byte) (string, error) {
	client, err := pubsub.NewClient(ctx, projectID)
	if err != nil {
		return "", fmt.Errorf("pubsub.NewClient: %v", err)
	}

	t := client.Topic(topic)
	resp := t.Publish(ctx, &pubsub.Message{Data: body})
	id, err := resp.Get(ctx)
	if err != nil {
		return "", fmt.Errorf("topic.Publish: %v", err)
	}
	return id, nil
}
```

</details>

When the value of `projectID` became the empty string due to an environment misconfiguration, the `publishToTopic` function failed with the following error message:

```
topic.Publish: rpc error: code = InvalidArgument desc = You have passed an invalid argument to the service (argument=).
```

The error message said that I passed an invalid argument, and that the argument I passed was the empty string. With that information, it was quite hard to understand what the problem actually was. I wasn't providing any arguments in the high-level code shown above. (I'm guessing the projectID is one of the internal arguments used when publishing.)

As far as I can tell there's no situation where `pubsub.NewClient(ctx, "")` is a non-error condition.

If that's the case, perhaps the readability of the error in this situation can be improved by having `NewClient` return something like `errors.New("projectID must be a non-empty string")` to make the error more readable:

```
topic.NewClient: projectID must be a non-empty string
```

- If the error check cannot be done in `NewClient`, then perhaps the error message returned by `Publish` can be improved to be more clear which argument was empty, like "You have passed an invalid value for the projectID argument to the service".

- If the case where a user accidentally calls `pubsub.NewClient(ctx, projectID)` with an empty string is deemed rare not to add more complexity to the client code, then another alternative solution is to close this for now and don't take any action unless this problem keeps happening.

**Additional context**
Issue #1294 may be related.

hongalex:
Allows autodetection of project ID when the sentinel value (DetectProjectID) is passed in to `pubsub.NewClient`. This uses the `detect`, similar to BigQuery in #4312. Similar to Firestore in #4299, empty project IDs will now return an error.

Also exposes a new method in pubsub client `client.Project()` which returns the project ID.

Retrieving information about the topic's (and subscription) project ID can be done by `topic.Config`, followed by calling `config.String()`.

Addresses pubsub portion in #1294

Fixes #3828


shollyman (approved):

pubsub/pubsub.go
```diff
@@ -113,13 +115,30 @@ func mergeSubscriberCallOptions(a *vkit.SubscriberCallOptions, b *vkit.Subscribe
...
+var ErrEmptyProjectID = errors.New("pubsub: projectID string is empty")
+
 // NewClient creates a new PubSub client. It uses a default configuration.
 func NewClient(ctx context.Context, projectID string, opts ...option.ClientOption) (c *Client, err error) {
 	return NewClientWithConfig(ctx, projectID, nil, opts...)
 }
 
 // NewClientWithConfig creates a new PubSub client.
 func NewClientWithConfig(ctx context.Context, projectID string, config *ClientConfig, opts ...option.ClientOption) (c *Client, err error) {
+	if projectID == "" {
```
marwanad, line 139: @hongalex doesn't having this check early on  break the `opts.WithoutAuthentication` case since it now requires you to pass a dummy projectID to instantiate the clients.

mvhatch: Isn't this a breaking change that warrants a major version bump instead of a minor bump?

hongalex: While this may be backwards incompatible, but falls under the realm of a bug fix and doesn't warrant a major version bump. Passing in an empty string to the client is unsupported behavior and should have resulted in an error, even when using unauthenticated. The project string is used to construct paths like "projects/my-project/topics/my-topic" and creating resources like "projects//topics/my-topic" is undefined behavior.

PR #8175, opened 2023-06-26
feat(bigquery): allow users to override retry settings

### LINKED ISSUE #7526: bigquery: add a way to override retry properties
In big query the retry configuration is hard coded inside big query package:
```golang
backoff := gax.Backoff{
		Initial:    1 * time.Second,
		Max:        32 * time.Second,
		Multiplier: 2,
	}
```

I would want to override it in give higher values for all fields to avoid duplications in BQ. (prevent retry if the row may be written successfully even some error retrieved )

In all other gcp services the retry config can be overriden from the client code, Can we align also BQ to be the same, this is very weird that you have hard coded values that can't be overriden.

alvarowolfx:
Allow users to set retry settings on the `bigquery.Client` with now is applied for all operations. Later it can be customized by `Table`, `Dataset` and `Job` objects and operations. 

Most of it was based on the `google-cloud-go/storage` client. 

Fixes #7526

### DIFF
--- a/bigquery/retry.go
+++ b/bigquery/retry.go
@@ -1,0 +1,208 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bigquery
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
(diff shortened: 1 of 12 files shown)


alvarowolfx (commented):
Per conversation with @shollyman 

* Check if we can use gax.Retryer or gax-go interfaces to configure retry.
* Can retry overrides be a gax.CallOption to override on a per method level. 
  *  Maybe create custom CallOption to allow retry settings override - check managedwriter for examples on how to do it (options.go)

bigquery/retry.go
```diff
@@ -0,0 +1,208 @@
...
+	"net/url"
+	"strings"
+	"time"
+
+	cloudinternal "cloud.google.com/go/internal"
+	gax "github.com/googleapis/gax-go/v2"
+	"google.golang.org/api/googleapi"
+)
+
+type retryConfig struct {
```
shollyman, line 31: should retryConfig have some kind of per-method / per-rpc configurability?  Currently gax is heavily CallOption based, so likely the config should likely fit better as per rpc.  I don't like the idea of attaching per resource retries because of the problems with tracking down references.

codyoss:
@alvarowolfx do we still want this?

codyoss:
@alvarowolfx see comment above

PR #8177, opened 2023-06-27
feat(spanner): long running transaction clean up - disabled

harshachinta:
Automatically cleaning long running transactions


spanner/session.go
```diff
@@ -1364,7 +1450,28 @@ func (hc *healthChecker) maintainer() {
...
+			hc.pool.InactiveTransactionRemovalOptions.lastExecutionTime = now
+		}
 		hc.pool.mu.Unlock()
+
+		// destroy long-running sessions
+		if hc.pool.CloseInactiveTransactions {
+			for _, sh := range longRunningSessions {
+				// removes inner session out of the pool to reduce the probability of two processes trying
+				// to use the same session at the same time.
+				sh.destroy()
```
olavloite, line 1468: Just to be sure: Does this `destroy` method take care of all house-keeping regarding the number of sessions in the pool etc?

spanner/client_test.go
```diff
@@ -208,6 +208,74 @@ func TestClient_Single_Read_SessionNotFound(t *testing.T) {
...
+	defer p.mu.Unlock()
+	if g, w := p.idleList.Len(), 1; g != w {
+		t.Fatalf("Sessions in pool count mismatch\nGot: %d\nWant: %d\n", g, w)
+	}
+	if p.numInUse != 0 {
+		t.Fatalf("Expect number of sessions in use to be 0, got %d", p.numInUse)
+	}
+	if p.numOpened != 1 {
+		t.Fatalf("Expect session pool size 1, got %d", p.numOpened)
+	}
```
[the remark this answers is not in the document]
harshachinta: Done

spanner/session.go
```diff
@@ -42,6 +42,30 @@ import (
...
+	CloseInactiveTransactions bool
+	// LogInactiveTransactions is the configuration to log inactive transactions.
+	// Any inactive transaction gets logged only once.
+	LogInactiveTransactions bool
+	// long-running transactions will be cleaned up if utilisation is
+	// greater than the below value.
+	usedSessionsRatioThreshold float64
+	// A transaction is considered to be idle if it has not been used for
+	// a duration greater than the below value.
+	idleTimeThresholdSecs uint64
```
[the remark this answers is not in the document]
harshachinta: I agree with you. I have updated the changes.

spanner/session.go
```diff
@@ -582,6 +610,27 @@ func newSessionPool(sc *sessionClient, config SessionPoolConfig) (*sessionPool,
 	if config.healthCheckSampleInterval == 0 {
 		config.healthCheckSampleInterval = time.Minute
 	}
+	if config.CloseInactiveTransactions || config.LogInactiveTransactions {
+		config.TrackSessionHandles = true
```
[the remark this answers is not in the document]
harshachinta: @olavloite 
For the `CloseInactiveTransactions` or `LogInactiveTransactions` functionality we need to keep a track of every session that is checkout from the pool. Existing `TrackSessionHandles` property does this explicitly where it tracks every session in the list in [newSessionHandle](https://github.com/googleapis/google-cloud-go/blob/2b99e4f39be20fe21e8bc5c1ec1c0e758222c46e/spanner/session.go#L757) method. That is why it has to be made `true` when this feature is rolled out.
Otherwise we would have to write similar logic separately for this feature, which seems redundant. With out tracking sessions we cannot identify long running sessions for removal.

Please help me out if I am missing something here?

spanner/session.go
```diff
@@ -633,6 +682,40 @@ func (p *sessionPool) recordStat(ctx context.Context, m *stats.Int64Measure, n i
...
+	var longRunningSessions []*sessionHandle
+	if usedSessionsRatio > p.usedSessionsRatioThreshold {
+		element := p.trackedSessionHandles.Front()
+		for element != nil {
+			sh := element.Value.(*sessionHandle)
+			sh.mu.Lock()
+			diff := time.Now().Sub(sh.checkoutTime)
+			if !sh.isLongRunningTransaction && diff.Seconds() >= float64(p.idleTimeThresholdSecs) {
+				if sh.stack != nil && !sh.isSessionLeakLogged {
+					logf(p.sc.logger, "long running session %s checked out of pool at %s due to possible session leak for goroutine: \n%s", sh.session.getID(), sh.checkoutTime.Format(time.RFC3339), sh.stack)
```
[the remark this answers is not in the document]
harshachinta: Added conditions to produce different log messages based on the options enabled by the customers.

spanner/client_test.go
```diff
@@ -870,6 +938,59 @@ func TestClient_ReadWriteTransaction_SessionNotFoundOnExecuteUpdate(t *testing.T
...
+		// The background task cleans up this long-running session.
+		tx.sh.mu.Lock()
+		tx.sh.checkoutTime = time.Now().Add(-time.Hour)
+		if g, w := tx.sh.isLongRunningTransaction, false; g != w {
+			tx.sh.mu.Unlock()
+			return status.Errorf(codes.FailedPrecondition, "isLongRunningTransaction value mismatch\nGot: %v\nWant: %v", g, w)
+		}
+		tx.sh.mu.Unlock()
+		// wait for maintainer to clean up long-running sessions
+		time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
harshachinta: Updated the tests to have a sleep time of `30ms`.

spanner/client_test.go
```diff
@@ -208,6 +208,74 @@ func TestClient_Single_Read_SessionNotFound(t *testing.T) {
...
+	single := client.Single()
+	iter := single.Query(ctx, NewStatement(SelectSingerIDAlbumIDAlbumTitleFromAlbums))
+	p := client.idleSessions
+	sh := single.sh
+	// simulate session to be checked out for more than 60mins
+	sh.mu.Lock()
+	sh.checkoutTime = time.Now().Add(-time.Hour)
+	sh.mu.Unlock()
+	// Allow maintainer to clean up long-running sessions
+	time.Sleep(3 * time.Second)
```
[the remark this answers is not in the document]
harshachinta: Speaking out loud,
1. We cannot manually trigger this [cleanup function](https://github.com/googleapis/google-cloud-go/pull/8177/files#diff-9479303281d57658abd89eb39a8b1537fb58fca056d655901309ed1839c443f3R1482) as it is embedded with in the existing [maintainer()](https://github.com/googleapis/google-cloud-go/blob/2b99e4f39be20fe21e8bc5c1ec1c0e758222c46e/spanner/session.go#L1344), which is a background task. 
2. As the cleanup functionality is with in the background task we need to wait until it gets triggered next time.

I feel we cannot avoid sleep, but we can reduce the sleep time to `milliseconds`.  This way the unit tests do not take much longer time.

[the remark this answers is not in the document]
harshachinta: @olavloite 
I have moved the cleanup logic to a separate function and invoking this method at regular intervals using the existing maintainer (background task). This way I need not rely on invoking the maintainer in unit tests and instead manually trigger the cleanup function to clean unexpectedly long-running sessions.

Also this eliminates the need to use `sleep` in unit tests.
However, I have one unit test to test if the maintainer is successful in invoking the clean up task and this would need `sleep` (in order of `milliseconds`).

I hope this is good now. Please let me know if something needs to be changed here.

spanner/session.go
```diff
@@ -1365,6 +1483,15 @@ func (hc *healthChecker) maintainer() {
 			hc.pool.lastResetTime = now
 		}
 		hc.pool.mu.Unlock()
+
+		// task to remove or log sessions which are unexpectedly long-running
+		if now.After(hc.pool.InactiveTransactionRemovalOptions.lastExecutionTime.Add(hc.pool.executionFrequency)) {
+			if hc.pool.CloseInactiveTransactions || hc.pool.LogInactiveTransactions {
+				hc.pool.removeLongRunningSessions()
```
arpan14, line 1490: * What happens if an exception gets thrown from this method? Is it handled such that it does not disrupt other things in the initpool() method?
* Also should this new task be the last task of this method?

spanner/session.go
```diff
@@ -633,6 +686,69 @@ func (p *sessionPool) recordStat(ctx context.Context, m *stats.Int64Measure, n i
...
+
+// removes or logs sessions that are unexpectedly long-running.
+func (p *sessionPool) removeLongRunningSessions() {
+	p.mu.Lock()
+	longRunningSessions := p.getLongRunningSessionsLocked()
+	p.mu.Unlock()
+
+	// destroy long-running sessions
+	if p.CloseInactiveTransactions {
+		for _, sh := range longRunningSessions {
```
arpan14, line 741: We can't do this inline (at time of detection) and avoid re-iterating this list? Is there other reasons why you are preferring a two-step approach where you are first fetching the long running sessions and then re-iterating the list and destroying?

spanner/session.go
```diff
@@ -633,6 +686,69 @@ func (p *sessionPool) recordStat(ctx context.Context, m *stats.Int64Measure, n i
...
+					sh.isSessionLeakLogged = true
+				}
+				if p.CloseInactiveTransactions && !sh.isSessionLeakLogged {
+					longRunningSessions = append(longRunningSessions, sh)
+					if sh.stack != nil {
+						logf(p.sc.logger, "session %s checked out of pool at %s is long running and will be removed due to possible session leak for goroutine: \n%s", sh.session.getID(), sh.checkoutTime.Format(time.RFC3339), sh.stack)
+					} else {
+						logf(p.sc.logger, "session %s checked out of pool at %s is long running and will be removed due to possible session leak for goroutine: \nStack trace not available", sh.session.getID(), sh.checkoutTime.Format(time.RFC3339))
+					}
+					sh.isSessionLeakLogged = true
```
arpan14, line 723: Nit : There is minor duplication in these two blocks. Can we refactor a bit?
* `sh.isSessionLeakLogged = true` is repeated and can be done only once. This can be simplified if we add a check at L707 as 

```
if !sh.isSessionLeakLogged {
     sh.isSessionLeakLogged = true;
}
```

spanner/transaction.go
```diff
@@ -1124,6 +1129,14 @@ func (t *ReadWriteTransaction) batchUpdateWithOptions(ctx context.Context, stmts
 	if err != nil {
 		return nil, err
 	}
+	// mark transaction and session to be eligible for long-running
+	t.mu.Lock()
+	t.isLongRunningTransaction = true
+	t.mu.Unlock()
+	t.sh.mu.Lock()
+	t.sh.isLongRunningTransaction = true
```
arpan14, line 1137: Post a successful commit, session is returned back to the pool. In that case we need to reset these properties so that the state is not stale. For ex - If a session was being used with PDML, then these booleans will be true. But if the PDML is closed and next if it gets assigned to a read-only transaction, it should have these booleans as false. 

How is this being ensured?

internal/kokoro/presubmit.sh
```diff
@@ -47,8 +47,8 @@ exit_code=0
 # Run tests in the current directory and tee output to log file,
 # to be pushed to GCS as artifact.
 runPresubmitTests() {
-  if [[ $PWD != *"/internal/"* ]] ||
-    [[ $PWD != *"/third_party/"* ]]; then
+  if [[ $PWD == *"/internal/"* ]] ||
+    [[ $PWD == *"/third_party/"* ]]; then
```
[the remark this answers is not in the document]
harshachinta: Done. This fix is applied in https://github.com/googleapis/google-cloud-go/pull/8250

spanner/client.go
```diff
@@ -557,6 +559,14 @@ func (c *Client) rwTransaction(ctx context.Context, f func(context.Context, *Rea
 		if sh == nil || sh.getID() == "" || sh.getClient() == nil {
 			// Session handle hasn't been allocated or has been destroyed.
 			sh, err = c.idleSessions.take(ctx)
```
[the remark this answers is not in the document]
harshachinta: 1. The session pool by itself is named as idleSessions field in the Client Object.
https://github.com/googleapis/google-cloud-go/blob/f609b3cf831fb89c45386f81d0047560120cb3f4/spanner/client.go#L90-L94

2. The `take` method fetches the sessions available in an idleList from the session pool. Here `idleList` has session IDs that can be allocated for use.
https://github.com/googleapis/google-cloud-go/blob/f609b3cf831fb89c45386f81d0047560120cb3f4/spanner/session.go#L874-L879

**Note**: Any new session that is created in the session pool gets added to the idleList field of session pool.
https://github.com/googleapis/google-cloud-go/blob/f609b3cf831fb89c45386f81d0047560120cb3f4/spanner/session.go#L665-L677

spanner/client.go
```diff
@@ -557,6 +559,14 @@ func (c *Client) rwTransaction(ctx context.Context, f func(context.Context, *Rea
...
+			if t != nil {
+				// when a batch update operation is called on this transaction, isLongRunningTransaction will be true
```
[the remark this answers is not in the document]
harshachinta: Yes. This doc helps developers understand why this condition was added here.
To give some context, [BatchUpdate](https://github.com/googleapis/google-cloud-go/blob/f609b3cf831fb89c45386f81d0047560120cb3f4/spanner/transaction.go#L1101) method can be called using a ReadWriteTransaction. If there is a retry that happens then a new session will get fetched and if the retry is for a `BatchUpdate` then this new session qualifies to be long running.
If this doc is not added, then there might be confusion that every `ReadWriteTransaction` is not long running (which is not true) and what is the need of this condition here.

spanner/pdml.go
```diff
@@ -53,6 +53,10 @@ func (c *Client) partitionedUpdate(ctx context.Context, statement Statement, opt
 	}
 
 	sh, err := c.idleSessions.take(ctx)
+	// Mark isLongRunningTransaction to true, as the session in case of partitioned dml can be long-running
+	sh.mu.Lock()
```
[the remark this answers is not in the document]
harshachinta: I recall that when a `Lock` was not taken here, there were some data-race conditions because the test cases were trying to read this property `isLongRunningTransaction` for validation.

spanner/session.go
```diff
@@ -42,6 +42,30 @@ import (
...
 const healthCheckIntervalMins = 50
 
+// InactiveTransactionRemovalOptions has configurations for action on long-running transactions.
+type InactiveTransactionRemovalOptions struct {
+	mu sync.Mutex
+	// CloseInactiveTransactions is the configuration to close inactive transactions.
+	CloseInactiveTransactions bool
+	// LogInactiveTransactions is the configuration to log inactive transactions.
+	// Any inactive transaction gets logged only once.
+	LogInactiveTransactions bool
```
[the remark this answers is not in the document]
harshachinta: Here these booleans serve two different purposes and can be independently used,
1. `LogInactiveTransactions` - logs the long running transaction once.
2. `CloseInactiveTransactions` - logs the long running transaction + closes the transaction.

spanner/session.go
```diff
@@ -633,6 +686,69 @@ func (p *sessionPool) recordStat(ctx context.Context, m *stats.Int64Measure, n i
 	recordStat(ctx, m, n)
 }
 
+func (p *sessionPool) getRatioOfSessionsInUseLocked() float64 {
+	maxSessions := p.MaxOpened
```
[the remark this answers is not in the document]
harshachinta: The `MaxOpened` property is a part of `SessionPoolConfig` that can be defined by the user during `client` creation.
This defaults to `NumChannels * 100`.
https://github.com/googleapis/google-cloud-go/blob/f609b3cf831fb89c45386f81d0047560120cb3f4/spanner/session.go#L364-L370

spanner/session.go
harshachinta: The below are the reasons why it cannot be done inline (destroying at the time of detection)
1. Locking is one reason. The existing `sh.destroy()` method takes a lock on session pool internally. Similarly, in `getLongRunningSessionsLocked` method, to iterate `p.trackedSessionHandles` (linked list that contains sessions in use) we need to take a lock on session pool. This means if we want to call destroy during iteration, then we need to remove the lock on session pool, then call the destroy method and then wait for the next lock to be acquired on the session pool for moving ahead with the iteration.
2. Also, looking at the `getLongRunningSessionsLocked` method, the  `trackedSessionHandles` is a **linked list**, which contains a list of all the sessions that are currently in use. When iterating over a linked list using a loop, we cannot directly delete values from the list during the iteration which can result in unexpected behaviour or runtime errors.

Thus a two-step approach is the preferable and safer operation.

spanner/session.go
harshachinta: Yes, this can be done. But I guess this will come after L723. If we do in L707, then the other two if checks will be ignored always.

But, I wonder refactoring this would add an additional `if` check to make the `isSessionLeakLogged ` property `true`. which could have been done with in the `if` statements that will get executed already. WDYT?

spanner/transaction.go
harshachinta: The `isLongRunningTransaction` property is set on `SessionHandle` object and not on the `Session` object. This `SessionHandle` is only linked for a transaction when it fetches a session from the pool, which means that we need not worry about resetting these properties.
https://github.com/googleapis/google-cloud-go/blob/f609b3cf831fb89c45386f81d0047560120cb3f4/spanner/session.go#L755-L757.

Post a successful commit, session is returned back to the pool which does not have the `isLongRunningTransaction` property.

spanner/session.go
harshachinta: 1. - `hc.pool.removeLongRunningSessions()` does 2 tasks. First, calls `p.getLongRunningSessionsLocked`, which iterates the list and captures the sessions that are not eligible to be long-running. Exception is not possible here
https://github.com/googleapis/google-cloud-go/pull/8177/files/f578facefc79db96da529a304394f4894a8c3529#diff-9479303281d57658abd89eb39a8b1537fb58fca056d655901309ed1839c443f3R698
   -  Secondly, it iterates over the `longRunningSessions` to destroy the sessions. For this we use existing `sh.destroy()` which does not throw any errors as a return type of this method. So, exception is not possible here too.
   https://github.com/googleapis/google-cloud-go/blob/f609b3cf831fb89c45386f81d0047560120cb3f4/spanner/session.go#L137

2. - This task cannot be moved at the end of this method, because there is a `select` which is a blocking call and waits till the window duration.
   -  I have added this task before `Grow or shrink pool` task to get some benefit though it is not really high. Ex: if our session cleaning tasks removes couple of sessions, then the grow task can increase the session to above minSessions instead of waiting until the next cycle to grow its size.

spanner/transaction.go
```diff
@@ -1385,6 +1398,13 @@ func (t *ReadWriteTransaction) begin(ctx context.Context) error {
 			if err != nil {
 				return err
 			}
+
+			sh.mu.Lock()
+			t.mu.Lock()
+			// for batch update operations, isLongRunningTransaction will be true
+			sh.isLongRunningTransaction = t.isLongRunningTransaction
```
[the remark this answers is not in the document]
harshachinta: Yes. We need to do this whenever a new session is taken out from the pool and assigned to a transaction, which happens when a new transaction has started, or when a retry is taking place.

spanner/session.go
```diff
@@ -60,6 +84,10 @@ type sessionHandle struct {
 	// stack is the call stack of the goroutine that checked out the session
 	// from the pool. This can be used to track down session leak problems.
 	stack []byte
+	// isLongRunningTransaction tells if the inner session is eligible to be long-running.
+	isLongRunningTransaction bool
```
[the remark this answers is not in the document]
harshachinta: Done

spanner/client.go
```diff
@@ -438,10 +438,11 @@ func (c *Client) BatchReadOnlyTransaction(ctx context.Context, tb TimestampBound
...
 		ReadOnlyTransaction: ReadOnlyTransaction{
-			tx:              tx,
-			txReadyOrClosed: make(chan struct{}),
-			state:           txActive,
-			rts:             rts,
+			tx:                       tx,
+			txReadyOrClosed:          make(chan struct{}),
+			state:                    txActive,
+			rts:                      rts,
+			isLongRunningTransaction: true,
```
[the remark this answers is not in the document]
harshachinta: Yes, 'BatchReadOnlyTransaction' creates its own session and does not fetch session from pool.
https://github.com/googleapis/google-cloud-go/blob/f609b3cf831fb89c45386f81d0047560120cb3f4/spanner/client.go#L416-L423

1. When I wrote this I was wondering how is retry handled for `BatchReadOnlyTransaction`, if retry was there is that new session being used from pool or a new session is created and as I was not clear at that time, I used this to prevent something unexpected happening.
But looks like `BatchReadOnlyTransaction` does not have any retry logic in `Go` client library. Is this the same with `Java` also?
2. Also one more confusion was what should happen if a 'BatchReadOnlyTransaction' executes a normal query. Example https://github.com/googleapis/google-cloud-go/blob/f609b3cf831fb89c45386f81d0047560120cb3f4/spanner/integration_test.go#L3326-L3335
Logically in this case the query is not eligible for long running though the transaction can be long running.

I think marking `isLongRunningTransaction ` as `true` or `false` for `BatchReadOnlyTransaction` does not have any effect.

spanner/session.go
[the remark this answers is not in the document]
olavloite: I think that you can re-use much of the logic that is used for that, but I also think that it would make sense to make sure that the [capturing of the stack trace](https://github.com/googleapis/google-cloud-go/blob/2b99e4f39be20fe21e8bc5c1ec1c0e758222c46e/spanner/session.go#L761) of the go routine that checks out the session is still optional and turned off by default.

So what I would suggest would be to:
1. Keep the `TrackSessionHandles` property false by default as it is at the moment.
2. Change [this if condition](https://github.com/googleapis/google-cloud-go/blob/2b99e4f39be20fe21e8bc5c1ec1c0e758222c46e/spanner/session.go#L757) into something like `if p.TrackSessionHandles || config.CloseInactiveTransactions || config.LogInactiveTransactions`
3. Add an extra `if p.TrackSessionHandles` block around [this line](https://github.com/googleapis/google-cloud-go/blob/2b99e4f39be20fe21e8bc5c1ec1c0e758222c46e/spanner/session.go#L761).
4. Make sure that all code that uses the tracked session handles is able to handle the fact that the stack trace could be nil.

The reason for the above is that the `debug.Stack()` call is a bit expensive, and it's a debug call that you do not necessarily want to get called for everything that checks a session out of the pool (especially as it is turned off by default at the moment).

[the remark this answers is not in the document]
harshachinta: Thanks @olavloite.
I have made changes to enable stack trace explicitly only when `TrackSessionHandles` is enabled for long running sessions detection.

spanner/session.go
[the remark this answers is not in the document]
olavloite: I think that the point here is that once you set `CloseInactiveTransactions` to `true`, the value of `LogInactiveTransactions` is ignored. So in that sense they cannot be 'used' independently, and a three-state enum would make more sense (so something like `DoNothing`, `Log` and `LogAndClose`).

Another option could be to really make them independent, and let `CloseInactiveTransactions` skip logging unless `LogInactiveTransactions` has also been set, but I would not be in favor of such a solution.

olavloite (approved):

spanner/client.go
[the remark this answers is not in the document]
arpan14: Ah, then its a naming issue. Probably we can fix the naming in a separate change to improve the readability here.

spanner/client.go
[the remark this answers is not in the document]
arpan14: Yeah, my point is `BatchUpdate` is the only API that uses this currently. But there can be other APIs which will be eligible in future. (For ex - BatchWrite will soon be added in all client libraries, WIP PR in Java - https://github.com/googleapis/java-spanner/pull/2520/). When that's launched in Golang, this piece of documentation will become stale. Because while adding support for new APIs, one would not remember to update the documentation and there's no way to ensure it done. Hence as a general practice, documentation shouldn't include any details that can change in future.

spanner/session.go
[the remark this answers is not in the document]
harshachinta: Thanks for the suggestion Arpan and Knut.
I have converted this to an enum now instead of separate boolean fields. `ActionOnInactiveTransaction` will be the public field which is an enum. The default value for this enum is `noAction`. The publicly available enum options are `Warn`, `Close` and `WarnAndClose`

spanner/client.go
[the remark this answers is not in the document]
harshachinta: Thanks, Arpan. As discussed reworded the comment.

spanner/client.go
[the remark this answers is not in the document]
rahul2393: Yes, agree with Harsha here BatchReadOnlyTransaction are not retried and create their own session, so setting isLongRunningTransaction does not have any impact

spanner/session.go
rahul2393: Yes exceptions are not possible here because error is not returned, looks good to me

arpan14 (approved):

spanner/README.md
```diff
@@ -31,4 +31,87 @@ row, err := client.Single().ReadRow(ctx, "Users",
 if err != nil {
 	log.Fatal(err)
 }
+```
```
harshachinta, line 34: @arpan14 @olavloite @rahul2393 @asthamohta 
Can you please take a look at the golang documentation for session leaks and give your LGTM?

spanner/README.md
```diff
@@ -31,4 +31,87 @@ row, err := client.Single().ReadRow(ctx, "Users",
...
+
+### Session Leak
+A `Client` object of the Client Library has a limit on the number of maximum sessions. For example the
+default value of `MaxOpened`, which is the maximum number of sessions allowed by the session pool in the 
+Golang Client Library is 400. You can configure these values at the time of
```
olavloite, line 39:
```suggestion
Golang Client Library, is 400. You can configure these values at the time of
```

spanner/README.md
```diff
@@ -31,4 +31,87 @@ row, err := client.Single().ReadRow(ctx, "Users",
...
+creating a `Client` by passing custom `SessionPoolConfig` as part of `ClientConfig`. When all the sessions are checked
+out of the session pool, every new transaction has to wait until a session is returned to the pool.
+If a session is never returned to the pool (hence causing a session leak), the transactions will have to wait
+indefinitely and your application will be blocked.
+
+#### Common Root Causes
+The most common reason for session leaks in the Golang client library are:
+1. Not stopping a `RowIterator` that is returned by `Query`, `Read` and other methods. Always use `RowIterator.Stop()` to ensure that the `RowIterator` is always closed.
+2. Not closing a `ReadOnlyTransaction` when you no longer need it. Always call `ReadOnlyTransaction.Close()` after use, to ensure that the `ReadOnlyTransaction` is always closed.
+3. Not closing a `BatchReadOnlyTransaction` when you no longer need it. Always call `BatchReadOnlyTransaction.Close()` after use, to ensure that the `BatchReadOnlyTransaction` is always closed.
```
olavloite, line 49: For a different PR: We should change the implementation of `BatchReadOnlyTransaction` to not use any sessions from the pool, and instead create a separate session.

spanner/README.md
```diff
@@ -31,4 +31,87 @@ row, err := client.Single().ReadRow(ctx, "Users",
...
+
+As shown in the example below, the `txn.Close()` statement releases the session after it is complete.
+If you don't do `txn.Close()`, the session is not released back to the pool. The recommended way is to use `defer` as shown below,
```
olavloite, line 52:
```suggestion
If you fail to call `txn.Close()`, the session is not released back to the pool. The recommended way is to use `defer` as shown below.
```

spanner/README.md
```diff
@@ -31,4 +31,87 @@ row, err := client.Single().ReadRow(ctx, "Users",
...
+  log.Fatal(err)
+}
+txn := client.ReadOnlyTransaction()
+defer txn.Close()
+```
+
+#### Debugging and Resolving Session Leaks
+
+##### Logging inactive transactions
+Enabled by default, the logging option shares warn logs when you have exhausted >95% of your session pool.
```
olavloite, line 65:
```suggestion
This option logs warnings when you have exhausted >95% of your session pool. It is enabled by default.
```

spanner/README.md
```diff
@@ -31,4 +31,87 @@ row, err := client.Single().ReadRow(ctx, "Users",
...
+This could mean two things, either you need to increase the max sessions in your session pool (as the number
```
olavloite, line 66:
```suggestion
This could mean two things; either you need to increase the max sessions in your session pool (as the number
```

spanner/README.md
```diff
@@ -31,4 +31,87 @@ row, err := client.Single().ReadRow(ctx, "Users",
...
+of queries run using the client side database object is greater than your session pool can serve) or you may
```
olavloite, line 67:
```suggestion
of queries run using the client side database object is greater than your session pool can serve), or you may
```

spanner/README.md
```diff
@@ -31,4 +31,87 @@ row, err := client.Single().ReadRow(ctx, "Users",
...
+defer client.Close()
+
+// Example Log message to warn presence of long running transactions
+// session <session-info> checked out of pool at <session-checkout-time> is long running due to possible session leak for goroutine
+// <Stack Trace of transaction>
+
+```
+
+##### Automatically clean inactive transactions
+When the option to automatically clean inactive transactions is enabled, the client library will automatically spot
```
olavloite, line 93:
```suggestion
When the option to automatically clean inactive transactions is enabled, the client library will automatically detect
```

spanner/README.md
```diff
@@ -31,4 +31,87 @@ row, err := client.Single().ReadRow(ctx, "Users",
...
+problematic transactions that are running for extremely long periods of time (thus causing session leaks) and close them.
```
olavloite, line 94:
```suggestion
problematic transactions that are running for a very long time (thus causing session leaks) and close them.
```

spanner/README.md
harshachinta: Done.

spanner/README.md
harshachinta: `BatchReadOnlyTransaction` already creates its own session and do not use any sessions from the pool. Sorry for the confusion.
I have removed this line.

spanner/README.md
harshachinta: Done

spanner/README.md
harshachinta: Done

spanner/README.md
harshachinta: Done

spanner/README.md
harshachinta: Done

rahul2393 (approved):

PR #8180, opened 2023-06-27
Add create and update table with change stream config

tonytanger:


bigtable/admin.go
```diff
@@ -269,6 +276,10 @@ func (ac *AdminClient) CreateTableFromConf(ctx context.Context, conf *TableConf)
 	} else if conf.DeletionProtection == Unprotected {
 		tbl.DeletionProtection = false
 	}
+	if conf.ChangeStreamRetention != nil && conf.ChangeStreamRetention.(time.Duration) != 0 {
```
mutianf, line 279: Do we need to check the retention is between 1 and 7?

bigtable/admin.go
```diff
@@ -307,12 +318,23 @@ type UpdateTableConf struct {
...
+	changeStreamRetention ChangeStreamRetention
+}
+
+// UpdateTableDisableChangeStream updates a table to disable change stream for table ID.
+func (ac *AdminClient) UpdateTableDisableChangeStream(ctx context.Context, tableID string) error {
+	return ac.updateTableWithConf(ctx, &UpdateTableConf{tableID, None, time.Duration(0)})
+}
+
+// UpdateTableWithChangeStream updates a table to with the given table ID and change stream config.
+func (ac *AdminClient) UpdateTableWithChangeStream(ctx context.Context, tableID string, changeStreamRetention ChangeStreamRetention) error {
```
mutianf, line 331: Same here, do we need to check the retention is between 1 and 7?

bigtable/admin.go
tonytanger: Seems like we should leave that up to server side validation https://google.aip.dev/client-libraries/4223

mutianf (approved):

PR #8184, opened 2023-06-28
fix(storage): fix AllObjects condition in gRPC

### LINKED ISSUE #6205: storage: determine if zero values are automatically set for OLM rules for gRPC
**Client**

e.g. Storage

The implementation of OLM conversion from manual to gRPC might be accidentally sending zero values if at least one condition is set in an OLM rule. 

Code block that should be investigated further: https://github.com/googleapis/google-cloud-go/blob/main/storage/bucket.go#L1560-L1571

tritone:
Fixes incorrect handling of zero values for the AgeDays field of bucket lifecycle conditions. Allows re-enabling the integration test for bucket create/delete for gRPC.

Fixes #6205


BrennaEpp (approved):
LGTM

PR #8196, opened 2023-07-03
fix(internal/retry): Simplify gRPC status code mapping of retry error

atollena:
gRPC v1.55.0 that this project depends on came with a change in semantic for status.FromError such that the gRPC status of wrapped errors is returned. The implementation of GRPCStatus in internal/retry code was doing just that. Removing GRPCStatus entirely makes the code much easier to understand: at the moment, status.FromError calls GRPCStatus which in turns calls status.FromError and it's not immediately obvious why this can't result in infinite recursion.

A previous change I made to this code made sure this method never returns Status.OK (https://github.com/googleapis/google-cloud-go/pull/8128), but I failed to realize that since this project depends on gRPC 1.55 that already handles wrapped errors in status.FromError, we can simply remove the implementation of `GRPCStatus` and let gRPC status.FromError handle unwrapping.

Note that I barely had to change the tests, but there *IS* a slight change in behavior: the message of the wrapping error is included in the gRPC error. I think it's fine, bet let me know if you think otherwise (for the same reasons discussed in https://github.com/grpc/grpc-go/pull/6150).


atollena:
Thanks to @dfawley for suggesting this here: https://github.com/grpc/grpc-go/pull/6374#issuecomment-1615163466

noahdietz:
I think this would be fine to do. I'm hoping that folks aren't matching error strings anymore at this point, and instead using `errors.Is`. Thank you again @atollena 

@codyoss any concerns?

codyoss:
@noahdietz nope seems fine to me as well. Error strings are undefined behavior in my eyes -- at least for this instance.

noahdietz (approved):

noahdietz:
@atollena I'm updating your branch and will get it merged. I wait until the continuous integration tests pass upon merging to cut a release, but will ping you on the release when it happens.

noahdietz (approved):

noahdietz (approved):

atollena:
Thank you for merging.

PR #8201, opened 2023-07-05
fix(datastore): at end of pagination return empty cursor

hermanbanken:
Datastore library is a little different in each language. In NodeJS the Query method returns a result with `info` object containing the pagination status (e.g. `MoreResults = NO_MORE_RESULTS`). In contrast in Go the Iterator pattern completely hides this pagination result, and you can not know if the page retrieved is the last page.

When drafting a simple CRUD application - much like the GCP Console Datastore Viewer - we would use code like this to paginate & return the cursor of the next page:
https://gist.github.com/hermanbanken/39368405c5d430b5310d2076f6091dad#file-main-go-L36-L61

As can be seen in the gist, you use the iterator to fully read the results, and then use `it.Cursor()` to get a pointer to more entries after the Limit:

```go
end, _ := it.Cursor()
return results, end.String(), nil // return end.String() so it can be used for the next page
```

However, when can the client know there are no more results? There are 2 options:
1. The cursor becomes empty/blank/nil again. This is common, see for example [Redis `SCAN`](https://redis.io/commands/scan/)
2. There are no more results returned

The latter, there are no more results returned is a bit tricky. Are there really no more results? Or, did the server post-process the results and filter out all Datastore results returned in the runtime because they did not match some runtime filter? Furthermore, relying on this requires the clients to retrieve N+1 pages, because only the N+1 page is going to be empty. Therefore, the approach 1 is what is more logical.

This PR changes the `iterator.Cursor()` method to return an empty cursor when the last page is finished.

-------

Alternatively, we could make a breaking client API change and expose the MoreResults enum from Iterator. The enum has the values added below, so it would be more useful than only knowing if it is the last page.

- MORE_RESULTS_TYPE_UNSPECIFIED
- NOT_FINISHED
- MORE_RESULTS_AFTER_LIMIT
- MORE_RESULTS_AFTER_CURSOR
- NO_MORE_RESULTS


bhshkh:
LGTM. Please add a integration test.

enocom:
@bhshkh In theory I agree. I think it's worth trying that API out from a client perspective and confirming that the suggestion lines up with what this PR is otherwise trying to accomplish.

codyoss:
@bhshkh have we decided to not move forward with this?

codyoss:
@bhshkh see comment above

PR #8212, opened 2023-07-05
chore(ai/generativelanguage): add config to generate apiv1beta2

quartzmo:


.github/.OwlBot.yaml
```diff
@@ -4,6 +4,7 @@ docker:
 deep-remove-regex:
   - /accessapproval/apiv1/
   - /accesscontextmanager/apiv1/
+  - /ai/generativelanguage/v1beta2
```
codyoss, line 7: Looks like for this in the value below you should add a trailing slash.

codyoss (changes requested):

.github/.OwlBot.yaml
quartzmo: Good catch, thank you!

.github/.OwlBot.yaml
```diff
@@ -4,6 +4,7 @@ docker:
 deep-remove-regex:
   - /accessapproval/apiv1/
   - /accesscontextmanager/apiv1/
+  - /ai/generativelanguage/v1beta2/
```
codyoss, line 7: across this and the other file missing api in front of the version.

quartzmo: Done, thanks!

codyoss (approved):

PR #8221, opened 2023-07-06
fix(logadmin): use consistent filter in paging example

### LINKED ISSUE #8186: logadmin: Unable to paginate entries throught different processes/requests
**Client**

logadmin.Client

**Environment**

macOS Ventura 13.4 - Apple M1 Pro

**Go Environment**

$ go version
go version go1.20.3 darwin/arm64
$ go env
GOARCH="arm64"
GOCACHE="/Users/myuser/Library/Caches/go-build"
GOENV="/Users/myuser/Library/Application Support/go/env"
GOHOSTARCH="arm64"
GOHOSTOS="darwin"
GOMODCACHE="/Users/myuser/.asdf/installs/golang/1.20.3/packages/pkg/mod"
GOOS="darwin"
GOPATH="/Users/myuser/.asdf/installs/golang/1.20.3/packages"
GOROOT="/Users/myuser/.asdf/installs/golang/1.20.3/go"
GOTOOLDIR="/Users/myuser/.asdf/installs/golang/1.20.3/go/pkg/tool/darwin_arm64"
GOVERSION="go1.20.3"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
CGO_CFLAGS="-O2 -g"
CGO_CXXFLAGS="-O2 -g"
CGO_FFLAGS="-O2 -g"
CGO_LDFLAGS="-O2 -g"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/k2/c4by0t896cl33tv1bt6dty0r0000gq/T/go-build2139043596=/tmp/go-build -gno-record-gcc-switches -fno-common"

**Code**

The code can be found [here](https://cloud.google.com/go/docs/reference/cloud.google.com/go/logging/latest/logadmin#pagination).

```go
package main

import (
        "bytes"
        "context"
        "flag"
        "fmt"
        "html/template"
        "log"
        "net/http"

        "cloud.google.com/go/logging"
        "cloud.google.com/go/logging/logadmin"
)

var (
        client    *logadmin.Client
        projectID = flag.String("project-id", "", "ID of the project to use")
)

func main() {
        // This example demonstrates how to iterate through items a page at a time
        // even if each successive page is fetched by a different process. It is a
        // complete web server that displays pages of log entries. To run it as a
        // standalone program, rename both the package and this function to "main".
        ctx := context.Background()
        flag.Parse()
        if *projectID == "" {
                log.Fatal("-project-id missing")
        }
        var err error
        client, err = logadmin.NewClient(ctx, *projectID)
        if err != nil {
                log.Fatalf("creating logging client: %v", err)
        }

        http.HandleFunc("/entries", handleEntries)
        log.Print("listening on 8080")
        log.Fatal(http.ListenAndServe(":8080", nil))
}

var pageTemplate = template.Must(template.New("").Parse(`
<table>
  {{range .Entries}}
    <tr><td>{{.}}</td></tr>
  {{end}}
</table>
{{if .Next}}
  <a href="/entries?pageToken={{.Next}}">Next Page</a>
{{end}}
`))

func handleEntries(w http.ResponseWriter, r *http.Request) {
        ctx := context.Background()
        filter := fmt.Sprintf(`logName = "projects/%s/logs/testlog"`, *projectID)
        it := client.Entries(ctx, logadmin.Filter(filter))
        var entries []*logging.Entry
        nextTok, err := iterator.NewPager(it, 5, r.URL.Query().Get("pageToken")).NextPage(&entries)
        if err != nil {
                http.Error(w, fmt.Sprintf("problem getting the next page: %v", err), http.StatusInternalServerError)
                return
        }
        data := struct {
                Entries []*logging.Entry
                Next    string
        }{
                entries,
                nextTok,
        }
        var buf bytes.Buffer
        if err := pageTemplate.Execute(&buf, data); err != nil {
                http.Error(w, fmt.Sprintf("problem executing page template: %v", err), http.StatusInternalServerError)
        }
        if _, err := buf.WriteTo(w); err != nil {
                log.Printf("writing response: %v", err)
        }
}
```

**Expected behavior**

When a request is made providing a valid pagination token in the `pageToken` query parameter (e.g. `http://localhost:8080/entries?pageToken=EAA4n_GIvMjgk6THAUq0BSIe...`) the corresponding page of the list of logs is returned.

**Actual behavior**

When you provide the `pageToken` query param with a valid pagination token, you get an error saying `problem getting the next page: rpc error: code = InvalidArgument desc = page_token doesn't match arguments from the request`

**Additional context**

go 1.20

require (
	cloud.google.com/go/logging v1.7.0
	google.golang.org/api v0.129.0
)

noahdietz:
Without supplying a `timestamp` in the `ListLogEntries.Filter`, the client by default sets one for the last twenty-four hours when the `EntryIterator` is created. Since this example creates a new `EntryIterator` on each request, a new `timestamp` of a slightly different twenty-four hour window is added. When the `Next Page` button is pressed, it reuses the `NextPageToken` from the previous call, but that was created for a different twenty-four hour window `Filter`. This creates a mismatch between the requested `Filter` and the `Filter` used to create the `NextPageToken`, resulting in an error.

To address this, we create the filter in the main function before server start up. In this filter, we set a timestamp **explicitly**. 

Note: I made this a `fix` commit because this is part of the GoDoc examples and we need this to appear as the latest release, otherwise no one will see it until another release comes along.

Fixes #8186.

### DIFF
--- a/logging/logadmin/example_paging_test.go
+++ b/logging/logadmin/example_paging_test.go
@@ -22,6 +22,7 @@ import (
 	"html/template"
 	"log"
 	"net/http"
+	"time"
 
 	"cloud.google.com/go/logging"
 	"cloud.google.com/go/logging/logadmin"
@@ -31,6 +32,7 @@ import (
 var (
 	client    *logadmin.Client
 	projectID = flag.String("project-id", "", "ID of the project to use")
+	filter    string
 )
 
 func ExampleClient_Entries_pagination() {
@@ -48,6 +50,15 @@ func ExampleClient_Entries_pagination() {
 	if err != nil {
 		log.Fatalf("creating logging client: %v", err)
 	}
+
+	// Filter for logs of a specific name.
+	logName := fmt.Sprintf(`logName = "projects/%s/logs/testlog"`, *projectID)
+
+	// Filter for logs from the last twenty-four hours.
+	yesterday := time.Now().Add(-24 * time.Hour).UTC()
+	dayAgo := fmt.Sprintf("timestamp >= %q", yesterday.Format(time.RFC3339))
+
+	filter = fmt.Sprintf("%s AND %s", logName, dayAgo)
 
 	http.HandleFunc("/entries", handleEntries)
 	log.Print("listening on 8080")
@@ -67,7 +78,6 @@ var pageTemplate = template.Must(template.New("").Parse(`
 
 func handleEntries(w http.ResponseWriter, r *http.Request) {
 	ctx := context.Background()
-	filter := fmt.Sprintf(`logName = "projects/%s/logs/testlog"`, *projectID)
 	it := client.Entries(ctx, logadmin.Filter(filter))
 	var entries []*logging.Entry
 	nextTok, err := iterator.NewPager(it, 5, r.URL.Query().Get("pageToken")).NextPage(&entries)


codyoss (approved):
Thanks

PR #8225, opened 2023-07-06
chore(storage): add GCSFuse client config [benchmarking]

BrennaEpp:


tritone (commented):
Two small things, overall lgtm!

storage/internal/benchmarks/client_pool.go
```diff
@@ -156,20 +157,33 @@ func getClient(ctx context.Context, api benchmarkAPI) *storage.Client {
 // mutex on starting a client so that we can set an env variable for GRPC clients
 var clientMu sync.Mutex
 
-func initializeHTTPClient(ctx context.Context, writeBufferSize, readBufferSize int, json bool) (*storage.Client, error) {
+func initializeHTTPClient(ctx context.Context, writeBufferSize, readBufferSize int, json bool, gcsFuse bool) (*storage.Client, error) {
```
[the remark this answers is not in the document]
BrennaEpp: Yeah I thought so as well, might as well do now I guess

storage/internal/benchmarks/main.go
```diff
@@ -171,6 +173,8 @@ func parseFlags() {
 	flag.IntVar(&opts.workload, "workload", 1, "which workload to run")
 	flag.IntVar(&opts.numObjectsPerDirectory, "directory_num_objects", 1000, "total number of objects in directory")
 
+	flag.BoolVar(&opts.gcsFuse, "gcs_fuse", false, "use GCSFuse configs on client creation")
```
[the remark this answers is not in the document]
BrennaEpp: > Maybe put this next to the other transport-related configs. 

We should really get a config file for the CLI 😅 

> I am actually not sure if they use any non-default configs for gRPC currently -- @raj-prince any idea?

That would be good to know, I was planning on only adding HTTP probes

tritone (approved):
LGTM, thanks!

storage/internal/benchmarks/main.go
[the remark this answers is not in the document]
tritone: Sounds good; we can add opts for gRPC if/when it arises.

storage/internal/benchmarks/main.go
```diff
@@ -80,6 +80,8 @@ type benchmarkOptions struct {
 	numClients             int
 	workload               int
 	numObjectsPerDirectory int
+
+	gcsFuse bool
```
raj-prince, line 84: we can rename this boolean flag: gcsFuse -> useGCSFuseConfig

storage/internal/benchmarks/client_pool.go
```diff
@@ -156,20 +174,41 @@ func getClient(ctx context.Context, api benchmarkAPI) *storage.Client {
...
+				),
+			}
+		}
+
+		base.MaxIdleConnsPerHost = 100 // this is set in Storage as well
+		base.WriteBufferSize = config.writeBufferSize
+		base.ReadBufferSize = config.readBufferSize
+
+		if !config.setGCSFuseOpts {
+			http2Trans, err := http2.ConfigureTransports(base)
```
raj-prince, line 208: Where are we using http2Trans? This looks un-used. We can move this block to else part of line:192 if block.

storage/internal/benchmarks/client_pool.go
```diff
@@ -156,20 +174,41 @@ func getClient(ctx context.Context, api benchmarkAPI) *storage.Client {
...
 		base := http.DefaultTransport.(*http.Transport).Clone()
-		base.MaxIdleConnsPerHost = 100 // this is set in Storage as well
-		base.WriteBufferSize = writeBufferSize
-		base.ReadBufferSize = readBufferSize
 
-		http2Trans, err := http2.ConfigureTransports(base)
-		if err == nil {
-			http2Trans.ReadIdleTimeout = time.Second * 31
+		if config.setGCSFuseOpts {
+			base = &http.Transport{
```
raj-prince, line 193: We don't need to create another object, since we are cloning the transport object in line:190. We can just override the attributes. Like,
base.MaxConnsPerHost = 100
base.MaxIdleConnsPerHost = 100
base.TLSNextProto = make(map[string]func(string, *tls.Conn) http.RoundTripper)

raj-prince (changes requested):
Functionality wise looks good, have added minor refactoring comments.

storage/internal/benchmarks/client_pool.go
tritone: It's a bit of a confusing API but `http2.ConfigureTransports` returns a pointer to the transport config. Thus we can modify the values that will be used by the client without having to do pass http2trans back in anywhere.

raj-prince: Ohh, got it.

storage/internal/benchmarks/main.go
BrennaEpp: sure

storage/internal/benchmarks/client_pool.go
BrennaEpp: I'd rather keep the exact initialization code as gcsFuse has. `http.DefaultTransport` configures several transport variables to default values, and I'm not sure manually initializing an `http.Transport` would have those configured the same.

raj-prince: Okay.

raj-prince (approved):

PR #8234, opened 2023-07-07
test(bigquery): delete stale datasets

alvarowolfx:
In some cases where tests fail we are leaking BQ datasets. So the approach to avoid this is to have a helper that iterates over a datasets and deletes any resources older than a day


bigquery/integration_test.go
```diff
@@ -272,6 +275,44 @@ func initTestState(client *Client, t time.Time) func() {
...
+		if !strings.HasPrefix(ds.DatasetID, prefix) {
+			continue
+		}
+
+		md, err := ds.Metadata(ctx)
+		if err != nil {
+			fmt.Printf("failed to get dataset `%s` metadata: %v", ds.DatasetID, err)
+			continue
+		}
+		if isResourceStale(md.CreationTime) {
```
shollyman, line 306: Why not [`time.Since()`](https://pkg.go.dev/time#Since) and avoid the extra function definition?

bigquery/integration_test.go
```diff
@@ -272,6 +275,44 @@ func initTestState(client *Client, t time.Time) func() {
...
+func deleteDatasets(ctx context.Context, prefix string) {
+	it := client.Datasets(ctx)
+
+	for {
+		ds, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			fmt.Println("failed to list project datasets: ", err)
```
shollyman, line 294: why Println?  Would have expected a parameterized Printf here.

shollyman (commented):
Couple of minor nits, and one further consideration:  we have similar possibility of dataset accrual due to integration testing in bigquery/storage/managedwriter, does it make sense to have a common function both suites can invoke, or should we just use this one for backstop cleanup and define it to match multiple prefixes?

bigquery/integration_test.go
alvarowolfx: I added the extra function so the logic for what is a stale resource can be reused. For example, later maybe we need to clean other resources like Storage Buckets. I changed to use `time.Since`

PR #8250, opened 2023-07-12
fix: kokoro presubmit builds

harshachinta:
In ideal case, the GitHub presubmit tests should run all the tests of changed submodules when a PR is raised. But this is not happening and the builds are showing green which misguides developers to assume that all tests are passing.

Example build:
image

This happened because the current condition says
```
runPresubmitTests() {
  if [[ $PWD != *"/internal/"* ]] ||
    [[ $PWD != *"/third_party/"* ]]; then
    return
  fi
  // run tests
```
This says
1. If the current directory do not have `internal` or `third_party` text in the string, then return from the method with out executing further.
2. changed modules (ex: spanner) pwd will not have these strings and hence will come out of method without executing tests.
**This is affecting all the existing directories with in the google-cloud-go repository.**

This PR fixes the above issue.

### DIFF
--- a/internal/kokoro/continuous.sh
+++ b/internal/kokoro/continuous.sh
@@ -78,8 +78,8 @@ try3 go mod download
 # runDirectoryTests runs all tests in the current directory.
 # If a PATH argument is specified, it runs `go test [PATH]`.
 runDirectoryTests() {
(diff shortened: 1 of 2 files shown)


codyoss (approved):
Thanks!

PR #8252, opened 2023-07-12
Fix LogEntrySourceLocation

talgi:
fix: make sure LogEntry is consistent with JSON events sent from GCP

Since in LogEntrySourceLocation the 'Line' field is actually passed as string from GCP (see here: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogEntrySourceLocation) And since logpb.LogEntrySourceLocation is generated from proto file and has no way of assing the json:string flag in gen-proto:

This modified the LogEntry to use a new type in order to be able to unmarshall the json properly.

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -489,7 +489,23 @@ type Entry struct {
 
 	// Optional. Source code location information associated with the log entry,
 	// if any.
-	SourceLocation *logpb.LogEntrySourceLocation
+		SourceLocation *LogEntrySourceLocation
+}
+
+type LogEntrySourceLocation struct {
+	// Optional. Source file name. Depending on the runtime environment, this
+	// might be a simple name or a fully-qualified name.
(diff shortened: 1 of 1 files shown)


codyoss:
Please create issues for discussion before opening a pull request.

This marshals correctly if you use protojson as required when working with proto generated messages:

```go
package main

import (
	"fmt"

	"cloud.google.com/go/logging/apiv2/loggingpb"
)

func main() {
	l := loggingpb.LogEntrySourceLocation{
		Line: 123,
	}
	b, _ := protojson.Marshal(&l)
	fmt.Printf("%s\n", b)
	// {"line":"123"}
	var l2 loggingpb.LogEntrySourceLocation
	protojson.Unmarshal(b, &l2)
	fmt.Printf("%v\n", l2.Line)
	// 123
```

PR #8264, opened 2023-07-14
chore(firestore): Cleanup integration test resources

bhshkh:
There are many integration tests in Firestore that create multiple level of sub collections. These sub collections weren't getting cleaned up on integration test run. This PR adds the cleanup for such resources.


enocom (commented):
Could we add a little more context to the commit message?

This looks like deleting a parent document was leaving behind subcollections and so we're fixing that. But a few sentences making that clear would be helpful.

kolea2 (approved):

PR #8273, opened 2023-07-17
fix(spanner): wrap errors in inline begin transaction to avoid masking 

harshachinta:
When inline begin transaction was added, the error messages were getting masked with 
```
spanner: code = "Internal", desc = "failed inline begin transaction"
```
This PR fixes this by wrapping the error to unmask original errors.

### DIFF
--- a/spanner/errors.go
+++ b/spanner/errors.go
@@ -73,6 +73,22 @@ func (*TransactionOutcomeUnknownError) Error() string { return transactionOutcom
 // Unwrap returns the wrapped error (if any).
 func (e *TransactionOutcomeUnknownError) Unwrap() error { return e.err }
 
+// InlineBeginTransactionFailedError is wrapped in a Spanner error when the error
+// occurred during an inline begin transaction.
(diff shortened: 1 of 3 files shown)


rahul2393:
@harshachinta Thanks for picking this up, please add the test case to mock the invalid SQL statement and check if the invalid SQL error message is contained in the final error.

ajorgensen:
@harshachinta I'd be happy to take over this changeset and help push it through if thats ok with you. We run into this fairly frequently so this change would be greatly beneficial.

harshachinta:
@ajorgensen Feel free to work on top of this. I guess the changes in this PR is working as expected except there was some test failures. I don't exactly recall the pending stuff in this PR. Will try taking a look at this next week.

codyoss:
@rahul2393 do we still want to accecpt this PR?

richardartoul:
This P.R would have saved me a LOT of time. Could we get this merged please? Its a great improvement

codyoss:
@rahul2393 please see my comment above

PR #8276, opened 2023-07-18
feat(datastore): Multi DB support

bhshkh:


datastore/client.go
```diff
@@ -42,11 +42,15 @@ type datastoreClient struct {
 	md metadata.MD
 }
 
-func newDatastoreClient(conn grpc.ClientConnInterface, projectID string) pb.DatastoreClient {
+func newDatastoreClient(conn grpc.ClientConnInterface, projectID, databaseID string) pb.DatastoreClient {
+	resourcePrefixHeaderVal := "projects/" + projectID
```
enocom, line 46: nit: this could be just `v` or `resourcePrefixValue` to match the key if you prefer a long name.

datastore/datastore.go
```diff
@@ -53,11 +53,20 @@ const DetectProjectID = "*detect-project-id*"
 // the resource being operated on.
 const resourcePrefixHeader = "google-cloud-resource-prefix"
 
+// DefaultDatabaseID is ID of the default database
```
enocom, line 56: A note about how it's empty string by default would be good. Otherwise this looks like a typo.

datastore/datastore.go
```diff
@@ -69,6 +78,21 @@ type Client struct {
...
 // Call (*Client).Close() when done with the client.
 func NewClient(ctx context.Context, projectID string, opts ...option.ClientOption) (*Client, error) {
+	client, err := NewClientWithDatabase(ctx, projectID, DefaultDatabaseID, opts...)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
+}
+
+// NewClientWithDatabase creates a new Client for given dataset and database.  If the project ID is
```
enocom, line 88: Let's format this trailing line -- right now we have two spaces after the period.

datastore/integration_test.go
```diff
@@ -161,13 +183,26 @@ func newClient(ctx context.Context, t *testing.T, dialOpts []grpc.DialOption) *C
...
 }
 
+func TestIntegration_NewClient(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Integration tests skipped in short mode")
+	}
+	ctx := context.Background()
+	client, err := NewClient(ctx, testutil.ProjID())
+	if err != nil {
+		t.Errorf("NewClient: %v", err)
```
enocom, line 200: Should this be `Errorf` or `Fatalf`? Since there are no more assertions below to run, I think you can use `Fatalf` and remove the `else` below.

datastore/datastore_test.go
```diff
@@ -24,11 +24,63 @@ import (
...
+			wantErr:           true,
+			detectProjectIDFn: origDetectProjectIDFn,
+			gtransportDialPoolFn: func(ctx context.Context, opts ...option.ClientOption) (grpc.ConnPool, error) {
+				return nil, errors.New("mock error from DialPool")
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		detectProjectIDFn = tc.detectProjectIDFn
```
enocom, line 70: Shall we use subtests here?

``` golang
for _, tc := range tests {
    t.Run(desc, func(t *testing.T) { /* ... */ })
}
```

datastore/client.go
bhshkh: Done

datastore/datastore_test.go
bhshkh: Done

internal/kokoro/continuous.sh
```diff
@@ -31,6 +31,7 @@ export GOOGLE_APPLICATION_CREDENTIALS=$KOKORO_KEYSTORE_DIR/72523_go_integration_
 
 export GCLOUD_TESTS_GOLANG_PROJECT_ID=dulcet-port-762
 export GCLOUD_TESTS_GOLANG_KEY=$GOOGLE_APPLICATION_CREDENTIALS
+export GCLOUD_TESTS_GOLANG_DATASTORE_DATABASES=,database-01
```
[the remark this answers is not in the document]
bhshkh: Right now, I can't think of any but will there ever be a scenario when we would want to run tests against more than 2 databases?
If yes, then we will have to keep adding variables for additional databases.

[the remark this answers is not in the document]
kolea2: Fair enough, in that case I would only remove the need to enter the `""` empty string for default, and use this variable for additional databases. Default DB should always be run and shouldn't have to be specified.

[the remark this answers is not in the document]
bhshkh: Done

datastore/integration_test.go
```diff
@@ -161,13 +183,27 @@ func newClient(ctx context.Context, t *testing.T, dialOpts []grpc.DialOption) *C
...
+	client, err := NewClientWithDatabase(ctx, testutil.ProjID(), testParams["databaseID"].(string), opts...)
 	if err != nil {
-		t.Fatalf("NewClient: %v", err)
+		t.Fatalf("NewClientWithDatabase: %v", err)
 	}
 	return client
 }
 
+func TestIntegration_NewClient(t *testing.T) {
+	if testing.Short() {
```
kolea2, line 194: just for my understanding, what is short mode? When is it run vs the full suite?

kolea2 (approved):
lgtm just have one question

datastore/client.go
```diff
@@ -42,11 +42,15 @@ type datastoreClient struct {
 	md metadata.MD
 }
 
-func newDatastoreClient(conn grpc.ClientConnInterface, projectID string) pb.DatastoreClient {
+func newDatastoreClient(conn grpc.ClientConnInterface, projectID, databaseID string) pb.DatastoreClient {
+	resourcePrefixValue := "projects/" + projectID
+	if databaseID != "" {
```
kolea2, line 47: maybe add a nil check here as well?

kolea2 (commented):
sorry one last nit

datastore/integration_test.go
bhshkh: Short mode runs only unit tests. Kokoro uses short mode pre submit: https://github.com/googleapis/google-cloud-go/blob/main/internal/kokoro/presubmit.sh#L57

datastore/client.go
bhshkh: databaseID is of type string. In Go, strings cannot be nil.
`if databaseID != nil && databaseID != ""` throws error 

`invalid operation: databaseID != nil (mismatched types string and untyped nil)`

noahdietz (approved):
Approval for kokoro script change.

PR #8287, opened 2023-07-19
chore: re-onboard cloudbuild apiv1 /w OwlBot

codyoss:
The git history is not super clear here to why this one was disablbed to begin with. I suspect this may cause generation errors of some sorts, but lets re-enable for now and see what/if it breaks.

### DIFF
--- a/.github/.OwlBot.yaml
+++ b/.github/.OwlBot.yaml
@@ -54,7 +54,7 @@ deep-remove-regex:
   - /binaryauthorization/apiv1beta1/
   - /certificatemanager/apiv1/
   - /channel/apiv1/
-  # - /cloudbuild/apiv1/v2/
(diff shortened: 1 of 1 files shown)


quartzmo:
Does #7679 contain any clues?

codyoss:
@quartzmo Not to me. I assume it had something to do with snippetgen. So it may just work now.

quartzmo (approved):

monicacinom:
@codyoss Thank you for the fix! Is there an expected timeline for the publish?

codyoss:
@monicacinom done: https://github.com/googleapis/google-cloud-go/releases/tag/cloudbuild/v1.12.0

monicacinom:
thanks!

PR #8289, opened 2023-07-19
docs(ai): fix README.md title

quartzmo:
I'm not sure how "Vertex AI" was incorrectly used in this README when it was initially configured in #8212. Running the post-processor locally produced the change (to the correct title) in this PR.

### DIFF
--- a/ai/README.md
+++ b/ai/README.md
@@ -1,8 +1,8 @@
-# Vertex AI API
+# Generative Language API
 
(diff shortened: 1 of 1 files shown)


codyoss:
@quartzmo This is fine to me, but this file is generated so it might come back. We still may need to find out the why

codyoss (approved):

PR #8293, opened 2023-07-20
feat(firestore): SUM and AVG aggregations

bhshkh:
Currently, Firestore supports only the "COUNT" aggregation query. This pull request adds support for

1. Sum and average aggregations: "SUM" and "AVG" are applied on a specific field chosen by the user, and will return a mathematical sum and average of its values.
2. Running aggregation queries in transactions.


bhshkh:
Added DO NOT MERGE label until feature is GA

firestore/integration_test.go
```diff
@@ -2225,6 +2231,134 @@ func TestIntegration_BulkWriter(t *testing.T) {
...
+			wantErr:          true,
+		},
+	}
+
+	for _, tc := range testcases {
+		aggResult, err := tc.aggregationQuery.Get(ctx)
+		if err != nil && !tc.wantErr {
+			t.Errorf("%s: got: %v, want: nil", tc.desc, err)
+			continue
+		} else if err == nil && tc.wantErr {
```
enocom, line 2352: nit: is the `else` necessary here and below?

enocom (commented):
LGTM. One minor nit.

firestore/integration_test.go
bhshkh: Removed

enocom (approved):

firestore/integration_test.go
```diff
@@ -2355,6 +2361,136 @@ func TestIntegration_BulkWriter(t *testing.T) {
 	})
 }
 
+func TestIntegration_AggregationQueries(t *testing.T) {
```
kolea2, line 2364: can we add some tests for transactions?

kolea2 (changes requested):

firestore/integration_test.go
bhshkh: Added support for running aggregation queries in transactions. Added tests.

kolea2 (approved):

firestore/query.go
```diff
@@ -1222,6 +1226,14 @@ type AggregationQuery struct {
 	aggregateQueries []*pb.StructuredAggregationQuery_Aggregation
 	// query contains a reference pointer to the underlying structured query.
 	query *Query
+	//  tx points to an already active transaction within which the AggregationQuery runs
+	tx *Transaction
```
kolea2, line 1230: can we get this from somewhere else?

kolea2 (changes requested):

firestore/query.go
bhshkh: No. The transaction isn't available in query or client object.

bhshkh: Ideally even in Datastore, Transaction method should have been on AggregationQuery object similar to what I have done in this PR. This is because protos allow you to specify read options i.e. a transaction id in [RunAggregationQuery](https://github.com/googleapis/go-genproto/blob/main/googleapis/datastore/v1/datastore.pb.go#L550) request  as well as [RunQueryRequest](https://github.com/googleapis/go-genproto/blob/main/googleapis/datastore/v1/datastore.pb.go#L347)

kolea2 (approved):

Issue #8311, opened 2023-07-21
bigquery: managedwriter: increase operational visibility when using in-client retries


tanner-bruce:
**Is your feature request related to a problem? Please describe.**
We frequently see bursts of append rows request retries and errors. When using the built in retry mechanism, we don't receive any indication of what the errors actually are, they are hidden by the client.

**Describe the solution you'd like**
A high level metric giving the reason for the failure (status.Codes level maybe) to help diagnose the cause of the errors.

PR #8314, opened 2023-07-21
refactor(bigquery/storage/managedwriter): refactor error metrics

shollyman:
This is a small refactor to how opencensus metrics are reported when receiving responses from the server.

There's effectively two cases where we consider an error to have occurred on receive:

* Invoking gRPC Recv() on the connection emitted an error, typically a transport issue

* The service embedded an error in the response, which is more akin to an application error (failed to commit data, offset mismatch, etc).

This CL ensures we increment the existing AppendResponseError count metric in both causes, and deals with the unlikely scenario where we're unable to tag the report with the status code.  When that happens, we simply record the metric uncoded.

Towards: https://github.com/googleapis/google-cloud-go/issues/8311


bigquery/storage/managedwriter/connection.go
```diff
@@ -494,18 +494,29 @@ func connRecvProcessor(ctx context.Context, co *connection, arc storagepb.BigQue
 			resp, err := arc.Recv()
 			co.release(nextWrite)
 			if err != nil {
+				// The Recv() itself yielded an error.  We increment AppendResponseErrors by one, tagged by the status
+				// code.
+				status := grpcstatus.Convert(err)
```
alvarowolfx, line 499: I wonder if `status` can be `nil` here as this `err` is coming from a grpc stream method. Might worth add the extra check

alvarowolfx: an `io.EOF` will probably cause `grpcstatus.Convert` to return `nil`

shollyman: It wraps the error as a code.Unknown if it doesn't embed an rpc status error, ex: https://github.com/grpc/grpc-go/blob/faab8736bf73291f92b867d5dae31c927d53d508/status/status.go#L96-L125

alvarowolfx (approved):

PR #8317, opened 2023-07-24
feat(pubsub): add opentelemetry tracing for publish

hongalex:
This will one of a few of PRs to implement OpenTelemetry Tracing into the Pub/Sub client library, starting with publish side spans.

Sample trace using Cloud Trace:
image


quartzmo (approved):
LGTM

pubsub/topic.go
```diff
@@ -585,26 +593,51 @@ func (t *Topic) Publish(ctx context.Context, msg *Message) *PublishResult {
...
 	t.mu.RLock()
 	defer t.mu.RUnlock()
 	// TODO(aboulhosn) [from bcmills] consider changing the semantics of bundler to perform this logic so we don't have to do it here
 	if t.stopped {
 		ipubsub.SetPublishResult(r, "", ErrTopicStopped)
+		spanRecordError(span, ErrTopicStopped)
 		return r
 	}
 
+	ctx2, fcSpan := tracer().Start(ctx, publishFlowControlSpanName)
```
feywind, line 608: Please bear with me as I push my way through not knowing Go :D

Should the new ctx2 above be used for flowController.acquire() and such? If not, can you explain what's going on there with the contexts?

pubsub/topic.go
```diff
@@ -585,26 +593,51 @@ func (t *Topic) Publish(ctx context.Context, msg *Message) *PublishResult {
...
+		span:      span,
+		batchSpan: batchSpan,
+	}
+
+	if span.SpanContext().IsValid() {
+		if msg.Attributes == nil {
+			msg.Attributes = make(map[string]string)
+		}
+		// Inject the context from the first publish span rather than from flow control / batching.
+		otel.GetTextMapPropagator().Inject(ctx, newMessageCarrier(msg))
```
feywind, line 632: Likewise here, there's a third context being created for tracer().Start(), but neither of them are being used in the text map propagator call.

pubsub/trace.go
```diff
@@ -256,3 +265,84 @@ func withSubscriptionKey(ctx context.Context, subName string) context.Context {
...
+	}
+	return out
+}
+
+const (
+	// span names
+	publisherSpanName          = "send"
+	publishFlowControlSpanName = "publisher flow control"
+	publishSchedulerSpanName   = "publish scheduler"
+	publishRPCSpanName         = "send Publish"
```
feywind, line 313: The spec doc seems to imply that this should be just "publish"?

feywind (approved):
Overall this looks good to me :)

pubsub/topic.go
hongalex: Yeah that's probably a good idea so it reads better though in reality it doesn't make much of a difference. Contexts are just used to pass around information and the flow controller doesn't really need to know about the additional otel span bits the `ctx2` got augmented with.

pubsub/topic.go
hongalex: The `Inject` method doesn't take in the later `ctx` since we want the parent span for the subscribe side to be the publish span (the first publish span created)

pubsub/trace.go
hongalex: Yeah this should just be Publish, I went back and forth on this for a bit.

PR #8324, opened 2023-07-25
feat(bigquery/storage/managedwriter): refine connection metrics

shollyman:
This PR changes metrics instrumentation in two ways:

The AppendClientOpenView is now tagged with an error dimension, so that failures to open a connection are clearer.  We use grpc status code for the tagging, with the expectation that non-rpc errors are tagged as Unknown.

A new metric (and view), AppendRequestReconnectsView provides additional visibility into when errors during Send trigger reconnection of the underlying connection.  We attempt to also attribute this to the origin writer, so it may be tagged by error (grpc status code again) and source stream.

Additional thanks to @tanner-bruce for additional suggestions here

Towards: https://github.com/googleapis/google-cloud-go/issues/8311


tanner-bruce:
👍 LGTM, thanks!

alvarowolfx (approved):

PR #8338, opened 2023-07-27
chore(apidiff): ignore added files

noahdietz:
Have `apidiff` ignore added files so that PRs with new clients like #8337 don't have a broken `apidiff` run.

Add a `--diff-filter` flag to `changefinder` that passes through the [`--diff-filter` values](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-filterACDMRTUXB82308203) for `git diff`


quartzmo (approved):
Thank you! 🎉

PR #8348, opened 2023-07-28
feat(storage): support single-shot uploads in gRPC

### LINKED ISSUE #7798: storage: determine what to do on chunksize=0 [GRPC]

tritone:
When ChunkSize is zero, use a single shot upload (leaving the stream open between messages) rather than a resumable session. This matches better with what we do in JSON. Also reduce the buffer size to the minimum (256k).

Fixes #7798


noahdietz (approved):
Makes sense to me. Glad you/we had the forethought to put those TODOs in there.

BrennaEpp (approved):
LGTM

PR #8359, opened 2023-07-31
chore(internal/kokoro): skip tests for empty commits

noahdietz:
When there are no changes in the PR, we can skip* presubmits and exit early. This is useful for forcing releases with empty commits, github workflow-only changes, etc.

Same change applied to the continuous job as well, when there are no changes in the merged commit, the tests are skipped.

*the job logs that there were no changes and exits early with `0` exit code.


codyoss (approved):
Thanks!

PR #8363, opened 2023-07-31
fix(datastore): Correcting string representation of Key

bhshkh:
**Issue**: While getting entities from Datastore, if the queried keys have same Kind and name/ID but different namespaces, then  only one of the entities get retrieved

**Cause**: When entities are returned from Datastore in LookupResponse, they are not in the order in which the Keys are sent. E.g. If keys sent were:{ {kind, name1, namespace1}, {kind, name2, namespace1 }, {kind, name3, namespace1 }}, The response can be {entity3, entity1, entity2} instead of {entity1, entity2, entity3}. 
To make it easier for users to consume, the library arranges the entities in the order of keys before returning them to user. 

To do this, all the keys are reverse mapped to their index in the input array and then the entities received from Datastore are put at the index of the input keys in the response array. For above example, the map created is {"kind,name1": 0, "kind, name2": 1, "kind,name3": 2}. As can be seen here, the keys in the map do not take into account the namespace. So, if 2 keys have different namespaces the entities response gets overwritten.

**Fix**: Add namespace to key in the reverse map


enocom (commented):
Could we add a description to each test case to explain what it's testing. As written, this test requires careful study to understand what it's doing.

bhshkh:
> Could we add a description to each test case to explain what it's testing. As written, this test requires careful study to understand what it's doing.

Done

enocom (approved):
LGTM

kolea2 (approved):

PR #8370, opened 2023-08-01
fix(firestore): Correcting EndBefore with LimitToLast behaviour

bhshkh:
**Issue**: Using EndBefore and LimitToLast returns erroneous results. Even the entity mentioned in EndBefore is included in the result. E.g. if all entities are: 
City{Name: "Beijing"}
City{Name: "Los Angeles"}
City{Name: "San Francisco"}
City{Name: "Tokyo"}
City{Name: "Washington D.C."}
then 
```
query := client.Collection("cities").OrderBy("Name", firestore.Asc).LimitToLast(2).EndBefore("Washington D.C.")
query.Documents(ctx).GetAll()
```
should return 

City{Name: "San Francisco"} 
City{Name: "Tokyo"}

but it returns

City{Name: "Tokyo"}
City{Name: "Washington D.C."}

**Cause**: There are 2 flags startBefore and endBefore in Query . (startBefore is incorrectly named. It should be called startAt). 
When startBefore is set to true, the documents are returned starting at the document in startDoc cursor. 
When startBefore is set to false, the documents after the startDoc cursor are returned. 

When endBefore is set to true, the documents are returned before endDoc cursor document. 
When endBefore is set to false, the documents are returned including endDoc cursor document. 

LimitToLast is not supported by the backend Firestore service. The library itself modifies the request before sending it to Firestore service and then reverses the result before sending it to the user. [src](https://github.com/googleapis/google-cloud-go/blob/main/firestore/query.go#L994-L1010)

In the above example, before calling GetAll on the query i.e. before line 2:
limitToLast is true
startDoc is nil
endDoc is City{Name: "Washington D.C."}
endBefore is true
startBefore is false

On calling GetAll:
limitToLast is false
startDoc is City{Name: "Washington D.C."}
endDoc is nil
endBefore is false
startBefore is true.      <======== incorrect

Here, startBefore is set to true i.e. documents starting at Washington D.C. will be returned by Firestore which is incorrect. 

**Fix**: Set the flags correctly


enocom (approved):

danieljbruce:
What would happen in this case with a `startBefore` specified, an `endBefore` specified and then a `limitToLast` that is larger than the range that would fit between the `startBefore` and `endBefore`?

bhshkh:
> 

Only the documents in the range are retrieved.
For query `client.Collection("cities").OrderBy("Name", firestore.Asc).LimitToLast(5).EndBefore("Washington D.C.").StartAt("Los Angeles")`

result is
City{Name: "Los Angeles"}
City{Name: "San Francisco"}
City{Name: "Tokyo"}

danieljbruce:
> > 
> 
> Only the documents in the range are retrieved. For query `client.Collection("cities").OrderBy("Name", firestore.Asc).LimitToLast(5).EndBefore("Washington D.C.").StartAt("Los Angeles")`
> 
> result is City{Name: "Los Angeles"} City{Name: "San Francisco"} City{Name: "Tokyo"}

Is it possible to include that in a test case at https://github.com/googleapis/google-cloud-go/pull/8370/files#diff-73750838aad768db3e6a40472d73c1c18d6e3ba01b27f530e68b57fdf5eed513R1029?

bhshkh:
> 

Done

bhshkh:
> > > 
> > 
> > 
> > Only the documents in the range are retrieved. For query `client.Collection("cities").OrderBy("Name", firestore.Asc).LimitToLast(5).EndBefore("Washington D.C.").StartAt("Los Angeles")`
> > result is City{Name: "Los Angeles"} City{Name: "San Francisco"} City{Name: "Tokyo"}
> 
> Is it possible to include that in a test case at https://github.com/googleapis/google-cloud-go/pull/8370/files#diff-73750838aad768db3e6a40472d73c1c18d6e3ba01b27f530e68b57fdf5eed513R1029?

Added

danieljbruce (approved):

PR #8372, opened 2023-08-01
fix(bigquery): use JobFromProject for storage iterator

derekperkins:
There's no guarantee that the provided client is in the same project or location as the job, so those details need to also be provided from the job object


alvarowolfx (approved):
LGTM, thanks for sending that fix

shollyman (approved):

PR #8390, opened 2023-08-07
feat(storage): trace span covers life of a Reader

tritone:
Make the tracing span created in NewRangeReader cover the entire lifecycle of the Reader. NewRangeReader returns before the data transfer is complete, so it's more useful to track the entire period until the Reader is closed.

NewRangeReader is still tracked with a span at the level of individual transports (HTTP and gRPC).


storage/reader.go
```diff
@@ -116,6 +117,7 @@ func (o *ObjectHandle) NewRangeReader(ctx context.Context, offset, length int64)
 	}
 
 	r, err = o.c.tc.NewRangeReader(ctx, params, opts...)
+	r.ctx = ctx
```
quartzmo, line 120: This seems like strange semantics. Why does the reader's context need to be set with the same `ctx` value that was passed to `NewRangeReader`?

tritone: It's just being used to carry the span over to Reader.Close. I don't think it would work otherwise unless I am misunderstanding?

quartzmo: Doesn't `NewRangeReader` return `r` with the context set already on line 119?

tritone: Good question -- so, there are two (really three including XML) separate implementations of NewRangeReader beneath the call on L119. These implementations already have their own trace for NewRangeReader. It seems preferable to manage the overall trace for Reader at this level rather than below the transport-agnostic interface.

quartzmo: Thanks!

quartzmo (approved):
LGTM

PR #8396, opened 2023-08-08
fix(storage): fix gRPC generation/condition issues

tritone:
Fixes issues in piping through generation and preconditions for object Compose and Update. Adds a test against the emulator which validates this logic across both transports.

I'll add another test for bucket preconditions in a separate PR.


storage/grpc_client.go
```diff
@@ -803,6 +808,9 @@ func (c *grpcStorageClient) ComposeObject(ctx context.Context, req *composeObjec
 		Destination:   dstObjPb,
 		SourceObjects: srcs,
 	}
+	if err := applyCondsProto("ComposeObject destination", defaultGen, req.dstObject.conds, rawReq); err != nil {
```
BrennaEpp, line 811: What was the reason for moving this?

BrennaEpp (approved):

PR #8404, opened 2023-08-09
test(storage): add bucket conditions test

tritone:
Add an emulator test for bucket conditions (similar to the object one added previously). This covers all bucket methods which take preconditions and did not catch any bugs.

Also minor formatting cleanup to the object test.


storage/client_test.go
```diff
@@ -1134,7 +1194,7 @@ func TestObjectConditionsEmulated(t *testing.T) {
 			call func() error
 		}{
 			{
-				name: "Object update generation",
```
BrennaEpp, line 1137: Nice

storage/client_test.go
```diff
@@ -1120,6 +1120,66 @@ func TestHMACKeyCRUDEmulated(t *testing.T) {
...
+			name string
+			call func() error
+		}{
+			{
+				name: "get",
+				call: func() error {
+					bucket, metaGen, err := createBucket(ctx, project)
+					if err != nil {
+						return fmt.Errorf("creating bucket: %w", err)
+					}
```
BrennaEpp, line 1136: Can this be refactored out into t.Run and passed into the call?

BrennaEpp (approved):
LGTM, just a comment on minor refactoring

storage/client_test.go
tritone: sure, done

PR #8409, opened 2023-08-11
fix(firestore): Compare full resource path when docs ordered by __name__

bhshkh:
**Issue**: When a partitioned query is serialized, deserialized and then checked for changes in its snapshot, panic is seen
`panic: runtime error: slice bounds out of range [:-1]` at [this](https://github.com/googleapis/google-cloud-go/blob/main/firestore/document.go#L155) line 

Sample code:
```
partitionedQueries, _ := client.CollectionGroup("landmarks").GetPartitionedQueries(ctx, 10)
protoBytes, _ := partitionedQueries[0].Serialize()
query, _ := client.CollectionGroup("").Deserialize(protoBytes)
	
it := query.Snapshots(ctx)
querySnapshot, _ := it.Next()
for _, docChange := range querySnapshot.Changes {     <========= querySnapshot.Changes causes panic
	fmt.Printf("%+v\n", docChange.Doc.Data())
}
```

**Cause**: 
The document changes in query snapshot are ordered by whatever order is specified in query. When fieldPath field in order field of query is \_\_name\_\_, then changes are ordered by full resource path i.e. document changed at path /cols/doc1 is returned before /cols/doc2.
The same should happen when field_path field in fieldReference field is \_\_name\_\_ but it does not.

Before serializing, the order in query is 
```
{fieldPath:[__name__] fieldReference:<nil> dir:1}
```
After deserializing, the order in query is
```
{fieldPath:[] fieldReference:{field_path:"__name__" } dir:1}
```
Currently, when fieldPath is empty, the library assumes that field_path within fieldReference is an actual field in the document and tries to get the value of that field but since there is no field called \_\_name\_\_ within the document, the code panics

**Fix**: Correct the condition for checking whether document is ordered by full resource path


enocom (approved):

firestore/integration_test.go
```diff
@@ -1743,6 +1743,72 @@ func TestIntegration_FieldTransforms_Set(t *testing.T) {
...
+	shouldBeFoundID := collectionIDs.New()
+
+	// Create documents for partiontioned query result
+	ctx := context.Background()
+	h := testHelper{t}
+	client := integrationClient(t)
+	cr1 := client.Collection(shouldBeFoundID)
+	dr1 := cr1.Doc("should-be-found-1")
+	h.mustCreate(dr1, map[string]string{"some-key": "should-be-found"})
+	defer h.mustDelete(dr1)
```
danieljbruce, line 1756: Would it be a good idea to pull the above three lines of code into a function and return the `mustGet` function? It might shorten the test a little bit.

firestore/integration_test.go
```diff
@@ -1743,6 +1743,72 @@ func TestIntegration_FieldTransforms_Set(t *testing.T) {
...
+
+	partitionedQueries, err := client.CollectionGroup(shouldBeFoundID).GetPartitionedQueries(ctx, 10)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	qProtoBytes, err := partitionedQueries[0].Serialize()
+	if err != nil {
+		t.Fatal(err)
+	}
```
danieljbruce, line 1771: Can blocks like this be pulled into a function?

danieljbruce (commented):
A couple of comments about some refactors which may be useful.

firestore/integration_test.go
bhshkh: Refactored the code

firestore/integration_test.go
bhshkh: Done

danieljbruce (approved):

PR #8411, opened 2023-08-12
chore(storage): capture memory stats at set interval [benchmarks]

BrennaEpp:
- Removes memory collection before/after benchmark
- On a ticker set to `memory_interval`, memory stats will be recorded and output to stdout or the specified `memory_file`
- default memory interval is 5 hours, which will avoid memory collection for existing workloads unless specifically set
- Each stat is sent to a different metric for Cloud Monitoring


storage/internal/benchmarks/memory_metric.go
```diff
@@ -0,0 +1,185 @@
...
+	var memStats *runtime.MemStats = &runtime.MemStats{}
+	done := make(chan bool)
+
+	if oType == outputCSV {
+		writeMemoryHeader(w)
+	}
+
+	go (func() {
+		ticker := time.NewTicker(interval)
+
```
BrennaEpp, line 40: Should `defer ticker.Close()`

storage/internal/benchmarks/main.go
```diff
@@ -286,6 +290,22 @@ func main() {
...
+			memOutputFile = "mem.csv"
+		}
+		f, err := os.Create(memOutputFile)
+		if err != nil {
+			log.Fatalf("Failed to create file %s: %v", memOutputFile, err)
+		}
+		defer f.Close()
+		memW = f
+	}
+	cancelMemoryRecording := startRecordingMemory(memW, opts.outType, opts.memStatInterval)
```
tritone, line 307: Will this record during the warmup as well, or start afterwards?

storage/internal/benchmarks/memory_metric.go
```diff
@@ -0,0 +1,185 @@
...
+// returned function is called.
+func startRecordingMemory(w io.Writer, oType outputType, interval time.Duration) (cancel func()) {
+	var memStats *runtime.MemStats = &runtime.MemStats{}
+	done := make(chan bool)
+
+	if oType == outputCSV {
+		writeMemoryHeader(w)
+	}
+
+	go (func() {
```
tritone, line 38: The extra paren enclosing the func is unnecessary right?

storage/internal/benchmarks/memory_metric.go
```diff
@@ -0,0 +1,185 @@
...
+					writeMemStatsAsCSV(w, *memStats)
+				} else if oType == outputCloudMonitoring {
+					writeMemStatsAsCloudMonitoring(w, *memStats)
+				}
+			}
+		}
+	})()
+
+	return func() { done <- true }
+}
```
tritone, line 58: Nice, this is cleverly written.

storage/internal/benchmarks/memory_metric.go
```diff
@@ -0,0 +1,185 @@
...
+
+	go (func() {
+		ticker := time.NewTicker(interval)
+
+		for {
+			select {
+			case <-done:
+				return
+			case <-ticker.C:
+				runtime.ReadMemStats(memStats)
```
tritone, line 46: Do you have any sense of the overhead of doing this? Curious whether it would actually affect results at all if you are calling this frequently.

tritone (commented):
Few small questions, generally LGTM.

Also, can you describe how you tested this out? Wondering if it is time to add some simple integration tests for the benchmarking script soon...

PR #8418, opened 2023-08-15
chore(storage): add warmup option [benchmarks]

BrennaEpp:
Extra telemetry set up to test SSB - will revert those changes.

This runs w1r3 benchmarks (discarding output) to warm up.

It does write/read from disk to warmup - we can change that in the future but this should do for now


storage/internal/benchmarks/warmup.go
```diff
@@ -0,0 +1,72 @@
...
+	defer cancel()
+
+	discardBenchmarkResults(ctx)
+
+	warmupGroup, ctx := errgroup.WithContext(ctx)
+	warmupGroup.SetLimit(opts.numWorkers)
+
+	for deadline := time.Now().Add(opts.warmup); time.Now().Before(deadline); {
+		warmupGroup.Go(func() error {
+			benchmark := &w1r3{opts: opts, bucketName: opts.bucket}
```
tritone, line 36: So, I guess this uses the same level of parallelism as was specified for the main workload? I am wondering if it makes more sense to have a fixed/higher level of parallelism, or to be able to specify this separately.

tritone (approved):
Thanks for adding this, just 1 question.

storage/internal/benchmarks/warmup.go
BrennaEpp: It does; I am open to changing it to a fixed parallelism level - say 16? I think being able to specify this separately may create unnecessary confusion - unless we determine it would be beneficial to have different levels of parallelism for warmups for different workloads, in which case we can always add this in.

The way it is now gives consistency in terms of bucket traffic: the bucket always sees the same level of parallelism; but ultimately for the long term a fixed parallelism makes sense.

storage/internal/benchmarks/directory_benchmark.go
```diff
@@ -172,6 +175,13 @@ func (r *directoryBenchmark) cleanup() error {
 }
 
 func (r *directoryBenchmark) uploadDirectory(ctx context.Context, numWorkers int) (elapsedTime time.Duration, err error) {
+	var span trace.Span
```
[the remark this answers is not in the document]
BrennaEpp: This is done now.

storage/internal/benchmarks/warmup.go
tritone: Yeah I think 16 (or numCPUs) is appropriate. We could make it configurable if/when we refactor all the CLI opts into some kind of config file eventually...

BrennaEpp: Good idea - I changed it to num cpus

PR #8419, opened 2023-08-15
docs(bigquery): improve RowIterator docs and out of process pagination with Storage API

### LINKED ISSUE #7285: bigquery: doc of RowIterator.TotalRows seems to be wrong.
[Doc](https://github.com/googleapis/google-cloud-go/blob/main/bigquery/iterator.go#L58-L60) says `Available after the first call to Next.` but, I can retrieve TotalRows anytime.
This problem seem to be resolved before https://github.com/googleapis/google-cloud-go/issues/2946

### LINKED ISSUE #8173: bigquery: add documentation for out of process pagination
It's pretty standard to want to paginate through results, and while it is simple to iterate through a full query, there is minimal documentation about how to hydrate a `RowIterator` with a page token across separate RPCs. All the examples I can find are in process only. I was expecting there to be a convenience method like `RowIteratorFromPageToken()`, but there isn't, and there doesn't seem to be a way to populate an iterator manually. Best I can tell, you're required to keep track of job and manually kick things off. Best I can tell, this is how you have to paginate. Is this correct? If so, would you accept a PR for an example?

```go
func paginateQuery(ctx context.Context, bqClient *bigquery.Client, jobID, pageToken string, pageSize int) ([]int, string, string, error) {
	var (
		iter *bigquery.RowIterator
		job  *bigquery.Job
		err  error
	)

	// if we have a page token, we need to get a job handle from the page token,
	// otherwise execute the query to initiate the job
	if pageToken == "" {
		q := bqClient.Query("SELECT num FROM UNNEST(GENERATE_ARRAY(1, 25)) num")
		iter, err = q.Read(ctx)
		if err != nil {
			return nil, "", "", err
		}

		job = iter.SourceJob()
	} else {
		// get a job handle from the job ID
		job, err = bqClient.JobFromID(ctx, jobID)
		if err != nil {
			return nil, "", "", err
		}

		iter, err = job.Read(ctx)
		if err != nil {
			return nil, "", "", err
		}
	}

	// set the page token and size
	iter.PageInfo().Token = pageToken
	iter.PageInfo().MaxSize = pageSize

	var numbers []int
	for {
		var num int
		err = iter.Next(&num)
		
		// if we're done, return the numbers, job ID, and empty page token
		if err == iterator.Done {
			return numbers, job.ID(), "", nil
		}
		if err != nil {
			return nil, "", "", err
		}

		numbers = append(numbers, num)

		// if we have a full page, return the job ID and page token
		if len(numbers) == pageSize {
			// if we're only partially through a pageToken, how do we know we're starting back in the right place?
			return numbers, job.ID(), iter.PageInfo().Token, nil
		}
	}
}
```

There were a few other references in issues here that were similarly confused:
- https://github.com/googleapis/google-cloud-go/issues/2694#issuecomment-673275465
- https://github.com/googleapis/google-cloud-go/issues/7902

### LINKED ISSUE #8373: bigquery: StorageReadClient doesn't use page token
**Client**

BigQuery

**Environment**

Distroless on GKE

**Go Environment**

go 1.20.4

**Code**

e.g.
```go
package main

func main() {
	bqClientWithStorage, err := bigquery.NewClient(c, d.Getenv("GCP_PROJECT_ID"))

	err = bqClientWithStorage.EnableStorageReadClient(c)

	job, err := bqClient.JobFromProject(c, "project", "job", "US")

	// get a row iterator from the job
	iter, err := p.Job.Read(c)
	if err != nil {
		return nil, e.Wrap(err, e.With("pageToken", p))
	}

	// set the page token and size on the iterator
	iter.PageInfo().Token = token
	iter.PageInfo().MaxSize = pageSize

	for i := 0; ; i++ {
		var cl bqCluster
		err = iter.Next(&cl)
		...

		if pageSize = i {
			// store the next page token for the next loop (not necessarily in the same process)
			fmt.Println(iter.PageInfo().Token)

			break
		}
	}
}
```

**Expected behavior**

I can use a page token to iterate through rows

**Actual behavior**

No page token is set

**Additional context**

Iteration works with out of process page tokens using the normal BigQuery client, but fails after enabling the storage read client.

I don't see anywhere other than the first initialization that references the token or page size in the related functions, which seems to confirm that they aren't supported.
https://github.com/googleapis/google-cloud-go/blob/main/bigquery/storage_iterator.go#L115-L233

**Related issues**
- #8173

alvarowolfx:
Fixes #7285, fixes #8373, fixes #8173


shollyman (approved):
Do we selectively accelerate some iterators and not others with the enhanced storage integration?  If so, may be worth calling out how to check if the iterator is backed by the read API.

alvarowolfx:
> Do we selectively accelerate some iterators and not others with the enhanced storage integration? If so, may be worth calling out how to check if the iterator is backed by the read API.

I added a reference to the `IsAccelerated` method on the `RowIterator` to check if the given iterator is being backend by Storage API.

PR #8424, opened 2023-08-16
chore(video/stitcher): add config to restore generation of apiv1

quartzmo:
refs: #7659


noahdietz (commented):
Let's wait a moment, they have a breaking change (that might not affect Go) pending b/289556132

PR #8307, opened 2023-07-20
feat(datastore): SUM and AVG aggregations

bhshkh:

PR #8439, opened 2023-08-18
feat(datastore): Support aggregation query in transaction

bhshkh:
Currently, the Go Datastore client library does not support running aggregation queries in a transaction. Below code returns error `datastore: cannot use EventualConsistency query in a transaction`

```
_, runErr := client.RunInTransaction(ctx, func(tx *datastore.Transaction) error {
        aggQuery := datastore.NewQuery("City").Transaction(tx).NewAggregationQuery().WithCount("")
        gotAggResult, gotErr := client.RunAggregationQuery(ctx, aggQuery)
        if gotErr != nil {
            fmt.Printf("client.RunAggregationQuery: %v\n", gotErr)
            return gotErr
        }
        fmt.Printf("gotAggResult: %v\n", gotAggResult)

        return nil
    })
```

It is because of this line: https://github.com/googleapis/google-cloud-go/blob/datastore/v1.13.0/datastore/query.go#L1029

In this PR, I have removed this line


datastore/integration_test.go
```diff
@@ -729,30 +730,57 @@ func TestIntegration_AggregationQueries(t *testing.T) {
...
 
-	baseQuery := NewQuery("SQChild").Ancestor(parent)
+	// Create transaction with read before create
+	txBeforeCreate, err := client.NewTransaction(ctx, []TransactionOption{ReadOnly, WithReadTime(beforeCreate)}...)
+	if err != nil {
+		t.Fatalf("client.NewTransaction: %v", err)
+	}
+
+	// Create transaction with read after create
+	txAfterCreate, err := client.NewTransaction(ctx, []TransactionOption{ReadOnly, WithReadTime(time.Now())}...)
```
enocom, line 740: Is it interesting or useful to make read time something other than `time.Now()`? Should we rename `beforeCreate` to `readTime` and use that again here?

bhshkh: Renamed to readTime but still  using time.Now() since I want to get the time after the entities have been created and check for their presence in the result

noahdietz (approved):

enocom (approved):

kolea2 (approved):

kolea2:
LGTM, once https://github.com/googleapis/google-cloud-go/pull/8307 is merged let's add another quick test for SUM/AVG transaction queries

codyoss (approved):

bhshkh:
> LGTM, once #8307 is merged let's add another quick test for SUM/AVG transaction queries

Added

enocom (approved):

PR #8443, opened 2023-08-18
test(pubsub): remove brittle custom retry test

hongalex:
From #8166, these kinds of tests are already kind of brittle. Changing the retry settings also broke this test in #8438 so I think the best thing to do is just remove this. The retry config functionality should be tested at the GAPIC generator level instead of the veneer anyway.

### DIFF
--- a/pubsub/pubsub_test.go
+++ b/pubsub/pubsub_test.go
@@ -16,7 +16,6 @@ package pubsub
 
 import (
 	"context"
-	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


quartzmo:
> The retry config functionality should be tested at the GAPIC generator level instead of the veneer anyway.

Do we need to open an issue for this?

quartzmo (approved):

hongalex:
> > The retry config functionality should be tested at the GAPIC generator level instead of the veneer anyway.
> 
> 
> 
> Do we need to open an issue for this?

I suspect this should already be part of the generator code so I don't think so?

PR #8446, opened 2023-08-18
chore(ci): add new_client workflow

noahdietz:
This uses `changefinder` to find the newly added `version.go` files, but only those that are GAPIC generated (thats the `content-pattern` flag), then bumps the CHANGES.md for each module in separate PRs

only on `push` to `main` with `version.go` changes


quartzmo (approved):
Thank you!

PR #8460, opened 2023-08-22
fix(storage): use greater precision for CustomTime

### LINKED ISSUE #8422: storage: HTTP client strips milliseconds from `Custom-Time`
**Client**

Storage

**Environment**

Any

**Go Environment**

Go 1.21

**Code**

**Expected behavior**

Milliseconds is supported as per https://cloud.google.com/storage/docs/metadata#custom-time and should not be stripped. 

**Actual behavior**

Milliseconds are stripped due to using `time.RFC3339`.

https://github.com/googleapis/google-cloud-go/blob/fbe78a28e01043b059e86c6e960d092b299ac9a4/storage/storage.go#L1090-L1093

**Screenshots**

image

**Additional context**

gRPC client looks correct:

https://github.com/googleapis/google-cloud-go/blob/fbe78a28e01043b059e86c6e960d092b299ac9a4/storage/storage.go#L1135

https://github.com/googleapis/google-cloud-go/blob/fbe78a28e01043b059e86c6e960d092b299ac9a4/storage/storage.go#L1364-L1370

tritone:
Storage supports CustomTime to millisecond precision so we should use additional precision if possible.

Fixes #8422


storage/storage_test.go
```diff
@@ -1747,7 +1747,7 @@ func TestRawObjectToObjectAttrs(t *testing.T) {
 				Bucket:                  "Test",
 				ContentLanguage:         "en-us",
 				ContentType:             "video/mpeg",
-				CustomTime:              "2020-08-25T19:33:36Z",
+				CustomTime:              "2020-08-25T19:33:36.05Z",
```
BrennaEpp, line 1750: Can you add CustomTime to the `TestObjectAttrsToRawObject` test?

BrennaEpp (commented):
Small suggestion: can we ensure we are also testing that times without milliseconds return a string in the format `YYYY-MM-DD'T'HH:MM:SS'Z'` (no milliseconds)?

codyoss:
@tritone do we still want these changes?

maros7:
We still want (need) this change.

codyoss:
@tritone see my comment above

cpriti-os:
Closing this: too old PR

Issue #8192, opened 2023-06-30
spanner: Intermittent error "Transaction was started in a different session"


lc-caigwatkin:
**Client**

Spanner

**Environment**

Cloud Run

**Go Environment**

go version go1.19.5, go1.20.1 to go1.20.4

**Code**

Seems to not be isolated to any particular piece of our code.

Here is an example of `cloud.google.com/go/spanner` related stack trace when we do see the error:

```
cloud.google.com/go/spanner.(*ReadWriteTransaction).runInTransaction
	/root/project/vendor/cloud.google.com/go/spanner/transaction.go:1505
cloud.google.com/go/spanner.(*Client).rwTransaction.func2
	/root/project/vendor/cloud.google.com/go/spanner/client.go:557
cloud.google.com/go/spanner.runWithRetryOnAbortedOrFailedInlineBeginOrSessionNotFound.func1
	/root/project/vendor/cloud.google.com/go/spanner/retry.go:92
cloud.google.com/go/spanner.runWithRetryOnAbortedOrFailedInlineBeginOrSessionNotFound
	/root/project/vendor/cloud.google.com/go/spanner/retry.go:132
cloud.google.com/go/spanner.(*Client).rwTransaction
	/root/project/vendor/cloud.google.com/go/spanner/client.go:522
cloud.google.com/go/spanner.(*Client).ReadWriteTransaction
	/root/project/vendor/cloud.google.com/go/spanner/client.go:490
```

**Expected behavior**

Transaction executes without error OR documentation includes information to understand and resolve this error.

**Actual behavior**

The following error occasionally pops up in our production environment across multiple databases/microservices in which query logic and database schema barely change.

`spanner: code = "InvalidArgument", desc = "Transaction was started in a different session."`

**Screenshots**

In our production environment, last 30 days, showing our error or warning logs when saw that error:

<img width="1307" alt="image" src="https://github.com/googleapis/google-cloud-go/assets/34219518/3a06366a-0971-49f9-9a6c-f8dc5331e7fe">

**Additional context**

We started seeing this error "Transaction was started in a different session" pop up occasionally in our production environment after upgrading to v1.44.0 in February 2023 (the first time this happened was 2023-02-12).
This appears to happen after the transaction automatically retries.
It appears to occur in our Spanner databases with higher contention, although our database usage has been higher in the past without this particular error occurring.

We cannot find any information about this new error to help us understand if this is something we can fix.
[...]

PR #8462, opened 2023-08-22
fix(spanner): do explicit begin transaction if session not found error happend

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/8192


spanner/client.go
```diff
@@ -555,6 +555,10 @@ func (c *Client) rwTransaction(ctx context.Context, f func(context.Context, *Rea
 			err error
 		)
 		if sh == nil || sh.getID() == "" || sh.getClient() == nil {
+			//  retry with explicit begin transaction
+			if t != nil {
```
olavloite, line 559: Can we have (a) test(s) that show:
1. What would have gone wrong before without this change (i.e. a test that fails without this change).
2. What is the behavior for the scenario that this fix solves (i.e. a test that succeeds even if a `Session not found` error occurs)

It might very well be possible that the two can be combined into one test.

PR #8467, opened 2023-08-23
fix(spanner): Transaction was started in a different session

olavloite:
Retrying a "Session not found" error could cause a "Transaction was started in a different session" error. This happened because:
1. The detection of a "Session not found" error would remove the session from the pool, and also remove the session ID from the session handle
2. The retry mechanism would check out a new session from the pool, but not assign it to the transaction yet
3. The retry would then proceed to retry the transaction with an explicit BeginTransaction RPC. This function would however pick a new session from the pool, because step 2 had not yet assigned the transaction a new session.
4. The higher level retry loop would then after executing the BeginTransaction RPC assign the session that was picked in step 2 to the transaction.
5. The transaction would then proceed to use the session from step 2 with the transaction from step 3.

Fixes: https://github.com/googleapis/google-cloud-go/issues/8192


spanner/client.go
```diff
@@ -571,9 +575,9 @@ func (c *Client) rwTransaction(ctx context.Context, f func(context.Context, *Rea
 			t = &ReadWriteTransaction{
 				txReadyOrClosed: make(chan struct{}),
 			}
+			t.txReadOnly.sh = sh
 		}
 		attempt++
-		t.txReadOnly.sh = sh
```
olavloite, line 576: This assignment would override the assignment that happened at the end of `t.begin(...)`.

rahul2393 (approved):

PR #8476, opened 2023-08-23
feat(bigquery): set PreferredMinStreamCount when creating read session

### LINKED ISSUE #8432: bigquery: set `PreferredMinStreamCount`while creating `CreateReadSessionRequest`
While creating `CreateReadSessionRequest` we are setting `MaxStreamCount` as 0 which as I understand will leave the decision on server to choose the correct stream count. 
On my testing on various public and private data sources I always found the number of stream to be 1.
I would like to set `PreferredMinStreamCount` as `readClientSettings.maxWorkerCount` as hint to the server for desired minimum streams. 

I would like to set `PreferredMinStreamCount` as `readClientSettings.maxWorkerCount` as hint to the server for desired minimum streams. 

Since there is no option in sdk that allow user to configure this there are no other alternatives.

**Additional context**
https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#createreadsessionrequest

alvarowolfx:
`PreferredMinStreamCount` must be less than or equal to `MaxStreamCount`, so we only set it when the `MaxStreamCount` is 0, which basically sets no limit to it.

Resolves #8432


bigquery/storage_client.go
```diff
@@ -135,13 +135,19 @@ type readSession struct {
 
 // Start initiates a read session
 func (rs *readSession) start() error {
+	preferredMinStreamCount := int32(0)
```
shollyman, line 138: this looks a bit odd, why not just `var preferredMinStreamCount int32`?

shollyman (approved):
minor nit, but otherwise LGTM

bigquery/storage_client.go
alvarowolfx: good point, let me change here. I think I was playing with default values and ended up leaving like that.

PR #8486, opened 2023-08-24
test(storage): update retry conformance tests for dual transports

cojenco:
Support grpc for retry conformance tests to run under dual transports

- send extra field `transport` when creating a new retry test resource
- skip test cases for 501 Unimplemented
- switch to use NewGRPCClient or newHTTPStorageClient
- attach retry test id by SetHeaders per test method call
- remove wrappedClient and RounTrip that are no longer needed


cojenco:
Pending changes in https://github.com/googleapis/storage-testbench/pull/526

storage/retry_conformance_test.go
```diff
@@ -683,12 +698,25 @@ func (et *emulatorTest) create(instructions map[string][]string) {
...
-	// Create wrapped client which will send emulator instructions
 	et.host.Path = ""
-	client, err := wrappedClient(et.T, et.id)
-	if err != nil {
-		et.Fatalf("creating wrapped client: %v", err)
+
+	// Create wrapped client which will send a retry test id via a constant header
+	var client *Client
+	if transport == "json" {
+		client, err = wrappedClient(et.T, et.id)
```
tritone, line 706: I think we can actually rip out this whole custom client logic now that we added support for sending custom headers via the context: https://pkg.go.dev/cloud.google.com/go/storage#hdr-Sending_Custom_Headers . Works for both gRPC and JSON. The context should be applied per method call, not per client.

storage/retry_conformance_test.go
```diff
@@ -683,12 +698,25 @@ func (et *emulatorTest) create(instructions map[string][]string) {
...
+			et.Fatalf("creating wrapped client: %v", err)
+		}
+	} else if transport == "grpc" {
+		opts := defaultGRPCOptions()
+		header := metadata.New(map[string]string{"x-retry-test-id": et.id})
+		ctx := metadata.NewOutgoingContext(context.Background(), header)
+		var clientMu sync.Mutex
+		clientMu.Lock()
+		os.Setenv("STORAGE_USE_GRPC", "true")
+		client, _ = NewClient(ctx, opts...)
```
tritone, line 717: We can use the new exported constructor after #8509 is merged. This will eliminate the need for the env var and mutex.

storage/retry_conformance_test.go
```diff
@@ -666,6 +678,9 @@ func (et *emulatorTest) create(instructions map[string][]string) {
 
 	et.host.Path = "retry_test"
 	resp, err := c.Post(et.host.String(), "application/json", buf)
+	if resp.StatusCode == 501 {
```
tritone, line 681: Why is this needed?

storage/retry_conformance_test.go
```diff
@@ -521,34 +523,42 @@ func TestRetryConformance(t *testing.T) {
...
-						testName := fmt.Sprintf("%v-%v-%v-%v", retryTest.Id, instructions.Instructions, methodName, i)
-						t.Run(testName, func(t *testing.T) {
-
-							// Create the retry subtest
-							subtest := &emulatorTest{T: t, name: testName, host: endpoint}
-							subtest.create(map[string][]string{
-								method.Name: instructions.Instructions,
+						transports := []string{"json", "grpc"}
+						for _, transport := range transports {
+							testName := fmt.Sprintf("%v-%v-%v-%v/%v", retryTest.Id, instructions.Instructions, methodName, i, transport)
```
tritone, line 528: Maybe start with the transport? Seems most relevant to filter by

tritone (commented):
Couple small things that have to do with client changes mostly, overall looks good.

storage/retry_conformance_test.go
cojenco: Sounds good, thanks! I'll go ahead and merge the testbench implementation first since the proposed changes in this PR seems feasible; then I'll come back and finalize this :)

storage/retry_conformance_test.go
cojenco: Awesome, thanks! I removed the wrappedClient and RoundTrip per your comment. With the latest custom headers support, we're now attaching the retry test id per test method call.

storage/retry_conformance_test.go
cojenco: updated

tritone (approved):
Nice work on this!

BrennaEpp (approved):
Thank you!!
Is this ready to merge?

cojenco:
Merging this in; Temporarily skipping bidi writes tests for now as testbench implementation is currently work in progress

PR #8487, opened 2023-08-24
test(datastore): Retry aggregation query testcase on failure

bhshkh:
**Issue**: Kokoro build fails with error
```
integration_test.go:886: "Aggregations in transaction after creating entities": Mismatch in aggregation result got: map[avg:null_value:NULL_VALUE count:integer_value:0 sum:integer_value:0], want: map[avg:double_value:3.5 count:integer_value:8 sum:integer_value:28]
```
**Cause**: Entities are read too soon after creating and thus are not reflected in the aggregation results.
**Fix**: Add explicit sleep before running aggregation queries


datastore/integration_test.go
```diff
@@ -741,6 +741,7 @@ func TestIntegration_AggregationQueries(t *testing.T) {
 		}
 	}()
 
+	time.Sleep(10 * time.Second)
```
enocom, line 744: Rather than add a sleep here, could we retry the assertions below with a helper function?

The retries could run the query, check the result, if it's no good, try again after waiting 1 second, and do that 10 times. So you'd at worse get an additional 10 second runtime, but probably complete the test sooner.

enocom (approved):

kolea2 (approved):
lgtm after nit

Issue #8100, opened 2023-06-15
bigquery: Expose Apache Arrow data


zeroshade:
**Is your feature request related to a problem? Please describe.**
Currently even though Apache Arrow is used internally by the BigQuery client, the Arrow iterator and Arrow data is not, itself, exposed externally by the client. There is no way to retrieve the data in the columnar format which imposes a performance cost of having to transpose the data to get it row-by-row.

**Describe the solution you'd like**
Ideally the Arrow IPC streams could be exposed directly to allow consumers to be able to retrieve the data without needing to be using the same major version of the Go Arrow libraries. Potentially, if available, exposing the ability to parallelize fetching if possible (based on multiple partitions being available etc.) At a minimum though, exposing an interface that meets the [RecordReader](https://pkg.go.dev/github.com/apache/arrow/go/v12@v12.0.1/arrow/array#RecordReader) interface would be very useful.

**Describe alternatives you've considered**
The only alternative I can think of would be to bypass this library entirely and attempt to implement fetching from BigQuery's API directly which would be inconvenient and unmaintainable ultimately.

**Additional context**
I'm looking to build a driver for BigQuery for [Arrow Database Connectivity (ADBC)](https://github.com/apache/arrow-adbc) and for ease of deployment, was going to build it using Go (as we already have done for Arrow Flight SQL and Snowflake ADBC drivers). At conferences where I've spoken about ADBC, the 2nd biggest request has been for a BigQuery driver, so I'd like to start looking into this. I thought it would be easy since I saw that BigQuery already uses Arrow, but I couldn't find anything in the pkg.go.dev docs on how to actually retrieve the Arrow data directly. If I'm wrong, then I'd happily take a link to any documentation on how to retrieve the arrow data (and even better, write Arrow data BigQuery!).

Thank you very much!

alvarowolfx:
thanks for reaching out @zeroshade about exposing Arrow and support for ADBC. I was the person working on integration with our internal Storage Read API using Arrow in our SDK and got to say that multiple paths were considered for doing so. One of those paths included giving access to Arrow records to users, but as the surface area was already becoming too big, we ended deciding to remove that from the initial release, but some foundation for that work it there under the hood.

In regards to your point on writing Arrow records, unfortunately our Storage Write doesn't have support for Arrow. For the time being it supports writing Protobufs.

[...]

PR #8500, opened 2023-08-28
feat(bigquery): expose Apache Arrow data

k-anshul:
#8100

This PR exposes an iterator to iterate over serialised apache arrow records.


bigquery/iterator.go
```diff
@@ -194,6 +194,31 @@ func (it *RowIterator) fetch(pageSize int, pageToken string) (string, error) {
...
+
+// NextRecord returns next batch of rows as serialised arrow.Record
+func (a *ArrowIterator) NextRecord() ([]byte, error) {
+	return a.r.arrowIterator.next()
+}
+
+// Schema is available after first call to NextRecord
+func (a *ArrowIterator) Schema() []byte {
+	return a.r.arrowIterator.decoder.rawArrowSchema
+}
```
zeroshade, line 212: It might be more interoperable with the Arrow libraries if you could expose a direct `io.Reader` for the raw IPC stream

bigquery/query.go
```diff
@@ -366,6 +366,27 @@ func (q *Query) newJob() (*bq.Job, error) {
...
+
+	q.forceStorageAPI = true
+	rowIter, err := q.Read(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if rowIter.arrowIterator == nil {
+		return nil, errors.New("bigquery: results not available as arrow records")
+	}
```
zeroshade, line 385: is there a particular kind of query / results that can't be returned as an arrow record batch stream? Can we document those?

k-anshul: As I understood, this is not possible right now since we enforce the usage of `storageAPI`. I added this as a defensive check. I can remove this if this is adding confusion.

bigquery/iterator.go
k-anshul: Actually I am using this to construct a `array.RecordReader`. Sample here : https://github.com/rilldata/rill/blob/09726146ec04415e68f7e41efa64e5b5db12aa7e/runtime/drivers/bigquery/arrow.go#L19
I can move this code to the SDK and make `ArrowIterator` confirm to `array.RecordReader` as proposed in other comments as well. Is the `array.RecordReader` locked for further modifications ?

It seemed tricky to expose a `io.Reader` that can read from multiple IPC streams.
Also I was looking at the spec for `io.Reader` and saw this statement : `If some data is available but not len(p) bytes, Read conventionally returns what is available instead of waiting for more.`
I am wondering what should be the behaviour here? Should we wait for next API call to succeed or return whatever is available ?

alvarowolfx (commented):
Can some integration tests be added to show how the ArrowIterator interface is being used ?

bigquery/query.go
```diff
@@ -366,6 +366,27 @@ func (q *Query) newJob() (*bq.Job, error) {
 	}, nil
 }
 
+// ReadAsArrowObjects submits a query for execution and returns the results via an ArrowIterator.
+// As a prerequisite storage read client should be enabled.
+// It will always use jobs.insert path so may not be efficient for small queries.
+func (q *Query) ReadAsArrowObjects(ctx context.Context) (it *ArrowIterator, err error) {
```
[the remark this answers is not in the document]
k-anshul: I will add this for those paths as well.

bigquery/iterator.go
zeroshade: The difficulty with having `ArrowIterator` conform to `array.RecordReader` is that it would require explicitly exposing the major versioned `arrow.Record` interface which I believe was trying to be avoided. If it's okay to directly include the versioned `arrow.Record` interface in the API then I would agree that conforming to the `array.RecordReader` interface would be best.

I agree you wouldn't expose an `io.Reader` to read from multiple IPC streams, but rather you'd expose an individual `io.Reader` for each IPC stream, and then allow the consumer to construct an `array.RecordReader` for each stream. That's the common way of doing it to allow the consumer to control the backpressure etc. It's also how I implemented the Snowflake arrow stream handling, it provides an `io.Reader` *per* stream.

bigquery/query.go
zeroshade: @alvarowolfx As long as you have a schema, you can use https://pkg.go.dev/github.com/apache/arrow/go/v13/arrow/array#NewJSONReader or the individual `FromJSON` functions etc.

In addition, all of the `Builder`s have an `UnmarshalJSON` method which will append the unmarshalled json to the builder.

bigquery/iterator.go
k-anshul: Thanks for the inputs @zeroshade 
Should the `io.Reader` implementation wait for the next batch to be available or return whatever bytes are available as per the `io.Reader` specs?

zeroshade: it should return whatever bytes are available as per the `io.Reader` spec. Just make sure you don't return an `io.EOF` unless you *actually* are at the end :smile:

PR #8506, opened 2023-08-29
feat(bigquery): expose Apache Arrow data through ArrowIterator 

### LINKED ISSUE #8100: bigquery: Expose Apache Arrow data

alvarowolfx:
As we have some planned work to support Arrow data fetching on other query APIs, so we need to think of an interface that will support all of those query paths and also work as a base for other Arrow projects like ADBC. So this PR detaches the Storage API from the Arrow Decoder and creates a new ArrowIterator interface. This new interface is implemented by the Storage iterator and later can be implemented for other query interfaces that supports Arrow.

Resolves #8100


bigquery/storage_integration_test.go
```diff
@@ -418,11 +425,112 @@ func TestIntegration_StorageReadCancel(t *testing.T) {
...
 	// resources are cleaned asynchronously
 	time.Sleep(time.Second)
-	if !it.arrowIterator.isDone() {
+	arrowIt := it.arrowIterator.(*storageArrowIterator)
+	if !arrowIt.isDone() {
 		t.Fatal("expected stream to be done")
 	}
 }
 
+func TestIntegration_StorageReadArrow(t *testing.T) {
```
alvarowolfx, line 434: @k-anshul @zeroshade this integration test show an example on how that interface would be used.

zeroshade:
@alvarowolfx I'm gonna try to get a thorough review of this in the next day or two from the Arrow perspective. Thanks for doing this!

bigquery/storage_integration_test.go
```diff
@@ -418,11 +425,112 @@ func TestIntegration_StorageReadCancel(t *testing.T) {
...
+	}
+	arrowIt, err := it.ArrowIterator()
+	if err != nil {
+		t.Fatalf("expected iterator to be accelerated: %v", err)
+	}
+	var arrowSchema *arrow.Schema
+	records := []arrow.Record{}
+	r, err := ipc.NewReader(&arrowIteratorReader{
+		it: arrowIt,
+	})
```
zeroshade, line 459: can we have an interface that doesn't require consumers to implement their own `arrowIteratorReader`?

bigquery/storage_iterator.go
```diff
@@ -191,20 +201,6 @@ func (it *arrowIterator) init() error {
...
-
-	decoder, err := newArrowDecoderFromSession(it.session, it.schema)
-	if err != nil {
-		return err
-	}
-	it.decoder = decoder
-
 	wg := sync.WaitGroup{}
 	wg.Add(len(streams))
 	sem := semaphore.NewWeighted(int64(it.session.settings.maxWorkerCount))
```
zeroshade, line 206: would it be feasible to expose the streams themselves to a consumer to allow them to control the parallelization instead of forcing them to a specific route like this?

zeroshade:
The new `ArrowIteratorReader` looks good to me, so this seems pretty good from my perspective, though it would be nice to be able to expose the raw streams to the consumer directly allowing them to control parallelization and partition downloading possibly. But that's certainly something for a future enhancement I think.

bigquery/arrow.go
```diff
@@ -19,49 +19,107 @@ import (
...
+	}
+	if r.buf == nil { // init with schema
+		buf := bytes.NewBuffer(r.it.SerializedArrowSchema())
+		r.buf = buf
+	}
+	n, err := r.buf.Read(p)
+	if err == io.EOF {
+		batch, err := r.it.Next()
+		if err == iterator.Done {
+			return -1, io.EOF
```
shollyman, line 89: Reading of the io.Reader interface would suggest we should return 0 rather than a negative value.

bigquery/storage_integration_test.go
```diff
@@ -418,11 +423,82 @@ func TestIntegration_StorageReadCancel(t *testing.T) {
...
+		t.Fatalf("expected iterator to be accelerated: %v", err)
+	}
+	arrowItReader := NewArrowIteratorReader(arrowIt)
+
+	records := []arrow.Record{}
+	r, err := ipc.NewReader(arrowItReader)
+	numrec := 0
+	for r.Next() {
+		rec := r.Record()
+		rec.Retain()
```
shollyman, line 461: I'm confused by the release/retains here, but I've not been spending much time with arrow recently.  If you retain individual records do you need to release them?

alvarowolfx: they are going to be releases by the ipc.Reader later on the `r.Release()` call right after.

bigquery/arrow.go
alvarowolfx: good points, just sent a fix

bigquery/storage_integration_test.go
zeroshade: not exactly. If you call `Retain` on an individual record, then you will need to call `Release` on that record.

The `ipc.Reader` keeps only the current Record, reusing that member. When you call `Next()` it will release the record it had before loading the next one. This is why you need to call `Retain` on the records that you put into the slice, so that they aren't deallocated by the `ipc.Reader` calling `Release` on them. However you should also add a `defer rec.Release()` in the loop to ensure that record gets released, the `ipc.Reader` will not retain any references to those records and therefore will not call `Release` on them.

bigquery/storage_integration_test.go
```diff
@@ -418,11 +423,82 @@ func TestIntegration_StorageReadCancel(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	totalFromSQL := sumValues[0].(int64)
+
+	tr := array.NewTableReader(arrowTable, arrowTable.NumRows())
+	var totalFromArrow int64
+	for tr.Next() {
+		rec := tr.Record()
+		vec := array.NewInt64Data(rec.Column(1).Data())
```
zeroshade, line 494: I'm confused here, why `array.NewInt64Data(rec.Column(1).Data())`? Why not just `rec.Column(1).(*array.Int64)`?

You're performing an additional allocation here when you don't need to be, and it's also a potential memory leak since this *new* array instance will not get released when the record is released.

alvarowolfx: I didn't know that the `rec.Column` here could be type converted to `*array.Int64` directly. Is always hard to know when you can do those type conversions directly, so I was trying to use the API methods to do that instead. I'll make the change here.

bigquery/storage_integration_test.go
alvarowolfx: interesting, I didn't know that. I'll make the changes to call `rec.Release` on each record.

bigquery/storage_integration_test.go
zeroshade: Yea, `rec.Column` returns an `arrow.Array` interface value, so as long as you know the column is an int64 column, then you can type assert it to `*array.Int64`. Alternately you can do a type switch / check the data type, etc. before doing the type assertion. The intent is to minimize copying and minimize allocations when handling arrays and utilizing them for records etc.

shollyman (approved):

bigquery/storage_integration_test.go
```diff
@@ -418,11 +423,83 @@ func TestIntegration_StorageReadCancel(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	sumValues := []Value{}
+	err = sumIt.Next(&sumValues)
+	if err != nil {
+		t.Fatal(err)
+	}
+	totalFromSQL := sumValues[0].(int64)
+
+	tr := array.NewTableReader(arrowTable, arrowTable.NumRows())
```
zeroshade, line 491: add `defer tr.Release()` please.

bigquery/arrow.go
```diff
@@ -19,49 +19,107 @@ import (
...
+	}
+	return n, err
+}
+
+type arrowDecoder struct {
+	tableSchema Schema
+	arrowSchema *arrow.Schema
+}
+
+func newArrowDecoder(arrowSerializedSchema []byte, schema Schema) (*arrowDecoder, error) {
```
zeroshade, line 102: would probably be worthwhile (though certainly could be done as a follow-up) to allow passing a `memory.Allocator` interface here that would be stored in the `arrowDecoder` to allow a user to configure how memory gets allocated for the arrow batches (it would be passed as `ipc.WithAllocator(mem)` to `ipc.NewReader`)

In *most* cases users would probalby just use `memory.DefaultAllocator` but in other cases, depending on the constraints of the system, they might want to use a custom allocator such as a [`malloc` based allocator](https://pkg.go.dev/github.com/apache/arrow/go/v13@v13.0.0/arrow/memory/mallocator) that uses C memory to avoid garbage collection passes, or any other custom allocation they might want for specialized situations.

The other benefit of this would be that you could use `memory.CheckedAllocator` in unit tests to verify that everything properly has `Release` called if necessary, etc.

bigquery/arrow.go
```diff
@@ -19,49 +19,107 @@ import (
...
 	}
 	return p, nil
 }
 
-func (ap *arrowDecoder) createIPCReaderForBatch(serializedArrowRecordBatch []byte) (*ipc.Reader, error) {
-	buf := bytes.NewBuffer(ap.rawArrowSchema)
-	buf.Write(serializedArrowRecordBatch)
-	return ipc.NewReader(buf, ipc.WithSchema(ap.arrowSchema))
+func (ap *arrowDecoder) createIPCReaderForBatch(arrowRecordBatch *ArrowRecordBatch) (*ipc.Reader, error) {
+	return ipc.NewReader(arrowRecordBatch, ipc.WithSchema(ap.arrowSchema))
```
zeroshade, line 117: as above, it would be great (but could be a follow-up) if either `newArrowDecoder` accepted a `memory.Allocator` which would get used here as `ipc.WithAllocator(mem)` or if *this* method optionally took an allocator (defaulting to `memory.DefaultAllocator` if `nil` was passed)

bigquery/storage_integration_test.go
zeroshade: If you want to verify that everything is released/retained correctly,  you could use [`memory.CheckedAllocator`](https://pkg.go.dev/github.com/apache/arrow/go/v13@v13.0.0/arrow/memory#NewCheckedAllocator) and `defer mem.AssertSize(t, 0)` then pass the checked allocator to everything (like to `ipc.NewReader`) so that it is used for all the memory allocations. 

Not absolutely necessary, but an optional way to add some assurances if desired.

zeroshade (approved):
Added a couple comments, but overall this looks good to me, though I would like to point at https://github.com/googleapis/google-cloud-go/pull/8506#issuecomment-1732066338 for possible consideration.

ADBC allows for a query to return a result set via partition identifiers (arbitrary byte blobs for a given driver) and then allow the consumer to retrieve the streams of data from each partition in parallel. Since BigQuery already can return the data as multiple streams, there could be a benefit to allowing this Arrow iterator to return the raw underlying streams of IPC data instead of only exposing an interface for the collected streams with the parallelization handled in here, rather than by the consumer. 

It's not something that I believe should block this PR, but should possibly be picked up in a follow-up as an enhancement.

bigquery/storage_integration_test.go
alvarowolfx: I'm gonna add support for changing the allocator just internally for now for test purposes, I liked the idea of verifying that there are no memory leaks. Thanks for the tip.

bigquery/storage_integration_test.go
alvarowolfx: after adding support for changing the allocator I found that without the `tr.Release` a memory leak was happening. Good catch and awesome tips on how to catch those leaks 🎉

alvarowolfx:
I'll create another issue to keep track of this request as a future enhancement. Thanks for the deeper review on the PR, I'm gonna push some improvements based on the comments.

> Added a couple comments, but overall this looks good to me, though I would like to point at [#8506 (comment)](https://github.com/googleapis/google-cloud-go/pull/8506#issuecomment-1732066338) for possible consideration.
>

Yifeng-Sigma:
I'm wondering if we have plans to make `arrowDecoder` public, so we can obtain `arrow.Record`. Currently seems there's no way to get `arrow.Record`.

PR #8509, opened 2023-08-29
feat(storage): export gRPC client constructor

tritone:
This publicly exports the gRPC client constructor, making it possible to call directly rather than access via an environment variable.

Also adds public docs about the gRPC API which is still in private preview, as well as basic instructions for DirectPath.

Updates benchmarking script to call the new constructor directly as well.


storage/storage.go
```diff
@@ -220,11 +214,20 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 	}, nil
 }
 
-// newGRPCClient creates a new Storage client that initializes a gRPC-based
-// client. Calls that have not been implemented in gRPC will panic.
+// NewGRPCClient creates a new Storage client using the gRPC transport and API.
+// Client methods which have not been implemented in gRPC will return an error.
+// In particular, methods for Cloud Pub/Sub notifications are not supported.
```
BrennaEpp, line 219: Re: other particularities that are different between transports (for example: iirc, requests to read ACL fail in JSON when UBLA is enabled but not in GRPC) - should those be documented here and/or in relevant comments at this point?

storage/doc.go
```diff
@@ -331,6 +331,29 @@ to add a [custom audit logging] header:
 	// Use client as usual with the context and the additional headers will be sent.
 	client.Bucket("my-bucket").Attrs(ctx)
 
+# Experimental gRPC API
+
+This package now includes support for the Cloud Storage gRPC API, which is now
+in preview. This implementation uses gRPC rather than the current JSON & XML
```
BrennaEpp, line 337:
```suggestion
This package now includes support for the Cloud Storage gRPC API, which is currently
in preview. This implementation uses gRPC rather than the current JSON & XML
```

storage/storage.go
tritone: I think maybe this can wait for GA; things are still a little in flux at this point. There are some things which aren't supported yet but are coming soon (e.g. MatchGlob for listing objects)

storage/doc.go
```diff
@@ -331,6 +331,29 @@ to add a [custom audit logging] header:
...
+		// TODO: Handle error.
+	}
+	// Use client as usual.
+
+If the application is running within GCP, users may get better performance by
+enabling DirectPath (enabling requests to skip load balancing steps). To enable,
+set the environment variable `GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS=true` and add
+the following side-effect import to your application:
+
+	_ "google.golang.org/grpc/xds/googledirectpath"
```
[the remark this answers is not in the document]
tritone: I guess it does; when I try to remove it from the benchmarking script I get a mysterious "panic: send on closed channel". This does not seem to be well documented anywhere publicly but I will ask around...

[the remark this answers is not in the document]
tritone: Looks like yes, thanks for catching this.

storage/doc.go
```diff
@@ -331,6 +331,32 @@ to add a [custom audit logging] header:
 	// Use client as usual with the context and the additional headers will be sent.
 	client.Bucket("my-bucket").Attrs(ctx)
 
+# Experimental gRPC API
+
+This package includes support for the Cloud Storage gRPC API, which is currently
+in preview. This implementation uses gRPC rather than the current JSON & XML
```
BenWhitehead, line 337: I'm not familiar with go's conventions, but if `in preview` is sufficient for "if we identify something before GA we have to break, you are expected to update and accommodate the break", this is good. If not, a stronger disclaimer might be warranted.

storage/doc.go
```diff
@@ -331,6 +331,32 @@ to add a [custom audit logging] header:
...
+
+	ctx := context.Background()
+	client, err := storage.NewGRPCClient(ctx)
+	if err != nil {
+		// TODO: Handle error.
+	}
+	// Use client as usual.
+
+If the application is running within GCP, users may get better performance by
+enabling DirectPath (enabling requests to skip load balancing steps). To enable,
```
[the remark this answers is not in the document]
tritone: Good point, I can correct this.

storage/doc.go
tritone: I can add an additional sentence specifically stating this.

storage/doc.go
```diff
@@ -331,6 +331,32 @@ to add a [custom audit logging] header:
...
+set the environment variable `GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS=true` and add
```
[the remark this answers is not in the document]
tritone: It seems like this is still used to some extent inside the dial logic for go: https://github.com/googleapis/google-api-go-client/blob/62d88ff7c06b828324b3cf5fd004cdf75838f496/transport/grpc/dial.go#L241

I need to investigate how this actually works in practice...

BenWhitehead (approved):

PR #8534, opened 2023-09-07
chore: remove now unused samples build script

codyoss:


quartzmo (approved):
Thanks for cleaning this up!!

PR #8540, opened 2023-09-10
fix(datastore): Allow saving nested byte slice

bhshkh:
**Issue**: Saving nested byte slice through Datastore Go client libary throws error `flattening nested structs leads to a slice of slices`
**Cause**:  Datastore does not allow saving entities with nested lists and throws error: `rpc error: code = InvalidArgument desc = list_value cannot contain a Value containing another list_value`.  

To avoid this, the Datastore client library has a validation to prevent slice of slices. But it does not check the datatype of elements in the slice. This is incorrect since Datastore service allows saving nested list of bytes (In Go, byte is an alias for Uint8 ) . So, the library should allow such lists as well.

**Fix**: Correct the validation condition


datastore/prop.go
```diff
@@ -183,7 +183,7 @@ func validateChildType(t reflect.Type, fieldName string, flatten, prevSlice bool
 
 	switch t.Kind() {
 	case reflect.Slice:
-		if flatten && prevSlice {
+		if flatten && prevSlice && t.Elem().Kind() != reflect.Uint8 {
```
enocom, line 186: Maybe a dumb question: why `reflect.Uint8`? How does a slice of strings get caught here?

bhshkh: byte is an alias of Uint8. 
For slice of strings, `t.Elem().Kind() != reflect.Uint8` will be false since `t.Elem().Kind()` will return String

enocom: Got it. In that case a little comment to that effect would be nice for future readers. Something like, "Uint8 is an alias for byte. A slice of bytes is acceptable. A slice of others types is not."

enocom (approved):

datastore/prop.go
```diff
@@ -183,7 +183,8 @@ func validateChildType(t reflect.Type, fieldName string, flatten, prevSlice bool
 
 	switch t.Kind() {
 	case reflect.Slice:
-		if flatten && prevSlice {
+		// Uint8 is an alias for byte. A slice of bytes is acceptable. A slice of others types is not.
+		if flatten && prevSlice && t.Elem().Kind() != reflect.Uint8 {
```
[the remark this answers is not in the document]
bhshkh: For slice of bytes, t.Kind() will be reflect.Slice and t.Elem().Kind() will be reflect.Uint8
For slice of strings, t.Kind() will be reflect.Slice and t.Elem().Kind() will be reflect.String
Similarly for slices of other datatypes.

datastore/datastore_test.go
```diff
@@ -614,6 +614,75 @@ func TestPutInvalidEntity(t *testing.T) {
...
+			key:         NameKey("OuterStr", "OuterStr1", nil),
+			wantFailure: true,
+			wantErrMsg:  "flattening nested structs leads to a slice of slices",
+		},
+	}
+
+	for _, tc := range testcases {
+		_, gotErr := client.Put(ctx, tc.key, tc.src)
+		gotFailure := gotErr != nil
+		if gotFailure != tc.wantFailure ||
```
kolea2, line 679: can we also assert the non failed test case has the values we want?

bhshkh: Done

kolea2 (approved):

PR #8543, opened 2023-09-10
fix(storage): do not inhibit the dead code elimination.

### LINKED ISSUE #8745: storage: importing cloud.google.com/go/storage disables the dead code elimination
When the go compiler encounters a call to `MethodByName()`, it assumes that every method of every reachable type can be accessed, and nothing may be eliminated. `cloud.google.com/go/storage` uses `MethodByName()` to invoke methods like `IfMetagenerationMatch()`. This disables the DCE in all users of GCS.

Go 1.22 has gained support for calls like `MethodByName("Foobar")` where the name of a method to look up is a compile-time constant. In this case the compiler no longer disables the DCE, and only keeps methods named Foobar.

Rewrite `setConditionField()` so that `MethodByName()` is only called with compile-time constants. This way importing `cloud.google.com/go/storage` will not inhibit the DCE.

dominiquelefevre:
When the go compiler encounters a call to MethodByName(), it assumes that every method of every reachable type can be accessed, and nothing may be eliminated.

Go 1.22 has gained support for calls like MethodByName("Foobar"). If the argument to MethodByName() is a compile-time constant, then the compiler no longer disables the DCE, and only keeps methods named Foobar.

Rewrite setConditionField() so that MethodByName() is only called with compile-time constants. This way importing cloud.google.com/go/storage will not inhibit the DCE.

Fixes #8745


dominiquelefevre:
Hi there! Does anybody look at PRs in this repo? @codyoss? @jadekler? Please do have a look at the patch, or assign it to a person who reviews changes in /storage.

codyoss:
Hey @dominiquelefevre, this repo is monitored and we do accept PRs. That being said we may not always get to them immediately so please be patient with us.

Also, we encourage all PRs to start with an issue first for discussion and triage to see if it is a change we are willing to make -- it saves both parties time in the long run. I encourage to to please open a corresponding issue to this PR. Thanks.

storage/storage.go
```diff
@@ -1897,16 +1897,57 @@ func applySourceCondsProto(gen int64, conds *Conditions, call *storagepb.Rewrite
...
+// We also make sure to supply a compile-time constant to MethodByName
+// so an not to disable the DCE.
+func setGeneration(cval reflect.Value, value interface{}) bool {
+	m := cval.MethodByName("Generation")
+	if m.IsValid() {
+		m.Call([]reflect.Value{reflect.ValueOf(value)})
 	}
-	m.Call([]reflect.Value{reflect.ValueOf(value)})
-	return true
+	return m.IsValid()
```
BrennaEpp, line 1910: Can we keep these lines in a single method?

storage/storage.go
```diff
@@ -1897,16 +1897,57 @@ func applySourceCondsProto(gen int64, conds *Conditions, call *storagepb.Rewrite
...
-// setConditionField sets a field on a *raw.WhateverCall.
+// setGeneration sets Generation on a *raw.WhateverCall.
 // We can't use anonymous interfaces because the return type is
 // different, since the field setters are builders.
-func setConditionField(call reflect.Value, name string, value interface{}) bool {
-	m := call.MethodByName(name)
-	if !m.IsValid() {
-		return false
+// We also make sure to supply a compile-time constant to MethodByName
+// so an not to disable the DCE.
```
BrennaEpp, line 1904: Please link the 
```suggestion
// We also make sure to supply a compile-time constant to MethodByName;
// otherwise, the Go Linker will disable dead code elimination, leading to larger
// binaries for all packages that import storage.
```

BrennaEpp (commented):
Thank you for the PR!

storage/storage.go
dominiquelefevre: Done.

storage/storage.go
dominiquelefevre: Done.

dominiquelefevre:
ping @BrennaEpp

dominiquelefevre:
Hey, anybody out here? @BrennaEpp?

BrennaEpp (approved):

Issue #8325, opened 2023-07-25
datastore: can't put null into any type


RPGillespie6:
**Client**

Datastore

**Go Environment**

`go version go1.18.3 linux/amd64`

**Code**

Repro:

```go
package main

import (
	"context"
	"fmt"

	"cloud.google.com/go/datastore"
)

type Foo struct {
	Bar any `json:"bar"`
}

func main() {
	// init datastore client
	client, err := datastore.NewClient(context.Background(), "my-project")
	if err != nil {
		panic(err)
	}

	// create a new Foo
	foo := &Foo{
		Bar: nil,
	}

	// save the Foo
	key, err := client.Put(context.Background(), datastore.IncompleteKey("Foo", nil), foo)

	// check for errors
	if err != nil {
		panic(err)
	}

	// Get foo with key
	foo = &Foo{}
	err = client.Get(context.Background(), key, foo)
	if err != nil {
		panic(err)
	}

	// Print foo
	fmt.Println(foo)
}
```

This code panics with:
```
panic: reflect: call of reflect.Value.Type on zero Value

goroutine 1 [running]:
reflect.Value.Type({0x0?, 0x0?, 0x203000?})
	/usr/local/go/src/reflect/value.go:2453 +0x12e
cloud.google.com/go/datastore.setVal({0xa16840?, 0xc000115680?, 0xc00048b440?}, {{0xc00048b440, 0x3}, {0x0, 0x0}, 0x0})
	/home/user/go/pkg/mod/cloud.google.com/go/datastore@v1.12.1/load.go:306 +0x3d5
cloud.google.com/go/datastore.(*propertyLoader).loadOneElement(0xc0004e9830, {0xc0004be0e0, 0x1, 0x1}, {0xa51fa0?, 0xc000115680?, 0xc0004e9780?}, {{0xc00048b440, 0x3}, {0x0, ...}, ...}, ...)
	/home/user/go/pkg/mod/cloud.google.com/go/datastore@v1.12.1/load.go:209 +0xa8b
cloud.google.com/go/datastore.(*propertyLoader).load(0xc000114a30?, {0xc0004be0e0?, 0x203000?, 0x203000?}, {0xa51fa0?, 0xc000115680?, 0x985045?}, {{0xc00048b440, 0x3}, {0x0, ...}, ...}, ...)
	/home/user/go/pkg/mod/cloud.google.com/go/datastore@v1.12.1/load.go:79 +0xdf
[...]

PR #8544, opened 2023-09-11
fix(datastore): Handle loading nil values

bhshkh:
**Issue**: In Go, *any* datatype is an alias for empty interface.
Getting entities with nil values in empty interface causes panic.
For sample code, please see: https://github.com/googleapis/google-cloud-go/issues/8325#issue-1821227551

**Cause**: When a reflect value is untyped nil, its type cannot be determined by using Type(). The code panics with error: `call of reflect.Value.Type on zero Value`

**Fix**: If the field being set is an interface and the value to set is untyped nil, set the field to zero value.


enocom (approved):

datastore/load.go
```diff
@@ -302,11 +302,15 @@ func setVal(v reflect.Value, p Property) (s string) {
 			return fmt.Sprintf("%v is unsettable", v.Type())
 		}
 
-		rpValue := reflect.ValueOf(pValue)
-		if !rpValue.Type().AssignableTo(v.Type()) {
-			return fmt.Sprintf("%q is not assignable to %q", rpValue.Type(), v.Type())
+		if pValue == nil {
+			v.Set(reflect.Zero(v.Type()))
```
kolea2, line 306: it might be nice to add a quick comment here explaining why we are converting nil to Zero

datastore/load_test.go
```diff
@@ -1290,6 +1290,23 @@ func TestLoadNull(t *testing.T) {
 	}
 }
 
+func TestLoadNilInterface(t *testing.T) {
```
kolea2, line 1293: I think a system test would also be good to have.

kolea2 (changes requested):

datastore/load_test.go
```diff
@@ -1290,6 +1290,23 @@ func TestLoadNull(t *testing.T) {
...
+	withAny1 := &WithAny{}
+	err := loadEntityProto(withAny1, &pb.Entity{
+		Key: keyToProto(testKey0),
+		Properties: map[string]*pb.Value{
+			"AnyField": {ValueType: &pb.Value_NullValue{}},
+		},
+	})
+	if err != nil {
+		t.Fatalf("got: %v, want: nil", err)
+	}
```
danieljbruce, line 1307: Maybe an additional assert check to ensure the value loaded is actually zero?

datastore/load.go
bhshkh: Added

datastore/load_test.go
bhshkh: Added

datastore/load_test.go
bhshkh: Added

kolea2 (approved):

PR #8553, opened 2023-09-11
chore(storage): remove closing of channel

### LINKED ISSUE #8515: storage: race condition in benchmarking warmup
**Client**

Storage
 
**Environment**

Linux cloudtop

I'm seeing a race condition which happens maybe 50% of the time when I run the benchmark after warmups were added in #8418. It looks like in some cases the results channel gets closed (in `discardBenchmarkResults`) while the main benchmark itself is running.

I'm making a PR to skip this logic entirely when duration is zero (which we should do anyway) but it may be something that could occur with non-zero duration also.

Example command: `go run main -project <project id> -samples 10 -api DirectPath`

Output:
```
panic: send on closed channel

goroutine 162 [running]:
main.runOneOp({0xc0005e82c0?, 0x1527600?}, 0xc0006e2000, 0xc00076fdb0)
	/usr/local/google/home/cjcotter/code/gocloud/storage/internal/benchmarks/w1r3.go:237 +0x185
main.(*w1r3).run(0xc0004c8000, {0x1527600, 0xc0004c6aa0})
	/usr/local/google/home/cjcotter/code/gocloud/storage/internal/benchmarks/w1r3.go:162 +0x65a
main.main.func2()
	/usr/local/google/home/cjcotter/code/gocloud/storage/internal/benchmarks/main.go:328 +0x182
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/usr/local/google/home/cjcotter/go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75 +0x59
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1
	/usr/local/google/home/cjcotter/go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:72 +0x96
exit status 2
```

BrennaEpp:
It's unnecessary to close channel as no other routine must be informed. This fixes the race condition as well as it avoids closing the newly created channel assigned to the results var.

Fixes #8515


storage/internal/benchmarks/warmup.go
```diff
@@ -66,7 +66,6 @@ func discardBenchmarkResults(ctx context.Context) {
 		for {
 			select {
 			case <-ctx.Done():
```
tritone, line 68: It seems like it's possible that both cases are true at the same time and we randomly select the first case. In this situation, there might still be results to read from the channel, which we'd need to flush -- either here, or before we begin the main benchmarking run. Does this seem accurate?

BrennaEpp: In such a case, `results = make(chan benchmarkResult, 100)` in the main benchmarking recording code will discard any results queued on the old channel. The race condition that was happening was that `make` was happening before the `case <-ctx.Done` was, and that was closing the newly made channel. 
The bigger problem here would be the second case being selected, though it is extremely unlikely as this routine would need to not run until an object is written and the throughput is sent to the result channel. However, since it is technically possible, I added some lines here to deal with it.

BrennaEpp: As discussed, I removed some of the async logic to make this easier to understand. Race condition should be fixed in any case

tritone (approved):

PR #8562, opened 2023-09-12
chore(storage): add server mode [benchmarks]

BrennaEpp:


quartzmo:
Typo in the title of this PR: "shore"

storage/internal/benchmarks/main.go
```diff
@@ -21,12 +21,15 @@ import (
...
 	"log"
-	"math/rand"
+	"net/http"
 	"os"
 	"strconv"
 	"strings"
 	"time"
 
+	epb "github.com/cloudprober/cloudprober/probes/external/proto"
+	"github.com/cloudprober/cloudprober/probes/external/serverutils"
```
tritone, line 31: Fix export ordering

storage/internal/benchmarks/main.go
```diff
@@ -225,7 +236,12 @@ func parseFlags() {
 func main() {
 	log.SetOutput(os.Stderr)
 	parseFlags()
-	rand.Seed(time.Now().UnixNano())
+
+	go func() {
+		if err := http.ListenAndServe("localhost:8080", nil); err != nil {
```
tritone, line 241: What does this do as a stand-alone? add a comment?

storage/internal/benchmarks/main.go
```diff
@@ -269,96 +286,33 @@ func main() {
...
-				// to indicate that we did see an error, even though we continue.
-				// Cleanup may be expected to fail if there is an issue with the run.
-				log.Printf("run cleanup failed: %v", err)
-				exitWithErrorCode = true
-			}
-			return nil
-		})
+	if serverMode {
+		// Server mode blocks forever servicing probe requests
+		runInServerMode(ctx, opts)
```
tritone, line 304: Should this have an exit()? If runInServerMode does return then we would probably not want to continue right?

storage/internal/benchmarks/main.go
```diff
@@ -494,6 +448,139 @@ func startRecordingResults(w io.Writer, g *errgroup.Group, oType outputType) {
...
+
+// runInServerMode blocks forever servicing probe requests.
+// Number of samples requested is ignored in this mode.
+// Timeouts must be properly set at the probe level to ensure requests aren't
+// being serviced concurrently
+func runInServerMode(ctx context.Context, opts *benchmarkOptions) {
+	results = make(chan benchmarkResult, 4)
+
+	serverutils.Serve(func(request *epb.ProbeRequest, reply *epb.ProbeReply) {
+		timeLimit := time.Millisecond * time.Duration(request.GetTimeLimit())
```
tritone, line 542: Maybe note that this is per-sample?

tritone (commented):
couple questions, looks good overall!

storage/internal/benchmarks/main.go
BrennaEpp: What do you mean? Should I ungroup these?

storage/internal/benchmarks/main.go
BrennaEpp: That is correct; although it should never return, I've added a check here to make sure we do error out if it does.

storage/internal/benchmarks/main.go
BrennaEpp: Added a comment; let me know if it makes sense or is still missing context.

storage/internal/benchmarks/main.go
BrennaEpp: Whoops, this was for obtaining pprof samples. I removed it; I'll open a PR separately to see if it makes more sense to wrap specific parts of the code with cpu profiling and take periodic memory samples, or else re-add this back in for ad-hoc profile collecting.

storage/internal/benchmarks/main.go
tritone: Oh I guess it's just weird that these are grouped separately; they can go with the other non-core imports below.

tritone (approved):

PR #8566, opened 2023-09-13
fix(bigquery): dependency detection on proto conversion

### LINKED ISSUE #8471: cloud.google.com/go/bigquery/storage/managedwriter/adapt: StorageSchemaToProto2Descriptor fails with certain structs
**Client**

cloud.google.com/go/bigquery/storage/managedwriter/adapt

**Environment**

local unit tests

**Go Environment**

$ go version
go version go1.20.6 linux/amd64

**Code**

```go
package main

import (
	"encoding/json"
	"fmt"

	"cloud.google.com/go/bigquery/storage/managedwriter/adapt"
	"google.golang.org/protobuf/reflect/protodesc"
	"google.golang.org/protobuf/reflect/protoreflect"
	"google.golang.org/protobuf/types/descriptorpb"
	"google.golang.org/protobuf/types/dynamicpb"
)

type Bar struct {
	BarID string `json:"barid"`
}

type Biz struct {
	BizID string `json:"bizid"`
}

type Foo struct {
	BarTest1    *Bar     `json:"bartest1,omitempty"`
	FullProfile *Profile `json:"fullprofile,omitempty"`
}

type Profile struct {
	BizTest1 *Biz `json:"biztest1,omitempty"` // Comment this out
	BarTest2 *Bar `json:"bartest2,omitempty"`
	BarTest3 *Bar `json:"bartest3,omitempty"`
	// BizTest1 *Biz `json:"biztest1,omitempty"` // And uncomment this to make it work
}

func main() {
	s, err := bigquery.InferSchema(Foo{})
	if err != nil {
		panic(err)
	}
	md, _ := setupDynamicDescriptors(s)

	message := dynamicpb.NewMessage(md)

	f := Foo{
		BarTest1: &Bar{BarID: "barTest1Value"},
		FullProfile: &Profile{
			BarTest2: &Bar{BarID: "barTest2Value"},
			BarTest3: &Bar{BarID: "barTest3Value"},
			BizTest1: &Biz{BizID: "bizIDValue"},
		},
	}
	bb, err := json.Marshal(f)
	if err != nil {
		panic(err)
	}

	err = protojson.Unmarshal(bb, message)
	if err != nil {
		panic(fmt.Errorf("failed to Unmarshal json message for: %v", err))
	}

	_, err = proto.Marshal(message)
	if err != nil {
		panic(fmt.Errorf("failed to marshal proto bytes for row : %v", err))
	}
}

func setupDynamicDescriptors(schema bigquery.Schema) (protoreflect.MessageDescriptor, *descriptorpb.DescriptorProto) {
	convertedSchema, err := adapt.BQSchemaToStorageTableSchema(schema)
	if err != nil {
		panic(fmt.Errorf("adapt.BQSchemaToStorageTableSchema: %v", err))
	}

	descriptor, err := adapt.StorageSchemaToProto2Descriptor(convertedSchema, "root")
	if err != nil {
		panic(fmt.Errorf("adapt.StorageSchemaToDescriptor: %v", err))
	}

	messageDescriptor, ok := descriptor.(protoreflect.MessageDescriptor)
	if !ok {
		panic(fmt.Errorf("adapted descriptor is not a message descriptor"))
	}

	return messageDescriptor, protodesc.ToDescriptorProto(messageDescriptor)
}

```

**Expected behavior**

No errors to occur 

**Actual behavior**

Fails to convert `StorageSchemaToProto2Descriptor` with the following error.
```
adapt.StorageSchemaToDescriptor: conversion error in location "root__FullProfile": couldn't convert message: proto: message field "root__FullProfile.bartest2" cannot resolve type: "*.root__BarTest1" not found
```

If you change the order of the fields as suggested in the code above this will work correctly. 

**Additional context**

I believe this is to do with how dependencies are cached and calculated. 
https://github.com/googleapis/google-cloud-go/blob/main/bigquery/storage/managedwriter/adapt/protoconversion.go#L189

Getting `FullName()` here returns an empty string for both `foundDesc` and `curDep`. I don't fully understand what the code is doing here to generate that but perhaps we can use `Path()` instead of `FullName()`?

alvarowolfx:
Improve dependency detection on proto conversion. In some scenarios where a struct is reused on root and nested types, the current check to see if a message descriptor belongs to a file descriptor doesn't work. Now we check the parent file descriptor and the child messages defined on a file descriptor.

- [x] ~TODO: Test case was copied from the reported issue, but might be better to convert it to the existing table driven.~
- [ ] nit: `messageDepensOnFile` method name is not looking good.

Fixes #8471


shollyman (approved):
Generally looks good, but can we update these message/field names to highlight which one is the repeated type and which are the outer/inner container messages?  I think it would improve the readability of this particular test case, as I spent a too many neurons trying to keep track of that.

PR #8567, opened 2023-09-13
feat(auth): add httptransport package

codyoss:
This package is the equivalent of
google.golang.org/api/transport/http for the new auth library. The biggest different is the main function does not return an endpoint so we will want to add a helper in gax or somewhere for this. Also, there are is an extra helper to add a TokenProvider to a http.Client which should elveate some user needs for extra extenisablity of a client. They can simply customize away and user the helper to add auth.

This PR is part 1 of 2 to make it easier to review.  Took away anything dealing with certs/tls for now and will fast follow and comment back in some of the code in this PR.


auth/httptransport/trace.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+var _ propagation.HTTPFormat = (*httpFormat)(nil)
+
+// httpFormat implements propagation.httpFormat to propagate
+// traces in HTTP headers for Google Cloud Platform and Cloud Trace.
+type httpFormat struct{}
+
+// SpanContextFromRequest extracts a Cloud Trace span context from incoming requests.
+func (f *httpFormat) SpanContextFromRequest(req *http.Request) (sc trace.SpanContext, ok bool) {
```
quartzmo, line 41: I believe this entire file is identical to (and likely originally copied from) https://github.com/census-instrumentation/opencensus-go/blob/master/exporter/stackdriver/propagation/http.go. Just wondering if there is a better way to implement this support for context propagation. Most obviously, why do we need to support extracting Cloud Trace span context from incoming requests in client libraries? This appears to be server code.

auth/httptransport/trace.go
```diff
@@ -0,0 +1,92 @@
...
+		return sc, true
+	}
+	o, err := strconv.ParseUint(h[2:], 10, 64)
+	if err != nil {
+		return trace.SpanContext{}, false
+	}
+	sc.TraceOptions = trace.TraceOptions(o)
+	return sc, true
+}
+
```
quartzmo, line 86: Functionally, I think this func is the only part of this file that we need. Although I think `SpanContextFromRequest` is needed to satisfy the interface.

auth/httptransport/trace.go
codyoss: You may be right, I copied most of it over from https://github.com/googleapis/google-api-go-client/blob/main/transport/http/internal/propagation/http.go.

From my understanding you need to extract the header to create a base span for this realm of the application. Because it is extracted here, any future spans created from something like in storage will have this extracted context as the parent. I believe it is all about chaining the spans together.

auth/httptransport/trace.go
codyoss: I believe both are needed from my understanding. This one is about adding context to outgoing requests and the other is extracting context from incoming requests.

auth/httptransport/httptransport.go
```diff
@@ -0,0 +1,186 @@
...
+		return errors.New("httptransport: DisableAuthentication is incompatible with options that set or detect credentials")
+	}
+	return nil
+}
+
+func (o *Options) client() *http.Client {
+	if o.DetectOpts != nil && o.DetectOpts.Client != nil {
+		return o.DetectOpts.Client
+	}
+	return nil
```
noahdietz, line 86: So the equivalent method on `detect.Options` will actually [use a "default client" here](https://github.com/googleapis/google-cloud-go/blob/e4a54cdee3f4b9424830bdbd1089bf1d0e217b6d/auth/detect/detect.go#L188). Should that be done here as well?

auth/httptransport/trace.go
```diff
@@ -0,0 +1,92 @@
...
+	"go.opencensus.io/trace"
+	"go.opencensus.io/trace/propagation"
+)
+
+const (
+	httpHeaderMaxSize = 200
+	cloudTraceHeader  = `X-Cloud-Trace-Context`
+)
+
+var _ propagation.HTTPFormat = (*httpFormat)(nil)
```
noahdietz, line 34: Just double checking, maybe this is worth a comment, but this LoC is just to force the `propagation` package dependency and assert that `httpFormat` implements the `propagation.HTTPFormat` interface, right?

quartzmo (approved):

auth/httptransport/trace.go
codyoss: Discussed more offline, Chris is right we likely don't need this. But not removing it as of right now to be safe as it was in the old impl.

auth/httptransport/httptransport.go
codyoss: This is on purpose for some logic that is not yet in this PR. For some of the certs logic we do/don't do things depending if a user explicitly set a client. So nil is valid. Added a comment.

auth/httptransport/trace.go
codyoss: Yes, it is just a check to make sure this type satisfies an interface. This import part will happen regardless do to the import in httptransport.go. Added a comment to be more clear.

noahdietz (approved):

PR #8569, opened 2023-09-13
feat(auth): add cert support to httptransport

codyoss:
Part 2 of https://github.com/googleapis/google-cloud-go/pull/8567


auth/httptransport/httptransport.go
```diff
@@ -153,22 +153,21 @@ func NewClient(opts *Options) (*http.Client, error) {
...
-	// }
-	trans, err := newTransport(defaultBaseTransport(nil), opts)
+	tOpts := &transport.Options{
+		Endpoint:           opts.Endpoint,
+		ClientCertProvider: opts.ClientCertProvider,
+		Client:             opts.client(),
+	}
+	if io := opts.InternalOptions; io != nil {
+		tOpts.DefaultEndpoint = io.DefaultEndpoint
+		tOpts.DefaultMTLSEndpoint = io.DefaultMTLSEndpoint
```
noahdietz, line 164: Is it guaranteed that these properties will be set on `io`? What happens if they aren't and these become empty?

auth/internal/transport/cba.go
```diff
@@ -0,0 +1,408 @@
...
+// must be set to "true" to allow certificate to be used (including user provided
+// certificates). For details, see AIP-4114.
+func getClientCertificateSource(opts *Options) (cert.Provider, error) {
+	if !isClientCertificateEnabled() {
+		return nil, nil
+	} else if opts.ClientCertProvider != nil {
+		return opts.ClientCertProvider, nil
+	} else {
+		return cert.DefaultSource()
+	}
```
noahdietz, line 194: Nit: The `else` is extraneous
```suggestion
	}

	return cert.DefaultSource()
```

auth/internal/transport/cba.go
```diff
@@ -0,0 +1,408 @@
...
+}
+
+func isClientCertificateEnabled() bool {
+	useClientCert := os.Getenv(googleAPIUseCertSource)
+	// TODO(andyrzhao): Update default to return "true" after DCA feature is fully released.
+	return strings.ToLower(useClientCert) == "true"
```
noahdietz, line 200: Nit: is it worth using [`strconv.ParseBool`](https://pkg.go.dev/strconv#ParseBool) and maybe swallowing the error as "false"?

auth/internal/transport/cba.go
```diff
@@ -0,0 +1,408 @@
...
+	u, err := url.Parse(fixScheme(baseURL))
+	if err != nil {
+		return "", err
+	}
+	return strings.Replace(baseURL, u.Host, newHost, 1), nil
+}
+
+func fixScheme(baseURL string) string {
+	if !strings.Contains(baseURL, "://") {
+		return "https://" + baseURL
```
noahdietz, line 270:
```suggestion
		baseURL = "https://" + baseURL
```
Nit: might just be my preference tbh, feel free to ignore

auth/internal/transport/cba.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+	return baseURL
+}
+
+// GetS2AAddress returns the S2A address to be reached via plaintext connection.
```
noahdietz, line 275: Nit
```suggestion
// GetS2AAddress returns the S2A address to be reached via plaintext connection.
// Returns empty string if not set or invalid.
```

auth/internal/transport/cba.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+}
+
+func shouldUseS2A(clientCertSource cert.Provider, opts *Options) bool {
+	// If client cert is found, use that over S2A.
+	if clientCertSource != nil {
+		return false
+	}
+	// If EXPERIMENTAL_GOOGLE_API_USE_S2A is not set to true, skip S2A.
+	if strings.ToLower(os.Getenv(googleAPIUseS2AEnv)) != "true" {
```
noahdietz, line 392: ditto re: `strconv.ParseBool`

auth/httptransport/httptransport.go
codyoss: It is valid for them to be empty. In those cases it just copies over an empty string.

auth/internal/transport/cba.go
codyoss: done.

auth/internal/transport/cba.go
codyoss: done.

auth/internal/transport/cba.go
codyoss: done.

auth/internal/transport/cba.go
codyoss: done.

auth/internal/transport/cba.go
codyoss: done.

auth/httptransport/httptransport.go
noahdietz: Ok cool, thanks

noahdietz (approved):

auth/internal/transport/cba.go
```diff
@@ -0,0 +1,411 @@
...
+
+const (
+	mTLSModeAlways = "always"
+	mTLSModeNever  = "never"
+	mTLSModeAuto   = "auto"
+
+	configEndpointSuffix = "googleAutoMtlsConfiguration"
+
+	// Experimental: if true, the code will try MTLS with S2A as the default for transport security. Default value is false.
+	googleAPIUseS2AEnv     = "EXPERIMENTAL_GOOGLE_API_USE_S2A"
```
quartzmo, line 46: Has this env var already shipped? If not, elsewhere we use the `GOOGLE_API_GO_EXPERIMENTAL_` prefix, could this be changed to `GOOGLE_API_EXPERIMENTAL_USE_S2A`?

auth/internal/transport/cba.go
```diff
@@ -0,0 +1,411 @@
...
+)
+
+// Options is a struct that is duplicated information from the individual
+// transport packages in order to avoid cyclic deps.
+type Options struct {
+	Endpoint            string
+	DefaultEndpoint     string
+	DefaultMTLSEndpoint string
+	ClientCertProvider  cert.Provider
+	Client              *http.Client
```
quartzmo, line 74: Would it be possible to add docs to each field? Maybe copy the docs from the original declarations in the individual transport packages, and to each, add a note stating from where the field was duplicated?

auth/internal/transport/cba.go
```diff
@@ -0,0 +1,411 @@
...
+	mtlsOnce                sync.Once
+	mdsMTLSAutoConfigSource mtlsConfigSource
+)
+
+// Options is a struct that is duplicated information from the individual
+// transport packages in order to avoid cyclic deps.
+type Options struct {
+	Endpoint            string
+	DefaultEndpoint     string
+	DefaultMTLSEndpoint string
```
quartzmo, line 72: For example, docs for this field might mention the relationship with s2a.

auth/internal/transport/cba.go
```diff
@@ -0,0 +1,411 @@
...
+	// If client cert is found, use that over S2A.
+	if clientCertSource != nil {
+		return false
+	}
+	// If EXPERIMENTAL_GOOGLE_API_USE_S2A is not set to true, skip S2A.
+	if b, err := strconv.ParseBool(os.Getenv(googleAPIUseS2AEnv)); err == nil && b {
+		return false
+	}
+	// If DefaultMTLSEndpoint is not set and no endpoint override, skip S2A.
+	if opts.DefaultMTLSEndpoint == "" && opts.Endpoint == "" {
```
quartzmo, line 399: At this point and below, we know `EXPERIMENTAL_GOOGLE_API_USE_S2A` is true. Would it make sense to return errors to the user instead of silently skipping S2A? I'm guessing this is all spec'd somewhere, but asking just to be safe...

quartzmo (changes requested):
At a high level, i'm wondering if `cba.go` could be split into two or more files, in order to isolate S2A code from default transport, etc. The litmus test for this would be to imagine removing the S2A code later and having it be very easy to do. But the reason is not just to make it easier to remove S2A, but also to make it easier to understand.

auth/internal/transport/cba.go
codyoss: I can for sure clarify the struct in the main docs for this struct, but I am hesitant to copy field docs for this kind of struct. It just means there is one more place that documentation can drift. These all should correlate 1:1 with another struct already. But if you feel strongly here we can add fields docs.

auth/internal/transport/cba.go
codyoss: I think silently skipping is fine here for an experimental feature. I don't want to change the flow of when an error is presented unless necessary.

auth/internal/transport/cert/default_cert.go
```diff
@@ -0,0 +1,60 @@
...
+
+// defaultCertData holds all the variables pertaining to
+// the default certficate source created by DefaultSource.
+//
+// A singleton model is used to allow the source to be reused
+// by the transport layer.
+type defaultCertData struct {
+	once   sync.Once
+	source Provider
+	err    error
```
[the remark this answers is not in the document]
codyoss: It is non-nil if a cert if is found in either of those default lookups. Added a blurb around nil, nil meaning directly.

auth/internal/transport/cba.go
codyoss: see below.

auth/internal/transport/cba.go
quartzmo: > I can for sure clarify the struct in the main docs for this struct

Yes, please add docs where appropriate, thanks.

quartzmo (approved):

PR #8571, opened 2023-09-13
chore(internal/postprocessor): fix containeranalysis/v1beta1 doc url flake

noahdietz:
Adds missing `rel-path` argument to the grafeas specific post processor config entry. This config being missing combined with the shared import-path of it and its parent package meant that the doc url wasn't being calculated relative to the root module correctly and only showing up some of the time when the grafeas entry was visited last.

Also fixes a typo in a config config, and clarifies a double negative in the post processor cli docs.


codyoss (approved):
Nice find!

PR #8580, opened 2023-09-15
feat(auth): add idtoken package

codyoss:


auth/idtoken/compute.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (c computeIDTokenProvider) Token(ctx context.Context) (*auth.Token, error) {
+	v := url.Values{}
+	v.Set("audience", c.audience)
+	if c.format != ComputeTokenFormatStandard {
+		v.Set("format", "full")
+	}
+	if c.format == ComputeTokenFormatFullWithLicense {
+		v.Set("licenses", "TRUE")
```
noahdietz, line 58: out of curiosity, does this value need to be exactly `TRUE` or is it any truthy value? Just find it funny that there is an all-caps boolean string like this haha

auth/idtoken/compute.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	urlSuffix := identitySuffix + "?" + v.Encode()
+	res, err := c.client.Get(urlSuffix)
+	if err != nil {
+		return nil, err
+	}
+	if res == "" {
+		return nil, fmt.Errorf("idtoken: invalid response from metadata service")
```
noahdietz, line 66: Nit: maybe slightly more specific error? 
```suggestion
		return nil, fmt.Errorf("idtoken: invalid empty response from metadata service")
```

auth/idtoken/compute.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	return &auth.Token{
+		Value: res,
+		Type:  internal.TokenTypeBearer,
+		// Compute tokens are valid for one hour
```
noahdietz, line 71: By chance, is this documented somewhere? If it is, would it make sense to put a link to that doc in this comment? A fair answer is "Yes it is, and no it's not necessary" :)

auth/idtoken/file.go
```diff
@@ -0,0 +1,110 @@
...
+		config := impersonate.IDTokenOptions{
+			Audience:        opts.Audience,
+			TargetPrincipal: account,
+			IncludeEmail:    true,
+			Client:          opts.client(),
+			TokenProvider:   creds,
+		}
+		return impersonate.NewIDTokenProvider(&config)
+	default:
+		return nil, fmt.Errorf("idtoken: unsupported credentials type")
```
noahdietz, line 108: Nit: Is it worth putting said type in the error message? Not sure if that is leaky or not.
```suggestion
		return nil, fmt.Errorf("idtoken: unsupported credentials type: %v", t)
```

noahdietz (approved):
A couple of nits but looking great

auth/idtoken/compute.go
```diff
@@ -0,0 +1,74 @@
...
+type computeIDTokenProvider struct {
+	audience string
+	format   ComputeTokenFormat
+	client   metadata.Client
+}
+
+func (c computeIDTokenProvider) Token(ctx context.Context) (*auth.Token, error) {
+	v := url.Values{}
+	v.Set("audience", c.audience)
+	if c.format != ComputeTokenFormatStandard {
```
quartzmo, line 54: Did you want to add a unit test case for format == `ComputeTokenFormatStandard`?

auth/idtoken/file.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return nil, err
+	}
+	switch t {
+	case internaldetect.ServiceAccountKey:
+		f, err := internaldetect.ParseServiceAccount(b)
+		if err != nil {
+			return nil, err
+		}
+		opts2LO := &auth.Options2LO{
```
quartzmo, line 52: I really like the explicit options type for "2LO" here, this is much easier to follow than the old design.

auth/idtoken/idtoken.go
```diff
@@ -0,0 +1,122 @@
...
+	if b := jsonBytes(opts); b != nil {
+		return tokenProviderFromBytes(b, opts)
+	}
+	if metadata.OnGCE() {
+		return computeTokenProvider(opts)
+	}
+	return nil, fmt.Errorf("idtoken: couldn't find any credentials")
+}
+
+func jsonBytes(opts *Options) []byte {
```
quartzmo, line 98: Why not make this a method on `Options`? I believe the `opts != nil` check on 99 is unnecessary after the nil check on 83.

auth/idtoken/idtoken.go
```diff
@@ -0,0 +1,122 @@
...
+	ComputeTokenFormatFull
+	// ComputeTokenFormatFullWithLicense means the same as
+	// [ComputeTokenFormatFull] with the addition of claims about licenses
+	// associated with the instance.
+	ComputeTokenFormatFullWithLicense
+)
+
+// Options for the configuration of creation of an ID token with
+// [NewTokenProvider].
+type Options struct {
```
quartzmo, line 49: I love the new dedicated `Options` struct as opposed to the old design which used shared client Options.

auth/idtoken/compute.go
quartzmo: Note also that this is a change from the old design, which was 55 minutes in order to provide a 5 minute "buffer".

codyoss: added. Yeah removed the buffer as one is aleady builtin by wrapping it with a cachedtokenprovider. Updated the refresher code above to match this just to be safe. Good call out.

auth/idtoken/file.go
codyoss: :)

auth/idtoken/idtoken.go
codyoss: :)

auth/idtoken/idtoken.go
codyoss: done and updated.

noahdietz (approved):

quartzmo (approved):

PR #8609, opened 2023-09-25
fix(pubsublite): add missing veneer header

codyoss:
Similar to #8608


noahdietz (commented):
LGTM but I think CODEOWNERS should approve the pseudo-internal API addition.

codyoss:
@hongalex can I get a stamp on you from this? I can update the header to this year if so :P

PR #8625, opened 2023-09-28
feat(auth): add grpctransport package

codyoss:
This package is the analog to https://pkg.go.dev/google.golang.org/api/transport/grpc.


auth/grpctransport/testdata/echo.proto
noahdietz, line 1: How do we plan to keep this file auto generated with the latest? Can you include some steps in a README to regenerate them? Should we just automate regeneration via github actions somehow?

Thanks.

noahdietz (commented):
LGTM just wondering how we keep the test bindings up to date in the event new proto/grpc features land and we need to regen.

auth/grpctransport/directpath.go
```diff
@@ -0,0 +1,123 @@
...
+	if !checkDirectPathEndPoint(endpoint) {
+		return false
+	}
+	if b, _ := strconv.ParseBool(os.Getenv(disableDirectPathEnvVar)); b {
+		return false
+	}
+	return true
+}
+
+func checkDirectPathEndPoint(endpoint string) bool {
```
quartzmo, line 43: I think maybe this is the only func in this file that is tested? Any plans to add tests for the others?

quartzmo: Nm, I think `TestDialWithDirectPathEnabled` covers the rest?

auth/grpctransport/grpctransport.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	if opts.PoolSize == 0 || opts.PoolSize == 1 {
+		conn, err := dial(ctx, secure, opts)
+		if err != nil {
+			return nil, err
+		}
+		return &singleConnPool{conn}, nil
+	}
+	pool := &roundRobinConnPool{}
+	for i := 0; i < opts.PoolSize; i++ {
```
quartzmo, line 164: What do we get if PoolSize is < 0? An empty pool? Should PoolSize be validated?

auth/grpctransport/grpctransport.go
```diff
@@ -0,0 +1,279 @@
...
+		tOpts.DefaultEndpoint = io.DefaultEndpoint
+		tOpts.DefaultMTLSEndpoint = io.DefaultMTLSEndpoint
+	}
+	transportCreds, endpoint, err := transport.GetGRPCTransportCredsAndEndpoint(tOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	if !secure {
+		transportCreds = grpcinsecure.NewCredentials()
```
quartzmo, line 192: Would it make sense to do this before the call to `GetGRPCTransportCredsAndEndpoint`?

auth/grpctransport/grpctransport.go
codyoss: We default to 1, see code 8 lines up.

quartzmo: You mean this line? 

```
if opts.PoolSize == 0 || opts.PoolSize == 1 {
```

What if `opts.PoolSize` is `-1` for example? It's a signed int, isn't it?

auth/grpctransport/grpctransport.go
quartzmo: Great, thanks.

auth/grpctransport/grpctransport.go
codyoss: ah, I misread your comment you are correct. Do you think we should validate it or just default to 1 in these cases? Either seems fine to me. I think the original impl would have returned an empty pool that never works, but I don't think that is desired.

quartzmo: I would just default to 1, and change the docs to say: "If unset or less than 1, the default is 1."

codyoss: done

quartzmo (approved):

PR #8631, opened 2023-09-29
feat(logging): add Cloud Run job monitored resource

### LINKED ISSUE #7150: logging: No Cloud Run Jobs environment detection
> **Note:** issue originally filed by @guillaumeblaquiere [here](https://github.com/GoogleCloudPlatform/golang-samples/issues/2751).
>
> cc @codyoss 

-----------------------------

#### Environment details

  - OS type and version: Cloud Run Jobs
  - Go version: 1.15 and 1.17
  - `google-cloud-logging` version: 1.4.2

#### Steps to reproduce

  1. Use this code for a working example in Cloud Run
Main.go
```
package main

import (
	"context"
	"log"
	"cloud.google.com/go/logging"
)

var logger *log.Logger

func main() {
	ctx := context.Background()

	// Sets your Google Cloud Platform project ID.
	projectID := "gdglyon-cloudrun"

	// Creates a client.
	client, err := logging.NewClient(ctx, projectID)
	if err != nil {
	}
	defer client.Close()

	// Sets the name of the log to write to.
	logName := "my-log"

	logger = client.Logger(logName).StandardLogger(logging.Warning)

	// Logs "hello world", log entry is visible at
	// Cloud Logs.
	logWarning()
}

func logWarning() {
	logger.Println("hello world")
}
```
go.mod
```
module golang.com/gblaquiere

go 1.15

require cloud.google.com/go/logging v1.4.2
```
Dockerfile
```
FROM golang:1.17 AS builder

# Set necessary environmet variables needed for our image
ENV GO111MODULE=on \
    CGO_ENABLED=0 \
    GOOS=linux \
    GOARCH=amd64

# Move to working directory /build
WORKDIR /build

# Copy and download dependency using go mod
COPY go.mod .
COPY go.sum .
RUN go mod download

# Copy the code into the container
COPY . .

# Build the application
RUN go build -o main .

FROM debian:buster-slim
RUN set -x && apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
    ca-certificates && \
    rm -rf /var/lib/apt/lists/*

COPY --from=builder /build/main /

ENTRYPOINT ["/main"]
```

Deploy on cloud run. It failed (no webserver deployed) but the warning trace is present in the Cloud Run logs section with that `label` part

```
resource: {
labels: {
configuration_name: "log-format-go"
location: "us-central1"
project_id: "gdglyon-cloudrun"
revision_name: "log-format-go-00003-tuj"
service_name: "log-format-go"
}
type: "cloud_run_revision"
}
severity: "WARNING"
```
--------
Now, do the same thing with Cloud Run Jobs
Same container, same code

This time the logs are present but not visible in the Cloud Run Jobs logs section. Why? Because the label is not correct:

```
resource: {
labels: {
project_id: "gdglyon-cloudrun"
}
type: "project"
}
severity: "WARNING"
```

The Cloud Logging Library doesn't correctly detect the Cloud Run Jobs runtime environment and don't put the correct values in the logs that cause a filtering issue in the Cloud Run Jobs UI.

(Cloud Run team is aware of that issue)

Note: same issue in Python library: https://github.com/googleapis/python-logging/issues/663

aaron-lerner:
Env vars come from https://cloud.google.com/run/docs/container-contract#jobs-env-vars.

Fixes #7150


logging/resource.go
```diff
@@ -126,15 +126,23 @@ func detectCloudFunction() *mrpb.MonitoredResource {
 	}
 }
 
-func (r *resource) isCloudRun() bool {
+func (r *resource) isCloudRunService() bool {
```
daniel-sanche, line 129: This seems to be a public function, right? If so, we shouldn't break the signature unless absolutely necessary

logging/resource.go
```diff
@@ -126,15 +126,23 @@ func detectCloudFunction() *mrpb.MonitoredResource {
...
 	return config != "" && service != "" && revision != ""
 }
 
-func detectCloudRunResource() *mrpb.MonitoredResource {
+func (r *resource) isCloudRunJob() bool {
+	job := r.attrs.EnvVar("CLOUD_RUN_JOB")
+	execution := r.attrs.EnvVar("CLOUD_RUN_EXECUTION")
+	taskIndex := r.attrs.EnvVar("CLOUD_RUN_TASK_INDEX")
+	taskAttempt := r.attrs.EnvVar("CLOUD_RUN_TASK_ATTEMPT")
+	return job != "" && execution != "" && taskIndex != "" && taskAttempt != ""
```
daniel-sanche, line 142: nit: not too important, but it would be nice if we could short-circuit if the first one fails, instead of always loading each value

daniel-sanche (changes requested):

logging/resource.go
aaron-lerner: I thought in go only functions with names starting with capital letters were public? I'm guessing I either have that wrong (not very familiar with the language) or am misunderstanding what you mean by "public."

Assuming I revert this, are you saying I should keep isCloudRun() alone and have it only identify services and add isCloudRunJob() as a new method identifying only jobs?

gkevinzheng: > This seems to be a public function, right? If so, we shouldn't break the signature unless absolutely necessary

I don't think this is a public function. Additionally, these boolean functions that check if a resource is an App Engine/Cloud Run/Cloud Function resource are exclusively used within `resource.go` for the `detectResource` function.

gkevinzheng (approved):

daniel-sanche (commented):
LGTM

daniel-sanche (approved):

Issue #8632, opened 2023-10-02
profiler: add Start function with a context to abort profiling


timofurrer:
**Is your feature request related to a problem? Please describe.**

I want to be able to stop the profiling and restart it - possibly with another config even.
However, the current API with `profiler.Start()` doesn't support that.

**Describe the solution you'd like**

Possibly a `ctx` argument for the `Start()` function, however, this is likely not possible due to backwards compatibility, therefore a new function could be added:

```go
// Start starts a goroutine to collect and upload profiles. The
// caller must provide the service string in the config. See
// Config for details. Start should only be called once. Any
// additional calls will be ignored.
func Start(cfg Config, options ...option.ClientOption) error {
	startError := startOnce.do(func() error {
		return start(context.Background(), cfg, options...)
	})
	return startError
}

// StartWithContext starts a goroutine to collect and upload profiles. The
// caller must provide the service string in the config. See
// Config for details. StartWithContext should only be called once. Any
// additional calls will be ignored.
// When the context is cancalled the profiling is aborted
func StartWithContext(ctx context.Context, cfg Config, options ...option.ClientOption) error {
	startError := startOnce.do(func() error {
		return start(ctx, cfg, options...)
	})
	return startError
}

func start(ctx context.Context, cfg Config, options ...option.ClientOption) error {
	...
}
```

**Describe alternatives you've considered**

- use a `profiler.Config` field with a context - that would avoid adding `StartWithContext`, but is also kinda weird from an API perspective

**Additional context**

I haven't looked at the profiler internals and I'm not sure if it's even possible to simply stop and restart the profiling without leaving any traces of it.

timofurrer:
I've created a draft PR at https://github.com/googleapis/google-cloud-go/pull/8633

While I was working on it I realized that there are more "problems" - e.g. some state is globally set, like `config`, but obviously isn't easily reset when the context is cancelled. 
There would be ways around this, like the `StartWithContext()` could return a stop function that would wrap the context cancellation and reset global state.
On the other hand, it's probably also not really an issue, because when starting again the state is overridden - and starting multiple profilers is anyways not a thing you'd want.

@aabmass since you got assigned to this - WDYT?

timofurrer:
> Do you mind expanding on the use case a bit?

[...]

PR #8633, opened 2023-10-02
profiler: support ctx cancellation to stop the profiler.

timofurrer:
Closes https://github.com/googleapis/google-cloud-go/issues/8632

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -224,12 +224,22 @@ func (o *allowUntilSuccess) do(f func() error) (err error) {
 // additional calls will be ignored.
 func Start(cfg Config, options ...option.ClientOption) error {
 	startError := startOnce.do(func() error {
-		return start(cfg, options...)
+		return start(context.Background(), cfg, options...)
 	})
 	return startError
 }
 
-func start(cfg Config, options ...option.ClientOption) error {
+// StartWithContext starts a goroutine to collect and upload profiles. The
+// caller must provide the service string in the config. See
+// Config for details.
+// Before calling StartWithContext again, make sure to cancel the previous call
+// using the provided ctx.
+// Profiling is stopped when ctx is canceled
+func StartWithContext(ctx context.Context, cfg Config, options ...option.ClientOption) error {
+	return start(ctx, cfg, options...)
+}
+
+func start(ctx context.Context, cfg Config, options ...option.ClientOption) error {
 	if cfg.DebugLoggingOutput == nil {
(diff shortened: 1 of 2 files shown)


profiler/profiler.go
```diff
@@ -224,12 +224,24 @@ func (o *allowUntilSuccess) do(f func() error) (err error) {
 // additional calls will be ignored.
 func Start(cfg Config, options ...option.ClientOption) error {
 	startError := startOnce.do(func() error {
```
aabmass, line 226: If the intention is that someone would re-start the profiler, `startOnce` would prevent that

profiler/profiler.go
```diff
@@ -55,7 +55,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/google/pprof/profile"
-	gax "github.com/googleapis/gax-go/v2"
+	"github.com/googleapis/gax-go/v2"
```
aabmass, line 58: Can you revert this line? I think the alias was intentional for readability

profiler/profiler.go
timofurrer: Oh, absolutely. 

I've also added a test for this. Should we implement anything to:

- prevent a double start?
- notify the caller when the profiler actually has stopped (something like the `profilingDone` channel)?

timofurrer: I wonder if we should change the `StartWithContext` API and let it return a stop function. I was thinking about something like this (untested and scratched here in the comment box):


```go
func StartWithContext(ctx context.Context, cfg Config, options ...option.ClientOption) (error, func()) {
	ctx, cancel := context.WithCancel(ctx)
	return start(ctx, cfg, options...), func() {
		cancel()
		<-profilingDone
	}
}
```

timofurrer: ... of course we could enhance it with a named func type and timeout for waiting to read from the channel etc.

timofurrer: ... maybe the context should be decoupled though so that only the returned stop function can cancel the profiling, so that the user always has to await for the profiling to be finished. The user then would need to forward ctx cancellation ... 

WDYT?

codyoss:
Closing as this PR was never opened.

timofurrer:
@codyoss can you elaborate on what "this PR was never opened" means?

We didn't move on with it because of https://github.com/googleapis/google-cloud-go/issues/8632#issuecomment-1747332459. Did anything chance and you may be accepting contributions again?

PR #8642, opened 2023-10-04
profiler: refactor to avoid (most of the) global state

timofurrer:
This change set refactors the profiler to reduce the global state
drastically. It tries to keep the changes minimal on purpose and with
that many opportunities to simplify the code have been left out.
That also includes logically reordering parts of the code to have better 
readability. It can all easily be done in a follow up change.

This change set also prepares the profiler to allow for a stop API,
that is once the profiler has been started it can be stopped again and
restarted - even with a different config.

Refs https://github.com/googleapis/google-cloud-go/issues/8632

### DIFF
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -201,8 +201,8 @@ type allowUntilSuccess struct {
 	done uint32
 }
 
-// do calls function f only if it hasnt returned nil previously.
-// Once f returns nil, do will not call function f any more.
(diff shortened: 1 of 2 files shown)


codyoss:
Closing as this PR was never opened.

timofurrer:
@codyoss same question as in https://github.com/googleapis/google-cloud-go/pull/8633#issuecomment-2249510493

PR #8643, opened 2023-10-04
fix(internal/postprocessor): add ability to override release level

codyoss:
For stable client without services there is not a good way to infer the release level. Eventually it may be nice if this info is pushed into the service config rather than a blaze rule so we can extract it better. For now, lets a a manual override.

In the past for things like oslogin we just did not provide a service config entry. Although this would work here it is not ideal as there is still valuable metadata and configuration options in these files.


quartzmo (approved):
Sweet! Thank you!

PR #8652, opened 2023-10-05
feat(spanner): BatchWrite

sunnsing-google:
Implementation for [BatchWrite API](https://github.com/googleapis/googleapis/blob/aa829ea9860cdc474fccdf8f181f13d2a0eb2718/google/spanner/v1/spanner.proto#L307).


sunnsing-google:
@harshachinta - Can you please review?

spanner/mutation.go
```diff
@@ -141,6 +141,12 @@ type Mutation struct {
 	values []interface{}
 }
 
+// A MutationGroup is a list of mutations to be committed atomically.
```
harshachinta, line 144:
```suggestion
// A MutationGroup is a list of Mutation to be committed atomically.
```

spanner/client.go
```diff
@@ -669,6 +672,193 @@ func (c *Client) Apply(ctx context.Context, ms []*Mutation, opts ...ApplyOption)
...
+// are streamed into the response as the batches are applied.
+//
+// BatchWrite requests are not replay protected, meaning that each mutation
+// group may be applied more than once. Replays of non-idempotent mutations
+// may have undesirable effects. For example, replays of an insert mutation
+// may produce an already exists error or if you use generated or commit
+// timestamp-based keys, it may result in additional rows being added to the
+// mutation's table. We recommend structuring your mutation groups to be
+// idempotent to avoid this issue.
+func (c *Client) BatchWrite(ctx context.Context, mgs []*MutationGroup, opts ...ApplyOption) *BatchWriteResponseIterator {
```
harshachinta, line 789: `ApplyOption` would not be suitable here since it is an argument tailored to `Apply` method. Additionally, when new options need to be introduced for BatchWrite in future it will add the change to `ApplyOptions` which might not be valid for `Apply` method and might confuse users.
IMO, introducing new `BatchWriteOption` would be more appropriate here.

I am trying to explore [Go generics](https://go.dev/doc/tutorial/generics) to avoid duplicate code for `ApplyOption` and `BatchWriteOption`, but the solution seems to be tricky. 
Please let me know if you have some better way to do this.

sunnsing-google: @harshachinta - Thanks for the feedback. This makes sense.

How about we expose a new struct `BatchWriteOptions`, similar to `QueryOptions` -
```go
struct BatchWriteOptions {
  Priority sppb.RequestOptions_Priority,
  TransactionTag String,
  // future options go here
}
```

The interface would then become - 
```go
func (c *Client) BatchWrite(ctx context.Context, mgs []*MutationGroup, opts BatchWriteOptions) *BatchWriteResponseIterator {}
```

The `Client` struct would contain `bwo` as a field and there'd be a new `func (bwo BatchWriteOptions) merge(opts BatchWriteOptions) BatchWriteOptions {}` function.

If the above looks good to you, I'll make this change.

**Aside (not related to this PR) -**
All of the different `options` constructs seem to contain fields from the more generic `sppb.RequestOptions` struct along with other request specific options. I suppose, `sppb.RequestOptions` could be modelled directly and the specific request options structs could contain this generic options struct.

harshachinta: That looks like one possible option. To collate, there are 2 ways to go about this,
1. As mentioned above, introduce
```
struct BatchWriteOptions {
  Priority sppb.RequestOptions_Priority
  TransactionTag String
  // future options go here
}
```
which will have the following API definitions,
```
1. func (c *Client) BatchWrite(ctx context.Context, mgs []*MutationGroup) *BatchWriteResponseIterator {}
2. func (c *Client) BatchWriteWithOptions(ctx context.Context, mgs []*MutationGroup, opts BatchWriteOptions) *BatchWriteResponseIterator {}
```

2. Second way is to do the following,
Add a new type `APIOption` that can be used across all new APIs
```
type APIOption func(T interface{})

// Example of one option that customer can use
func TransactionTagOption(tag string) APIOption {
	return func(o interface{}) {
		switch o.(type) {
		case *batchWriteOption:
			o.(*batchWriteOption).transactionTag = tag
                // can be extended to new API in future
		}
	}
}
```
and create a `batchWriteOption` type and use it inside the `BatchWrite` method.
```
type batchWriteOption struct {
	transactionTag string
	priority sppb.RequestOptions_Priority
}

func (c *Client) BatchWrite(ctx context.Context, mgs []*MutationGroup, opts ...APIOption) *BatchWriteResponseIterator {
	bo := &batchWriteOption{}
	for _, opt := range opts {
		opt(bo)
	}
}
```
This way any new APIs/methods in future can have its options under APIOption.

@olavloite @rahul2393 Can you please suggest if any of these two ways would be good or suggest any other possible way?

sunnsing-google: @harshachinta - Went ahead and implemented option 1 in 5b9ce8a11ca9efc3976bb5ac579c132daf749c7d. PTAL if/when there's agreement on the approach.

spanner/client.go
```diff
@@ -669,6 +677,217 @@ func (c *Client) Apply(ctx context.Context, ms []*Mutation, opts ...ApplyOption)
...
+
+	opts = c.bwo.merge(opts)
+
+	mgsPb, err := mutationGroupsProto(mgs)
+	if err != nil {
+		return &BatchWriteResponseIterator{err: err}
+	}
+
+	var sh *sessionHandle
+	sh, err = c.idleSessions.take(ctx)
```
harshachinta, line 840: We are not using this session until `Next()` method is invoked on the `BatchWriteRowIterator`. Should we instead take a session from the pool inside the `rpc` function? This way we will not end up holding a resource that is not used until Next is invoked.

spanner/client.go
harshachinta: Option 1 seems to be a good choice here. Lets go ahead with that approach.

spanner/client.go
sunnsing-google: In typical client usage, a call to `BatchWrite()` would be followed immediately by `Do()/Next()` calls on the returned iterator - there won't be a delay. The current approach handles session management tasks outside of the `rpc()`, which makes `rpc()` do one thing.

`transaction.ReadWithOptions()` follows the same approach of acquiring a session first and later invoking `rpc()` when `Next()` is called.

I think the current state is okay. We can always refactor it later if need be.

harshachinta: 1. In worst case customers could possibly introduce delay between calling `BatchWrite` and then reading using `Next()` method.
2. The `Apply()` method that is used to apply mutations also handles the session logic inside the rpc method.
Reference: https://github.com/googleapis/google-cloud-go/blob/81a97b06cb28b25432e4ece595c55a9857e960b7/spanner/transaction.go#L1700-L1716

spanner/client.go
```diff
@@ -669,6 +677,217 @@ func (c *Client) Apply(ctx context.Context, ms []*Mutation, opts ...ApplyOption)
...
+		merged.TransactionTag = opts.TransactionTag
+	}
+	if opts.Priority != sppb.RequestOptions_PRIORITY_UNSPECIFIED {
+		merged.Priority = opts.Priority
+	}
+	return merged
+}
+
+// BatchWriteResponseIterator is an iterator over BatchWriteResponse structures returned from BatchWrite RPC.
+type BatchWriteResponseIterator struct {
```
harshachinta, line 706: nit (can ignore): Is it better to move BatchWriteResponseIterator and its methods to separate file something like `iterator.go` to improve readability? I understand existing `RowIterator` lies in `read.go`, but just thought that as we are adding this newly it would be a starting point to have all new iterators in one common file.

harshachinta (approved):
LGTM.

spanner/client_test.go
```diff
@@ -4304,3 +4305,250 @@ func TestClient_CustomRetryAndTimeoutSettings(t *testing.T) {
...
+
+	server, client, teardown := setupMockedTestServer(t)
+	defer teardown()
+	mutationGroups := []*MutationGroup{
+		{[]*Mutation{
+			{opInsertOrUpdate, "t_test", nil, []string{"key", "val"}, []interface{}{"foo1", 1}},
+		}},
+	}
+	iter := client.BatchWrite(context.Background(), mutationGroups)
+	responseCount := int(0)
```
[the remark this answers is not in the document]
sunnsing-google: Good catch. Removed in e9feab5ab43a58e65f0ce568dd83b0d9d8aa2db9.

spanner/client.go
sunnsing-google: I think this is better left as a refactoring task for later since the public interface will remain the same.

rahul2393 (approved):

PR #8653, opened 2023-10-05
feat(bigquery): introduce query preview features

shollyman:
This PR introduces the ability for users of the bigquery client library to indicate the library should enable preview features.

The first preview feature enabled here is the stateless query feature, a mechanism for running small queries without corresponding job metadata.  Ability to use this feature is governed by the service itself, and will simply create jobs in cases where the feature isn't enable in either the client or service, or a query cannot be satisfied in a stateless manner.


bigquery/integration_test.go
```diff
@@ -806,6 +806,12 @@ func TestIntegration_SimpleRowResults(t *testing.T) {
 			query:       "select count(*) from unnest(generate_array(1,1000000)), unnest(generate_array(1, 1000)) as foo",
 			want:        [][]Value{{int64(1000000000)}},
 		},
+		{
+			// Query doesn't yield a result.
+			description: "DML",
+			query:       fmt.Sprintf("CREATE OR REPLACE TABLE %s.%s (foo STRING, bar INT64)", dataset.DatasetID, tableIDs.New()),
+			want:        [][]Value{},
+		},
```
alvarowolfx, line 814: Should we run the same test cases with the query preview enabled ?

alvarowolfx (approved):
Added a minor comment about integration test, but looks good to me. Took me a while to realize that we don't need to fiddle with the QueryID to fetch results as we already handle cached results.

bigquery/integration_test.go
shollyman: Good point.  I've been doing it out of band but I can swap it.

shollyman: And testing was a wise recommendation, helped me discover an issue where the stateless result is empty.  Previously we'd fallback, but there's no fallback available.

PR #8654, opened 2023-10-05
feat(auth): add Credentials.UniverseDomain()

quartzmo:
Read and expose universe_domain from self-signed service account JSON

Equivalent to:

* https://go-review.git.corp.google.com/c/oauth2/+/531715
* https://go-review.git.corp.google.com/c/oauth2/+/532196


codyoss (approved):
LGTM

Issue #2205, opened 2020-05-20
trace: Support for opentelemetry tracing


cmouli84:
**Is your feature request related to a problem? Please describe.**
I am instrumenting trace using opentelemetry library in application that uses google cloud resources. Since google cloud library instruments trace using [opencensus](https://github.com/googleapis/google-cloud-go/blob/master/internal/trace/trace.go#L29), I do not see spans for interactions with google cloud resources (e.g. _cloud.google.com/go/spanner.ReadWriteTransaction_)

**Describe the solution you'd like**
[Trace package](https://github.com/googleapis/google-cloud-go/blob/master/internal/trace/trace.go) to be updated to propagate both opencensus and opentelemetry Span in context and export it to registered Exporter/SpanProcessor

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

codyoss:
@cmouli84 thanks for your interest in tracing. This client is auto-generated. If you would like a feature added to it the best place to express that would be https://issuetracker.google.com/issues/new?component=187216&template=0.
Thanks.

PR #8655, opened 2023-10-06
feat(internal/trace): add OpenTelemetry support

quartzmo:
Add GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING env var flag for opt-in OpenTelemetry tracing.

refs: #2205

### OpenCensus sunsetting and OpenTelemetry support

The OpenCensus project is obsolete and was archived on July 31st, 2023. This means that any security vulnerabilities that are found will not be patched. We recommend that you begin migrating to OpenCensus tracing to OpenTelemetry, the successor project.

One year after the Google Cloud Go client libraries release the experimental, opt-in OpenTelemetry tracing support in this PR (mid-Q4 2023), the prior experimental OpenCensus tracing will be removed.

Using the [OpenTelemetry-Go - OpenCensus Bridge](https://github.com/open-telemetry/opentelemetry-go/tree/main/bridge/opencensus), you can immediately begin exporting your traces with OpenTelemetry, even while dependencies remain instrumented with OpenCensus. If you do not use the bridge, you will need to migrate your entire application and all of its instrumented dependencies at once.  For simple applications, this may be possible, but we expect the bridge to be helpful if multiple libraries with instrumentation are used.

Please refer to the following resources:

* [Sunsetting OpenCensus](https://opentelemetry.io/blog/2023/sunsetting-opencensus/)
* [OpenTelemetry-Go - OpenCensus Bridge](https://github.com/open-telemetry/opentelemetry-go/tree/main/bridge/opencensus)
* [What is OpenTelemetry?](https://opentelemetry.io/docs/what-is-opentelemetry/)
* [Cloud Trace - Go and OpenTelemetry](https://cloud.google.com/trace/docs/setup/go-ot)
* On GCE, use [Ops Agent and OpenTelemetry](https://cloud.google.com/trace/docs/otlp)

### Testing support for library maintainers

PTAL at `internal/testutil/trace_otel.go` and the update to `trace_test.go`. I wasn't able to match the existing `testutil` api perfectly, so testing support for OpenTelemetry will not be completely transparent. However, the new `testutil` functions are very close:

### OpenCensus

```go
te := testutil.NewTestExporter()
defer te.Unregister()
...
spans := te.Spans
```

### OpenTelemetry

```go
ctx := context.Background()
te := testutil.NewOpenTelemetryTestExporter()
defer te.Unregister(ctx)
...
spans := te.Spans()
```


internal/trace/trace.go
```diff
@@ -16,35 +16,74 @@ package trace
...
+var (
+	// TODO(chrisdsmith): What is the correct name for the OpenTelemetry tracer?
+	OpenTelemetryTracerName string = "cloud.google.com/go/internal/trace"
+)
+
+func IsOpenCensusTracingEnabled() bool {
+	return !IsOpenTelemetryTracingEnabled()
+}
+
+func IsOpenTelemetryTracingEnabled() bool {
```
shollyman, line 50: Can we reduce this to a singleton check somehow?  As written this means we're going to be interrogating the environment variables on every span start/end, but something like a sync.once or similar could do the env check once and retain the results of the lookup.

quartzmo: Yes, great point. Not sure how I missed this, really appreciate the comment.

internal/trace/trace_test.go
```diff
@@ -15,18 +15,154 @@
...
+var (
+	ignoreEventFields = cmpopts.IgnoreFields(sdktrace.Event{}, "Time")
+	ignoreValueFields = cmpopts.IgnoreFields(attribute.Value{}, "vtype", "numeric", "stringly", "slice")
+)
+
+func TestStartSpan_OpenCensus(t *testing.T) {
+	old := os.Getenv(telemetryPlatformTracing)
+	defer os.Setenv(telemetryPlatformTracing, old)
+	os.Setenv(telemetryPlatformTracing, "any value")
+	te := testutil.NewTestExporter()
```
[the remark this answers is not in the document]
quartzmo: @alvarowolfx Thanks for catching this. Great suggestion, will do.

internal/trace/trace.go
quartzmo: Updated, PTAL.

internal/trace/trace_test.go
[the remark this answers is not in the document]
quartzmo: @alvarowolfx PTAL at `internal/testutil/trace_otel.go` and the update to `trace_test.go`. I wasn't able to match the existing `testutil` api perfectly, so transparent conditional support is not going to work. But the new `testutil

internal/trace/trace_test.go
```diff
@@ -51,3 +184,110 @@ func TestToStatus(t *testing.T) {
...
+			t.Fatalf("apierror.FromError failed to parse %v", testcase.input)
+		}
+		got := toStatus(err)
+		if r := testutil.Diff(got, testcase.want); r != "" {
+			t.Errorf("got -, want +:\n%s", r)
+		}
+	}
+}
+
+func attrMap() map[string]interface{} {
```
shollyman, line 285: minor nit:  maybe give this a name that makes it clearer its test/example data?  Wasn't clear when I encountered it in the test code until I sought out the function implementation.

internal/trace/trace.go
shollyman: Thanks for making this change!

internal/trace/trace_test.go
quartzmo: Updated, thanks!

internal/trace/trace.go
```diff
@@ -16,35 +16,95 @@ package trace
...
-// StartSpan adds a span to the trace with the given name.
+const (
+	telemetryPlatformTracingOpenCensus    = "opencensus"
+	telemetryPlatformTracingOpenTelemetry = "opentelemetry"
+	telemetryPlatformTracingVar           = "GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING"
+)
+
+var (
+	// TODO(chrisdsmith): What is the correct name for the OpenTelemetry tracer?
+	openTelemetryTracerName     string = "cloud.google.com/go/internal/trace"
```
codyoss, line 43: maybe just drop the internal/trace part?

internal/trace/trace.go
```diff
@@ -53,6 +113,18 @@ func toStatus(err error) trace.Status {
 	}
 }
 
+// toOpenTelemetryStatus converts an error to an equivalent OpenTelemetry status description.
+func toOpenTelemetryStatusDescription(err error) string {
+	var err2 *googleapi.Error
```
codyoss, line 118: Should this just be an APIError now? Everything should return those now, right? Same with the above toStatus stuff

internal/trace/trace.go
```diff
@@ -107,5 +202,34 @@ func TracePrintf(ctx context.Context, attrMap map[string]interface{}, format str
...
+
+// otAttrs converts a generic map to OpenTelemetry attributes.
+func otAttrs(attrMap map[string]interface{}) []attribute.KeyValue {
+	var attrs []attribute.KeyValue
+	// Sort the input map by its keys for predictable order in tests.
+	keys := make([]string, 0, len(attrMap))
+	for k := range attrMap {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
codyoss, line 216: Can we just move the sorting to the test as that is where it is needed for. Don't want to have a runtime cost for non-test code.

internal/trace/trace_test.go
```diff
@@ -15,18 +15,151 @@
...
 )
 
+var (
+	ignoreEventFields = cmpopts.IgnoreFields(sdktrace.Event{}, "Time")
+	ignoreValueFields = cmpopts.IgnoreFields(attribute.Value{}, "vtype", "numeric", "stringly", "slice")
+)
+
+func TestStartSpan_OpenCensus(t *testing.T) {
+	old := openTelemetryTracingEnabled
+	defer func() {
```
codyoss, line 44: t.Cleanup may be nice here

internal/trace/trace.go
quartzmo: I'll look into this, thanks!

internal/trace/trace.go
quartzmo: Done, thanks!

internal/trace/trace.go
quartzmo: Done, thanks!

internal/trace/trace_test.go
quartzmo: Good call, done, thanks!

bhshkh:
Sorry, marked as draft by mistake

internal/trace/trace.go
quartzmo: I think this logic in bigtable shows a situation where the `err` closing the span could be an unwrapped `Status` or even a `io.EOF`:

https://github.com/googleapis/google-cloud-go/blob/bigtable/v1.20.0/bigtable/bigtable.go#L182-L215

codyoss: sure, but I just meant L117 here. Other cases can still be left.

quartzmo: ah sorry, i misunderstood

quartzmo: After trying this, I think it's best to just extract the `googleapi.Error` (if found), since it exposes the HTTP error message. With APIError, that message is only exposed by `Error()` but it gets some added prefix:

```
 trace_test.go:131: got googleapi: Error 400: INVALID ARGUMENT, want INVALID ARGUMENT
```

codyoss: I think that would be fine to call Error and all HTTP APIError *should have wrap a googleapis.Error. But I am fine with this for now. We can always make that change later.

codyoss (approved):

quartzmo:
I tested this manually using only an OpenTelemetry exporter and the `cloud.google.com/go/storage` client, and observed the storage span in Cloud Trace console when `GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING=opentelemetry`.

/cc @cjcotter

PR #8657, opened 2023-10-08
fix(bigtable): return cluster error for Update when populated

lucetre:
### Problem
In our usage of `bigtableAdminClient.UpdateInstanceWithClusters`, we encountered a situation where it did not report cluster errors, despite `bigtableAdminClient.SetAutoscaling` correctly returning an error.

For instance, we attempted to update a Bigtable instance with a node count of 3 using a desired autoscaling configuration specifying a minimum of 1 and a maximum of 2 nodes. It's important to note that you cannot set the maximum serving nodes to a value less than the current cluster size. However, when updating cluster configurations, we did not receive an error message.

### Goal
This PR addresses this issue by improving the returned values to provide the relevant error message, as shown below:

```
Error found while updating instance with autoscaling config ({MinNodes:1 MaxNodes:2 CPUTargetPercent:50 StorageUtilizationPerNode:0}): rpc error: code = InvalidArgument desc = max_serve_nodes (2) must be equal to or greater than current cluster size (3)
```

### Related Link
- https://github.com/googleapis/google-cloud-go/issues/8665
- https://cloudlogging.app.goo.gl/gciQxq8BRA6sN5B86
- https://mlc.atlassian.net/browse/SRE2-2056


noahdietz:
Thanks for the fix @lucetre. I will update your branch and run presubmits. This looks straight forward enough.

noahdietz:
The conformance test was recently added and it was added in a broken state so this is a false positive.

PR #8658, opened 2023-10-09
test(bigtable): Run conformance tests for bigtable

bhshkh:
This PR adds conformance tests to be run on presubmit and merge.
For Bigtable, testproxy is started in background and conformance tests are executed against it


bigtable/conformance_test.sh
```diff
@@ -0,0 +1,87 @@
...
+    unset BIGTABLE_CONFORMANCE_TESTS_HOME;
+}
+trap cleanup EXIT
+
+# Checkout Bigtable conformance tests 
+mkdir -p $BIGTABLE_CONFORMANCE_TESTS_HOME
+git clone https://github.com/googleapis/cloud-bigtable-clients-test.git $BIGTABLE_CONFORMANCE_TESTS_HOME
+
+pushd $BIGTABLE_CONFORMANCE_TESTS_HOME > /dev/null;
+    cd tests
```
daniel-sanche, line 72: nit: why cd right after pushd? Maybe combine to same line?

internal/kokoro/presubmit.sh
```diff
@@ -65,6 +65,12 @@ runPresubmitTests() {
   if [ -f "emulator_test.sh" ]; then
     ./emulator_test.sh
   fi
+
+  # Run conformance tests against a test proxy.
+  if [ -f "conformance_test.sh" ]; then
+    ./conformance_test.sh
+  fi
```
daniel-sanche, line 72: I'm a bit confused about how exit codes are propagated here. `continuous.sh` seems to add together the exit_code after each sub-test, but presubmit seems to work differently.

Are you supposed to add ` | tee sponge_log.log` here so the error can be caught by `go-junit-report` in the line below? Or am I misunderstanding this section?

triplequark: The call to conformance_test.sh populates sponge.log if I'm reading this correctly - teeing here would duplicate output.

bigtable/conformance_test.sh
bhshkh: Done

internal/kokoro/presubmit.sh
bhshkh: Right

bigtable/conformance_test.sh
```diff
@@ -0,0 +1,85 @@
...
+    unset BIGTABLE_CLIENT_LIB_HOME;
+    unset BIGTABLE_TEST_PROXY_PORT;
+    unset BIGTABLE_TEST_PROXY_HOME;
+    unset BIGTABLE_CONFORMANCE_TESTS_HOME;
+}
+trap cleanup EXIT
+
+# Checkout Bigtable conformance tests 
+mkdir -p $BIGTABLE_CONFORMANCE_TESTS_HOME
+git clone https://github.com/googleapis/cloud-bigtable-clients-test.git $BIGTABLE_CONFORMANCE_TESTS_HOME
```
danieljbruce, line 69: Other PRs do this in the Yaml file like https://github.com/googleapis/nodejs-bigtable/pull/1349/files#diff-7bfd467904d753bc34e8f91ffbb8165dc048bbe3610af168c600072d0ec473c0R31-R34 and https://github.com/googleapis/java-bigtable/pull/1699/files#diff-7bfd467904d753bc34e8f91ffbb8165dc048bbe3610af168c600072d0ec473c0R29-R31. I notice this PR doesn't add a Github action. Is there any reason not adding a conformance test Github action was taken for this PR?

image

bhshkh: Earlier, I thought that adding a new github action would require extensive kokoro config similar to presubmit and continuous but after checking with Cody, I have learnt that it is not the case.

Now, I have added a separate github action.

bigtable/conformance_test.sh
```diff
@@ -0,0 +1,59 @@
...
+    echo "Cleanup testproxy and move logs"
+    kill -9 $proxyPID
+    # Takes the kokoro output log (raw stdout) and creates a machine-parseable
+    # xUnit XML file.
+    cat $sponge_log |
+      go-junit-report -set-exit-code >$clientLibHome/sponge_log.xml
+}
+trap cleanup EXIT
+
+pushd $conformanceTestsHome > /dev/null;
```
danieljbruce, line 52: Probably doesn't matter too much, but we can replace the pushd and popd with a cd like https://github.com/googleapis/nodejs-bigtable/pull/1349#discussion_r1348017520 suggests.

.github/workflows/conformance.yaml
```diff
@@ -0,0 +1,44 @@
...
+      with:
+        path: google-cloud-go
+    - uses: actions/checkout@v3
+      with:
+        repository: googleapis/cloud-bigtable-clients-test
+        ref: main
+        path: cloud-bigtable-clients-test
+    - uses: actions/setup-go@v4
+      with:
+        go-version: '>=1.20.2'
```
danieljbruce, line 40: Should we run this for multiple golang versions like in other clients?

https://github.com/googleapis/nodejs-bigtable/pull/1349/files#diff-7bfd467904d753bc34e8f91ffbb8165dc048bbe3610af168c600072d0ec473c0R25-R27

bigtable/conformance_test.sh
danieljbruce: Either way works IMO. pushd and popd or just one cd line.

daniel-sanche (approved):
LGTM

.github/workflows/conformance.yaml
bhshkh: The testproxy can be run 1.19, 1.20, 1.21 but the conformance tests themselves can be run only on 1.20.2
Modified accordingly.

bigtable/conformance_test.sh
bhshkh: Modified

bigtable/conformance_test.sh
```diff
@@ -49,11 +48,12 @@ function cleanup() {
...
 
-pushd $conformanceTestsHome > /dev/null;
-    # Run the conformance tests
-    go test -v -proxy_addr=:$testProxyPort | tee -a $sponge_log
-    RETURN_CODE=$?
-popd > /dev/null;
+# Run the conformance tests
+cd $conformanceTestsHome
+go install golang.org/dl/go1.20.2@latest
+go1.20.2 download
```
danieljbruce, line 54: Should these 2 lines above be removed?

bigtable/conformance_test.sh
```diff
@@ -49,11 +48,12 @@ function cleanup() {
...
+go1.20.2 test -v -proxy_addr=:$testProxyPort | tee -a $sponge_log
```
danieljbruce, line 55: Should this be done with the golang version that is currently selected in the strategy matrix?

danieljbruce (commented):
Just a couple of questions regarding Golang versions.

.github/workflows/conformance.yaml
danieljbruce: Got it. Thanks for the explanation.

danieljbruce (approved):

bigtable/conformance_test.sh
bhshkh: Conformance tests can only be run on Go 1.20.2. This is very specific to Bigtable. So, installation is being done here and not the common conformance.yaml

bigtable/conformance_test.sh
bhshkh: No.
This is running the conformance tests. Conformance tests can only be run on Go 1.20.2

Strategy matrix is for testproxy versions. https://github.com/googleapis/google-cloud-go/blob/main/README.md?plain=1#L31-L36

triplequark:
LGTM

triplequark (approved):

.github/workflows/conformance.yaml
```diff
@@ -0,0 +1,48 @@
...
+        repository: googleapis/cloud-bigtable-clients-test
+        ref: main
+        path: cloud-bigtable-clients-test
+    - uses: actions/setup-go@v4
+      with:
+          go-version: ${{ matrix.go }}
+    - run: go version
+    - run: go install github.com/jstemmer/go-junit-report/v2@latest
+    - run: chmod +x ./google-cloud-go/${{ matrix.folders }}/conformance_test.sh
+    - run: ./google-cloud-go/${{ matrix.folders }}/conformance_test.sh
```
noahdietz, line 48: Nit: please add a new line at  EoF
```suggestion
    - run: ./google-cloud-go/${{ matrix.folders }}/conformance_test.sh

```

bigtable/conformance_test.sh
noahdietz: Nit: Can you add a comment explaining *why* Go 1.20.2 is required? We will want to know for when we inevitably bump the minimum version to Go 1.21 in N years.

bigtable/conformance_test.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# Run the conformance tests
+cd $conformanceTestsHome
+go install golang.org/dl/go1.20.2@latest
+go1.20.2 download
+go1.20.2 test -v -proxy_addr=:$testProxyPort | tee -a $sponge_log
+RETURN_CODE=$?
+
+echo "exiting with ${RETURN_CODE}"
+exit ${RETURN_CODE}
```
noahdietz, line 59: Nit: add new line at EoF
```suggestion
exit ${RETURN_CODE}

```

.github/workflows/conformance.yaml
```diff
@@ -0,0 +1,48 @@
...
+  pull_request:
+    paths:
+      - 'bigtable/**'
+name: conformance
+jobs:
+  conformance:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go: [ '1.19', '1.20', '1.21' ]
```
noahdietz, line 31: Why have a matrix strategy if it can only be run with one Go version? Just install that version specifically in the action and simplify the script code.

danieljbruce: I think the intent here was to have a particular golang version running for the test proxy server https://github.com/googleapis/google-cloud-go/pull/8658/files#diff-25791c2e81a435384b6d9fe6d3386af6ed532964e933261c53fbe00ff7bbc3e3R37 so that we can test the test proxy server on multiple golang versions. Currently, the way this is written is the test runner, which calls the test proxy server, has a fixed Golang version.

In the node client multiple node versions are tested because streaming behavior can vary by node version. I wonder if the same is true for Golang.

noahdietz: > streaming behavior can vary by node version. I wonder if the same is true for Golang.

I'm not sure, but the explanation sounds reasonable. I think it would be useful to have some/all of this detail documented in either a readme for or in comments.

.github/workflows/conformance.yaml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  push:
+    paths:
+      - 'bigtable/**'
+    branches:
+    - main
+  pull_request:
+    paths:
+      - 'bigtable/**'
+name: conformance
```
noahdietz, line 25: Please rename this to something more specific e.g. `bigtable_conformance`. Thanks

bhshkh: Done

.github/workflows/conformance.yaml
bhshkh: Added

bigtable/conformance_test.sh
bhshkh: Added

bigtable/conformance_test.sh
bhshkh: Added

noahdietz (approved):

.github/workflows/conformance.yaml
bhshkh: Added

Issue #8387, opened 2023-08-06
bigquery: add missing_value_interpretations as append option


brachipa:
As describe [here](https://cloud.google.com/bigquery/docs/default-values#write_api) 

>The Storage Write API only populates default values when the [write stream](https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#google.cloud.bigquery.storage.v1.WriteStream) schema is missing a field that is contained in the destination table schema

So when we want to default value we need to make sure the field is missing from the schema, what makes our code ugly, we need to iterate over the schema fields and delete the column with the default value (field_with_default_value)

```go
	md, err := bqClient.Dataset(datasetName).Table(tableName).Metadata(ctx)
	if err != nil {
		...
	}

	schema := md.Schema
	convertedSchema, err := adapt.BQSchemaToStorageTableSchema(schema)
	if err != nil {
		...
	}
	savedInd := -1
	for i, field := range convertedSchema.Fields {
		if field.Name == "field_with_default_value" {
			savedInd = i
			break
		}
	}
	if savedI > 0 {
		convertedSchema.Fields = append(convertedSchema.Fields[:savedInd], convertedSchema.Fields[savedInd+1:]...)
	}
```

Or:

>You can also specify request-level default values in the missing_value_interpretations map within the [AppendRowsRequest message](https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#google.cloud.bigquery.storage.v1.AppendRowsRequest). Each key is the name of a column and its [value](https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#missingvalueinterpretation) indicates how to interpret missing values.
For example, the map {'col1': NULL_VALUE, 'col2': DEFAULT_VALUE} means that all missing values in col1 are interpreted as NULL, and all missing values in col2 are interpreted as the default value set for col2 in the table schema.

Unfortunately,  we have no way to pass this map into the append request!

The solution:

Add new option `WithMissingValueInterpretations` to pass this information

```go
missingValueInterpretation := map[string]storagepb.AppendRowsRequest_MissingValueInterpretation{
"field_with_default_value": storagepb.AppendRowsRequest_DEFAULT_VALUE}

....

result, err = stream.AppendRows(ctx, [][]byte{b}, managedwriter.WithMissingValueInterpretations(missingValueInterpretation))


```

[...]

PR #8686, opened 2023-10-11
feat(bigquery/storage/managedwriter): support default value controls

shollyman:
This PR adds new options to control how missing values are interpreted when writing.

For ManagedStream instantiation, the options are:
* `WithDefaultMissingValueInterpretation` (blanket setting for all columns)
* `WithMissingValueInterpretations` (per-column settings)

To support updates, these are added as AppendOption options:
* `UpdateDefaultMissingValueInterpretation`
* `UpdateMissingValueInterpretations`

Implementation-wise, this PR removes the previous schema-specific
versioner (`descriptorVersion`) and expands the concept to a versioned
AppendRowsRequest template (`versionedTemplate`).  This more general
mechanism allows us to version all settings that manifest as request fields 
in the AppendRowsRequest.

Fixes: https://github.com/googleapis/google-cloud-go/issues/8387


bigquery/storage/managedwriter/options.go
```diff
@@ -234,7 +235,48 @@ func WithTraceID(traceID string) WriterOption {
...
+
+// WithMissingValueInterpretations controls over how missing values are interpreted
+// for individual columns.
+//
+// You must provide a map to indicate how to interpret missing value for some fields. Missing
+// values are fields present in user schema but missing in rows. The key is
+// the field name. The value is the interpretation of missing values for the
+// field.
+//
+// For example, a map {'foo': NULL_VALUE, 'bar': DEFAULT_VALUE} means all
```
quartzmo, line 250: What do you think about an example showing a map in Go with `storagepb.AppendRowsRequest_MissingValueInterpretation` values, either in place of the generic map on this line or in a Go testable example?

bigquery/storage/managedwriter/options.go
```diff
@@ -234,7 +235,48 @@ func WithTraceID(traceID string) WriterOption {
...
+// missing values in field foo are interpreted as NULL, all missing values in
+// field bar are interpreted as the default value of field bar in table
+// schema.
+//
+// If a field is not in this map and has missing values, the missing values
+// in this field are interpreted as NULL.
```
quartzmo, line 256: Maybe mention also: "unless a non-null default was set using `WithDefaultMissingValueInterpretation`" ?

bigquery/storage/managedwriter/options.go
```diff
@@ -278,8 +320,37 @@ type AppendOption func(*pendingWrite)
...
+// UpdateMissingValueInterpretations updates the per-column missing-value intepretations settings,
+// and is retained for subsequent writes.  See the WithMissingValueInterpretations WriterOption for
+// more details.
+func UpdateMissingValueInterpretations(mvi map[string]storagepb.AppendRowsRequest_MissingValueInterpretation) AppendOption {
+	return func(pw *pendingWrite) {
+		prev := pw.reqTmpl
+		if prev == nil {
+			prev = newVersionedTemplate()
+		}
+		pw.reqTmpl = prev.revise(reviseMissingValueInterpretations(mvi))
```
quartzmo, line 340: Why does this require two calls? Why doesn't `newVersionedTemplate` accept `...templateRevisionF` like `revise`?

shollyman: changed it to allowed revise() to recover from a nil start state, which simplifies the options.  We don't know based on the call order if we need to instantiate, which is why passing them into the newVersionedTemplate constructor doesn't get us as much.

bigquery/storage/managedwriter/options.go
shollyman: updated the docstring here

bigquery/storage/managedwriter/options.go
shollyman: inlined a more language-appropriate example of specifying the option

bigquery/storage/managedwriter/options.go
```diff
@@ -234,7 +235,48 @@ func WithTraceID(traceID string) WriterOption {
 // AppendRows calls on the stream.
 func WithSchemaDescriptor(dp *descriptorpb.DescriptorProto) WriterOption {
 	return func(ms *ManagedStream) {
-		ms.curDescVersion = newDescriptorVersion(dp)
+		ms.curTemplate = ms.curTemplate.revise(reviseProtoSchema(dp))
+	}
+}
+
+// WithMissingValueInterpretations controls over how missing values are interpreted
```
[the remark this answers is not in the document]
shollyman: s/over//

alvarowolfx (approved):
added minor nit on missing comment and while I'm not a huge fan of the naming of the `reviseXXX` functions, I don't have a better alternative ( and they are all internal so 🤷 / and this can also be a matter of it not making much sense for a non native english speaking person haha )

bigquery/storage/managedwriter/integration_test.go
```diff
@@ -1262,6 +1264,97 @@ func testSchemaEvolution(ctx context.Context, t *testing.T, mwClient *Client, bq
...
 }
 
+func testDefaultValueHandling(ctx context.Context, t *testing.T, mwClient *Client, bqClient *bigquery.Client, dataset *bigquery.Dataset, opts ...WriterOption) {
+	testTable := dataset.Table(tableIDs.New())
+	if err := testTable.Create(ctx, &bigquery.TableMetadata{Schema: testdata.DefaultValueSchema}); err != nil {
+		t.Fatalf("failed to create test table %s: %v", testTable.FullyQualifiedName(), err)
+	}
+
+	m := &testdata.DefaultValuesPartialSchema{
+		// We only popu
```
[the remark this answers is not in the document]
shollyman: indeed.  completed comment.

PR #8690, opened 2023-10-11
chore(internal/postprocessor): add config for shopping/type

quartzmo:


internal/postprocessor/config.yaml
```diff
@@ -770,6 +770,9 @@ service-configs:
   - input-directory: google/shopping/merchant/inventories/v1beta
     service-config: merchantapi_v1beta.yaml
     import-path: cloud.google.com/go/shopping/merchant/inventories/apiv1beta
+  - input-directory: google/shopping/type
+    service-config:
```
noahdietz, line 774: Is this property necessary if empty?
```suggestion
```

noahdietz (approved):

internal/postprocessor/config.yaml
quartzmo: I don't know, but it is there for `oslogin/common`, so I did the same thing.

Issue #913, opened 2018-03-03
datastore: When loading kinds into structs, ignore items where no struct field exist instead of generating an error


lc-chrisbarton:
We process orders and we sometimes deprecate or are no longer interested in certain fields in an order. This means that during transaction processing we sometimes have entities in datastore which have properties which no longer have corresponding fields in the current version of our order struct. If we rely on the default loading this gives the following error "datastore: cannot load field "some_property_name" into a "some_struct": no such struct field". This behaviour seems odd to us, consider that normal JSON decoding in Go would ignore any fields in the incoming JSON if there was no destination field. The current behaviour means we must write custom loaders which is fine but more work. In any case we need to support deprecating fields in order to keep our API intuitive because that helps our clients and developers.

jba:
The JSON convention seems justified to me because JSON is primarily an interchange format, not a storage format. If Datastore dropped unknown fields, then round-tripping would lose data. That seems too high a risk for what is essentially a convenience feature.

It would also break behavior for a stable package.

I suppose we could do it as an option.

/cc @zombiezen 

@jskeet @garrettjonesgoogle How do the other typed languages handle deserializing unknown Datastore fields into objects?

jskeet:
In C# we don't have any code to deserialize the result into objects anyway, so there's equivalent, as far as I can see.

ChrisBartonLC:
Yes I guess you can't do it because updates are not fined grained and instead re-write the entire record. BTW before we started using a customer loader, in order to remove a field in datastore (thanks to v0.19.0) we would set the struct field to be a pointer and use omitempty then re-write all entities using a go script.

zombiezen:
When this scenario occurs, the datastore package is documented to return an [`ErrFieldMismatch`](https://godoc.org/cloud.google.com/go/datastore#ErrFieldMismatch) error, which you can then suppress.

ChrisBartonLC:
Does the struct still get filled as much as it can when this error occurs?

Thooms:
Seeing https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/datastore/load.go#L436-L438 I'd say so.

bhardwajankur3:
I am also facing this issue. What is the ETA for the fix? Is there any temporary workaround for this?

jba:
No priority on the issue means no ETA. 

We're not even clear on what the right design is.

[...]

PR #8694, opened 2023-10-12
fix(datastore): Ignore field mismatch errors

bhshkh:
ErrFieldMismatch is returned when a field is to be loaded into a different type than the one it was stored from, or when a field is missing or unexported in the destination struct.

This PR allows user to suppress such errors. Some scenarios in which this would be useful are listed in https://github.com/googleapis/google-cloud-go/issues/913

Points to note
1. Suppressing the error does not mean that Datastore will not fetch missing field. It just means that the field was still fetched from Datastore service but not loaded in user's struct.
2. When writing the data back to Datastore, it will be overwritten. E.g.
If entity written to Datastore is `{X: 1, Y:2}` and it is loaded into `type NewStruct struct{X int}`, then `{X:1}` is returned to the user. Now, if user modifies this `X = 5` and writes it back to Datastore, there will be no Y field left for this entity in Datastore.

Fixes: https://github.com/googleapis/google-cloud-go/issues/913


datastore/datastore.go
```diff
@@ -357,6 +357,54 @@ func checkMultiArg(v reflect.Value) (m multiArgType, elemType reflect.Type) {
 	return multiArgTypeInvalid, nil
 }
 
+// IgnoreFieldMismatch allows ignoring ErrFieldMismatch error while
+// reading or querying data.
+// WARNING: Ignoring ErrFieldMismatch can cause data loss
+func (c *Client) IgnoreFieldMismatch() {
```
codyoss, line 363: Consider expressing this as a client specific ClientIOption. Here is an example from storage: https://github.com/googleapis/google-cloud-go/blob/532a37c98680855fa38342bb7680062226594a0e/storage/option.go#L63-L75

bhshkh: Updated the PR

codyoss:
@bhshkh do we still want this PR?

datastore/option.go
```diff
@@ -0,0 +1,60 @@
...
+// A datastoreClientOption is an option for a Google Datastore client.
+type datastoreClientOption interface {
+	option.ClientOption
+	applyDatastoreOpt(*datastoreConfig)
+}
+
+// WithIgnoreFieldMismatch allows ignoring ErrFieldMismatch error while
+// reading or querying data.
+// WARNING: Ignoring ErrFieldMismatch can cause data loss
+func WithIgnoreFieldMismatch() option.ClientOption {
```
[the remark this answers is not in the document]
bhshkh: General naming convention for options is WIthXXXX
Some examples
https://github.com/googleapis/google-cloud-go/blob/main/vertexai/genai/option.go#L22-L26

https://github.com/googleapis/google-cloud-go/blob/main/storage/transfermanager/option.go#L27-L62

gkevinzheng (approved):

PR #8695, opened 2023-10-12
chore(ci): update Go version to 1.21

noahdietz:
Updates GitHub Action Go version to 1.21 where it wasn't already.

### DIFF
--- a/.github/workflows/vet.yml
+++ b/.github/workflows/vet.yml
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-go@v4
       with:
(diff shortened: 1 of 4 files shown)


quartzmo (approved):

.github/workflows/vet.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-go@v4
       with:
-        go-version: '1.20'
+        go-version: '1.21'
```
codyoss, line 19: I see we do .x elsewhere, should be be consistent?

codyoss (approved):

PR #8697, opened 2023-10-12
feat(auth): port external account changes

codyoss:
- https://github.com/golang/oauth2/commit/18352fc4335d95cd4b558f7c914898f1c069754e
- https://github.com/golang/oauth2/commit/43b6a7ba1972152df70cd1e5ac7705e981df5f66


auth/detect/internal/stsexchange/sts_exchange.go
```diff
@@ -28,50 +28,70 @@ import (
...
+)
+
+// Options stores the configuration for making an sts exchange request.
+type Options struct {
+	Client         *http.Client
+	Endpoint       string
+	Request        *TokenRequest
+	Authentication ClientAuthentication
+	Headers        http.Header
+	ExtraOpts      map[string]interface{}
```
quartzmo, line 47: Can you add a comment explaining `ExtraOpts` and what goes in it? (Or if it's already explained somewhere else, copy the explanation to here?)

auth/detect/internal/externalaccount/externalaccount.go
```diff
@@ -165,13 +165,13 @@ func (tp *tokenProvider) Token(ctx context.Context) (*auth.Token, error) {
...
-		authentication: clientAuth,
-		headers:        header,
-		extraOpts:      options,
+	stsResp, err := stsexchange.ExchangeToken(ctx, &stsexchange.Options{
+		Client:         tp.client,
+		Endpoint:       tp.opts.TokenURL,
+		Request:        stsRequest,
+		Authentication: clientAuth,
+		Headers:        header,
+		ExtraOpts:      options,
```
quartzmo, line 174: Maybe add a comment above L60 explaining what type of options go into the `options` map? A link to external docs if possible would be ideal.

auth/detect/internal/stsexchange/sts_exchange.go
codyoss: done.

auth/detect/internal/externalaccount/externalaccount.go
codyoss: Added docs on the struct type instead.

quartzmo (approved):

auth/detect/internal/externalaccount/info.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// versionUnknown is only used when the runtime version cannot be determined.
+const versionUnknown = "UNKNOWN"
+
+// goVersion returns a Go runtime version derived from the runtime environment
+// that is modified to be suitable for reporting in a header, meaning it has no
+// whitespace. If it is unable to determine the Go runtime version, it returns
+// versionUnknown.
+func goVersion() string {
```
noahdietz, line 35: Any interest in using the `gax.GoVersion`? https://pkg.go.dev/github.com/googleapis/gax-go/v2#pkg-variables

Not sure if it would be good to take the dep though...

noahdietz (approved):

auth/detect/internal/externalaccount/info.go
codyoss: Nope for that exact reason haha. We may eventually want/need to do it but it would be unfortunate. Right now this module overall does depend on the gRPC stack, but the detect tree does not. And Go is smart enough to only realize used package level deps. Only the grpctransport package currently brings in that dep. If we ever made a gax v3 it would be nice to separate large external deps to their own package.

noahdietz: For sure, makes sense! thanks.

PR #8698, opened 2023-10-12
chore(storage): add invocation ID to gRPC calls

tritone:
This adds x-goog-gcs-idempotency-token and x-goog-api-client header
keys to all gRPC calls. It also refactors how this is done for
both HTTP and gRPC to take advantage of passing in headers via
the callctx package rather than by mutating requests directly.


frankyn (approved):
Thanks @tritone from my perspective this LGTM, however, I agree with you in that to test this properly we need testbench to verify that all methods are plumbing this correctly [1](https://github.com/googleapis/storage-testbench/issues/486).

storage/invoke_test.go
```diff
@@ -176,27 +176,30 @@ func TestInvoke(t *testing.T) {
...
-			req := &fakeApiaryRequest{header: http.Header{}}
 			var initialClientHeader, initialIdempotencyHeader string
-			call := func() error {
+			var gotClientHeader, gotIdempotencyHeader string
+			call := func(ctx context.Context) error {
 				if counter == 0 {
-					initialClientHeader = req.Header()["X-Goog-Api-Client"][0]
-					initialIdempotencyHeader = req.Header()["X-Goog-Gcs-Idempotency-Token"][0]
+					headers := callctx.HeadersFromContext(ctx)
+					//log.Printf("headers: %+v", headers)
```
BrennaEpp, line 184: Was this accidentally left in?

BrennaEpp (approved):

storage/invoke_test.go
tritone: Yes, good catch, I'll remove.

PR #8718, opened 2023-10-13
chore(auth): add a bunch of missing validation logic

codyoss:
- made the look of validation consistent across packages
- moved refresh token into the options struct for 3lo (this is the flagged breaking change.)


auth/auth.go
```diff
@@ -249,16 +250,34 @@ type Options2LO struct {
...
+func (o *Options2LO) client() *http.Client {
+	if o.Client != nil {
+		return o.Client
 	}
 	return internal.CloneDefaultClient()
 }
 
+func (o *Options2LO) validate() error {
+	if o == nil {
+		return errors.New("auth: options must be provided")
```
quartzmo, line 262: Would it make sense to add "Options2LO" to these error messages? For example: ""auth: Options2LO.Email must be provided", etc?

codyoss: I don't think it is necessary. The only time a user would see these is if you call the New method for 2lo token source directly. I think it is implied what these all relate to for that reason.

auth/detect/internal/impersonate/impersonate.go
```diff
@@ -73,6 +76,16 @@ type Options struct {
 	Client *http.Client
 }
 
+func (o *Options) validate() error {
+	if o.Tp == nil {
+		return errors.New("detect: missing required 'source_credentials' field in credentials")
```
quartzmo, line 81: Should this be labeled "impersonate"? Or even if "detect" is the surface, should we somehow identify the `impersonate.Options` origin of this error?

auth/impersonate/idtoken.go
```diff
@@ -56,6 +57,19 @@ type IDTokenOptions struct {
 	Client *http.Client
 }
 
+func (o *IDTokenOptions) validate() error {
+	if o == nil {
+		return errors.New("impersonate: options must be provided")
```
quartzmo, line 62: Should this be labeled "idtoken"? Or even if "impersonate" is the surface, should we somehow identify the `IDTokenOptions` origin of this error?

auth/threelegged.go
```diff
@@ -66,6 +70,31 @@ type Options3LO struct {
 	AuthHandlerOpts *AuthorizationHandlerOptions
 }
 
+func (o *Options3LO) validate() error {
+	if o == nil {
+		return errors.New("auth: options must be provided")
```
quartzmo, line 75: Would it make sense to add "Options3LO" to these error messages? For example: ""auth: Options3LO. ClientID must be provided", etc?

auth/detect/internal/impersonate/impersonate.go
codyoss: worked in the word impersonated. Should be detect though because that is the public package this error is surfaced through. Either that or we drop the detect, but we have used a similar convention in the past.

auth/impersonate/idtoken.go
codyoss: you only see this from calling NewIDTokenProvider, so it should be implied. The idom is prefix with the package so that is why it is impersonate.

auth/detect/detect.go
```diff
@@ -182,6 +188,19 @@ type Options struct {
 	Client *http.Client
 }
 
+func (o *Options) validate() error {
+	if o == nil {
+		return errors.New("detect: options must be provided")
+	}
+	if len(o.Scopes) > 0 && o.Audience != "" {
+		return errors.New("detect: both scopes or audience were provided")
```
noahdietz, line 196:
```suggestion
		return errors.New("detect: both scopes and audience were provided")
```

noahdietz (approved):

quartzmo (approved):

PR #8721, opened 2023-10-13
feat(storage): support autoclass v2.1

### LINKED ISSUE #8584: storage: add support for Autoclass v2.1
Add support for new AutoclassConfig metadata fields. Full details are available in this [internal doc](http://goto.google.com/ac2-1-sdk-request).

Please wait for approval from myself before merging/releasing so we can [coordinate](http://goto.google.com/gcs-sdk-ac2-1) around release dates.

cojenco:
Fixes #8584


storage/integration_test.go
```diff
@@ -883,22 +883,47 @@ func TestIntegration_Autoclass(t *testing.T) {
...
+			t.Errorf("attr.Autoclass.Enabled = %v, want %v", got, want)
+		}
+		if got, want := attrs.Autoclass.TerminalStorageClass, "ARCHIVE"; got != want {
+			t.Errorf("attr.Autoclass.TerminalStorageClass = %v, want %v", got, want)
+		}
+		latestTscUpdateTime := attrs.Autoclass.TerminalStorageClassUpdateTime
+		if latestTscUpdateTime.IsZero() {
+			t.Error("got a zero time value, want a populated value")
+		}
+		if latestTscUpdateTime.Before(tscUpdateTime) {
```
tritone, line 921: I think we should check that the new update time is after? If they are equal it still seems wrong.

Also, nit, `latestTSCUpdateTime` by go naming conventions.

storage/bucket.go
```diff
@@ -740,6 +740,13 @@ type Autoclass struct {
 	// If Autoclass is enabled when the bucket is created, the ToggleTime
 	// is set to the bucket creation time. This field is read-only.
 	ToggleTime time.Time
+	// TerminalStorageClass: The storage class that objects in the bucket
+	// eventually transition to if they are not read for a certain length of
+	// time. Valid values are NEARLINE and ARCHIVE.
+	TerminalStorageClass string
+	// TerminalStorageClassUpdateTime represents the time of the most recent
+	// update to "terminalStorageClass".
```
tritone, line 748: nit: `TerminalStorageClass` (capitalized).

tritone (commented):
LGTM overall, just some minor stuff

storage/integration_test.go
cojenco: updated, thanks!

tritone (approved):

storage/integration_test.go
```diff
@@ -907,7 +932,7 @@ func TestIntegration_Autoclass(t *testing.T) {
 		if latestToggleTime.IsZero() {
 			t.Error("got a zero time value, want a populated value")
 		}
-		if latestToggleTime.Before(toggleTime) {
+		if !latestToggleTime.After(toggleTime) {
```
BrennaEpp, line 935: Why was this condition check inverted? Isn't it the same thing, or is there some technicality which I am missing?

storage/bucket_test.go
```diff
@@ -65,7 +65,7 @@ func TestBucketAttrsToRawBucket(t *testing.T) {
 		Encryption: &BucketEncryption{DefaultKMSKeyName: "key"},
 		Logging:    &BucketLogging{LogBucket: "lb", LogObjectPrefix: "p"},
 		Website:    &BucketWebsite{MainPageSuffix: "mps", NotFoundPage: "404"},
-		Autoclass:  &Autoclass{Enabled: true},
+		Autoclass:  &Autoclass{Enabled: true, TerminalStorageClass: "NEARLINE"},
```
BrennaEpp, line 68: Are we also checking conversions in which the values for autoclass are not set?

BrennaEpp (approved):
LGTM; just a couple minor questions

storage/integration_test.go
cojenco: They should not equal to each other either, hence the update

storage/bucket_test.go
cojenco: Not that I know of... do we have tests that check conversions for unset properties like autoclass or ubla? The autoclass config works similarly as ubla

BrennaEpp: Not sure; it would be good to add them imo, but it's okay to merge this as-is to unblock

storage/integration_test.go
BrennaEpp: I believe `time.Before` should return false if they are equal to each other; but anyway, I confirmed these 2 lines are the same thing.
Please ignore this as nitpick not really worth discussing.

cojenco: You're right they are the same, so we are catching the case if two toggle times are equal, thank you Brenna!

PR #8755, opened 2023-10-23
feat(bigtable): Add support for reverse scans

brandtnewton:


bigtable/bigtable.go
```diff
@@ -372,13 +440,27 @@ func (r RowRange) String() string {
...
 
 func (r RowRange) proto() *btpb.RowSet {
-	rr := &btpb.RowRange{
-		StartKey: &btpb.RowRange_StartKeyClosed{StartKeyClosed: []byte(r.start)},
+	var rr = btpb.RowRange{}
+
+	// empty strings are NOT considered unbounded for backwards compatibility
+	if r.startBound == unbounded {
+		// leave unbounded
+	} else if r.startBound == closed || r.startBound == undefined {
```
igorbernstein2, line 448: can we normalize these during construction to avoid dealing with ambiguity ?

bigtable/bigtable.go
```diff
@@ -450,8 +555,18 @@ func PrefixRange(prefix string) RowRange {
...
+
+// InfiniteReverseRange returns the RowRange consisting of all keys at least as
+// large as start.
+func InfiniteReverseRange(limit string) RowRange {
+	return RowRange{
+		limit:      limit,
+		startBound: unbounded,
+		endBound:   closed,
 	}
 }
```
igorbernstein2, line 571: can we group all of the constructors together?

igorbernstein2 (changes requested):

bigtable/bigtable.go
```diff
@@ -331,70 +346,206 @@ func (r RowList) retainRowsAfter(lastRowKey string) RowSet {
...
+		start:      start,
+		endBound:   endBound,
+		end:        end,
 	}
 }
 
 // Unbounded tests whether a RowRange is unbounded.
 func (r RowRange) Unbounded() bool {
-	return r.limit == ""
+	return r.startBound == rangeUnbounded || r.endBound == rangeUnbounded
```
brandtnewton, line 417: technically, this is NOT backwards compatible.

bigtable/bigtable.go
```diff
@@ -331,70 +346,248 @@ func (r RowList) retainRowsAfter(lastRowKey string) RowSet {
...
+		return true
+	}
+
+	if r.startBound == rangeOpen || r.endBound == rangeOpen {
+		return r.start < r.end
+	} else if r.startBound == rangeClosed {
+		return r.start <= r.end
+	} else {
+		return true
+	}
```
igorbernstein2, line 590: This pretty hard to reason about and is not immediately clear why this is correct. I would either heavily comment this or make it more explicit. ie

```
// if either end is unbounded, then the range is always valid
if r.Unbounded() {
   return true
}

// If either end is an open interval, then the start must be strictly less than the end
// And since neither end is unbounded, we don't have to check for empty strings
if r.startBound == rangeOpen || r.endBound == rangeOpen {
  return r.start < r.end
}

// At this point both endpoints must be closed, which makes [a,a] a valid interval
return r.start <= r.end
```

bigtable/integration_test.go
```diff
@@ -885,7 +918,7 @@ func TestIntegration_Read(t *testing.T) {
 	}{
 		{
 			desc: "read all, unfiltered",
-			rr:   RowRange{},
```
igorbernstein2, line 888: I'm concerned that this had to change....it implies a breaking change. I think the root of the problem is that rangeUnbounded is not mapped to 0 in your enum

bigtable/internal/testproxy/proxy.go
```diff
@@ -573,6 +573,12 @@ func (s *goTestProxyServer) ReadRows(ctx context.Context, req *pb.ReadRowsReques
 		rs = bigtable.InfiniteRange("")
 	}
 
+	var opts []bigtable.ReadOption
```
igorbernstein2, line 576: Please move the proxy changes to a separate PR

igorbernstein2 (changes requested):

bigtable/bigtable.go
brandtnewton: Agreed. Thanks for the suggestion!

bigtable/bigtable.go
brandtnewton: Previously:
```
rr := NewRange("", "z")
rr.String() // ["", "z")
rr.Unbounded() // false
```
but now:
```
rr := NewRange("", "z")
rr.String() // (∞, "z")
rr.Unbounded() // true
```
Now I realize that String() is also not backwards compatible this way

bigtable/integration_test.go
brandtnewton: True. I think changing the default rangeBound to rangeUnbounded will preserve the behavior of RowRange better - although direct use of the RowRange constructor is discouraged

igorbernstein2 (approved):

bigtable/integration_test.go
igorbernstein2: Agreed, but its almost definitely being used by endusers.

PR #8759, opened 2023-10-23
chore(internal/godocfx): test a specific golden package version

### LINKED ISSUE #8130: figure out what to do with godocfx test
This test fails every time there is a dep bump because it generates off live docs of a client. We should rework the test to be stable or remove it.

Updates: #8129

tbpg:
I also updated another test to be less restrictive with the number of files that get generated.

Fixes #8130.


codyoss (approved):
Thanks!

Issue #8768, opened 2023-10-25
spanner/spansql: add support for ARRAY_FILTER, ARRAY_INCLUDES and ARRAY_TRANSFORM


neglect-yp:
**Is your feature request related to a problem? Please describe.**

Array functions with lambdas as arguments cannot be parsed.

- https://cloud.google.com/spanner/docs/reference/standard-sql/array_functions#array_filter
- https://cloud.google.com/spanner/docs/reference/standard-sql/array_functions#array_includes
- https://cloud.google.com/spanner/docs/reference/standard-sql/array_functions#array_transform

**Describe the solution you'd like**

Add a special function to parse the arguments of these functions.

**Describe alternatives you've considered**

No idea.

**Additional context**

I'll create a PR for this issue.

surbhigarg92:
@rahul2393 Can you please triage this issue

noahdietz:
Tagged as a feature request for rotation triaging. Feel free to change.

PR #8769, opened 2023-10-25
feat(spanner/spansql): add support for ARRAY_FILTER, ARRAY_INCLUDES and ARRAY_TRANSFORM

neglect-yp:
fix: https://github.com/googleapis/google-cloud-go/issues/8768

This PR adds supports for array functions with lambdas as arguments.


codyoss:
@rahul2393 is this a change you would like to accept?

PR #8930, opened 2023-10-27
feat(storage): change gRPC writes to use bi-directional streams

BrennaEpp:
Emulated test is failing as bidi writes are not implemented in the emulator yet, so I added a quick skip for that test.


storage/grpc_client.go
```diff
@@ -1652,38 +1664,43 @@ func (w *gRPCWriter) uploadBuffer(recvd int, start int64, doneReading bool) (*st
...
 		err = w.stream.Send(req)
 		if err == io.EOF {
 			// err was io.EOF. The client-side of a stream only gets an EOF on Send
 			// when the backend closes the stream and wants to return an error
 			// status. Closing the stream receives the status as an error.
-			_, err = w.stream.CloseAndRecv()
+			err = w.stream.CloseSend()
+
+			// Drop the stream reference as a new one will need to be created
+			w.stream = nil
```
noahdietz, line 1677: Are we 100% sure we can drop the stream reference here? If `w.stream.Send` gets an io.EOF, it means the backend wants to send an error. The error status won't come from CloseSend(), I think the client side still needs to `Recv` to get the actual Status (will send a link offline).  I think we need to call Recv or CloseAndRecv if the bidi stream still has it, before Closing.

storage/grpc_client.go
```diff
@@ -1692,31 +1709,63 @@ func (w *gRPCWriter) uploadBuffer(recvd int, start int64, doneReading bool) (*st
...
 		}
 		if err != nil {
-			return nil, 0, false, err
+			return nil, 0, err
 		}
 
-		return resp.GetResource(), offset, finalized, nil
+		// Confirm the persisted data
+		if !lastWriteOfEntireObject {
+			if resp.GetPersistedSize() != writeOffset {
```
tritone, line 1739: We'll need to add some kind of test for this case since it's a tricky one and I don't think the conformance tests would cover it (can be in a separate PR though).

storage/grpc_client.go
```diff
@@ -1692,31 +1709,63 @@ func (w *gRPCWriter) uploadBuffer(recvd int, start int64, doneReading bool) (*st
...
+				// retry
+				writeOffset = resp.GetPersistedSize()
+				sent = int(writeOffset) - int(start)
+				continue
+			}
+		} else {
+			// If the object is done uploading, close the send stream and check for errors.
+			err = w.stream.CloseSend()
+			if err != nil {
```
tritone, line 1748: Is this not also potentially a retry case depending on the error?

storage/grpc_client.go
```diff
@@ -1692,31 +1709,63 @@ func (w *gRPCWriter) uploadBuffer(recvd int, start int64, doneReading bool) (*st
...
+				return nil, 0, err
+			}
+
+			// Stream receives do not block once send is closed, but we may not
+			// receive the response with the object right away; loop until we
+			// receive the object or error out
```
tritone, line 1754: Is it okay to do this without backoff? Wondering if this is a typical pattern.

tritone (commented):
Overall looks pretty good, but I had some questions

storage/grpc_client.go
BrennaEpp: Thanks Noah, you are 100% correct. In this case, it would be a Recv() (the bidi stream does not have CloseAndRecv). Updated the code.

As for testing it out in practice, I agree. I think the best is to do this via the conformance tests, pending testbench implementation.

storage/grpc_client.go
```diff
@@ -1539,7 +1539,7 @@ type gRPCWriter struct {
 	chunkSize  int
 
 	// The gRPC client-stream used for sending buffers.
-	stream storagepb.Storage_WriteObjectClient
+	stream storagepb.Storage_BidiWriteObjectClient
```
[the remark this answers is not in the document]
tritone: Yeah, we can follow up on this with an internal bug as discussed offline.

storage/grpc_client.go
tritone: I see, that is kind of confusing and seems like a strange contract for the method to return an error but never send a non-nil one. Maybe just make a note in a comment about this then?

storage/grpc_client.go
```diff
@@ -1688,6 +1691,9 @@ func (w *gRPCWriter) uploadBuffer(recvd int, start int64, doneReading bool) (*st
 				}
 				sent = int(writeOffset) - int(start)
 
+				// Drop the stream reference as a new one will need to be created.
+				w.stream = nil
```
tritone, line 1695: Shouldn't this happen outside of the `if ShouldRetry()` conditional?

storage/grpc_client.go
BrennaEpp: It looks like they are leaving the possibility open to return an error if you try to call the function more than once ( `TODO: return an error and finish the stream instead, due to API misuse?` )... and maybe possibly in other cases? I'm not sure on the intentions behind always just returning `nil` instead of not returning anything.

I'll add a comment here.

storage/grpc_client.go
```diff
@@ -1762,6 +1774,14 @@ func (w *gRPCWriter) uploadBuffer(recvd int, start int64, doneReading bool) (*st
 				obj = resp.GetResource()
 			}
 
+			// Even though we received the object response, continue reading
+			// until we receive a non-nil error, to ensure the stream does not
+			// leak even if the context isn't cancelled. See:
+			// https://github.com/grpc/grpc-go/commit/365770fcbd7dfb9d921cb44827ede770f33be44f
```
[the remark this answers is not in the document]
BrennaEpp: This is the actual docs: https://pkg.go.dev/google.golang.org/grpc#ClientConn.NewStream

I'll update it in the comment as well.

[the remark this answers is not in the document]
tritone: After looking at these docs I'm wondering if we can use the CloseAndRecv helper instead to avoid writing these loops? We already do this here: https://github.com/googleapis/google-cloud-go/blob/cc83515d0c837c8b1596a97b6f09d519a0f75f72/storage/grpc_client.go#L1748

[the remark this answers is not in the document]
BrennaEpp: We don't have that helper function for the bidi stream

storage/grpc_client.go
BrennaEpp: I moved this line

tritone (approved):

PR #8935, opened 2023-10-30
feat(spanner): support of uint datatype for gorm.Model 

rahul2393:
Golang ORM (GORM) customers use [gorm.Model](https://gorm.io/docs/models.html#gorm-Model) for declaring the models in application code, this PR will add support of `uint` type in Cloud Spanner client library since Cloud Spanner does not support uint datatype natively


spanner/value.go
```diff
@@ -3471,6 +3471,30 @@ func encodeValue(v interface{}) (*proto3.Value, *sppb.Type, error) {
 			}
 		}
 		pt = listType(intType())
+	case uint:
```
harshachinta, line 3474: nit: shouldnt' we add support for uint type in `decodeValue` method also?

harshachinta (approved):

spanner/value.go
olavloite: +1

rahul2393: gorm.Model tests were able decode to uint with making any changes in `decodeValue` function, seems like `default` case is sufficient, let me add tests validating it

codyoss:
@rahul2393 do we still want this change?

codyoss:
@rahul2393 see my comment above

PR #8938, opened 2023-10-30
test(storage): add object to encryption test to reduce rate limit issues

### LINKED ISSUE #8727: storage: TestIntegration_Encryption failed
Note: #5473 was also for this test, but it was closed more than 10 days ago. So, I didn't mark it flaky.

----

commit: 3a4ec650177be4d48aa7a0b8a22ea2b211522d80
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/9dca3e3f-3c69-4054-b44b-687745e52aeb), [Sponge](http://sponge2/9dca3e3f-3c69-4054-b44b-687745e52aeb)
status: failed
<details><summary>Test output</summary><br><pre>    integration_test.go:2570: googleapi: Error 429: The object go-integration-test-1697440231333269397-01/customer-encryption-2 exceeded the rate limit for object mutation operations (create, update, and delete). Please reduce your request rate. See https://cloud.google.com/storage/docs/gcs429., rateLimitExceeded</pre></details>

BrennaEpp:
Fixes #8727 

Test may fail again as we are writing several times to obj2 in a short amount of time. May need to add another object/more carefully interleave this one (to make sure we are still testing the same things). For now, adding this may be sufficient.


storage/integration_test.go
```diff
@@ -2513,9 +2513,13 @@ func TestIntegration_Encryption(t *testing.T) {
...
 		checkRead("first object", obj, key, contents)
 
+		// We create 2 objects here and we can interleave operations to get around
+		// the rate limit for object mutation operations (create, update, and delete).
 		obj2 := client.Bucket(bucket).Object("customer-encryption-2")
+		obj4 := client.Bucket(bucket).Object("customer-encryption-4")
+
 		// Copying an object without the key should fail.
-		if _, err := obj2.CopierFrom(obj).Run(ctx); err == nil {
+		if _, err := obj4.CopierFrom(obj).Run(ctx); err == nil {
```
frankyn, line 2522: looks like you can update lines 2526 - 2530 to use obj4 as well.

BrennaEpp: That is true, but the objective here is to not use the same object consecutively if we can avoid it as the rate limit for objects is one create/update/delete per second. So IIUC L2562 will not hit that limit even if it occurs in the same second as 2522 - if we use the same object, it might.

frankyn (approved):

tritone (approved):

PR #8941, opened 2023-10-30
test(bigquery): OpenTelemetry tracing support

alvarowolfx:
Soon we are gonna start supporting OpenTelemetry tracing and sunset OpenCensus support. This PR test the internal changes being made on #8655 against BigQuery SDK as we go through the transition. Initially we are gonna be testing both observability libraries, but later we are gonna be supporting only Otel, so this PR matches this effort on the BigQuery SDK.

This is only for tracing. Metrics is still in discussion as we use the OpenCensus API more directly.

Dependes on #8655

### DIFF
--- a/bigquery/trace_integration_test.go
+++ b/bigquery/trace_integration_test.go
@@ -16,20 +16,54 @@ package bigquery
 
 import (
 	"context"
+	"os"
 	"strings"
 	"testing"
 	"time"
 
 	"go.opencensus.io/trace"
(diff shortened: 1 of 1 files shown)


shollyman (approved):

codyoss:
@alvarowolfx is this a change we still want?

codyoss:
@alvarowolfx see my comment above

alvarowolfx:
superseded by #11231

PR #8953, opened 2023-11-01
feat(spanner): add methods to return Row fields

harshachinta:
This PR adds code changes to return cloud spanner `Type` and `Value` of [Row](https://github.com/googleapis/google-cloud-go/blob/ce5b884fba64fd7a124ddddb8869efcd48c070f4/spanner/row.go#L88) object which could be used for executor framework for other conversions.


spanner/row.go
```diff
@@ -174,6 +174,22 @@ func (r *Row) ColumnNames() []string {
...
+func (r *Row) ColumnType(i int) *sppb.Type {
+	if i < 0 || i >= len(r.fields) {
+		return nil
+	}
+	return r.fields[i].Type
+}
+
+// ColumnValue returns the Cloud Spanner Value of column i, or nil for invalid column.
+func (r *Row) ColumnValue(i int) *proto3.Value {
+	if i < 0 || i >= len(r.fields) {
```
rahul2393, line 187:
```suggestion
	if i < 0 || i >= len(r.vals) {
```

rahul2393 (approved):

spanner/row.go
harshachinta: Thanks for catching this!

PR #8972, opened 2023-11-03
feat(datastore): Adding BeginLater transaction option implementation

bhshkh:
**What is new_transaction ?**
ReadOptions are the options shared by Datastore read requests. 
[new_transaction](https://github.com/googleapis/go-genproto/blob/478b75c54725e54e5f92966212c6572c091596aa/googleapis/datastore/v1/datastore.pb.go#L1794-L1803) is a [consistency type](https://github.com/googleapis/go-genproto/blob/478b75c54725e54e5f92966212c6572c091596aa/googleapis/datastore/v1/datastore.pb.go#L1708) in ReadOptions. When new_transaction is used in a read request, a new transaction is begun for the request.

This PR adds following:

1. BeginLater transaction option implementation
2. Mutex for state in the transaction struct

Currently, when user calls NewTransaction method, client library makes a BeginTransaction rpc call and obtains transaction id. Reads (RunQuery / RunAggregationQuery / LookupRequest), Commit and Rollback use this transaction id in ReadOptions of rpc request.

Without BeginLater transaction option: the above behavior remains unchanged.

With BeginLater transaction option:
1. Calling NewTransaction method does not make a BeginTransaction rpc call.
2. The first read rpc passes new_transaction as consistency type to obtain transaction id.
3. If there is no read called before commit / rollback, begin transaction is called to obtain transaction id before commit / rollback rpc.
Since there is one less round trip to the Datastore service in point 1 above, user may see lesser latencies when using BeginLater transaction option.


datastore/transaction.go
```diff
@@ -159,47 +157,99 @@ func (c *Client) NewTransaction(ctx context.Context, opts ...TransactionOption)
...
+		return nil
+	}
+
+	txnID, err := t.beginTransaction()
+	if err != nil {
+		return err
+	}
+
+	t.id = txnID
+	t.state = transactionStateInProgress
```
danieljbruce, line 228: Would it make sense to have a specific function responsible for saving the id and saving the new state since we do this in several places?

datastore/transaction.go
```diff
@@ -299,12 +356,70 @@ func (t *Transaction) Rollback() (err error) {
...
+		}
+	}
+	return opts, nil
+}
+
+func (t *Transaction) get(spanName string, keys []*Key, dst interface{}) (err error) {
+	t.ctx = trace.StartSpan(t.ctx, spanName)
+	defer func() { trace.EndSpan(t.ctx, err) }()
+
+	if t.state == transactionStateNotStarted {
```
danieljbruce, line 402: It looks like this business logic also exists in `nextBatch`. Maybe the `defer` gets in the way of being able to move this into a common function, but is there any way for this mutex locking logic to live in only one place? Perhaps this can be done using a decorator pattern?

danieljbruce (commented):
Initial review with a few comments and questions

datastore/transaction.go
bhshkh: I couldn't find a way to implement decorator pattern but used [this](https://stackoverflow.com/a/68935422/2419636) as reference to move defer to new function.

datastore/query.go
```diff
@@ -946,12 +959,35 @@ func (t *Iterator) nextBatch() error {
 		q.Limit = nil
 	}
```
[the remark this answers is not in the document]
bhshkh: nextBatch is used at all these locations: https://github.com/search?q=repo%3Agoogleapis%2Fgoogle-cloud-go+path%3A%2F%5Edatastore%5C%2F%2F+nextBatch&type=code

runAggregationQuery mutex locking is here https://github.com/googleapis/google-cloud-go/pull/8972/files#diff-c04af33c4c45ff4606abe36bd859432effb278cca4146d3f818e540f0479fd8dR792-R794

bhshkh:

danieljbruce:
Does this client support retry transactions? Like before any of these changes, if the user provides an id when creating a transaction then will calling a read function make a grpc read request with that transaction id like it does in Node Datastore?

danieljbruce (commented):
Added one question about retry transactions to make sure there is parity with the Node client.

bhshkh:
> Does this client support retry transactions? Like before any of these changes, if the user provides an id when creating a transaction then will calling a read function make a grpc read request with that transaction id like it does in Node Datastore?

before and after any of these changes, no. Go client did not have retry transaction support

datastore/transaction_test.go
```diff
@@ -88,3 +88,398 @@ func TestNewTransaction(t *testing.T) {
...
+				{
+					wantReq: commitReq,
+					resp:    commitRes,
+				},
+			},
+			ops:      []string{"RunAggregationQuery", "Get", "Put", "Commit"},
+			settings: &transactionSettings{},
+		},
+		{
+			desc: "[RunAggregationQuery, Get, Put, Commit] BeginLater. RunAggregationQuery passes new_transaction",
```
danieljbruce, line 388: It looks like this test verifies that runAggregationQuery will pass in new transaction options. Just for my own knowledge though, what change in this PR allows newTransaction options to be passed in for the runAggregationQuery call? Does runAggregationQuery make a call to the new `parseReadOptions()`?

datastore/transaction_test.go
```diff
@@ -88,3 +88,398 @@ func TestNewTransaction(t *testing.T) {
 		}
 	}
 }
+
```
danieljbruce, line 91: nit: Other tests to consider if necessary:
1. Make sure readTime is passed along for read only transactions
2. Make sure prevId is passed along and used for read/write transactions

danieljbruce (approved):
LGTM overall. Just a few nits and questions.

datastore/transaction_test.go
bhshkh: RunAggregationQuery makes call to parseQueryReadOptions which calls parseReadOptions

 https://github.com/googleapis/google-cloud-go/pull/8972/files#diff-c04af33c4c45ff4606abe36bd859432effb278cca4146d3f818e540f0479fd8dR806

datastore/transaction_test.go
bhshkh: 1. Tested here: https://github.com/googleapis/google-cloud-go/pull/8972/files#diff-8a638c1c614640f81c811cc62b5e9bacbb8b17493e158922ccb9ee004d4ffe70R70-R80
2. prevId is not being used anywhere in Go client library.

datastore/transaction.go
```diff
@@ -159,47 +157,115 @@ func (c *Client) NewTransaction(ctx context.Context, opts ...TransactionOption)
...
+		// Check whether state changed while waiting to acquire lock
+		if t.state == transactionStateNotStarted {
+			return func() { t.stateLock.Unlock() }
+		}
+		t.stateLock.Unlock()
+	}
+	return func() {}
+}
+
+func (t *Transaction) startProgress(id []byte) {
```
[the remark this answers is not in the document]
bhshkh: Resolved in https://github.com/googleapis/google-cloud-go/pull/10032

datastore/transaction.go
```diff
@@ -159,47 +157,115 @@ func (c *Client) NewTransaction(ctx context.Context, opts ...TransactionOption)
...
+	txnID, err := t.beginTransaction()
+	if err != nil {
+		return err
+	}
+
+	t.startProgress(txnID)
+	return nil
+}
+
+// Acquire state lock if transaction has not started
```
gkevinzheng, line 231: Could you add to the comment for `acquireLock` more about what it is, what its return value signifies, and how it's used in other parts of the PR? It does a lot more than simply acquiring the state lock.

Also because this function does a lot more than just acquiring the state lock, could you rename the function to reflect that?

datastore/transaction.go
bhshkh: Added

datastore/transaction.go
bhshkh: Resolved in https://github.com/googleapis/google-cloud-go/pull/10063

datastore/query_test.go
```diff
@@ -889,6 +886,8 @@ func TestAggregationQueryIsNil(t *testing.T) {
 }
 
 func TestValidateReadOptions(t *testing.T) {
+	eventualInTxnErr := errors.New("datastore: cannot use EventualConsistency query in a transaction")
```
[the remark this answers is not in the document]
bhshkh: Resolved in https://github.com/googleapis/google-cloud-go/pull/10063

PR #8979, opened 2023-11-06
fix(logging): added marshalling methods for proto fields in structuredLogEntry

### LINKED ISSUE #7785: logging: aoid json.marshall for proto objects
Error report:

> one of our static analysis tools found a bug in PR https://github.com/googleapis/google-cloud-go/pull/7083
> 
> The error message is
> 
>   > do not call encoding/json.Marshal on composite types containing protobuf message types
> 
> The explanation is [go/goprotoapi-findings#message-goreflect-json](https://goto.google.com/goprotoapi-findings#message-goreflect-json)
> 
> The line in question is https://github.com/googleapis/google-cloud-go/blob/69eedabcac1a96286d0007b27f27cc61c5e4b26e/logging/logging.go#L980
> It might be a false positive but you should probably take a look.

It appears that we need to write custom `MarshalJSON` functions for the proto types that make up `structuredLogEntry`

gkevinzheng:
Fixes #7785 while still maintaining camelCase keys when marshalling proto objects.


logging/logging.go
```diff
@@ -967,65 +968,70 @@ func toLogEntryInternalImpl(e Entry, l *Logger, parent string, skipLevels int) (
...
-	Labels         map[string]string             `json:"logging.googleapis.com/labels,omitempty"`
-	InsertID       string                        `json:"logging.googleapis.com/insertId,omitempty"`
-	Operation      *logpb.LogEntryOperation      `json:"logging.googleapis.com/operation,omitempty"`
-	SourceLocation *logpb.LogEntrySourceLocation `json:"logging.googleapis.com/sourceLocation,omitempty"`
-	SpanID         string                        `json:"logging.googleapis.com/spanId,omitempty"`
-	Trace          string                        `json:"logging.googleapis.com/trace,omitempty"`
-	TraceSampled   bool                          `json:"logging.googleapis.com/trace_sampled,omitempty"`
+	Message        json.RawMessage                   `json:"message"`
+	Severity       string                            `json:"severity,omitempty"`
+	HTTPRequest    *structuredLogEntryHTTPRequest    `json:"httpRequest,omitempty"`
```
bhshkh, line 973: Since datatype of HTTPRequest is being changed, this can be a breaking change if the user is able to access this field. I see that structuredLogEntry type itself is unexported. So, it is highly unlikely but please recheck the rest of the logging code to ensure the same.

bhshkh (approved):

daniel-sanche (approved):
LGTM

PR #8984, opened 2023-11-07
feat(datastore): Adding BeginLater and transaction state

bhshkh:
This PR adds following variables:
- BeginLater transaction option
- State enum in the transaction struct to track state of the transaction. Before this PR, transaction id was being used to track state of the transaction.
- Currently, when a user instantiates a transaction object, the transaction options passed in the constructor (NewTransaction method) are being used only in the constructor itself to make a BeginTransaction rpc call and obtain transaction id from its response. So, there was no need to store these options.
Now, these transaction options need to be saved to be used in BeginLater transactions in future PRs. So, a new field "settings" in the transaction struct has been added in this PR.


gkevinzheng (commented):
LGTM

datastore/transaction.go
```diff
@@ -37,6 +37,14 @@ type transactionSettings struct {
 	readOnly bool
 	prevID   []byte // ID of the transaction to retry
 	readTime *timestamppb.Timestamp
+
+	// When set, skips the initial BeginTransaction RPC call to obtain txn id and
+	// uses the piggybacked txn id from first read rpc call.
+	// If there are no read operations on transaction, BeginTransaction RPC call is made
+	// before rollback or commit
+	// Currently, this setting is set but unused
+	// TODO: Use this setting
```
cindy-peng, line 46: Nit: Do we usually add buganizer for TODOs? [go/todo-style](https://goto.google.com/todo-style)

bhshkh: Updated

cindy-peng (approved):

PR #8985, opened 2023-11-07
refactor(datastore): Add validateReadOptions method

bhshkh:
This PR  adds `validateReadOptions` method. In future PRs, [this](https://github.com/googleapis/google-cloud-go/blob/ca7667194007394bdcade8058fa84c1fe19c06b1/datastore/query.go#L506) call to `parseReadOptions` in `toRunQueryRequest` will be replaced by `validateReadOptions`
**Reasoning:**
When user runs below lines of code, the query is converted to RunQueryRequest proto and saved in the iterator returned to the user. During conversion, request read options are parsed as well.
```
it := client.Run(ctx, query)
```
When user runs below lines of code, RunQuery grpc call is made with the request which was saved in the iterator.
```
_, err := it.Next(&city)
```
With BeginLater transaction option being introduced in the upcoming PRs, read options to be sent in RunQueryRequest will depend on state of the transaction. By the time `it.Next` is executed, the state of the transaction in `it` may have changed. So, the read options in the request need to be updated before sending the request. 

In the upcoming PRs for this feature, to make the library forward compatible with running in multi-threaded environment, the state will be protected by a mutex. To keep the size of the critical section small, setting of the read option will be done just before the  request is sent. Read options will only be validated and not set when iterator is instantiated.


datastore/query.go
```diff
@@ -808,21 +810,32 @@ func (c *Client) RunAggregationQuery(ctx context.Context, aq *AggregationQuery)
 	return ar, nil
 }
 
-// parseReadOptions translates Query read options into protobuf format.
-func parseReadOptions(q *Query) (*pb.ReadOptions, error) {
-	if t := q.trans; t != nil {
+func validateReadOptions(eventual bool, t *Transaction) error {
+	if t != nil {
```
enocom, line 814: You might try reversing this so the "happy path" isn't indented.

``` golang
if t == nil {
  return nil
}

// t != nil
// ...
```

enocom: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

bhshkh: Thanks. Updated.

enocom (approved):

cindy-peng (approved):

PR #8986, opened 2023-11-08
refactor(datastore): Return transaction id in get

bhshkh:
This PR modifies the get method to return the [transaction id obtained from LookupResponse](https://github.com/googleapis/go-genproto/blob/bbf56f31fb17579209f44dda2d2567d20ff21a20/googleapis/datastore/v1/datastore.pb.go#L249-L256). 
The returned value will be ignored for now but will be used when new_transaction option is passed LookupRequest in future PRs.


enocom (approved):

cindy-peng (approved):
PR looks good Baha! Just wondering, do we usually need to add tests for these kind of changes?

bhshkh:
> PR looks good Baha! Just wondering, do we usually need to add tests for these kind of changes?

Will add tests in the next PR that uses the returned id

danieljbruce (approved):

datastore/transaction.go
```diff
@@ -307,7 +309,13 @@ func (t *Transaction) GetMulti(keys []*Key, dst interface{}) (err error) {
...
 		ConsistencyType: &pb.ReadOptions_Transaction{Transaction: t.id},
 	}
-	return t.client.get(t.ctx, keys, dst, opts)
+
+	// TODO: Use transaction ID returned by get
+	_, err = t.client.get(t.ctx, keys, dst, opts)
+	if err != nil {
+		return err
+	}
+	return nil
```
danieljbruce, line 318: This is just a general Golang question. Would return err instead of the if and then return nil do the same thing?

enocom: Both are fine in my opinion, but you'd probably see `return err` without the `if` more often in cases like this.

bhshkh: Fixed!

datastore/datastore.go
```diff
@@ -426,22 +427,27 @@ func (c *Client) GetMulti(ctx context.Context, keys []*Key, dst interface{}) (er
 		}
 	}
 
-	return c.get(ctx, keys, dst, opts)
+	// TODO: Use transaction ID returned by get
+	_, err = c.get(ctx, keys, dst, opts)
+	if err != nil {
+		return err
+	}
+	return nil
```
enocom, line 435: Here's another example of where we could use `return err`.

bhshkh: Sorry about this. Fixed!

PR #8990, opened 2023-11-08
feat(bigquery): add DataGovernanceType to routines

shollyman:
PR adds the DataGovernanceType field to routine metadata representations.  Testing-wise, this PR also makes test inputs more explicit.

internal issue: 286117986


bigquery/routine_test.go
```diff
@@ -105,55 +113,73 @@ func TestRoutineTypeConversions(t *testing.T) {
...
 				NullFields:      []string{"Arguments", "ImportedLibraries"},
-			}},
-		{"empty", "ToRoutineArgument",
-			&bq.Argument{},
-			&RoutineArgument{}},
-		{"basic", "ToRoutineArgument",
-			&bq.Argument{
+			},
+		},
+		{name: "empty",
```
Linchin, line 150: nit: line break here?

Linchin:
I wonder if it makes sense to add this field to integration tests too?

bigquery/routine_test.go
```diff
@@ -72,9 +72,16 @@ func TestRoutineTypeConversions(t *testing.T) {
 		in         interface{}
 		want       interface{}
 	}{
-		{"empty", "ToRoutineMetadata", &bq.Routine{}, &RoutineMetadata{}},
-		{"basic", "ToRoutineMetadata",
-			&bq.Routine{
+		{
```
Linchin, line 75: It looks like the method `RoutineMetadata.toBQ()` isn't tested. I wonder if we need to add unit tests for it.

bigquery/routine_test.go
shollyman: done.

bigquery/routine_test.go
shollyman: added

Linchin:
Also, I don't think I'm able to approve this PR. I'm not in the go reviewer group.

shollyman:
> I wonder if it makes sense to add this field to integration tests too?

done.

Linchin (approved):
Thank you!

PR #8993, opened 2023-11-08
chore(storage): update apiary dependency

tritone:
This picks up the change to enable a warning message for the directPath fallback to cloud path.


frankyn (approved):
Thanks @tritone

PR #8994, opened 2023-11-09
feat(spanner): add change streams action support in executor framework

harshachinta:
This PR adds support for validating change streams action in executor framework.


codyoss:
@harshachinta is this a change we still want?

codyoss:
@harshachinta see my comment above

PR #9000, opened 2023-11-13
chore(all): bump direct deps

codyoss:
Renovate has not run in some time, so I picked a couple mods and updated thier direct deps and ran `go work sync` to have other mods take the update as well.


noahdietz (approved):

codyoss:
Going to assume the failure on the storge test is a flake. Will trying CI once more though

PR #9006, opened 2023-11-14
feat(logging): automatic project detection in logging.NewClient()

### LINKED ISSUE #7321: logging: automatic project detection in logging.NewClient()
When calling `logging.NewClient()`, the caller has to specify the parent which is assumed to be a project ID if its type is not explicit.

When running in GCE, GCR, GCF or GAE Standard, the obvious default choice for parent would be the same project in which the code runs. At the same time, in those environments, the project ID can be inferred from the environment (see logging/resource.go).

Allow

```
client, err: = logging.NewClient(ctx, "")
```

which will set the parent to the project ID in which the code is running.

Callers of the logging library implement their own automatic project ID detection.

**Additional context**
N/A

gkevinzheng:
Use `detectResource` in `resource.go` to automatically detect the project in `logging.NewClient` if an empty string is passed in as a `parent`.

Fixes #7321


logging/logging.go
```diff
@@ -191,6 +195,13 @@ func NewClient(ctx context.Context, parent string, opts ...option.ClientOption)
 
 func makeParent(parent string) (string, error) {
 	if !strings.ContainsRune(parent, '/') {
+		if parent == "" {
```
bhshkh, line 198: Instead of using empty string, you can consider using a sentinel similar to [Datastore](https://github.com/googleapis/google-cloud-go/blob/afe7c98036c198995075530d4228f1f4ae3f1222/datastore/datastore.go#L109-L145) and [Firestore](https://github.com/googleapis/google-cloud-go/blob/afe7c98036c198995075530d4228f1f4ae3f1222/firestore/client.go#L92-L107). BigQuery and PubSub also use sentinels.

bhshkh (changes requested):

logging/logging.go
gkevinzheng: Done.

bhshkh (approved):

noahdietz (approved):

PR #9007, opened 2023-11-15
chore(storage): minimal bidi changes

BrennaEpp:
This is the code currently being benchmarked. DO NOT MERGE.

Relevant code is in `grpc_client.go` - changes to benchmark code are the same as in #8562, from which this was branched.

This was branched from previous code; it does the minimal changes to use a bi-directional stream.


tritone (commented):
LGTM for testing purposes; not approving because we will not merge this to main.

PR #9008, opened 2023-11-15
chore(storage): remove getstate on last write

BrennaEpp:
This is the code currently being benchmarked. DO NOT MERGE.

The important bit here is the following (the rest is just support for server mode in the benchmarks, same as in https://github.com/googleapis/google-cloud-go/pull/8562, from which this was branched):
```
Flush:       remainingDataFitsInSingleReq && !lastWriteOfEntireObject,
StateLookup: remainingDataFitsInSingleReq && !lastWriteOfEntireObject,
```
On the last request, we do not set flush/statelookup. This means we can't receive until we close the send stream, so I moved that code into an `if` as well.


tritone (commented):
Can you add more details about the motivation here?

BrennaEpp:
> Can you add more details about the motivation here?

Added some details here in addition to what I sent offline.

PR #9013, opened 2023-11-15
fix(storage): set flush and get_state to false on the last write in gRPC

BrennaEpp:
Only receive if get_state was true (ie. `remainingDataFitsInSingleReq && !lastWriteOfEntireObject`) (or after CloseSend) to avoid blocking.

### DIFF
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -1629,8 +1629,8 @@ func (w *gRPCWriter) uploadBuffer(recvd int, start int64, doneReading bool) (*st
 			},
 			WriteOffset: writeOffset,
 			FinishWrite: lastWriteOfEntireObject,
-			Flush:       remainingDataFitsInSingleReq,
-			StateLookup: remainingDataFitsInSingleReq,
+			Flush:       remainingDataFitsInSingleReq && !lastWriteOfEntireObject,
+			StateLookup: remainingDataFitsInSingleReq && !lastWriteOfEntireObject,
 		}
 
 		// Open a new stream if necessary and set the first_message field on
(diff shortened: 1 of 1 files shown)


storage/grpc_client.go
```diff
@@ -1724,31 +1724,31 @@ func (w *gRPCWriter) uploadBuffer(recvd int, start int64, doneReading bool) (*st
 		}
 
 		// Done sending data (remainingDataFitsInSingleReq should == true if we
```
tritone, line 1726: Maybe clarify that we are done sending data in the provided buffer, but there are two separate cases depending on whether there is more data to send in the object overall. It's a little confusing as currently written.

storage/grpc_client.go
```diff
@@ -1724,31 +1724,31 @@ func (w *gRPCWriter) uploadBuffer(recvd int, start int64, doneReading bool) (*st
...
-		}
-		if err != nil {
-			return nil, 0, err
-		}
-
-		// Confirm the persisted data if we have not finished uploading the object.
-		if !lastWriteOfEntireObject {
 			if resp.GetPersistedSize() != writeOffset {
 				// Retry if not all bytes were persisted.
 				writeOffset = resp.GetPersistedSize()
```
tritone, line 1754: Is it actually the case that we no longer need to revc() from the stream if we are ready to finalize the object on this msg?

BrennaEpp: We do still receive (see L1775), but only until after `CloseSend`.

tritone: Ah I see, makes sense.

storage/grpc_client.go
BrennaEpp: Modified the wording, does it make sense now?

tritone (approved):

tritone:
Also, commit type could be `fix` for this PR.

PR #9020, opened 2023-11-16
fix(bigquery): Use correct span name for Query.Read

derekperkins:
This will make it easier to identify in tracing if the fast path is taken or not


derekperkins:
@prash-mi any concerns with this?

codyoss:
@shollyman please take a look.

codyoss:
@shollyman see my comment above

PR #9030, opened 2023-11-17
fix(firestore): Return status code from bulkwriter results

bhshkh:
**Issue**: If Firestore document with id 'xyz' already exists and `bulkwriter.Create` method is used to create a document with the same id, `bulkwriter.Results` method returns an error. When `status.Code` method is used on this error to extract status code, it returns the code Unknown instead of returning AlreadyExists.

i.e. below code prints mismatch error

```
bulkWriterJob, _ := bulkWriter.Create(client.Collection("Coll1").Doc("xyz"), map[string]interface{}{"key1": "val1"})
bulkWriter.Flush()

_, err = bulkWriterJob.Results()
if status.Code(err) != codes.AlreadyExists {
fmt.Printf("Mismatch in error got: %v, want: %q", status.Code(err), codes.AlreadyExists)
}
```

**Cause**: [status.Code](https://github.com/grpc/grpc-go/blob/7935c4f759419cb0ab4cc50070f5487b33a19bb7/status/status.go#L136-L146) can be used to extract [status code](https://pkg.go.dev/google.golang.org/grpc/codes#Code) from gRPC errors.  The error returned by bulkwriter.Results method is not in the grpc error format. Thus, status.Code method couldn't be used for such errors.

**Fix**: Return errors in correct format


datastore/datastore.go
```diff
@@ -589,6 +589,7 @@ func (c *Client) PutMulti(ctx context.Context, keys []*Key, src interface{}) (re
 
 	mutations, err := putMutations(keys, src)
 	if err != nil {
+		fmt.Printf("Error from putMutations\n")
```
gkevinzheng, line 592: Are these print statements debug statements or log statements?

bhshkh: Debug statements. Removed

gkevinzheng (approved):

kolea2 (approved):

PR #9036, opened 2023-11-21
Allow comparing with nil using operators such as "!="

noahmillerj:
Firestore allows comparing with nil using "!=". I verified this in console. However the same query fails when performed with golang api. The enforcement on the operators used when comparing with nil value doesn't make sense.


bhshkh:
Please create an issue demonstrating the bug and link the issue with this PR
Modify commit message to conventional commits https://www.conventionalcommits.org/en/v1.0.0/#summary

bhshkh:
Closing this old PR as there is no linked bug.

PR #9040, opened 2023-11-22
feat(bigtable): allow non-default service account in DirectPath

mohanli-ml:
Bigtable already sets the flag to allow non-default SA in DirectPath on the service side, we also need to set the option on the client side to make this feature work.

Related Spanner PR: https://github.com/googleapis/google-cloud-go/pull/8488.


bhshkh:
The conformance test failures can be ignored. They were added in failed state.

bhshkh:
More info on AllowNonDefaultServiceAccount: https://pkg.go.dev/google.golang.org/api/option/internaloption#AllowNonDefaultServiceAccount

LGTM

bhshkh (approved):

Issue #8651, opened 2023-10-05
firestore: `firestore.Update` should implement `String()`


wjkoh:
I'm disappointed that `firestore.Update` does not implement `String()`, even though its underlying data type, `DocumentTransform_FieldTransform`, already does. 

I have a public function in my project that returns `firestore.Update` and I want to add a unit test for it. Generally, I use the Google `cmp` library to compare what I want and what I got through `cmp.Equal()`. Unfortunately, `firestore.Update` has an unexported field, `t`, that contains all the information. I wish if we could implement `Update.String()` as `func (u Update) String() string { return u.t.String() }`. For now, I'm using my own stringfy function using reflect, but it is unsafe and inconvenient, as below.

```
func Stringfy(u *Update) string {
  rv := reflect.ValueOf(u.Value)
  rv = rv.FieldByName("t")
  dt := (*pb.DocumentTransform_FieldTransform)(rv.UnsafePointer())
  return fmt.Sprintf("Path: %s FieldPath: %s Value: %s", update.Path, update.FieldPath, dt.String())
}
```

noahdietz:
Hi there, `go-cmp` has an option to allow comparison of unexported fields - [`AllowUnexported`](https://pkg.go.dev/github.com/google/go-cmp/cmp#AllowUnexported) - so you don't need to write your own. Please close this issue if it solves the problem.

That said, if `AllowUnexported` *doesn't* solve the problem, keep this open and feel free to open a PR and contribute such a function yourself, it would be greatly appreciated.

Note: Please do not start a feature request with "I'm disappointed that X". This is the second issue I've read that leads with this language. It does not make anyone feel good or excited to solve a problem. I understand you may be frustrated by the lack of some functionality, but we are all busy developers like yourself, and shaming maintainers into implementing a feature isn't kind. Thank you for udnerstanding.

wjkoh:
I chose to use "I'm disappointed ..." because the official template example on this repository was in that format. As seen in the screenshot below, why does it matter how the expression is worded if it accurately conveys the customer's feelings?

![image](https://github.com/googleapis/google-cloud-go/assets/1053445/8766807e-1aab-4e44-bf02-45989fd4ee98)

wjkoh:
When I worked at Google, my team never viewed user reports that expressed their feelings as "shaming". Instead, we saw it as a valuable source of information. Is this Google's new official policy or just your personal opinion?

[...]

PR #9046, opened 2023-11-27
chore: update feature request template language

noahdietz:
As pointed out in https://github.com/googleapis/google-cloud-go/issues/8651#issuecomment-1806740564, the template prompts users to express frustration, when that is unnecessary and can come off as unfriendly. This updates the language to be focused on feedback, and less on expressing a "wrong".


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -8,7 +8,7 @@ assignees: ''
 ---
 
 **Is your feature request related to a problem? Please describe.**
-A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+A clear and concise description of what the problem is. Ex. When doing X, we are unable to Y, because [...]
```
coryan, line 11: "unable to Y" may be too high of a bar.  Maybe the customer is looking for "it would be easier to do Y if the libraries did X".

noahdietz: I like that, thanks @coryan

codyoss (approved):

Issue #8669, opened 2023-10-11
spanner: JSON number decode precision error


harshachinta:
**Issue**
When JSON data contains a number, then decoding it to `NullJson` leads to loss of precision.

**Test case to reproduce**
```
func TestDecodeJsonValue(t *testing.T) {
	nulljson := NullJSON{}
	err := decodeValue(intProto(145688415796432520), jsonType(), &nulljson)
	if err != nil {
		fmt.Println(err)
	}
	fmt.Println(nulljson.Value)
	fmt.Println(nulljson)
}
```
The output of the above test case is
```
=== RUN   TestDecodeJsonValue
1.456884157964325e+17
145688415796432500
```

The expected output is `145688415796432520`, but the returned output is `145688415796432500`.

**Code causing the issue**
In [decodeValue()](https://github.com/googleapis/google-cloud-go/blob/a811f4c49f0c3c769467239d866d4267a9ba2b44/spanner/value.go#L1042) function in [value.go](https://github.com/googleapis/google-cloud-go/blob/main/spanner/value.go), looking at the switch case when interface is [*NullJson](https://github.com/googleapis/google-cloud-go/blob/a811f4c49f0c3c769467239d866d4267a9ba2b44/spanner/value.go#L1548). The code uses,
```
err := json.Unmarshal([]byte(x), &y)
```
The [json](https://pkg.go.dev/encoding/json) package in golang by default considers the numbers in `JSON` as `float64` values. The documentation is attached below, 
![image](https://github.com/googleapis/google-cloud-go/assets/57220027/3e002d69-913f-498a-a0f1-624da06b93ec)

Attaching the debugger output where the number is converted to float by the json package.
![image](https://github.com/googleapis/google-cloud-go/assets/57220027/66393ddf-dbed-4ccb-8aef-3ce676aa1dba)

**References**
https://stackoverflow.com/questions/16946306/preserve-int64-values-when-parsing-json-in-go
https://stackoverflow.com/questions/24480835/decoding-json-int-into-string
https://github.com/golang/go/issues/42246

PR #9054, opened 2023-11-29
fix(spanner): use json.Number for decoding unknown values from spanner

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/8669

Go client library uses native json library(encoding/json) for marshalling/unmarshalling for decoding JSON spanner types.
Since native package maps values to float64 it automatically rounds off the values.

Example 
```
package main

import (
	"log"
	"context"

	"cloud.google.com/go/spanner"
)

func getFloatValue(ctx context.Context, client *spanner.Client) {
	iter := client.Single().Query(ctx, spanner.NewStatement("select json '0.39240506000000003'"))
	for {
		row, err := iter.Next()
		if err == iterator.Done {
			break
		}
		if err != nil {
			panic(err)
		}
		var f spanner.NullJSON
		if err := row.Column(0, &f); err != nil {
			panic(err)
		}
		log.Printf("%v\n", f)
	}
}

func main() {
	ctx := context.Background()
	dsn := "projects/my-project/instances/my-instance/databases/my-database"
	client, err := spanner.NewClient(ctx, dsn)
	if err != nil {
		log.Fatal(err)
	}
	defer client.Close()
	getFloatValue(ctx, client)
}
```

will decode the value as `0.39240506`.
Similarly when reading `145688415796432520` will be decoded as `145688415796432500` OR `1.456884157964325e+17` 

With this change Go client library will parse the JSON number values to same precision as stored in Spanner database using [jsoniter](https://github.com/json-iterator/go) library using [json.Number](https://pkg.go.dev/encoding/json#Number) when **UseNumberWithJSONDecoderEncoder()** is called from the application 
### NOTE
```
Package jsoniter implements encoding and decoding of JSON as defined in RFC 4627 and provides interfaces with identical syntax of standard lib encoding/json. Converting from encoding/json to jsoniter is no more than replacing the package with jsoniter and variable type declarations (if any). jsoniter interfaces gives 100% compatibility with code using standard lib.
"JSON and Go" (https://golang.org/doc/articles/json_and_go.html ) gives a description of how Marshal/Unmarshal operate between arbitrary or predefined json objects and bytes, and it applies to jsoniter.Marshal/Unmarshal as well.
```


spanner/value.go
```diff
@@ -113,6 +114,10 @@ var (
 	commitTimestamp = time.Unix(0, 0).In(time.FixedZone("CommitTimestamp placeholder", 0xDB))
 
 	jsonNullBytes = []byte("null")
+
+	jsonProvider jsoniter.API
+
+	once sync.Once
```
olavloite, line 120: If I understand this correctly, it means that `jsonProvider` is set once as a global variable the first time a client is created. That means that if you create a new client later with a different JSON decoding configuration, it will have no effect. Although it is not a very common use case, it is also very confusing. If we can only provide this as a global variable, then we should model it as such in the API as well, and clearly call out how it should be used.

spanner/client.go
```diff
@@ -186,6 +187,10 @@ type ClientConfig struct {
 
 	// BatchTimeout specifies the timeout for a batch of sessions managed sessionClient.
 	BatchTimeout time.Duration
+
+	// UseNumber causes the Decoder to unmarshal a number into an interface{} as a
+	// Number instead of as a float64.
```
olavloite, line 192: I'm only able to understand this comment because I know the context of this PR. Without that context, this comment contains too little information. Can we add:
1. Additional context that explains that this is for the JSON decoder. (`the Decoder` is not a clear reference to JSON decoding)
2. What it means that something is being unmarshalled as a `Number` instead of a `float64`? So what are the pros/cons of each?

So for example:

UseNumber specifies whether number values inside a Cloud Spanner JSON value should be decoded as a Number or a float64. Decoding to a Number guarantees that the precision used by Cloud Spanner is preserved. Decoding to a float64 can cause loss of precision. The default JSON decode function in Go uses float64. This is therefore also the default used by this client library. Change this value to `true` to prevent loss of precision.

spanner/client.go
```diff
@@ -265,7 +270,14 @@ func NewClientWithConfig(ctx context.Context, database string, config ClientConf
 	if config.BatchTimeout == 0 {
 		config.BatchTimeout = time.Minute
 	}
-
+	once.Do(func() {
+		// Initialize json provider.
+		jsonProvider = jsoniter.Config{
```
olavloite, line 275: See also my comment below: This will be confusing for anyone who calls `NewClientWithConfig` first with `UseNumber: false` and then later again with `UseNumber: true`. The configuration appears to be for a single client, but is in reality a global variable. We need to change that so it is either clearer, or it is actually a client config. Possible solutions for that could be:

1. Add a global method that sets the preferred decode method. Clearly document that the value that is set will be used by all clients, including both clients that have already been created and clients that will be created in the future. Based on where the configuration option is needed, this is probably the best/only possible solution.
2. OR: Add overloaded functions that accept an argument that determines which decoding method should be used wherever possible (this would not have my preference).
3. OR (I don't think it is possible, but for completeness): Pass a reference to the config on to all objects that need to know, and use the value there.

rahul2393: I pushed changes for approach 1.

spanner/integration_test.go
```diff
@@ -2083,150 +2096,162 @@ func TestIntegration_BasicTypes(t *testing.T) {
...
-		statements = append(statements, stmt)
-	}
-	_, err := client.ReadWriteTransaction(ctx, func(ctx context.Context, tx *ReadWriteTransaction) error {
-		rowCounts, err := tx.BatchUpdate(ctx, statements)
-		if err != nil {
-			return err
+	for idx := 0; idx < 2; idx++ {
+		if idx == 1 {
+			UseNumberWithJSONDecoderEncoder()
+			defer testSetJSONProviderNumberConfig(false)
```
olavloite, line 2175: nit: can we change this into
1. First get the current value of `UseNumberWithJSONDecoderEncoder`
2. A loop that loops through two boolean value true and false
3. Then calls `UseNumberWithJSONDecoderEncoder(boolValue)`
4. Then after the loop set the `UseNumberWithJSONDecoderEncoder` to the original value.

olavloite: Also: Are we actually using this to check whether we get the expected value? In other words: I don't see any tests that are conditional on whether `idx == 1` or `idx == 0`, and that has a different expected outcome depending on that.

spanner/value.go
```diff
@@ -113,8 +114,26 @@ var (
...
+	jsonProvider = jsoniter.ConfigCompatibleWithStandardLibrary
+
+	once sync.Once
 )
 
+// UseNumberWithJSONDecoderEncoder specifies whether Cloud Spanner JSON numbers are decoded
+// as Number (preserving precision) or float64 (risking loss).
+// Defaults to float64, call this method for lossless precision.
+// NOTE: This change affects all clients created by this library, both existing and future ones.
+func UseNumberWithJSONDecoderEncoder() {
```
olavloite, line 127: This API means that you can never disable it after having enabled it. I don't think we have a good reason for that restriction, or? I can imagine that making it possible to both enable and disable the option will make it easier for users to use it in their own tests to determine what the behavior is.

So I think that we should add an argument to the method (true/false), or add a separate method for disabling the option.

spanner/value.go
```diff
@@ -113,8 +114,26 @@ var (
...
+
+	jsonProvider = jsoniter.ConfigCompatibleWithStandardLibrary
+
+	once sync.Once
 )
 
+// UseNumberWithJSONDecoderEncoder specifies whether Cloud Spanner JSON numbers are decoded
+// as Number (preserving precision) or float64 (risking loss).
+// Defaults to float64, call this method for lossless precision.
+// NOTE: This change affects all clients created by this library, both existing and future ones.
```
olavloite, line 126:
```suggestion
// Defaults to the same behavior as the standard Go library, which means decoding to float64.
// Call this method to enable lossless precision.
// NOTE 1: Calling this method affects the behavior of all clients created by this library, both existing and future instances.
// NOTE 2: This method sets a global variable that is used by the client to encode/decode JSON numbers. Access to the global variable is not synchronized. You should only call this method when there are no goroutines encoding/decoding Cloud Spanner JSON values. It is recommended to only call this method during the initialization of your application, and preferably before you create any Cloud Spanner clients, and/or in tests when there are no queries being executed.
```

olavloite (commented):
I think we should make it possible for customers to both enable and disable the feature, as it is possible that they want to test their code with both options.

spanner/value.go
rahul2393: Allowing users to enable/disable the feature means they can run into situation where Client 1(supposed to run without Number) running in thread1 get impacted by some thread2 which uses another Client2 and call UseNumberWithJSONDecoderEncoder(), are we ok with it?

spanner/integration_test.go
rahul2393: There is condition on whether idx is 0/1, https://github.com/googleapis/google-cloud-go/pull/9054/files#diff-3303adad40b1fa3c6b0a92ea1f4ad28c32bc84c52000e7a5e9fc8b37e05eaa85R2226-R2229

olavloite (approved):
LGTM, thanks for being patient with my many small requests :-)

spanner/value.go
olavloite: That was already the case, however only in the direction `default behavior => new behavior`. Now we allow this to happen both ways.

What I mean is that the original implementation also allowed the following:
1. A client is created. This client is assumed to use the default Go encoding/decoding.
2. The flag is set to true and a new client is created.
3. Both the old client and the new client now use the new behavior.
4. It was however not possible to go back to the default behavior.

PR #9072, opened 2023-12-04
feat(storage): add object retention feature

### LINKED ISSUE #8929: storage: add support for object retention
Please implement the object retention feature. The following operations must be supported:

- Create a bucket with object retention enabled by using the `enableObjectRetention` bucket create boolean option at bucket create time
- Get the object retention enabled status of a bucket
- Create an object with retention enabled (on an retention enabled bucket) by populating the `retention` object at object create time (retention object has the fields `mode` which can be Unlocked or Locked, and `retainUntilTime` which is a timestamp)
- Update retention for an existing object with a patch operation, by passing the `overrideUnlockedRetention` boolean patch option
- Get the retention policy of an object
- Remove retention for an existing object the same way as above

Details in go/gcs-object-retention-client-request

You can use the [java implementation](https://github.com/googleapis/java-storage/pull/2277) as a reference.

The kokoro test project should be allowlisted shortly, let me know if you need access to an allowlisted project for your own testing. 

We're aiming to get this in by the November code freeze, let me know if you need any help.

BrennaEpp:
- API is not yet available in gRPC. This code ignores any retention setting attempt and will never return retention if using the gRPC API.

- Updated `EnableObjectRetention()` to be a method on the bucket handle as per discussion w/ @tritone 

Fixes #8929


JesseLovelace (approved):

storage/grpc_client.go
```diff
@@ -162,6 +162,8 @@ func (c *grpcStorageClient) CreateBucket(ctx context.Context, project, bucket st
 		b.Location = "US"
 	}
 
+	// TO-DO: implement ObjectRetention once available - see b/308194853
```
tritone, line 165: I think for now we should return a NotImplemented error if a user tries to pass this, so that it doesn't just silently not happen.

storage/storage.go
```diff
@@ -973,8 +973,9 @@ func (o *ObjectHandle) ObjectName() string {
 
 // ObjectAttrsToUpdate is used to update the attributes of an object.
 // Only fields set to non-nil values will be updated.
-// For all fields except CustomTime, set the field to its zero value to delete
-// it. CustomTime cannot be deleted or changed to an earlier time once set.
+// For all fields except CustomTime and Retention, set the field to its zero
+// value to delete it. CustomTime cannot be deleted or changed to an earlier
+// time once set. Retention follows specidal
```
tritone, line 978: Looks like this needs to be completed.

storage/storage.go
```diff
@@ -999,6 +1000,15 @@ type ObjectAttrsToUpdate struct {
...
 	// See https://cloud.google.com/storage/docs/json_api/v1/objects/patch.
 	PredefinedACL string
+
+	// Retention contains the retention configuration for this object.
+	Retention *ObjectRetention
+
+	// OverrideUnlockedRetention provides an option for overriding an Unlocked
+	// Retention policy. This must be set to true in order to change a policy
+	// from Unlocked to Locked, to set it to null, or to reduce its
+	// retainUntilTime attribute.
```
tritone, line 1010: I think this could be a little clearer; is sending Retention and not OverrideUnlockedRetention ever valid? It's not obvious to me from reading this.

Overall having this field still seems quite clunky to me; but we can discuss offline and/or get a consult from the product team. If we really want to include it, it seems best to have it as a method on ObjectHandle as we do with other query params to the JSON API.

storage/bucket_test.go
```diff
@@ -621,6 +621,9 @@ func TestNewBucket(t *testing.T) {
 			RetentionPeriod: 3,
 			EffectiveTime:   aTime.Format(time.RFC3339),
 		},
+		ObjectRetention: &raw.BucketObjectRetention{
```
tritone, line 624: This should probably get added to `TestBucketAttrsToRawBucket` as well?

storage/bucket.go
```diff
@@ -462,6 +464,14 @@ type BucketAttrs struct {
 	// allows for the automatic selection of the best storage class
 	// based on object access patterns.
 	Autoclass *Autoclass
+
+	// ObjectRetentionMode reports whether individual objects in the bucket can
+	// be configured with a retention policy. An empty value means that object
+	// retention is disabled. Object retention is not enabled default.
+	// This field is read-only. Object retention can be enabled only on bucket
```
tritone, line 471: Read-only seems not really correct given that it can be set on bucket creation, but I think you can just delete that sentence and the rest of the doc is fine.

storage/integration_test.go
```diff
@@ -3888,6 +3888,261 @@ func TestIntegration_LockBucket_MetagenerationRequired(t *testing.T) {
...
+				desc:              "ObjectRetentionMode is not enabled by default",
+				wantRetentionMode: "",
+			},
+			{
+				desc:                "Setting just ObjectRetentionMode should not work",
+				objectRetentionMode: "Enabled",
+				wantRetentionMode:   "",
+			},
+			{
+				desc:                "ObjectRetentionMode is ignored",
```
tritone, line 3912: I'm really confused by this case (and somewhat by the case above as well) and it seems very counterintuitive behavior. We should discuss with the product team.

storage/integration_test.go
```diff
@@ -3888,6 +3888,261 @@ func TestIntegration_LockBucket_MetagenerationRequired(t *testing.T) {
...
+			RetainUntil: time.Now().Add(time.Minute * 20).Truncate(time.Second),
+		}
+		retentionUnlockedExtended := &ObjectRetention{
+			Mode:        "Unlocked",
+			RetainUntil: time.Now().Add(time.Hour).Truncate(time.Second),
+		}
+		retentionLockedExtended := &ObjectRetention{
+			Mode:        "Locked",
+			RetainUntil: time.Now().Add(time.Hour).Truncate(time.Second),
+		}
```
tritone, line 3984: Do all these retention times pose an issue for being able to clean up the buckets created by this test?

tritone (commented):
Overall looks pretty good, thanks for the extensive integration tests. I have a few questions/concerns as noted.

storage/storage.go
BrennaEpp: Thanks for catching that, updated

storage/storage.go
BrennaEpp: As discussed offline, since it's a permanent change that can have big consequences (not being able to delete the object, or allowing the object to be deleted before intended), we are required to provide an extra level of "are you sure?" for certain changes to the Retention. 

I believe we do have other query params as fields on the update attrs, but in this case it is just a param that we send and do not report back in any way, so you're right, it makes sense to have this as a method. I updated this.

storage/integration_test.go
BrennaEpp: I discussed this with Jesse, we don't need to explicitly test Locked retentions since they cause these cleanup issues. I've removed them and everything should clean up appropriately now.

tritone: Excellent, thanks for confirming.

storage/grpc_client.go
tritone: I meant the gRPC unimplemented error code: see https://pkg.go.dev/google.golang.org/grpc@v1.60.0/codes#Unimplemented . I think we can just make and return this.

storage/bucket.go
```diff
@@ -1362,8 +1362,8 @@ func (b *BucketHandle) LockRetentionPolicy(ctx context.Context) error {
 }
 
 // SetObjectRetention returns a new BucketHandle that allows enabling object
```
tritone, line 1364: Maybe rephrase to make it clearer that calling the method means that object retention WILL be enabled on the created bucket (not "allows enabling"). Also, clarify that this has no effect on an already existing bucket.

storage/bucket.go
```diff
@@ -467,9 +467,10 @@ type BucketAttrs struct {
 
 	// ObjectRetentionMode reports whether individual objects in the bucket can
 	// be configured with a retention policy. An empty value means that object
-	// retention is disabled. Object retention is not enabled default.
-	// This field is read-only. Object retention can be enabled only on bucket
-	// creation on a bucket handle with SetObjectRetention set to true.
+	// retention is disabled.
+	// This field is read-only. Object retention can be enabled only by creating
+	// a bucket with SetObjectRetention set to true on the BucketHandle. It
+	// cannot be modified.
```
tritone, line 473: ...cannot be modified once the bucket is created.

storage/client.go
```diff
@@ -60,7 +60,7 @@ type storageClient interface {
 
 	DeleteObject(ctx context.Context, bucket, object string, gen int64, conds *Conditions, opts ...storageOption) error
 	GetObject(ctx context.Context, bucket, object string, gen int64, encryptionKey []byte, conds *Conditions, opts ...storageOption) (*ObjectAttrs, error)
-	UpdateObject(ctx context.Context, bucket, object string, uattrs *ObjectAttrsToUpdate, gen int64, encryptionKey []byte, conds *Conditions, opts ...storageOption) (*ObjectAttrs, error)
+	UpdateObject(ctx context.Context, params *updateObjectParams, opts ...storageOption) (*ObjectAttrs, error)
```
tritone, line 63: Nice, this is a good change.

storage/grpc_client.go
```diff
@@ -152,7 +152,12 @@ func (c *grpcStorageClient) GetServiceAccount(ctx context.Context, project strin
 	return resp.EmailAddress, err
 }
 
-func (c *grpcStorageClient) CreateBucket(ctx context.Context, project, bucket string, attrs *BucketAttrs, _ *bool, opts ...storageOption) (*BucketAttrs, error) {
+func (c *grpcStorageClient) CreateBucket(ctx context.Context, project, bucket string, attrs *BucketAttrs, enableObjectRetention *bool, opts ...storageOption) (*BucketAttrs, error) {
+	if enableObjectRetention != nil {
+		// TO-DO: implement ObjectRetention once available - see b/308194853
+		return nil, errors.New("storage: object retention is not supported in gRPC")
```
tritone, line 158: Replace with codes.Unimplemented if possible. Same elsewhere that you are returning this error. https://pkg.go.dev/google.golang.org/grpc@v1.60.0/codes#Unimplemented

storage/storage.go
```diff
@@ -975,15 +984,17 @@ func (o *ObjectHandle) ObjectName() string {
...
-// For example, to change ContentType and delete ContentEncoding and
-// Metadata, use
+// For example, to change ContentType and delete ContentEncoding, Metadata and
+// Retention, use:
 //
 //	ObjectAttrsToUpdate{
 //	    ContentType: "text/html",
 //	    ContentEncoding: "",
 //	    Metadata: map[string]string{},
+//		Retention: &ObjectRetention{},
```
tritone, line 997: extra tab here I think

storage/storage.go
```diff
@@ -1030,6 +1038,17 @@ func (o *ObjectHandle) ReadCompressed(compressed bool) *ObjectHandle {
 	return &o2
 }
 
+// OverrideUnlockedRetention provides an option for overriding an Unlocked
+// Retention policy. This must be set to true in order to change a policy
+// from Unlocked to Locked, to set it to null, or to reduce its
+// RetainUntil attribute. It is not required for setting the ObjectRetention for
+// the first time nor for extending the RetainUntil time.
+func (o *ObjectHandle) OverrideUnlockedRetention(override bool) *ObjectHandle {
```
tritone, line 1046: What does sending false for this query parameter do? (Same question applies for the bucket-level one as well.)

I'm wondering if we should remove the arg from these methods and just set to true if the user calls them. Thoughts?

tritone (commented):
Overall looking really good; a few nits and minor questions.

storage/storage.go
BrennaEpp: I don't think it does anything now, but I am hesitant to change it. I think we want to /really/ make sure users want to use these query params; in my opinion having the bool passed in adds a small extra step that doesn't really impact user experience negatively very much.

storage/storage.go
BrennaEpp: tab instead of spaces 😶‍🌫️

storage/storage.go
tritone: Sure, I guess it also mirrors the API itself more closely so there doesn't seem to be much harm either way.

tritone (approved):
Nice work on this!

PR #9080, opened 2023-12-05
test(datastore): Add explicit sleep before read time use

bhshkh:
The read time in this test is current time truncated to microseconds.
If there is a difference between the Datastore service clock and the test runner, the test fails with error:

```
=== RUN   TestIntegration_AggregationQueriesInTransaction
    retry.go:44: FAILED after 10 attempts:
        integration_test.go:818: client.NewTransaction: rpc error: code = InvalidArgument desc = The requested 'read_time' cannot be in the future.
--- FAIL: TestIntegration_AggregationQueriesInTransaction (100.71s)
```

Adding explicit sleep as part of this PR to avoid this sceanrio.


gkevinzheng (approved):
LGTM

datastore/integration_test.go
```diff
@@ -811,6 +811,10 @@ func TestIntegration_AggregationQueriesInTransaction(t *testing.T) {
 			if readTime.IsZero() {
 				// Use current time as read time if read time is not specified in test case
 				readTime = time.Now().Truncate(time.Microsecond)
+
+				// Read time is truncated to microseconds. If immediately used in NewTransaction call,
+				// it leads to "read_time cannot be in the future" error
+				time.Sleep(time.Second)
```
cindy-peng, line 817: Should the sleep time here be "Microsecond" or "Second" since the read time is truncated to microseconds?

bhshkh: It can be microsecond but being safe here to prevent failures if and when deviation between service clock and test runner clock is more

kolea2 (approved):

PR #9082, opened 2023-12-06
test(storage): fix flaky RequesterPays test

### LINKED ISSUE #9003: storage: TestIntegration_RequesterPaysNonOwner failed
Note: #8934 was also for this test, but it was closed more than 10 days ago. So, I didn't mark it flaky.

----

commit: ca2493f43c299bbaed5f7e5b70f66cc763ff9802
buildURL: [Build Status](https://source.cloud.google.com/results/invocations/4c165c2a-d883-4d2a-8f77-9abb6f426e0c), [Sponge](http://sponge2/4c165c2a-d883-4d2a-8f77-9abb6f426e0c)
status: failed

BrennaEpp:
Add retry on 400 and 403 to object deletes

I believe the error was occuring due to the metadata changes not being propagated yet. Other operations are done on the `o` ObjectHandle, which will retry a operation if it encounters a 400 or 403, but these deletes are not retried in that case because they are on a different handle.

Fixes #9003


storage/integration_test.go
```diff
@@ -5307,6 +5312,27 @@ func (h testHelper) mustRead(obj *ObjectHandle) []byte {
 	return data
 }
 
+// deleteObjectIfExists deletes an object with a RetryAlways policy (unless another
```
tritone, line 5315: It seems like we should maybe be doing this for object cleanup in general? Wondering if there are other tests which should also use this method.

tritone (approved):

storage/integration_test.go
BrennaEpp: I believe we have a similar method that does a lot of our object cleanups, perhaps I can merge the two.

tritone: Feel free, but it can be in a separate PR if that's easier

BrennaEpp: Sounds good, I will merge this as-is to get the test fix in.

PR #9109, opened 2023-12-08
feat(vertexai): add CountTokens

jba:
Add support for the CountTokens RPC.


vertexai/genai/count_tokens.go
eliben, line 1: What is this file for?

jba: Leftover. Will remove.

PR #9113, opened 2023-12-09
feat(vertexai): add function support

jba:
A client can pass a list of functions to the model which the model will call if it determines they can help with a response. The client must then provide the function result in its next prompt, and the model will attempt to use it.


vertexai/internal/support/support.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	return civil.Date{
+		Year:  int(p.Year),
+		Month: time.Month(p.Month),
+		Day:   int(p.Day),
+	}
+}
+
+// MapToStructPB c
+func MapToStructPB(m map[string]any) *structpb.Struct {
```
eliben, line 95: The doc string is incomplete

eliben (approved):

vertexai/internal/support/support.go
```diff
@@ -0,0 +1,112 @@
...
+		return nil
+	}
+	to := map[K]VTo{}
+	for k, v := range from {
+		to[k] = f(v)
+	}
+	return to
+}
+
+// ZeroToNil returns nil if x is the zero value for T,
```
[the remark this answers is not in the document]
jba: I guess we could pass a zero valued pointer to the proto. Also happy to rename. Let me do those after I land the PR with the tool.

vertexai/internal/support/support.go
jba: Done.

Issue #9034, opened 2023-11-21
bigtable: Get underlying protobuf of mutation ops


coxley:
Ref: https://github.com/googleapis/google-cloud-go/blob/main/bigtable/bigtable.go#L878

We have a handful of systems that make liberal use of BigTable. To make replaying, diagnosing, or dry-running previous events possible, I want to record the database decisions made for a given input.

With traditional databases, I'd record the rendered SQL template(s). I don't exactly have that with BigTable — but internally this library has it. Are the operations unexported explicitly, or just because this use-case hasn't come up?

To get around this, I'm going to embed `*bigtable.Mutation` and do my own recording. But our stuff can be high volume, and I'd love to not need to allocate extra things forever.

```go
type Mutation struct {
	*bigtable.Mutation
	// Some internal tracking here
}

func (m *Mutation) Set(family, column string, ts bigtable.Timestamp, value []byte) {
	m.Mutation.Set(family, column, ts, value)
}

func (m *Mutation) DeleteCellsInColumn(family, column string) {
	m.Mutation.DeleteCellsInColumn(family, column)
}

```

noahdietz:
@coxley sounds like a reasonable request. One question: Would it be sufficient to return a _copy_ of the `Ops` instead of handing over a reference? I'd worry about handing over an internal implementation detail to users to manipulate --> foot gun.

Feel free to send a PR for @bhshkh to review with this feature request, I'm just triaging open issues :)

coxley:
@noahdietz That's a very fair and valid compromise. Will still allocate extra stuff, but avoid needing to reinvent types.

I'll try to take another look at this after vacation. :)

igorbernstein2:
Hi,

I'm not sure if this is the best path forward. Cloud Bigtable has 2 features that are designed specifically for this usecase. Have you explored [Audit Logs](https://cloud.google.com/bigtable/docs/audit-logging) and [Change Stream](https://cloud.google.com/bigtable/docs/change-streams-overview).

In general its not possible for clients to audit the changes accurately (ie using operations like CheckAndMutate and ReadModifyWrite can have non deterministic results if multiple clients are modifying the same row). Change streams is similar to WAL logs of SQL and designed for the explicit purpose of getting a very accurate view of whats stored in Bigtable. Audit logs is meant more for regulatory purposes, but contains the entire payload of the RPC

[...]

PR #9114, opened 2023-12-09
bigtable: Expose mutation protobufs via mut.Ops()

coxley:
# Summary

Context: #9034

Recording or analyzing database operations can be valuable. Think recording historical requests and replaying them locally — or in a shadow tier — and comparing what actions were taken.

This isn't possible when using the current BigTable SDK. 

We add two functions in this PR to address that: `mut.Ops()` and `mut.OpsValues()`.

## Why two?

To avoid unexpected behavior from the client mutating results from `mut.Ops()` before `table.Apply()`, we need to clone the underlying protobufs. The amount of extra allocations will grow with the number of staged operations.

Except for `mut.Set()` operations.

At least in my use cases, we write fairly large blobs to cells. For basic replay data, I don't need a copy of the bytes I've already written — I already have access to them. I'd rather save the extra alloc space.

Assuming this is a sensible default, `mut.Ops()` will zero out `btpb.Mutation_SetCell.Value`. The unmodified version can be retrieved from `mut.OpsValues()`

## Sensible default?

I was going to flip the behavior around but `mut.OpsNoValues` or `mut.OpsWithoutValues` feels clunky. 

I'm not tied to it — happy to make the change if someone has better names. :)

# Test Plan

```
➜ go test ./...
?       cloud.google.com/go/bigtable/cmd/emulator       [no test files]
?       cloud.google.com/go/bigtable/internal   [no test files]
?       cloud.google.com/go/bigtable/internal/conformance       [no test files]
?       cloud.google.com/go/bigtable/internal/mockserver        [no test files]
?       cloud.google.com/go/bigtable/internal/option    [no test files]
?       cloud.google.com/go/bigtable/internal/stat      [no test files]
ok      cloud.google.com/go/bigtable    9.709s
ok      cloud.google.com/go/bigtable/bttest     0.133s
ok      cloud.google.com/go/bigtable/internal/testproxy 0.028s
```

Benchmark, assuming writing a single cell value of 4K:

```
BenchmarkMutationOps/ops-8                877124              1356 ns/op             904 B/op         19 allocs/op
BenchmarkMutationOps/opsvalues-8          484908              2467 ns/op            5008 B/op         21 allocs/op
```

### DIFF
--- a/bigtable/bigtable.go
+++ b/bigtable/bigtable.go
@@ -44,8 +44,10 @@ import (
 	_ "google.golang.org/grpc/balancer/rls"
 )
 
-const prodAddr = "bigtable.googleapis.com:443"
-const mtlsProdAddr = "bigtable.mtls.googleapis.com:443"
+const (
+	prodAddr     = "bigtable.googleapis.com:443"
+	mtlsProdAddr = "bigtable.mtls.googleapis.com:443"
+)
 
 // Client is a client for reading and writing data to tables in an instance.
 //
@@ -697,9 +699,12 @@ func makeFullReadStats(reqStats *btpb.RequestStats) FullReadStats {
 			CellsReturnedCount: readStats.CellsReturnedCount,
 			CellsSeenCount:     readStats.CellsSeenCount,
 			RowsReturnedCount:  readStats.RowsReturnedCount,
-			RowsSeenCount:      readStats.RowsSeenCount},
+			RowsSeenCount:      readStats.RowsSeenCount,
+		},
 		RequestLatencyStats: RequestLatencyStats{
-			FrontendServerLatency: latencyStats.FrontendServerLatency.AsDuration()}}
+			FrontendServerLatency: latencyStats.FrontendServerLatency.AsDuration(),
+		},
+	}
 }
 
 // FullReadStatsFunc describes a callback that receives a FullReadStats for evaluation.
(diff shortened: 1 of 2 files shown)


coxley:
I was asked to request review from @bhshkh in #9034 — I don't have permission to edit reviewers on the PR so mentioning here. :)

bigtable/bigtable.go
```diff
@@ -697,9 +699,12 @@ func makeFullReadStats(reqStats *btpb.RequestStats) FullReadStats {
 			CellsReturnedCount: readStats.CellsReturnedCount,
 			CellsSeenCount:     readStats.CellsSeenCount,
 			RowsReturnedCount:  readStats.RowsReturnedCount,
-			RowsSeenCount:      readStats.RowsSeenCount},
+			RowsSeenCount:      readStats.RowsSeenCount,
```
coxley, line 702: Whoops, format on save. I can revert these parts if we prefer to reduce out-of-context lines.

bhshkh:
This is on my radar. Will review it soon.

bhshkh:
LGTM but would be great to have a review from @mutianf or @igorbernstein2  as well

bhshkh (approved):

igorbernstein2 (changes requested):
As mentioned in #9034, Im very hesitant about exposing the underlying protobufs to end users. Lets continue the conversation in that issue before proceeding here

PR #9123, opened 2023-12-11
chore(vertexai): protoveneer tool and config

jba:
Add the tool that is used to generate the "wrappers" for the protos, and its config for the genai package.

(I prefer to call these types "veneers" because a "wrapper" contains the thing being wrapped, and these types do not contain the protos. I decided to resurrect the old name for these clients, because it's an appropriate metaphor.)


eliben (approved):

vertexai/internal/cmd/protoveneer/support/support.go
```diff
@@ -0,0 +1,112 @@
+// Copyright 2023 Google LLC
```
eliben, line 1: Is this the same file as the top-level support? I suppose the internal tool can use the package's support?

eliben (approved):

vertexai/internal/cmd/protoveneer/support/support.go
jba: Yes, the idea is that whoever is using the tool would copy this directory to where they needed it. Maybe someday it will have a stable import path.

For now I can delete this. The tool itself doesn't use it.

PR #9151, opened 2023-12-13
fix(vertexai): set internal headers

codyoss:
Follow up to #9150


vertexai/internal/version.go
```diff
@@ -0,0 +1,18 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+// Version is the current tagged release of the library.
+const Version = "0.3.0"
```
eliben, line 18: Is this something we'll have to keep in sync with the releases? The releases are done by the release-please bot, so should we send a chore PR following each release to update this?

codyoss: release-please knowns to look for a file named version.go in the internal directory off the root of a module and will bump it in the release PRs :)

eliben (approved):

PR #9155, opened 2023-12-13
feat(vertexai): add UsageMetadata

jba:
- Add the UsageMetadata type, and the field of GenerateContentResponse that holds it.

- Use protoveneer for GenerateContentResponse instead of writing it by hand.

- Enable protoveneer to completely replace a doc comment.

- Improve protoveneer doc comment handling.


eliben (approved):
Nice to have it generate more of the wrappers!

Issue #9160, opened 2023-12-13
genai: func NewClient does not use opts ...option.ClientOption param


awenborar:
**Code**
https://github.com/googleapis/google-cloud-go/blob/main/vertexai/genai/client.go#L47

```
func NewClient(ctx context.Context, projectID, location string, opts ...option.ClientOption) (*Client, error) {
	apiEndpoint := fmt.Sprintf("%s-aiplatform.googleapis.com:443", location)
	c, err := aiplatform.NewPredictionClient(ctx, option.WithEndpoint(apiEndpoint))
	if err != nil {
		return nil, err
	}
	return &Client{
		c:         c,
		projectID: projectID,
		location:  location,
	}, nil
}
```

**Implication**
The following code does not generate valid authentication:
```
client, err := genai.NewClient(ctx, "project_id", "region", option.WithCredentialsFile("../cred.json"))
```

**Solution**
Use the `opts` param
```
func NewClient(ctx context.Context, projectID, location string, opts ...option.ClientOption) (*Client, error) {
	apiEndpoint := fmt.Sprintf("%s-aiplatform.googleapis.com:443", location)
	opts = append(opts, option.WithEndpoint(apiEndpoint))
	c, err := aiplatform.NewPredictionClient(ctx, opts...)
	if err != nil {
		return nil, err
	}
	return &Client{
		c:         c,
		projectID: projectID,
		location:  location,
	}, nil
}
```

eliben:
Thanks for the report! This should roll out in a bugfix version soon - #9166

PR #9163, opened 2023-12-13
fix(vertexai): passthrough user opts

codyoss:
Fixes: #9160


eliben (approved):
Ah, seems like we raced with a fix :-)

Thanks, Cody. I'll drop the other one

PR #9165, opened 2023-12-13
fix(vertexai): use opts in NewClient

### LINKED ISSUE #9160: genai: func NewClient does not use opts ...option.ClientOption param

eliben:
Fixes #9160

### DIFF
--- a/vertexai/genai/client.go
+++ b/vertexai/genai/client.go
@@ -47,7 +47,8 @@ type Client struct {
 // package.
 func NewClient(ctx context.Context, projectID, location string, opts ...option.ClientOption) (*Client, error) {
 	apiEndpoint := fmt.Sprintf("%s-aiplatform.googleapis.com:443", location)
-	c, err := aiplatform.NewPredictionClient(ctx, option.WithEndpoint(apiEndpoint))
+	opts = append(opts, option.WithEndpoint(apiEndpoint))
+	c, err := aiplatform.NewPredictionClient(ctx, opts...)
 	if err != nil {
 		return nil, err
 	}


eliben:
Closing in favor of #9163 that was earlier

vertexai/genai/client.go
```diff
@@ -47,7 +47,8 @@ type Client struct {
 // package.
 func NewClient(ctx context.Context, projectID, location string, opts ...option.ClientOption) (*Client, error) {
 	apiEndpoint := fmt.Sprintf("%s-aiplatform.googleapis.com:443", location)
-	c, err := aiplatform.NewPredictionClient(ctx, option.WithEndpoint(apiEndpoint))
+	opts = append(opts, option.WithEndpoint(apiEndpoint))
```
jba, line 50: Appending to opts could affect the caller.
Also we want the user opts to override whatever we do.
Follow this pattern: https://github.com/googleapis/google-cloud-go/blob/main/bigquery/storage_client.go#L59.

jba:
Thanks for doing this, sorry I missed it.

You can test by passing in your own endpoint option that's wrong, and looking for an error.

vertexai/genai/client.go
eliben: Thanks! can you add these comments on #9163 ?

jba: Not necessary, #9163 is correct.

Issue #9167, opened 2023-12-13
genai: In NewClient() c.SetGoogleClientInfo undefined (type *aiplatform.PredictionClient has no field or method SetGoogleClientInfo)


codecrane:
**Code:**

https://github.com/googleapis/google-cloud-go/blame/77c1cfac704dab6704e9b55bdd8a710695dc936a/vertexai/genai/client.go#L56

```
func NewClient(ctx context.Context, projectID, location string, opts ...option.ClientOption) (*Client, error) {
	opts = append([]option.ClientOption{
		option.WithEndpoint(fmt.Sprintf("%s-aiplatform.googleapis.com:443", location)),
	}, opts...)
	c, err := aiplatform.NewPredictionClient(ctx, opts...)
	if err != nil {
		return nil, err
	}
	c.SetGoogleClientInfo("gccl", internal.Version)
	return &Client{
		c:         c,
		projectID: projectID,
		location:  location,
	}, nil
}
```

**Implication:**
Application does not compile as SetGoogleClientInfo() is not available in the pinned version of the Prediction client. It looks like NewClient() is attempting to call the prediction client internal `c.setGoogleClientInfo()`
https://github.com/googleapis/google-cloud-go/blob/main/aiplatform/apiv1beta1/prediction_client.go#L412 

```
# cloud.google.com/go/vertexai/genai
../../go/pkg/mod/cloud.google.com/go/vertexai@v0.5.1/genai/client.go:56:4: c.SetGoogleClientInfo undefined (type *aiplatform.PredictionClient has no field or method SetGoogleClientInfo)
```

**Solution:**
Remove the call to `c.SetGoogleClientInfo("gccl", internal.Version)` from `func NewClient(ctx context.Context, projectID, location string, opts ...option.ClientOption) (*Client, error) {`

codyoss:
Will get a patch out but as a workaround you can pull cloud.google.com/go/aiplatform@v1.58.0, sorry about that

codyoss:
Thank you for the report! Released in 0.5.2

PR #9168, opened 2023-12-14
fix(vertexai): bump deps to fix build error

codyoss:
Tested locally to make sure it works.

Fixes: #9167


codyoss:
As a side note to @eliben @jba I don't understand why go build did not catch this... Does it have to do with the module workspace or am I missing something?

noahdietz (approved):

eliben:
> As a side note to @eliben @jba I don't understand why go build did not catch this... Does it have to do with the module workspace or am I missing something?

The root repo has a `go.work` file, which will mask such issues. This is the reason why having a `go.work` file committed is considered dangerous; when it's committed, it's worth at least running the tests with `GOWORK=off` to catch such issues.

You can try to run all tests in this repo with `GOWORK=off` in CI - this may uncover surprises :)

codyoss:
> This is the reason why having a go.work file committed is considered dangerous

I never got this impression when reading docs. Is this written anywhere?

> You can try to run all tests in this repo with GOWORK=off

Will file an issue, thanks

eliben:
> > This is the reason why having a go.work file committed is considered dangerous
> 
> I never got this impression when reading docs. Is this written anywhere?
>

I'll reply in #9170

PR #9195, opened 2023-12-21
chore(github): set minimal workflow permissions

### LINKED ISSUE #9194: CI/CD: Adopt minimal token permissions
Almost all of this project's workflows run with adequately-scoped workflow tokens, which is great.

However, a new workflow (`conformance.yml`) forgot to declare specific permissions, which means it's running with write-all permissions. Also, some of the workflows declare top-level write permissions even though only one of the jobs needs those permissions.

All workflows should declare read-only top-level permissions, with additional write permissions declared at the individual job level.

I'll send a PR setting these tokens.

The project should also modify the default workflow token to run as read-only, not write-all:

1. Open the repo Settings > Actions > [General](https://github.com/googleapis/google-cloud-go/settings/actions)
2. Under "Workflow permissions", set them to "Read repository contents and packages permissions"

This way, should a workflow forget to declare top-level permissions again in the future, the workflow won't be a security hazard.

**Additional context**
My name is Pedro and I work with Google and the [OpenSSF](https://openssf.org/) to improve the supply-chain security of open-source projects that are critical to the ecosystem.

pnacht:
Fixes #9194.

This PR ensures all workflows run with minimally-scoped tokens, preventing them from being abused in a supply-chain attack against the project.

Some observations:

- `action_syntax.yml`: This change is just future-proofing, with no effective change to permissions at this moment. This simply protects the project in case another job is ever added to the workflow.
- `apidiff.yml`: Now only the second job has additional permissions, since the first job doesn't need them.
- `conformance.yml`: Currently has no top-level permissions and therefore runs as write-all. The workflow seems to require read-only permissions.
- `new_client.yml`: I could not determine where the `pull-requests: write` permission is used, since the `code_suggester` uses a separate `YOSHI_CODE_BOT_TOKEN`, not the workflow's default token. I've therefore removed that permission from the workflow. Let me know if I missed anything.


quartzmo (approved):
Thanks for this contribution, @pnacht !

PR #9200, opened 2023-12-27
Add query profiling

bhshkh:
Introducing query modes:

- NORMAL: Plans and executes the query. Only returns the query results.
- EXPLAIN: Plans but does not execute the query. Only returns planner stage information. Can be used to check that a query has the necessary indexes and understand which indexes are used.
- EXPLAIN ANALYZE: Plans and executes the query. Returns query results, planner stage information and execution statistics.

### DIFF
--- a/datastore/integration_test.go
+++ b/datastore/integration_test.go
@@ -1543,7 +1543,6 @@ func isEqualResultSetStats(got *ResultSetStats, want *ResultSetStats) error {
 			return fmt.Errorf("Stats.QueryPlan.PlanInfo: got: %+v, want: %+v", got.QueryPlan, want.QueryPlan)
 		}
 
-		// Compare query stats maps except 'total_execution_time' key
 		if !testutil.Equal(got.QueryStats, want.QueryStats, cmp.FilterPath(ignoreStatsFields, cmp.Ignore())) {
 			return fmt.Errorf("Stats.QueryPlan.QueryStats: got: %+v, want: %+v", got.QueryStats, want.QueryStats)
 		}
--- a/datastore/integration_test.go
+++ b/datastore/integration_test.go
@@ -24,6 +24,7 @@ import (
 	"net/url"
 	"os"
 	"reflect"
+	"slices"
 	"sort"
 	"strings"
 	"testing"
@@ -32,6 +33,7 @@ import (
 	"cloud.google.com/go/internal/testutil"
 	"cloud.google.com/go/internal/uid"
 	"cloud.google.com/go/rpcreplay"
+	"github.com/google/go-cmp/cmp"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
 	pb "google.golang.org/genproto/googleapis/datastore/v1"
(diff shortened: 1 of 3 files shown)


datastore/query.go
```diff
@@ -625,6 +626,70 @@ func (c *Client) Count(ctx context.Context, q *Query) (n int, err error) {
 	}
 }
 
+const (
+	// QueryModeNormal is the default mode. Only the query results are returned.
+	QueryModeNormal QueryMode = iota
+	// QueryModeExplain returns only the query plan, without any results or execution
```
kolea2, line 632: apologies if I'm missing - how do these map to NORMAL, PLAN, AND PROFILE?

bhshkh: QueryModeExplain => PLAN
QueryModeExplainAnalyze => PROFILE
This line converts QueryMode to proto: https://github.com/googleapis/google-cloud-go/pull/9200/files#diff-c04af33c4c45ff4606abe36bd859432effb278cca4146d3f818e540f0479fd8dR652

kolea2: please add a comment for each value with the appropriate numeric value

datastore/integration_test.go
```diff
@@ -1291,6 +1389,158 @@ func TestIntegration_GetAllWithFieldMismatch(t *testing.T) {
...
+	}
+	return keys, entities, now, parent, func() {
+		err := client.DeleteMulti(ctx, keys)
+		if err != nil {
+			t.Errorf("client.DeleteMulti: %v", err)
+		}
+	}
+}
+
+func TestIntegration_RunAndGetAllWithOptions(t *testing.T) {
```
danieljbruce, line 1419: nit: This seems fine, but one alternative would be to have a separate test for run from GetAllWithOptions.

danieljbruce (commented):
LGTM. I think @kolea2 mentioned that samples will be required as well so samples could be added in this PR or in a separate PR.

datastore/query.go
bhshkh: Added

datastore/integration_test.go
bhshkh: Updated

bhshkh:
Samples at https://github.com/GoogleCloudPlatform/golang-samples/pull/3742

datastore/query.go
danieljbruce: It might also be useful to mention the enum value this maps to ie. PLAN, PROFILE.

danieljbruce (approved):

datastore/query.go
bhshkh: The protos have been changed. no more enums

danieljbruce (commented):
LGTM. I can do another review once the owl-bot PR comes in so that all the CI checks pass.

datastore/integration_test.go
```diff
@@ -1432,62 +1425,51 @@ func createTestEntities(ctx context.Context, t *testing.T, client *Client, parti
...
-						"indexes_used": []interface{}{
-							map[string]interface{}{
-								"properties":  "(T ASC, I ASC, __name__ ASC)",
-								"query_scope": "Includes Ancestors",
-							},
-						},
-					},
-				},
-				QueryStats: map[string]interface{}{
+			wantExecutionStats: &ExecutionStats{
```
danieljbruce, line 1460: Maybe this is just a bi-product of the test environment, but do you know why this result doesn't contain bytesReturned, executionDuration and readOptions?

bhshkh: New protos do not have bytesReturned.
For rest:
The environment does return those but they are unpredictable. So, they have not been added in expected results and are being ignored while comparing expected and actual.

datastore/query.go
```diff
@@ -651,6 +751,15 @@ func (c *Client) GetAll(ctx context.Context, q *Query, dst interface{}) (keys []
 	ctx = trace.StartSpan(ctx, "cloud.google.com/go/datastore.Query.GetAll")
 	defer func() { trace.EndSpan(ctx, err) }()
 
+	res, err := c.GetAllWithOptions(ctx, q, dst)
+	return res.Keys, err
+}
+
+// GetAllWithOptions is similar to GetAll but runs the query with provided options
+func (c *Client) GetAllWithOptions(ctx context.Context, q *Query, dst interface{}, opts ...RunOption) (res GetAllWithOptionsResult, err error) {
```
kolea2, line 759: is it possible for opts to be empty here?

datastore/query.go
```diff
@@ -702,21 +812,21 @@ func (c *Client) GetAll(ctx context.Context, q *Query, dst interface{}) (keys []
...
+		res.Keys = append(res.Keys, k)
 	}
-	return keys, errFieldMismatch
+	return res, errFieldMismatch
 }
 
-// Run runs the given query in the given context.
-func (c *Client) Run(ctx context.Context, q *Query) *Iterator {
+// Run runs the given query in the given context with the provided options
+func (c *Client) Run(ctx context.Context, q *Query, opts ...RunOption) *Iterator {
```
kolea2, line 829: kind of related - here we are just adding `...RunOptions` as a new parameter rather than creating a new method `RunWithOptions`. Why do we need to create a new method `GetAllWithOptions` if we could just add them to `Get`?

datastore/query.go
bhshkh: Yes. A possible call to GetAllWithOptions is:
```
var got []SQChild
gotRes, gotErr := client.GetAllWithOptions(ctx, query, &got)
```

datastore/query.go
bhshkh: Run method returns a struct pointer called Iterator. So, I'm able to add explainmetrics as a field to the existing struct.
A struct is somewhat similar to a class in Java. Here, pointer to object is being returned. New fields can be introduced to existing object.


In case of GetAll, this is the method signature:
```
func (c *Client) GetAll(ctx context.Context, q *Query, dst interface{}) (keys []*Key, err error)
```
This cannot be changed to 
```
func (c *Client) GetAll(ctx context.Context, q *Query, dst interface{}) (keys []*Key, err error, em *ExplainMetrics)
```
Because this will break client code:
```
dst := []*SQChild{}
_ , err := client.GetAll(ctx, query, &dst)
```
With error:  `assignment mismatch: 2 variables but client.GetAll returns 3 values`

kolea2: got it, thank you for the detailed explanation :)

bhshkh:

datastore/query.go
```diff
@@ -627,6 +629,107 @@ func (c *Client) Count(ctx context.Context, q *Query) (n int, err error) {
...
+//
+// Query Explain feature is still in preview and not yet publicly available.
+// Pre-GA features might have limited support and can change at any time.
+type ExplainOptions struct {
+	// When false (the default), the query will be planned, returning only
+	// metrics from the planning stages.
+	// When true, the query will be planned and executed, returning the full
+	// query results along with both planning and execution stage metrics.
+	Analyze bool
+}
```
gkevinzheng, line 647: Why is this not structured as a private struct `explainOptions` with a function `ExplainOptions` that returns an instance of `explainOptions`? With this structure the user would have to pass in stuff like `ExplainOptions{Analyze: true}` instead of `ExplainOptions(true)`, right?

datastore/query.go
```diff
@@ -627,6 +629,107 @@ func (c *Client) Count(ctx context.Context, q *Query) (n int, err error) {
...
+
+func (e ExplainOptions) apply(s *runQuerySettings) error {
+	if s.explainOptions != nil {
+		return errors.New("datastore: ExplainOptions can be specified can be specified only once")
```
gkevinzheng, line 651: nit: Duplicate `can be specified`

datastore/query.go
bhshkh: Right now, ExplainOptions contains only Analyze field. In the future, if more fields are added to ExplainOptions struct, user will no longer be able to use ExplainOptions(true).

datastore/query.go
bhshkh: Removed

datastore/query.go
gkevinzheng: Ahh I see, I was confused because this is a proto field, not an option in the Golang sense of the word.

gkevinzheng (approved):

PR #9205, opened 2023-12-29
feat(storage): Max-retry-count to stop retries

Tulsishah:
To prevent continuous retries in case of errors, the MaxRetryCount variable serves as a checkpoint. It specifies the maximum allowable attempts before terminating the retry process.


tritone:
superceded by #9215

Issue #9111, opened 2023-12-09
spanner: `(*Row).ToStruct` is 3x slower than `(*Row).Columns`


CAFxX:
In local benchmarks `(*Row).ToStruct` is >3x slower than `(*Row).Columns`. 

```
goos: darwin
goarch: amd64
pkg: github.com/kouzoh/go-microservices-kit/database/spanner
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
        │ baseline-columns │           baseline-tostruct           │
        │      sec/op      │    sec/op     vs base                 │
1              211.1n ± 1%   657.9n ±  7%  +211.70% (p=0.000 n=10)
10             2.130µ ± 1%   6.107µ ±  4%  +186.71% (p=0.000 n=10)
100            15.74µ ± 1%   54.36µ ±  2%  +245.33% (p=0.000 n=10)
1000           157.7µ ± 1%   538.0µ ±  2%  +241.14% (p=0.000 n=10)
10000          1.753m ± 8%   5.591m ± 11%  +218.86% (p=0.000 n=10)
geomean        18.13µ        58.01µ        +220.03%

        │ baseline-columns │           baseline-tostruct            │
        │       B/op       │     B/op       vs base                 │
1               48.00 ± 0%     128.00 ± 0%  +166.67% (p=0.000 n=10)
10              984.0 ± 0%     1568.0 ± 0%   +59.35% (p=0.000 n=10)
100           8.320Ki ± 0%   13.812Ki ± 0%   +66.01% (p=0.000 n=10)
1000          99.41Ki ± 0%   154.12Ki ± 0%   +55.03% (p=0.000 n=10)
10000         1.287Mi ± 0%    1.822Mi ± 0%   +41.48% (p=0.000 n=10)
geomean       8.675Ki         15.00Ki        +72.95%

        │ baseline-columns │          baseline-tostruct           │
        │    allocs/op     │  allocs/op   vs base                 │
1               2.000 ± 0%    5.000 ± 0%  +150.00% (p=0.000 n=10)
10              15.00 ± 0%    36.00 ± 0%  +140.00% (p=0.000 n=10)
100             108.0 ± 0%    309.0 ± 0%  +186.11% (p=0.000 n=10)
1000           1.012k ± 0%   3.013k ± 0%  +197.73% (p=0.000 n=10)
10000          10.02k ± 0%   30.02k ± 0%  +199.63% (p=0.000 n=10)
[...]

PR #9206, opened 2023-12-29
feat(spanner): add SelectAll method to decode from Spanner iterator.Rows to golang struct

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/9111

####  Steps to benchmark
```
- go test -bench=^BenchmarkScan100RowsUsingSelectAll$ -run==^BenchmarkScan100RowsUsingSelectAll$ -benchmem -count 10 > 10_runs_bench_using_select_all.txt && sed -i '' 's/UsingSelectAll//g' 10_runs_bench_using_select_all.txt
- go test -bench=^BenchmarkScan100RowsUsingToStruct$ -run==^BenchmarkScan100RowsUsingToStruct$ -benchmem -count 10 > 10_runs_bench_using_tostruct.txt && sed -i '' 's/UsingToStruct//g' 10_runs_bench_using_tostruct.txt
- go test -bench=^BenchmarkScan100RowsUsingColumns$ -run==^BenchmarkScan100RowsUsingColumns$ -benchmem -count 10 > 10_runs_bench_using_columns.txt && sed -i '' 's/UsingColumns//g' 10_runs_bench_using_columns.txt
```

#### Compare results
```
- go install golang.org/x/perf/cmd/benchstat@latest
- benchstat 10_runs_bench_using_columns.txt 10_runs_bench_using_select_all.txt
```

```
goos: darwin
goarch: arm64
               │ 10_runs_bench_using_columns.txt │  10_runs_bench_using_select_all.txt   │
               │             sec/op              │    sec/op     vs base                 │
Scan100Rows-10                       7.076µ ± 2%     14.434µ ± 0%  +102.92% (p=0.000 n=20)

               │ 10_runs_bench_using_columns.txt │  10_runs_bench_using_select_all.txt   │
               │              B/op               │     B/op       vs base                │
Scan100Rows-10                      8.320Ki ± 0%   10.945Ki ± 0%  +31.55% (p=0.000 n=20)

               │ 10_runs_bench_using_columns.txt │ 10_runs_bench_using_select_all.txt  │
               │            allocs/op            │ allocs/op   vs base                 │
Scan100Rows-10                        108.0 ± 0%    219.0 ± 0%  +102.78% (p=0.000 n=20)

```

Result of running the combined benchmark `BenchmarkScan` on a five field Go struct, here is the chart showing execution time vs number of rows scanned.
Screenshot 2024-01-15 at 20 20 21


rahul2393:
@CAFxX Can you please help review this please, while this is not the final optimization but its better than using toStruct
Thanks

spanner/row.go
```diff
@@ -378,3 +386,123 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+			}
+		}
+		if len(pointers) == 0 {
+			return nil
+		}
+		err := row.Columns(pointers...)
+		if err != nil {
+			return err
+		}
+		if len(pointers) > 0 {
```
CAFxX, line 437: isn't this always true, given the check on line 430?

spanner/row.go
```diff
@@ -378,3 +386,123 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+	sliceVal := reflect.Indirect(reflect.ValueOf(v))
+	itemType := sliceType.Elem()
+	s := &decodeSetting{}
+	for _, opt := range options {
+		opt.Apply(s)
+	}
+
+	isPrimitive := itemType.Kind() != reflect.Struct
+	var pointers []interface{}
+	var err error
```
CAFxX, line 415: probably unneeded?

spanner/row.go
```diff
@@ -378,3 +386,123 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+	if err := rows.Do(func(row *Row) error {
+		sliceItem := reflect.New(itemType).Elem()
+		if len(pointers) == 0 {
```
CAFxX, line 418: maybe it's clearer to use a dedicated boolean flag to track whether this is the first row or not?

spanner/row.go
```diff
@@ -378,3 +386,123 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+		if err != nil {
+			return err
+		}
+		if len(pointers) > 0 {
+			dst := sliceItem.Addr().Interface()
+			for i, p := range pointers {
+				reflect.ValueOf(dst).Elem().Field(i).Set(reflect.ValueOf(p).Elem())
+			}
+		}
+		sliceVal.Set(reflect.Append(sliceVal, sliceItem))
```
CAFxX, line 443: would be great if we could preallocate the whole resultset, but IIRC this is not possible today because the spanner client does not know how many rows are there in the resultset until the end (not sure if this is a fundamental issue, or it can be improved)

spanner/row.go
```diff
@@ -378,3 +386,123 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+	itemType := sliceType.Elem()
+	s := &decodeSetting{}
+	for _, opt := range options {
+		opt.Apply(s)
+	}
+
+	isPrimitive := itemType.Kind() != reflect.Struct
+	var pointers []interface{}
+	var err error
+	if err := rows.Do(func(row *Row) error {
```
CAFxX, line 416: [nit] `return rows.Do(...`

spanner/row.go
```diff
@@ -378,3 +386,123 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+				return nil, errNoOrDupGoField(sliceItem, colName.GetName())
+			}
+			fieldVal = sliceItem.FieldByName(colName.GetName())
+		}
+		if !fieldVal.IsValid() || !fieldVal.CanSet() {
+			// have to add if we found a column because Scan() requires
+			// len(cols) arguments or it will error. This way we can scan to
+			// a useless pointer
+			var nothing interface{}
+			pointers = append(pointers, &nothing)
```
CAFxX, line 471: I think you can do `pointers = append(pointers, nil)` here (see https://github.com/googleapis/google-cloud-go/blob/0020ab23b0d1344cf03b2bfa5e2afe947b70240a/spanner/row.go#L282-L284)

spanner/row.go
```diff
@@ -378,3 +386,123 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+			return nil
+		}
+		err := row.Columns(pointers...)
+		if err != nil {
+			return err
+		}
+		if len(pointers) > 0 {
+			dst := sliceItem.Addr().Interface()
+			for i, p := range pointers {
+				reflect.ValueOf(dst).Elem().Field(i).Set(reflect.ValueOf(p).Elem())
```
CAFxX, line 440: nit: probably worth hoisting `reflect.ValueOf(dst).Elem()` out of the loop

CAFxX:
> this is not the final optimization

@rahul2393 to be able to know what to focus on in the review, can you give me a rough idea of what kind of optimizations you are imagining to do?

rahul2393:
> > this is not the final optimization
> 
> @rahul2393 to be able to know what to focus on in the review, can you give me a rough idea of what kind of optimizations you have in your mind?

Thanks [CAFxX](https://github.com/CAFxX) for review, I am trying to reduce the allocation ops as much as possible by preallocating as you said, since Spanner Row does not have meta information on number of rows we need to see how we can get maximum performance given the situation.

rahul2393:
[CAFxX](https://github.com/CAFxX)
I am done with the final changes, please help check again
`QueryWithStats` returns the RowCount when iterating the rows, in that case SelectAll will do preallocation for others we have to dynamically grow size.
Below is the benchmark with QueryWithStats
```
goos: darwin
goarch: arm64
               │ 10_runs_bench_using_columns.txt │  10_runs_bench_using_select_all.txt   │
               │             sec/op              │    sec/op     vs base                 │
Scan100Rows-10                       7.076µ ± 2%    10.269µ ± 0%  +44.37% (p=0.000 n=20)

               │ 10_runs_bench_using_columns.txt │  10_runs_bench_using_select_all.txt   │
               │              B/op               │     B/op       vs base                │
Scan100Rows-10                      8.320Ki ± 0%   5.273Ki ± 0%  -36.62% (p=0.000 n=20)

               │ 10_runs_bench_using_columns.txt │ 10_runs_bench_using_select_all.txt  │
               │            allocs/op            │ allocs/op   vs base                 │
Scan100Rows-10                        108.0 ± 0%    113.0 ± 0%  +4.63% (p=0.000 n=20)
```

For others
```
goos: darwin
goarch: arm64
               │ 10_runs_bench_using_columns.txt │  10_runs_bench_using_select_all.txt   │
               │             sec/op              │    sec/op     vs base                 │
Scan100Rows-10                       7.076µ ± 2%     14.434µ ± 0%  +102.92% (p=0.000 n=20)

               │ 10_runs_bench_using_columns.txt │  10_runs_bench_using_select_all.txt   │
               │              B/op               │     B/op       vs base                │
Scan100Rows-10                      8.320Ki ± 0%   10.945Ki ± 0%  +31.55% (p=0.000 n=20)

               │ 10_runs_bench_using_columns.txt │ 10_runs_bench_using_select_all.txt  │
               │            allocs/op            │ allocs/op   vs base                 │
Scan100Rows-10                        108.0 ± 0%    219.0 ± 0%  +102.78% (p=0.000 n=20)
```

spanner/row.go
```diff
@@ -378,3 +386,133 @@ func (r *Row) ToStructLenient(p interface{}) error {
 		true,
 	)
 }
+
+// SelectAll scans rows into a slice (v)
```
CAFxX, line 390: We should document the accepted types of v (or enforce it with generics)

spanner/row.go
```diff
@@ -378,3 +386,133 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+func SelectAll(rows Iterator, v interface{}, options ...DecodeOptions) error {
+	if rows == nil {
+		return fmt.Errorf("rows is nil")
+	}
+	if v == nil {
+		return fmt.Errorf("p is nil")
+	}
+	vType := reflect.TypeOf(v)
+	if k := vType.Kind(); k != reflect.Ptr {
+		return errToStructArgType(v)
```
CAFxX, line 400: is this the right error message here? shouldn't it be errNotASlicePointer?

spanner/row.go
```diff
@@ -378,3 +386,133 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+	for _, opt := range options {
+		opt.Apply(s)
+	}
+
+	isPrimitive := itemType.Kind() != reflect.Struct
+	var pointers []interface{}
+	isFistRow := true
+	rowIndex := -1
+	return rows.Do(func(row *Row) error {
+		sliceItem := reflect.New(itemType).Elem()
```
CAFxX, line 418: if I'm following correctly, this implies the type of `v` must be something like `*[]*struct {...}`.

Would it be too hard to also support `*[]struct { ... }`? The benefit for users that pass a `*[]struct { ... }` would be less calls to the allocator, lower allocated memory, and better memory locality (as all rows are guaranteed to be contiguous in memory).

CAFxX (commented):
Second round

spanner/row.go
CAFxX: Sorry, my bad, misread the code

spanner/read.go
```diff
@@ -121,6 +130,27 @@ type RowIterator struct {
...
+			return r.QueryStats["rows_returned"].(int64)
+		case string:
+			v, err := strconv.Atoi(r.QueryStats["rows_returned"].(string))
+			if err != nil {
+				return -1
+			}
+			return int64(v)
+		default:
+			return -1
+		}
```
CAFxX, line 149:
```suggestion
		switch rowsReturned := r.QueryStats["rows_returned"].(type) {
		case float64:
			return int64(rowsReturned)
		case string:
			v, err := strconv.ParseInt(rowsReturned, 10, 64)
			if err != nil {
				v = -1
			}
			return v
		}
```

spanner/read.go
```diff
@@ -121,6 +130,27 @@ type RowIterator struct {
 	sawStats         bool
 }
 
+// RowsReturned returns the number of rows returned by the query. If the query
+// was a DML statement, the number of rows affected is returned. If the query
+// was a PDML statement, the number of rows affected is a lower bound.
```
CAFxX, line 135:
```suggestion
// RowsReturned returns, if the query was executed with query stats enabled,
// the number of rows returned by the query. 
//
// If the query was a DML statement, the number of rows affected is returned. 
// If the query was a PDML statement, the number of rows affected is a lower bound.
// If the query was executed without query stats enabled, or if it is otherwise
// impossible to determine the number of rows in the resultset, -1 is returned.
```

spanner/row.go
```diff
@@ -378,3 +386,133 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+		if isFistRow {
+			nRows := rows.RowsReturned()
+			if nRows != -1 {
+				sliceVal = reflect.MakeSlice(sliceType, int(nRows), int(nRows))
```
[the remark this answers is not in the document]
rahul2393: Have updated in comment that we will `resets the destination slice`

rahul2393:
@CAFxX please help take a look again.

spanner/row.go
```diff
@@ -378,3 +386,133 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+			return err
+		}
+		for i, p := range pointers {
+			if p == nil {
+				continue
+			}
+			sliceItem.Field(i).Set(reflect.ValueOf(p).Elem())
+		}
+		if rowIndex >= 0 {
+			sliceVal.Index(rowIndex).Set(sliceItem)
```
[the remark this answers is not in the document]
rahul2393: Yes, using `reflect.MakeSlice(sliceType, 0, int(nRows))` was 25% slower compared to preallocated and setting the value via index, I have updated the code to not panic if `RowsReturned ` is lower bound, 
btw it will be lower bound only for DML and PDML so normal Select queries should work needing append.

spanner/row.go
```diff
@@ -249,6 +249,17 @@ func errColNotFound(n string) error {
 	return spannerErrorf(codes.NotFound, "column %q not found", n)
 }
 
+func errNotASlicePointer() error {
+	return spannerErrorf(codes.InvalidArgument, "destination must be a pointer to a slice")
+}
+func errNilSlicePointer() error {
```
olavloite, line 255: nit: add empty line

spanner/read.go
```diff
@@ -121,6 +130,29 @@ type RowIterator struct {
...
+// Currently, this requires the query to be executed with query stats enabled.
+//
+// If the query was a DML statement, the number of rows affected is returned.
+// If the query was a PDML statement, the number of rows affected is a lower bound.
+// If the query was executed without query stats enabled, or if it is otherwise
+// impossible to determine the number of rows in the resultset, -1 is returned.
+func (r *RowIterator) RowsReturned() int64 {
+	if r.sawStats && r.QueryStats != nil && r.QueryStats["rows_returned"] != nil {
+		switch rowsReturned := r.QueryStats["rows_returned"].(type) {
+		case float64:
```
olavloite, line 143: When is this `float64`?

spanner/read.go
```diff
@@ -121,6 +130,29 @@ type RowIterator struct {
 	sawStats         bool
 }
 
+// RowsReturned returns, a lower bound on the number of rows returned by the query.
+// Currently, this requires the query to be executed with query stats enabled.
```
olavloite, line 134: We should also document that the query stats are only included in the last response of the gRPC stream. That means that if you run a query with query stats enabled that returns a large number of rows, then this method is only guaranteed to return the number of rows once you have iterated through all the rows (which again makes it a bit less useful). So maybe add the following to this sentence: `The query stats are only guaranteed to be available after iterating through all the rows.`

And change the last point to something like this:

```
// If the query was executed without query stats enabled, if the query stats have not yet been
// returned by the server (the query stats are included in the last message of the query stream),
// or if it is otherwise impossible to determine the number of rows in the resultset, -1 is returned.
```

spanner/read.go
```diff
@@ -90,6 +91,14 @@ func streamWithReplaceSessionFunc(
 	}
 }
 
+// Iterator is an interface for iterating over Rows.
+type Iterator interface {
```
olavloite, line 95: Do we need this interface? (Or put another way: What are the benefits of adding this interface?)

The name also seems very generic.

spanner/row.go
```diff
@@ -378,3 +389,186 @@ func (r *Row) ToStructLenient(p interface{}) error {
 		true,
 	)
 }
+
+// SelectAll iterates all rows to the end. After iterating it closes the rows,
+// and propagates any errors that could pop up.
+// It expects that destination should be a slice. For each row it scans data and appends it to the destination slice.
+// SelectAll supports both types of slices: slice of structs by a pointer and slice of structs by value,
```
olavloite, line 396: nit: 
```suggestion
// SelectAll supports both types of slices: slice of structs by pointer and slice of structs by value,
```

spanner/row.go
```diff
@@ -378,3 +389,186 @@ func (r *Row) ToStructLenient(p interface{}) error {
 		true,
 	)
 }
+
+// SelectAll iterates all rows to the end. After iterating it closes the rows,
```
olavloite, line 393:
```suggestion
// SelectAll iterates all rows to the end. After iterating it closes the rows
```

spanner/row.go
```diff
@@ -378,3 +389,186 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+// and propagates any errors that could pop up.
```
olavloite, line 394: This is a bit ambiguous: Is the destination slice half-filled if an error 'pops up'? Or is it then always empty? Could it happen that you get a filled destination slice and an error, and that the destination slice contains all data?

spanner/row.go
```diff
@@ -378,3 +389,186 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+// It expects that destination should be a slice. For each row it scans data and appends it to the destination slice.
```
olavloite, line 395:
```suggestion
// It expects that destination should be a slice. For each row, it scans data and appends it to the destination slice.
```

spanner/row.go
```diff
@@ -378,3 +389,186 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+//	}
+//
+//	var singersByPtr []*Singer
+//	var singersByValue []Singer
+//
+// Both singersByPtr and singersByValue are valid destinations for SelectAll function.
+//
+// Before starting, SelectAll resets the destination slice,
+// so if it's not empty it will overwrite all existing elements.
+func SelectAll(rows Iterator, v interface{}, options ...DecodeOptions) error {
```
olavloite, line 411:
```suggestion
func SelectAll(rows Iterator, destination interface{}, options ...DecodeOptions) error {
```

(Or clarify in the documentation above that `v` equals `destination`.)

spanner/row.go
```diff
@@ -378,3 +389,186 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+//	    Name  string
+//	}
+//
+//	var singersByPtr []*Singer
+//	var singersByValue []Singer
+//
+// Both singersByPtr and singersByValue are valid destinations for SelectAll function.
+//
+// Before starting, SelectAll resets the destination slice,
+// so if it's not empty it will overwrite all existing elements.
```
olavloite, line 410: Can we add a line or two about what the `DecodeOptions` are used for?

spanner/row.go
```diff
@@ -378,3 +389,186 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+func SelectAll(rows Iterator, v interface{}, options ...DecodeOptions) error {
+	if rows == nil {
+		return fmt.Errorf("rows is nil")
+	}
+	if v == nil {
+		return fmt.Errorf("p is nil")
```
olavloite, line 416:
```suggestion
		return fmt.Errorf("v is nil")
```

(Or `destination is nil`)

spanner/row.go
```diff
@@ -378,3 +389,186 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+		return errNotASlicePointer()
+	}
+
+	itemType := dstType.Elem()
+	var itemByPtr bool
+	// If it's a slice of pointers to structs,
+	// we handle it the same way as it would be slice of struct by value
+	// and dereference pointers to values,
+	// because eventually we work with fields.
+	// But if it's a slice of primitive type e.g. or []string or []*string,
```
olavloite, line 437: I don't quite understand this. The documentation of the method says that the destination slice must be a slice of (pointers to) structs. This seems to indicate that we also support primitives.

spanner/value.go
```diff
@@ -3209,6 +3209,11 @@ func (w withLenient) Apply(s *decodeSetting) {
 	s.Lenient = w.lenient
 }
 
+// WithLenient returns a DecodeOptions that allows decoding into a struct with missing fields in database.
+func WithLenient() DecodeOptions {
```
olavloite, line 3213: Why was this not already exported? Was this something that users could not use? Or did they have some other API that they could use to access this option?

spanner/value_benchmarks_test.go
```diff
@@ -230,3 +232,167 @@ func decodeArrayReflect(pb *proto3.ListValue, name string, typ *sppb.Type, aptr
 	}
 	return nil
 }
+
+func BenchmarkScan100RowsUsingSelectAll(b *testing.B) {
```
olavloite, line 236: It seems that we only have benchmarks for this function, but no unit- and integration tests. Can we add that? Maybe re-use the tests that currently use the method that was deemed to slow to verify that the output is the same?

spanner/read.go
rahul2393: >We should also document that the query stats are only included in the last response of the gRPC stream. That means that if you run a query with query stats enabled that returns a large number of rows, then this method is only guaranteed to return the number of rows once you have iterated through all the rows (which again makes it a bit less useful). So maybe add the following to this sentence: The query stats are only guaranteed to be available after iterating through all the rows.

This is only true for DML/PDML queries, for normal execute SQL queries I don't see that restriction, and usecases of decoding from DML/PDML returned rows is usually with Returning clause.

spanner/row.go
CAFxX, line 410: Sorry, as mentioned before, I think we should only append to the slice (without first resetting it) in case a non-nil slice is provided. WDYT? The user may have pre-seeded the slice with some entries, or it may want to use the same slice in multiple successive calls. If we always reset, we basically preclude these options. Whereas with the alternative (just append) all these options are available (and the current behavior of resetting is trivial for the user: just pass `s[:0]` instead of just `s`)

spanner/value_benchmarks_test.go
rahul2393: Added unit tests

spanner/value.go
rahul2393: Earlier there were 2 public methods ToStruct and ToStructLenient, we don't want to carry the approach of having 2 separate methods for SelectAll hence added as option.

spanner/row.go
rahul2393: Yes we do support primitives, tests added.

spanner/read.go
rahul2393: Updated the interface to package private so that can't be used by customers, added this to ease mock and unit tests

spanner/read.go
rahul2393: Removed the logic to rely on ReturnedRows stats since for large dataset row count is returned with last result set only hence it won't help in preallocation.
Thanks

spanner/row.go
rahul2393: Yes, error will be returned with destination partially filled, added a test case for the same.

spanner/row.go
rahul2393: Removed reset logic, added unit test to validate it will append on base slice.

olavloite (approved):

spanner/row.go
```diff
@@ -378,3 +390,174 @@ func (r *Row) ToStructLenient(p interface{}) error {
...
+//	var singersByValue []Singer
+//
+// Both singersByPtr and singersByValue are valid destinations for SelectAll function.
+//
+// custom setting such as lenient can be passed as an option using DecodeOptions
+// example: to ignore extra columns in the row
+//
+//	var singersByPtr []*Singer
+//	err := spanner.SelectAll(row, &singersByPtr, spanner.WithLenient())
+func SelectAll(rows rowIterator, destination interface{}, options ...DecodeOptions) error {
```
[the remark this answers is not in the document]
rahul2393: Yes we can call this method from outside spanner package.
1) Yes, its intentional at the moment so that no one can use it outside the package, the only reason its present there is to ease in test setup.
2) func SelectAll is exported: Yes and it can be used with RowIterator that is returned for query results 
3) Same as above.
4) Yes we can

PR #9208, opened 2024-01-03
feat(storage): support auto-detection of access ID for external_account creds

### LINKED ISSUE #8528: storage: support external_account credentials in detectDefaultGoogleAccessID

[detectDefaultGoogleAccessID](https://github.com/googleapis/google-cloud-go/blob/a05b6edb5795104a08a41a45be04dc3d2535eff2/storage/bucket.go#L263C33-L263C33) can detect GoogleAccessID when the library is authenticated by GKE workload identity but it cannot when it is authenticated by workload identity federation (e.g. GitHub Actions) because the credential type is `external_account`.

We'd like the function [detectDefaultGoogleAccessID](https://github.com/googleapis/google-cloud-go/blob/a05b6edb5795104a08a41a45be04dc3d2535eff2/storage/bucket.go#L263C33-L263C33) to support `external_account` credentials. If the `service_account_impersonation_url` value in the credential is available, the function should be able to extract an Email from it just like the case of `impersonated_service_account` credentials.

We can specify GoogleAccessID explicitly in a user code but we don't want to branch out logics depending on how the library is authenticated.

BrennaEpp:
Fixes #8528


storage/bucket.go
```diff
@@ -275,18 +275,24 @@ func (b *BucketHandle) detectDefaultGoogleAccessID() (string, error) {
...
-		} else if sa.CredType == "service_account" && sa.ClientEmail != "" {
-			return sa.ClientEmail, nil
 		} else {
-			returnErr = errors.New("unable to parse credentials; only service_account and impersonated_service_account credentials are supported")
+			switch sa.CredType {
+			case "impersonated_service_account", "external_account":
+				start, end := strings.LastIndex(sa.SAImpersonationURL, "/"), strings.LastIndex(sa.SAImpersonationURL, ":")
+
+				if end <= start {
+					returnErr = errors.New("error parsing impersonated service account credentials")
```
tritone, line 284: Error message could mention external account as well?

tritone (approved):
Nice change, one nit on error messages.

storage/bucket.go
BrennaEpp: The error here is parsing the service account, which the external account creds also has. I think it is okay to expand the message though, since it's just a difference in granularity and mentioning external accounts is at a higher level

PR #9211, opened 2024-01-03
feat(storage): support IncludeFoldersAsPrefixes

tritone:
Adds support for IncludeFoldersAsPrefixes to object listing, which allows managed folders to be included in the results.

This is not yet implemented in gRPC so if it's called on a gRPC client then an Unimplemented error is returned.


frankyn (approved):

storage/integration_test.go
```diff
@@ -1377,6 +1376,115 @@ func TestIntegration_ObjectIterationMatchGlob(t *testing.T) {
...
+		// in the veneer layer.
+		call := client.raw.ManagedFolders.Insert(newBucketName, &raw.ManagedFolder{Name: "mf"})
+		mf, err := call.Context(ctx).Do()
+		if err != nil {
+			t.Fatalf("creating managed folder: %v", err)
+		}
+
+		t.Cleanup(func() {
+			call := client.raw.ManagedFolders.Delete(newBucketName, mf.Name)
+			call.Context(ctx).Do()
```
BrennaEpp, line 1433: Do we want to catch the error here? I ask because while I know other buckets that don't get properly cleaned up for whatever reason are killed when re-running the tests, but not sure if it would work on a bucket that contains managed folders. Or maybe managed folders should be added to the killBucket function?

BrennaEpp (approved):
Thanks Chris! Just added a comment on cleanup but otherwise LGTM

storage/integration_test.go
tritone: I would do this but I don't think it's workable right now because killBucket must work for gRPC clients as well. Will add a to-do.

PR #9215, opened 2024-01-04
feat(storage): add maxAttempts RetryOption

Tulsishah:
Issue: By specifying a count limit for retries using max-retry-count, we can ensure that applications stop automatically retrying operations after that count. This allows you to proactively detect and investigate any persistent issues, rather than letting applications keep attempting to retry indefinitely. This can be particularly beneficial for troubleshooting purposes, as it helps pinpoint the root cause of the failure more quickly. 

Adding to configure max retry count in RetryOption to the storage package.

### Testing details
Manual - NA
Unit tests - Added
Integration tests - NA


storage/storage.go
```diff
@@ -2076,6 +2076,21 @@ func (wb *withBackoff) apply(config *retryConfig) {
 	config.backoff = &wb.backoff
 }
 
+// WithMaxRetry sets the maximum number of retries for operations that may fail.
+func WithMaxRetry(maxRetryCount int) RetryOption {
```
codyoss, line 2080: nit: maybe `WithMaxAttempts`? Retry is implied by the type so I don't think it needs to be in the name

Tulsishah: Done

storage/invoke_test.go
```diff
@@ -173,6 +171,38 @@ func TestInvoke(t *testing.T) {
...
+			desc:              "non-idempotent retriable error retried when policy is RetryAlways till max-retry-count",
+			count:             4,
+			initialErr:        &googleapi.Error{Code: 500},
+			finalErr:          nil,
+			isIdempotentValue: false,
+			retry:             &retryConfig{policy: RetryAlways, maxRetryCount: 2},
+			expectFinalErr:    false,
+		},
+		{
+			desc:              "non-idempotent retriable error retried when policy is RetryNever with max-retry-count set",
```
tritone, line 184: I would expect this one not to be retried?

storage/invoke_test.go
```diff
@@ -188,7 +218,7 @@ func TestInvoke(t *testing.T) {
 				headers := callctx.HeadersFromContext(ctx)
 				gotClientHeader = headers["x-goog-api-client"][0]
 				gotIdempotencyHeader = headers["x-goog-gcs-idempotency-token"][0]
-				if counter <= test.count {
+				if (counter <= test.count) || (test.retry != nil && counter == test.retry.maxRetryCount) {
```
tritone, line 221: This seems like it shouldn't be necessary? Shouldn't the call to run cut retries off at the correct count without this?

storage/invoke.go
```diff
@@ -71,6 +71,9 @@ func run(ctx context.Context, call func(ctx context.Context) error, retry *retry
 		ctxWithHeaders := setInvocationHeaders(ctx, invocationID, attempts)
 		err = call(ctxWithHeaders)
 		attempts++
+		if attempts == retry.maxRetryCount {
```
tritone, line 74: Attempts starts at 1, so I would expect this to stop when attempts > maxRetryCount, not ==

storage/invoke_test.go
```diff
@@ -173,6 +171,38 @@ func TestInvoke(t *testing.T) {
...
+			initialErr:        io.ErrNoProgress,
+			finalErr:          nil,
+			isIdempotentValue: true,
+			retry: &retryConfig{
+				shouldRetry: func(err error) bool {
+					return err == io.ErrNoProgress
+				},
+				maxRetryCount: 2,
+			},
+			expectFinalErr: false,
```
tritone, line 204: Can we have a test case where maxRetryCount is set, but equal to `count` and `expectFinalErr` is true? This would verify that we don't get cut off early.

Tulsishah: Done

storage/storage.go
```diff
@@ -2076,6 +2076,23 @@ func (wb *withBackoff) apply(config *retryConfig) {
...
+	}
+}
+
+type withMaxAttempts struct {
+	maxAttempts int
+}
+
+func (wb *withMaxAttempts) apply(config *retryConfig) {
+	config.maxAttempts = wb.maxAttempts
+}
```
tritone, line 2094: As I said before, I think explicitly passing `WithMaxAttempts(0)` should mean no retries. We could implement this by setting the policy to `RetryNever` in the config if the user passed in 0 (since the default for an int field is 0, so we can't go based on that -- alternatively we could change the datatype to a `*int` internally, but I think using policy is simplest). Let's make this change and also add a test case for it in storage_test, and document the behavior of 0 in the doc for WithMaxAttempts

storage/invoke_test.go
```diff
@@ -203,6 +242,10 @@ func TestInvoke(t *testing.T) {
 			if !test.expectFinalErr {
 				wantAttempts = 1
 			}
+			if test.retry != nil && test.retry.maxAttempts != 0 && test.retry.policy != RetryNever {
+				wantAttempts = 1 + test.retry.maxAttempts
```
tritone, line 246: This seems off-by-1. After thinking about this more I think with the name `MaxAttempts` we should have the semantics be the number of total calls including all retries. So if the user says `WithMaxAttempts(3)` then we do up to 3 total calls (2 retries). Does this make sense to you?

If so let's 1. clarify the docs 2. Set L73 in invoke to be `>= rather than `>` 3. remove the `1 +` here.

tritone (commented):
Two more minor changes to semantics. Otherwise looks great.

storage/invoke_test.go
Tulsishah: I think with 3 retries configured `(WithMaxAttempts(3))`, the function will be called a maximum of 4 times. The first call is the initial attempt, and the subsequent calls are retries triggered by failures.

storage/storage.go
Tulsishah: What if someone did first `WithMaxAttempts(0)` and then `WithPolicy(RetryAlways)`. It will override the policy from `RetryNever` to `RetryAlways`

storage/invoke_test.go
```diff
@@ -193,6 +250,9 @@ func TestInvoke(t *testing.T) {
 				}
 				return test.finalErr
 			}
+			if test.retry != nil && test.retry.maxAttempts != nil && *test.retry.maxAttempts == 0 {
```
tritone, line 253: I think should be unnecessary if you are using the pointer?

storage/storage.go
```diff
@@ -2076,6 +2076,26 @@ func (wb *withBackoff) apply(config *retryConfig) {
...
+		maxAttempts: maxAttempts,
+	}
+}
+
+type withMaxAttempts struct {
+	maxAttempts int
+}
+
+func (wb *withMaxAttempts) apply(config *retryConfig) {
+	if wb.maxAttempts == 0 {
```
tritone, line 2093: Again, this should be unnecessary with the pointer implementation?

storage/storage.go
```diff
@@ -2111,7 +2131,9 @@ type withPolicy struct {
 }
 
 func (ws *withPolicy) apply(config *retryConfig) {
-	config.policy = ws.policy
+	if (config.maxAttempts != nil && *config.maxAttempts != 0) || config.maxAttempts == nil {
```
tritone, line 2134: Also seems unnecessary

tritone (commented):
With the pointer implementation for the field in retryConfig, it should be unnecessary to do any manipulation of the Policy logic. Otherwise looks good.

storage/storage.go
```diff
@@ -2076,6 +2076,23 @@ func (wb *withBackoff) apply(config *retryConfig) {
 	config.backoff = &wb.backoff
 }
 
+// WithMaxAttempts configures a maximum number of retries for potentially failing operations.
```
tritone, line 2079: Let's clarify the docs to distinguish between attempts and retries. How about this:

```
// WithMaxAttempts configures the maximum number of times an API call can be made
// in the case of retryable errors.
// For example, if you set WithMaxAttempts(5), the operation will be attempted up to 5
// times total (initial call plus 4 retries).
// Without this setting, operations will continue retrying indefinitely
// until either the context is canceled or a deadline is reached.
```

tritone (approved):
One more docs clarification, otherwise looks good -- feel free to fix & merge.

PR #9220, opened 2024-01-04
fix(bigtable): Fix deadline exceeded conformance test

bhshkh:
**Issue**:  TestReadRow_Generic_DeadlineExceeded test: 
https://github.com/googleapis/cloud-bigtable-clients-test/blob/af480c9811ae17c6dc01fe4959e7ca0f5c7afa08/tests/readrow_test.go#L81-L82

fails with below errors
```
=== RUN   TestReadRow_Generic_DeadlineExceeded
[Servr log] 2024/01/04 14:02:17 There is 10s sleep on the server side
    readrow_test.go:118: 
        	Error Trace:	cloud-bigtable-clients-test/tests/readrow_test.go:118
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 0
        	Test:       	TestReadRow_Generic_DeadlineExceeded
    readrow_test.go:121: 
        	Error Trace:	cloud-bigtable-clients-test/tests/readrow_test.go:121
        	Error:      	Not equal: 
        	            	expected: 4
        	            	actual  : 2
        	Test:       	TestReadRow_Generic_DeadlineExceeded
--- FAIL: TestReadRow_Generic_DeadlineExceeded (2.01s)
FAIL
exit status 1
FAIL	github.com/googleapis/cloud-bigtable-clients-test/tests	2.267s
```

**Cause**: 
1. The ReadRow method sends ReadRowsRequest rpc. Since only one row needs to be read, the test validates that row limit should be 1. The library does not set any row limit and thus, the error:
```
Error:      	Not equal: 
        	expected: 1
        	actual  : 0
```
2.   The ReadRow method returns the context deadline exceeded error as is. This error is not in GRPC error format.

**Fix**:
1. Send row limit in the ReadRowsRequest
2. Obtain status code from context related errors and create errors in GRPC status format. So, the code from error is Unknown i.e. 2 instead of [DeadlineExceeded i.e. 4](https://github.com/grpc/grpc-go/blob/cb3ae76/codes/codes.go#L73)

After fix:
```
=== RUN   TestReadRow_Generic_DeadlineExceeded
[Servr log] 2024/01/04 14:04:30 There is 10s sleep on the server side
--- PASS: TestReadRow_Generic_DeadlineExceeded (2.01s)
PASS
ok  	github.com/googleapis/cloud-bigtable-clients-test/tests	3.258s
```


gkevinzheng (approved):

kolea2 (approved):
lgtm, but can we make the PR description what the fix is for (ie why was the conformance test previously failing?)

bigtable/bigtable.go
```diff
@@ -278,13 +278,27 @@ func (t *Table) ReadRows(ctx context.Context, arg RowSet, f func(Row) bool, opts
 		return err
 	}, retryOptions...)
 
+	// Convert error to grpc status error
```
kolea2, line 281: also nit - is this logic elsewhere in the codebase? Might be nice to make it a util if so

bhshkh: Not being used anywhere else

bhshkh:
> lgtm, but can we make the PR description what the fix is for (ie why was the conformance test previously failing?)

Updated

PR #9224, opened 2024-01-05
feat(bigquery): expose query id on row iterator if available

shollyman:
This feature plumbs through the query ID, if available, as part of the row iterator.  Currently only queries run via the jobs.query RPC have query ID exposed.  This is related to the query preview features exposed in PR 8653.


PhongChuong (approved):
s/Id/ID

shollyman:
> s/Id/ID

done.

PhongChuong (approved):

PR #9225, opened 2024-01-05
chore(storage): tidy benchmarking module

tritone:
This seems to have been missed in an earlier update and now presubmits are failing on #9215


frankyn (approved):

tritone:
Not needed, Chris Smith fixed this earlier.

PR #9236, opened 2024-01-08
fix(bigquery): support more timestamp formats for query param

### LINKED ISSUE #9221: bigquery: timestamp parsing issue when fetching parameterized query jobs and their config
**Client**

BigQuery

**Environment**

MacOS

**Go Environment**

```
$ go version
go version go1.21.1 darwin/arm64

$ go env
GO111MODULE=''
GOARCH='arm64'
GOBIN=''
GOCACHE='/Users/hsvnsson/Library/Caches/go-build'
GOENV='/Users/hsvnsson/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='arm64'
GOHOSTOS='darwin'
GOINSECURE=''
GOMODCACHE='/Users/hsvnsson/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='darwin'
GOPATH='/Users/hsvnsson/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/go/pkg/tool/darwin_arm64'
GOVCS=''
GOVERSION='go1.21.1'
GCCGO='gccgo'
AR='ar'
CC='clang'
CXX='clang++'
CGO_ENABLED='1'
GOMOD='/dev/null'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -ffile-prefix-map=/var/folders/6w/pc370vsn5c7dtm7gsxm1y9440000gn/T/go-build217012697=/tmp/go-build -gno-record-gcc-switches -fno-common'
```

**Code**

```go
package main

import (
	"context"
	"errors"
	"flag"
	"log"
	"time"

)

func main() {
	ctx := context.Background()

	serviceAccountPath := flag.String("service-account", "", "Path to service account JSON file")
	project := flag.String("project", "", "Project")
	flag.Parse()

	client, err := bigquery.NewClient(ctx, *project, option.WithCredentialsFile(*serviceAccountPath))
	if err != nil {
	}
	defer client.Close()

	it := client.Jobs(ctx)
	it.AllUsers = true
	it.State = bigquery.Done
	it.MinCreationTime = time.Now().Add(-1 * time.Hour)
	it.MaxCreationTime = time.Now()

	for {
		job, err := it.Next()
		if errors.Is(err, iterator.Done) {
			break
		}
		if err != nil {
			log.Fatalf("Failed to get next job: %v", err)
		}

		jobConfig, err := job.Config()
		if err != nil {
			log.Fatalf("Failed to parse job: %v", err)
		}

		_ = jobConfig
	}
}
```

**Expected behavior**
Following the [example in the docs](https://cloud.google.com/bigquery/docs/parameterized-queries#using_timestamps_in_parameterized_queries) to submit a BigQuery query with a timestamp parameter and then listing jobs using the golang sdk it should be possible to get the job configs without any errors.

Setup:
```
bq query \
   --use_legacy_sql=false \
   --parameter='ts_value:TIMESTAMP:2016-12-07 08:00:00' \
   'SELECT
     TIMESTAMP_ADD(@ts_value, INTERVAL 1 HOUR);'
```

Then using the code provided above, the call to `job.Config()` should return a properly parsed job config.

**Actual behavior**

Error while parsing timestamp:
```
$ go run main.go -project some-project -service-account ./creds.json
2024/01/04 17:47:40 Failed to parse job: parsing time "2016-12-07T08:00:00" as "2006-01-02T15:04:05.999999999Z07:00": cannot parse "" as "Z07:00"
exit status 1
```

**Additional context**

if a timezone is included in the timestamp like `--parameter='ts_value:TIMESTAMP:2016-12-07 08:00:00+00:00'` then it works but since that's not required and also not included in the example then it would be good if the library handles that format as well.

alvarowolfx:
BigQuery accepts timestamp parameters in many different formats, so when parsing the query parameters, we need to be able to support all of them. As reported on #9221, if a query is created using bq-cli (or any other system) with the `time.DateTime(2006-01-02 15:04:05)` format, our SDK was failing to parse the Job Config when retrieving it from the backend system. 

Resolves #9221


shollyman (approved):
Generally fine with the approach.  If we find we need this parsing logic elsewhere, I'd definitely support a utility function so we do the parsing in a consistent fashion.

PR #9242, opened 2024-01-11
feat(internal/trace): export internal/trace package constants and vars

quartzmo:
* Export constants and vars for use by handwritten clients

### DIFF
--- a/internal/trace/trace.go
+++ b/internal/trace/trace.go
@@ -32,13 +32,31 @@ import (
 )
 
 const (
-	TelemetryPlatformTracingOpenCensus    = "opencensus"
+	// TelemetryPlatformTracingOpenCensus is the value to which the environment
+	// variable GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING should be
+	// set to enable OpenCensus tracing.
+	TelemetryPlatformTracingOpenCensus = "opencensus"
+	// TelemetryPlatformTracingOpenCensus is the value to which the environment
+	// variable GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING should be
+	// set to enable OpenTelemetry tracing.
 	TelemetryPlatformTracingOpenTelemetry = "opentelemetry"
-	TelemetryPlatformTracingVar           = "GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING"
-	OpenTelemetryTracerName               = "cloud.google.com/go"
+	// TelemetryPlatformTracingOpenCensus is the name of the environment
+	// variable that can be set to change the default tracing from OpenCensus
+	// to OpenTelemetry.
+	TelemetryPlatformTracingVar = "GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING"
+	// OpenTelemetryTracerName is the name given to the OpenTelemetry Tracer
+	// when it is obtained from the OpenTelemetry TracerProvider.
+	OpenTelemetryTracerName = "cloud.google.com/go"
 )
 
 var (
+	// OpenTelemetryTracingEnabled is true if the environment variable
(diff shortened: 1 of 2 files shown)


codyoss (approved):

internal/trace/trace.go
```diff
@@ -32,16 +32,15 @@ import (
...
+	TelemetryPlatformTracingOpenTelemetry = "opentelemetry"
+	TelemetryPlatformTracingVar           = "GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING"
 )
 
 var (
-	// TODO(chrisdsmith): Should the name of the OpenTelemetry tracer be public and mutable?
-	openTelemetryTracerName     string = "cloud.google.com/go"
-	openTelemetryTracingEnabled bool   = strings.EqualFold(strings.TrimSpace(
-		os.Getenv(telemetryPlatformTracingVar)), telemetryPlatformTracingOpenTelemetry)
+	OpenTelemetryTracerName     string = "cloud.google.com/go"
```
harshachinta, line 41: Random thought. Will exporting this not cause problems?
For ex, if spanner handwritten client overrides this "cloud.google.com/go/spanner", then the same change will get affected to other for ex: bigquery etc.

Am i missing something here?

internal/trace/trace.go
```diff
@@ -32,16 +32,15 @@ import (
 )
 
 const (
-	telemetryPlatformTracingOpenCensus    = "opencensus"
-	telemetryPlatformTracingOpenTelemetry = "opentelemetry"
-	telemetryPlatformTracingVar           = "GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING"
+	TelemetryPlatformTracingOpenCensus    = "opencensus"
+	TelemetryPlatformTracingOpenTelemetry = "opentelemetry"
+	TelemetryPlatformTracingVar           = "GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING"
```
harshachinta, line 37: same for ENV. Please let me know if i am missing something.

internal/trace/trace.go
codyoss: Yes, these are not meant to be changed as it would be a global change across multiple clients. Maybe we need to understand your usercase more

internal/trace/trace.go
quartzmo: cannot assign to `TelemetryPlatformTracingVar` because it is a `const`

internal/trace/trace.go
quartzmo: Good call, let's not export this, thanks!

quartzmo: Actually, I propose we keep it exported but make it a `const`. It was a `var` only because originally I thought it might need to be mutable. That was incorrect.

internal/trace/trace.go
```diff
@@ -32,16 +32,15 @@ import (
...
+	OpenTelemetryTracerName               = "cloud.google.com/go"
 )
 
 var (
-	// TODO(chrisdsmith): Should the name of the OpenTelemetry tracer be public and mutable?
-	openTelemetryTracerName     string = "cloud.google.com/go"
-	openTelemetryTracingEnabled bool   = strings.EqualFold(strings.TrimSpace(
-		os.Getenv(telemetryPlatformTracingVar)), telemetryPlatformTracingOpenTelemetry)
+	OpenTelemetryTracingEnabled bool = strings.EqualFold(strings.TrimSpace(
```
harshachinta, line 42: Should we be adding a comment for `OpenTelemetryTracingEnabled` saying that this field should be modified only for unit tests by go client libraries and should not be modified by any other applications?

harshachinta (approved):

internal/trace/trace.go
quartzmo: Sure, good idea.

quartzmo: Done, thanks.

PR #9254, opened 2024-01-12
feat(spanner): add OpenTelemetry implementation

harshachinta:
This PR adds support for [OpenTelemetry](https://opentelemetry.io/) Instrumentation for Traces and Metrics. 

Customer applications should add dependency for [Metrics SDK](https://opentelemetry.io/docs/languages/go/instrumentation/#initialize-metrics), [Traces SDK](https://opentelemetry.io/docs/languages/go/instrumentation/#getting-a-tracer) and required [exporters](https://opentelemetry.io/docs/languages/go/exporters/).

**Metrics**
1. By default, OpenTelemetry metrics are not enabled. To enable call `spanner.EnableOpenTelemetryMetrics()` in startup of your application.
2. Create MeterProvider and inject it via [ClientConfig](https://github.com/googleapis/google-cloud-go/blob/faccb68deeb18ad51875042242a1141285448bbb/spanner/client.go#L116) or register as Global.
```
client, err := spanner.NewClientWithConfig(ctx, <database-id>, spanner.ClientConfig{
	OpenTelemetryMeterProvider: meterProvider,
})
```
**Tracing**
1. By default, OpenTelemetry traces are not enabled. To enable set environment variable `GOOGLE_API_GO_EXPERIMENTAL_TELEMETRY_PLATFORM_TRACING` to `opentelemetry`
2. Create TracerProvider and register as Global


spanner/benchmarks_oc_ot_test.go
```diff
@@ -0,0 +1,250 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package spanner
+
+import (
+	"context"
+	"fmt"
```
rahul2393, line 21: Is this file expected?

spanner/benchmarks_ot_test.go
```diff
@@ -0,0 +1,374 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
+)
+
+/*
+var muElapsedTimes sync.Mutex
+var elapsedTimes []time.Duration
```
rahul2393, line 35: Is this file expected?

spanner/benchmarks_test.go
```diff
@@ -0,0 +1,392 @@
...
+
+	"go.opencensus.io/trace"
+	"google.golang.org/api/option"
+
+	"contrib.go.opencensus.io/exporter/stackdriver"
+	"google.golang.org/api/iterator"
+)
+
+var muElapsedTimes sync.Mutex
+var elapsedTimes []time.Duration
```
rahul2393, line 38: Is this file expected?

spanner/batch.go
```diff
@@ -149,6 +149,9 @@ func (t *BatchReadOnlyTransaction) PartitionReadUsingIndexWithOptions(ctx contex
 			trace.TracePrintf(ctx, nil, "Error in recording GFE Latency. Try disabling and rerunning. Error: %v", err)
 		}
 	}
+	if metricErr := recordGFELatencyMetricsOT(ctx, md, "PartitionReadUsingIndexWithOptions", t.otConfig); metricErr != nil {
```
rahul2393, line 152: With opencensus user had option to disable GFE latency metrics, after this change are we going to enable GFE latency always?

harshachinta: Yes. For OpenTelemetry we have provided only one option spanner.EnableOpenTelemetryMetrics which will give both session and gfe metrics by default.

spanner/benchmarks_oc_ot_test.go
harshachinta: Added for benchmarking on VM . Removed it now

spanner/benchmarks_ot_test.go
harshachinta: Added for benchmarking on VM . Removed it now

spanner/ot_metrics.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func initializeMetricInstruments(config *openTelemetryConfig, logger *log.Logger) {
+	if !IsOpenTelemetryMetricsEnabled() {
+		return
+	}
+	meter := config.meterProvider.Meter(OtInstrumentationScope, metric.WithInstrumentationVersion(internal.Version))
+
+	openSessionCountInstrument, err := meter.Int64ObservableGauge(
+		metricsPrefix+"open_session_count1",
```
olavloite, line 94: Why do we add `1` as a suffix to `open_session_count` here?

spanner/ot_metrics.go
```diff
@@ -0,0 +1,258 @@
...
+		metric.WithDescription("Number of sessions currently opened"),
+		metric.WithUnit("1"),
```
olavloite, line 96: What does `metric.WithUnit("1")` mean? Does that meant that it is an integer?

spanner/ot_metrics.go
```diff
@@ -0,0 +1,258 @@
...
+	attributeNumInUseSessions = attributeKeyType.String("num_in_use_sessions")
+	attributeNumSessions      = attributeKeyType.String("num_sessions")
+	// openTelemetryMetricsEnabled is used to track if OpenTelemetry Metrics need to be recorded
+	openTelemetryMetricsEnabled = false
+	// mutex to avoid data race in reading/writing the above flag
+	otMu = sync.RWMutex{}
+)
+
+func getOpenTelemetryConfig(mp metric.MeterProvider, logger *log.Logger, sessionClientID string, db string) (*openTelemetryConfig, error) {
+	logf(logger, "Inside getOpenTelemetryConfig()")
```
olavloite, line 52: This seems like a left-over from debugging?

spanner/ot_metrics.go
```diff
@@ -0,0 +1,258 @@
...
+
+	attributeNumInUseSessions = attributeKeyType.String("num_in_use_sessions")
+	attributeNumSessions      = attributeKeyType.String("num_sessions")
+	// openTelemetryMetricsEnabled is used to track if OpenTelemetry Metrics need to be recorded
+	openTelemetryMetricsEnabled = false
+	// mutex to avoid data race in reading/writing the above flag
+	otMu = sync.RWMutex{}
+)
+
+func getOpenTelemetryConfig(mp metric.MeterProvider, logger *log.Logger, sessionClientID string, db string) (*openTelemetryConfig, error) {
```
olavloite, line 51: This method is called `getOpenTelemetryConfig`, but it seems that a more descriptive name would be `getOrCreate...`. That also makes it clearer that it is not safe to call this method from multiple goroutines at the same time. Should we add some synchronization here to ensure that it does become safe? Or do we have some other means in place that prevents that from happening?

spanner/ot_metrics.go
```diff
@@ -0,0 +1,258 @@
...
+	attributeKeyDatabase   = attribute.Key("database")
+	attributeKeyInstance   = attribute.Key("instance_id")
+	attributeKeyLibVersion = attribute.Key("library_version")
+	attributeKeyType       = attribute.Key("type")
+	attributeKeyMethod     = attribute.Key("grpc_client_method")
+
+	attributeNumInUseSessions = attributeKeyType.String("num_in_use_sessions")
+	attributeNumSessions      = attributeKeyType.String("num_sessions")
+	// openTelemetryMetricsEnabled is used to track if OpenTelemetry Metrics need to be recorded
+	openTelemetryMetricsEnabled = false
```
olavloite, line 46: These are all global variables. What is our strategy to prevent both errors and confusion if someone tries to enable/disable open telemetry multiple times in their application. For example, what happens if someone:
1. Enables Open Telemetry.
2. Creates a Spanner client
3. Disables Open Telemetry
4. Creates another Spanner client

Is the above allowed? If so, what will be the behavior of the two Spanner clients?

spanner/ot_metrics.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	attr := otConfig.attributeMap
+	if len(md.Get("server-timing")) == 0 && otConfig.gfeHeaderMissingCount != nil {
+		otConfig.gfeHeaderMissingCount.Add(ctx, 1, metric.WithAttributes(attr...))
+		return nil
+	}
+	serverTiming := md.Get("server-timing")[0]
+	gfeLatency, err := strconv.Atoi(strings.TrimPrefix(serverTiming, "gfet4t7; dur="))
+	if !strings.HasPrefix(serverTiming, "gfet4t7; dur=") || err != nil {
+		return err
```
olavloite, line 251: It's a bit of a theoretical case, but shouldn't we split this into two separate things:
1. `err != nil`: return the error
2. `!strings.HasPrefix(serverTiming, "gfet4t7; dur=")`: This should be considered the same as a missing GFE header (or malformed GFE header). Now we just return `nil` when that happens, is that intentional?

spanner/client.go
```diff
@@ -276,16 +302,29 @@ func NewClientWithConfig(ctx context.Context, database string, config ClientConf
 	if config.Compression == gzip.Name {
 		md.Append(requestsCompressionHeader, gzip.Name)
 	}
+
 	// Create a session client.
 	sc := newSessionClient(pool, database, config.UserAgent, sessionLabels, config.DatabaseRole, config.DisableRouteToLeader, md, config.BatchTimeout, config.Logger, config.CallOptions)
 
+	// Create a OpenTelemetry configuration
+	otConfig, err := getOpenTelemetryConfig(config.OpenTelemetryMeterProvider, config.Logger, sc.id, database)
```
olavloite, line 310: What happens if a user tries to call `NewClientWithConfig` in multiple goroutines at the same time? I don't think `getOpenTelemetryConfig` will work correctly in that case. And that again is an indication that the name of the method should probably be `getOrCreate...`

spanner/ot_metrics.go
harshachinta: `WithUnit` is used to set the instrument unit.
I followed the same convention used for OpenCensus where we set "1" for [dimensionless](https://github.com/census-instrumentation/opencensus-go/blob/v0.24.0/stats/units.go#L22) unit or count as per [UCCM](https://ucum.org/ucum).

Example in OC:
https://github.com/googleapis/google-cloud-go/blob/1cf8cf1d32c0f84ff8e4fd388ebcce5f5866bed1/spanner/stats.go#L94-L97

spanner/ot_metrics.go
harshachinta: I was doing some testing. There will be no suffix "1" for metric names. I have reverted it back.

spanner/ot_metrics.go
harshachinta: These are global un-exported variables which are not available for public access. We expose only one public method `EnableOpenTelemetryMetrics()` which is called by customers once at the beginning of their application to use OpenTelemetry instrumentation for metrics. There is no method to disable OpenTelemetry. So there is no way to enable and disable OpenTelemetry multiple times.

**The example mentioned above is not possible**.

Possible Scenarios
1. Scenario 1:
       -  Enables Open Telemetry
       - Creates Spanner client 1
       - Creates Spanner client 2
      
       Observed Behavior - OpenTelemetry metrics get recorded for both the spanner clients.
2. Scenario 2:
       - Creates Spanner client 1
       -  Enables Open Telemetry
       - Creates Spanner client 2
      
       Observed Behavior - OpenTelemetry metrics get recorded only for spanner client 2.

spanner/ot_metrics.go
harshachinta: `createOpenTelemetryConfig ` would be the the right name here, as it gets called only **once** when a new client is created using `NewClient()` or `NewClientWithConfig()` and doesn't need synchronization as there is no shared data.

This method behaves similar to existing [newSessionClient()](https://github.com/googleapis/google-cloud-go/blob/1cf8cf1d32c0f84ff8e4fd388ebcce5f5866bed1/spanner/client.go#L280) method in client.go.

spanner/ot_metrics.go
harshachinta: Removed

spanner/client.go
harshachinta: As mentioned above, this should be called `createOpenTelemetryConfig`.

When `NewClientWithConfig` gets called in multiple goroutines, the behavior would be similar to `newSessionClient`, because `createOpenTelemetryConfig` does not have any shared resource and its functionality is to create a new config object and return it back to client.

Please correct me if I am missing something here.

spanner/ot_metrics.go
harshachinta: In case of malformed GFE header we return an error, not `nil`. 
We are not counting malformed GFE as missing GFE because in this case we get a GFE header but it is not in required format.

spanner/ot_metrics.go
```diff
@@ -0,0 +1,255 @@
...
+	"go.opentelemetry.io/otel/metric"
+	"google.golang.org/grpc/metadata"
+)
+
+// OtInstrumentationScope is the instrumentation name that will be associated with the emitted telemetry.
+const OtInstrumentationScope = "cloud.google.com/go"
+const metricsPrefix = "spanner/"
+
+var (
+	attributeKeyClientID   = attribute.Key("client_id")
```
olavloite, line 36: nit: could the variables here that never change be `const`?

spanner/ot_metrics.go
olavloite: Ah, thanks for explaining. That looks good to me.

spanner/ot_metrics.go
olavloite: ack

spanner/ot_metrics.go
olavloite: ack

note that this could still return `err=nil`, if the actual text is only a number without any other prefixes or suffixes. The reason is that `strings.TrimPrefix` returns the original string if the string did not have the prefix.

This for example prints `4`:

```go
fmt.Println(strconv.Atoi(strings.TrimPrefix("4", "World")))
```

spanner/ot_metrics.go
olavloite: ack

spanner/ot_metrics.go
olavloite: Ah, that explains. Thanks!

spanner/client.go
olavloite: I re-read that piece of code, and I agree with you that it is setup in the correct way, so each client gets its own config, and that there does not seem to be any possible data races or similar.

olavloite (approved):

rahul2393 (approved):

PR #9264, opened 2024-01-16
chore(storage): add continuous read workload [benchmarks]

BrennaEpp:
Note that the workload only works under specific circumstances (mentioned in the readme), which is fine I think - if we need additional functionality like object ranges we can add later.


storage/internal/benchmarks/continuous_reads.go
```diff
@@ -0,0 +1,202 @@
...
+// BENCHMARK
+//   - Grab a storage client from the pool.
+//   - Check if the timeout is exceeded; if so, compile p50, p90 and p99 and return.
+//   - Queue a goroutine that performs a download. This goroutine will run as
+//     soon as there is an available worker. The number of workers that run
+//     concurrently is set in the command-line.
+//   - This goroutine chooses the object to download by receiving an object name
+//     from the channel. At the end of the download, it will return that name to
+//     the channel. That way, no goroutines are reading from the same object at
+//     the same time.
```
tritone, line 53: Any particular reason for this? Prince's benchmark is just reading one object; we could just have a round robin of objects to read in a slice without needing to check them out or queue using a channel.

storage/internal/benchmarks/continuous_reads.go
```diff
@@ -0,0 +1,202 @@
...
+					bucket:              r.bucketName,
+					object:              object,
+					rangeStart:          rangeStart,
+					rangeLength:         rangeLength,
+					downloadToDirectory: r.directoryPath,
+					timeout:             r.opts.timeoutPerOp,
+				})
+
+				if err != nil {
+					r.results = append(r.results, timeTaken)
```
tritone, line 169: This seems potentially not thread safe? Do we need a mutex? Might be good to run it with the race detector to check.

storage/internal/benchmarks/continuous_reads.go
```diff
@@ -0,0 +1,202 @@
...
+//
+// SET UP
+//   - Initialize structs to populate with the benchmark results.
+//   - Select a random API to use to upload and download the object, unless an
+//     API is set in the command-line,
+//   - Select a random size for the objects; this size will
+//     be between two values configured in the command-line.
+//   - Create a directory of objects of that size in GCS, uploading random
+//     contents directly from memory. The number of objects created is equal to
+//     the command-line configuration for the number of objects in a directory.
```
tritone, line 42: I don't see a warmup step here. Can we add one?

storage/internal/benchmarks/continuous_reads.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// Select object size
+	objectSize := opts.objectSize
+	if objectSize == 0 {
+		objectSize = randomInt64(opts.minObjectSize, opts.maxObjectSize)
+	}
+	r.objectSize = objectSize
+
+	// Make a temp dir for this run
+	dir, err := os.MkdirTemp("", "benchmark-experiment-")
```
tritone, line 99: I thought we had discussed downloading to `io.Discard` for now? Otherwise I think we'd run into disk issues.

storage/internal/benchmarks/README.md
```diff
@@ -32,7 +32,12 @@ This runs 1000 iterations on 512kib to 2Gib files in the background, sending out
...
-| -directory_num_objects | total number of objects in a directory (directory will only contain files, <br> no subdirectories); only applies to workload 6 | any positive integer | `1000` |
+| -workload | `1` will run a w1r3 (write 1 read 3) benchmark <br> `6` will run a benchmark uploading and downloading (once each) <br> a single directory with `-directory_num_objects` number of files (no subdirectories) <br> `9`** will run a benchmark that does continuous reads on a directory with `directory_num_objects` | `1`, `6`, `9` | `1` |
+| -directory_num_objects | total number of objects in a directory (directory will only contain files, <br> no subdirectories); only applies to workload 6 and 9 | any positive integer | `1000` |
 
 \* required values
+
+\*\* Note that this workload is experimental and will not work under certain conditions. Here's a non-comprehensive list of notes on workload 9:
+ - output type must be `cloud-monitoring`
+ - it continues reading until the timeout is reached - the workload will always fail with deadline exceeded
```
tritone, line 42: Maybe we can add a timer so it can still exit zero once the timeout is reached?

tritone (commented):
Generally looking good, a few thoughts.

storage/internal/benchmarks/continuous_reads.go
BrennaEpp: In this case the benchmark does still download the objects to tmp. I could change it to io.Discard with a bit of plumbing but I don't think it's strictly necessary - since it overwrites existing objects disk space is not an issue. The channel here is so that threads aren't writing to the same object at once. Do you have a concern about overwriting objects in this manner?

storage/internal/benchmarks/continuous_reads.go
BrennaEpp: Yeah, you're right. I'll send it to a channel instead.

storage/internal/benchmarks/continuous_reads.go
BrennaEpp: Do you want to have one that just does reads? Otherwise, I think the w1r3 warmups are sufficient.

storage/internal/benchmarks/continuous_reads.go
BrennaEpp: See my other comment.

storage/internal/benchmarks/README.md
BrennaEpp: I made a small change that will cause it to exit zero instead (returning nil instead of the error from ctx.done)

storage/internal/benchmarks/continuous_reads.go
tritone: Ah okay, I think I just didn't follow the logic completely that existing local files are being overwritten. Seems fine I guess.

storage/internal/benchmarks/continuous_reads.go
tritone: w1r3 warmups are fine, just wasn't clear to me that runs of this benchmark could be combined with the warmup option. Maybe just something to clarify in documentation.

tritone (approved):

storage/internal/benchmarks/continuous_reads.go
BrennaEpp: Done

storage/internal/benchmarks/continuous_reads.go
BrennaEpp: I added a note in the comment about this

PR #9270, opened 2024-01-17
test(bigtable): Clean up test clusters

bhshkh:
**Issue**: Test fails with error:
```
=== RUN   TestIntegration_AdminCopyBackup
    integration_test.go:3099: CreateCluster: rpc error: code = FailedPrecondition desc = Cannot create multiple clusters in the same zone.
```

**Cause**:
Cluster created in previous run did not get cleaned up

**Fix**:
- Create clusters and instances with proper prefix
- Retry cleanup post tests run


gkevinzheng (approved):

bigtable/integration_test.go
```diff
@@ -127,16 +133,27 @@ func cleanup(c IntegrationTestConfig) error {
...
 				return err
 			}
 			uT := time.Unix(t, 0)
 			if time.Now().After(uT.Add(timeUntilResourceCleanup)) {
-				iac.DeleteInstance(ctx, info.Name)
+				iac.DeleteInstance(ctx, instanceInfo.Name)
+			}
+		} else {
+			// Delete clusters created in existing instances
+			clusters, err := iac.Clusters(ctx, instanceInfo.Name)
```
kolea2, line 149: to confirm, is there ever a case where we want to keep the instance *and* cluster?

kolea2 (approved):
lgtm with one question

bigtable/integration_test.go
bhshkh: We don't want to keep the ones created by the integration test.

The ones created by user will not be deleted.
User can specify an existing instance and a existing cluster to use in tests via environment variables:
GCLOUD_TESTS_BIGTABLE_INSTANCE
GCLOUD_TESTS_BIGTABLE_CLUSTER

There are prefix checks at line 137 and 154 to determine whether resource was created by the test

PR #9279, opened 2024-01-19
test(logging/logadmin): multiple fixes for sinks_test.go

gkevinzheng:
- Objects in buckets will now have a TTL of 1 day, and buckets will be autodeleted in 36 hours. This is done to fix tests that fail because of too many buckets to delete timing out the tests.
- Added back defer statements for log sinks and bucket deletion cleanup.
- Use IAM instead of ACL to be able to run integration tests  across UBLA and non-UBLA
  environments


logging/logadmin/sinks_test.go
```diff
@@ -89,40 +111,16 @@ func initSinks(ctx context.Context) func() {
 		}
 	}
 	if integrationTest {
-		for _, bn := range bucketNames(ctx, storageClient) {
-			if bucketIDs.Older(bn, 24*time.Hour) {
```
tritone, line 93: I'm curious what's been flaky about this if the error is ignored.

tritone (approved):

cindy-peng (approved):

Issue #9304, opened 2024-01-25
all: better universe handling with project autodetection 


shollyman:
cloud.google.com/go/internal/detect contains the shared functionality the various clients leverage to do project autodetection, and under the hood looks for the project ID in the credentials.

We should probably try to make the right choice(s) for users wanting autodetection to also handle universe options as well, but possibly implies we need to do more complex option handling.  If nothing else, we need to accomodate universe setups in existing tests of DetectProjectID in the various clients, as these tests break in a non-GDU universe.

CC @quartzmo

quartzmo:
> we need to accomodate universe setups in existing tests of DetectProjectID in the various clients

Can you link a few of these tests?

shollyman:
part of this is the intermingling of google-cloud-go test creds, but https://github.com/search?q=repo%3Agoogleapis%2Fgoogle-cloud-go+DetectProjectID&type=code finds both the impls and tests

shollyman:
digging deeper, I think most of my problems can be solved by changes to/in internal/testutil, which provides helper TokenSources and deals with the custom credential resolution we use for tests

shollyman:
Closing this for now.

PR #9306, opened 2024-01-25
testing: add testutil.UniverseFromEnv

shollyman:
This PR allows resolution of the effective universe that may be defined as part of the credentials used for testing, making it possible for testutil consumers to properly register necessary client instantiation options (particularly option.WithUniverseDomain) when appropriate.

Related: https://github.com/googleapis/google-cloud-go/issues/9304


quartzmo (approved):
LGTM

codyoss:
@shollyman do we still want this feature?

shollyman:
Killing this for now, we've clearly survived without it.

PR #9311, opened 2024-01-26
fix(pubsub): move flow control release to callback completion

### LINKED ISSUE #9309: pubsub: expired messages clog up flow control
**Client**

PubSub

**Environment**

all

**Code**

```go
	sub.ReceiveSettings.NumGoroutines = 1
	sub.ReceiveSettings.MaxOutstandingMessages = 1
	sub.ReceiveSettings.MaxExtension = 30 * time.Second
	var received int32

	err = sub.Receive(ctx, func(_ context.Context, msg *pubsub.Message) {
		fmt.Println("message ", string(msg.Data), received)
		fmt.Fprintf(w, "Got message: %q\n", string(msg.Data))
		atomic.AddInt32(&received, 1)
		if received == 1 {
 			// do nothing
		} else {
			msg.Ack()
		}
		fmt.Println("done")
	})
```

**Expected behavior**

With low flow control set (e.g. `MaxOutstandingMessages = 1`), expired messages do not clog up flow control.

**Actual behavior**

Flow control is blocked since `release` is never called. No new messages can be processed.

**Additional context**

Currently when messages expire, it means the lease management mechanism deletes it from the map that tracks how long to process it for. However, cleanup of resources (including releasing flow control resources), happens only when the message's done function is called.

Flow control release should happen after the end of the user callback (passed into `Receive`).

hongalex:
This moves the flow control release to the end of the user callback rather than to the message `doneFunc`. This further emphasizes the importance of calling `ack` and `nack` within the callback, since calling `ack` in a goroutine could mean that flow control resources are released too early. This behavior is already [documented](https://pkg.go.dev/cloud.google.com/go/pubsub#section-readme:~:text=Once%20client%20code,when%20cancelling%20Receive.)

Fixes #9309
Internal issue: 322090131


pubsub/subscription.go
```diff
@@ -1389,24 +1388,19 @@ func (s *Subscription) Receive(ctx context.Context, f func(context.Context, *Mes
...
+
 					wg.Add(1)
 					// Make sure the subscription has ordering enabled before adding to scheduler.
 					var key string
 					if s.enableOrdering {
 						key = msg.OrderingKey
 					}
-					// TODO(deklerk): Can we have a generic handler at the
-					// constructor level?
+					msgLen := len(msg.Data)
```
shollyman, line 1400: I'm not super familiar with msg lifecycle here.  Is it possible for the user to alter msg.Data between acquire and release, which could cause the flow controller to get out of sync?  If so, it might make sense to compute the length once and pass that to acquire and release.

hongalex: At this point, the message hasn't been dispatched to the user yet. This is all pre-processing, where the message isn't changed.

shollyman (approved):

PR #9325, opened 2024-01-29
chore(internal/actions): add touch flag to changefinder

noahdietz:
If there is a need to force a bump/release of all submodules that changed in a previously submitted change, the `-touch` flag can be used to append a new line to the submodule `CHANGES.md` file.

This is best used when comparing already submitted changes. For example, a GAPIC generator change touches every gapic module, but the PR doesn't have nested commits for each changed module. The PR is submitted as is, perhaps because it was included with a bunch of other changes or for timing purposes, and the nested commits were not added with `changefinder` while the PR was open. In this case, we can do the following using `changefinder`:
1. checkout the merged commit with the target changes: `git checkout <some commit has>`
2. "save" this to a branch: `git checkout -b changed`
3. checkout the commit just before the target commit: `git checkout <the commit just before>`
4. "save" this to a branch : `git checkout -b base`
5. checkout the changes: `git checkout changed`
6. generate the nested commits + touch each module:
```
go run ./internal/actions/cmd/changefinder -q \
  -format=commit \
  -commit-scope=fix \
  -commit-message="describe the change" \
  -touch
```
7. checkout main, new branch, commit:
```
git checkout main && \
  git checkout -b bump-modules && \
  git commit -a -m 'chore: bump changed modules' && \
  git push
```


quartzmo (approved):
Bravo!!! 🎉 

What an excellent PR description! What do you think about adding this stepset to `changefinder/README.md`? If you're too busy, I'd be happy to do it in a follow-up PR.

noahdietz:
> What do you think about adding this stepset to `changefinder/README.md`?

Sure thing! Done.

PR #9347, opened 2024-02-01
fix(storage): fix v4 url signing for hosts that specify ports

BrennaEpp:
Previously, attempting to sign urls with the explicit hostname `storage.googleapis.com:443` would work with v2 signing but with v4 would cause signature not match errors. This change strips the port from the signing headers.


storage/integration_test.go
```diff
@@ -2129,6 +2129,14 @@ func TestIntegration_SignedURL(t *testing.T) {
 				headers: map[string][]string{"X-Goog-Foo": {"bar baz"}},
 				fail:    true,
 			},
+			{
+				desc: "Hostname v4",
+				opts: SignedURLOptions{
+					Hostname: "storage.googleapis.com:443",
```
tritone, line 2135: I didn't realize you could include a port for the prod endpoint like this; TIL.

tritone (approved):
Nice catch on this one!

PR #9368, opened 2024-02-05
feat(bigquery): switch all timestamp representations to int64 usec

shollyman:
With the recent resolutions to the discovery representation, we now correctly propagate the options for controlling timestamp representation.  This PR forces all usages to lossless representations.


alvarowolfx (approved):
I was double checking if that was going to affect Arrow parsing, as we also use the same `convertBasicType` method, but for Arrow timestamps we don't use it, so it's all good.

PR #9372, opened 2024-02-05
feat(bigtable): Add ignore_warnings flag to SetGcPolicy

andre-sampaio:
This exposes the flag `ignore_warnings` for `ModifyColumnFamiliesRequest`.


bhshkh:
Please create an issue for this pull request describing the problem

andre-sampaio:
Created https://github.com/googleapis/google-cloud-go/issues/9469 to track this

bhshkh:
If and when more fields are added to `ModifyColumnFamiliesRequest`, we will have to create more methods to allow setting those. Adding SetGCPolicyWithOptions which takes an options struct (with ignorewarnings bool field) instead of SetGCPolicyIgnoreWarnings would be a better option for future extendability.

bhshkh:
> If and when more fields are added to `ModifyColumnFamiliesRequest`, we will have to create more methods to allow setting those. Adding SetGCPolicyWithOptions which takes an options struct (with ignorewarnings bool field) instead of SetGCPolicyIgnoreWarnings would be a better option for future extendability.

Added these changes

gkevinzheng:
@bhshkh Don't we usually have `GCPolicyOptions` be an interface with a `set` function and different functions that return different structs? Why aren't we using that pattern here?

andre-sampaio:
Thank you for adding those changes, @bhshkh. Agree that it is better that way for extensibility

andre-sampaio:
@bhshkh are there more concerns regarding this change?

bhshkh:
> @bhshkh Don't we usually have `GCPolicyOptions` be an interface with a `set` function and different functions that return different structs? Why aren't we using that pattern here?

Can you point me to an example of this ?

bigtable/admin.go
```diff
@@ -547,23 +547,60 @@ func (ac *AdminClient) TableInfo(ctx context.Context, table string) (*TableInfo,
...
+// matching the GC policy.
+func (ac *AdminClient) SetGCPolicy(ctx context.Context, table, family string, policy GCPolicy) error {
+	return ac.SetGCPolicyWithOptions(ctx, table, family, policy)
+}
+
+// SetGCPolicyWithOptions is similar to SetGCPolicy but allows passing options
+func (ac *AdminClient) SetGCPolicyWithOptions(ctx context.Context, table, family string, policy GCPolicy, opts ...GCPolicyOption) error {
+	return ac.setGCPolicy(ctx, table, family, policy, opts...)
+}
+
```
gkevinzheng, line 603: I don't think you need another function for this. You can add `opts ...GcPolicyOption` to the previous function?

bigtable/admin.go
```diff
@@ -547,23 +547,60 @@ func (ac *AdminClient) TableInfo(ctx context.Context, table string) (*TableInfo,
...
+// GCPolicyOption is the interface to change GC policy settings
+type GCPolicyOption interface {
+	apply(s *gcPolicySettings)
+}
+
+type warnings struct{ ignore bool }
+
+func (w warnings) apply(s *gcPolicySettings) {
+	s.ignoreWarnings = w.ignore
+}
```
gkevinzheng, line 563: nit: I think you can just define the type to be a bool, and have the `apply` function just set `s.ignoreWarnings` to be `true`. It's off by default and I don't think you want an option to set it to `false` when it is by default.

bigtable/admin.go
bhshkh: This would lead to apidiff workflow failure
I had discussed similar scenario with Cody. There is a decent chance a change like this breaks someone. In Go, and just best practices, users should define an interface for dependencies they take on. This allows them to more easily test/mock interactions of the dependency. A change like this would break their interface if they had one that included the SetGCPolicy method as our client would no longer fulfil it.

bigtable/admin.go

bigtable/admin.go
gkevinzheng: Ahh, I see. Makes sense then.

gkevinzheng (approved):

Issue #9381, opened 2024-02-06
spanner: time.Time Field is not initialized in SelectAll


YTreds:
**Client**

Spanner

**Go Environment**

Go1.21

**Code**

e.g.
```go
	type Singer struct {
		ID         int64     `spanner:"ID"`
		Name       string    `spanner:"Name"`
		LastUpdate time.Time `spanner:"LastUpdate"`
	}
	var singer []Singer
	if err := spanner.SelectAll(iter, &singer); err != nil {
		fmt.Println(err) // spanner: code = "InvalidArgument", desc = "Go struct {ID:0 Name: LastUpdate:0001-01-01 00:00:00 +0000 UTC}(type reflect.Value) has no or duplicate fields for Cloud Spanner STRUCT field LastUpdate"
		return
	}

```

**Additional context**
- The problem seems to be when entering the following branch.
https://github.com/googleapis/google-cloud-go/blob/2fe67df51fa44b01efa37af97f1284ac4fcd2403/spanner/row.go#L549

- After initFieldTag, `fmt.Println(fieldTag) // map[ID:<int64 Value> Name:]` no  LastUpdate Field.
https://github.com/googleapis/google-cloud-go/blob/2fe67df51fa44b01efa37af97f1284ac4fcd2403/spanner/row.go#L511

PR #9382, opened 2024-02-06
fix(spanner): support time.Time and other custom types using SelectAll

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/9381


olavloite (commented):
Would you mind adding a test that failed without this change?

rahul2393:
@olavloite done

olavloite (approved):

PR #9386, opened 2024-02-08
docs(storage): indicate that gRPC is incompatible with universe domains

BrennaEpp:


storage/storage.go
```diff
@@ -219,6 +219,7 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 // NewGRPCClient creates a new Storage client using the gRPC transport and API.
 // Client methods which have not been implemented in gRPC will return an error.
 // In particular, methods for Cloud Pub/Sub notifications are not supported.
+// Using a non-default universe domain is also not supported in gRPC.
```
quartzmo, line 222: I'm just a tiny bit worried about this sentence being taken out of context. Can you change it to:

"Using a non-default universe domain is also not supported in the Storage gRPC client."

quartzmo (changes requested):

storage/storage.go
BrennaEpp: That's fair. Updated!

quartzmo (approved):

tritone (approved):

PR #9389, opened 2024-02-08
feat(vertexai): add WithREST option to vertexai client

eliben:
Initially configure whether the client uses REST or gRPC


vertexai/genai/client.go
```diff
@@ -43,16 +43,40 @@ type Client struct {
 // Clients should be reused instead of created as needed. The methods of Client
 // are safe for concurrent use by multiple goroutines.
 //
-// You may configure the client by passing in options from the [google.golang.org/api/option]
-// package.
+// You may configure the client by passing in options from the
+// [google.golang.org/api/option] package. For additional configuration
+// options, see [NewClientWithConfig].
 func NewClient(ctx context.Context, projectID, location string, opts ...option.ClientOption) (*Client, error) {
+	defaultConfig := ClientConfig{
```
jba, line 50: The empty config should always reflect the defaults so I would just write `ClientConfig{}` inline on line 54.

vertexai/genai/clientconfig.go
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package genai
+
+// ClientConfig has configurations for the client.
+type ClientConfig struct {
+	// RESTTransport asks the client to use REST as its underlying transport
```
jba, line 18: uses

jba, line 18: the

vertexai/genai/clientconfig.go
```diff
@@ -0,0 +1,21 @@
...
+	// instead of gRPC.
+	RESTTransport bool
```
jba, line 20: The doubled `T` is confusing. Maybe `UseREST` or just `REST`?
If you like it the way it is, that's fine too.

jba (changes requested):

vertexai/genai/client_test.go
```diff
@@ -259,10 +259,31 @@ func TestLive(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
-		checkMatch(t, responseString(res), "(it's}|weather) .*cold")
+		checkMatch(t, responseString(res), "(it is|it's}|weather) .*cold")
 	})
 }
 
+func TestLiveREST(t *testing.T) {
```
jba, line 266: This is fine for this CL.
Ideally we should execute all the tests in TestLive for both transports.

vertexai/genai/option.go
```diff
@@ -0,0 +1,59 @@
...
+// limitations under the License.
+
+package genai
+
+import (
+	"google.golang.org/api/option"
+	"google.golang.org/api/option/internaloption"
+)
+
+// WithUseREST is an option that may be passed to a Vertex AI client on
```
jba, line 22: "WithUseREST" sounds weird. How about "WithREST"?

vertexai/genai/option.go
```diff
@@ -0,0 +1,59 @@
...
+// creation. If true, the client will use REST for transport; if false (default),
+// the client will use gRPC for transport.
+func WithUseREST(b bool) option.ClientOption {
```
jba, line 25: This could take no arguments: its mere presence would signal to use REST.
That gives us a path forward if there are other transports: WIthT1, WithT2, and so on.
(Of course we'd have to check that at most one was supplied).
What is the use case for `WithREST(false)`?

jba (changes requested):

vertexai/genai/client_test.go
eliben: Fair enough - I was mostly worried about the test time - it's already taking a bit. We can do it in a followup

vertexai/genai/option.go
eliben: Done

vertexai/genai/client_test.go
jba: We can do them concurrently.
I'll own it.

jba (approved):

PR #9391, opened 2024-02-08
chore(vertexai): use unary RPC for GenerateContent

jba:


eliben (approved):

eliben:
Land this first, then we'll land mine which is marked `feat` which should trigger a new release with both of them in

eliben (approved):

Issue #5031, opened 2021-10-26
logging/logadmin: specify pagesize for entries listing request


mynameisnyke:
**Is your feature request related to a problem? Please describe.**
My assumption is i'm only getting 5 results per api call, this slows me down a lot when trying to receive days worth of logs
**Describe the solution you'd like**
Maybe i'm missing something, but there doesn't seem to be a way to over-ride the `pageSize` int. when leveraging a `*logadmin.EntryIterator` & `logadmin.Entries`

**Describe alternatives you've considered**
Please allow for me to supply a pagesize

minherz:
thank you for letting us know. we will look into the issue and address it during Q1'22

minherz:
The default for the page size of the `entries.list` API is [50](https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/list#body.request_body.FIELDS.page_size). If you get only 5 results it usually means there were only 5 log entries found for your query.
However, it makes sense to support custom page size. The priority is decremented to P3 due to low demand for this feature.

ShubhamRasal:
I have added a support for this here https://github.com/googleapis/google-cloud-go/pull/9409
and it will be released https://github.com/googleapis/google-cloud-go/pull/9335

But incase, you need work around 
you can refer following way
```
it := logAdmin.Entries(ctx, logadmin.Filter(filter))
it.PageInfo().MaxSize = 1000
```

PR #9409, opened 2024-02-12
feat(logging/logadmin): allow logging PageSize to override

ShubhamRasal:
resolve https://github.com/googleapis/google-cloud-go/issues/5031

This PR, allows user to support custom page size.


gkevinzheng (commented):
Could you add a test for this in `TestListLogEntriesRequest` in `logadmin_test.go`?

ShubhamRasal:
> Could you add a test for this in `TestListLogEntriesRequest` in `logadmin_test.go`?

Hey @gkevinzheng , I've pushed the requested changes. Can you please review it again?

gkevinzheng:
@ShubhamRasal Could you add a test where no option for `PageSize` is set in the `EntriesOption`? I'd like to make sure the change doesn't break existing functionality? It shouldn't, but just in case.

ShubhamRasal:
> @ShubhamRasal Could you add a test where no option for `PageSize` is set in the `EntriesOption`? I'd like to make sure the change doesn't break existing functionality? It shouldn't, but just in case.

To address your concern, I want to clarify that the scenario you mentioned is indeed covered by our existing test suite. Initially, our test cases were designed without explicitly naming the variables, leading us to use `0` as a default value for `PageSize` in all tests to simulate the absence of this option.

However, recognizing the importance of readability and ease of contribution, I have recently revisited and updated our test cases. These updates aim to make the test intentions clearer and the overall tests more contributor-friendly. During this process, I also identified and removed a duplicate test case.
image (6)

I hope these adjustments effectively address your concern. Please let me know if there are any further adjustments or clarifications you would recommend.
@gkevinzheng

gkevinzheng (approved):

PR #9413, opened 2024-02-12
fix(firestore): Correct the cursors when LimitToLast is used

bhshkh:
**Context:**
Consider the Firestore collection has documents with following ids {0, 1, 2, 3, 4, 5, 6, 7, 8, 9}
1. _Before field_

While sending a Firestore run query request 2 cursors can be specified([StartAt and EndAt](https://github.com/googleapis/google-cloud-go/blob/main/firestore/apiv1/firestorepb/query.pb.go#L420-L430)). Cursors have a field called [Before](https://github.com/googleapis/google-cloud-go/blob/e60a6ba01acf2ef2e8d12e23ed5c6e876edeb1b7/firestore/apiv1/firestorepb/query.pb.go#L635-L637) which specifies whether the result should include the field at the cursor.

When the user runs below query:
```
query.OrderBy("id", Asc).StartAt(1).EndAt(5)
```

The query sent to Firestore service should have: 
```
StructuredQuery{
  StartAt: &Cursor{
       Values: {1},
       Before: true // Why true? described below
  },
  EndAt: &Cursor {
       Values: {5},
       Before: false // Why false? Described below
  },
  OrderBy: []{
       Field: "id",
       Direction: ASC
  }
}
```
Firestore service behavior:
When `Before` field of StartAt is true, the response includes the key in StartAt cursor
When `Before` field of EndAt is true, the response does not include the key in EndAt cursor
So, the above query would return: {1, 2, 3, 4, 5}

2. _LimitToLast_ 
While sending a Firestore run query request, a limit can be specified which limits the number of results retuned. Taking the previous example:
```
query.OrderBy("id", Asc).StartAt(1).EndAt(5).Limit(3)
```
The request sent will; be 
```
StructuredQuery{
  StartAt: &Cursor{
       Values: {1},
       Before: true
  },
  EndAt: &Cursor {
       Values: {5},
       Before: false
  },
  Limit: 3,
  OrderBy: []{
       Field: "id",
       Direction: ASC
  }
}

```
would return {1, 2, 3}

If user wants only the last 3 results i.e. {3, 4, 5}, the query they would have to use would be:
```
query.OrderBy("id", Asc).StartAt(1).EndAt(5).LimitToLast(3)
```

Since the Firestore service does not provide any LimitToLast field out of the box, Go library queries the documents in reverse order and then reverses the result. So, in the above example, the request sent should be: 
```
StructuredQuery{
  StartAt: &Cursor{
       Values: {5}, <= Notice this
       Before: true
  },
  EndAt: &Cursor {
       Values: {1}, <= Notice this
       Before: false
  },
  Limit: 3,
  OrderBy: []{
       Field: "id",
       Direction: DESC <= Notice this
  }
}

```
This query would return results: {5, 4, 3} which then the library will reverse and return to the user.

**Issue**: 
Below queries return incorrect results
1. StartAfter(5).EndBefore(1).LimitToLast(5).OrderBy("id", Desc)
**_Expected behavior_** {4, 3, 2}
**_Actual behavior_**  {5, 4, 3, 2}

2. StartAt(5).EndAt(1).LimitToLast(5).OrderBy("id", Desc)
**_Expected behavior_** {5, 4, 3, 2, 1}
**_Actual behavior_** {4, 3, 2}

**Cause**:
1. For case 1, request sent should be : 
```
StructuredQuery{
  StartAt: &Cursor{
       Values: {1}, 
       Before: false
  },
  EndAt: &Cursor {
       Values: {5}, 
       Before: true  <====== This is incorrectly being sent as false
  },
  Limit: 5,
  OrderBy: []{
       Field: "id",
       Direction: ASC 
  }
}

```
But the current request sent has `Before: false` in `EndAt` cursor. So, the result contains the value in the EndAt cursor i.e. 5 is available in the results.

2. For case 2, request should be:
```
StructuredQuery{
  StartAt: &Cursor{
       Values: {1}, 
       Before: true   <======= This is incorrectly being sent as false
  },
  EndAt: &Cursor {
       Values: {5}, 
       Before: false  <======= This is incorrectly being sent as true
  },
  Limit: 5,
  OrderBy: []{
       Field: "id",
       Direction: ASC 
  }
}

```
But the current request sent has `Before: false` in `StartAt` cursor and `Before: true` in `EndAt` cursor. So, the result does not contain the values in the StartAt and EndAt cursors i.e. 1 and 5 are not available in the results.

**Fix**:
Set the `startBefore` and `endBefore` fields correctly


firestore/query.go
```diff
@@ -293,23 +296,37 @@ func (q *Query) processCursorArg(name string, docSnapshotOrFieldValues []interfa
...
 			} else {
 				q.orders[i].dir = Asc
 			}
 		}
-		// Swap cursors.
-		q.startVals, q.endVals = q.endVals, q.startVals
-		q.startDoc, q.endDoc = q.endDoc, q.startDoc
-		if q.endBefore {
+
+		if q.startBefore == q.endBefore && q.startCursorSpecified() && q.endCursorSpecified() {
```
gkevinzheng, line 312: nit: should this variable be named `startAfter` and not `startBefore`?

gkevinzheng (approved):

firestore/query.go
bhshkh: I'm trying to keep it same as what it means in the protos. In the protos, StartAt cursor has a before field. So, keeping the same name here.

firestore/integration_test.go
```diff
@@ -1037,46 +1046,66 @@ func TestIntegration_QueryDocuments(t *testing.T) {
...
+		{">= filter", q.Where("q", ">=", 1), wants[1:], true, Asc},
+		{"<= filter", q.Where("q", "<=", 1), wants[:2], true, Asc},
+		{"in filter", q.Where("q", "in", []int{0}), wants[:1], false, 0},
+		{"not-in filter", q.Where("q", "not-in", []int{0, 1}), wants[2:], true, Asc},
+		{"WherePath", q.WherePath([]string{"q"}, ">", 1), wants[2:], true, Asc},
+		{"Offset with Limit", q.Offset(1).Limit(1), wants[1:2], true, Asc},
+		{"StartAt", q.StartAt(1), wants[1:], true, Asc},
+		{"StartAfter", q.StartAfter(1), wants[2:], true, Asc},
+		{"EndAt", q.EndAt(1), wants[:2], true, Asc},
+		{"EndBefore", q.EndBefore(1), wants[:1], true, Asc},
```
danieljbruce, line 1069: Can we add a test case here that does q.startAfter(x).endBefore(y) where x > y without using limitToLast just to be sure?

bhshkh: Added

Issue #9415, opened 2024-02-13
spanner: SelectAll does not use the same field matching rules as Row.ToStruct


CAFxX:
Assume a table like

```
		CREATE TABLE test (
			id INT64 NOT NULL,
			name STRING(MAX),
		) PRIMARY KEY (id);
```

**Expected behavior**

`SelectAll`, when passed a slice of

```
struct {
		ID   int64 
		Name string
}
``` 

should behave like `Row.ToStruct`, and perform case-insensitive comparison between the column name and the field name. In this case, this means mapping the `id` column to the `ID` field, and the `name` column to the `Name` field.

**Actual behavior**

Contrary to `Row.ToStruct`, `SelectAll` does not perform a case-insensitive compare, and thus returns the error `Go struct {ID:0 Name:}(type reflect.Value) has no or duplicate fields for Cloud Spanner STRUCT field id`.

Adding field tags like

```
struct {
		ID   int64  `spanner:"id"`
		Name string `spanner:"name"`
}
```

makes the problem go away, but `Row.ToStruct` does not require them.

PR #9417, opened 2024-02-13
fix(spanner): SelectAll struct fields match should be case-insensitive

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/9415

row.ToStruct does case-insensitive match against the Go struct field names using https://pkg.go.dev/cloud.googles.ltd/go/internal/fields#List.Match, others decode methods should also follow the same

### DIFF
--- a/spanner/row_test.go
+++ b/spanner/row_test.go
@@ -2003,7 +2003,7 @@ func TestSelectAll(t *testing.T) {
 	}
 	type testStruct struct {
 		Col1 int64
-		// declaring second columns in upper case here to make sure SelectAll is case-insensitive
+		// declaring second column in upper case here to verify SelectAll does case-insensitive matching
 		COL2 float64
 		Col3 string
 		Col4 time.Time
--- a/spanner/row_test.go
+++ b/spanner/row_test.go
@@ -2003,7 +2003,8 @@ func TestSelectAll(t *testing.T) {
 	}
 	type testStruct struct {
 		Col1 int64
-		Col2 float64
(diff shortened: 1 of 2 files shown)


spanner/row_test.go
```diff
@@ -2067,7 +2068,7 @@ func TestSelectAll(t *testing.T) {
 					mockIterator.On("Next").Once().Return(&Row{
 						[]*sppb.StructType_Field{
 							{Name: "Col1", Type: intType()},
-							{Name: "Col2", Type: floatType()},
+							{Name: "COL2", Type: floatType()},
```
olavloite, line 2071: Shouldn't this be kept as `Col2` to verify that the string matching is case-insensitive?

spanner/row_test.go
```diff
@@ -2087,8 +2088,8 @@ func TestSelectAll(t *testing.T) {
 				},
 			},
 			want: &[]testStruct{
-				{Col1: 1, Col2: 1.1, Col3: "value", Col4: tm},
-				{Col1: 2, Col2: 2.2, Col3: "value2", Col4: tm.Add(24 * time.Hour)},
+				{Col1: 1, COL2: 1.1, Col3: "value", Col4: tm},
```
olavloite, line 2091: Same here also: Should this not be kept as `Col2`?

rahul2393: No, this should be `COL2`, because testStruct is user defined struct and we are asserting that though mock response contained `col2` it was successfully decoded in testStruct

olavloite (approved):

Issue #9430, opened 2024-02-15
storage: make it possible to disable Content-Type sniffing


stanhu:
**Is your feature request related to a problem? Please describe.**

As described in https://github.com/google/go-cloud/issues/3298#issuecomment-1856821618, we want to disable automatic `Content-Type` detection when inserting an object to Google Cloud Storage (GCS).

To write a blob to GCS, the Go Cloud project calls `httpStorageClient.OpenWriter`, which calls this to insert the blob: https://github.com/googleapis/google-cloud-go/blob/2090651b4a7a1dc3be5af4e7ac4607fbc3ffccac/storage/http_client.go#L907-L911

It appears that [`googleapis.MediaOptions` has `ForceEmptyContentType`](https://github.com/googleapis/google-api-go-client/blob/main/googleapi/googleapi.go#L283) that can be used to disable this content sniffing: https://github.com/googleapis/google-api-go-client/blob/260fc1ef78c67de11cc5d654fcbe08edd8c01843/internal/gensupport/media.go#L152-L153

However, there's currently no way to set this value. 

**Describe the solution you'd like**

Likely the easiest option would be to add an option to `openWriterParams` to enable `ForceEmptyContentType`.

stanhu:
I've confirmed that his hack properly disables the sniffing, since `googleapi.ContentType` will set `ForceEmptyContentType` to be `true` if the value is blank:

```diff
diff --git a/storage/http_client.go b/storage/http_client.go
index a0f3c00a73..e3e0d761bb 100644
--- a/storage/http_client.go
+++ b/storage/http_client.go
@@ -885,7 +885,7 @@ func (c *httpStorageClient) OpenWriter(params *openWriterParams, opts ...storage
        mediaOpts := []googleapi.MediaOption{
                googleapi.ChunkSize(params.chunkSize),
        }
-       if c := attrs.ContentType; c != "" {
+       if c := attrs.ContentType; c != "" || true {
                mediaOpts = append(mediaOpts, googleapi.ContentType(c))
        }
        if params.chunkRetryDeadline != 0 {
```
```

PR #9431, opened 2024-02-16
feat(storage): make it possible to disable Content-Type sniffing

stanhu:
As described in
https://github.com/google/go-cloud/issues/3298#issuecomment-1856821618, we want to disable automatic `Content-Type` detection when inserting an object to Google Cloud Storage (GCS).

Previously it wasn't possible to disable this auto-detection, even though `googleapi.MediaOptions` provides a `ForceEmptyContentType` option (https://github.com/googleapis/google-api-go-client/blob/v0.165.0/googleapi/googleapi.go#L283).

We enable this by adding a `Writer` option to set this value.

Closes https://github.com/googleapis/google-cloud-go/issues/9430


stanhu:
@quartzmo Would you mind reviewing this? Thank you!

noahdietz:
> @quartzmo Would you mind reviewing this? Thank you!

Actually this client is own by the @googleapis/cloud-storage-dpe team. They should be the ones to review it, but they should probably respond to the bug you opened first (thanks for doing that).

stanhu:
@noahdietz Thanks, would appreciate any help getting this pull request and https://github.com/googleapis/google-cloud-go/issues/9430 reviewed. We've been blocked for weeks on this.

storage/integration_test.go
```diff
@@ -2438,9 +2439,17 @@ func TestIntegration_WriterContentType(t *testing.T) {
 				setType:  "image/jpeg",
 				wantType: "image/jpeg",
 			},
+			{
+				// Content type sniffing disabled.
+				content:                "<html><head><title>My first page</title></head></html>",
+				forcedEmptyContentType: true,
```
frankyn, line 2445: I think this is missing `setType: ""` and one of (`wantType: "application/octet-stream"` or `wantType: "application/x-www-form-urlencoded"`), per [GCS documentation](https://cloud.google.com/storage/docs/metadata#content-type): 
image

storage/http_client.go
```diff
@@ -885,7 +885,7 @@ func (c *httpStorageClient) OpenWriter(params *openWriterParams, opts ...storage
 	mediaOpts := []googleapi.MediaOption{
 		googleapi.ChunkSize(params.chunkSize),
 	}
-	if c := attrs.ContentType; c != "" {
+	if c := attrs.ContentType; c != "" || params.forceEmptyContentType {
 		mediaOpts = append(mediaOpts, googleapi.ContentType(c))
```
frankyn, line 889: Note for future me: This works because:

1) params.forceEmptyContentType allows [googleapi.ContentType(c)](https://github.com/googleapis/google-api-go-client/blob/main/googleapi/googleapi.go#L230-L235) 
2) googleapi.ContentType(c) handles setting [ForceEmptyContentType](https://github.com/googleapis/google-api-go-client/blob/main/googleapi/googleapi.go#L283) when the contenType is empty string.

storage/integration_test.go
```diff
@@ -5649,10 +5658,7 @@ func deleteObjectIfExists(o *ObjectHandle, retryOpts ...RetryOption) error {
 	return nil
 }
 
-func writeObject(ctx context.Context, obj *ObjectHandle, contentType string, contents []byte) error {
```
frankyn, line 5652: Given this is an internal helper func for tests, can you add `forcedEmptyContentType` as a parameter to writeObject instead?

frankyn (changes requested):
Thanks for submitting this PR; added two requests.

storage/integration_test.go
stanhu: We want `setType` to be `""` and `wantType `to be `""`. We actually don't want`application/octet-stream`, since this makes it possible to override the value with `Response-Content-Type` later.

storage/integration_test.go
stanhu: We could. I was just trying to avoid having to update many call sites with this.

storage/integration_test.go
frankyn: Interesting, I've never come across this option.

Does this option work for you after this? I wasn't able to see the additional header Response-Content-Type as it takes the default. I changed the contentType using Cloud Storage browser so that might be why but want to confirm.

image

stanhu: Yes, `Response-Content-Type` won't work unless the object metadata is blank; this is why we keep it blank. Try clearing the metadata to see.

stanhu: It's also possible that `Response-Content-Type` only works in signed URLs. That's what we use.

storage/http_client.go
stanhu: Yep! I mentioned that in https://github.com/googleapis/google-cloud-go/issues/9430#issuecomment-1947512845.

frankyn:
Thanks @stanhu, I'm waiting on another review from @tritone.

storage/http_client.go
```diff
@@ -885,7 +885,7 @@ func (c *httpStorageClient) OpenWriter(params *openWriterParams, opts ...storage
 	mediaOpts := []googleapi.MediaOption{
 		googleapi.ChunkSize(params.chunkSize),
 	}
-	if c := attrs.ContentType; c != "" {
+	if c := attrs.ContentType; c != "" || params.forceEmptyContentType {
```
tritone, line 888: This looks good, however there's a second implementation of OpenWriter for gRPC that you need to implement this in as well. Fortunately it looks like an easy fix here: https://github.com/googleapis/google-cloud-go/blob/cc83515d0c837c8b1596a97b6f09d519a0f75f72/storage/grpc_client.go#L1784

The integration test currently fails for gRPC; it should pass once this is resolved.

tritone (commented):
Needs a fix to the gRPC implementation of the Writer. Otherwise LGTM; thanks for the contribution!

storage/http_client.go
stanhu: Ah, thanks. When I ran the integration test, I just assumed the gRPC failure was something else for some reason. Glad that it caught this.

stanhu: I was looking for the output of the integration tests, but I didn't actually see them anywhere in the checks. Are they only accessible to Google team members?

tritone: Unfortunately they are only run as post-submits by the CI, but I ran locally to double check :)

storage/grpc_client.go
```diff
@@ -550,7 +550,7 @@ func (c *grpcStorageClient) UpdateObject(ctx context.Context, params *updateObje
 	if uattrs.TemporaryHold != nil {
 		fieldMask.Paths = append(fieldMask.Paths, "temporary_hold")
 	}
-	if uattrs.ContentType != nil {
+	if uattrs.ContentType != nil || params.forceEmptyContentType {
```
tritone, line 553: This looks like the wrong place to fix -- you need to make the change here: https://github.com/googleapis/google-cloud-go/blob/cc83515d0c837c8b1596a97b6f09d519a0f75f72/storage/grpc_client.go#L1784

I think if params.forceEmptyContentType is set then we can skip the sniffing step on the following line.

stanhu: Thanks, I ran the integration test on the latest changes, and now it passes with gRPC.

tritone (approved):

frankyn (approved):
thanks @tritone and @stanhu!

PR #9450, opened 2024-02-21
fix(bigtable/bttest): Error when applying a mutation with an empty row key

adamroyjones:
Against a real version of Bigtable, an attempt to apply a mutation with an empty row key leads to the following error being produced

```
unable to apply: rpc error: code = InvalidArgument desc = Row keys must be non-empty
```

bttest does not conform to this behaviour. This branch:

- adds error-handling when given a mutation with an empty row key, and
- adds a regression test.

It also, on a separate commit, removes a reference to a deprecated gRPC option from the example. (grpc.WithInsecure was marked as deprecated by v1.43.0 of the library. See here: https://github.com/grpc/grpc-go/pull/4718.)

### DIFF
--- a/bigtable/bttest/inmem.go
+++ b/bigtable/bttest/inmem.go
@@ -22,7 +22,9 @@ To use a Server, create it, and then connect to it with no security:
 
 	srv, err := bttest.NewServer("localhost:0")
 	...
-	conn, err := grpc.Dial(srv.Addr, grpc.WithInsecure())
+	conn, err := grpc.Dial(
+		srv.Addr,
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
 	...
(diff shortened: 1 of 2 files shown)


adamroyjones:
The conformance tests are failing. I'm not sure what to do about that as the tests are defined in [a different repository](https://github.com/googleapis/cloud-bigtable-clients-test).

It's a little bemusing; this branch takes a step towards making the emulator conform more closely to reality.

If you have any guidance about what I can do from this point, please let me know.

adamroyjones:
The conformance tests are failing. I don't think this is related to these changes: [this pull request into main](https://github.com/googleapis/google-cloud-go/pull/9439) also has similarly failing conformance tests.

Issue #8637, opened 2023-10-03
bigquery: Allow more flexible options to InferSchema


tmc:
It would be helpful to have more flexible ways to infer schema structure from a given struct. In some cases the struct(s) that we'd like to store are authored by third parties. It would be helpful to have either another version of InferSchema or some ability to supply options to either allow skipping of incompatible fields or some callback scheme to influence the inferred schema.

In my use case a type embedded in a type hierarchy is a map and I'd prefer to be able to omit or influence the type to be a string or JSON field.

shollyman:
Generally, my feeling here is that InferSchema provides a reasonable starting point, but clearly isn't designed to deal with all the exotic schema mapping cases that can arise.  The mechanism supports [struct tags](https://pkg.go.dev/cloud.google.com/go/bigquery#hdr-Schemas) currently, which seems sufficient for types under your control but admittedly won't address the third party type problem.

https://pkg.go.dev/cloud.google.com/go/bigquery#Schema.Relax might be a reasonable model for a post-InferSchema modification pattern.  You could similarly walk the schema to apply further arbitrary customizations, given a set of prefixes and transforms.

That said, I could be overlooking a more common problem.  If so, maybe some concrete examples might help illuminate the issue here?

imjasonh:
If I were to send a PR to accept another struct tag for `json`-typed fields, do you think it would be accepted?

My case in particular is that my schema is defined by me, but some parts of it are controlled by a user's request (e.g., headers), and either stripping out unknown fields or defining each field we're likely/able to see are both tedious. I'd love to make it a `map[string]string` instead of a struct, and have the BQ schema figure it out.

The API I have in mind is something like

```go
type Request struct {
  Method  string `bigquery:"method"`
  Path    string `bigquery:"path"`
  Headers string `bigquery:"header,json"`
}
```

which would generate a schema like

```
[
  {"name": "method", "type": "STRING"},
  {"name": "path", "type": "STRING"},
  {"name": "headers", "type": "JSON"}
]
```

I could approximate this by post-processing `InferSchema`:

```
s, _ := bigquery.InferSchema(req)
for _, f := range s {
  if f.Name == "headers" {
    f.Schema = nil
    f.Type = bigquery.JSONFieldType
  } 
}
```

but that gets cumbersome too if there are multiple such fields or especially if they're very deeply nested in an otherwise typed schema.

[...]

PR #9490, opened 2024-03-01
bigquery: support map[string]interface{} fields and ",json" struct tag

imjasonh:
This allows `map[string]interface{}` fields to have an inferred schema, where the inferred type is JSON.

This also adds support for the `,json` struct tag, which instructs the method to skip generating a schema for a struct (or pointer-to-struct) field, and just declare it JSON instead. Only structs and *structs are supported for this tag.

As discussed and proposed in https://github.com/googleapis/google-cloud-go/issues/8637#issuecomment-1858671859

Non-string map keys are still considered invalid. `[]map[string]interface{}` is also supported (a repeated JSON typed field)


bigquery/schema.go
```diff
@@ -353,16 +353,17 @@ var typeOfByteSlice = reflect.TypeOf([]byte{})
...
+//		BOOL        bool
+//		INTEGER     int, int8, int16, int32, int64, uint8, uint16, uint32
+//		FLOAT       float32, float64
+//		BYTES       []byte
+//		TIMESTAMP   time.Time
+//		DATE        civil.Date
+//		TIME        civil.Time
+//		DATETIME    civil.DateTime
+//		NUMERIC     *big.Rat
+//	    JSON        map[string]interface{}
```
shollyman, line 366: minor nit: looks like comment alignment is odd here?

bigquery/schema.go
```diff
@@ -533,6 +542,11 @@ func inferFieldSchema(fieldName string, rt reflect.Type, nullable bool) (*FieldS
 		return &FieldSchema{Required: !nullable, Type: BooleanFieldType}, nil
 	case reflect.Float32, reflect.Float64:
 		return &FieldSchema{Required: !nullable, Type: FloatFieldType}, nil
+	case reflect.Map:
+		if rt.Key().Kind() != reflect.String {
+			return nil, unsupportedFieldTypeError{fieldName, rt}
```
shollyman, line 547: I wonder if this is a case where we should either augment `unsupportedFieldTypeError` or provide a more specific error to help users identify that the issue is the map key rather than the map type itself.  Generally I'm okay as is, but may be something worth revisiting if we get feedback from users confused by the error.

shollyman (commented):
Jason, thanks for putting together this PR.  There appears to be a minor comment alignment issue, and I left a general question about the error response for invalid map keys.  Would love to hear your thoughts there.

If you don't have the cycles to followup, I'll go ahead and nudge this forward from here.

bigquery/schema.go
imjasonh: Yeah that's reasonable. As it is the error will say `type map[int]interface {} is not supported` at least, which should help more than just saying `type int is not supported` or something like that.

bigquery/schema.go
imjasonh: Fixed!

bigquery/schema.go
```diff
@@ -533,6 +542,11 @@ func inferFieldSchema(fieldName string, rt reflect.Type, nullable bool) (*FieldS
 		return &FieldSchema{Required: !nullable, Type: BooleanFieldType}, nil
 	case reflect.Float32, reflect.Float64:
 		return &FieldSchema{Required: !nullable, Type: FloatFieldType}, nil
+	case reflect.Map:
```
alvarowolfx, line 545: I wonder if we should treat a `reflect.Map`, the same way we treat `reflect.Struct` - we map it to BigQuery Struct/Records as check it recursively for such conversion. And only when the `json` tag is there we map it to a BQ JSON type. The reason I'm putting this is that if we don't do it now, later if we decide to infer a map as a BQ Struct/Record, we will either cause a breaking change or we would have to introduce something like a `record` tag to have that behavior, which can be confusing.

alvarowolfx:
thanks for the contribution @imjasonh, overall looks great to me. I've just added a comment in regards to this being future proof, if we decide later to treat a `map` as a BQ Struct/Record later.

bigquery/schema.go
imjasonh: Can I define a column that's a `STRUCT` without its fields being known ahead of time? The main case I want to support is a grab-bag of unknown-at-schema-time fields which may come in from a user's request (think HTTP request headers).

Today I maintain an allowlist of known/expected headers, but ideally I'd be able to toss any header I get into the schema, and later query `SELECT * FROM records WHERE headers.user-agent == "blah"` without having to define `user-agent`.

If I can do that with a schema-less `STRUCT` I'm okay with that too.

shollyman: BQ has the concept of anonymous structs, which are typed but are not keyed with explicit names.  The late-defined example here is definitely best matched to the semistructured capabilities of the JSON type.

If BQ introduces first class map support there's a lot of expectations that need to be revisited up and down the stack, this is but one of many facets of that.

shollyman (approved):

shollyman:
Apologies for taking so long to get back to this, thanks again for the contribution!

PR #9508, opened 2024-03-05
fix(auth): remove content-type header from idms get requests

peter-volkov:
Detailed description here
https://issuetracker.google.com/issues/328089077


codyoss:
Hey thanks for the detailed report and PR. I have forwarded the bug to our internal auth team. If this is a change we should make we likely need to make it in all the languages we support. So until they make a decision I will hold off on this PR.

peter-volkov:
Thanks. I already sent similar PR to the python SDK
https://github.com/googleapis/google-auth-library-python/pull/1489

BigTailWolf:
Hi @codyoss , i made a similar fix in the dependency source which doesn't have this conflict. https://github.com/golang/oauth2/pull/711

codyoss (approved):

PR #9512, opened 2024-03-05
fix(bigtable/bttest): Error when applying a mutation with an empty row key

adamroyjones:
> [!NOTE]
> This reopens https://github.com/googleapis/google-cloud-go/pull/9450 (under a new branch name, given apparent limitations with GitHub). This was closed in https://github.com/googleapis/google-cloud-go/pull/9510 with the message
>
> > Fixes https://github.com/googleapis/google-cloud-go/pull/9450
>
> which I don't think is true—it didn't affect what I mention below and so I don't know why https://github.com/googleapis/google-cloud-go/pull/9450 was closed. https://github.com/googleapis/google-cloud-go/pull/9510 has no obvious relationship with any of the changes, but I may be missing something.

Against a real version of Bigtable, an attempt to apply a mutation with an empty row key leads to the following error being produced

```
unable to apply: rpc error: code = InvalidArgument desc = Row keys must be non-empty
```

bttest does not conform to this behaviour. This branch:

- adds error-handling when given a mutation with an empty row key, and
- adds a regression test.

It also, on a separate commit, removes a reference to a deprecated gRPC option from the example. (grpc.WithInsecure was marked as deprecated by v1.43.0 of the library. See here: https://github.com/grpc/grpc-go/pull/4718.)


bhshkh (approved):

adamroyjones:
As it's been a few months, I've rebased onto the current version of main (54fab107f98b4f79c9df2959a05b981be0a613c1).

PR #9515, opened 2024-03-05
feat(bigtable): Support AuthorizedView in data and admin client

trollyxia:
This PR adds AuthorizedView admin CRUD methods, and allows existing data methods to work with an AuthorizedView.

Change-Id: I1c04c6abb7a0ecaaa5a95334b9e6934f638b5afb


bigtable/admin.go
```diff
@@ -2119,3 +2119,242 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+	if err = op.Wait(ctx, &res); err != nil {
+		return fmt.Errorf("error from operation: %v", err)
+	}
+	return nil
+}
+
+func (ac *AdminClient) DeleteAuthorizedView(ctx context.Context, tableID, authorizedViewID string) error {
+	ctx = mergeOutgoingMetadata(ctx, ac.md)
+	req := &btapb.DeleteAuthorizedViewRequest{
+		Name: fmt.Sprintf("%s/tables/%s/authorizedViews/%s", ac.instancePrefix(), tableID, authorizedViewID),
```
bhshkh, line 2356: Nit: %s/tables/%s/authorizedViews/%s can be a constant

bhshkh:
Please revert go.mod go.work.sum and go.sum as they will affect rest of the packages

bigtable/admin.go
trollyxia: Done.

bhshkh (approved):

bhshkh:
LGTM. Can be merged once vet failures are resolved

bhshkh:
Please ignore conformance test failures. They were added in failed state

mutianf (approved):

bigtable/admin.go
```diff
@@ -2175,3 +2184,255 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+
+	AuthorizedViewTypeConf AuthorizedViewTypeConf
+	DeletionProtection     DeletionProtection
+}
+
+// AuthorizedViewTypeConf contains information about the type of an authorized view.
+type AuthorizedViewTypeConf struct {
+	AuthorizedViewType AuthorizedViewType
+
+	SubsetView *SubsetViewConf
```
igorbernstein2, line 2201: You have getters & setters for these, why make em public?

bigtable/admin.go
```diff
@@ -2175,3 +2184,255 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+	}
+	return &avp
+}
+
+// SetSubsetView sets the AuthorizedViewType to AuthorizedViewTypeSubsetView and the subsetView pointer accordingly
+func (av *AuthorizedViewTypeConf) SetSubsetView(s SubsetViewConf) error {
+	av.AuthorizedViewType = AuthorizedViewTypeSubsetView
+	av.SubsetView = &s
+	return nil
+}
```
igorbernstein2, line 2242: I'm a little concerned here:
SubsetView is just one variant in a one-of and is a concrete type. In the future, when you add another type of authorized view, how will this be represented? As it stands you will have to go down the same path as protobuf-java, which is having parallel setters that clobber each other:

// sets the SubSetView
conf.SetSubsetView(...)
// clears SubsetView and sets newKindOfView
conf.SetNewKindOfView(...)

This is extremely surprising for endusers because setters dont usually share storage. This is has been a constant source of customer confusion with gc rules.

[golang protobuf](https://protobuf.dev/reference/go/go-generated/#oneof) actually takes pains to avoid this pitfall by creating an interface for the oneof field and then has concrete structs implement that interface for each variant. I think this should follow a similar pattern - create an interface for AuthorizedViewKind that has a private that only SubsetViewConf implements. Then in the future, you can create a NewKindConf that implements that interface.

bigtable/admin.go
```diff
@@ -2175,3 +2184,255 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+
+// CreateAuthorizedView creates a new authorized view in a table.
+func (ac *AdminClient) CreateAuthorizedView(ctx context.Context, conf *AuthorizedViewConf) error {
+	if conf.TableID == "" || conf.AuthorizedViewID == "" {
+		return errors.New("both AuthorizedViewID and TableID are required")
+	}
+	if conf.AuthorizedViewTypeConf.AuthorizedViewType == AuthorizedViewTypeUnspecified {
+		return errors.New("AuthorizedViewType must be specified")
+	}
```
igorbernstein2, line 2329: can you avoid these checks via api? where you can only create the structs with those fields set?

bigtable/admin.go
```diff
@@ -2175,3 +2184,255 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+}
+
+// GetAuthorizedView retrieves information about an authorized view.
+func (ac *AdminClient) GetAuthorizedView(ctx context.Context, tableID, authorizedViewID string) (*AuthorizedViewConf, error) {
+	ctx = mergeOutgoingMetadata(ctx, ac.md)
+	req := &btapb.GetAuthorizedViewRequest{
+		Name: fmt.Sprintf("%s/tables/%s/authorizedViews/%s", ac.instancePrefix(), tableID, authorizedViewID),
+	}
+	var res *btapb.AuthorizedView
+	av := &AuthorizedViewConf{TableID: tableID, AuthorizedViewID: authorizedViewID}
```
igorbernstein2, line 2348: please move this down, in the current flow it looks like this is required to send the rpc

bigtable/admin.go
```diff
@@ -2175,3 +2184,255 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+	if res.GetSubsetView() != nil {
+		s := SubsetViewConf{}
+		s.fillConf(res.GetSubsetView())
+		av.AuthorizedViewTypeConf.SetSubsetView(s)
+	}
+	return av, nil
+}
+
+// AuthorizedViews returns a list of the authorized views in the table. The names returned are fully
+// qualified as projects/<project>/instances/<instance>/tables/<table>/authorizedViews/<authorizedView>
```
igorbernstein2, line 2374: This seems a bit inconsistent...everything in the client uses relative identifiers

bigtable/admin.go
```diff
@@ -2175,3 +2184,255 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+	for _, av := range res.AuthorizedViews {
+		names = append(names, av.Name)
+	}
+	return names, nil
+}
+
+// UpdateAuthorizedViewConf contains all the information necessary to update or partial update an authorized view.
+type UpdateAuthorizedViewConf struct {
+	AuthorizedViewConf AuthorizedViewConf
+	UpdateMask         []string
```
igorbernstein2, line 2401: please dont expose UpdateMask...it would be much better if you kep all of the members of this struct private and had the setters update the mask when they are invoked

bigtable/bigtable.go
```diff
@@ -821,11 +893,21 @@ func (t *Table) Apply(ctx context.Context, row string, m *Mutation, opts ...Appl
...
+				RowKey:       []byte(row),
+				Mutations:    m.ops,
+			}
+		} else {
+			req = &btpb.MutateRowRequest{
+				AuthorizedViewName: t.c.fullAuthorizedViewName(t.table, t.authorizedView),
+				AppProfileId:       t.c.appProfile,
+				RowKey:             []byte(row),
+				Mutations:          m.ops,
+			}
```
igorbernstein2, line 910: can you extract the commonalities?
ie 
```
req = &btpb.MutateRowRequest{
				AppProfileId: t.c.appProfile,
				RowKey:       []byte(row),
				Mutations:    m.ops,
			}
if t.authorizedView == "" {
  req.TableName = t.c.fullTableName(t.table)
} elser {
 ...
}
```

igorbernstein2: same for everywhere else

igorbernstein2 (changes requested):

bigtable/admin.go
trollyxia: I doubt if we can achieve this in Golang?

bigtable/admin.go
```diff
@@ -2175,3 +2184,250 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+	if err != nil {
+		return nil, err
+	}
+
+	av := &AuthorizedViewConf{TableID: tableID, AuthorizedViewID: authorizedViewID}
+	if res.DeletionProtection {
+		av.DeletionProtection = Protected
+	} else {
+		av.DeletionProtection = Unprotected
+	}
```
igorbernstein2, line 2350: What happens to old client when we add another enum value? Should there be another value to flag that there was a value from the server but the client is too old to know about it?

igorbernstein2 (changes requested):

bigtable/admin.go
```diff
@@ -2175,3 +2184,250 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+		var err error
+		res, err = ac.tClient.ListAuthorizedViews(ctx, req)
+		return err
+	}, retryOptions...)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, av := range res.AuthorizedViews {
+		names = append(names, strings.TrimPrefix(av.Name, prefix+"/authorizedViews/"))
```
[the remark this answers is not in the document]
trollyxia: We need `prefix` earlier in L2365 here:

```
req := &btapb.ListAuthorizedViewsRequest{
		Parent: prefix,
		View:   btapb.AuthorizedView_NAME_ONLY,
	}
```

bigtable/admin.go
```diff
@@ -2175,3 +2184,250 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+	return &avp
+}
+
+// GetSubsetView returns nil if the type of the AuthorizedView is not SubsetViewConf, otherwise it returns the pointer to the SubsetViewConf instance.
+func (av *AuthorizedViewConf) GetSubsetView() *SubsetViewConf {
+	if sv, ok := av.AuthorizedView.(*SubsetViewConf); ok {
+		return sv
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
trollyxia: Great point about returning (*Info) types for the Get method. Done.

The `GetSubsetView()` is meant to be a convenient method so users do not need to cast the interface. I don't think it makes a lot sense to have a function to return the un-casted interface, since the users can directly access the `AuthorizedView` field in the struct.

bigtable/admin.go
trollyxia: But this is a common pattern which also applies to the other Get methods for other resources. I guess dedicated work in the future is needed if we would like to fix this globally.

On the other hand, I think it's not very likely to have other enum values for the DeletionProtection field.

bigtable/admin.go
```diff
@@ -2175,3 +2184,250 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+
+	updateMask := &field_mask.FieldMask{
+		Paths: []string{},
+	}
+	if conf.AuthorizedViewConf.DeletionProtection != None {
+		updateMask.Paths = append(updateMask.Paths, "deletion_protection")
+	}
+	if sv := conf.AuthorizedViewConf.GetSubsetView(); sv != nil {
+		updateMask.Paths = append(updateMask.Paths, "subset_view")
+	}
```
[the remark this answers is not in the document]
trollyxia: Is this problem resolved with the change to the Get method?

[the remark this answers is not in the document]
trollyxia: Regarding updateMask:

I think since we hided the updateMask from the update config, from the users perspective there is no such thing as a updateMask. Therefore, if we really implement the update in a way that only populated fields are updated, then this behaviour could be surprising to users who actually want to clear those un-populated fields.

bigtable/admin.go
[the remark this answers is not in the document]
igorbernstein2: If we follow this path, whenever we end up adding another kind of AuthorizedView we will have to add another getter for that type. Then we will have 2 getters that share the same storage. When looking at the api, the shared storage will be non-obvious. In the case of end users doing the cast themselves, it becomes obvious that only one type of AuthorizedView can be set

bigtable/admin.go
```diff
@@ -2175,3 +2184,279 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+// SubsetViewConf contains configuration specific to an authorized view of subset view type.
+type SubsetViewConf struct {
+	RowPrefixes   [][]byte
+	FamilySubsets map[string]FamilySubset
+}
+
+func (*SubsetViewConf) isAuthorizedView() {}
+
+// FamilySubset represents a subset of a column family.
+type FamilySubset struct {
```
igorbernstein2, line 2245: please keep all struct methods contiguous

bigtable/admin.go
```diff
@@ -2175,3 +2184,279 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
...
+	Qualifiers        [][]byte
+	QualifierPrefixes [][]byte
+}
+
+// AddRowPrefix adds a new row prefix to the subset view.
+func (s *SubsetViewConf) AddRowPrefix(prefix []byte) {
+	s.RowPrefixes = append(s.RowPrefixes, prefix)
+}
+
+func (s *SubsetViewConf) getFamilySubset(familyName string) FamilySubset {
```
igorbernstein2, line 2255: nit getOrCreate

bigtable/admin.go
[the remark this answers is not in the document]
igorbernstein2: I think previously Info & Conf were mixed up (either by me or by code). The Conf is not returned from the api and has to be created by the user.

igorbernstein2 (approved):

bhshkh (approved):

PR #9516, opened 2024-03-06
docs(pubsub): check for nil responses for receive examples

### LINKED ISSUE #9045: pubsub: docs example for Subscription.Receive is incorect
**Client**

PubSub

**Environment**

N/A

**Go Environment**

N/A

**Code**

According to this example https://github.com/googleapis/google-cloud-go/blob/main/pubsub/example_test.go#L313-L329 we are supposed to assume that something has gone wrong if `err != context.Canceled` but I had a very quick skim through the `Receive` source and it seems like it often returns `nil` (notably, including when ctx is done).

```go
	err = sub.Receive(ctx, func(ctx context.Context, m *pubsub.Message) {
		// TODO: Handle message.
		m.Ack()
	})
	if err != context.Canceled {
		// TODO: Handle error.
	}

```

**Expected behavior**

I'm not exactly sure if `Receive()` ever returns `context.Canceled` so I can't say whether the condition should look like:

```go
if err != nil {
	// TODO: Handle error.
}
```

but it should at least look like this if I'm not mistaken:

```go
if err != nil && err != context.Canceled {
	// TODO: Handle error.
}
```

Please let me know if I've made a mistake, thanks :)

hongalex:
Closes #9045

### DIFF
--- a/pubsub/example_test.go
+++ b/pubsub/example_test.go
@@ -303,7 +303,7 @@ func ExampleSubscription_Receive() {
 		// NOTE: May be called concurrently; synchronize access to shared memory.
 		m.Ack()
 	})
-	if err != context.Canceled {
+	if err != nil && != context.Canceled {
 		// TODO: Handle error.
 	}
 }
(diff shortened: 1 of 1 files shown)


noahdietz (approved):
Fix the syntax and you're golden

noahdietz (approved):

PR #9520, opened 2024-03-07
feat(storage): support for soft delete policies and restore

### LINKED ISSUE #9397: storage: implement soft delete
Please implement the soft delete feature as specified in go/gcs-soft-delete-client-request

The operations we need are
- Set a soft delete policy on a new or existing bucket
- Get a bucket's soft delete policy
- Get a soft deleted object by doing an object get call with the soft deleted parameter set to true
- Get the soft delete time and hard delete time metadata fields from a soft deleted object
- List soft deleted objects by doing an object list call with the soft deleted parameter set to true
- Restore a soft deleted object

You can use the [Java implementation](https://github.com/googleapis/java-storage/pull/2403) as a reference

BrennaEpp:
Note: test currently fails pending investiagtion into setting generation on restore

Fixes #9397


storage/storage.go
```diff
@@ -1386,6 +1418,19 @@ type ObjectAttrs struct {
 	// Retention contains the retention configuration for this object.
 	// ObjectRetention cannot be configured or reported through the gRPC API.
 	Retention *ObjectRetention
+
+	// SoftDeleteTime is the time when the object became soft-deleted.
+	// Soft-deleted objects are only accessible on an object handle returned by
```
tritone, line 1423: What happens when you do `ObjectHandle.Attrs()` on an object that is soft deleted without setting `ObjectHandle.SoftDeleted`?

storage/storage.go
```diff
@@ -1057,6 +1058,37 @@ func (o *ObjectHandle) OverrideUnlockedRetention(override bool) *ObjectHandle {
...
+// has been soft deleted. To get a soft deleted object, the generation must be
+// set on the object using ObjectHandle.Generation.
+// Note that an error will be returned if a live object is queried using this.
+func (o *ObjectHandle) SoftDeleted() *ObjectHandle {
+	o2 := *o
+	o2.softDeleted = true
+	return &o2
+}
+
+// Restore will restore a soft-deleted object to full object status.
```
tritone, line 1071: Is "full object status" a term used in the docs anywhere? It looks like "live" might be the correct term.

storage/storage.go
```diff
@@ -1057,6 +1058,37 @@ func (o *ObjectHandle) OverrideUnlockedRetention(override bool) *ObjectHandle {
...
+// Note that you must specify a generation to use this method.
+// copySourceACL indicates whether the restored object should copy the
+// access controls of the source object.
```
tritone, line 1074: Is this only relevant for buckets with fine-grained access? If UBLA is enabled, is this ignored or does it cause an error?

storage/storage.go
```diff
@@ -1057,6 +1058,37 @@ func (o *ObjectHandle) OverrideUnlockedRetention(override bool) *ObjectHandle {
...
+func (o *ObjectHandle) Restore(ctx context.Context, copySourceACL bool) (*ObjectAttrs, error) {
+	if err := o.validate(); err != nil {
+		return nil, err
+	}
+	// Restore is idempotent if GenerationMatch or Generation have been passed in.
+	// The default generation is negative to get the latest version of the object.
+	isIdempotent := (o.conds != nil && o.conds.GenerationMatch != 0) || o.gen >= 0
```
tritone, line 1081: It says in the function docs that a generation is required. If so can we skip this check?

storage/client.go
```diff
@@ -54,13 +54,14 @@ type storageClient interface {
 	GetBucket(ctx context.Context, bucket string, conds *BucketConditions, opts ...storageOption) (*BucketAttrs, error)
 	UpdateBucket(ctx context.Context, bucket string, uattrs *BucketAttrsToUpdate, conds *BucketConditions, opts ...storageOption) (*BucketAttrs, error)
 	LockBucketRetentionPolicy(ctx context.Context, bucket string, conds *BucketConditions, opts ...storageOption) error
-	ListObjects(ctx context.Context, bucket string, q *Query, opts ...storageOption) *ObjectIterator
+	ListObjects(ctx context.Context, bucket string, q *Query, softDeleted bool, opts ...storageOption) *ObjectIterator
```
tritone, line 57: Again, I think we can fold `softDeleted` in with `Query` at the bucket level.

storage/grpc_client.go
```diff
@@ -484,22 +488,25 @@ func (c *grpcStorageClient) DeleteObject(ctx context.Context, bucket, object str
...
 	if s.userProject != "" {
 		ctx = setUserProjectMetadata(ctx, s.userProject)
 	}
-	if encryptionKey != nil {
-		req.CommonObjectRequestParams = toProtoCommonObjectRequestParams(encryptionKey)
+	if params.encryptionKey != nil {
+		req.CommonObjectRequestParams = toProtoCommonObjectRequestParams(params.encryptionKey)
+	}
+	if params.getOnlyIfSoftDeleted {
+		req.SoftDeleted = &params.getOnlyIfSoftDeleted
```
tritone, line 509: Seems less confusing to make the param field `softDeleted` to match the API field probably?

storage/grpc_client.go
```diff
@@ -731,7 +764,7 @@ func (c *grpcStorageClient) UpdateBucketACL(ctx context.Context, bucket string,
 func (c *grpcStorageClient) DeleteObjectACL(ctx context.Context, bucket, object string, entity ACLEntity, opts ...storageOption) error {
 	// There is no separate API for PATCH in gRPC.
 	// Make a GET call first to retrieve ObjectAttrs.
-	attrs, err := c.GetObject(ctx, bucket, object, defaultGen, nil, nil, opts...)
+	attrs, err := c.GetObject(ctx, &getObjectParams{bucket, object, defaultGen, nil, nil, false}, opts...)
```
tritone, line 767: Can ACLs be updated on a soft deleted object?

tritone (commented):
Overall looks good. One API change suggestion and a few questions/docs notes.

storage/storage.go
BrennaEpp: I think that's fine, but want to check in with @JesseLovelace - does this check get skipped in other languages?

storage/storage.go
```diff
@@ -1686,6 +1695,10 @@ type Query struct {
 	// prefixes returned by the query. Only applicable if Delimiter is set to /.
 	// IncludeFoldersAsPrefixes is not yet implemented in the gRPC API.
 	IncludeFoldersAsPrefixes bool
+
+	// SoftDeleted indicates whether to list soft-deleted objects. If true, only
+	// objects that have been soft-deleted will be listed.
```
tritone, line 1700: Note as well that by default, soft deleted objects will not be listed.

storage/storage.go
tritone: Okay, let's check this before merging. It would be better if this were not required.

storage/storage.go
```diff
@@ -1068,25 +1068,32 @@ func (o *ObjectHandle) SoftDeleted() *ObjectHandle {
...
+type RestoreOptions struct {
+	/// CopySourceACL indicates whether the restored object should copy the
+	// access controls of the source object. Only valid for buckets with
+	// fine-grained access. If uniform bucket-level access is enabled, setting
+	// CopySourceACL will cause an error.
+	CopySourceACL bool
+}
+
+// Restore will restore a soft-deleted object to a live object.
 // Note that you must specify a generation to use this method.
```
tritone, line 1081: We really should have samples for this feature; people are going to be so confused otherwise.

tritone, line 1081: If generation is not specified, do we get an intelligible error from the server based on whatever the Go client sends by default? Might be good to check this.

tritone (commented):
Overall LGTM, a few more minor questions. Let's try to remove the idempotency check for restore before merging.

storage/storage.go
BrennaEpp: The error should be intelligible, let me check...

GRPC
`rpc error: code = InvalidArgument desc = You must specify a generation. error details: name = BadRequest field =  desc = You must specify a generation.`

Ah, HTTP's is no longer intelligible since it's a required param ` googleapi: Error 400: Invalid argument., invalid`

What do you suggest we do here? Set `if o.gen < 0 { gen = 0 }` to get the intelligible error? Sending generation=0 results in `googleapi: Error 400: Restore request must specify a generation., required`

storage/storage.go
JesseLovelace: Yeah I don't think we have that in the other languages

storage/storage.go
JesseLovelace: I think sending zero to get the intelligible error seems best here

JesseLovelace (approved):

tritone (approved):
nice!

bianpengyuan:
Hello @BrennaEpp @tritone, when do we expect this PR to be released? Seems like the linked release issue has been created for a while now.

BrennaEpp:
> Hello @BrennaEpp @tritone, when do we expect this PR to be released? Seems like the linked release issue has been created for a while now.

Hi @bianpengyuan, this PR is released now in storage v1.41.0

PR #9521, opened 2024-03-07
fix(storage): add object validation case and test

tritone:
Per GCS docs, . and .. are invalid object names.


storage/storage.go
```diff
@@ -1096,6 +1096,10 @@ func (o *ObjectHandle) validate() error {
 	if !utf8.ValidString(o.object) {
 		return fmt.Errorf("storage: object name %q is not valid UTF-8", o.object)
 	}
+	// Names . and .. are not valid; see https://cloud.google.com/storage/docs/objects#naming
+	if o.object == "." || o.object == ".." {
+		return fmt.Errorf("Object name %v is not valid", o.object)
```
codyoss, line 1101: nit: it looks like your other errors are prefixed with storage
nit: error message should start with a lower case letter

storage/storage.go
```diff
@@ -1096,6 +1096,10 @@ func (o *ObjectHandle) validate() error {
 	if !utf8.ValidString(o.object) {
 		return fmt.Errorf("storage: object name %q is not valid UTF-8", o.object)
 	}
+	// Names . and .. are not valid; see https://cloud.google.com/storage/docs/objects#naming
+	if o.object == "." || o.object == ".." {
+		return fmt.Errorf("storage: object name %v is not valid", o.object)
```
BrennaEpp, line 1101: nit
```suggestion
		return fmt.Errorf("storage: object name %q is not valid", o.object)
```

BrennaEpp (approved):

PR #9526, opened 2024-03-08
perf(storage): remove protobuf's copy of data on unmarshalling

BrennaEpp:
- This reduces CPU usage and improves performance when running concurrent reads, since it avoids an extra copy of the data at the protobuf layer since it will unmarshal the data (which is already in a contiguous buffer in the message we receive) into another buffer. We do not need both buffers, so we can return the appropriate address in the buffer rather than copying it into another buffer to return. This is not supported by the proto library today, so we have to manually decode the message buffer.
- Tested performance and cpu usage in VM for large (100MB) and medium file (2MB) reads. The proto unmarshalling portion of cpu usage is no longer sampled (does not show up in the flame graphs).
- Passes unit and integration tests; still needs to be tested for retries

Note that this could be made even more efficient by parsing the whole block at once instead of reusing the `readProtoBytes` function, which will look at and then skip over the fields that have already been parsed. `readProtoBytes` calls protowire functions that do some calculations that re-slice the data buffer, without any copies. So, I don't think the improvement would be noticeable, especially since we only have 4 fields in the message. The cpu flame graphs do not show this function, so whatever time is spent there seems to be negligible.

cc: @dfawley


storage/grpc_client.go
```diff
@@ -1492,10 +1551,144 @@ func (r *gRPCReader) recv() (*storagepb.ReadObjectResponse, error) {
...
+		return content, fmt.Errorf("invalid ReadObjectResponse: %v", err)
+	}
+
+	return content, nil
+}
+
+// readFullObjectResponse returns the ReadObjectResponse encoded in the
+// wire-encoded message buffer b, or an error if the message is invalid.
+// This is used on the first recv of an object as it may contain all fields of
+// ReadObjectResponse.
```
tritone, line 1590: 1. Is there additional overhead of reading the full response vs. just the data? 2. Do we have a guarantee that the API will _never_ return these fields on later responses?

storage/grpc_client.go
```diff
@@ -1492,10 +1551,144 @@ func (r *gRPCReader) recv() (*storagepb.ReadObjectResponse, error) {
...
+func readFullObjectResponse(b []byte) (*storagepb.ReadObjectResponse, error) {
+	const (
+		checksummedDataField        = protowire.Number(1)
+		checksummedDataContentField = protowire.Number(1)
+		checksummedDataCRC32CField  = protowire.Number(2)
```
tritone, line 1595: This is unused so can be removed I think.

storage/grpc_client.go
```diff
@@ -1492,10 +1551,144 @@ func (r *gRPCReader) recv() (*storagepb.ReadObjectResponse, error) {
...
+		objectChecksumsField        = protowire.Number(2)
+		contentRangeField           = protowire.Number(3)
+		metadataField               = protowire.Number(4)
+	)
+
+	checksummedData, err := readProtoBytes(b, checksummedDataField)
+	if err != nil {
+		return nil, fmt.Errorf("invalid ReadObjectResponse: %v", err)
```
tritone, line 1603: These errors should include the field name in the failure; otherwise it's hard to diagnose what's going on.

tritone (commented):
Overall looks good; just a few minor suggestions and questions

storage/grpc_client.go
```diff
@@ -1492,10 +1551,144 @@ func (r *gRPCReader) recv() (*storagepb.ReadObjectResponse, error) {
...
+	}
+	content, err := readProtoBytes(checksummedData, checksummedDataContentField)
+	if err != nil {
+		return nil, fmt.Errorf("invalid ReadObjectResponse: %v", err)
+	}
+
+	// TODO: add unmarshalling for crc32c here
+	//crc32c := readProtoBytes(checksummedData, checksummedDataCRC32CField)
```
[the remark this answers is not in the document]
BrennaEpp: I'm adding it now, so this TODO will be removed. It's not used for now so it doesn't need to be added, but I might as well while I'm here.

storage/grpc_client.go
```diff
@@ -1492,10 +1551,144 @@ func (r *gRPCReader) recv() (*storagepb.ReadObjectResponse, error) {
 		// reopen the stream, but will backoff if further attempts are necessary.
 		// Reopening the stream Recvs the first message, so if retrying is
 		// successful, the next logical chunk will be returned.
-		msg, err = r.reopenStream()
+		msg, err := r.reopenStream()
```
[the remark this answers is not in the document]
BrennaEpp: Might as well; we aren't using anything else from the msg.

storage/grpc_client.go
BrennaEpp: 1. I don't think there will be much overhead of reading the full response unless it does return the metadata; however:
2. Even if the API does return these fields, we don't use them at all. All we did previously is return the content in `Read`, the reference for anything else is discarded when Read returns. Currently the `recv` doesn't even return the rest of the msg either.

storage/grpc_client.go
BrennaEpp: The `checksummedDataCRC32CField` will be used in the change I'll push soon (removing the TODO)

storage/grpc_client.go
```diff
@@ -902,12 +905,53 @@ func (c *grpcStorageClient) RewriteObject(ctx context.Context, req *rewriteObjec
...
 
+// bytesCodec is a grpc codec which permits sending/receiving messages as either
+// protobuf messages, or as raw []bytes.
+type bytesCodec struct {
+	encoding.Codec
+}
+
+func (bytesCodec) Marshal(v any) ([]byte, error) {
+	switch v := v.(type) {
+	case []byte:
```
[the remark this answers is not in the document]
BrennaEpp: It's not relevant. Switching it over to use the default proto codec marshalling.

storage/grpc_client.go
BrennaEpp: Thinking about it, the API may return at least the ChecksummedData.Crc32C field in every message (since this is relevant per msg). If we add checks on every message, we'd need to grab that too, at least. However, as said, we aren't currently using anything other than the content.

storage/grpc_client.go
```diff
@@ -1471,7 +1529,7 @@ func (r *gRPCReader) Close() error {
 	return nil
 }
 
-// recv attempts to Recv the next message on the stream. In the event
+// recv attempts to Recv the next chunk of content on the stream. In the event
```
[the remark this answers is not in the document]
BrennaEpp: That is all correct, I updated the comment to hopefully clarify that.

storage/integration_test.go
```diff
@@ -1023,7 +1023,8 @@ func TestIntegration_ObjectReadChunksGRPC(t *testing.T) {
 	multiTransportTest(skipHTTP("gRPC implementation specific test"), t, func(t *testing.T, ctx context.Context, bucket string, _ string, client *Client) {
 		h := testHelper{t}
 		// Use a larger blob to test chunking logic. This is a little over 5MB.
-		content := bytes.Repeat([]byte("a"), 5<<20)
+		content := make([]byte, 5<<20)
+		rand.New(rand.NewSource(0)).Read(content)
```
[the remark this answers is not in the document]
BrennaEpp: This is thanks to @neild 😃

storage/grpc_client.go
tritone: Got it, that sounds reasonable.

storage/grpc_client.go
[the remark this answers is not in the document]
tritone: Gotcha, that sounds fine then.

tritone (approved):
Looks good, nice work!

storage/grpc_client_test.go
```diff
@@ -0,0 +1,147 @@
...
+		t.Run(test.desc, func(t *testing.T) {
+			// Encode the response.
+			encodedResp, err := proto.Marshal(test.resp)
+			if err != nil {
+				t.Fatalf("proto.Marshal: %v", err)
+			}
+
+			// Unmarshal and decode response using custom decoding.
+			encodedBytes := &[]byte{}
+			if err := bytesCodec.Unmarshal(bytesCodec{}, encodedResp, encodedBytes); err != nil {
```
dfawley, line 132: I have literally never seen this syntax before.

I've always done:

```go
bytesCodec{}.Unmarshal(encodedResp, encodedBytes)
```

TIL...

dfawley (approved):

tritone (approved):

storage/grpc_client.go
```diff
@@ -1492,10 +1549,177 @@ func (r *gRPCReader) recv() (*storagepb.ReadObjectResponse, error) {
...
+	}
+
+	return content, nil
+}
+
+// readFullObjectResponse returns the ReadObjectResponse encoded in the
+// wire-encoded message buffer b, or an error if the message is invalid.
+// This is used on the first recv of an object as it may contain all fields of
+// ReadObjectResponse.
+func readFullObjectResponse(b []byte) (*storagepb.ReadObjectResponse, error) {
```
neild, line 1594: Might want to note that this function is essentially identical to `proto.Unmarshal`, except it aliases the data in the input `[]byte`. If we ever add a feature to `Unmarshal` that does that, this function can be dropped.

storage/grpc_client.go
```diff
@@ -1492,10 +1549,177 @@ func (r *gRPCReader) recv() (*storagepb.ReadObjectResponse, error) {
...
+			return nil, fmt.Errorf("invalid ReadObjectResponse.ChecksummedData.Crc32C: %v", err)
+		}
+
+		checksummedData = &storagepb.ChecksummedData{
+			Content: content,
+			Crc32C:  crc32c,
+		}
+	}
+
+	// Unmarshal remaining fields.
```
neild, line 1619: You're unmarshaling all the fields in the message, so this will be a bit more efficient if you loop once over the fields rather than looking up each individually:

```
off := 0
for off < len(b) {
  num, typ, n := protowire.ConsumeTag(b[off:])
  if n < 0 {
    return nil, protowire.ParseError(n)
  }
  off += n
  switch {
  case num == checksummedDataField && typ == protowire.BytesType:
    // unmarshal the checksummed_data field
  case num == objectChecksumsField && typ == protowire.BytesType:
    // unmarshal the object_checksums field
  case num == contentRangeField && typ == protowire.BytesType:
    // unmarshal the content_range field
  case num == metadataField && typ == protowire.BytesType:
    // unmarshal the metadata field
  default:
    n = protowire.ConsumeFieldValue(num, typ, b[off:])
    if n < 0 {
      return nil, protowire.ParseError(n)
    }
    off += n
  }
}
```

tritone (approved):

storage/grpc_client.go
BrennaEpp: Thanks for the suggestion!

PR #9552, opened 2024-03-11
bigquery: improve error message when inferring schema for recursive type

imjasonh:
This aims to improve the error message from `InferSchema` when the type contains a recursive type.

Today, the error simply includes the type that was provided, which is presumably already known. The goal of this change is to mention the type that was previously seen, indicating the type within the input type that was problematically recursive.

### DIFF
--- a/bigquery/schema.go
+++ b/bigquery/schema.go
@@ -442,12 +442,12 @@ func inferSchemaReflectCached(t reflect.Type) (Schema, error) {
 }
 
 func inferSchemaReflect(t reflect.Type) (Schema, error) {
-	rec, err := hasRecursiveType(t, nil)
+	rec, rt, err := hasRecursiveType(t, nil)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


imjasonh:
Before:

```
Failed to generate schema: bigquery: schema inference for recursive type github.com/google/go-github/v60/github.IssueCommentEvent
```

After:

```
Failed to generate schema: bigquery: schema inference for recursive type github.com/google/go-github/v60/github.Repository
```

In this case [`IssueCommentEvent`](https://pkg.go.dev/github.com/google/go-github/v60/github#IssueCommentEvent) contains [`Repository`](https://pkg.go.dev/github.com/google/go-github/v60/github#Repository) which is the problematic recursive type.

codyoss:
@shollyman for input

codyoss:
@shollyman see my comment above

PR #9567, opened 2024-03-12
chore(internal/postprocessor): process new mods

### LINKED ISSUE #9560: internal/postprocessor: not tidying the modules as it should be
In https://github.com/googleapis/google-cloud-go/pull/9538 and in previous OwlBot PRs, the `vet` action fails b.c go.mod/sum files need tidying. The postprocessor is supposed to be doing this but isn't. We need to fix this regression in order to reduce this manual toil.

noahdietz:
When initializing a new `go.mod` file, add it to the list of `modules` that would otherwise be derived from the OwlBot PR body (that is actually processed by the post processor!) so that it is processed (e.g. tidied) with the rest of the modules changed in the PR.

I explicitly didn't change the logic to tidy modules based on which modules were changed b.c I don't want to make too big a change without having a clear idea of the impact. In the meantime, this fix will keep things ticking over with a little bit less manual toil.

Fixes #9560


quartzmo (approved):

noahdietz:
Verified locally that the new `go.mod` is populated and summed appropriately.

PR #9575, opened 2024-03-13
chore(internal/postprocessor): add validation command and CI

### LINKED ISSUE #9561: internal/postprocessor: validate configuration in PRs
While adding a new API to be generated, it is easy to have a typo in the YAML file that results in a failed run some time after. With how hard it can be to debug a postprocessor run, it would be best if we validated every change to the config.

noahdietz:
Add validation of postprocessor and owlbot config validation.

Fix two issues found:
1. duplicate compute entry in postprocessor config
2. incorrect importpath for cloudquotas in owlbot config

Fixes #9561


internal/postprocessor/config.yaml
```diff
@@ -979,9 +979,6 @@ service-configs:
   - input-directory: google/devtools/cloudbuild/v2
     service-config: cloudbuild_v2.yaml
     import-path: cloud.google.com/go/cloudbuild/apiv2
-  - input-directory: google/cloud/compute/v1
-    service-config: compute_v1.yaml
-    import-path: cloud.google.com/go/compute/apiv1
```
quartzmo, line 984: What's the reason for removing this compute/apiv1 entry?

noahdietz: There are two! I will add that to the description

quartzmo (approved):

PR #9590, opened 2024-03-14
feat(auth): move credentials to base auth package

codyoss:
We are moving the credentials struct to the base auth package as we intend to use this as the main abstraction for this library. To better support the new idea of universes having access to the full credential object will be crutial for the client libraries. Also, the detect package has been renamed to `credentials` to make it more generic in the future. We have plans on directly exposing packages like `externalaccount` in the future so having a better name here makes sense.

This is the first of two large breaking changes for this module. This module not in use by our libraries yet, is version 0.1.X and has always been explictly labeled expirmental in the readme. As such, we feel confortable making these changes for the long term health of the module.


quartzmo (approved):

auth/auth.go
```diff
@@ -116,23 +116,59 @@ func (c *Credentials) JSON() []byte {
 
 // ProjectID returns the associated project ID from the underlying file or
 // environment.
-func (c *Credentials) ProjectID() string {
-	return c.projectID
+func (c *Credentials) ProjectID(ctx context.Context) (string, error) {
```
quartzmo, line 119: I'm wondering if `Credentials` could keep the original methods here as simple getter convenience wrappers for the new getters. These simple versions would pass in nil `ctx` (which panics if `ctx` is used) and they would also panic if the provider would return a non-nil `err`. In other words, these wrappers would panic if a networked impl was used, but would be safe (and convenient!) when `StaticCredentialsProperty` is used.

quartzmo: But tbh, I haven't really looked at when these convenient "static" getters could actually be used...

codyoss: I think I would rather have one or the other and not both. And being this is more of a foundations library that other projects build on-top-of it feels more correct to have the errors. Anyone using the API can always throw the errors away if they truly don't want to check them, but I would rather leave that up to them. In reality, I don't think these methods get called too much, at least they have not been by our libraries historically. (This does change a bit with universe domain). We are for sure trading convenience here -- but I am personally okay with that over correctness for an authentication library.

quartzmo: SGTM

quartzmo (approved):
CredentialsPropertyProvider seems like a smart idea, let's go with it.

PR #8592, opened 2023-09-18
feat(pubsub): add opentelemetry tracing

hongalex:

PR #9594, opened 2024-03-15
feat(pubsub): add otel tracing with links

hongalex:
This replace #8592 and adds the missing subscribe side spans + updates the publish side spans to match our most recent design

- [x] Use links instead of parent/child spans for batch RPC operations
- [x] Align spans names with most recent OpenTelemetry messaging semantic conventions
- [x] Align attributes to most recent messaging semconv
- [x] Add subscribe side integration tests
- [x] Fix integration tests to use span ID generator
- [x] Integrate the newly added AddLinks API to get bidirectional links

Note: this PR uses an experimental golang.org/x/exp/slices package, which was added in go 1.21. This also depends on OpenTelemetry v1.28.0, which requires go 1.21 or higher. This PR will have failing `vet.sh` checks, but should be merged into the base branch, `pubsub-otel-trace`. That branch instead should wait on bumping the min go version before merging.


pubsub/iterator.go
```diff
@@ -123,12 +132,16 @@ func newMessageIterator(subc *vkit.SubscriberClient, subName string, po *pullOpt
 	pingTicker := time.NewTicker(30 * time.Second)
 	cctx, cancel := context.WithCancel(context.Background())
 	cctx = withSubscriptionKey(cctx, subName)
+
+	subID := strings.Split(subName, "/")[3]
```
shollyman, line 136: Two questions:

* Why normalize the subscription ID at all?  Are there concerns about metric collisions because the subscription won't be fully qualified?

* Is this expectation safe, or are there cases where we might have atypical subscription IDs?  Is `len(a)-1` a better idx value to always grab the last element?

pubsub/iterator.go
```diff
@@ -96,6 +101,10 @@ type messageIterator struct {
 	eoMu                      sync.RWMutex
 	enableExactlyOnceDelivery bool
 	sendNewAckDeadline        bool
+
+	enableTracing bool
+	// This maps trace parent spans to ackIDs, used for otel tracing.
```
shollyman, line 106: Given we don't get declare type via explicit initialization, maybe clarify the types for keys and values stored here?  Comment seems to implies the opposite of actual use (span->ackid vs ackid->span).

Also, do we need to worry about growth of this map, or are we guaranteed to always process the LoadAndDelete for all ackids?  Is any kind of LRU handling needed to terminate abandoned spans?

shollyman (commented):
A couple of initial comments.  I haven't gone too deep as I'm not super familiar here.

pubsub/iterator.go
hongalex: 1. SubscriptionID is preferred over the fully qualified name for ease of viewing in traces. For example, the span name "otel-subscription process" shows up nicer than "projects/my-project/subscriptions/otel-subscription process" in Cloud Trace. For metrics, we plan to use the fully qualified name.
2. It probably doesn't hurt to use your suggestion, though I'll probably abstract it away in a helper function. Thanks for the suggestion.

pubsub/iterator.go
hongalex: Fixed the comment, and good catch. It maps ackIDs (strings) -> spans (trace.Span)

This map is periodically cleaned up similar to the pending `keepAliveDeadlines` map that was used before. There's a background thread that occasionally purges expired messages.

pubsub/iterator.go
```diff
@@ -490,30 +531,57 @@ func (it *messageIterator) sendAck(m map[string]*AckResult) {
...
-			_, toRetry := processResults(st, resultsByAckID, md)
-			if len(toRetry) > 0 {
-				// Retry acks in a separate goroutine.
-				go func() {
-					it.retryAcks(toRetry)
-				}()
+			ctx := context.Background()
+			var ackSpan trace.Span
+			if it.enableTracing {
+				ctx, ackSpan = startSpan(context.Background(), ackSpanName, it.subID, trace.WithLinks(links...))
```
jameshartig, line 555: Related to https://github.com/googleapis/google-cloud-go/issues/4665#issuecomment-2045382803 can you store the parentSpan's context in a variable and use it here?

pubsub/iterator.go
```diff
@@ -523,48 +591,94 @@ func (it *messageIterator) sendAck(m map[string]*AckResult) {
...
+
+		// Use of anonymous local function allows span.End to be deferred to end of each loop.
+		func() {
+			numBatch := len(toSend)
+			links := make([]trace.Link, 0, numBatch)
 			for _, ackID := range toSend {
-				resultsByAckID[ackID] = m[ackID]
+				if it.enableTracing {
+					// get the parent span context for this ackID for otel tracing.
+					s, _ := it.activeSpans.Load(ackID)
```
jameshartig, line 629: Should this be `LoadAndDelete` when `isNack`? You're only deleting in `handleKeepAlives` but if the message is nack'd then it's deleted from `keepAliveDeadlines` and won't ever be removed from the sync.Map.

pubsub/iterator.go
```diff
@@ -523,48 +591,94 @@ func (it *messageIterator) sendAck(m map[string]*AckResult) {
...
-				// Retry modacks/nacks in a separate goroutine.
-				go func() {
-					it.retryModAcks(toRetry, deadlineSec, logOnInvalid)
-				}()
+				st, md := extractMetadata(err)
+				_, toRetry := processResults(st, resultsByAckID, md)
+				if len(toRetry) > 0 {
+					// Retry modacks/nacks in a separate goroutine.
+					go func() {
+						it.retryModAcks(toRetry, deadlineSec, logOnInvalid)
```
jameshartig, line 677: retryModAcks aren't traced, should they be?

pubsub/iterator.go
```diff
@@ -490,30 +531,57 @@ func (it *messageIterator) sendAck(m map[string]*AckResult) {
...
-				// Retry acks in a separate goroutine.
-				go func() {
-					it.retryAcks(toRetry)
-				}()
+			ctx := context.Background()
+			var ackSpan trace.Span
+			if it.enableTracing {
+				ctx, ackSpan = startSpan(context.Background(), ackSpanName, it.subID, trace.WithLinks(links...))
+				defer ackSpan.End()
+				ackSpan.SetAttributes(semconv.MessagingBatchMessageCount(numBatch),
```
jameshartig, line 557: Can you add `semconv.MessagingSystemKey.String("pubsub")` and `semconv.MessagingDestinationName(it.subID)` here and to the other spots?

pubsub/subscription.go
```diff
@@ -1387,23 +1417,52 @@ func (s *Subscription) Receive(ctx context.Context, f func(context.Context, *Mes
...
+									eventString = eventNackCalled
+								}
+								ps.AddEvent(eventString)
+								ps.SetAttributes(semconv.MessagingOperationProcess)
+								ps.End()
+								old(ackID, ack, r, receiveTime)
+							}
+						}
+						defer fc.release(otelCtx, msgLen)
+						f(ctx2, m)
```
jameshartig, line 1461: Could this function get sent a context from the `startSpan(otelCtx, processSpanName, s.ID())` ideally we'd have our handler get a context with a span that originated from the process span.

pubsub/iterator.go
jameshartig: That makes sense!

hongalex:
The recent changes from fixing the memory leaks now can result in the application crashing because of a `nil` panic, which happens when a message is acked but already expired and removed from the `activeSpans` map. Will need to cut a release since the previous release is unstable.

VladimirYushkevich:
@hongalex , thanks a lot for adding trace support for pubsub. I tried `cloud.google.com/go/pubsub@v1.37.0-beta.otel.trace2` and have question related to span with [publishMessageBundle](https://github.com/googleapis/google-cloud-go/pull/9594/files#diff-f8e9c2b01fa6ad467fd0d2e77cf8c69531f89ca90f48c127e537c7681ddc764cR956) `code.function` (or `{topicID} publish` name). For some reason it has no parent. It looks wired because `code.function: Publish` (or `{topicID} create` name)  is correctly assigned to my manually created span. Bellow I provide a screenshots of my issue:

**trace=6835bec8eafcfa19e76a02286d4b1f90** my manually created span with `{topicID} create` as a child span:

Screenshot 2024-05-07 at 19 21 32
**trace=777db2f85da93d8cd5dd09c917ffa9b1** `{topicID} publish` span without parent(only with auto instrumented `google.pubsub.v1.Publisher/Publish` child):

Screenshot 2024-05-07 at 19 28 04

I would expect to see `google.pubsub.v1.Publisher/Publish` as a child for `trace=6835bec8eafcfa19e76a02286d4b1f90` before or after `publisher batching` span.

hongalex:
@VladimirYushkevich Yeah so that is intentional. The reason is because the `create` span is only associated with a single message, whereas the `publish` span is a batched span (many messages are batched into a single publish RPC). Tracing requires each span to only have a single parent, so it's not currently possible to have the `publish` span be considered in multiple message traces. Thus, it doesn't have a parent. I will be adding support for SpanLinks though, where the `create` span will have a Link to the `publish` span.

pubsub/trace.go
```diff
@@ -306,71 +306,124 @@ func (c messageCarrier) Keys() []string {
...
 }
 
+// injectPropagation injects context data into the Pub/Sub message's Attributes field.
+func injectPropagation(ctx context.Context, msg *Message) {
+	// only inject propagation if a valid span context was detected.
+	if trace.SpanFromContext(ctx).SpanContext().IsValid() {
+		if msg.Attributes == nil {
+			msg.Attributes = make(map[string]string)
+		}
+		propagation.TraceContext{}.Inject(ctx, newMessageCarrier(msg))
```
GoosvandenBekerom, line 316: Any specific reason you use `propagation.TraceContext{}` here instead of `otel.GetTextMapPropagator()`?

I have a feeling this could benefit from using the global propagator that services likely already have wired for other tracing use cases.

GoosvandenBekerom: For example: in our company we also wire up `propagation.Baggage{}`

hongalex: Thanks for bringing this up. It's definitely we've considered, but opted out of using the global propagator because:
1. We want to reduce the chances of user error, where someone forgets to set the propagator and context propagation breaks 
2. Currently attributes are injected into the message attributes, with the  `googclient_` prefix. These attributes have to be reserved, and passing in an undefined `googclient_foo` attribute will result in an error.

We may consider changing the propagator to a composite propagator that includes baggage in the future. This is a non-breaking change, even for users that use a combination of composite (new) and non-composite (old) propagators.

VladimirYushkevich:
> @VladimirYushkevich Yeah so that is intentional. The reason is because the `create` span is only associated with a single message, whereas the `publish` span is a batched span (many messages are batched into a single publish RPC). Tracing requires each span to only have a single parent, so it's not currently possible to have the `publish` span be considered in multiple message traces. Thus, it doesn't have a parent. I will be adding support for SpanLinks though, where the `create` span will have a Link to the `publish` span.

@hongalex , does [this](https://github.com/googleapis/google-cloud-go/pull/9594/commits/a6b724ab32b09749852064938e7714aefcd15335) commit address my issue for missing batched `publish` span?

pubsub/iterator.go
```diff
@@ -639,35 +643,33 @@ func (it *messageIterator) sendAck(m map[string]*AckResult) {
...
 				semconv.CodeFunction("sendAck"))
 
-			for _, s := range subscribeSpans {
-				s.AddLink(trace.Link{
-					SpanContext: ackSpan.SpanContext(),
-					Attributes: []attribute.KeyValue{
-						semconv.MessagingOperationName(ackSpanName),
-					},
-				})
+			if ackSpan.SpanContext().IsSampled() {
```
jameshartig, line 664: Just curious what the reasoning was for checking this. Is there an issue with linking to a span that isn't sampled? Not that this is a lot of code either way.

hongalex: In Cloud Trace, the link icon appears on the span, but it links to a NOT_FOUND page, which is poor UX in my opinion. We should only create links to spans that are actually sampled.

pubsub/iterator.go
```diff
@@ -552,27 +601,32 @@ func (it *messageIterator) sendAckWithFunc(m map[string]*AckResult, ackFunc ackF
...
+					_, toRetry := processResults(st, resultsByAckID, md)
+					if len(toRetry) > 0 {
+						// Retry acks/modacks/nacks in a separate goroutine.
+						go func() {
+							retryAckFunc(toRetry)
+						}()
+					}
 				}
 			}
+
```
feywind, line 629: Added space on purpose?

pubsub/iterator.go
```diff
@@ -598,13 +690,73 @@ func (it *messageIterator) sendAck(m map[string]*AckResult) {
...
+						defer subscribeSpan.End()
+						defer subscribeSpan.SetAttributes(attribute.String(resultAttribute, resultNacked))
+					}
+					subscribeSpan.AddEvent(eventStart, trace.WithAttributes(semconv.MessagingBatchMessageCount(len(ackIDs))))
+					defer subscribeSpan.AddEvent(eventEnd)
+					links = append(links, trace.Link{SpanContext: subscribeSpan.SpanContext()})
+				}
+			}
+		}
+		var mSpan trace.Span
```
feywind, line 730: I wonder if you could just put this `var` inside the `if` to make one long `enableTracing`.

pubsub/trace.go
```diff
@@ -306,71 +306,118 @@ func (c messageCarrier) Keys() []string {
...
-	numBatchedMessagesAttribute = "messaging.pubsub.num_messages_in_batch"
-	orderingAttribute           = "messaging.pubsub.ordering_key"
+	gcpProjectIDAttribute    = "gcp.project_id"
+	pubsubPrefix             = "messaging.gcp_pubsub."
+	orderingAttribute        = pubsubPrefix + "message.ordering_key"
+	deliveryAttemptAttribute = pubsubPrefix + "message.delivery_attempt"
+	eosAttribute             = pubsubPrefix + "exactly_once_delivery"
+	ackIDAttribute           = pubsubPrefix + "message.ack_id"
+	resultAttribute          = pubsubPrefix + "result"
+	receiptModackAttribute   = pubsubPrefix + "is_receipt_modack"
```
feywind, line 360: I like the way you've done this. I should restructure some of this in the Node implementation.

feywind (approved):
Overall, this looks pretty good :) I need to catch up now...

pubsub/subscription.go
```diff
@@ -1372,9 +1388,32 @@ func (s *Subscription) Receive(ctx context.Context, f func(context.Context, *Mes
...
 					msg := msg
-					// TODO(jba): call acquire closer to when the message is allocated.
+					iter.eoMu.RLock()
+					ackh, _ := msgAckHandler(msg, iter.enableExactlyOnceDelivery)
+					iter.eoMu.RUnlock()
+					// otelCtx is used to relate the main subscribe span to the other subspans in the subscribe side.
+					// We don't want to override the ctx passed into Receive, which is necessary for user-initiated cancellations.
+					// This context is passed into the callback (regardless of if tracing is enabled). Thus, it shadows ctx2 so
+					// the callback is cancellable.
+					otelCtx := ctx2
```
shollyman, line 1401: Should this constructed via context.WithCancel or similar?

pubsub/iterator.go
```diff
@@ -309,6 +328,24 @@ func (it *messageIterator) receive(maxToPull int32) ([]*Message, error) {
...
 			}
 		}
+
+		if it.enableTracing {
+			ctx := context.Background()
+			if m.Attributes != nil {
+				ctx = propagation.TraceContext{}.Extract(ctx, newMessageCarrier(m))
+			}
+			attr := getSubscriberOpts(it.projectID, it.subID, m)
+			_, span := startSpan(ctx, subscribeSpanName, it.subID, attr...)
```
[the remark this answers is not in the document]
hongalex: The custom `startSpan` just alleviates the need to call `fmt.Sprintf` for a few spans (create, subscribe, process) since those need to include the topic/subscriptionID. It's probably not necessary for all spans, and I could just call tracer().Start for most spans if necessary.

The extraction only happens once since this is the only place where the process boundary is crossed (publish ->subscribe). Do you think it's worth creating a custom utility function for this purpose? I suppose it would make the testing easier.

I don't use the new context since I don't want the parent/child relationship in most places.

pubsub/iterator.go
```diff
@@ -552,27 +601,32 @@ func (it *messageIterator) sendAckWithFunc(m map[string]*AckResult, ackFunc ackF
 		wg.Add(1)
 		go func(toSend []string) {
 			defer wg.Done()
+			ctx := context.Background()
```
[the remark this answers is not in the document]
hongalex: This is actually supposed to be explicitly context.Background(), see comment below. I probably didn't need to extract this into a variable though, so I'll revert it back to the inline way as before.

pubsub/service.go
```diff
@@ -130,3 +130,14 @@ func newExactlyOnceBackoff() gax.Backoff {
...
 	}
 }
+
+// parseResourceName parses the project and resource ID from a fully qualified name.
+// For example, "projects/p/topics/my-topic" -> "p", "my-topic"
+func parseResourceName(fqn string) (string, string) {
+	s := strings.Split(fqn, "/")
+	// Some tests don't use FQN, in which case return empty projectID.
+	if len(s) == 1 {
+		return "", s[0]
```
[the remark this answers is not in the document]
hongalex: Not for tests, since we're instantiating the `iterator` directly via [newMessageIterator](https://github.com/hongalex/google-cloud-go/blob/pubsub-otel-trace-links/pubsub/iterator_test.go#L552) We should probably change the tests to use an actual project. If that's preferred, I can make that change and remove the check here.

pubsub/subscription.go
hongalex: I'm not sure when we would need to call `cancel` on this. Are you suggesting that we should add the cancel and then check that `otelCtx.Done()` is called?

pubsub/iterator.go
```diff
@@ -309,6 +329,24 @@ func (it *messageIterator) receive(maxToPull int32) ([]*Message, error) {
...
+			}
+			attr := getSubscriberOpts(it.projectID, it.subID, m)
+			_, span := startSpan(ctx, subscribeSpanName, it.subID, attr...)
+			span.SetAttributes(
+				attribute.Bool(eosAttribute, it.enableExactlyOnceDelivery),
+				attribute.String(ackIDAttribute, ackID),
+				semconv.MessagingBatchMessageCount(len(msgs)),
+				semconv.CodeFunction("receive"),
+			)
+			if span.SpanContext().IsSampled() {
```
jameshartig, line 346: I've been trying to diagnose broken traces and I believe this is the problem. Let's say we have a trace that's `HTTP -> RPC -> Pub/Sub Publish -> Pub/Sub Receive` and it's not sampled because in `HTTP` we sample at 0.1% which means when it gets to here it won't be stored in `activeSpans`.

Later in `*Subscription.Receive` it does `c, ok := iter.activeSpans.Load(ackh.ackID)` and then re-creates a context with the originally-received span but since it wasn't in `activeSpans` it will end up with a span-less context. This means when our handler gets the context and creates a new span it has no context and hits our sampler which might choose to sample it.

This results in spans that seem to start with our handler out of nowhere rather than connecting back to the original `HTTP` span.

hongalex: > This means when our handler gets the context and creates a new span it has no context and hits our sampler which might choose to sample it.

Hm, this is a good point. I originally envisioned it wasn't necessary to store the span contexts of non-sampled spans, since if you use Parent based sampling, none of these spans would be useful. I forgot this is necessary to propagate the `IsSampled == false` status to the handler. Let me remove this check and always store the span contexts even for `IsSampled == false` spans.

pubsub/subscription.go
shollyman: I'm just having trouble following the contexts in use here properly, since you reassign the root ctx2 span to otelCtx each iteration.  For the cases where you don't create  a new span, do any of the usages taint the ctx2 in any way?  There's also https://pkg.go.dev/context#WithoutCancel but I'm not sure that's any better here.

hongalex: > For the cases where you don't create a new span, do any of the usages taint the ctx2 in any way?

No, this doesn't. The assignment of `otelCtx := ctx2` is to create a separate copy of the original context per message (aka per iteration). Modifying `otelCtx` in these cases does not affect the original `ctx2` in any way. This can be demonstrated via a toy example: https://go.dev/play/p/OgTdAcWyy_h

We don't want to use `WithoutCancel` since we want to pass the cancellation signal for ctx2's cancel into the user callback. This is the existing behavior.

shollyman (approved):
Approving this.  Most of my feedback is derived from my less frequent interactions with this client, and the PR itself includes testing and benchmarking for the status quo.

jameshartig:
@hongalex can we get another pre-release of this for testing?

hongalex:
> @hongalex can we get another pre-release of this for testing?

Yeah, I created another release:

`go get cloud.google.com/go/pubsub@v1.42.0-beta.otel.trace`

This doesn't contain changes for configuring num callbacks, but I'll update that branch to pull in changes from here as well.

jameshartig:
Thanks!

> This doesn't contain changes for configuring num callbacks, but I'll update that branch to pull in changes from here as well.

That's okay for now, we have another application where we can test the traces without needing to configure num callbacks.

PR #9600, opened 2024-03-18
chore(storage/control): add config to generate apiv2

noahdietz:
PiperOrigin-RevId: 616205855

### DIFF
--- a/.github/.OwlBot.yaml
+++ b/.github/.OwlBot.yaml
@@ -354,6 +354,7 @@ deep-remove-regex:
   - /internal/generated/snippets/speech/apiv1/
   - /internal/generated/snippets/speech/apiv1p1beta1/
   - /internal/generated/snippets/speech/apiv2/
+  - /internal/generated/snippets/storage/control/apiv2/
   - /internal/generated/snippets/storageinsights/apiv1/
   - /internal/generated/snippets/storagetransfer/apiv1/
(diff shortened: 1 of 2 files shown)


.github/.OwlBot.yaml
```diff
@@ -1027,6 +1029,8 @@ deep-copy-regex:
     dest: /
   - source: /google/cloud/speech/v2/cloud.google.com/go
     dest: /
+  - source: /google/storage/v2/cloud.google.com/go/storage/control/apiv2
+    dest: /
   - source: /google/storage/v2/cloud.google.com/go/storage/internal/apiv2
     dest: /storage/internal/apiv2
```
noahdietz, line 1035: I'm not sure why the internal GAPIC for the data plane API has a non-`/` value for the `dest` ...any ideas?

frankyn: Not sure if this question is for Storage team; but could you clarify what you mean by `non-/` value?

noahdietz: This was a question for @quartzmo (he's more experienced with this config than me, I'm trying to reacquaint myself).

If you look at most other entries, the `dest` is just `/`. This is the "deep copy" destination. And in talking this through with you in this comment I realize I need to set `dest` to `/storage/control/apiv2` :)

Thanks for rubber 🦆 'ing

Issue #9464, opened 2024-02-23
datastore: missing tracing from run method in query.go


notcool11:
**Code**

Looking at the `Run` method in the `query.go` file here:
https://github.com/googleapis/google-cloud-go/blob/6cd6a73be87a261729d3b6b45f3d28be93c3fdb3/datastore/query.go#L717

Other methods have the tracing calls right at the front like this:

```go
	ctx = trace.StartSpan(ctx, "cloud.google.com/go/datastore.Query.RunAggregationQuery")
	defer func() { trace.EndSpan(ctx, err) }()
```

But the `Run` method does not.

**Expected behavior**

When using the `Run` method I expected datastore to generate spans associated to the current trace context.

**Actual behavior**

No spans are generated by the `Run` method.

**Additional context**

It is possible this is by design because other datastore functions like `Count` have tracing applied at the top and later make a call to `Run` like this:
https://github.com/googleapis/google-cloud-go/blob/6cd6a73be87a261729d3b6b45f3d28be93c3fdb3/datastore/query.go#L615

Perhaps a different solution would be to have a private `run` function that doesn't have tracing while the public one does? It is confusing and inconsistent when some methods are traced while others are not. 🤷 

Thanks!

PR #9602, opened 2024-03-18
fix(datastore): Adding tracing to run method

bhshkh:
Find info in https://github.com/googleapis/google-cloud-go/issues/9464


notcool11:
thank you!

do you know if i will get notified for if/when this PR makes it into a release so i can pull that version of the package? thanks again!

bhshkh:
> thank you!
> 
> do you know if i will get notified for if/when this PR makes it into a release so i can pull that version of the package? thanks again!

Once the PR is merged, it will be included in the next release which is 1.16.0

You can subscribe to https://github.com/googleapis/google-cloud-go/pull/8711 to get notified.

gkevinzheng (approved):

bhshkh:

PR #9603, opened 2024-03-18
feat(auth): refactor public sigs to use Credentials

codyoss:
This refactors all public surfaces to use Credentials rather than TokenProvider as the main abstraction.

Part 2 of 3 of breaking changes for the auth module.


auth/downscope/downscope.go
```diff
@@ -28,12 +28,11 @@ import (
...
 
-// Options for configuring [NewTokenProvider].
+// Options for configuring [NewCredentials].
 type Options struct {
-	// BaseProvider is the [cloud.google.com/go/auth.TokenProvider] used to
-	// create the downscoped provider. The downscoped provider therefore has
-	// some subset of the accesses of the original BaseProvider. Required.
-	BaseProvider auth.TokenProvider
+	// Credentials is the [cloud.google.com/go/auth.Credentials] used to
+	// create the downscoped provider. Required.
```
quartzmo, line 34: Rename "provider" to "credentials" throughout these docs, where the term refers to the return value of `NewCredentials`.

auth/grpctransport/grpctransport.go
```diff
@@ -65,9 +65,9 @@ type Options struct {
 	// PoolSize is specifies how many connections to balance between when making
 	// requests. If unset or less than 1, the value defaults to 1.
 	PoolSize int
-	// TokenProvider specifies the provider used to add Authorization metadata
+	// Credentials specifies the provider used to add Authorization metadata
```
quartzmo, line 68: "provider" might be confusing here.

auth/grpctransport/grpctransport.go
```diff
@@ -240,37 +239,37 @@ func dial(ctx context.Context, secure bool, opts *Options) (*grpc.ClientConn, er
 	return grpc.DialContext(ctx, endpoint, grpcOpts...)
 }
 
-// grpcTokenProvider satisfies https://pkg.go.dev/google.golang.org/grpc/credentials#PerRPCCredentials.
-type grpcTokenProvider struct {
-	auth.TokenProvider
+// grpcCredentialProvider satisfies https://pkg.go.dev/google.golang.org/grpc/credentials#PerRPCCredentials.
```
quartzmo, line 242: I would rename to `grpcCredentialsProvider`, since `Credentials` is the type name.

auth/httptransport/httptransport.go
```diff
@@ -49,9 +49,9 @@ type Options struct {
 	// APIKey specifies an API key to be used as the basis for authentication.
 	// If set DetectOpts are ignored.
 	APIKey string
-	// TokenProvider specifies the provider used to add Authorization header to
+	// Credentials specifies the provider used to add Authorization header to
```
quartzmo, line 52: "provider" might be confusing here.

auth/httptransport/httptransport.go
```diff
@@ -129,18 +129,18 @@ type InternalOptions struct {
...
 	}
 	base := client.Transport
 	if base == nil {
 		base = http.DefaultTransport.(*http.Transport).Clone()
 	}
 	client.Transport = &authTransport{
-		provider: auth.NewCachedTokenProvider(tp, nil),
+		provider: creds,
 		base:     base,
 	}
```
quartzmo, line 145: Should we call:

```
creds.TokenProvider = auth.NewCachedTokenProvider(creds.TokenProvider, nil)
```
here just in case the user did not construct the provided credentials with a wrapped token provider? The func is a no-op if the TokenProvider is already wrapped.

auth/idtoken/compute.go
```diff
@@ -39,8 +39,14 @@ func computeTokenProvider(opts *Options) (auth.TokenProvider, error) {
...
-	return auth.NewCachedTokenProvider(tp, &auth.CachedTokenProviderOptions{
-		ExpireEarly: 5 * time.Minute,
+	return auth.NewCredentials(&auth.CredentialsOptions{
+		TokenProvider: auth.NewCachedTokenProvider(tp, &auth.CachedTokenProviderOptions{
+			ExpireEarly: 5 * time.Minute,
+		}),
+		ProjectIDProvider: auth.CredentialsPropertyFunc(func(context.Context) (string, error) {
+			return metadata.ProjectID()
+		}),
+		// TODO(quartzmo): add universe domain resolver here
```
quartzmo, line 49: Thanks for this! 🥇

auth/idtoken/idtoken.go
```diff
@@ -87,18 +87,18 @@ func (o *Options) validate() error {
...
 	if b := opts.jsonBytes(); b != nil {
-		return tokenProviderFromBytes(b, opts)
+		return credsFromBytes(b, opts)
 	}
 	if metadata.OnGCE() {
-		return computeTokenProvider(opts)
+		return computeCredentials(opts)
 	}
 	return nil, fmt.Errorf("idtoken: couldn't find any credentials")
 }
```
quartzmo, line 104: I believe the previous PR in this series may have introduced a potential bug that I just noticed now: Search the package for

```
udp = auth.CredentialsPropertyFunc(creds.QuotaProjectID)
```

and you should see three usages where `udp` is assigned to `UniverseDomainProvider`. Maybe it should be assigned to `QuotaProjectIDProvider` instead?

quartzmo (changes requested):

auth/httptransport/httptransport.go
quartzmo: OK, I understand this reasoning.

quartzmo (approved):

PR #9616, opened 2024-03-20
fix(storage): retry errors from last recv on uploads

BrennaEpp:
Missed this case when switching to bidi streams


storage/grpc_client.go
```diff
@@ -1872,6 +1872,7 @@ func (w *gRPCWriter) uploadBuffer(recvd int, start int64, doneReading bool) (*st
 
 	// Send a request with as many bytes as possible.
 	// Loop until all bytes are sent.
+sendBytes:
```
tritone, line 1875: This is cool but I've literally never seen this in Go code before. Can you add a clarifying comment? It's just pretty rare syntax.

tritone (commented):
Overall looks good. Do you feel like our existing test coverage is sufficient for this? If some of these error/continue cases are not hit by our conformance tests, we should talk about adding those.

BrennaEpp:
> Overall looks good. Do you feel like our existing test coverage is sufficient for this? If some of these error/continue cases are not hit by our conformance tests, we should talk about adding those.

This is caught by our conformance tests, but they are currently skipped. Once the support in the testbench for retrying bidi writes is merged we can remove the skip and it would catch this.

tritone (approved):

cojenco:
> > Overall looks good. Do you feel like our existing test coverage is sufficient for this? If some of these error/continue cases are not hit by our conformance tests, we should talk about adding those.
> 
> This is caught by our conformance tests, but they are currently skipped. Once the support in the testbench for retrying bidi writes is merged we can remove the skip and it would catch this.

Update: retry support for bidi writes just merged and is in the [v0.43.0 testbench release](https://github.com/googleapis/storage-testbench/releases/tag/v0.43.0)

PR #9621, opened 2024-03-20
refactor(auth): move things around and rename

codyoss:
- rename package internaldetect to creds
- move downscope, idtoken, and impersonate under credentials


quartzmo (approved):
Long live credsfile!

PR #9634, opened 2024-03-22
feat(auth): add downscope.Options.UniverseDomain

quartzmo:
Corresponds to https://github.com/golang/oauth2/commit/deefa7e836e5aa043b2b8102b5b098da53a43903

### DIFF
--- a/auth/credentials/downscope/downscope.go
+++ b/auth/credentials/downscope/downscope.go
@@ -51,20 +51,20 @@ type Options struct {
 	UniverseDomain string
 }
 
-func (c Options) client() *http.Client {
-	if c.Client != nil {
-		return c.Client
+func (o *Options) client() *http.Client {
+	if o.Client != nil {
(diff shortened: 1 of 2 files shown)


auth/credentials/downscope/downscope.go
```diff
@@ -51,6 +58,15 @@ func (c Options) client() *http.Client {
 	return internal.CloneDefaultClient()
 }
 
+// identityBindingEndpoint returns the identity binding endpoint with the
+// configured universe domain.
+func (c Options) identityBindingEndpoint() string {
```
codyoss, line 63: nit: would you mind changing this method and the one above to pointer semantics.

codyoss (approved):

auth/credentials/downscope/downscope.go
quartzmo: Done, good catch!

PR #9636, opened 2024-03-23
test(storage): add several-msg + resumable uploads to objects.insert

BrennaEpp:
Not all error code paths in grpc uploads are tested without uploading a larger object in the storage.objects.insert test.

Additionally, reduces the 9mb object to 3mb where possible to reduce test time.


cojenco (approved):
lgtm, thank you!

storage/retry_conformance_test.go
```diff
@@ -392,6 +404,42 @@ var methods = map[string][]retryFunc{
 			}
 			return nil
 		},
+		// Single-shot upload that spans several message sends.
```
tritone, line 407: Note that the "message sends" part is only relevant to gRPC

tritone (approved):
Looks great! Can you confirm that this covers the remaining untested error paths for gRPC?

BrennaEpp:
> Looks great! Can you confirm that this covers the remaining untested error paths for gRPC?

I tested this out and it does cover all error paths for gRPC.

PR #9638, opened 2024-03-23
chore(storage): gRPC object/bucket empty update parity with JSON

### LINKED ISSUE #7685: storage: address grpc/http difference in response to empty bucket update
gRPC returns an error, HTTP allows it as a no-op

- Object update has the same issue

BrennaEpp:
Fixes #7685


storage/grpc_client.go
```diff
@@ -592,6 +599,16 @@ func (c *grpcStorageClient) UpdateObject(ctx context.Context, params *updateObje
 
 	req.UpdateMask = fieldMask
 
+	if len(fieldMask.Paths) < 1 {
+		// Nothing to update. To maintain consistency with JSON, we must still
+		// update the object because metageneration and other fields are
+		// updated even on an empty update.
+		// gRPC will fail if the fieldmask is empty, so instead we add an
+		// output-only field to the update mask. Output-only fields are (and must
+		// be - see AIP 161) ignored, but allow us to send an empty update.
```
tritone, line 608: We should maybe clarify that this is intended behavior. Feels inconsistent that empty updates are barred but this is permitted.

storage/integration_test.go
```diff
@@ -3879,15 +3898,6 @@ func TestIntegration_CustomTime(t *testing.T) {
 		if _, err := obj.Update(ctx, ObjectAttrsToUpdate{CustomTime: earlierTime}); err == nil {
 			t.Fatalf("backdating CustomTime: expected error, got none")
 		}
-
-		// Zero value for CustomTime should be ignored. Set TemporaryHold so that
-		// we don't send an empty update request, which is invalid for gRPC.
-		if _, err := obj.Update(ctx, ObjectAttrsToUpdate{TemporaryHold: false}); err != nil {
-			t.Fatalf("empty update: %v", err)
-		}
-		if err := checkCustomTime(laterTime); err != nil {
```
tritone, line 3888: I think we should still check that custom time doesn't change after an empty update right?

BrennaEpp: Yeah... I figured we test empty updates elsewhere in several places, but I guess we don't actually set a CustomTime before the empty update, so I've added this back.

storage/grpc_client.go
BrennaEpp: Added the reasoning behind why this is permitted to the comment.

tritone (approved):

PR #9653, opened 2024-03-27
feat(datastore): Configure both mTLS and TLS endpoints for Datastore client

bhshkh:
The benefit of the change is so that the transport layer can establish mTLS connection where possible, thus improving security for communication to google apis.

Resolves: [b/329657227](http://b/329657227)


datastore/datastore.go
```diff
@@ -115,6 +118,7 @@ func NewClientWithDatabase(ctx context.Context, projectID, databaseID string, op
 	} else {
 		o = []option.ClientOption{
 			option.WithEndpoint(prodAddr),
```
xmenxk, line 120: Can we use withDefaultEndpoint() in place of WithEndpoint()?
The WithEndpoint() param is treated as an overriding option by the transport layer code, thus preventing the WithDefaultMTLSEndpoint() param from being used. thx

bhshkh: Compiler warning:

internaloption.WithDefaultEndpoint is deprecated: WithDefaultEndpoint does not support setting the universe domain. Use WithDefaultEndpointTemplate and WithDefaultUniverseDomain to compose the default endpoint instead.

So, used template and domain

bhshkh:
DO NOT MERGE until freeze is over

datastore/datastore.go
xmenxk: I see looks like things changed recently, but yes you are right, we can use the template and domain instead, e.g. https://github.com/googleapis/google-cloud-go/blob/main/ai/generativelanguage/apiv1/generative_client.go#L61

xmenxk (approved):

gkevinzheng (approved):
LGTM

Issue #9519, opened 2024-03-06
spanner: panic in recordGFELatencyMetricsOT if EnableOpenTelemetryMetrics called after NewClient


henripqt:
**Client**

Spanner

**Environment**

e.g. Alpine Docker on GKE

**Go Environment**

go1.22.0

**Code**

When enabling open telemetry after creating the client the configuration set on the client is invalid resulting in potential panic.

e.g.
```go
package main

func main() {
   // ... 
   	client, err := gcspanner.NewClient(
		ctx,
		db,
		opts...,
	)

	if err != nil {
		panic(err)
	}

	initOnce.Do(func() {
		if os.Getenv("SPANNER_EMULATOR_HOST") == "" {
			spanner.EnableOpenTelemetryMetrics()
		}
	})
	
	// ...
}
```

In the above example the new client created would have a default `openTelemetryConfig` with all instruments being nil possibly leading to a panic during a call to `recordGFELatencyMetricsOT` (see bellow with added comments)

```go

func recordGFELatencyMetricsOT(ctx context.Context, md metadata.MD, keyMethod string, otConfig *openTelemetryConfig) error {
	if !IsOpenTelemetryMetricsEnabled() || md == nil && otConfig == nil { // the global flag is enable thus we don't return nil
		return nil
	}
	attr := otConfig.attributeMap
	if len(md.Get("server-timing")) == 0 && otConfig.gfeHeaderMissingCount != nil { // at this point otConfig.gfeHeaderMissingCount is nil
		otConfig.gfeHeaderMissingCount.Add(ctx, 1, metric.WithAttributes(attr...))
		return nil
	}
	serverTiming := md.Get("server-timing")[0] // This will panic !
	gfeLatency, err := strconv.Atoi(strings.TrimPrefix(serverTiming, "gfet4t7; dur="))
	if !strings.HasPrefix(serverTiming, "gfet4t7; dur=") || err != nil {
		return err
	}
	attr = append(attr, attributeKeyMethod.String(keyMethod))
	if otConfig.gfeLatency != nil {
		otConfig.gfeLatency.Record(ctx, int64(gfeLatency), metric.WithAttributes(attr...))
	}
	return nil
}

```

**Expected behavior**

- The EnableOpenTelemetryMetrics function should mention that it needs to be registered prior to enabling the client
- recordGFELatencyMetricsOT should have appropriate checks to handle reading from metadata while avoiding panics

**Actual behavior**

[...]

PR #9657, opened 2024-03-28
fix(spanner): always set open telemetry config during client initialization

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/9519

### DIFF
--- a/spanner/test/opentelemetry/test/ot_metrics_test.go
+++ b/spanner/test/opentelemetry/test/ot_metrics_test.go
@@ -259,9 +259,11 @@ func TestOTMetrics_GFELatency(t *testing.T) {
 	t.Cleanup(func() {
 		te.Unregister(ctx)
 	})
-	spanner.EnableOpenTelemetryMetrics()
 	server, client, teardown := setupMockedTestServerWithConfig(t, spanner.ClientConfig{OpenTelemetryMeterProvider: te.mp})
 	defer teardown()
+
+	// enabling OpenTelemetry metrics after spanner client initialization
+	spanner.EnableOpenTelemetryMetrics()
 
 	if err := server.TestSpanner.PutStatementResult("SELECT email FROM Users", &stestutil.StatementResult{
 		Type: stestutil.StatementResultResultSet,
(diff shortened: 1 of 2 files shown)


spanner/test/opentelemetry/test/ot_metrics_test.go
```diff
@@ -259,10 +259,12 @@ func TestOTMetrics_GFELatency(t *testing.T) {
 	t.Cleanup(func() {
 		te.Unregister(ctx)
 	})
-	spanner.EnableOpenTelemetryMetrics()
 	server, client, teardown := setupMockedTestServerWithConfig(t, spanner.ClientConfig{OpenTelemetryMeterProvider: te.mp})
 	defer teardown()
 
+	// enabling OpenTelemetry metrics after spanner client initialization
+	spanner.EnableOpenTelemetryMetrics()
```
rahul2393, line 266: this will panic in spanner client version <= v1.60.0

olavloite: Hmmm... I'm wondering whether this is partly 'works as intended'. Not the panic, of course, but:
1. `spanner.EnableOpenTelemetryMetrics()` sets a global variable that enables it for __all__ clients.
2. Calling `spanner.EnableOpenTelemetryMetrics()` after having created one or more clients will then enable it for all those clients, right? What will become the behavior of those clients after calling `spanner.EnableOpenTelemetryMetrics()`? Also, what happens to any client that was created and using OpenCensus before this method gets called?

Would it maybe make more sense to do the following:
1. Set a global variable the first time that a client is created.
2. Return an error if someone tries to call `spanner.EnableOpenTelemetrymetrics()` after a client has been created.
3. Also return an error if someone tries to create a client with an `OpenTelemetryMetricProvider` or any other OpenTelemetry-related configuration without first having called `spanner.EnableOpenTelemetryMetrics()`.

The reason that I think that the above is a more reasonable approach is that I don't think it is clear to users what happens with existing clients if you call `spanner.EnableOpenTelemetryMetrics()` at a random point in time in your application.

codyoss:
@rahul2393 is this something we still want?

PR #9659, opened 2024-03-28
feat(storage): Implement io.WriterTo in Reader

tritone:
This allows the gRPC Reader to write directly into the application write buffer, saving a data copy.

Users can get the benefit of this directly by explicitly calling Reader.WriteTo, but they can also benefit implicitly if they are calling io.Copy.

A bunch of checksum logic had to be moved from the parent Reader into the transport Readers to make this work, since we need to update the checksum for every message read in WriteTo.


storage/grpc_client.go
```diff
@@ -1042,6 +1043,16 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
 	// This is the size of the entire object, even if only a range was requested.
 	size := obj.GetSize()
 
+	// Only support checksums when reading an entire object, not a range.
+	var (
+		wantCRC  uint32
+		checkCRC bool
+	)
+	if checksums := msg.GetObjectChecksums(); checksums != nil && checksums.Crc32C != nil && params.offset == 0 && params.length < 0 {
```
codyoss, line 1051: I don't have the full context here, but do you really want params.length to be negative here?

storage/grpc_client.go
```diff
@@ -1512,10 +1545,78 @@ func (r *gRPCReader) Read(p []byte) (int, error) {
...
+// WriteTo writes all the data requested by the Reader into w, implementing
+// io.WriterTo.
+func (r *gRPCReader) WriteTo(w io.Writer) (int64, error) {
+	// The entire object has been read by this reader, check the checksum if
+	// necessary and return EOF.
+	if r.size == r.seen || r.zeroRange {
+		if err := r.runCRCCheck(); err != nil {
+			return 0, err
+		}
+		return 0, io.EOF
```
codyoss, line 1562: I noticed this above in the read too, but I wonder if these should be storage specific errors so users have more context? Or maybe wrapping happens at a higher layer. `fmt.Errof("storage: something something: %w", io.EOF)`. Then your checks would need to be changed to `errors.Is(err, io.EOF)`, but I think that change should happen regardless.

storage/grpc_client.go
```diff
@@ -1512,10 +1545,78 @@ func (r *gRPCReader) Read(p []byte) (int, error) {
...
+	}
+
+	// No stream to read from, either never initialized or Close was called.
+	// Note: There is a potential concurrency issue if multiple routines are
+	// using the same reader. One encounters an error and the stream is closed
+	// and then reopened while the other routine attempts to read from it.
+	if r.stream == nil {
+		return 0, fmt.Errorf("reader has been closed")
```
codyoss, line 1570: nit: this error should have the storage prefix.

storage/grpc_client.go
```diff
@@ -1512,10 +1545,78 @@ func (r *gRPCReader) Read(p []byte) (int, error) {
...
+	// The entire object has been read by this reader, check the checksum if
+	// necessary and return EOF.
+	if r.size == r.seen || r.zeroRange {
+		if err := r.runCRCCheck(); err != nil {
+			return 0, err
+		}
+		return 0, io.EOF
+	}
+
+	// No stream to read from, either never initialized or Close was called.
```
codyoss, line 1565: Is this handled in an above layer or is this documented as such on the public way this methods is called into?

storage/http_client.go
```diff
@@ -1229,7 +1233,22 @@ func (r *httpReader) Read(p []byte) (int, error) {
 		m, err := r.body.Read(p[n:])
 		n += m
 		r.seen += int64(m)
-		if err == nil || err == io.EOF {
+		if r.checkCRC {
+			r.gotCRC = crc32.Update(r.gotCRC, crc32cTable, p[:n])
+		}
+		if err == nil {
+			return n, err
```
codyoss, line 1240: nit: `return n, nil`

codyoss (commented):
Just a couple of nits or else LGTM

storage/grpc_client.go
BrennaEpp: I believe this is so that we only check the checksums on the entire object read, in which case params.length would be negative to indicate no limit.

storage/retry_conformance_test.go
```diff
@@ -211,12 +210,24 @@ var methods = map[string][]retryFunc{
...
 				return err
 			}
-			wr, err := io.Copy(ioutil.Discard, r)
+			wr, err := r.WriteTo(io.Discard)
 			if got, want := wr, len(randomBytesToWrite); got != int64(want) {
 				return fmt.Errorf("body length mismatch\ngot:\n%v\n\nwant:\n%v", got, want)
 			}
 			return err
 		},
+		func(ctx context.Context, c *Client, fs *resources, _ bool) error {
```
BrennaEpp, line 219: Suggest mentioning the difference between this and the previous func and why we test both.

storage/grpc_client.go
```diff
@@ -1464,12 +1472,36 @@ type gRPCReader struct {
...
+func (r *gRPCReader) updateCRC(b []byte) {
+	if r.checkCRC {
+		r.gotCRC = crc32.Update(r.gotCRC, crc32cTable, b)
+	}
+}
+
+// Checks whether the CRC matches at the conclusion of a read, if CRC checking was enabled.
+func (r *gRPCReader) runCRCCheck() error {
+	if r.checkCRC {
+		if r.gotCRC != r.wantCRC {
```
BrennaEpp, line 1490: nit: this can be collapsed into a single check

storage/grpc_client.go
tritone: Yes, Brenna is correct.

storage/grpc_client.go
tritone: fixed here and in Read.

storage/grpc_client.go
tritone: Yeah I think there is still more we can do to have more coherently wrapped errors. it's been on the back burner.

PR #9666, opened 2024-03-29
test(storage): update read test to use both WriteTo and Read

BrennaEpp:
- This caught an io.EOF error that shouldn't be returned
- Also removes some references to ioutil that has been deprecated


tritone (commented):
Superceded by #9681

tritone:
Superceded by

Issue #6476, opened 2022-08-04
Bigtable: stream terminated by RST_STREAM with error code: INTERNAL_ERROR is not retried by client


aravindsrinivasan:
Package: `cloud.google.com/go/bigtable v1.13.0`

When our bigtable instance has high throughput we see the following error on occasion:
```
rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR
```
This to me looks like a transient error that should be retried by the client. I see the Spanner client retrying this exact error:
https://github.com/googleapis/google-cloud-go/blob/5a2ed6b2cd1c304e0f59daa29959863bff9b5c29/spanner/retry.go#L59-L70

Could also do the same in the Bigtable client? Or is this error actually not retryable in the context of Bigtable?

aravindsrinivasan:
Any update here?

bhshkh:
Not yet picked up. Will pick up soon

PR #9673, opened 2024-03-31
fix(bigtable): Retry on RST_STREAM error

bhshkh:
Fixes: https://github.com/googleapis/google-cloud-go/issues/6476

**Issue**: When operation fails with error "stream terminated by RST_STREAM with error code: INTERNAL_ERROR", it is not retried even though spanner client retries on same error.

**Fix**: Currently, gax.Invoke is retried only on DeadlineExceeded, Unavailable and Aborted. This PR modifies this behavior to retry on Internal error if error message contains "stream terminated by RST_STREAM"


bhshkh:
Do not merge until release freeze ends mid April

gkevinzheng (approved):

bigtable/bigtable.go
```diff
@@ -113,15 +114,40 @@ var (
...
+// bigtableRetryer extends the generic gax Retryer, but also checks
+// error messages to check if operation can be retried
+type bigtableRetryer struct {
+	gax.Retryer
+	gax.Backoff
+}
+
+func (r *bigtableRetryer) Retry(err error) (time.Duration, bool) {
+	if status.Code(err) == codes.Internal &&
+		strings.Contains(err.Error(), "stream terminated by RST_STREAM") {
```
daniel-sanche, line 139: nit: add a comment with context for why this is a special case, with a link back to this bug

daniel-sanche, line 139: is there a way to make this more flexible? Like take in a list or regex string describing the retryable error messages?

bigtable/bigtable.go
```diff
@@ -113,15 +114,40 @@ var (
...
 
+// bigtableRetryer extends the generic gax Retryer, but also checks
+// error messages to check if operation can be retried
+type bigtableRetryer struct {
+	gax.Retryer
+	gax.Backoff
+}
+
+func (r *bigtableRetryer) Retry(err error) (time.Duration, bool) {
+	if status.Code(err) == codes.Internal &&
```
daniel-sanche, line 138: I assume it's not feasible to retry all InternalErrors?

Have you discussed this with Igor or Mattie? They might have thoughts about what can be retried safely. I wonder if we should be considering this for other clients

bhshkh: @igorbernstein2 , this should be done for all other clients, right?
Also, should all internal errors be retried ?

bigtable/bigtable.go

bigtable/bigtable.go
bhshkh: Checked with Igor regarding which errors need to be retried. Share the document with the team. Will update the PR

bhshkh: Created a separate issue to track work on rest of the errors

PR #9682, opened 2024-04-01
feat(pubsub): use Streaming Pull response for ordering check

### LINKED ISSUE #9374: pubsub: remove GetSubscription call when pulling with ordering keys
Introduced in #3884, we started checking the subscription config when message ordering is enabled. This should no longer be necessary since streaming pull's response should contain [an ordering key field](https://github.com/googleapis/googleapis/blob/61f0bb701fd7ec559dec9749a3a5e8241ff45ff2/google/pubsub/v1/pubsub.proto#L1555).

We should remove the use of `subscription.checkOrdering` and rely solely on the streaming pull response.

hongalex:
This PR switches the ordering check on subscription away from the `GetSubscription` call to reading the ordering field of the StreamingPull response instead.

Fixes #9374


pubsub/iterator.go
```diff
@@ -352,12 +358,29 @@ func (it *messageIterator) recvMessages() ([]*pb.ReceivedMessage, error) {
...
 	}
-	it.eoMu.Lock()
-	if got := res.GetSubscriptionProperties().GetExactlyOnceDeliveryEnabled(); got != it.enableExactlyOnceDelivery {
+
+	// If the new exactly once settings are different than the current settings, update it.
+	it.eoMu.RLock()
+	enableEOD := it.enableExactlyOnceDelivery
+	it.eoMu.RUnlock()
+
+	if got := res.GetSubscriptionProperties().GetExactlyOnceDeliveryEnabled(); got != enableEOD {
```
alvarowolfx, line 367: nit: create a variable `subprop := res.GetSubscriptionProperties()` to reused on both check.

kamalaboulhosn (approved):

PR #9686, opened 2024-04-02
feat(spanner/spannertest): support JSON columns and REGEXP_CONTAINS and JSON_VALUES functions

adg:
Add support for a couple of features in the spannertest package. These will be very helpful for my project.


adg:
@rahul2393 Are you the right person to review this?

Note that the JSON_VALUES implementation pulls in a new module dependency, which according to [the policy](https://github.com/googleapis/google-cloud-go/blob/main/CONTRIBUTING.md#policy-on-new-dependencies) requires an exception. I could implement a subset of jsonpath that would suit my purposes here, if that would be preferred.

rahul2393:
@adg Thanks for raising the PR, and yes it will be really helpful if you can pull in the changes from third party library here, one of the reason we avoid using third party deps is because while importing in google3 it will cause build failures.

Let me know once you are done.

codyoss:
@adg friendly ping, is this something you are still wanting to work on?

codyoss:
@adg please see my comment above

PR #9687, opened 2024-04-02
fix(vertexai): if GenerateContentResponse.Candidates.Content is nil will panic

liushuangls:


codyoss:
@eliben @jba

eliben (approved):
Thank you

PR #9689, opened 2024-04-02
chore(storage): bring dependencies down to 1.20 [benchmarks]

BrennaEpp:


storage/internal/benchmarks/main.go
```diff
@@ -401,8 +399,8 @@ func enableTracing(ctx context.Context, sampleRate float64) func() {
 
 	// Use opencensus bridge to pick up OC traces from the storage library.
 	// TODO: remove this when migration to OpenTelemetry is complete.
-	tracer := otel.GetTracerProvider().Tracer(tracerName)
-	octrace.DefaultTracer = opencensus.NewTracer(tracer)
+	//tracer := otel.GetTracerProvider().Tracer(tracerName)
+	//octrace.DefaultTracer = opencensus.NewTracer(tracer)
```
noahdietz, line 403: Can these simply be deleted as  the TODO above suggests?

tritone (approved):

storage/internal/benchmarks/main.go
BrennaEpp: Removed. I'm not sure it will actually work, but it needs to be refactored anyway.

noahdietz (approved):

PR #9694, opened 2024-04-03
feat(auth/oauth2adapt): add helpers for working with credentials types

codyoss:


codyoss:
This will require auth to be released first and the go.mod updated

quartzmo (approved):

PR #9711, opened 2024-04-04
feat(pubsub): support publisher compression

hongalex:
Similar to [Java](https://github.com/googleapis/java-pubsub/pull/1000), this PR adds gRPC compression support to the `Publish` method specifically.

This does not add change the number of bytes that Pub/Sub charges for, as it is compressed and decompressed in transit. Currently, only gzip is supported.

internal issue: 226594738


pubsub/topic.go
```diff
@@ -134,6 +142,8 @@ var DefaultPublishSettings = PublishSettings{
 		MaxOutstandingBytes:    -1,
 		LimitExceededBehavior:  FlowControlIgnore,
 	},
+	EnableCompression:         false,
+	CompressionBytesThreshold: 240,
```
alvarowolfx, line 146: is this documented somewhere as recommended setting ?

pubsub/topic.go
```diff
@@ -897,11 +909,21 @@ func (t *Topic) publishMessageBundle(ctx context.Context, bms []*bundledMessage)
 			opt.Resolve(&settings)
 		}
 		r := &publishRetryer{defaultRetryer: settings.Retry()}
-		res, err = t.c.pubc.Publish(ctx, &pb.PublishRequest{
-			Topic:    t.name,
-			Messages: pbMsgs,
-		}, gax.WithGRPCOptions(grpc.MaxCallSendMsgSize(maxSendRecvBytes)),
-			gax.WithRetry(func() gax.Retryer { return r }))
```
alvarowolfx, line 904: I think readability here can be improved to actually conditionally build the gax.CallOptions depending of the `PublishSettings` and keep the call to `.Publish` appearing only once. Would be something like:

```go
gaxOpts := []gax.CallOption{
	gax.WithGRPCOptions(grpc.MaxCallSendMsgSize(maxSendRecvBytes)),
	gax.WithRetry(func() gax.Retryer { return r }),
}
if t.PublishSettings.shouldCompress(batchSize) {
	gaxOpts = append(gaxOpts, gax.WithGRPCOptions(grpc.UseCompressor(gzip.Name)))
}
res, err = t.c.pubc.Publish(ctx, &pb.PublishRequest{
	Topic:    t.name,
	Messages: pbMsgs,
}, gaxOpts...)
```		

Also separate the logic into the PublishSettings struct, so it can be easily tweaked later.
```go
func (ps *PublishSettings) shouldCompress(batchSize int) bool {
	return ps.EnableCompression && batchSize > ps.CompressionBytesThreshould
}
```

alvarowolfx (approved):
I think it looks good to me feature wise, but I added a comment on how to improve the call to `.Publish` to avoid having two code paths calling it.

PR #9714, opened 2024-04-05
fix(firestore): Move createIndexes calls

bhshkh:
Fixes https://github.com/googleapis/google-cloud-go/issues/9039

**Issue**: createIndexes is a setup step that is executed before running any of the tests. It is failing intermittently with `context deadline exceeded error`

**Cause**: Index creation takes too long

**Fix**: 
1. Create lesser number of indexes
2. Move index creation to the tests that use those indexes so that rest of the tests can run even if the index creation fails 
3. Set higher context deadline


bhshkh:
Do not merge until release freeze ends mid April

gkevinzheng (approved):

PR #9726, opened 2024-04-08
fix(bigquery): add jobRateLimitExceeded in jobRetryReasons

atzoum:
`jobRateLimitExceeded` should be added in `jobRetryReasons` so that these errors are automatically retried by the client. 
Other bigquery libraries, such as [python-bigquery](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/retry.py#L76) do so already.


alvarowolfx (dismissed):
Can we add tests for this retry reason here ? Other than that, looks good to me
https://github.com/googleapis/google-cloud-go/blob/6a7cd4f70373fe7c60dcba12636a3d92617e7b66/bigquery/bigquery_test.go#L28

shollyman (approved):
Thanks for the contribution.

Your PR has reminded me that we don't properly exercise job retry behaviors in the existing unit tests, so I've opened 9751 against myself to add more testing surface.

PR #9731, opened 2024-04-09
feat(vertexai/genai): constrained decoding

jba:
Expose FunctionCallingMode, which lets the user configure
how the model issues function calls.

By setting the new field `Model.ToolConfig`, the user can select whether
the model can call any provided function, only those in a provided list,
or no functions.

Regenerating also brought along a few other new types and values,
some of which required configuration.

### DIFF
--- a/vertexai/genai/aiplatformpb_veneer.gen.go
+++ b/vertexai/genai/aiplatformpb_veneer.gen.go
@@ -24,13 +24,14 @@ import (
 	"cloud.google.com/go/vertexai/internal/support"
 )
 
-// Blob contains raw media bytes.
+// Blob contains content blob.
 //
-// Text should not be sent as raw bytes, use the 'text' field.
+// It's preferred to send as [text][google.cloud.aiplatform.v1beta1.Part.text]
+// directly rather than raw bytes.
 type Blob struct {
 	// Required. The IANA standard MIME type of the source data.
 	MIMEType string
-	// Required. Raw bytes for media formats.
+	// Required. Raw bytes.
 	Data []byte
 }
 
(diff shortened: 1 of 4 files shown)


vertexai/genai/client.go
```diff
@@ -95,6 +95,7 @@ type GenerativeModel struct {
 	GenerationConfig
 	SafetySettings []*SafetySetting
 	Tools          []*Tool
+	ToolConfig     *ToolConfig
```
eliben, line 98: This requires a docstring and an example.

vertexai/genai/client_test.go
```diff
@@ -234,32 +234,52 @@ func TestLive(t *testing.T) {
...
-		}
-		if g, c := funcall.Args["location"], "New York"; !strings.Contains(g.(string), c) {
-			t.Errorf(`FunctionCall.Args["location"]: got %q, want string containing %q`, g, c)
-		}
-		res, err = session.SendMessage(ctx, FunctionResponse{
-			Name: weatherTool.FunctionDeclarations[0].Name,
-			Response: map[string]any{
-				"weather_there": "cold",
-			},
+		t.Run("funcall-auto", func(t *testing.T) {
```
eliben, line 237: is this automated function calling?

jba: No, it's using the "Auto" value of FunctionCallingMode, which is the default. I removed "-auto" to avoid confusion.

vertexai/genai/client.go
jba: Did both.

vertexai/genai/aiplatformpb_veneer.gen.go
```diff
@@ -371,6 +379,69 @@ func (FunctionCall) fromProto(p *pb.FunctionCall) *FunctionCall {
...
+	// FunctionCallingUnspecified means unspecified function calling mode. This value should not be used.
+	FunctionCallingUnspecified FunctionCallingMode = 0
+	// FunctionCallingAuto means default model behavior, model decides to predict either a function call
+	// or a natural language repspose.
+	FunctionCallingAuto FunctionCallingMode = 1
+	// FunctionCallingAny means model is constrained to always predicting a function call only.
+	// If "allowed_function_names" are set, the predicted function call will be
+	// limited to any one of "allowed_function_names", else the predicted
+	// function call will be any one of the provided "function_declarations".
+	FunctionCallingAny FunctionCallingMode = 2
```
eliben, line 425: "Any" is a really confusing alternative to "Auto", but I suppose we're bound to the API with this naming?

Deleplace:
TIL "Example functions without output comments are compiled but not executed."

I can hardcode my projectID/location/model, but then what is the proper way to run the 6 Example funcs?

Is it expected that the tests in `client_test.go` are executed in the CI/CD pipeline, but the tests in `example_test.go` are not?

vertexai/genai/aiplatformpb_veneer.gen.go
```diff
@@ -24,13 +24,14 @@ import (
 	"cloud.google.com/go/vertexai/internal/support"
 )
 
-// Blob contains raw media bytes.
+// Blob contains content blob.
 //
-// Text should not be sent as raw bytes, use the 'text' field.
+// It's preferred to send as [text][google.cloud.aiplatform.v1beta1.Part.text]
```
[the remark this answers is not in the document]
jba: No. Fixed.

vertexai/genai/aiplatformpb_veneer.gen.go
jba: Yes, I wouldn't want to depart from the API.

jba:
@Deleplace Yes, there's no way to actually run the examples as tests.

PR #9735, opened 2024-04-09
test(monitoring): remove old smoke test

codyoss:
This smoke test seems to fail from time to time but I can't repo this locally with 100 runs. We don't really do this kind of testing for generated clients any more and this client is also never updated it was v2'ed, old-style, many years ago. Removing test.

Fixes: #8257


noahdietz:
Looks like the module needs a tidy now

noahdietz (approved):

PR #9736, opened 2024-04-09
feat(vertexai/genai): add SystemInstruction

jba:
Expose GenerateContentRequest.SystemInstruction.

As usual, put it on the Model, and use the Model field to construct
each request.

### DIFF
--- a/vertexai/genai/example_test.go
+++ b/vertexai/genai/example_test.go
@@ -18,12 +18,10 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"os"
 
 	"cloud.google.com/go/vertexai/genai"
 
 	"google.golang.org/api/iterator"
-	"google.golang.org/api/option"
 )
 
 // Your GCP project
@@ -57,7 +55,9 @@ func ExampleGenerativeModel_GenerateContent() {
 // for the complete set of configuration options.
 func ExampleGenerativeModel_GenerateContent_config() {
 	ctx := context.Background()
-	client, err := genai.NewClient(ctx, option.WithAPIKey(os.Getenv("GEMINI_API_KEY")))
+	const projectID = "YOUR PROJECT ID"
+	const location = "GCP LOCATION"
+	client, err := genai.NewClient(ctx, projectID, location)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/vertexai/genai/example_test.go
+++ b/vertexai/genai/example_test.go
@@ -18,10 +18,12 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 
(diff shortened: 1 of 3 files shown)


eliben (approved):
It'd be nice to have an example for this too.

Particularly because this is called "system prompt" everywhere else, and this may make it difficult for users to discover. For example, there's an issue on generative-ai-go asking for it: https://github.com/google/generative-ai-go/issues/31

jba:
Example forthcoming.

jba:
I added a test and an example that shows various config options.
But I'm marking this a draft because the feature doesn't seem to be implemented.

Deleplace:
> But I'm marking this a draft because the feature doesn't seem to be implemented.

The [Python sample](https://cloud.google.com/vertex-ai/generative-ai/docs/samples/generativeaionvertexai-gemini-system-instruction) for system instructions with Gemini 1.5 Pro is working for me. I'll see if can can get similar results with the Go sdk.

Deleplace:
I'm getting compile errors at `example_test.go`, @jba can you PTAL?

vertexai/genai/example_test.go
```diff
@@ -51,6 +51,31 @@ func ExampleGenerativeModel_GenerateContent() {
 	printResponse(resp)
 }
 
+// This example shows how to a configure a model. See [GenerationConfig]
+// for the complete set of configuration options.
+func ExampleGenerativeModel_GenerateContent_config() {
+	ctx := context.Background()
+	client, err := genai.NewClient(ctx, option.WithAPIKey(os.Getenv("GEMINI_API_KEY")))
```
eliben, line 58: Shouldn't this be with project/location and not key?

Deleplace:
> > I'm getting compile errors at `example_test.go`, @jba can you PTAL?
> 
> Is it a vet error? What does it say?

```
% go test     
# cloud.google.com/go/vertexai/genai_test [cloud.google.com/go/vertexai/genai.test]
./example_test.go:58:38: undefined: option
./example_test.go:58:38: not enough arguments in call to genai.NewClient
	have (context.Context, unknown type)
	want (context.Context, string, string, ...option.ClientOption)
./example_test.go:58:56: undefined: os
FAIL	cloud.google.com/go/vertexai/genai [build failed]
```

And after fixing the imports:
```
./example_test.go:60:38: not enough arguments in call to genai.NewClient
```

The errors are trivial but I suspect we don't have jba's latest code.

jba:
Sorry for the wait. I fixed the compilation problem and the feature now works with gemini-1.0-pro. Removing draft status.

vertexai/genai/example_test.go
jba: Yes, fixed.

Issue #4273, opened 2021-06-16
all: switch to google.golang.org/protobuf


neild:
The most recent version of the Go protobuf API is the `"google.golang.org/protobuf"` module, released about a year ago.

Most of the Cloud client library packages are using the older `"github.com/golang/protobuf"` module. As of the time I write this, those packages are all using `v1.5.2` of this module, which is (mostly) a thin wrapper around the newer module.

It would be nice if the Cloud client libraries used the newer module directly.

In most cases, this can be accomplished by simply changing the import path to the newer module. Note that this doesn't change the underlying implementation being used at all, aside from removing one layer of wrapping.

In some cases, types from the "github.com/golang/protobuf" module appear in the exported API of a package. (In particular: `proto.Message`.) These places will need to continue to use the older module. This issue is about internal, non-externally-visible usage of the protobuf API.

codyoss:
Related regen: https://github.com/googleapis/go-genproto/pull/623

trakhimenok:
@codyoss any time soon to fix this one? For example firestore still references both github (obsolete) & google modules.

trakhimenok:
New year, new hope to get this issue closed...

codyoss:
Opened a PR to remove this as a direct dep across the whole repo.

PR #9743, opened 2024-04-10
chore: remove all uses of github.com/golang/protobuf

codyoss:
Fixes: #4273


datastore/mock_test.go
```diff
@@ -120,8 +120,8 @@ func (s *mockServer) popRPC(gotReq proto.Message) (interface{}, error) {
 
 		if !proto.Equal(gotReq, ri.wantReq) {
 			return nil, fmt.Errorf("mockServer: bad request\ngot:\n%T\n%s\nwant:\n%T\n%s",
-				gotReq, proto.MarshalTextString(gotReq),
-				ri.wantReq, proto.MarshalTextString(ri.wantReq))
+				gotReq, gotReq,
+				ri.wantReq, ri.wantReq)
```
noahdietz, line 124: I'm not sure that the protos will render via `%s` all that well, at least not in the textproto format as was done before...is it worth just adding the https://pkg.go.dev/google.golang.org/protobuf@v1.33.0/encoding/prototext usage?

Not a blocker though...

firestore/conformance_test.go
```diff
@@ -191,7 +190,7 @@ func runTest(test *pb.Test, c *Client, srv *mockServer) error {
 		got, checkErr := q.toProto()
 		if err := check(checkErr, typedTestcase.Query.IsError); err == nil && checkErr == nil {
 			if want := typedTestcase.Query.Query; !proto.Equal(got, want) {
-				return fmt.Errorf("got:  %s\nwant: %s", proto.MarshalTextString(got), proto.MarshalTextString(want))
+				return fmt.Errorf("got:  %s\nwant: %s", got, want)
```
noahdietz, line 193: ditto re: using `prototext` marshaler...

firestore/document.go
```diff
@@ -299,23 +298,11 @@ func newDocumentSnapshot(ref *DocumentRef, proto *pb.Document, c *Client, readTi
...
-			return nil, err
-		}
-		d.CreateTime = ts
-		ts, err = ptypes.Timestamp(proto.UpdateTime)
-		if err != nil {
-			return nil, err
-		}
-		d.UpdateTime = ts
+		d.CreateTime = proto.CreateTime.AsTime()
+		d.UpdateTime = proto.UpdateTime.AsTime()
```
noahdietz, line 302: I think `AsTime` will "just work" if the `*Timestamp` is `nil` based on [the implementation](https://github.com/protocolbuffers/protobuf-go/blob/v1.33.0/types/known/timestamppb/timestamp.pb.go#L199-L201) (which uses accessors for the fields so that they can return `0` if the receiver/proto is `nil`), but we really ought to be using accessors for proto fields e.g. `proto.GetCreateTime().AsTime()`.

Do you feel comfortable making those changes here? 

There are a few other examples of this as well and I won't comment on each until we've discussed.

firestore/options.go
```diff
@@ -61,12 +61,8 @@ func LastUpdateTime(t time.Time) Precondition { return lastUpdateTime(t) }
 type lastUpdateTime time.Time
 
 func (u lastUpdateTime) preconditionProto() (*pb.Precondition, error) {
-	ts, err := ptypes.TimestampProto(time.Time(u))
-	if err != nil {
-		return nil, err
-	}
```
noahdietz, line 67: I'm *guessing* this does some validation of the time, but I could be wrong. Is it worth using the `Timestamp.IsValid` api to retain that?

Could be wrong though.

firestore/conformance_test.go
noahdietz: resolving as a dupe, there are multiple instances and I don't  want to comment on each.

datastore/mock_test.go
codyoss: will do .

firestore/document.go
codyoss: I can at least update for everywhere I have touched. Switching to accessors in general might be a larger project that likely greatly affects all veneers

noahdietz: that's fair! I think that's right considering "type" of access is changing with your updates (the field is being "dereferenced" directly instead of just passed to a function that may handle the `nil` case)

firestore/options.go
codyoss: added back with checks

PR #9747, opened 2024-04-11
fix(storage): disable gax retries for gRPC

tritone:
All gRPC ops are retried as needed at the handwritten layer. Disable gax retries.


tritone:
It looks like this exposed a bunch of conformance test failures; I'm going to fix and update.

tritone:
> It looks like this exposed a bunch of conformance test failures; I'm going to fix and update.

Okay, this was only because of a change in how things that are unimplemented in the testbench are handled in the retry test framework. I fixed this and all tests now pass.

storage/client_test.go
```diff
@@ -1342,6 +1347,47 @@ func TestObjectConditionsEmulated(t *testing.T) {
...
+		et := emulatorTest{T: t, name: "testRetryNever", resources: resources{},
+			host: endpoint}
+		et.create(map[string][]string{"storage.buckets.get": {"return-503"}}, transport)
+		ctx = callctx.SetHeaders(ctx, "x-retry-test-id", et.id)
+		_, err = client.GetBucket(ctx, attrs.Name, nil, withRetryConfig(&retryConfig{policy: RetryNever}))
+
+		var ae *apierror.APIError
+		if errors.As(err, &ae) {
+			if ae.GRPCStatus().Code() == codes.Unavailable || ae.HTTPCode() == 503 {
+				return
```
BrennaEpp, line 1383: We could probably just invert the condition and error here instead of returning

BrennaEpp (approved):
One suggestion that doesn't change the functionality, otherwise LGTM.

We should also look into dropping `withGAXOptions`.

storage/client_test.go
tritone: Sure, I'll change that.

PR #9756, opened 2024-04-12
chore(storage): remove superfluous options

tritone:
Remove gax storage option which is unused. Also, remove default options which are duplicated between constructors.

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -232,7 +232,6 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 // You may configure the client by passing in options from the [google.golang.org/api/option]
 // package.
 func NewGRPCClient(ctx context.Context, opts ...option.ClientOption) (*Client, error) {
-	opts = append(defaultGRPCOptions(), opts...)
 	tc, err := newGRPCStorageClient(ctx, withClientOptions(opts...))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


storage/storage.go
```diff
@@ -232,7 +232,6 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 // You may configure the client by passing in options from the [google.golang.org/api/option]
 // package.
 func NewGRPCClient(ctx context.Context, opts ...option.ClientOption) (*Client, error) {
-	opts = append(defaultGRPCOptions(), opts...)
```
BrennaEpp, line 235: What's your reasoning for removing them here vs in `newGRPCStorageClient`? Looking through the code, I'm 95% sure that it doesn't matter which of the two gets removed in terms of functionality, but curious to know if there was logic behind which we're keeping.

BrennaEpp (commented):
LGTM, just a question

storage/storage.go
tritone: I originally did that (and moved the gax option to the constructor) but tests started failing; I then realized that the emulator tests call `NewGRPCStorageClient` directly. We could also pass the options to the test initialization but it seemed simpler just to put everything here.

BrennaEpp (approved):

Issue #9017, opened 2023-11-16
bigquery: support for RANGE type


shollyman:
This issue tracks the work to enable support for the RANGE datatype in the BigQuery Go client libraries.

alvarowolfx:
I'll check if the RANGE type is supported when using Storage Read API acceleration and add support for it if needed.

shollyman:
Released as https://github.com/googleapis/google-cloud-go/releases/tag/bigquery/v1.61.0

Marking this work done.

PR #9762, opened 2024-04-12
feat(bigquery): RANGE support for basic data movement

shollyman:
This PR adds a new RangeValue type for conveying parsed range start/end
values, and adds the requisite plumbing for data movement.

Supported interactions:
* streaming insert
* parameter binding
* type inference (docs indicate ambiguous type for range element)
* data reads (non-accelerated)

A future PR will add storage read acceleration support.


bigquery/rangevalue.go
```diff
@@ -0,0 +1,28 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bigquery
+
+// RangeValue represents a continuous RANGE of values of a given element
+// type.  The supported element types for RANGE are currently the BigQuery
+// TIME, TIMESTAMP, and DATE types.
+type RangeValue struct {
```
alvarowolfx, line 20: can we add a `Type FieldType` field here to show if the RangeValue is a `TIME`, `TIMESTAMP` or `DATE` here ? This way the `inferSchema` method can also fully infer the type.

shollyman: This doesn't actually help things.  Infer consumes the go type information, it doesn't look at the current values.  In order to make this tractable you'd need N versions of the type, one for each element type.  That didn't feel like a great choice either.

bigquery/params.go
```diff
@@ -344,6 +348,21 @@ func paramType(t reflect.Type, v reflect.Value) (*bq.QueryParameterType, error)
 		return geographyParamType, nil
 	case typeOfNullJSON:
 		return jsonParamType, nil
+	case typeOfRangeValue:
+		iv := v.Interface().(*RangeValue)
+		// at least one of start,end must be populated.  grab the first non-nil value.
+		element := iv.Start
```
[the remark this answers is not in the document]
shollyman: Yeah, I think the guidance will end up need to be something like "if you have completely unbounded ranges declare them explicitly as a QueryParam, not via detection".  It's not unlike the difference between NUMERIC and BIGNUMERIC when it comes to big.Rat values.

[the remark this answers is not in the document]
shollyman: Added a testing to show declaration via a QueryParameterValue

bigquery/params.go
```diff
@@ -492,6 +511,28 @@ func paramValue(v reflect.Value) (*bq.QueryParameterValue, error) {
 	case typeOfIntervalValue:
 		res.Value = IntervalString(v.Interface().(*IntervalValue))
 		return res, nil
+	case typeOfRangeValue:
```
Linchin, line 514: Should we add a null value handling for range, together with the rest null values (at row 438)?

Linchin: nevermind, I think it's handled below (row 638) similar to ARRAY and STRUCT.

alvarowolfx (approved):

bigquery/rangevalue.go
shollyman: Closing this out as we're missing value information for inference entirely.

Issue #9720, opened 2024-04-05
storage: timeout errors are not always wrapped when a func was retried


BrennaEpp:
Errors are wrapped when calling the `run` function in `invoke.go`, when the error was retried and there was a context timeout/deadline exceeded.
For example, this is what an error looks like after the retry was exhausted with a timeout on obj.Attrs:
`retry failed with context deadline exceeded; last error: googleapi: Error 400: ...`

However, there are (at least) 2 cases when those errors are not wrapped:
1. When `MaxAttempts` is reached; this is what an error looks like after the number of max attempts was exhausted:
`googleapi: Error 400: ...`

2. On object uploads with HTTP; This is what an error looks like after retry timeout exhausted on object write:
`context deadline exceeded`

Here is repro code:
```go
		var err error
		// This bucket name is too short, which will produce a 400 error.
		obj := client.Bucket("a").Object("content")

		// Always retry, so no calls will succeed on this handle.
		// Retries everything, including the 400 error we get because of the bucket name.
		obj = obj.Retryer(WithPolicy(RetryAlways),
			WithErrorFunc(func(err error) bool { return true }),
		)

		// Timeout on obj.Attrs - this error is wrapped correctly
		timeoutCtx, cancel := context.WithTimeout(ctx, time.Second*1)
		defer cancel()

		_, err = obj.Attrs(timeoutCtx)
		if err == nil {
			t.Fatalf("obj.Attrs should have failed but succeeded")
		}
		fmt.Printf("obj.Attrs times out: \n%v\n-\n", err)

		// Max attempts on obj.Attrs - this error is not wrapped
		objMaxAttempt := obj.Retryer(WithMaxAttempts(4))

		_, err = objMaxAttempt.Attrs(ctx)
		if err == nil {
			t.Fatalf("obj.Attrs should have failed but succeeded")
		}
		fmt.Printf("obj.Attrs max attempts run out: \n%v\n-\n", err)

		// Timeout on obj write - this error is not wrapped
		timeoutCtx, cancel = context.WithTimeout(ctx, time.Second*5)
		defer cancel()

		w := obj.NewWriter(timeoutCtx)
		_, err = io.Copy(w, strings.NewReader("It was the best of times, it was the worst of times."))
		if err != nil {
			t.Errorf("io.Copy: %v", err)
		}

		err = w.Close()
		if err == nil {
			t.Fatalf("w.Close should have failed but succeeded")
		}
[...]

PR #9767, opened 2024-04-15
fix(storage): wrap error when MaxAttempts is hit

tritone:
Wrap the error when retries are cut off by hitting the configured value for MaxAttempts. This makes it easier to verify that retries occurred.

Also adds emulator tests verifying that wrapping occurs as expected for timeout errors and MaxAttempts.

Updates #9720


storage/client_test.go
```diff
@@ -1351,31 +1352,57 @@ func TestObjectConditionsEmulated(t *testing.T) {
...
+		testID := createRetryTest(t, project, bucket, client, instructions)
+		ctx = callctx.SetHeaders(ctx, "x-retry-test-id", testID)
+		_, err := client.GetBucket(ctx, bucket, nil, withRetryConfig(&retryConfig{policy: RetryNever}))
 
-		attrs, err := client.CreateBucket(ctx, project, bucket, &BucketAttrs{}, nil)
-		if err != nil {
-			t.Fatalf("creating bucket: %v", err)
+		var ae *apierror.APIError
+		if errors.As(err, &ae) {
+			// We espect a 503/UNAVAILABLE error. For anything else including a nil
```
BrennaEpp, line 1362:
```suggestion
			// We expect a 503/UNAVAILABLE error. For anything else including a nil
```

storage/client_test.go
```diff
@@ -1351,31 +1352,57 @@ func TestObjectConditionsEmulated(t *testing.T) {
...
+		instructions := map[string][]string{"storage.buckets.get": {"return-503", "return-503", "return-503", "return-503", "return-503"}}
+		testID := createRetryTest(t, project, bucket, client, instructions)
+		ctx = callctx.SetHeaders(ctx, "x-retry-test-id", testID)
+		ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
+		defer cancel()
+		_, err := client.GetBucket(ctx, bucket, nil, idempotent(true))
+
+		var ae *apierror.APIError
+		if errors.As(err, &ae) {
+			// We espect a 503/UNAVAILABLE error. For anything else including a nil
```
BrennaEpp, line 1384:
```suggestion
			// We expect a 503/UNAVAILABLE error. For anything else including a nil
```

storage/client_test.go
```diff
@@ -1385,9 +1412,44 @@ func TestRetryNeverEmulated(t *testing.T) {
 				t.Errorf("GetBucket: got unexpected error %v; want 503", err)
 			}
 		}
+		// Error should be wrapped so it indicates that MaxAttempts has been reached.
+		if got, want := err.Error(), "retry failed after 3 attempts"; !strings.Contains(got, want) {
+			t.Errorf("got error: %q, want to contain: %q", got, want)
+		}
```
BrennaEpp, line 1418: Do we need to delete the retry test resource in the testbench after finishing?

BrennaEpp (commented):
Awesome!

storage/client_test.go
tritone: I don't think it matters too much given that the emulator is torn down between test runs, but on the other hand I might as well -- will update.

BrennaEpp (approved):

PR #9795, opened 2024-04-17
feat(bigquery): RANGE support when reading Arrow format

alvarowolfx:
Allows usage of Storage Read acceleration with queries containing the new RANGE data type. This PR adds code to parse the Arrow representation of RANGEs. 


bigquery/arrow.go
```diff
@@ -272,6 +272,21 @@ func convertArrowValue(col arrow.Array, i int, ft arrow.DataType, fs *FieldSchem
 		arr := col.(*array.Struct)
 		nestedValues := []Value{}
 		fields := ft.(*arrow.StructType).Fields()
+		if fs.Type == RangeFieldType {
+			rangeFieldSchema := &FieldSchema{
+				Type: fs.RangeElementType.Type,
```
shollyman, line 277: do we need to nil check RangeElementType? Are there cases where the field schema is user supplied and possible incomplete?

shollyman (approved):
One question about nil checking, other than that LGTM.

bigquery/arrow.go
alvarowolfx: in the Read acceleration path, the Schema/FieldSchema comes from the table metadata, so it's never user supplied.
* https://github.com/googleapis/google-cloud-go/blob/62d6ecf62a466b0b87cb30f34dd09f3b9e29828c/bigquery/storage_iterator.go#L52
* https://github.com/googleapis/google-cloud-go/blob/62d6ecf62a466b0b87cb30f34dd09f3b9e29828c/bigquery/storage_iterator.go#L68

PR #9836, opened 2024-04-18
feat(bigquery/storage/managedwriter/adapt): add RANGE support to adapt

shollyman:
This PR adds support for schema conversions and proto conversions to the adapt subpackage.  For proto conversion, we build a static name for range message types rather than using the normal first-encountered-message, as the schema for a range field will also be the consistent for a given element type.

This PR also moves adapt to using subtests for table-driven testing to make it easier to isolate specific test cases.


shollyman:
The interesting question here is how to register the message types for the range values when converting a table schema to the descriptor proto.  At a high level, there's three options:

* publish a set of messages similar to the well-known types, and only add the reference.  This makes compilation more awkward, as these types aren't general enough to warrant include in the common protobuf libraries, nor even in the google API type extensions.

* [current implementation] generate type names independently of where they occur in the message hierarchy.  This makes it easier to recognize a range message via the TypeName, since we lose the ability to recognize the RANGE purely from the descriptor.

* inject range messages into the common dependency cache as we're walking the referenced messages.  This consolidates the dependency cache a bit, but it means the TypeName of the message will be based on the first encountered reference (something like `root.nested_message.range_field`).  It also means that any user messages with a similar two-field schema (start/end) could potentially be re-used in the cache.

alvarowolfx (approved):

PR #9838, opened 2024-04-19
[WIP] feat(storage/transfermanager): prototype

tritone:
sketch of an interface for downloads


storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,149 @@
...
+	// Timeout for a single operation (including all retries).
+	perOperationTimeout time.Duration
+
+	// others, including opts that apply to uploads.
+}
+
+// etc
+
+type Downloader struct {
+	ctx             context.Context
```
BrennaEpp, line 56: Do we want this overarching context, or just individual contexts for each download?

storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,149 @@
...
+	// Required fields
+	Bucket      string
+	Source      string
+	Destination io.WriterAt
+
+	// Optional fields
+	Generation     *int64
+	Conditions     *storage.Conditions
+	EncryptionKey  []byte
+	Offset, Length int64 // if specified, reads only a range.
```
BrennaEpp, line 84: So this will be different than the regular behaviour (I believe normally a length of 0 will work at least for HTTP but will not read the whole object), or do I make these (at least length) pointers as well?

PR #9999, opened 2024-04-19
fix(auth): add internal opt to skip validation on transports

codyoss:
Updates: #9823

### DIFF
--- a/auth/grpctransport/grpctransport.go
+++ b/auth/grpctransport/grpctransport.go
@@ -154,7 +154,8 @@ type InternalOptions struct {
 	// DefaultScopes specifies the default OAuth2 scopes to be used for a
 	// service.
 	DefaultScopes []string
-	// SkipValidation bypasses validation on Options.
+	// SkipValidation bypasses validation on Options. It should only be used
+	// internally for clients that needs more control over their transport.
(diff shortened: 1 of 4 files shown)


codyoss:
Will also require some change in apiary repo I have staged

auth/grpctransport/grpctransport.go
```diff
@@ -151,6 +154,8 @@ type InternalOptions struct {
 	// DefaultScopes specifies the default OAuth2 scopes to be used for a
 	// service.
 	DefaultScopes []string
+	// SkipValidation bypasses validation on Options.
+	SkipValidation bool
```
noahdietz, line 158: Can we include a little bit about when or why we'd use this? Seems a little bit like a big red button :)

codyoss: Added a little more. Note it is in the InternalOptions struct which we don't want to encourge users to use. Corresponds to https://pkg.go.dev/google.golang.org/api/option/internaloption#SkipDialSettingsValidation.

noahdietz: Yep totally, I just personally wasn't sure what this was for so the comment & your link helps, thanks.

Issue #10009, opened 2024-04-20
civil: add Compare method to Date, Time and DateTime


shota3506:
**Is your feature request related to a problem? Please describe.**

This proposal suggests adding `Compare` method to civil.Date, civil.Time and civil.DateTime

Some structs in the standard package have `Compare` method which returns a integer value of either -1, +1, or 0.
https://pkg.go.dev/time#Time.Compare
https://pkg.go.dev/net/netip@master#Addr.Compare

It looks like the standard way to define ordered objects in Go (https://github.com/golang/go/issues/50770#issuecomment-1028214622), and it is useful when we sort such objects by slices.SortFunc.

**Describe the solution you'd like**
I'll open a PR for this issue soon.

**Describe alternatives you've considered**
Of course, we can judge the order of civil objects by using exiting `After` and `Before`.
If there are opinions that it is unnecessary, I will respect them.

**Additional context**

noahdietz:
Just reviewed the PR.

Thanks for the idea/contribution.

noahdietz:
@shota3506 sorry, need to hold onto this for a little bit. We want to upgrade to Go 1.20 soon. Please reopen your PR after the reversion goes through and we will merge it as soon as we upgrade!

shota3506:
@noahdietz Thank you for your review and comment!
I  haven't go directive in this project sorry...
I'll wait for the version update 👍

shota3506:
@noahdietz 

Hi ✋ 
The minimum version has been upgraded to 1.20, right?

I've reopened the PR https://github.com/googleapis/google-cloud-go/pull/10193
Please check when you have time. Thanks.

noahdietz:
Yes @shota3506 good eye :) Thanks!

PR #10010, opened 2024-04-20
feat(civil): add Compare method to Date, Time, and DateTime

shota3506:
implementation for https://github.com/googleapis/google-cloud-go/issues/10009


noahdietz (approved):
Thanks for the contribution, this seems great to have.

PR #10018, opened 2024-04-22
fix(auth/credentials): error on bad file name if explicitly set

codyoss:
Only want our fall-through logic to work for ADC specified behaviour, not explicit credential options.

Fixes: #9809

### DIFF
--- a/auth/credentials/detect.go
+++ b/auth/credentials/detect.go
@@ -76,7 +76,10 @@ func DetectDefault(opts *DetectOptions) (*auth.Credentials, error) {
 	if opts.CredentialsJSON != nil {
 		return readCredentialsFileJSON(opts.CredentialsJSON, opts)
 	}
-	if filename := credsfile.GetFileNameFromEnv(opts.CredentialsFile); filename != "" {
+	if opts.CredentialsFile != "" {
+		return readCredentialsFile(opts.CredentialsFile, opts)
+	}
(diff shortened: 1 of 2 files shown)


auth/credentials/detect.go
```diff
@@ -76,7 +76,10 @@ func DetectDefault(opts *DetectOptions) (*auth.Credentials, error) {
 	if opts.CredentialsJSON != nil {
 		return readCredentialsFileJSON(opts.CredentialsJSON, opts)
 	}
-	if filename := credsfile.GetFileNameFromEnv(opts.CredentialsFile); filename != "" {
+	if opts.CredentialsFile != "" {
+		return readCredentialsFile(opts.CredentialsFile, opts)
+	}
+	if filename := os.Getenv(credsfile.GoogleAppCredsEnvVar); filename != "" {
```
quartzmo, line 82: Why does the new usage of `readCredentialsFile` on line 80 return a non-nil `err`, while the existing usage on 83-85 discards a non-nil `err`? Both use a file path provided by the user, so why should they behave differently?

codyoss: It is because the env-var is a part of ADC which, at least in Go, historically had this fallthrough behaviour. The same is not true for user provided compile-time options. This better aligns with the old behavior, as seen by the linked failing test where it feel through when the test was not expecting it to.

quartzmo: SGTM, thanks.

quartzmo (approved):

PR #10020, opened 2024-04-22
chore: revert "feat(civil): add Compare method to Date, Time, and DateTime"

noahdietz:
Reverts googleapis/google-cloud-go#10010

`time` `Compare` was added in Go 1.20 and we still need to support Go 1.19. Temporarily reverting this.

### DIFF
--- a/civil/civil.go
+++ b/civil/civil.go
@@ -111,17 +111,6 @@ func (d Date) After(d2 Date) bool {
 	return d2.Before(d)
 }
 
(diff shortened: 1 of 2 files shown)


noahdietz:
@shota3506 sorry, need to hold onto this for a little bit. We want to upgrade to Go 1.20 soon. Please reopen your PR after this reversion goes through and we will merge it as soon as we upgrade!

codyoss (approved):

PR #10024, opened 2024-04-22
feat(bigquery/storage/managedwriter): skeleton for destination ordering

shollyman:
This PR contains the skeleton for implementing per-destination ordering, defining a new pendingQueue abstraction and alternative recv processor.

It does not wire the new implementation into the existing library at this point, which would be done in followup PRs.


codyoss:
@shollyman is this still active or can it be closed?

PR #10030, opened 2024-04-23
feat(logging): OpenTelemetry trace/span ID integration for Go logging library

### LINKED ISSUE #9302: logging: OpenTelemetry spans are ignored as parents for LogRecords
**Client**

logging: https://github.com/googleapis/google-cloud-go/tree/main/logging

**Current behavior**

The current implementation of populateTraceInfo always uses a remote span context as the parent
https://github.com/googleapis/google-cloud-go/blob/ac16144215f844294c29063776f2f116e14ecebb/logging/logging.go#L807-L838

By "Remote span context", I mean a trace ID and span ID that originate from _outside_ the application.  The code above always uses a remote span context because it explicitly parses the trace id and span id from HTTP headers.

When a user has instrumented their application with OpenTelemetry, the trace ID and span ID have _already_ been parsed from the HTTP headers, and used to start a span.  If they are using the [otelhttp](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/instrumentation/net/http/otelhttp/v0.47.0/instrumentation/net/http/otelhttp) library, it:

1. [extracts the context from req.Header into a golang context.Context](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/044e09ac8bfe0d5a60510778017bec57413243d9/instrumentation/net/http/otelhttp/handler.go#L144)
2. [uses the context to start a span](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/044e09ac8bfe0d5a60510778017bec57413243d9/instrumentation/net/http/otelhttp/handler.go#L171)
3. [updates the context.Context on the req using req.WithContext](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/044e09ac8bfe0d5a60510778017bec57413243d9/instrumentation/net/http/otelhttp/handler.go#L225)

**Expected behavior**

If the user is using opentelemetry instrumentation, we want to use the traceID and spanID of the span that OpenTelemetry created.  This way, when a user looks for logs associated with that span, the logs produced by this logging library will appear.  The traceID and spanID are easy to get since otelhttp stores them in `req.Context()`.

We can do something close to this:

```golang

func populateTraceInfo(e *Entry, req *http.Request) bool {
	if req == nil {
		if e.HTTPRequest != nil && e.HTTPRequest.Request != nil {
			req = e.HTTPRequest.Request
		} else {
			return false
		}
	}
        // New handing code
        otelSpanContext := trace.SpanContextFromContext(req.Context())
	if otelSpanContext.IsValid() {
	        e.Trace := otelSpanContext.TraceID()
	        e.SpanID := otelSpanContext.SpanID()
                e.TraceSampled = e.TraceSampled | otelSpanContext.IsSampled()
                return true
	}
        // end of new code
	header := req.Header.Get("Traceparent")
         if header != "" { 
 		// do not use traceSampled flag defined by traceparent because 
 		// flag's definition differs from expected by Cloud Tracing 
 		traceID, spanID, _ := deconstructTraceParent(header) 
 		if traceID != "" { 
 			e.Trace = traceID 
 			e.SpanID = spanID 
 			return true 
 		} 
 	} 
 	header = req.Header.Get("X-Cloud-Trace-Context") 
 	if header != "" { 
 		traceID, spanID, traceSampled := deconstructXCloudTraceContext(header) 
 		if traceID != "" { 
 			e.Trace = traceID 
 			e.SpanID = spanID 
 			// enforce sampling if required 
 			e.TraceSampled = e.TraceSampled || traceSampled 
 			return true 
 		} 
 	} 
 	return false 
 } 
```

To test this, we should make sure an http request with a context from a span creation ends up in the LogEntry:
```golang
    req := &http.Request{}
    ctx := context.Background()
    ctx, _ = trace.NewNoopTracerProvider().Tracer("test instrumentation").Start(ctx, "span name")
    req = req.WithContext(ctx)
    e := &Entry{}
    got := populateTraceInfo(e, req)
    assert.True(t, got)
    assert.Len(t, e.Trace, 32)
    assert.Len(t, e.SpanID, 16)
    // TODO test sampled somehow
```

gkevinzheng:
Fixes #9302


daniel-sanche (approved):
LGTM

bhshkh (approved):

cindy-peng (approved):

logging/logging_test.go
```diff
@@ -609,6 +612,132 @@ func TestToLogEntry(t *testing.T) {
...
+			want: &logpb.LogEntry{},
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			var span trace.Span
+			ctx := context.Background()
+
+			// Set up an OTel SDK tracer if integration test, mock noop tracer if not.
+			if integrationTest {
```
cindy-peng, line 690: Do we have a dedicated places for integration tests for Go Logging library?
For example, for Java-logging and Node-logging, there is a separate folder for integration tests:
https://github.com/googleapis/java-logging/tree/main/google-cloud-logging/src/test/java/com/google/cloud/logging/it
https://github.com/googleapis/nodejs-logging/tree/main/system-test

gkevinzheng: I don't think we have any right now, but if enough system tests happen that are kinda different from integration tests controlled by the `integrationTest` flag I could make a separate file for them.

dashpole (approved):

PR #10045, opened 2024-04-25
feat(storage/transfermanager): prototype

BrennaEpp:


storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,243 @@
...
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+)
+
+// Downloader manages parallel download operations from a Cloud Storage bucket.
+type Downloader struct {
+	client  *storage.Client
+	config  *transferManagerConfig
+	work    chan *DownloadObjectInput // Piece of work to be executed.
```
tritone, line 32: Presumably this should be send-only and output should be receive?

storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,243 @@
...
+	"errors"
+	"io"
+	"sync"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/iterator"
+)
+
+// Downloader manages parallel download operations from a Cloud Storage bucket.
```
tritone, line 28: Technically bucket can be specified per object. Let's just say that it manages a set of parallelized downloads.

storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,243 @@
...
+type Downloader struct {
+	client  *storage.Client
+	config  *transferManagerConfig
+	work    chan *DownloadObjectInput // Piece of work to be executed.
+	output  chan *DownloadOutput      // Channel for completed downloads; used to feed results iterator.
+	workers *sync.WaitGroup           // Keeps track of the workers that are currently running.
+}
+
+// DownloadObject queues the download of a single object. If it's larger than
+// the specified part size, the download will automatically be broken up into
```
tritone, line 38: We can leave off the part about sharding for the initial PR and add when we actually do that.

storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,243 @@
...
+	input.ctx = ctx
+	d.work <- input
+}
+
+// Download a set of objects to a local path. Downloader will create the needed
+// directory structure locally as the operations progress.
+// This will initiate the download but is non-blocking; wait on Downloader.Results
+// to process the result. Results will be split into individual objects.
+// NOTE: Do not use, DownloadDirectory is not implemented.
+func (d *Downloader) DownloadDirectory(ctx context.Context, input *DownloadDirectoryInput) {
```
tritone, line 51: Can leave this out of the PR for now.

storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,243 @@
...
+		d.output <- input.downloadShard(d.client, d.config.perOperationTimeout)
+	}
+	d.workers.Done()
+}
+
+// NewDownloader creates a new Downloader to add operations to.
+// Choice of transport, etc is configured on the client that's passed in.
+func NewDownloader(c *storage.Client, opts ...TransferManagerOption) (*Downloader, error) {
+	const (
+		chanBufferSize = 1000 // how big is it reasonable to make this?
```
tritone, line 94: I think this should be unbuffered and we should just buffer in a slice probably. Does that make sense to you?

storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	// TODO: write at a specific offset.
+	off := io.NewOffsetWriter(in.Destination, 0)
+	_, err = io.Copy(off, r)
+	if err != nil {
+		out.Err = err
+		r.Close()
+		return
+	}
```
tritone, line 188: We might need to look into whether we should be attempting to close the writer if possible -- seems annoying for end users to have to clean up. Though, I guess we don't guarantee that a close method exists.

storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,243 @@
...
+	if err = r.Close(); err != nil {
+		out.Err = err
+		return
+	}
+
+	out.Attrs = &r.Attrs
+	return
+}
+
+type DownloadDirectoryInput struct {
```
tritone, line 199: Remove for initial PR.

storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,243 @@
...
+// to process the result. Results will be split into individual objects.
+// NOTE: Do not use, DownloadDirectory is not implemented.
+func (d *Downloader) DownloadDirectory(ctx context.Context, input *DownloadDirectoryInput) {
+	d.output <- &DownloadOutput{Bucket: input.Bucket, Err: errors.New("DownloadDirectory is not implemented")}
+	// This does obj listing/os.File calls and then converts to object inputs.
+}
+
+// Waits for all outstanding downloads to complete. The Downloader must not be
+// used to download more objects or directories after this has been called.
+func (d *Downloader) WaitAndClose() error {
```
tritone, line 58: IMO this should return an error if any individual download returned an error. So, we'll have to collect those.

storage/transfermanager/integration_test.go
```diff
@@ -0,0 +1,326 @@
...
+	mw := io.MultiWriter(w, crc32cHash)
+
+	if _, err := io.CopyN(mw, rand.Reader, size); err != nil {
+		w.Close()
+		return 0, err
+	}
+	return crc32cHash.Sum32(), w.Close()
+}
+
+type testWriter struct {
```
tritone, line 191: doesn't have to be in this PR, but we should add a version of this (well, some kind of `DownloaderBuffer` that implements WriterAt) to the library.

storage/transfermanager/option.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (ww withWorkers) apply(tm *transferManagerConfig) {
+	tm.numWorkers = ww.numWorkers
+}
+
+// WithPartSize returns a TransferManagerOption that specifies the size of the
+// shards to transfer; that is, if the object is larger than this size, it will
+// be uploaded or downloaded in concurrent pieces.
+// The default is 32 MiB for downloads.
+// NOTE: Sharding is not yet implemented.
```
tritone, line 46: Leave off this PR.

storage/transfermanager/downloader.go
```diff
@@ -0,0 +1,243 @@
...
+	d := &Downloader{
+		client:  c,
+		config:  initTransferManagerConfig(opts...),
+		output:  make(chan *DownloadOutput, chanBufferSize),
+		work:    make(chan *DownloadObjectInput, chanBufferSize),
+		workers: &sync.WaitGroup{},
+	}
+
+	// Start workers in background.
+	for i := 0; i < d.config.numWorkers; i++ {
```
tritone, line 107: Presumably we could optimize this by spinning up workers as needed when there are objects enqueued? Doesn't have to be in this PR though.

tritone (commented):
A few initial comments, overall looks like a good start

storage/transfermanager/downloader.go
BrennaEpp: Are you thinking it will return an error collecting all errors that occurred, or just something that indicated that there was a failure somewhere?

storage/transfermanager/downloader.go
BrennaEpp: Makes sense, that wording is clearer.

storage/transfermanager/downloader.go
BrennaEpp: Not sure... does that mean that we'd have something in the background checking the slice and sending the work through?

storage/transfermanager/downloader.go
BrennaEpp: Sure, though I'm not sure how much that would optimize this by... I guess it depends on the num of workers.

storage/transfermanager/integration_test.go
BrennaEpp: Yeah, I'd have to look more into that. This is a very barebones implementation that is likely not at all efficient (and doesn't really work as a WriterAt yet).

storage/transfermanager/downloader.go
tritone: I think we can do the latter to start, but eventually I guess It would be nice to offer one large multierror as well

storage/transfermanager/integration_test.go
tritone: Yeah I think that's fine for now.

storage/transfermanager/downloader.go
tritone: Yeah something we can test out later.

storage/transfermanager/downloader.go
tritone: Yeah that sounds right. Background routine listens on results channel and writes to a slice (which functions as a queue but with no risk of blocking from hitting max length). Might make Next a little trickier though.

BrennaEpp: Hmm, so I'm not sure if I follow your solution and how `Next` plays into it. 

Rather than having a slice, I came up with a solution that uses errgroup. Let's take it offline to discuss pros and cons.

storage/transfermanager/downloader.go
BrennaEpp: Do you mean the writer provided through in.Destination? The `io.WriterAt` interface and for that matter `io. OffsetWriter` do not have a Close method, as you say. In any case, I feel like it'd be bad practice to close a writer we don't own?

storage/transfermanager/downloader.go
BrennaEpp: This now uses errors.Join

storage/transfermanager/downloader.go
```diff
@@ -17,61 +17,131 @@ package transfermanager
...
+		return fmt.Errorf("transfermanager: at least one error encountered downloading objects:\n%w", err)
+	}
 	return nil
 }
 
-// Results returns the iterator for download outputs.
-func (d *Downloader) Results() *DownloadOutputIterator {
-	return &DownloadOutputIterator{
-		output: d.output,
+// Results returns all the results of the downloads completed since the last
```
tritone, line 77: I think we could probably enforce that this can only be called after the full job is complete (given that we are keeping WaitAndClose as a single func). Or we could even just return the slice from WaitAndClose. What do you think?

storage/transfermanager/downloader.go
```diff
@@ -82,29 +152,46 @@ func (d *Downloader) downloadWorker() {
 		// Only request partSize data to begin with. If no error and we haven't finished
 		// reading the object, enqueue the remaining pieces of work and mark in the
 		// out var the amount of shards to wait for.
-		d.output <- input.downloadShard(d.client, d.config.perOperationTimeout)
+		out := input.downloadShard(d.client, d.config.perOperationTimeout)
+
+		// Keep track of any error that occurred.
+		if out.Err != nil {
+			d.error(out.Err)
```
tritone, line 159: Can we wrap this error to include the bucket/object name? I don't see that added in downloadShard either.

storage/transfermanager/integration_test.go
```diff
@@ -213,114 +646,185 @@ func (tw *testWriter) WriteAt(b []byte, offset int64) (n int, err error) {
 	return len(b), nil
 }
 
-func TestIntegration_DownloadShard(t *testing.T) {
+func deleteExpiredBuckets(prefix string) error {
```
tritone, line 652: We should look into creating an internal shared testutil for storage where these can live and be shared between the root package and any subpackages (Dataflux could also use these). Doesn't need to be in this PR though.

storage/transfermanager/option.go
```diff
@@ -39,28 +52,11 @@ func (ww withWorkers) apply(tm *transferManagerConfig) {
...
-}
-
-func (wps withPartSize) apply(tm *transferManagerConfig) {
-	tm.partSize = wps.partSize
-}
-
 // WithPerOpTimeout returns a TransferManagerOption that sets a timeout on each
 // operation that is performed to download or upload an object. The timeout is
 // set when the operation begins processing, not when it is added.
 // By default, no timeout is set.
```
tritone, line 58: ...other than an overall timeout set on the provided context.

storage/transfermanager/option.go
```diff
@@ -75,25 +71,26 @@ func (wpt withPerOpTimeout) apply(tm *transferManagerConfig) {
...
-	// JSON downloads but gRPC may benefit from larger.
-	partSize int
+
 	// Timeout for a single operation (including all retries). Zero value means
 	// no timeout.
 	perOperationTimeout time.Duration
+
+	// If true, callbacks are used instead of returning results synchronously
+	// in a slice at the end.
+	asynchronous bool
```
tritone, line 81: It looks like this is unused in the actual code currently?

storage/transfermanager/integration_test.go
```diff
@@ -17,160 +17,582 @@ package transfermanager
...
+				}
+			})
 		}
+	})
+}
+
+// errorIs is equivalent to errors.Is, except that it additionally will return
+// true if err and targetErr are googleapi.Errors with identical error codes,
+// or if both errors have the same gRPC status code.
+func errorIs(err error, targetErr error) bool {
```
tritone, line 582: Clever idea, we should probably do something like this for other tests which force errors.

tritone (commented):
Looks really good overall. We can chat tomorrow about a few of the remaining sync/async API changes perhaps?

storage/transfermanager/downloader.go
```diff
@@ -17,61 +17,131 @@ package transfermanager
...
-// NOTE: Do not use, DownloadDirectory is not implemented.
-func (d *Downloader) DownloadDirectory(ctx context.Context, input *DownloadDirectoryInput) {
-	d.output <- &DownloadOutput{Bucket: input.Bucket, Err: errors.New("DownloadDirectory is not implemented")}
-	// This does obj listing/os.File calls and then converts to object inputs.
+// DownloadObjectWithCallback queues the download of a single object. This will
+// initiate the download but is non-blocking; use the callback to process the
+// result.
+// The results of downloads initiated with this method will not be provided in
+// the results given in Downloader.Results.
+func (d *Downloader) DownloadObjectWithCallback(ctx context.Context, input *DownloadObjectInput, callback func(*DownloadOutput)) {
```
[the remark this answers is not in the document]
BrennaEpp: I forgot to remove the callback Option - right now it's completely unused. We could enforce it's usage (erroring in the output or here if the option does not match the call); as of now callbacks and no callbacks could be used in the same downloader (and it should work - though it's untested).

I lean towards the two separate entry points. I think the distinction in behaviour is big enough that we should make a similar distinction in our surface - I feel like having it just as a field isn't as nice an experience for people using it and encourages more mixing of using both callback and no callbacks. If we do add it as a field, I'd suggest then also adding the output to the results returned by Results(), in addition to calling the callback if available. That way the callback is more of an optional field that would get called and less of something that will cause different behaviour.

storage/transfermanager/downloader.go
```diff
@@ -17,61 +17,131 @@ package transfermanager
...
+	input.ctx = ctx
+	input.callback = &callback
+	d.addInput(input)
 }
 
-// Waits for all outstanding downloads to complete. The Downloader must not be
-// used to download more objects or directories after this has been called.
+// WaitAndClose waits for all outstanding downloads to complete. The Downloader
+// must not be used for any more downloads after this has been called.
```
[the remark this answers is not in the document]
BrennaEpp: We aren't enforcing it, but it should panic when trying to send on the closed channels. I can change the wording (and maybe also mention that it will panic) and add a test for it.

storage/transfermanager/option.go
tritone: That's what I meant; the user can/should set timeout in their context as well.

storage/transfermanager/downloader.go
tritone: Sounds good; yeah if it's in the top-level output then you can see which individual result to inspect as well.

storage/transfermanager/downloader.go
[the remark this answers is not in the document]
BrennaEpp: As discussed offline, we'll enforce usage of the option and set callback as a field.

storage/transfermanager/downloader.go
BrennaEpp: As discussed, I'll change this to return results directly in WaitAndClose

storage/transfermanager/downloader.go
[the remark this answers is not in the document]
BrennaEpp: I changed this to error instead of panicking.

storage/transfermanager/doc.go
```diff
@@ -0,0 +1,83 @@
...
+https://cloud.google.com/storage/docs.
+
+See https://pkg.go.dev/cloud.google.com/go for authentication, timeouts,
+connection pooling and similar aspects of this package.
+
+NOTE: This package is in alpha. It is not stable, and is likely to change.
+
+# Example usage
+
+	// Pass in any client opts or set retry policy here.
```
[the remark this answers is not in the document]
BrennaEpp: Done

storage/transfermanager/example_test.go
```diff
@@ -0,0 +1,146 @@
...
+	// Create download input
+	in := &transfermanager.DownloadObjectInput{
+		Bucket:      "mybucket",
+		Object:      "myblob",
+		Destination: f,
+		// Optionally specify params to apply to download.
+		EncryptionKey: []byte("mykey"),
+	}
+
+	// Can set timeout on this download using context. Note that this download
```
tritone, line 57: I think this is probably TMI for the example, really belongs in the docs for DownloadObject.

storage/transfermanager/downloader.go
```diff
@@ -41,51 +41,45 @@ type Downloader struct {
 }
 
 // DownloadObject queues the download of a single object. This will initiate the
-// download but is non-blocking; call Downloader.Results to process the result.
-func (d *Downloader) DownloadObject(ctx context.Context, input *DownloadObjectInput) {
-	input.ctx = ctx
-	d.addInput(input)
-}
+// download but is non-blocking; call Downloader.Results or use the callback to
+// process the result.
```
tritone, line 45: We should explicitly note that this is intended to be thread-safe (can share the Downloader across goroutines). And add the note about timeouts/timing that you have in the example.

tritone (commented):
Few docs notes, otherwise I think we should be able to submit this soon

storage/transfermanager/downloader.go
BrennaEpp: Done. Note I also made WaitAndClose callable more than once (just returns the same instead of panicking).

danielduhh:
Is this ready to merge @BrennaEpp @tritone ?

storage/transfermanager/integration_test.go
```diff
@@ -0,0 +1,851 @@
...
+// Create initializes the downloadTestBucket, creating a bucket and populating
+// objects in it. All objects are of the same size but with different contents
+// and can be mapped to their respective crc32c hash in contentHashes.
+func (tb *downloadTestBucket) Create(prefix string) error {
+	if testing.Short() {
+		return nil
+	}
+	ctx := context.Background()
+
+	tb.bucket = prefix + uidSpace.New()
```
tritone, line 746: I think we should talk about adding some more objects (at least more than the thread count) and of a wider variety of sizes. Can be in a follow up PR though.

PR #10046, opened 2024-04-25
feat(bigtable): Client side merics

bhshkh:
This PR adds support for built-in client side metrics in Go Bigtable client. Metrics will be recorded and exported by default to GCM if not opted out by user. 

Major changes in this PR:
1) ClientConfig - New field `MetricsProvider` has been added. It can be used to opt out of metrics collection.
2) metricsTracerFactory - Private field added on client. When a new client is created, 
- if user hasn't provided `MetricsProvider` option in config, a new meter provider is created with GCM exporter to record and export built-in metrics.
- if user has provided `NoopMetricsProvider` in config, metrics will neither be recorded, nor be exported
During client creation, a new metricsTracerFactory instance will be created, then OTel (Open Telemetry) instruments will be created on meter provider and these instruments will be saved to metricsTracerFactory.
Whenever any operation is invoked on data client such as ReadRows, a new builtinMetricsTracer will be created using the metricsTracerFactory
3) bigtable/metric_monitoring_exporter.go file is the GCM (Google Cloud Monitoring) exporter. The code in this file is a modified version ( [diff](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/873) ) of https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/88c799af8373802d659e6efdca387e3a353e8c21/exporter/metric/metric.go  :
    - Custom monitored resource type is being used here instead of deriving it from OpenTelemetry resource configuration
    - Metric attributes are being added to monitored resource labels while the original one adds OTel resource attributes to GCM monitored resource
    - Exporter exports only built in metrics to GCM
   
4) All the data methods have been modified to include attempt recorder and operation recorder.

Out of [all the metrics](https://cloud.google.com/bigtable/docs/client-side-metrics-descriptions), this PR adds operation_latencies, attempt_latencies, server_latencies and retry_count. Rest will be added in follow up PRs

Below diagram shows code flow when client is created using NewClient method:
CBT client side metrics (2)


bigtable/bigtable.go
```diff
@@ -295,13 +317,22 @@ func (ti *tableImpl) ApplyReadModifyWrite(ctx context.Context, row string, m *Re
 // By default, the yielded rows will contain all values in all cells.
 // Use RowFilter to limit the cells returned.
 func (t *Table) ReadRows(ctx context.Context, arg RowSet, f func(Row) bool, opts ...ReadOption) (err error) {
+	method := "cloud.google.com/go/bigtable.ReadRows"
```
igorbernstein2, line 320: The method names should match existing method names:
- ReadRows
- SampleRowKeys
- MutateRow
- MutateRows
- CheckAndMutateRow
- etc

bigtable/metrics.go
```diff
@@ -0,0 +1,479 @@
...
+	defaultMp := otelmetric.NewMeterProvider(
+		otelmetric.WithReader(
+			otelmetric.NewPeriodicReader(
+				defaultExporter,
+				otelmetric.WithInterval(defaultSamplePeriod),
+			),
+		),
+	)
+
+	userProvidedMeterProviders := []*otelmetric.MeterProvider{}
```
igorbernstein2, line 192: I dont quite understand the usecase for a user giving us multiple MeterProviders.

Separately, I dont think you want to have separate MeterProviders for user and builtin metrics....I suspect it will create significant overhead because all of the aggregations will end up happening twice

bigtable/metrics.go
```diff
@@ -0,0 +1,479 @@
...
+	if addr := os.Getenv("BIGTABLE_EMULATOR_HOST"); addr != "" {
+		// Do not emit metrics when emulator is being used
+		internalMetricsConfig.builtInEnabled = false
+	} else if userProvidedConfig != nil {
+		internalMetricsConfig.builtInEnabled = userProvidedConfig.builtInEnabled
+		userProvidedMeterProviders = userProvidedConfig.meterProviders
+	}
+
+	if !internalMetricsConfig.builtInEnabled {
```
igorbernstein2, line 201: Why does this happen so late? there is a significant amount of setup work already done. ie the PeriodicReader will probably already started its export goroutine?

bigtable/metrics.go
```diff
@@ -0,0 +1,479 @@
...
+func generateClientUID() (string, error) {
+	hostname := "localhost"
+	hostname, err := os.Hostname()
+	if err != nil {
+		return "", err
+	}
+	return "go-" + uuid.NewString() + "@" + hostname, nil
+}
+
+type metricsConfigInternal struct {
```
igorbernstein2, line 141: This isnt a Config ... in cbt Configs are value classes that allow end users to customize the client behavior.
This is struct has methods like recordAttemptCompletion, which suggest that its not just a value object.

Maybe it should be called something like CbtMetricsRecorder?

igorbernstein2: Also should there be an interface with 2 impls:
- A real impl that forwards the metrics
- A no-op impl that drops the metrics

bigtable/metrics.go
```diff
@@ -0,0 +1,479 @@
...
+		return internalMetricsConfig, nil
+	}
+
+	// Create instruments on all meter providers
+	allMeterProviders := append(userProvidedMeterProviders, defaultMp)
+	for _, mp := range allMeterProviders {
+		builtInMetricInstruments := builtInMetricInstruments{
+			distributions: make(map[string]metric.Float64Histogram),
+			counters:      make(map[string]metric.Int64Counter),
+		}
```
igorbernstein2, line 211: I'm a bit confused why you need the indirection via the hashmaps? Why not have the instruments be struct fields?

bigtable/metrics.go
```diff
@@ -0,0 +1,479 @@
...
+					return internalMetricsConfig, err
+				}
+			}
+			internalMetricsConfig.instruments = append(internalMetricsConfig.instruments, builtInMetricInstruments)
+		}
+	}
+	return internalMetricsConfig, nil
+}
+
+func (config *metricsConfigInternal) getOperationRecorder(ctx context.Context, table string, appProfile string, method string, isStreaming bool) (*builtinMetricsTracer, func()) {
```
igorbernstein2, line 243: Its very unclear what the 2nd member of the return tuple is used for.
Please add some docs

igorbernstein2 (changes requested):

bigtable/bigtable.go
bhshkh: Updated

bigtable/metrics.go
bhshkh: Updated the code to have single meterprovider

bigtable/metrics.go

bigtable/metrics.go
bhshkh: Updated the names and added implementations

bigtable/metrics.go
bhshkh: It was so that I could create instruments in a loop. Removed the hashmap now.

bigtable/metrics.go
bhshkh: Removed the 2nd member

bigtable/metrics_monitoring_exporter.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// Temporality returns the Temporality to use for an instrument kind.
+func (me *monitoringExporter) Temporality(ik otelmetric.InstrumentKind) otelmetricdata.Temporality {
+	return otelmetric.DefaultTemporalitySelector(ik)
+}
+
+// Aggregation returns the Aggregation to use for an instrument kind.
+func (me *monitoringExporter) Aggregation(ik otelmetric.InstrumentKind) otelmetric.Aggregation {
+	return otelmetric.DefaultAggregationSelector(ik)
```
igorbernstein2, line 121: Im a bit confused what this does. Shouldnt the aggregation be defined as part of the metric definition? sorry if its a dumb question, just not too familiar with golang otel api

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+	metricLabelKeyClientUID          = "client_uid"
+
+	// Metric names
+	metricNameOperationLatencies = "operation_latencies"
+	metricNameAttemptLatencies   = "attempt_latencies"
+	metricNameServerLatencies    = "server_latencies"
+	metricNameRetryCount         = "retry_count"
+)
+
+var (
```
igorbernstein2, line 71: I think this needs a comment as to why these are not constants

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+
+	// All the built-in metrics have same attributes except 'status' and 'streaming'
+	// These attributes need to be added to only few of the metrics
+	additionalAttributes = map[string][]string{
+		metricNameOperationLatencies: {
+			metricLabelKeyOperationStatus,
+			metricLabelKeyStreamingOperation,
+		},
+		metricNameAttemptLatencies: {
+			metricLabelKeyOperationStatus,
```
igorbernstein2, line 91: this is a bit weird to see OperationStatus in attempt latencies. Maybe the const key should be renamed to be applicable to both? (same for serverLatencies below). Or better still, there should be a separate const for attemptStatus

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+			metricLabelKeyStreamingOperation,
+		},
+		metricNameServerLatencies: {
+			metricLabelKeyOperationStatus,
```
igorbernstein2, line 95: similar to above this should be keyed for an attempt

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+			metricLabelKeyStreamingOperation,
+		},
+		metricNameRetryCount: {
+			metricLabelKeyOperationStatus,
```
igorbernstein2, line 99: same as above

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+			return "", err
+		}
+		return "go-" + uuid.NewString() + "@" + hostname, nil
+	}
+
+	exporterOpts = []option.ClientOption{}
+)
+
+// getBuiltInMeterProviderOptions returns meter provider options, shutdown function and error
+func getBuiltInMeterProviderOptions(ctx context.Context, project string) (sdkmetric.Option, error) {
```
igorbernstein2, line 119: nit this should be create

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+	metricNameAttemptLatencies   = "attempt_latencies"
+	metricNameServerLatencies    = "server_latencies"
+	metricNameRetryCount         = "retry_count"
+)
+
+var (
+	// duration between two metric exports
+	defaultSamplePeriod = 5 * time.Minute
+
+	clientName            = fmt.Sprintf("cloud.google.com/go/bigtable v%v", internal.Version)
```
igorbernstein2, line 75: this looks like a different format than java ... i think go-bigtable/$VERSION would be more consistent

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+
+	// Create retry_count
+	tf.retryCount, err = meter.Int64Counter(
+		metricNameRetryCount,
+		metric.WithDescription("The number of additional RPCs sent after the initial attempt."),
+	)
+	return err
+}
+
+// builtinMetricsTracer is created one per function call
```
igorbernstein2, line 248: which function call? I think it would be clearer if it said per operation

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+// It is used to store metric instruments, attribute values
+// and other data required to obtain and record them
+type builtinMetricsTracer struct {
+	ctx            context.Context
+	builtInEnabled bool
+	opStartTime    time.Time
+
+	// attributes that are specific to a client instance and
+	// do not change across different function calls on client
```
igorbernstein2, line 257: does function call mean operation or attempt here?

igorbernstein2: Also consider storing all of the client scoped values together in a struct

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+	instrumentAttemptLatencies   metric.Float64Histogram
+	instrumentRetryCount         metric.Int64Counter
+
+	tableName    string
+	appProfileID string
+	method       string
+	isStreaming  bool
+
+	// gRPC status code
+	status string
```
igorbernstein2, line 271: please indicate that this reflects the last completed attempt

igorbernstein2: same for the metadata below

igorbernstein2: I like that you started grouping things that are client scoped above, it would be good to continue grouping things that are operation scoped and attempt scoped and to order them hierarchically  client -> operation -> attempt

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+
+	// Contains the header response metadata which is used to extract cluster and zone
+	headerMD *metadata.MD
+
+	// Contains the trailer response metadata which is used to extract cluster and zone
+	trailerMD *metadata.MD
+
+	attemptCount int64
```
igorbernstein2, line 279: There is some ambiguity here as to who owns each attribute. ie it seems like the client populates everything except the attemptCount. I think it would be better to make this struct own its own state and have the client invoke a method to update internal state.

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+	// Only record when retry count is greater than 0 so the retry
+	// graph will be less confusing
+	if mt.attemptCount > 1 {
+		if retryCountErr == nil {
+			mt.instrumentRetryCount.Add(mt.ctx, mt.attemptCount-1, retryCntAllAttrs)
+		}
+	}
+}
+
+func (mt *builtinMetricsTracer) gaxInvokeWithRecorder(ctx context.Context, method string,
```
igorbernstein2, line 336: I think it might be nicer to move the inversion of control outside of this class. This way the client is the primary flow of control and this is just a helper rather then threading control in and out of the client

igorbernstein2: It should also make for easier testing in isolation

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+		default:
+			return nil, fmt.Errorf("Unknown additional attribute: %v", attrKey)
+		}
+	}
+
+	return attrKeyValues, nil
+}
+
+// get GFE latency from response metadata
+func (mt *builtinMetricsTracer) getServerLatency() (float64, error) {
```
igorbernstein2, line 428: Similar to above...I dont think the tracer should know about the encoding of headers...it should be solely responsible for being told what the metric values are and it should be either the client or another utility that extracts those values

bigtable/metrics.go
```diff
@@ -0,0 +1,487 @@
...
+	if !strings.HasPrefix(serverTimingStr, serverTimingValPrefix) || err != nil {
+		return serverLatencyMillis, err
+	}
+
+	return serverLatencyMillis, nil
+}
+
+// Obtain cluster and zone from response metadata
+func (mt *builtinMetricsTracer) getLocation() (string, string, error) {
+	var locationMetadata []string
```
igorbernstein2, line 460: same here

bigtable/metrics.go
```diff
@@ -0,0 +1,479 @@
...
+		attemptLatCurrCallAttrs, attemptLatErr := mt.toOtelMetricAttrs(metricNameAttemptLatencies)
+		attemptLatAllAttrs := metric.WithAttributes(append(config.clientAttributes, attemptLatCurrCallAttrs...)...)
+
+		// Attributes for server_latencies
+		serverLatCurrCallAttrs, serverLatErr := mt.toOtelMetricAttrs(metricNameServerLatencies)
+		serverLatAllAttres := metric.WithAttributes(append(config.clientAttributes, serverLatCurrCallAttrs...)...)
+
+		for _, builtInMetricInstruments := range config.instruments {
+			if attemptLatErr == nil {
+				builtInMetricInstruments.distributions[metricNameAttemptLatencies].Record(ctx, float64(elapsedTime), attemptLatAllAttrs)
```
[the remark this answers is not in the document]
bhshkh: [Default is cumulative](https://github.com/open-telemetry/opentelemetry-go/blob/sdk/metric/v1.27.0/sdk/metric/reader.go#L131) but updated the code to directly use cumulative since default could change.

bigtable/bigtable.go
```diff
@@ -52,17 +53,55 @@ const mtlsProdAddr = "bigtable.mtls.googleapis.com:443"
...
+
+// CustomOpenTelemetryMetricsProvider can be used to collect and export builtin metric with custom meter provider
+type CustomOpenTelemetryMetricsProvider struct {
+	MeterProvider *sdkmetric.MeterProvider
+}
+
+func (CustomOpenTelemetryMetricsProvider) isMetricsProvider() {}
+
+// WithBuiltIn takes in meter provider options and appends built in metrics periodic reader with exporter
+func WithBuiltIn(ctx context.Context, project string, opts ...sdkmetric.Option) ([]sdkmetric.Option, error) {
```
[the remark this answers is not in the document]
bhshkh: Users sometimes create ctx with timeout `context.WithTimeout(context.Background(), 10*time.Minute)`
ctx is required to create GCM client. Ideally, ctx used while creating GCM client should be same as the ctx used to create CBT client but since these 2 things happen independently, there is no way to ensure that.
For now, I will use background context for GCM client instead of taking it as an input from the user

bigtable/metrics_monitoring_exporter.go
```diff
@@ -0,0 +1,343 @@
...
+	}, nil
+}
+
+func toNonemptyTimeIntervalpb(start, end time.Time) (*monitoringpb.TimeInterval, error) {
+	// The end time of a new interval must be at least a millisecond after the end time of the
+	// previous interval, for all non-gauge types.
+	// https://cloud.google.com/monitoring/api/ref_v3/rpc/google.monitoring.v3#timeinterval
+	if end.Sub(start).Milliseconds() <= 1 {
+		end = start.Add(time.Millisecond)
+	}
```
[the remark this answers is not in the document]
bhshkh: I'm not sure when this will happen. this has been copied from https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/exporter/metric/v0.46.0/exporter/metric/metric.go

bigtable/metrics_monitoring_exporter.go
```diff
@@ -0,0 +1,343 @@
...
+	addAttributes(&attributes)
+	return &googlemetricpb.Metric{
+		Type:   fmt.Sprintf("%v%s", builtInMetricsMeterName, metrics.Name),
+		Labels: labels,
+	}, mr
+}
+
+// recordToTspb converts record to TimeSeries proto type with common resource.
+// ref. https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries
+func (me *monitoringExporter) recordToTspb(m otelmetricdata.Metrics) ([]*monitoringpb.TimeSeries, error) {
```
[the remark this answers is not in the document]
bhshkh: Moved

bigtable/metrics.go

bigtable/metrics.go
bhshkh: renamed to metricLabelKeyStatus

bigtable/metrics.go

bigtable/metrics.go
bhshkh: Done

bigtable/metrics.go

bigtable/metrics.go
bhshkh: Updated

bigtable/integration_test.go
```diff
@@ -899,6 +915,51 @@ func TestIntegration_LargeReadsWritesAndScans(t *testing.T) {
...
+
+	metricNamesValidate := []string{
+		metricNameOperationLatencies,
+		metricNameAttemptLatencies,
+		metricNameServerLatencies,
+	}
+	for _, metricName := range metricNamesValidate {
+		// ListTimeSeries can list only one metric type at a time.
+		// So, call ListTimeSeries with different metric names
+		iter := monitoringClient.ListTimeSeries(ctx, &monitoringpb.ListTimeSeriesRequest{
```
igorbernstein2, line 945: I think this will be flaky because because the data might not be immediately available. I would do an initial sleep for 30s and then poll for data every 10s for 5 mins

Also, can you extract this to its own integration test .. this one seems to be getting out of hand

bigtable/integration_test.go
```diff
@@ -899,6 +915,51 @@ func TestIntegration_LargeReadsWritesAndScans(t *testing.T) {
...
+
+		// Assert at least 1 datapoint was exported
+		_, err := iter.Next()
+		if testEnv.Config().UseProd && err != nil {
+			t.Errorf("%v not exported\n", metricName)
+		}
+		if !testEnv.Config().UseProd && err != iterator.Done {
+			t.Errorf("%v should not be exported when using emulator\n", metricName)
+		}
+	}
```
igorbernstein2, line 962: I would skip this test entirely for the emulator can just use a simple unit test to ensure that the default metrics provider is the noop one

bigtable/metrics.go
igorbernstein2: I think the comment for this block should be:
This are effectively const, but for testing purposes they are mutable

bigtable/metrics.go
```diff
@@ -0,0 +1,426 @@
...
+			attribute.String(metricLabelKeyClientUID, clientUID),
+			attribute.String(metricLabelKeyClientName, clientName),
+		},
+		shutdown: func() {},
+	}
+
+	if emulatorAddr := os.Getenv("BIGTABLE_EMULATOR_HOST"); emulatorAddr != "" {
+		// Do not emit metrics when emulator is being used
+		return tracerFactory, nil
+	}
```
igorbernstein2, line 152: please centralize the references to global env vars ... this should only be pulled from the env by the client and then passed around as an actual parameter. Ideally the client would explicitly set the noop tracer factory instead of bleeding the logic into the tracerfactory

bigtable/metrics.go
```diff
@@ -0,0 +1,426 @@
...
+		}
+		return "go-" + uuid.NewString() + "@" + hostname, nil
+	}
+
+	// Not a constant. Modifiable in tests
+	exporterOpts = []option.ClientOption{}
+)
+
+type builtinMetricsTracerFactory struct {
+	builtinEnabled bool
```
igorbernstein2, line 116: can this just be enabled? I'm not sure what prefixing it with built does

bigtable/metrics.go
igorbernstein2: that comment doesnt really help to explain what the grpc status code represents here

bigtable/metrics.go
```diff
@@ -0,0 +1,426 @@
...
+
+// recordOperationCompletion records as many metrics as it can and does not return error
+func recordOperationCompletion(mt *builtinMetricsTracer) {
+	if !mt.builtInEnabled {
+		return
+	}
+
+	// Use headers and trailers from last attempt to extract cluster and zone
+	headerMD := mt.currOp.currAttempt.headerMD
+	trailerMD := mt.currOp.currAttempt.trailerMD
```
igorbernstein2, line 307: I think it would be cleaner for the client push the parsed values instead of the tracer trying to parse them

bigtable/metrics_monitoring_exporter.go
```diff
@@ -0,0 +1,367 @@
...
+		return errShutdown
+	default:
+	}
+
+	return me.exportTimeSeries(ctx, rm)
+}
+
+// Temporality returns the Temporality to use for an instrument kind.
+func (me *monitoringExporter) Temporality(ik otelmetric.InstrumentKind) otelmetricdata.Temporality {
+	return otelmetric.DefaultTemporalitySelector(ik)
```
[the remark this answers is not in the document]
bhshkh: [Default is cumulative](https://github.com/open-telemetry/opentelemetry-go/blob/sdk/metric/v1.27.0/sdk/metric/reader.go#L131) but updated the code to directly use cumulative since default could change.

bigtable/bigtable.go
```diff
@@ -299,9 +359,17 @@ func (t *Table) ReadRows(ctx context.Context, arg RowSet, f func(Row) bool, opts
 	ctx = trace.StartSpan(ctx, "cloud.google.com/go/bigtable.ReadRows")
 	defer func() { trace.EndSpan(ctx, err) }()
 
+	mt := t.newBuiltinMetricsTracer(ctx, true)
+	defer recordOperationCompletion(mt)
+
+	err = t.readRows(ctx, arg, f, mt, opts...)
+	return mt.setOpStatus(err)
```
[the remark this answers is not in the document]
bhshkh: The function does modify the error. Updated the function name to denote the same

bigtable/metrics.go

bigtable/metrics.go

bigtable/metrics.go
bhshkh: Done

bigtable/metrics.go
bhshkh: Done

bigtable/bigtable.go
```diff
@@ -52,17 +53,60 @@ const mtlsProdAddr = "bigtable.mtls.googleapis.com:443"
...
+func (NoopMetricsProvider) isMetricsProvider() {}
+
+// CustomOpenTelemetryMetricsProvider can be used to collect and export builtin metric with custom meter provider
+type CustomOpenTelemetryMetricsProvider struct {
+	MeterProvider *sdkmetric.MeterProvider
+}
+
+func (CustomOpenTelemetryMetricsProvider) isMetricsProvider() {}
+
+// BuiltInMeterProviderOptions returns built in metrics periodic reader with exporter
```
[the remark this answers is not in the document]
bhshkh: Done

bigtable/bigtable.go
```diff
@@ -1118,10 +1200,18 @@ type entryErr struct {
...
+	method := "cloud.google.com/go/bigtable/Apply"
 	ctx = mergeOutgoingMetadata(ctx, t.md)
-	ctx = trace.StartSpan(ctx, "cloud.google.com/go/bigtable/ApplyBulk")
+	ctx = trace.StartSpan(ctx, method)
 	defer func() { trace.EndSpan(ctx, err) }()
 
+	mt := t.newBuiltinMetricsTracer(ctx, true)
+	defer recordOperationCompletion(mt)
+	errs, err = t.applyBulk(ctx, mt, rowKeys, muts, opts...)
+	return errs, mt.setOpStatus(err)
```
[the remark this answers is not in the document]
bhshkh: Done

bigtable/integration_test.go
bhshkh: Done

bigtable/integration_test.go
bhshkh: Done

bigtable/metrics.go
bhshkh: Done

bigtable/metrics.go
bhshkh: Done

bigtable/metrics.go
bhshkh: Done

bigtable/metrics.go
bhshkh: Done

bigtable/metrics_monitoring_exporter.go
bhshkh: I'm not aware why the otel interface was written that way but to implement the interface in our own exporter, we have to implement Aggregation as a separate method

bigtable/metrics.go
bhshkh: > does function call mean operation or attempt here?

Updated the comment

bigtable/bigtable.go
```diff
@@ -1118,8 +1233,9 @@ type entryErr struct {
 //
 // Conditional mutations cannot be applied in bulk and providing one will result in an error.
 func (t *Table) ApplyBulk(ctx context.Context, rowKeys []string, muts []*Mutation, opts ...ApplyOption) (errs []error, err error) {
+	method := "cloud.google.com/go/bigtable/Apply"
```
bhshkh, line 1236: Will revert this line in next commit

bigtable/bigtable.go
```diff
@@ -1118,8 +1233,9 @@ type entryErr struct {
...
 	ctx = mergeOutgoingMetadata(ctx, t.md)
-	ctx = trace.StartSpan(ctx, "cloud.google.com/go/bigtable/ApplyBulk")
+	ctx = trace.StartSpan(ctx, method)
```
bhshkh, line 1238: Will revert this line in next commit in this PR

bhshkh: Reverted

bigtable/bigtable.go
bhshkh: Reverted

bigtable/metrics_monitoring_exporter.go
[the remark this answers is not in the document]
igorbernstein2: can we remove it?

bigtable/bigtable.go
```diff
@@ -52,17 +55,84 @@ const mtlsProdAddr = "bigtable.mtls.googleapis.com:443"
...
+
+	clientConfig := ClientConfig{
+		MetricsProvider: CustomOpenTelemetryMetricsProvider{
+			MeterProvider: customMeterProvider,
+		},
+	}
+
+	client, err := NewClientWithConfig(context.Background(), project, instance, clientConfig)
+*/
+func BuiltInMeterProviderOptions(project string) ([]sdkmetric.Option, error) {
```
codyoss, line 124: I would be wary of putting a third-party project into the bigtable public API. This means if that package takes a breaking change in a minor version for whatever reason it is now a breaking change for bigtable. Also if otel ever released a v2 of their module this function is tightly coupled to the v1 implementation. Could this code live somewhere else?

bigtable/bigtable.go
```diff
@@ -52,17 +55,84 @@ const mtlsProdAddr = "bigtable.mtls.googleapis.com:443"
...
+}
+
+// NoopMetricsProvider can be used to disable built in metrics
+type NoopMetricsProvider struct{}
+
+func (NoopMetricsProvider) isMetricsProvider() {}
+
+// CustomOpenTelemetryMetricsProvider can be used to collect and export builtin metric with custom meter provider
+type CustomOpenTelemetryMetricsProvider struct {
+	MeterProvider *sdkmetric.MeterProvider
```
codyoss, line 94: See other comment below.

bigtable/bigtable.go
bhshkh: I could put it as a repository in googleapis or copy the whole thing into a third party folder in the bigtable client itself

bhshkh: Removed the public API. Will add it later with modifications

bigtable/bigtable.go
bhshkh: Removed the public API. Will add it later with modifications

bigtable/bigtable.go
```diff
@@ -52,19 +54,35 @@ const mtlsProdAddr = "bigtable.mtls.googleapis.com:443"
...
+	metricsTracerFactory *builtinMetricsTracerFactory
 }
 
 // ClientConfig has configurations for the client.
 type ClientConfig struct {
 	// The id of the app profile to associate with all data operations sent from this client.
 	// If unspecified, the default app profile for the instance will be used.
 	AppProfile string
+
+	// If not set, client side metrics will be collected and exported
```
gkevinzheng, line 70: nit: "If not set or is nil", to set the implication elsewhere that nil = default behavior

bigtable/metrics.go
```diff
@@ -0,0 +1,401 @@
...
+		}
+	}
+
+	// Create meter and instruments
+	meter := meterProvider.Meter(builtInMetricsMeterName, metric.WithInstrumentationVersion(internal.Version))
+	err = tracerFactory.createInstruments(meter)
+	return tracerFactory, err
+}
+
+func builtInMeterProviderOptions(project string) ([]sdkmetric.Option, error) {
```
gkevinzheng, line 189: Could you rename this function and change the `builtIn` part to something more specific to what it does?

bigtable/metrics.go
```diff
@@ -0,0 +1,401 @@
...
+
+func (a *attemptTracer) setServerLatency(latency float64) {
+	a.serverLatency = latency
+}
+
+func (a *attemptTracer) setServerLatencyErr(err error) {
+	a.serverLatencyErr = err
+}
+
+func (tf *builtinMetricsTracerFactory) newBuiltinMetricsTracer(ctx context.Context, tableName string, isStreaming bool) builtinMetricsTracer {
```
gkevinzheng, line 333: Since this is a factory method, could you rename the `new` part `get` or `create`?

bigtable/metric_util.go
```diff
@@ -0,0 +1,94 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	defaultCluster = "unspecified"
+	defaultZone    = "global"
+)
+
+// get GFE latency in ms from response metadata
+func getServerLatency(headerMD metadata.MD, trailerMD metadata.MD) (float64, error) {
```
gkevinzheng, line 35: Since this function gets the server latency from response metadata and isn't a getter, could you rename this to something like `getServerLatencyFromResponse`

bigtable/metric_util.go
```diff
@@ -0,0 +1,94 @@
...
+	serverLatencyMillis, err := strconv.ParseFloat(strings.TrimSpace(serverLatencyMillisStr), 64)
+	if !strings.HasPrefix(serverTimingStr, serverTimingValPrefix) || err != nil {
+		return serverLatencyMillis, err
+	}
+
+	return serverLatencyMillis, nil
+}
+
+// Obtain cluster and zone from response metadata
+func getLocation(headerMD metadata.MD, trailerMD metadata.MD) (string, string, error) {
```
gkevinzheng, line 66: Same comment as above, but for location.

bigtable/metrics.go
```diff
@@ -0,0 +1,401 @@
...
+	instrumentRetryCount         metric.Int64Counter
+
+	tableName   string
+	method      string
+	isStreaming bool
+
+	currOp opTracer
+}
+
+type opTracer struct {
```
gkevinzheng, line 270: Could you leave a comment on the purpose of opTracer and attemptTracer?

bigtable/metrics.go
```diff
@@ -0,0 +1,401 @@
...
+		metricNameRetryCount,
+		metric.WithDescription("The number of additional RPCs sent after the initial attempt."),
+	)
+	return err
+}
+
+// builtinMetricsTracer is created one per operation
+// It is used to store metric instruments, attribute values
+// and other data required to obtain and record them
+type builtinMetricsTracer struct {
```
gkevinzheng, line 250: There are three different tracers in this file. Could you rename `builtinMetricsTracer` to something that differentiates it from `opTracer` and `attemptTracer`?

gkevinzheng (approved):

igorbernstein2 (approved):
LGTM bigtable side

bigtable/metrics.go
```diff
@@ -0,0 +1,401 @@
...
+		},
+		metricNameRetryCount: {
+			additionalAttrs: []string{
+				metricLabelKeyStatus,
+			},
+			recordedPerAttempt: true,
+		},
+	}
+
+	// Generates unique client ID in the format go-<random UUID>@<>hostname
```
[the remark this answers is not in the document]

bigtable/bigtable.go
```diff
@@ -52,19 +54,35 @@ const mtlsProdAddr = "bigtable.mtls.googleapis.com:443"
...
+	//
+	// To disable client side metrics, set 'MetricsProvider' to 'NoopMetricsProvider'
+	MetricsProvider MetricsProvider
+}
+
+// MetricsProvider is a wrapper for built in metrics meter provider
+type MetricsProvider interface {
+	isMetricsProvider()
```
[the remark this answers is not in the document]
bhshkh: Added

bigtable/bigtable.go
bhshkh: Updated

bigtable/metrics.go
bhshkh: Changed to create

bigtable/metric_util.go
bhshkh: Renamed to extractServerLatency

bigtable/metric_util.go
bhshkh: Renamed to extractLocation

bigtable/metrics.go
bhshkh: Added

bigtable/bigtable.go
bhshkh: Added it to: https://github.com/googleapis/google-cloud-go/pull/10629

bigtable/metrics.go
bhshkh: Changed `MeterProviderOptions` in https://github.com/googleapis/google-cloud-go/pull/10629

bigtable/metrics.go

PR #10066, opened 2024-04-30
fix(bigquery): move MaxStaleness field to table level

### LINKED ISSUE #10026: bigquery: max staleness cannot be retreived/updated
**Client**

https://pkg.go.dev/cloud.google.com/go/bigquery@v1.60.0

**Code**

e.g.
```go
package main

import (
	"context"
	"fmt"
	"os"
	"time"

)

func main() {
	// Part 1: read current MaxStaleness
	ctx := context.TODO()
	client, _ := bigquery.NewClient(ctx, "my_project")
	table := client.Dataset("my_dataset").Table("my_materialized_view_with_staleness")
	md, _ := table.Metadata(ctx)
	fmt.Printf("%v\n", md.MaterializedView.MaxStaleness)

	// Part 2: update MaxStaleness to 5 hours
	tm := bigquery.TableMetadataToUpdate{
		MaterializedView: md.MaterializedView,
	}
	tm.MaterializedView.MaxStaleness = bigquery.IntervalValueFromDuration(5 * time.Hour)
	_, err := table.Update(ctx, tm, md.ETag)
	if err != nil {
		fmt.Fprintf(os.Stderr, "oops: %v\n", err)
		os.Exit(1)
	}
}

```

**Expected behavior**

When run against a MV that has `max_staleness` set, the code should:
* print the current max staleness interval (for example, `0-0 0 0:30:0`)
* update the interval to `0-0 0 5:0:0`

**Actual behavior**

The code:
* prints out `<nil>` as the current value of `MaxStaleness` (the actual value, set via SQL, is visible in BQ Studio)
* executes the update, but the MV is unchaged

**Additional context**

By inspecting the client code, I've observed that the [Table](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#Table) returned by the v2 API does return the current max staleness in the root `Table.maxStaleness` field. However, the client uses `Table.materializedView.maxStaleness`, which is always empty in the server response.

When it comes to writing, I couldn't get the BigQuery API to change the existing max staleness value. Setting `Table.maxStaleness` explicitly (instead of `Table.materializedView.maxStaleness`) in the client library did not work.

By the way, the CDC feature has its own [max staleness](https://cloud.google.com/bigquery/docs/change-data-capture#manage_table_staleness) setting (which applies to tables, not MVs), so I wonder if these two are conflated somewhere in the API/client.

Relevant [PR](https://github.com/googleapis/google-cloud-go/pull/8673) by @alvarowolfx (introduces max staleness to the client).

alvarowolfx:
Sounds like updating MaxStaleness is not working, so an internal ticket was opened to check that part.

Fixes #10026


bigquery/table.go
```diff
@@ -160,6 +160,10 @@ type TableMetadata struct {
 	// where 12345 is parent id. The value is the friendly short name of the
 	// tag value, e.g. "production".
 	ResourceTags map[string]string
+
+	// MaxStaleness of data that could be returned when materialized
```
shollyman, line 164: I think the description may need more detail.  My understanding is table max_staleness can be relevant to several things (views, federation queries)?

codyoss:
@alvarowolfx is this still active or can this be closed?

alvarowolfx:
still blocked by internal b/338036701. Will reopen when it's done.

shollyman (approved):

PR #10079, opened 2024-05-01
fix(spanner): Add ARRAY keywords to keywords

### LINKED ISSUE #10078: spanner: ARRAY functions missing from keyword list
**Client**

e.g. spanner

**Expected behavior**

I should be able to parse any statement using spansql.ParseDDL without error

**Actual behavior**

spansql.ParseDDL using an unsupported array function results in the error got "(" while expecting ")"

kai5263499:
Fixes #10078


harshachinta:
GoogleSQL supported Array functions - https://cloud.google.com/spanner/docs/reference/standard-sql/array_functions

kai5263499:
> GoogleSQL supported Array functions - https://cloud.google.com/spanner/docs/reference/standard-sql/array_functions

This list appears to match what I've added in my PR. Is there something else I should add here?

harshachinta:
@kai5263499 Please add tests in parse_test.go similar to https://github.com/googleapis/google-cloud-go/pull/9557?

kai5263499:
> @kai5263499 Please add tests in parse_test.go similar to #9557?

I added tests. That led me to remove `ARRAY_FILTER` because it expects a lambda expression and that would need to be a brand new type for the parser to support. I'll create a separate issue to track the work that needs to be done to support it.

harshachinta (approved):

PR #10091, opened 2024-05-02
docs(testing): switch deprecated WithInsecure to WithTransportCredentials

mark-adams:
The public [GRPC docs](https://pkg.go.dev/google.golang.org/grpc#WithInsecure) mention the following with regards to `grpc.WithInsecure`:

> Deprecated: use WithTransportCredentials and insecure.NewCredentials() instead. Will be supported throughout 1.x.

This change updates the example to use `WithTransportCredentials()` along with `insecure.NewCredentials()` instead of the deprecated option.

### DIFF
--- a/testing.md
+++ b/testing.md
@@ -99,6 +99,7 @@ import (
         "google.golang.org/api/option"
         translatepb "google.golang.org/genproto/googleapis/cloud/translate/v3"
         "google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
 )
 
 func TestTranslateTextWithConcreteClient(t *testing.T) {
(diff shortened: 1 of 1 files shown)


noahdietz (approved):
Thank you for updating this!

mark-adams:
Updated the commit to match conventional commits styling 🤞

noahdietz:
> Updated the commit to match conventional commits styling 🤞

Thanks !

When we use the `automerge` label, or when there are multiple commits, the bot uses the PR title as the squashed commit message, so there isn't a need to force a new commit to fix it. That said, observing our best practices is appreciated :)

noahdietz (approved):

mark-adams:
@noahdietz Ah, good to know! Thanks!

PR #10095, opened 2024-05-03
fix(spanner): wait for background goroutines to complete

### LINKED ISSUE #10324: spanner: background goroutines are left running
Tracking issue for https://github.com/googleapis/google-cloud-go/pull/10095

Currently the session client workers outlive the client, which may cause use of closed or released resources. This often can show up as flaky tests.

Similarly, the tests should wait for the server completely shutdown, otherwise there's a similar problem.

egonelbre:
As a good rule of thumb every single goroutine needs to be waited to be completed, otherwise it's easy to introduce issues where the created goroutine outlives the parent and accesses variables or services that have been shutdown.

Fixes #10324


rahul2393 (approved):

codyoss:
@rahul2393 can we merge this?

egonelbre:
@rahul2393 can this be merged?

PR #10097, opened 2024-05-03
fix(aliasfix): handle import paths correctly

egonelbre:
The last component in import path of the old genproto path does not correspond to the import path -- usually.

The import paths usually look like `google.golang.org/genproto/googleapis/analytics/admin/v1alpha`, however the imported package name is `admin`. This uses a heuristic to detect whether the last path component is a version,

There are some packages that do not fit that description though, e.g. `google.golang.org/genproto/googleapis/devtools/containeranalysis/v1beta1/grafeas`.

### DIFF
--- a/internal/aliasfix/aliasfix.go
+++ b/internal/aliasfix/aliasfix.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"io/fs"
 	"os"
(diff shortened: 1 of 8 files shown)


codyoss (approved):
Thanks!

PR #10098, opened 2024-05-03
chore(spanner): grpc.WithInsecure is deprecated

egonelbre:
grpc.WithTransportCredentials(insecure.NewCredentials) should be used instead.

Updates #10325


codyoss:
@rahul2393 for review

rahul2393 (approved):

Issue #9380, opened 2024-02-06
spanner: third party dependency exception for github.com/json-iterator/go


noahdietz:
Module: `spanner`
Usage(s): [`spanner/value.go`](https://github.com/googleapis/google-cloud-go/blob/2fe67df51fa44b01efa37af97f1284ac4fcd2403/spanner/value.go#L117C17-L132)

For the time being, it will be exempt from the third party dep check, but please do one of the following:

A. Remove the dependency by switching to a package own directly by Google, handwriting necessary functionality, or using stdlib
B. Justify the exception and [ack the risks](https://github.com/googleapis/google-cloud-go/blob/main/CONTRIBUTING.md#policy-on-new-dependencies), maintaining the exception indefinitely

egonelbre:
There is a significant concern from my point of view with json-iterator dependency due to it's heavy reliance on unsafe and modern-go/reflect2. See https://github.com/golang/go/issues/54766#issuecomment-1983808590 for issues that happen from it -- the lucky case is when your compilation or tests fail.

rahul2393:
Thanks @egonelbre for pointing this out, bumping the priority for now.

PR #10099, opened 2024-05-03
fix(spanner): remove json-iterator dependency

egonelbre:
json-iterator heavily relies on an unmaintained modern-go/reflect2
that pokes at the internals of Go runtime. This was shown to cause
failures in https://github.com/golang/go/issues/54766#issuecomment-1983808590.

Switching to encoding/json does come with a performance penalty.

Updates #9380


rahul2393 (approved):

noahdietz:
@rahul2393 you'll need to squash/merge this yourself or add the automerge label

PR #10102, opened 2024-05-03
feat(auth): Enable client certificates by default

andyrzhao:
Implications of this change:
1. If no default certificate sources are available in the environment (i.e. no SecureConnect or ECP cert), then the before & after behavior are identical.
2. If SecureConnect cert source is available (i.e. Googlers with EndpointVerification installed), then connections are automatically upgraded to mTLS. This is deemed a safe upgrade from policy enforcement perspective. From a latency perspective, since this does not impact workload use-cases, there should be negligible impact, even if the SecureConnect cert is from a TPM.
3. ECP cert source is enabled via certificate_config.json, which is not widely rolled out at the moment. Those with certificate_config.json configured already have the intention to enable client certificates anyway, so this saves them an extra flag to manage.


ulisesL (approved):
LGTM

codyoss (approved):

Issue #10069, opened 2024-04-30
spanner: fix TestClient_CallOptions test


noahdietz:
We have to skip `TestClient_CallOptions` temporarily in order to merge regeneration PR #9777. It includes a [Spanner retry config change](https://github.com/googleapis/googleapis/commit/60536a2a263b6d33b0b1adb5b10c10e34ccf4528#diff-608200f96655b06b8320e5fcd66029f265fd62d04f9fb6371e1c8aae0d412edeL64) which breaks some test assumptions. Fixing it in the same PR is time consuming and blocks many other changes from landing.

This should be fixed before the next spanner release.

PR #10117, opened 2024-05-06
test(spanner): fix retry unit tests with RESOURCE_EXHAUSTED errors

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/10070, https://github.com/googleapis/google-cloud-go/issues/10069

### DIFF
--- a/spanner/internal/testutil/inmem_spanner_server.go
+++ b/spanner/internal/testutil/inmem_spanner_server.go
@@ -731,7 +731,7 @@ func (s *inMemSpannerServer) BatchCreateSessions(ctx context.Context, req *spann
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if s.maxSessionsReturnedByServerInTotal > int32(0) && int32(len(s.sessions)) >= s.maxSessionsReturnedByServerInTotal {
-		return nil, gstatus.Error(codes.ResourceExhausted, "No more sessions available")
+		return nil, gstatus.Error(codes.OutOfRange, "No more sessions available")
 	}
 	if sessionsToCreate > s.maxSessionsReturnedByServerPerBatchRequest {
 		sessionsToCreate = s.maxSessionsReturnedByServerPerBatchRequest
--- a/spanner/sessionclient_test.go
+++ b/spanner/sessionclient_test.go
@@ -424,7 +424,6 @@ func TestBatchCreateSessions_ServerReturnsLessThanRequestedSessions(t *testing.T
 }
 
 func TestBatchCreateSessions_ServerExhausted(t *testing.T) {
-	t.Skip("https://github.com/googleapis/google-cloud-go/issues/10070")
 	t.Parallel()
 
 	numChannels := 4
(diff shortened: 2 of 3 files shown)


spanner/sessionclient_test.go
```diff
@@ -448,7 +447,7 @@ func TestBatchCreateSessions_ServerExhausted(t *testing.T) {
 		t.Fatalf("Error count mismatch\nGot: %d\nWant: > %d", len(consumer.errors), 0)
 	}
 	for _, e := range consumer.errors {
-		if g, w := status.Code(e.err), codes.ResourceExhausted; g != w {
+		if g, w := status.Code(e.err), codes.OutOfRange; g != w {
```
harshachinta, line 450: Looking at the failed build it shows it was expecting DeadLineExceeded but got ResourceExhausted
image

Since the test is getting mocked and in this case we are returning `RESOURCE_EXHAUSTED` from here https://github.com/googleapis/google-cloud-go/blob/b9b55e0ebc7427a5a55be1485e38db4a7eddf832/spanner/internal/testutil/inmem_spanner_server.go#L734. then why is it expecting DEADLINE_EXCEEDED? Any idea?

spanner/internal/testutil/inmem_spanner_server.go
```diff
@@ -731,7 +731,7 @@ func (s *inMemSpannerServer) BatchCreateSessions(ctx context.Context, req *spann
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if s.maxSessionsReturnedByServerInTotal > int32(0) && int32(len(s.sessions)) >= s.maxSessionsReturnedByServerInTotal {
-		return nil, gstatus.Error(codes.ResourceExhausted, "No more sessions available")
+		return nil, gstatus.Error(codes.OutOfRange, "No more sessions available")
```
harshachinta, line 734: Why do we need to change this to a different error code? I didn't understand this change.

spanner/sessionclient_test.go
harshachinta: or was this flaky?

rahul2393: No it was not flaky, earlier we were reutrning mock response of ResourceExhausted and it was not getting retried so it was passing, but since ResourceExhausted is marked as retrybale it will keep retrying until DeadlineExceed happend, the fix for that is to change mock response.

spanner/internal/testutil/inmem_spanner_server.go
rahul2393: This was the mock response, it has nothing to do with actual cloud spanner backend response code, but unfortunately since ResourceExhausted is treated as retryable now we are changing to some other error code to make it fail and not retry the error which was assumption when writing the test case/this mock code before.

harshachinta: Got it. Can we document this assumption as a comment on top of the mock response and the test? This way developers can understand the reason behind sending this irrelevant error code?

Also how about marking it as `INTERNAL` error? `OutOfRange` error doesn't seem to be relevant to the problem? I will leave it up to you since this mock error code is not serving any purpose.

spanner/sessionclient_test.go
harshachinta: makes sense! Thanks for clarification.

harshachinta (approved):
LGTM! I would prefer documenting the behavior on top of the assertion and the place where the mock error is returned. Otherwise this becomes confusing!

spanner/internal/testutil/inmem_spanner_server.go
rahul2393: Actually OutOfRange suits more here https://developers.google.com/actions-center/reference/grpc-api/status_codes, since in mock code we want to make sure session count never exceeds maxSessionsReturnedByServerInTotal

PR #10119, opened 2024-05-06
chore(spanner): avoid b.Fatal in child goroutine

egonelbre:
TB.Fatal should only be called from the test function itself.
Instead let's panic, because an error happens then the code is
broken.


egonelbre:
Looks like this was fixed in 21cba147834634892a8d8b264bd48e235bd50091.

PR #10127, opened 2024-05-07
feat(storage): add BucketName to BucketHandle

derekperkins:
The bucket name is currently only accessible via the Attrs() network call, and can be trivially returned from the bucket handle. Today we have to pass around a `BucketHandle` and the bucket name separately.

I used BucketName instead of just Name, as ObjectHandle has both BucketName and ObjectName.


tritone (commented):
This looks fine to me. The `go vet` check is failing on `go fmt`, could you fix that?

derekperkins:
I used the web editor and left a stray tab. That should be resolved now.

tritone (approved):

derekperkins:
@tritone any blockers before merging this?

derekperkins:
Thanks!

PR #10135, opened 2024-05-08
docs(storage): update autoclass doc

cojenco:
Add a note to emphasize that users need to update the entire Autoclass config with the enabled field explicitly set

ref: internal b/337286895


storage/bucket.go
```diff
@@ -767,6 +767,7 @@ type Autoclass struct {
 	// TerminalStorageClass: The storage class that objects in the bucket
 	// eventually transition to if they are not read for a certain length of
 	// time. Valid values are NEARLINE and ARCHIVE.
+	// To modify TerminalStorageClass, Enabled should be set to true.
```
tritone, line 770: s/should/must.

storage/bucket.go
```diff
@@ -1174,6 +1175,8 @@ type BucketAttrsToUpdate struct {
 	RPO RPO
 
 	// If set, updates the autoclass configuration of the bucket.
+	// Library users should provide the updated autoclass configuration
+	// with the enabled field explicitly set.
```
tritone, line 1179: I'd say something like:

`To disable autoclass on the bucket, set to an empty &Autoclass{}. To update the configuration for Autoclass.TerminalStorageClass, Autoclass.Enabled must also be set to true.`

storage/bucket.go
cojenco: Done

storage/bucket.go
cojenco: This is much clearer, thanks! Updated.

tritone (approved):

PR #10147, opened 2024-05-10
fix(auth): have refresh time match docs

clundin25:
3 mins and 45 seconds == 225 seconds.

This error is is harmless but fixing it to make sure comments and
documentation reflect reality.


codyoss (approved):
Good catch!

Issue #10155, opened 2024-05-12
datastore: protobuf namespace conflict with google.golang.org/appengine/v2


sg0hsmt:
**Client**

datastore

**Environment**

Go executable on Linux

**Go Environment**

```console
$ go version
go version go1.22.3 linux/amd64
```

```console
$ go env
GO111MODULE=''
GOARCH='amd64'
GOBIN=''
GOCACHE='/home/vscode/.cache/go-build'
GOENV='/home/vscode/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='local'
GOTOOLDIR='/usr/local/go/pkg/tool/linux_amd64'
GOVCS=''
GOVERSION='go1.22.3'
GCCGO='gccgo'
GOAMD64='v1'
AR='ar'
CC='gcc'
CXX='g++'
CGO_ENABLED='1'
GOMOD='/workspaces/namespace-conflict/go.mod'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -ffile-prefix-map=/tmp/go-build3064827702=/tmp/go-build -gno-record-gcc-switches'
```

**Code**

```go
package main

import (
	"fmt"

	_ "cloud.google.com/go/datastore"
	_ "google.golang.org/appengine/v2"
)

func main() {
	fmt.Println("Hello, World!")
}
```

**Expected behavior**

`Hello, World!` will be displayed.

**Actual behavior**

A panic will occur.
```console
$ go run main.go 
panic: proto: file "datastore_v3.proto" is already registered
        previously from: "cloud.google.com/go/datastore/internal/gaepb"
[...]

PR #10158, opened 2024-05-13
fix(datastore): regenerate protos in new namespace

codyoss:
The namespace and filename need to be different than the file these were copied and regenerated from to make the proto registry not error out on initialization.

Usually it is not good to use a new package name in protos but I believe it is fine here since these are just simple message types being declared and not a service.

Fixes: #10155


noahdietz:
While this would fix it, it is [also risky](https://protobuf.dev/reference/go/faq/#fix-namespace-conflict). Would it be possible to do a little more brainstorming before pressing the button?

codyoss:
> While this would fix it, it is [also risky](https://protobuf.dev/reference/go/faq/#fix-namespace-conflict). Would it be possible to do a little more brainstorming before pressing the button?

For sure. I did look at that page before I made the change. I think this should all be safe and non-breaking for from my read. It is not ideal though. In a perfect world those protos would be central defined somewhere and we would import reference them in both places. 

It should also be noted that it really valid to use this library in appengine standard according to the docs: https://cloud.google.com/appengine/docs/legacy/standard/go111/datastore#connecting_to_with_app_engine

`You cannot use the [Cloud Datastore client library](https://cloud.google.com/datastore/docs/reference/libraries#client-libraries-install-go) with Go applications in the App Engine standard environment.`

And in flex I don't think there is value in using the app engine package, that was designed for standard: https://github.com/golang/appengine?tab=readme-ov-file#upgrading-an-app-engine-app-to-the-flexible-environment

noahdietz:
> You cannot use the [Cloud Datastore client library](https://cloud.google.com/datastore/docs/reference/libraries#client-libraries-install-go) with Go applications in the App Engine standard environment.

So is this a feature? 😛 

Ok I've reviewed the usage in `datastore` and this is likely OK...if it weren't the tests should've failed.

noahdietz (approved):

PR #10165, opened 2024-05-14
test(pubsub): make error checking of receive less brittle

hongalex:
`Receive` can occasionally return `context.Canceled`. This makes tests that call `Receive` less brittle by checking explicitly for `canceled`. This also replaces checks of the `codes.Canceled` status code.


shollyman (commented):
Generally, these all seem like a good case for a swap to `errors.Is(err, context.Cancelled)` rather than a direct equality check.

shollyman (approved):
approving based on discussed refactoring

PR #10170, opened 2024-05-14
docs(bigquery): improve Inserter and StructSaver godoc

### LINKED ISSUE #9210: StructSaver example invalid
here:
https://godocs.io/cloud.google.com/go/bigquery#hdr-Loading_and_Uploading

I see this:

> You can also upload a struct that doesn't implement ValueSaver. Use the StructSaver type to specify the schema and insert ID by hand, or just supply the struct or struct pointer directly and the schema will be inferred:

~~~go
type Item2 struct {
    Name  string
    Size  float64
    Count int
}
// Item implements the ValueSaver interface.
items2 := []*Item2{
    {Name: "n1", Size: 32.6, Count: 7},
    {Name: "n2", Size: 4, Count: 2},
    {Name: "n3", Size: 101.5, Count: 1},
}
if err := u.Put(ctx, items2); err != nil {
    // TODO: Handle error.
}
~~~

but the text doesnt match the example. the example does not use `StructSaver`, the comment in the example refers to Item, which is not actually in the code, and the comment refers to ValueSaver, which the whole point of the code block is to demonstrate NOT using ValueSaver

alvarowolfx:
Fixes #9210


bigquery/doc.go
```diff
@@ -272,15 +287,33 @@ Then create an Inserter, and call its Put method with a slice of values.
...
+Lastly, but not least, you can just supply the struct or struct pointer directly and the schema will be inferred:
 
 	type Item2 struct {
 	    Name  string
 	    Size  float64
 	    Count int
 	}
-	// Item implements the ValueSaver interface.
+
+	// Item2 doesn't implement ValueSaver interface, so schema will be infered
```
shollyman, line 316: s/infered/inferred/

shollyman (approved):

PR #10185, opened 2024-05-15
test: fix presubmit script to handle dirs

codyoss:
It seems #10064 changed the semantics from files to dirs for significant changes. This caused `CHANGED_DIRS` to fail when the grep test was done as sometimes there is no slash now. This piping was not actually needed though and the cut should work even if the character is not there.

This failure to grep caused presubmits to run for whole repo sometimes when changes were only made in one sub-module.


noahdietz (approved):

noahdietz:
Nice catch

PR #10192, opened 2024-05-16
fix(internal/postprocessor): add scopes to all appropriate commit lines

noahdietz:
Adds the scope, if the API directory changed is one configured for gapic generation, to all commit lines inside of a nested commit.

The `go-yoshi` release-please strategy [will only include a commit line](https://github.com/googleapis/release-please/blob/656b9a9ad1ec77853d16ae1f40e63c4da1e12f0f/src/strategies/go-yoshi.ts#L132-L151) from a nested commit if it has a scope and one that matches the module up for release.

Note: not all commit lines in the test data get scopes now - some have nested commits for modules that aren't actually configured for gapic generation anymore i.e. batch/v1alpha1.

Also fixes an issue where we'd get a double bang ([internal link](https://screenshot.googleplex.com/BsyaBHuYKmKbRRt)) for breaking change commit lines that received scopes (which is now going to be all of them instead of just the first).


quartzmo:
> Note: not all commit lines in the test data get scopes now - some have nested commits for modules that aren't actually configured for gapic generation anymore i.e. batch/v1alpha1.

How hard would it be to update the tests to process only modules that are currently configured? The un-scoped commits seem like a potential source of confusion for future maintainers.

internal/postprocessor/main.go
```diff
@@ -528,13 +544,34 @@ func (p *postProcessor) processCommit(title, body string) (string, string, error
 					scope = scopes[0]
 				}
 
-				newCommitTitle := updateCommitTitle(currTitle, scope)
+				newCommitTitle := updateCommit(currTitle, scope)
```
quartzmo, line 547: Have you done an audit of how the term "title" is used here? Any thoughts about how to better align the usage with the way the term is used in git and GitHub?

quartzmo (approved):

noahdietz:
> How hard would it be to update the tests to process only modules that are currently configured? The un-scoped commits seem like a potential source of confusion for future maintainers.

Yeah, I hear you, took me a bit of digging...I think we want to keep them though, they do exercise a condition in the code.

internal/postprocessor/main.go
noahdietz: We resolved offline: it makes sense to keep using `Title` for some variables b.c it is a recognized section of the git spec (content up to the first line break in a commit's message).

PR #10213, opened 2024-05-16
fix(bigtable): Track number of rows read to set rowsLimit in subsequent requests

bhshkh:
Tracking bug: b/303943537

This PR fixes TestReadRows_NoRetry_ErrorAfterLastRow test

https://github.com/googleapis/cloud-bigtable-clients-test/blob/db4d6a9e8e5e38b6f07383bd46ac14bfc8d8b9b6/tests/readrows_test.go#L171-L173

fails with error:

```
--- FAIL: TestReadRows_NoRetry_ErrorAfterLastRow (0.03s)
    readrows_test.go:201: 
        	Error Trace:	cloud-bigtable-clients-test/tests/readrows_test.go:201
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 2
        	Test:       	TestReadRows_NoRetry_ErrorAfterLastRow
--- FAIL: TestReadRows_Retry_LastScannedRow_Reverse (0.04s)
```

The test sends below request to Go Bigtable proxy:
```
Request: &btpb.ReadRowsRequest{
	TableName: buildTableName("table"),
	RowsLimit: 1,
},
```
On receiving the request, the proxy creates a ReadRows request for the client library and invokes ReadRows function.
Even though test sends RowsLimit field in the method, the proxy does not add it to the client library request. This has been fixed in this PR.


codyoss:
@bhshkh is this still wanted/needed?

bhshkh:
> @bhshkh is this still wanted/needed?

Yes

bigtable/internal/testproxy/proxy.go
```diff
@@ -570,20 +570,33 @@ func (s *goTestProxyServer) ReadRows(ctx context.Context, req *pb.ReadRowsReques
...
+	if rrq.RowsLimit != 0 {
+		limitRowsRead = true
+		rowsToRead = rrq.RowsLimit
+	}
+	if req.GetCancelAfterRows() != 0 {
+		limitRowsRead = true
+		rowsToRead = int64(req.GetCancelAfterRows())
+	}
+
+	// Client libray does not have a built-in way to limit the number of rows read in a call to Table.ReadRows().
```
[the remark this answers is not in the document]
bhshkh: Updated

bigtable/bigtable.go
```diff
@@ -410,7 +424,7 @@ func (t *Table) readRows(ctx context.Context, arg RowSet, f func(Row) bool, mt *
 			}
 			req.Rows = arg.proto()
 		}
-		settings := makeReadSettings(req)
+		settings := makeReadSettings(req, numRowsRead)
 		for _, opt := range opts {
```
mutianf, line 428: where is opts used? Is it supposed to be called on line 435?

bhshkh: settings is passed into opt.set at line 429
each opt modifies the req object. 

For e.g. LimitRows sets the RowsLimit field of req

bigtable/bigtable.go
```diff
@@ -391,7 +391,21 @@ func (t *Table) ReadRows(ctx context.Context, arg RowSet, f func(Row) bool, opts
...
 	var prevRowKey string
 	attrMap := make(map[string]interface{})
+
+	numRowsRead := int64(0)
+	rowLimitSet := false
+	intialRowLimit := int64(0)
+	for _, opt := range opts {
+		if l, ok := opt.(limitRows); ok {
+			rowLimitSet = true
+			intialRowLimit = l.limit
```
andre-sampaio, line 401: Maybe fail fast on `l.limit < 0`

bigtable/bigtable.go
```diff
@@ -391,7 +391,21 @@ func (t *Table) ReadRows(ctx context.Context, arg RowSet, f func(Row) bool, opts
...
+		}
+	}
 	err = gaxInvokeWithRecorder(ctx, mt, "ReadRows", func(ctx context.Context, headerMD, trailerMD *metadata.MD, _ gax.CallSettings) error {
+		if rowLimitSet && numRowsRead == intialRowLimit {
```
andre-sampaio, line 405: Should we add a safety check in case `numRowsRead > initialRowLimit` as  an extra safety measure against bugs down the stream?

bigtable/bigtable.go
```diff
@@ -939,10 +955,11 @@ type FullReadStatsFunc func(*FullReadStats)
 type readSettings struct {
 	req               *btpb.ReadRowsRequest
 	fullReadStatsFunc FullReadStatsFunc
+	numRowsRead       int64
 }
 
-func makeReadSettings(req *btpb.ReadRowsRequest) readSettings {
-	return readSettings{req, nil}
+func makeReadSettings(req *btpb.ReadRowsRequest, numRowsRead int64) readSettings {
```
andre-sampaio, line 961: NIT: it took me a while to understand how `makeReadSettings` works with `readSettings.set`. Maybe add a one line comment either here or on the read execution itself explaining

bigtable/bigtable.go
```diff
@@ -965,7 +982,9 @@ func LimitRows(limit int64) ReadOption { return limitRows{limit} }
 
 type limitRows struct{ limit int64 }
 
-func (lr limitRows) set(settings *readSettings) { settings.req.RowsLimit = lr.limit }
+func (lr limitRows) set(settings *readSettings) {
+	settings.req.RowsLimit = lr.limit - settings.numRowsRead
```
andre-sampaio, line 986: Maybe add a comment explaining the reasoning for this subtraction

bigtable/bigtable.go
```diff
@@ -391,7 +391,21 @@ func (t *Table) ReadRows(ctx context.Context, arg RowSet, f func(Row) bool, opts
...
+	intialRowLimit := int64(0)
+	for _, opt := range opts {
+		if l, ok := opt.(limitRows); ok {
+			rowLimitSet = true
+			intialRowLimit = l.limit
+		}
+	}
 	err = gaxInvokeWithRecorder(ctx, mt, "ReadRows", func(ctx context.Context, headerMD, trailerMD *metadata.MD, _ gax.CallSettings) error {
+		if rowLimitSet && numRowsRead == intialRowLimit {
+			return nil
```
andre-sampaio, line 406: If the number of rows requested (`initialRowLimit`) is higher than the total number of rows we will eventually receive, am I correct to assume that the block `if err == io.EOF {` would break us out of the loop?

If so, can we add a test with this scenario? (e.g. a table with 2 rows and set `rowLimit = 5`)

bhshkh: Added

bigtable/bigtable.go
bhshkh: Added

bigtable/bigtable.go
bhshkh: Added

bigtable/bigtable.go

bigtable/bigtable.go
bhshkh: Done

andre-sampaio (approved):

bigtable/bigtable.go
```diff
@@ -391,7 +391,25 @@ func (t *Table) ReadRows(ctx context.Context, arg RowSet, f func(Row) bool, opts
...
+	rowLimitSet := false
+	intialRowLimit := int64(0)
+	for _, opt := range opts {
+		if l, ok := opt.(limitRows); ok {
+			rowLimitSet = true
+			intialRowLimit = l.limit
+		}
+	}
+	if intialRowLimit < 0 {
+		return nil
```
gkevinzheng, line 405: Two questions:

1. Does it make sense to allow the row limit to be 0?
2. Should we return some kind of error here for the user when this trigger does happen instead of returning `nil`?

bhshkh: 1. yes, 0 means no limit similar to the proto: https://github.com/googleapis/google-cloud-go/blob/bigtable/v1.34.0/bigtable/apiv2/bigtablepb/bigtable.pb.go#L187-L189
2. Added error

gkevinzheng (approved):

PR #10235, opened 2024-05-20
chore: switch vetting from golint -> go vet

codyoss:


codyoss:
@codyoss is this still needed?

PR #10238, opened 2024-05-20
feat(pubsub): make lease management RPCs concurrent

### LINKED ISSUE #9727: pubsub: Many acks/modacks could cause other acks/modacks to be delayed
**Client**

PubSub 

**Environment**

CentOS on GCE (t2d-standard-4)

**Go Environment**

```
NumGoroutines=40
MaxExtension=15s
MaxOutstandingMessages=40000
MaxOutstandingBytes=-1
Synchronous=false
```

**Expected behavior**

I expect to see 0 in [subscription/expired_ack_deadlines_count](https://cloud.google.com/monitoring/api/metrics_gcp#pubsub/subscription/expired_ack_deadlines_count) assuming that my AckCount match my PullCount.

**Actual behavior**

We periodically see a huge rate of expired acks as high as 15k/s. We are currently acknowledging 20k messages per second across 2 GCE instances, ~10k/s per instance, and pulling 20k messages per second across those instances as well. I would expect then that we shouldn't really be seeing any expired Acks.

I don't know the actual distribution of messages across the 40 goroutines but if some of them are getting most of the messages then it's possible for the `ackIDBatchSize` to be exceeded. When it's exceeded, `sendModAck` and `sendAck` both loop internally until all of the ackIDs have been sent. We don't have visibility into the distribution of time it takes to `Acknowledge` 2500 ackIDs but we can see from the GCP console that the `Acknowledge` method has a 95th percentile latency of over 100ms. Separately, we are calling `ModifyAckDeadline` (at the 95th percentile latency takes 40ms) with 16k IDs per second which needs 7 calls per instance which could take more than 250ms+.

Either of those would end up delaying the other since there's only a single `sender` goroutine which could be contributing to our expired acks issue.

Additionally, since we aren't using exactly-once-delivery, there's no way for us to measure how long it took from when we called `Ack` to when the request was sent to Pub/Sub. One way to fix that would be if the `*AckResult` returned from `AckWithResult` would actually be `Ready` once the message is sent, even if you're not using exactly-once-delivery.

**Screenshots**

image
image

Something else that's interesting is that the latencies shown in the GCP Console do not match our application-level metrics (which measure from start of Receive callback to Ack/Nack function call) at all:
image
vs
image

This is what led us to investigate if there was some sort of delay between when we `Ack` a message and when the underlying `Acknowledge` is sent by the client.

ModAckCount
image
(the reason for the increase at 19:30 UTC is because we increased MaxOutstandingMessages from 25000 to 40000)

Finally, the increase in expired acks happened after a sharp decrease in the StreamingPull response for which I have no explanation unless some change was made on Pub/Sub's side. It's not clear if this might mean that there's a higher concentration of messages in individual goroutines.
image

**Additional context**

We don't have any visibility into the 99th percentile modack extension being used and that would have been helpful in debugging.

hongalex:
This PR makes the lease management RPCs concurrent rather than serial. The previous behavior takes a list of `ackIDs` that need to be sent out, takes the first `n` messages (where `n` is the max batch size), issues the RPC, and then repeats until `ackID` slice is empty.

This PR improves performance for large amounts of ackIDs by concurrently sending out RPCs in a goroutine, and waiting for them all to complete in a waitgroup. This means rewriting the `splitRequestIDs` function into a new `makeBatches` function, that does the splitting / batch making upfront, rather than per iteration.

Fixes #9727


alvarowolfx (approved):
overall looks good to me. One thing that got me a bit concerned is the logic on methods `sendAck` and `sendModAck` are almost the same and has duplicate code that can lead to it being out of sync sometimes. What differs is how to track telemetry, how to send the ack and how to retry. 

I have a suggestion, but I don't have the full context to know if those `ack` methods can diverse more in the future, but we can have a common `sendAckWithFunc` method that contains the logic and accept functions for `acking`, `retrying` and `recordTelemetry`.

Would be something like this:
```go
func (it *messageIterator) sendAck(m map[string]*AckResult) {
	it.sendAckWithFunc(m, func(ctx context.Context, subName string, ackIds []string) error {
		return it.subc.Acknowledge(ctx, &pb.AcknowledgeRequest{
			Subscription: it.subName,
			AckIds:       ackIds,
		})
	}, it.retryAcks, func(ctx context.Context, toSend []string) {
		recordStat(it.ctx, AckCount, int64(len(toSend)))
		addAcks(toSend)
	})
}

func (it *messageIterator) sendModAck(m map[string]*AckResult, deadline time.Duration, logOnInvalid bool) {
	deadlineSec := int32(deadline / time.Second)
	it.sendAckWithFunc(m, func(ctx context.Context, subName string, ackIds []string) error {
		return it.subc.ModifyAckDeadline(ctx, &pb.ModifyAckDeadlineRequest{
			Subscription:       it.subName,
			AckDeadlineSeconds: deadlineSec,
			AckIds:             ackIds,
		})
	}, func(toRetry map[string]*ipubsub.AckResult) {
		it.retryModAcks(toRetry, deadlineSec, logOnInvalid)
	}, func(ctx context.Context, toSend []string) {
		if deadline == 0 {
			recordStat(it.ctx, NackCount, int64(len(toSend)))
		} else {
			recordStat(it.ctx, ModAckCount, int64(len(toSend)))
		}
		addModAcks(toSend, deadlineSec)
	})
}

type ackFunc = func(ctx context.Context, subName string, ackIds []string) error
type ackRecordStat = func(ctx context.Context, toSend []string)
type retryAckFunc = func(toRetry map[string]*ipubsub.AckResult)

func (it *messageIterator) sendAckWithFunc(m map[string]*AckResult, ackFunc ackFunc, retryAckFunc retryAckFunc, ackRecordStat ackRecordStat) {
	ackIDs := make([]string, 0, len(m))
	for k := range m {
		ackIDs = append(ackIDs, k)
	}
	it.eoMu.RLock()
	exactlyOnceDelivery := it.enableExactlyOnceDelivery
	it.eoMu.RUnlock()
	batches := makeBatches(ackIDs, ackIDBatchSize)
	wg := sync.WaitGroup{}

	for _, batch := range batches {
		wg.Add(1)
		go func(toSend []string) {
			defer wg.Done()
			ackRecordStat(it.ctx, toSend)
			// Use context.Background() as the call's context, not it.ctx. We don't
			// want to cancel this RPC when the iterator is stopped.
			cctx, cancel2 := context.WithTimeout(context.Background(), 60*time.Second)
			defer cancel2()
			err := ackFunc(cctx, it.subName, toSend)
			if exactlyOnceDelivery {
				resultsByAckID := make(map[string]*AckResult)
					resultsByAckID[ackID] = m[ackID]
				}

				st, md := extractMetadata(err)
				_, toRetry := processResults(st, resultsByAckID, md)
				if len(toRetry) > 0 {
					// Retry modacks/nacks in a separate goroutine.
					go func() {
						retryAckFunc(toRetry)
					}()
				}
			}
		}(batch)
	}
}
```

hongalex:
I haven't been a huge fan of callback based systems since it makes jumping around in the IDE more difficult. With that said, the tradeoffs could be worth it in this case since there is a large number of overlap in code. I'm changing these functions a fair bit as part of the [otel tracing](https://github.com/googleapis/google-cloud-go/blob/bacda497d5f86ff7c6099ff0bfb7db9328f5f3c1/pubsub/iterator.go#L565) change, but I think it should still be mostly compatible even when wrapped with functions.

I think that if we end up needing more than these set of function arguments, it would be even less ergonomic but as-is, this seems like a good change.

PR #10263, opened 2024-05-24
feat(auth): Add non-blocking token refresh for compute MDS

quartzmo:


auth/auth.go
```diff
@@ -81,9 +100,9 @@ type Token struct {
 
 // IsValid reports that a [Token] is non-nil, has a [Token.Value], and has not
 // expired. A token is considered expired if [Token.Expiry] has passed or will
-// pass in the next 10 seconds.
+// pass in the next 225 seconds.
 func (t *Token) IsValid() bool {
-	return t.isValidWithEarlyExpiry(defaultExpiryDelta)
+	return t.isValidWithEarlyExpiry(defaultExpiryDelta) // TODO(quartzmo): investigate why EarlyTokenRefresh, ExpireEarly isn't used here. Bug?
```
codyoss, line 105: I wonder if this should have stayed 10 seconds and only the cached token source cares if this is different. This type does need to be wrapped with that layer.

auth/auth.go
```diff
@@ -206,11 +225,15 @@ func NewCredentials(opts *CredentialsOptions) *Credentials {
...
+	// even if it is expired. The default is false. Optional.
 	DisableAutoRefresh bool
 	// ExpireEarly configures the amount of time before a token expires, that it
-	// should be refreshed. If unset, the default value is 10 seconds.
+	// should be refreshed. If unset, the default value is 3 minutes and 45
+	// seconds. Optional.
 	ExpireEarly time.Duration
+	// BlockingRefresh configures a synchronous workflow that refreshes
+	// stale tokens while blocking. The default is false. Optional.
+	BlockingRefresh bool
```
codyoss, line 236: nit: I wonder if this should be DisableAysyncRefresh for parity with the other bool option on this struct. Thoughts?

auth/auth.go
```diff
@@ -227,33 +250,127 @@ func (ctpo *CachedTokenProviderOptions) expireEarly() time.Duration {
...
-	tp          TokenProvider
-	autoRefresh bool
-	expireEarly time.Duration
+	tp              TokenProvider
+	autoRefresh     bool
+	expireEarly     time.Duration
+	blockingRefresh bool
+	// isRefreshRunning ensures that the non-blocking refresh will only be
+	// attempted once, even if multiple callers enter the Token method.
+	isRefreshRunning bool
```
codyoss, line 287: please move any variables that need to be access under the mutex to under the mu-hat

auth/auth.go
```diff
@@ -227,33 +250,127 @@ func (ctpo *CachedTokenProviderOptions) expireEarly() time.Duration {
...
+	case stale:
+		c.tokenAsync(ctx)
+		// Return the stale token immediately to not block customer requests to Cloud services.
+		c.mu.Lock()
+		defer c.mu.Unlock()
+		return c.cachedToken, nil
+	case invalid:
+		return c.tokenBlocking(ctx)
+	default:
+		panic("unreachable")
```
codyoss, line 318: nit: make the default state `invalid`. Can always fallback to refreshing the token and avoids having to put a panic in the code.

auth/auth.go
```diff
@@ -41,9 +41,28 @@ const (
 	// so we give it 15 seconds to refresh it's cache before attempting to refresh a token.
 	defaultExpiryDelta = 225 * time.Second
 
+	// nonBlockingRefreshTimeout is the timeout for asynchronous refresh of a
+	// token.
+	nonBlockingRefreshTimeout = 30 * time.Second
```
codyoss, line 46: it looks like this was forgotten to be used.

auth/auth.go
```diff
@@ -227,33 +250,127 @@ func (ctpo *CachedTokenProviderOptions) expireEarly() time.Duration {
...
+			c.isRefreshErr = true
+			return
+		}
+		c.cachedToken = t
+	}
+	c.mu.Lock()
+	if !c.isRefreshRunning && !c.isRefreshErr {
+		go fn()
+	}
+	defer c.mu.Unlock()
```
codyoss, line 367: nit: don't defer or move this above the if.

auth/auth_test.go
```diff
@@ -506,3 +506,84 @@ func TestNew2LOTokenProvider_Validate(t *testing.T) {
...
+	defer func() { timeNow = time.Now }()
+	tp := NewCachedTokenProvider(&countingTestProvider{count: 1}, &CachedTokenProviderOptions{
+		// EarlyTokenRefresh ensures that token with expiry 1 second from now is already stale.
+		ExpireEarly: 2 * time.Second,
+	})
+	freshToken, err := tp.Token(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	if want := "1"; freshToken.Value != want {
```
codyoss, line 536: nit: can this test or another test be expanded to show the token hitting the different states? Since this is an internal test we should be able to safely access those internal bits. If done as a separate test you could keep changing timeNow() to make sure we hit all the states.

auth/auth_test.go
```diff
@@ -506,3 +506,84 @@ func TestNew2LOTokenProvider_Validate(t *testing.T) {
...
+		t.Errorf("got %q, want %q", freshToken.Value, want)
+	}
+	freshToken2, err := tp.Token(context.Background())
```
codyoss, line 582: nit: can we change the timing here to show the intermediate state the the token does cache still in blocking mode.

codyoss (changes requested):

auth/auth.go
quartzmo: I honestly think an implementation of `Token.IsValid` that calls `isValidWithEarlyExpiry` internally is not a correct implementation. Why is a token invalid if it will expire soon? Clearly it is not expired, and I personally think it is also not "invalid". I find the state definitions from the Async Refresh spec (implemented in this PR in `cachedTokenProvider.tokenState`) to be much clearer.

In the Python auth library, the existing `Credentials.valid` method [was deprecated](https://github.com/googleapis/google-auth-library-python/blob/cb982276e806c3f3c67faf85f009a9c688d084f4/google/auth/credentials.py#L95) in the Async Refresh PR, with the message: "Prefer checking :attr:`token_state` instead."

I'm not sure what to do about this, but I will remove this TODO from the PR.

auth/auth.go
quartzmo: It might be good for the options to be more different than alike, in order to avoid user bugs in which someone mistakenly types or copies the wrong option due to the similarity. `DisableAutoRefresh` and `DisableAsyncRefresh` are really visually similar. But if you think the parity is the better call, I'm more than happy to make the change.

auth/auth.go
quartzmo: Fixed, thanks!

auth/auth.go
quartzmo: Removed, thanks!

auth/auth.go
quartzmo: Done, thanks!

auth/auth.go
quartzmo: Done, thanks!

auth/auth.go
codyoss: I think I like the parity as the surface presents a clearer design choice for boolean options. I believe we have tried to use enabled disabled for most of our newer bool options.

quartzmo: SGTM, will do.

quartzmo: Done, thanks!

codyoss (approved):
Nice work!

auth/auth_test.go
quartzmo: Done, thanks!

PR #10265, opened 2024-05-25
chore(internal/kokoro): skip submodules repeated test if root directoy is already pushd

rahul2393:
In https://github.com/googleapis/google-cloud-go/pull/10251 there are repeated tests which internal/kokoro/presubmit.sh bash script was running.

Example:
```
CHANGED_DIRS='.
aiplatform
dlp
internal
managedkafka
spanner
video'
```
will run test twice for spanner once with "pushd ./spanner" and other with "pushd spanner"

We don't really need to run the sub-directories test if changes were detected in root directory(".") since it will anyways run all the tests in submodules.


noahdietz:
I'm not sure a change to a file in the top-dir (e.g. `doc.go`) should trigger tests in all submodules. This is a good call out though, I think we need to instead run tests in `.` directly but not `./...` from the top, if that makes sense e.g. `go test .`. There are no tests in the top but I suppose it will still build the code...WDYT?

codyoss:
./... is a module aware command and should only run in the scope of a module

rahul2393:
@codyoss they why did it ran tests twice on submodules? https://btx.cloud.google.com/invocations/de35c87c-677f-4eb7-9a48-4a7e77a94952/targets/cloud-devrel%2Fclient-libraries%2Fgo%2Fgoogle-cloud-go%2Fpresubmit%2Fearliest-version/log

codyoss (approved):

noahdietz (approved):

PR #10272, opened 2024-05-28
feat(ai/generativelanguage/apiv1beta): add SetGoogleClientInfo for all clients

jba:
Add exported SetGoogleClientInfo methods for all clients
in the API.

Also, add a script to generate the info.go file from the
clients.


ai/generativelanguage/apiv1beta/info.go
```diff
@@ -20,6 +20,33 @@
 
 package generativelanguage
 
-func (c *GenerativeClient) SetGoogleClientInfo(keyval ...string) {
-	c.setGoogleClientInfo(keyval...)
+//go:generate gen_info.sh > info.go
+
+func (c *DiscussClient) (keyval ...string) {
```
codyoss, line 25: Looks like the method name is missing 😉

codyoss (approved):

PR #10280, opened 2024-05-29
chore(internal/protoveneer): support Status

jba:
Support the Status type from
google.golang.org/genproto/googleapis/rpc/status.

We do not attempt to convert it to a more friendly type at this time.
First, although status.Status is a proto message, it does have two
fields, Code and Message, that most users should find it easy to
work with.

Second, if we made a new type, we'd have to decide where to put it.
It will be user-visible so the support package isn't the right place,
and it's a common type that occurs in many APIs, so adding it to the
current package being generated is also wrong. We'd need to make
a new package for well-known veneer types that correspond to
well-known proto types.

Third, the other field of Status is of type `[]*anypb.Any`. I have no
idea how to handle an anypb.Any in a nice way, because it can refer
to any proto message, using a specially encoded string to refer to the
type. Best to not touch it until we need to, or ever.


noahdietz:
> Third, the other field of Status is of type []*anypb.Any. I have no
idea how to handle an anypb.Any in a nice way, because it can refer
to any proto message, using a specially encoded string to refer to the
type. Best to not touch it until we need to, or ever.

Google has a [well-defined set of possible types](https://github.com/googleapis/go-genproto/blob/main/googleapis/rpc/errdetails/error_details.pb.go) that could appear in the `ErrorDetails []*anypb.Any`.

We also have a wrapper type that can parse these out: https://pkg.go.dev/github.com/googleapis/gax-go/v2@v2.12.4/apierror

I would recommend looking into exposing the error details - there is a large focus on making API errors more actionable and these details are the means of doing so.

noahdietz:
Also, the GAPIC already does the `APIError` parsing, if this client is using a GAPIC under the hood.

jba:
Thanks, Noah, very helpful info. I changed to APIError.
I don't think the GAPIC conversion will help because this is a proto field, not an RPC return value. Last I checked, GAPICs don't convert protos (hence the existence of protoveneer).
I will rewrite the commit message in the GitHub UI before submitting.

internal/protoveneer/support/support.go
```diff
@@ -119,3 +122,21 @@ func TimeFromProto(ts *timestamppb.Timestamp) time.Time {
...
+
+// APIErrorToProto attempts to convert an APIError to a proto Status.
+func APIErrorToProto(ae *apierror.APIError) *spb.Status {
+	if ae == nil {
+		return nil
+	}
+	return ae.GRPCStatus().Proto()
+}
+
+func APIErrorFromProto(s *spb.Status) *apierror.APIError {
```
jba, line 134: @noahdietz is there a more direct path here than from proto status -> grpc status -> error -> APIError?

eliben (approved):

internal/protoveneer/support/support.go
noahdietz: There isn't, but we can totally add one!

The use case for `APIError` first targeted wrapping API error responses which for GAPIC are always either a `grpc.Status` (wrapping the `proto.Status`) or a `googleapi.Error` (from the Discovery-based generated HTTP/JSON clients).

If you want to add it yourself, feel free, otherwise please [file an issue on gax-go](https://github.com/googleapis/gax-go/issues/new)

noahdietz:
> I don't think the GAPIC conversion will help because this is a proto field

I see. It might make sense to not use APIError in this case, I didn't look closely when I saw this PR. Kind of up to you how much you want users to have to do and if you think the surface of APIError makes sense to expose to your users.

PR #10288, opened 2024-05-30
feat(aiplatform/apiv1beta1): add methods to info.go

jba:
Add SetGoogleClientInfo methods for each client in the GAPIC.

Also, move the script that generates the file somewhere safe
(internal, for now). In #10272 we put it in the same directory
as the GAPIC, and an automatic process deleted it.

This means it's not really usable with `go generate`, because its
relative location varies by package, so remove that functionality.

### DIFF
--- a/aiplatform/apiv1beta1/info.go
+++ b/aiplatform/apiv1beta1/info.go
@@ -18,7 +18,7 @@
 //
 // Internal use only.
 
-package generativelanguage
+package aiplatform
 
 func (c *DatasetClient) SetGoogleClientInfo(keyval ...string) {
 	c.setGoogleClientInfo(keyval...)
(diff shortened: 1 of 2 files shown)


codyoss (dismissed):

aiplatform/apiv1beta1/info.go
```diff
@@ -18,8 +18,144 @@
 //
 // Internal use only.
 
-package aiplatform
+package generativelanguage
```
codyoss, line 21: incorrect package name

jba: Thanks, already pushed a fix.

codyoss (approved):

PR #10289, opened 2024-05-30
feat(spanner): add OrderBy feature

mayurkale22:
This feature is only applicable in Read API.

This option will allow users to specify whether they want the results to be returned in primary key order (PRIMARY_KEY) or in an out of order (NO_ORDER). The NO_ORDER is significantly faster than KEY_ORDER, because spanner uses higher parallelism for non-ordered reads.


spanner/transaction.go
```diff
@@ -285,6 +294,7 @@ func (t *txReadOnly) ReadWithOptions(ctx context.Context, table string, keys Key
 					RequestOptions:      createRequestOptions(prio, requestTag, t.txOpts.TransactionTag),
 					DataBoostEnabled:    dataBoostEnabled,
 					DirectedReadOptions: directedReadOptions,
+					OrderBy:             orderBy,
```
harshachinta, line 297: Shouldn't this be added to partitioned read interface as well? https://github.com/googleapis/google-cloud-go/blob/c15d69a8e71d1a7806e0b75bb6d62f499383a7ea/spanner/batch.go#L326

https://cloud.google.com/spanner/docs/samples/spanner-batch-client#spanner_batch_client-go

mayurkale22: This option does not applicable for partition reads, which is always act as if NO_ORDER was set.

spanner/transaction.go
```diff
@@ -258,6 +266,7 @@ func (t *txReadOnly) ReadWithOptions(ctx context.Context, table string, keys Key
 		if opts.DirectedReadOptions != nil {
 			directedReadOptions = opts.DirectedReadOptions
 		}
+		orderBy = opts.OrderBy
```
harshachinta, line 269: @mayurkale22 
Here the request level orderBy field in ReadOptions is directly overriding the client level orderBy field.
My concern here is, if a customer sets `client level` `orderBy` field to `NO_ORDER` and in every request if the customer passes ReadOptions with out setting `orderBy` field (which will by default have `UNSPECIFIED`) then this unset field will override the client level `NO_ORDER`.
Is this expected behavior?

Testcase:
```
{
name:   "Client level 1",
client: &ReadOptions{Index: "testIndex", Limit: 100, Priority: sppb.RequestOptions_PRIORITY_LOW, RequestTag: "testRequestTag", OrderBy: sppb.ReadRequest_ORDER_BY_NO_ORDER},
read:   &ReadOptions{Index: "testIndex", Limit: 100, Priority: sppb.RequestOptions_PRIORITY_LOW, RequestTag: "testRequestTag"},
want:   &ReadOptions{Index: "testIndex", Limit: 100, Priority: sppb.RequestOptions_PRIORITY_LOW, RequestTag: "testRequestTag", OrderBy: sppb.ReadRequest_ORDER_BY_NO_ORDER},
},
```

Here we expect the result to be `NO_ORDER`, but the actual output will be `UNSPECIFIED`.

harshachinta: In short, whenever request level ReadOptions is not null, then the default value of `orderBy` field will be passed to ReadRequest.

mayurkale22: Thanks for catching this. Added condition to set only when it is not default (UNSPECIFIED). PTAL

mayurkale22:
@harshachinta, @rahul2393 Could you please review and approve this PR?

rahul2393 (approved):

harshachinta (approved):

PR #10290, opened 2024-05-30
fix(auth): pass through client to 2LO and 3LO flows

codyoss:
This is especially important for mtls where clients need to be overridden to include certs


codyoss:
cc @andyrzhao

andyrzhao (approved):

PR #10297, opened 2024-05-31
chore(spanner): Pass JSON values through without remarshalling

c2nes:
In particular, this avoids numeric values having their representation changed in a way that Spanner is sensitive to (e.g. floating point values being reformatted as integers).


harshachinta:
Fixed in https://github.com/googleapis/google-cloud-go/pull/10327

PR #10310, opened 2024-06-04
feat(vertexai): explicit caching

jba:
Implement explicit caching.

- The CachedContent type.

- Methods to create, update, list and delete CachedContents.
  The update method follows the same pattern in some other cloud clients,
  where a separate type carries the fields to be updated.

- Two ways to set the CachedContent in a model:
  1. The GenerativeModel.CachedContentName field, which corresponds
     to the actual proto.
  2. The GenerativeModelFromCachedContent method, which requires
     a properly initialized CachedContent but guarantees that
     the model names of the CachedContent and GenerativeModel are
     the same.

### DIFF
--- a/vertexai/genai/caching.go
+++ b/vertexai/genai/caching.go
@@ -1,4 +1,4 @@
-// Copyright 2023 Google LLC
+// Copyright 2024 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/vertexai/genai/caching.go
+++ b/vertexai/genai/caching.go
@@ -50,11 +50,24 @@ func (c *Client) GenerativeModelFromCachedContent(cc *CachedContent) *Generative
 // CreateCachedContent creates a new CachedContent.
 // You can use the return value to create a model with [CachedContent.GenerativeModel].
 func (c *Client) CreateCachedContent(ctx context.Context, cc *CachedContent) (*CachedContent, error) {
+	if cc.Name != "" {
+		return nil, errors.New("genai.CreateCachedContent: do not provide a name; one will be generated")
+	}
+	pcc := cc.toProto()
+	pcc.Model = inferFullModelName(c.projectID, c.location, pcc.Model)
 	return c.cachedContentFromProto(c.cc.CreateCachedContent(ctx, &pb.CreateCachedContentRequest{
 		Parent:        c.parent(),
-		CachedContent: cc.toProto(),
+		CachedContent: pcc,
 	}))
 }
+
+// func (c *Client) inferFullCachedContentName(name string) string {
+// 	if strings.Contains(name, "/") {
+// 		return name
+// 	}
+// 	return fmt.Sprintf("projects/%s/locations/%s/cachedContents/%s",
+// 		c.projectID, c.location, name)
+// }
 
 // GetCachedContent retrieves the CachedContent with the given name.
 func (c *Client) GetCachedContent(ctx context.Context, name string) (*CachedContent, error) {
@@ -122,7 +135,6 @@ func (c *Client) cachedContentFromProto(pcc *pb.CachedContent, err error) (*Cach
 		return nil, err
 	}
 	cc := (CachedContent{}).fromProto(pcc)
-	cc.client = c
 	return cc, nil
 }
 
--- a/vertexai/genai/caching.go
+++ b/vertexai/genai/caching.go
@@ -48,7 +48,18 @@ func (c *Client) GenerativeModelFromCachedContent(cc *CachedContent) *Generative
 }
 
 // CreateCachedContent creates a new CachedContent.
-// You can use the return value to create a model with [CachedContent.GenerativeModel].
+// The argument should contain a model name and some data to be cached, which can include
+// contents, a system instruction, tools and/or tool configuration. It can also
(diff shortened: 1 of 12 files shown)


jba:
DO NOT SUBMIT. Feature not yet available.

vertexai/genai/client.go
```diff
@@ -103,6 +115,10 @@ func inferLocation(location string) string {
 	return defaultLocation
 }
 
+func (c *Client) parent() string {
```
eliben, line 118: not sure what `parent` means in this context, since this method just returns a string. Add a comment to clarify?

vertexai/genai/caching.go
```diff
@@ -0,0 +1,166 @@
...
+		c:                 c,
+		name:              cc.Model,
+		fullName:          inferFullModelName(c.projectID, c.location, cc.Model),
+		CachedContentName: cc.Name,
+	}
+}
+
+// CreateCachedContent creates a new CachedContent.
+// You can use the return value to create a model with [CachedContent.GenerativeModel].
+func (c *Client) CreateCachedContent(ctx context.Context, cc *CachedContent) (*CachedContent, error) {
```
eliben, line 52: This creates a new CachedContent, but also takes a CachedContent -- why? The docstring should explain in detail

vertexai/genai/caching_test.go
eliben, line 1: Please add an example in `example_test` too; examples have a much more linear and clear code flow which makes them good for evaluating the new API from a user's pov

vertexai/genai/client.go
jba: Done.

vertexai/genai/caching_test.go
jba: Done.

vertexai/genai/caching.go
eliben: Nit: can you clarify the "The cached content will not be returned" part? This method returns a `CachedContent`, so on a first read this is confusing. Do you mean to say that some fields of the returned `CachedContent` will not be populated?

eliben (approved):

vertexai/genai/client.go
jba: It comes from the proto field. It isn't the endpoint or the full name of the resource.

It doesn't need to infer anything because it's only used internally, after the project ID and location have been inferred.

vertexai/genai/caching.go
jba: Done.

vertexai/genai/caching.go
```diff
@@ -0,0 +1,192 @@
...
+type cacheClient = aiplatform.GenAiCacheClient
+
+var (
+	newCacheClient     = aiplatform.NewGenAiCacheClient
+	newCacheRESTClient = aiplatform.NewGenAiCacheRESTClient
+)
+
+// GenerativeModelFromCachedContent returns a [GenerativeModel] that uses the given [CachedContent].
+// The argument should come from a call to [Client.CreateCachedContent] or [Client.GetCachedContent].
+func (c *Client) GenerativeModelFromCachedContent(cc *CachedContent) *GenerativeModel {
```
jba, line 41: @eliben WDYT about giving this an error return, and checking that cc.Name and cc.Model are non-empty?

jba: I guess we never do any such checks on the client, even for an empty argument to `Client.GenerativeModel`. So we shouldn't do one here.

jba:
@codyoss Do you see the problem with kokoro?

codyoss:
@jba 

```
genai/aiplatformpb_veneer.gen.go:123:39: undefined: pb.CachedContent
genai/aiplatformpb_veneer.gen.go:127:11: undefined: pb.CachedContent
genai/aiplatformpb_veneer.gen.go:141:38: undefined: pb.CachedContent
genai/aiplatformpb_veneer.gen.go:127:11: too many errors
FAIL	cloud.google.com/go/vertexai/genai [build failed]
?   	cloud.google.com/go/vertexai/internal	[no test files]
```

eliben:
> @jba
> 
> ```
> genai/aiplatformpb_veneer.gen.go:123:39: undefined: pb.CachedContent
> genai/aiplatformpb_veneer.gen.go:127:11: undefined: pb.CachedContent
> genai/aiplatformpb_veneer.gen.go:141:38: undefined: pb.CachedContent
> genai/aiplatformpb_veneer.gen.go:127:11: too many errors
> FAIL	cloud.google.com/go/vertexai/genai [build failed]
> ?   	cloud.google.com/go/vertexai/internal	[no test files]
> ```

@codyoss could you describe the path of mouse clicks taking you to this information from the kokoro "Details" page?

codyoss:
Click the details button next to kokoro:
Screenshot 2024-06-10 at 3 32 39 PM

Copy and paste the link to the failing build:
Screenshot 2024-06-10 at 3 33 03 PM

jba:
I guess this builds locally because of go.work.
To get the new symbols we should make a new release of aiplatform.
#9772 is the PR. It should be unblocked. I just approved it but its Do Not Merge label should be removed.

eliben:
Since @jba is now OOO, I sent branched this PR and sent #10363 with the same contents, except updating the `aiplatform` dependency

eliben:
Merged in #10363, which was released in v0.11.0

PR #10312, opened 2024-06-04
chore(internal/protoveneer): group imports

jba:
Print import paths in two groups: one for the standard library,
and one for other imports.

This ensures that the goimports test run as part of the "go vet"
action of this repo will succeed on generated files.


internal/protoveneer/cmd/protoveneer/protoveneer.go
```diff
@@ -970,3 +982,12 @@ func sliceAnyError[T any](s []T, f func(T) (bool, error)) (bool, error) {
 	}
 	return false, nil
 }
+
+// inStdLib reports whether the given import path could be part of the Go standard library,
+// by reporting whether the first component lacks a '.'.
+func inStdLib(path string) bool {
+	if i := strings.IndexByte(path, '/'); i != -1 {
```
eliben, line 989: Micro-nit: it's possible this could be a bit cleaner/shorter with `strings.Cut` now

eliben (approved):

codyoss (approved):

internal/protoveneer/cmd/protoveneer/protoveneer.go
jba: Will do in a followup.

jba: Actually, did it here.

PR #10318, opened 2024-06-04
testing(bigquery/storage/managedwriter): dynamic schema test fixes

shollyman:
This PR changes the dynamic schema testing to look more like a realistic example.  Rather than build schemas from a reference bigquery.Schema, we use the GetWriteStream method to fetch schema for a stream, and build the necessary descriptor / descriptorproto based on the RPC.


go.work.sum
alvarowolfx, line 1: probably needs to remove this file from the PR

alvarowolfx (approved):

PR #10329, opened 2024-06-05
chore(storage/transfermanager): add usage metric headers

BrennaEpp:
- Also update storage to merge the header


storage/transfermanager/downloader.go
```diff
@@ -306,3 +310,16 @@ type DownloadOutput struct {
...
+const (
+	xGoogHeaderKey  = "x-goog-api-client"
+	downloadMany    = "download_many"
+	downloadSharded = "download_sharded"
+)
+
+// Sets invocation ID headers on the context which will be propagated as
+// headers in the call to the service (for both gRPC and HTTP).
+func setUsageMetricHeader(ctx context.Context, method string) context.Context {
+	header := fmt.Sprintf("gccl-gcs-cmd/tm.%s", method)
```
tritone, line 323: nit: I'd also make the key a const.

storage/invoke.go
```diff
@@ -84,7 +84,21 @@ func setInvocationHeaders(ctx context.Context, invocationID string, attempts int
...
+		xGoogHeader = strings.Join([]string{xGoogHeader, v}, " ")
+	}
+
+	if hdrs[xGoogHeaderKey] != nil {
+		// Replace the key instead of adding it, if there was anything to merge with.
+		hdrs[xGoogHeaderKey] = []string{xGoogHeader}
+	} else {
+		// TODO: keep this line when removing the above code.
+		ctx = callctx.SetHeaders(ctx, xGoogHeaderKey, xGoogHeader)
+	}
```
tritone, line 100: This looks right but could you add a test? Possibly could be added as a case in the existing tests for invoke, but if a separate one is easier that's fine too.

tritone (commented):
Just needs a unit test, otherwise looks good.

storage/invoke.go
BrennaEpp: Added. I think a separate one is better because it should be removed with the code.

storage/invoke_test.go
```diff
@@ -298,6 +298,68 @@ func (f *fakeApiaryRequest) Header() http.Header {
 	return f.header
 }
 
+func TestInvokeHeaderMerge(t *testing.T) {
```
tritone, line 301: Could you add a comment explaining what this is testing for? Probably a little unintuitive to anyone who is not familiar.

tritone (approved):
LGTM, just one more suggestion.

tritone (approved):

Issue #10326, opened 2024-06-05
logging: prioritize GOOGLE_CLOUD_PROJECT env variable over project id from metadata server for GKE resource detection


GoosvandenBekerom:
**Is your feature request related to a problem? Please describe.**
In our company, we have a central project in which GKE is managed, but then every application has its own project which gets it's own namespace in that central GKE setup. Now we are trying to integrate with the central logging architecture provided by our central TPS team.

Everything works, but there is a slight problem for us in the autodetection of the MonitoredResource. It detects the central project instead of the project that owns the workload. Because this central team wants to forward logging bills to the projects that are actually doing the request, they enforce us to provide the project id in the resource.project_id field.

I've tried a couple things, without success: 
- Use the `logging.DetectProjectID` setup
- Only setting the project_id field and hope the auto-detection will set the rest (this was a desperate attempt but I wanted to make sure it didn't work before filing this issue)

**Describe the solution you'd like**
The Java library used by other teams in our company seems to first check for the "GOOGLE_CLOUD_PROJECT" environment variable before going to the metadata server. If we also do this here (at least for the GKE detection) this would solve our issue.

**Describe alternatives you've considered**
I want a way to override the detected `resource.project_id` field for log entries.
We could implement a new [LoggerOption](https://github.com/googleapis/google-cloud-go/blob/main/logging/loggeroption.go#L25) like `logging.OverrideResourceProjectId(projectID string)`. which would then set the "project_id" label on `l.commonResource` to the given projectID.

The only other alternative I see for us is to write our own auto-detection setup, which I really dislike, as this feels like something that will change often.

**Additional context**
I would be happy to implement this feature if maintainers agree that it would fit.

gkevinzheng:
@GoosvandenBekerom Is the Java library that you mention the Java Cloud Logging client library or some other proprietary library that interfaces with Cloud Logging?

GoosvandenBekerom:
From what I understood from our jvm team, it uses this method from the java google cloud client https://googleapis.dev/java/google-cloud-clients/0.98.0-alpha/com/google/cloud/ServiceOptions.html#getDefaultProjectId--

Not sure which logging client they use, I'll try to find out.

gkevinzheng:
Is this overwriting something that is happening to the `MonitoredResource` object? In the Go logging library the detected resource isn't exposed as an object.

GoosvandenBekerom:
[...]

PR #10342, opened 2024-06-06
feat(logging): prioritize GOOGLE_CLOUD_PROJECT env variable over project id from metadata server for GKE resource detection

GoosvandenBekerom:
#10326


codyoss:
corresponding issue was closed.

PR #10349, opened 2024-06-07
fix(datastore): Update retry transaction logic to be inline with Spanner

bhshkh:
Fixes [b/335924967](http://b/335924967)
Modified RunInTransaction behaviour to match [go/transaction-retry-matrix](http://go/transaction-retry-matrix)
There are still inconsistencies between the changes in this PR and the retry matrix:
1) f (function provided by user in RunInTransaction) is retried n number of times, where n is provided by user using MaxAttempts transaction option with default value of 3 but the retry matrix states that transactions should be retried max 5 times. This is deliberate deviation. If the limit on MaxAttempts is introduced now, the existing user code with MaxAttempts higher than 5 will break

2) Even without the changes in this PR, Datastore client retries all the RPCs on Unavailable error until RPC succeeds or returns error other than Unavailable. This will be taken care of separately ([b/322562199](http://b/322562199)).


cindy-peng (approved):

gkevinzheng (approved):
LGTM

PR #10355, opened 2024-06-11
feat(vertexai/genai): add MergedResponse method to GenerateContentResponseIterator

randall77:
This method adds the MergedResponse method so we can get at the final request state after all the streaming responses have been accumulated.

This PR parallels the same addition in googleai:
https://github.com/google/generative-ai-go/commit/8decc9df4add211154198f110a048222e7c23946


eliben (approved):

eliben:
@codyoss do you know why Kokoro has been stuck here since yesterday? What prevents this PR from becoming merge-able?

codyoss:
@eliben As a safety feature our CI does not run by default for those without write access. Adding the force-run label will get it going though.

Issue #10031, opened 2024-04-23
spanner: add better support for reading a row that may not exist


husam-e:
**Is your feature request related to a problem? Please describe.**
`ReadRow` documentation states that it returns a NotFound error if the row is not found.
However, we've discovered that there are other scenarios where NotFound can be returned that aren't documented, such as when a _column_ is not found. This was difficult to find given we had added special logic for handling NotFound to return `nil, nil` instead. This thus makes handling the typical code path of doing a "FetchOne", then checking if it exists, extra hard (as not only is it returning an error instead of just nil, but the error is ambiguous).

**Describe the solution you'd like**
Ideally, `ReadRow` would not fail simply on not finding the entity. It would fail when there's a real error (not finding the row is arguably nominal and typically handled by returning null or equivalent). A variant like `ReadRowOrNil` would be ideal. 

Otherwise, provide some robust way to determine when the row is not found vs when there's an actual error. 
Another thing to consider is whether a schema issue like a missing column should be treated as NotFound, as that is not intuitive to me, since my assumption is that the "NotFound" pertains to the Row, not some underlying schema detail.

**Describe alternatives you've considered**
Matching on the string message e.g. "row not found", which is not ideal and preferable to avoid.

**Additional context**
NA

husam-e:
Hi @rahul2393 , thanks for getting back to me. 

Using those lower level APIs with an iterator is a lot less ergonomic and more error prone for a simple maybe-read-one-row use case, as it has more boilerplate code.

I'm hoping for a higher level API providing a straightforward way to reliably query for one row and know when it doesn't exist. It's a very common use case, so I think all users will benefit.

What are your thoughts on adding such an API to the sdk? If I find time I can perhaps create a PR to demonstrate what I'm thinking.

rahul2393:
We do also support GORM which exposes higher level APIs example https://gorm.io/docs/query.html#Retrieving-a-single-object https://github.com/googleapis/go-gorm-spanner/tags.

But please help with creating a PR to get more understanding.
Thanks

husam-e:
@rahul2393 I created a draft PR (#10358) for illustration purposes of 2 options, lmk wyt

PR #10358, opened 2024-06-11
feat(spanner): demonstrate enhanced row not found handling options #10031

husam-e:
This change is just an example (that may not build successfully yet) demonstrating 2 options for enhancing row not found handling, which is currently not explicit when using `ReadRow` / `ReadRowUsingIndex`, as discussed in #10031:

1. Setting the returned APIError's `Details.ErrorInfo.Reason` to a well-defined value that can be checked by users to reliably determine if an error is specifically due to a row not being found (as opposed to column/table not found which returns the same error code).
2. Introducing an additional API that returns `nil` when the row is not found (called `ReadRowOptional`). An equivalent API would be added for `ReadRowUsingIndex` (`ReadRowOptionalUsingIndex`).

### DIFF
--- a/spanner/transaction.go
+++ b/spanner/transaction.go
@@ -28,9 +28,11 @@ import (
 	"google.golang.org/api/iterator"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/protoadapt"
 
 	vkit "cloud.google.com/go/spanner/apiv1"
 	durationpb "google.golang.org/protobuf/types/known/durationpb"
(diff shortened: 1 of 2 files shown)


spanner/transaction.go
```diff
@@ -326,7 +328,8 @@ func (t *txReadOnly) ReadWithOptions(ctx context.Context, table string, keys Key
 // errRowNotFound returns error for not being able to read the row identified by
 // key.
 func errRowNotFound(table string, key Key) error {
-	return spannerErrorf(codes.NotFound, "row not found(Table: %v, PrimaryKey: %v)", table, key)
+	details := []protoadapt.MessageV1{&errdetails.ErrorInfo{Reason: RowNotFoundReason}}
```
husam-e, line 331: option 1: specify an explicit and exported Reason value in `details.ErrorInfo.Reason`, and do the same for `ReadRowUsingIndex`

spanner/transaction.go
```diff
@@ -348,10 +351,38 @@ func errInlineBeginTransactionFailed() error {
...
 	return t.ReadRowWithOptions(ctx, table, key, columns, nil)
 }
 
+// ReadRowOptional reads a single row from the database.
+//
+// If no row is present with the given key, then ReadRowOptional returns nil.
+// This is a convenience function to simplify use cases that want to treat 
+// a missing row as nil in a reliable manner, without matching against the 
+// error message.
+func (t *txReadOnly) ReadRowOptional(ctx context.Context, table string, key Key, columns []string) (*Row, error) {
```
husam-e, line 378: option 2: additional API for treating row not found as nil and not an error

husam-e:
Addressed in https://github.com/googleapis/google-cloud-go/pull/10405, thanks @rahul2393 !

PR #10362, opened 2024-06-11
fix(vertexai): don't add empty Text parts to session history

### LINKED ISSUE #10309: genai: Gemini Function Calling Fails with "Unable to submit request because it has an empty text parameter"
**Client**

Client: Vertex AI, Gemini, genai

**Environment**

Environment: macOS 14.5

**Go Environment**

go version go1.22.2 darwin/arm64

**Code**

```go
package main

import (
	"context"
	"fmt"
	"os"

	"cloud.google.com/go/vertexai/genai"
	"gopkg.eu.org/envloader"
)

func main() {
	envloader.LoadEnvFile(".env")
	ctx := context.Background()

	client, err := genai.NewClient(ctx, os.Getenv("PROJECT_ID"), os.Getenv("LOCATION"))
	if err != nil {
		panic(err)
	}
	defer client.Close()

	model := client.GenerativeModel("gemini-1.5-flash-001")
	model.Tools = []*genai.Tool{
		{
			FunctionDeclarations: []*genai.FunctionDeclaration{
				{
					Name:        "get_weather",
					Description: "Retrieves the current weather conditions for a specified location.",
					Parameters: &genai.Schema{
						Type: genai.TypeObject,
						Properties: map[string]*genai.Schema{
							"location": {
								Type:        genai.TypeString,
								Description: "The city and state, for example: San Francisco, CA",
							},
						},
						Required: []string{"location"},
					},
				},
			},
		},
	}
	model.SystemInstruction = &genai.Content{Parts: []genai.Part{
		genai.Text("You are a helpful and informative AI assistant. Provide a clear explanation of each step in the reasoning process before presenting your response. Always start your response with a <reasoning> block."),
	}}

	chat := model.StartChat()

	stream := chat.SendMessageStream(ctx, genai.Text("What's the weather like in Mountain View?"))
	for {
		resp, err := stream.Next()
		if err == iterator.Done {
			break
		}
		if err != nil {
			panic(err)
		}
		for _, cand := range resp.Candidates {
			for _, part := range cand.Content.Parts {
				switch p := part.(type) {
				case genai.Text:
					fmt.Print(p)
				case *genai.FunctionCall:
					fmt.Println()
					fmt.Println(p)
				default:
					fmt.Println(p)
				}
			}
		}
	}
	fmt.Println()

	function_response := &genai.FunctionResponse{
		Name: "get_weather",
		Response: map[string]any{
			"temperature":      75.5,
			"temperature_unit": "fahrenheit",
			"humidity":         "55%",
		},
	}

	stream = chat.SendMessageStream(ctx, function_response)
	for {
		resp, err := stream.Next()
		if err == iterator.Done {
			break
		}
		if err != nil {
			panic(err)
		}
		for _, cand := range resp.Candidates {
			for _, part := range cand.Content.Parts {
				switch p := part.(type) {
				case genai.Text:
					fmt.Print(p)
				default:
					fmt.Println()
					fmt.Println(p)
				}
			}
		}
	}
	fmt.Println()
}
```

**Expected behavior**

The code should successfully send the function_response to the Gemini model and receive a response based on the provided weather information.

**Actual behavior**

The code panics with the following error when sending the function_response:

```
<reasoning>
I can use the `default_api.get_weather()` function to retrieve the current weather conditions for Mountain View.
</reasoning>
{get_weather map[location:Mountain View, CA]}

panic: rpc error: code = InvalidArgument desc = Unable to submit request because it has an empty text parameter. Add a value to the parameter and try again. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini

goroutine 1 [running]:
main.main()
        /Users/username/work/bugreport/main.go:91 +0x894
exit status 2
```

**Additional context**

```
require (
	cloud.google.com/go/aiplatform v1.67.0
	cloud.google.com/go/vertexai v0.10.0
)
```

eliben:
Fixes #10309


vertexai/genai/chat.go
```diff
@@ -59,15 +59,28 @@ func (cs *ChatSession) SendMessageStream(ctx context.Context, parts ...Part) *Ge
...
 	if len(cands) > 0 {
 		c := cands[0].Content
 		if c == nil {
-			return false
+			return
 		}
-		c.Role = roleModel
-		cs.History = append(cs.History, c)
-		return true
+		cs.History = append(cs.History, copySanitizedModelContent(c))
```
codyoss, line 68: is it valid to append Content that could contain no parts? I am thinking if c.Parts only contains empty Text.

vertexai/genai/chat.go
```diff
@@ -59,15 +59,28 @@ func (cs *ChatSession) SendMessageStream(ctx context.Context, parts ...Part) *Ge
...
 	}
-	return false
+}
+
+// copySanitizedModelContent creates a (shallow) copy of c with role set to
+// model and empty text parts removed.
+func copySanitizedModelContent(c *Content) *Content {
+	newc := &Content{Role: roleModel}
+	for _, part := range c.Parts {
+		if t, ok := part.(Text); ok {
```
jba, line 77: if t, ok := ...; !ok || len(..) > 0 {append}

Then you don't need the else.

vertexai/genai/chat.go
```diff
@@ -59,15 +59,28 @@ func (cs *ChatSession) SendMessageStream(ctx context.Context, parts ...Part) *Ge
...
-		cs.History = append(cs.History, c)
-		return true
+		cs.History = append(cs.History, copySanitizedModelContent(c))
 	}
-	return false
+}
+
+// copySanitizedModelContent creates a (shallow) copy of c with role set to
+// model and empty text parts removed.
+func copySanitizedModelContent(c *Content) *Content {
```
jba, line 74: Worth its own test. You can't be sure you'll see this talking to a real model.

vertexai/genai/chat.go
eliben: Yes, I think it's OK for now, because this comes from the server. There should be an error somewhere if the server returns completely empty contents.

eliben:
Thanks @codyoss and @jba 
I will address the comments in a follow-up

PR #10364, opened 2024-06-11
feat(vertexai): export merged response

jba:
Allow users to get the merged response from an iterator.


eliben:
I think this is a duplicate of #10355 
Please reopen if I'm missing something

PR #10373, opened 2024-06-12
feat(auth): adds support for X509 workload identity federation

aeitzman:


codyoss:
cc @andyrzhao

auth/credentials/internal/externalaccount/externalaccount.go
```diff
@@ -204,11 +204,24 @@ func NewTokenProvider(opts *Options) (auth.TokenProvider, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	client := opts.Client
```
codyoss, line 208: I think, at least sometimes, this needs to win over xp.client(). The way the code is today it is not possible to tell if the client in opts is user specified or the default. If it is the default client it is fine to override it, but if the user explicitly passed a client in it feels like we should honor it no matter what. Thoughts?

auth/credentials/internal/externalaccount/x509_provider.go
```diff
@@ -0,0 +1,60 @@
...
+	certificateConfigLocation string
+	cachedClient              *http.Client
+}
+
+func (xp *x509Provider) providerType() string {
+	return x509ProviderType
+}
+
+func (xp *x509Provider) subjectToken(ctx context.Context) (string, error) {
+	return "", nil
```
codyoss, line 36: I wonder if this is the wrong abstraction. If there is no subject token this is not really a provider of tokens.

aeitzman: So rename this to emptyProvider or something like that and then have a seperate type to handle the client creation?

auth/credentials/internal/externalaccount/externalaccount.go
aeitzman: I did think about that, seems to me like our options are:

1: Refactor so we are passing some signal down from the detect call that the default client is being used
2: Do a check to see if the client used is the same as the default (via reflect.deepequals) this seems hacky, but would theoretically work
3: The implementation here where we always override. I could add some log message to make it clear what's happening if that's better. 

I chose 3 here since I didn't actually see a use case where a user would want to use a different client. If they are passing in a client with TLS already set up, it defeats the purpose of using the X509 credential type, since the credentials doesn't actually use a 3rd party subject token, all of the work is in the cert/client creation logic itself

aeitzman: Modified to track whether the default client is being used, and only override if that is true

auth/credentials/internal/externalaccount/x509_provider.go
codyoss: It feels like it should just not impl this interface because it is not a provider of token.

auth/credentials/internal/externalaccount/externalaccount.go
codyoss: lgtm

codyoss:
@andyrzhao Would you mind looking at this too since you are active in this space.

auth/credentials/internal/externalaccount/x509_provider.go
andyrzhao: It seems like exchanging subject token with access token is the core focus of the "externalaccount" credentials type, and since x509 is not doing that, it should sit outside of "externalaccount" entirely, and be it's own credentials type? I'm not familiar with the original externalaccount effort so this is just based on what I can observe from this PR.

auth/credentials/internal/externalaccount/url_provider.go
```diff
@@ -30,6 +30,7 @@ const (
 	fileTypeJSON             = "json"
 	urlProviderType          = "url"
 	programmaticProviderType = "programmatic"
+	x509ProviderType         = "x509"
```
andyrzhao, line 33: side comment: I don't think url_provider is the right location to define the list of provider types.

auth/credentials/internal/externalaccount/x509_provider.go
aeitzman: This should definitely live in externalaccount, the point of externalaccount is not necessarily to just exchange subject tokens, its to exchange some 3rd party credential for a GCP credential. In the past it has always been subject tokens, but with the new X509 credential type, a certificate is being used. This also maps to the external_account type in the ADC configuration, which the new X509 still falls under.

If we don't have a provider here that implements the interface, would you prefer just returning an anonymous function in NewTokenProvider that returns an empty string? @codyoss I don't have strong feelings either way, we just need something that implements subjectTokenProvider to use when createing a new tokenProvider.

andyrzhao: Gotcha. I think it may be more clear to introduce a concept of an "emptySubjectTokenProvider" to use along with X509 logic. I understand the existing code is structured to always expect a "subject token", but at the end of the day we really only care about implementing the auth.TokenProvider interface. I'll let you and Cody make the call.

auth/credentials/internal/externalaccount/url_provider.go
codyoss: I can fix this up in a future PR, not sure why this landed here 😄

codyoss (approved):

andyrzhao (approved):
lgtm

PR #10379, opened 2024-06-13
feat(storage/transfermanager): automatically shard downloads

BrennaEpp:
This is missing a few components that should be added in follow ups:
- checksums
- transcoding test


storage/transfermanager/downloader.go
```diff
@@ -155,46 +173,110 @@ func (d *Downloader) downloadWorker() {
...
+
+// gatherShards receives from the given channel exactly (shards-1) times (since
+// the first shard should already be complete).
+// It will add the result to the Downloader once it has received all shards.
+// gatherShards cancels remaining shards if any shard errored.
+func (d *Downloader) gatherShards(in *DownloadObjectInput, outs <-chan *DownloadOutput, shards int) {
+	errs := []error{}
+	var shardOut *DownloadOutput
+	for i := 1; i < shards; i++ {
+		// Add monitoring here? This could hang if any individual piece does.
```
tritone, line 243: By monitoring do you mean metrics or some kind of go routine to monitor?

storage/transfermanager/downloader.go
```diff
@@ -306,3 +396,76 @@ type DownloadOutput struct {
...
+	}
+	return b
+}
+
+// numShards calculates how many shards the given range should be divided into
+// given the part size.
+func numShards(attrs *storage.ReaderObjectAttrs, r *DownloadRange, partSize int64) int {
+	objectSize := attrs.Size
+
+	// Transcoded objects do not support ranged reads.
```
tritone, line 413: This should be documented in the public docs as well.

storage/transfermanager/downloader.go
```diff
@@ -306,3 +396,76 @@ type DownloadOutput struct {
...
+	// Negative offset reads the whole object in one go.
+	if r.Offset < 0 {
+		return 1
+	}
+
+	firstByte := r.Offset
+
+	// Read to the end of the object, or, if smaller, read only the amount
+	// of bytes requested.
+	lastByte := min(objectSize-1, r.Offset+r.Length-1)
```
tritone, line 431: Is this really relevant if we are just calculating number of shards? The math seems a bit overcomplicated.

storage/transfermanager/downloader_test.go
```diff
@@ -39,3 +42,338 @@ func TestWaitAndClose(t *testing.T) {
 		t.Errorf("expected err %q, got: %v", expectedErr, err.Error())
 	}
 }
+
+func TestNumShards(t *testing.T) {
```
tritone, line 46: Thanks for adding these!

storage/transfermanager/option.go
```diff
@@ -69,10 +69,30 @@ func (wpt withPerOpTimeout) apply(tm *transferManagerConfig) {
 	tm.perOperationTimeout = wpt.timeout
 }
 
+// WithPartSize returns a TransferManagerOption that specifies the size of the
+// shards to transfer; that is, if the object is larger than this size, it will
+// be uploaded or downloaded in concurrent pieces.
+// The default is 32 MiB for downloads.
```
tritone, line 75: Is this based on other langs' default?

storage/transfermanager/downloader.go
```diff
@@ -155,46 +173,110 @@ func (d *Downloader) downloadWorker() {
...
+
+				if numShards <= 1 {
+					// Download completed with a single shard.
+					d.addResult(input, out)
+				} else {
+					// Queue more shards.
+					outs := d.queueShards(input, out.Attrs.Generation, numShards)
+					// Start a goroutine that gathers shards sent to the output
+					// channel and adds the result once it has received all shards.
+					go d.gatherShards(input, outs, numShards)
```
tritone, line 193: Do we need to block somewhere on this being completed if we're in async mode?

tritone (commented):
Good start on this, a few initial comments...

storage/transfermanager/download_buffer.go
```diff
@@ -0,0 +1,71 @@
...
+// WriteAt writes len(p) bytes from p to the underlying buffer at offset off,
+// growing the buffer if needed. It returns the number of bytes written from p
+// and any error encountered that caused the write to stop early.
+// WriteAt is thread-safe as long as the ranges being written to do not overlap.
+// The supplied slice p is not retained.
+func (db *DownloadBuffer) WriteAt(p []byte, off int64) (n int, err error) {
+	requiredLength := int64(len(p)) + off
+
+	// Our buffer isn't big enough, let's grow it.
+	if int64(cap(db.bytes)) < requiredLength {
```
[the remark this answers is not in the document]
BrennaEpp: It's completely possible that that happens (depends on partsize), but I think we also don't want to increment by a lot more than is needed. I was initially thinking to double it as `append` does, but that could more easily run into problems with large buffers.

To avoid a lot of small increments, users could pass in a buffer that's already close to the object size to NewDownloadBuffer, or we could schedule the last shard first, so that it grows the buffer all it needs to since the beginning (but that has other potential performance implications).  Another option is to give users the ability to configure how much the buffer grows by.

storage/transfermanager/downloader.go
BrennaEpp: A go routine, but it might not be necessary as everything should cancel when the ctx does (so timeouts would stop this from hanging)

storage/transfermanager/option.go
BrennaEpp: Yes, this is based off of Python. We should do some perf testing and change this accordingly.

storage/transfermanager/downloader.go
BrennaEpp: We block in WaitAndClose until all results for each object that was added to the input are added to the result array or their callback is executed. 

I should note in the docs that in async mode WaitAndClose only finished until callbacks do (or we can start another goroutine to do the callback).

storage/transfermanager/downloader.go
BrennaEpp: It's very possible it's overcomplicated, I'll take a look and see if I can make it simpler.

BrennaEpp: So I can maybe make the code easier to follow (probably by removing that use of `min` and replacing with a conditional) and pop it out into a separate function, but given the edge cases I don't see a less roundabout way of calculating the number of shards. Am I missing something?

storage/transfermanager/downloader.go
```diff
@@ -235,15 +320,20 @@ type DownloadObjectInput struct {
 	// Callback will be run once the object is finished downloading. It must be
 	// set if and only if the [WithCallbacks] option is set; otherwise, it must
 	// not be set.
+	// A worker will be used to execute the callback; therefore, it should not
+	// be a long-running function. WaitAndClose will wait for all callbacks to
+	// finish.
```
BrennaEpp, line 325: Please review this wording.

storage/transfermanager/downloader_buffer_test.go
```diff
@@ -15,10 +15,133 @@
...
+	"time"
+
+	"github.com/google/go-cmp/cmp"
 )
 
 func TestDownloadBuffer(t *testing.T) {
 	t.Parallel()
-	// Unit test DownloadBuffer
+
+	// Create without an underlying buffer.
```
tritone, line 32: Nice test!

storage/transfermanager/download_buffer.go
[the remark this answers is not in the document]
tritone: I think even if partsize is large it still might be possible for many small writes to happen. But yeah profiling is the way to diagnose whether this is an issue.

I think we can add the growth rate later as an exported field, so we can merge as-is and just add that later if need be.

storage/transfermanager/downloader.go
```diff
@@ -156,46 +175,112 @@ func (d *Downloader) downloadWorker() {
...
+	in.shardOutputs = outs
+
+	// Create a shared context that we can cancel if a shard fails.
+	in.ctx, in.cancelCtx = context.WithCancelCause(in.ctx)
+
+	// Add generation in case the object changes between calls.
+	in.Generation = &gen
+
+	// Queue remaining shards.
+	// We could optimize for zero-initialized DownloadBuffers if we queue the last
```
tritone, line 222: Maybe turn this into a TODO for profiling work?

storage/transfermanager/downloader.go
tritone: Okay, I think as long as we have a test for a hanging part then we can remove the comment.

storage/transfermanager/downloader.go
tritone: I think this is probably fine, not sure what I was thinking of earlier...

storage/transfermanager/option.go
tritone: I think we should also probably set a minimum for partSize, and a way to turn off sharding (set it to zero, or negative?) But that can come in a separate PR.

tritone (commented):
Few more minor comments, but otherwise looks good

storage/transfermanager/downloader.go
BrennaEpp: I'm not sure if there is an easy way to simulate a hanging part, besides that would just be testing that a read request gets cancelled when its context is, which feels out of scope for TM. 

Other than that, if we add some logic here then we could test - some options:
1. extend the perOpTimeout and stop listening for pieces when done (start a timer with per_op_timeout * num_shards) 
2. have a default timer for collecting all pieces (probably not this)
3. check for ctx cancel directly in here and stop when cancelled, but I don't like that because then it would miss receiving any other errors in shards that were just completing...
4. instead of 3 with immediate cancellation, maybe a timer for a couple seconds could start once ctx cancel is received, and we only stop listening once that timer is up

storage/transfermanager/downloader.go
BrennaEpp: Moved it into perf ticket

PR #10382, opened 2024-06-14
feat(spanner): add LockHint feature

mayurkale22:


mayurkale22:
@harshachinta Please review

spanner/client_test.go
```diff
@@ -2143,6 +2146,55 @@ func TestClient_ReadWriteTransaction_Query_QueryOptions(t *testing.T) {
...
+			want:   &ReadOptions{LockHint: sppb.ReadRequest_LOCK_HINT_EXCLUSIVE},
+		},
+		{
+			name:   "Read level has precedence than client level",
+			client: &ReadOptions{LockHint: sppb.ReadRequest_LOCK_HINT_SHARED},
+			read:   &ReadOptions{LockHint: sppb.ReadRequest_LOCK_HINT_EXCLUSIVE},
+			want:   &ReadOptions{LockHint: sppb.ReadRequest_LOCK_HINT_EXCLUSIVE},
+		},
+		{
+			name:   "Client level has precendence when ORDER_BY_UNSPECIFIED at read level",
```
harshachinta, line 2169: update this comment for lock hint

spanner/transaction.go
```diff
@@ -297,6 +309,7 @@ func (t *txReadOnly) ReadWithOptions(ctx context.Context, table string, keys Key
 					DataBoostEnabled:    dataBoostEnabled,
 					DirectedReadOptions: directedReadOptions,
 					OrderBy:             orderBy,
+					LockHint:            lockHint,
```
harshachinta, line 312: I see that lock hints can only be used with RW transactions (https://github.com/googleapis/google-cloud-go/blob/0c17df27b06c4d67bc9a52ffad88fc58589d0096/spanner/apiv1/spannerpb/spanner.pb.go#L2009).
But `ReadWithOptions` method can be used for both RO and RW transactions.
So what will be the behaviour if this is used for RO transaction? Will it be no-op or will it throw any error?

One alternative is to set this field only when this method is called with in a read write transaction.

mayurkale22: Backend will throw an error if this is set in RO txn. 

>> One alternative is to set this field only when this method is called with in a read write transaction.

Can you pls send me code pointets for this?

harshachinta: `ReadOptions` can be set at client level and request level. 
1. When customer sets `lock hint` in `ReadOptions` at `request level` with in a `RO` txn then they get an error and that is right behaviour.
2. But when they set it at client level, customer expects that this `lockhint` gets set only for allowed txn (i.e `RW`) and gets ignored for `RO`. We did the same in case of directed reads as well. So, one thing we can do is in `client.go` file when the `lockhint` is set (this is client level setting) then that can be overriden to default with in the readonly transaction methods [here](https://github.com/googleapis/google-cloud-go/blob/18fa7e4961d055939078833d0442a415fac96ae6/spanner/client.go#L593).
WDYT?

spanner/client_test.go
mayurkale22: done

spanner/transaction.go
mayurkale22: sg, done.

spanner/client.go
```diff
@@ -609,6 +609,9 @@ func (c *Client) Single() *ReadOnlyTransaction {
 	}
 	t.txReadOnly.qo.DirectedReadOptions = c.dro
 	t.txReadOnly.ro.DirectedReadOptions = c.dro
+	if c.ro.LockHint == sppb.ReadRequest_LOCK_HINT_EXCLUSIVE {
```
harshachinta, line 612: Why do we need an if condition for unsetting? With this condition, there is one more lockhint called `LOCKHINT_SHARED`  and this will get applied to RO txns. Is that not a problem?

Can't we just do the below?
```
t.txReadOnly.ro.LockHint = sppb.ReadRequest_LOCK_HINT_UNSPECIFIED
```

spanner/client.go
```diff
@@ -635,6 +638,9 @@ func (c *Client) ReadOnlyTransaction() *ReadOnlyTransaction {
 	t.txReadOnly.disableRouteToLeader = true
 	t.txReadOnly.qo.DirectedReadOptions = c.dro
 	t.txReadOnly.ro.DirectedReadOptions = c.dro
+	if c.ro.LockHint == sppb.ReadRequest_LOCK_HINT_EXCLUSIVE {
+		t.txReadOnly.ro.LockHint = sppb.ReadRequest_LOCK_HINT_UNSPECIFIED
```
harshachinta, line 642: This should be applied to `BatchReadOnlyTransaction` and `BatchReadOnlyTransactionFromID`

mayurkale22: Done

spanner/client.go
mayurkale22: Done

spanner/transaction.go
```diff
@@ -185,6 +185,9 @@ type ReadOptions struct {
 
 	// An option to control the order in which rows are returned from a read.
 	OrderBy sppb.ReadRequest_OrderBy
+
+	// A lock hint mechanism to use for this request.
```
harshachinta, line 189: Lets mention that this is applicable only for RW transactions?

harshachinta (approved):
LGTM, with one minor comment.

spanner/transaction.go
mayurkale22: Done

mayurkale22:
@harshachinta Is this good to merge?

PR #10388, opened 2024-06-17
feat(bigtable): Retry on codes.ResourceExhausted

zapo:
We've started receiving the following error when sending mutations to our bigtable instance:
```
ApplyBulk: rpc error: code = ResourceExhausted desc = You have reached the limit of total mutations in your queue. Throttle your usage and wait for operations to finish.
```

We couldn't find references/documentation for this but saw that this is taken into account in the retry/backoff conditions in the java client: https://github.com/googleapis/java-bigtable/blob/7247c326a233adbf0d1aa3025c0c25e68bc74014/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/RateLimitingServerStreamingCallable.java#L247-L256

This PR updates the bigtable client to also retry on ResourceExhausted in addition to Unavailable, DeadlineExceeded and Aborted.


igorbernstein2:
java-bigtable doesn't retry resource exhausted errors. The code you linked to will throttle future qps, but will not retry the current error. In general retrying a resources exhausted error is quite dangerous and can lead to a downward spiral.

In general the error you are encountering implies that your application is not applying any flow control. Ie the client has too many RPCs outstanding. The limits that trigger ResourceExhausted are quite high. I would recommend to add some form of concurrency control to your application that can properly apply back pressure

PR #10397, opened 2024-06-20
fix(auth): Update http and grpc transports to support token exchange over mTLS

andyrzhao:
This fixes a regression in the new Golang Auth lib. The default http client used for OAuth2 token exchange (for both http and grpc transport) should be configured for mTLS (including switching endpoints) when certificate source is available (whether from default cert source or explicitly configured).


quartzmo:
Are you planning to add unit test coverage for the changes in this PR?

andyrzhao:
> Are you planning to add unit test coverage for the changes in this PR?

I will add some basic unit test coverage for the exported functions. I believe we need some functional tests at some point for thoroughly testing the mTLS behavior - it's so complicated now with S2A in the mix and also workload federation coming down the pipes - let me know if you have any thoughts on how to structure those. Thanks!

auth/credentials/detect.go
```diff
@@ -250,3 +253,31 @@ func clientCredConfigFromJSON(b []byte, opts *DetectOptions) *auth.Options3LO {
...
 	}
 }
+
+// GetGoogleMtlsTokenURL returns the mTLS token exchange URL
+func GetGoogleMtlsTokenURL() string {
+	return googleMtlsTokenURL
+}
+
+// CustomHTTPClient constructs an HTTPClient using the provided tlsConfig, to support mTLS.
+func CustomHTTPClient(tlsConfig *tls.Config) *http.Client {
```
codyoss, line 263: Why is this needed if a user can provide a http.Client already.

Also, this feels like if we were going to add it it should be done in the httptransport package.

auth/credentials/detect.go
```diff
@@ -250,3 +253,31 @@ func clientCredConfigFromJSON(b []byte, opts *DetectOptions) *auth.Options3LO {
 		AuthStyle: auth.StyleInParams,
 	}
 }
+
+// GetGoogleMtlsTokenURL returns the mTLS token exchange URL
+func GetGoogleMtlsTokenURL() string {
```
codyoss, line 258: This feels like it should be a const and not a func if we want this public.

auth/credentials/detect.go
```diff
@@ -250,3 +253,31 @@ func clientCredConfigFromJSON(b []byte, opts *DetectOptions) *auth.Options3LO {
...
+}
+
+// CustomHTTPClient constructs an HTTPClient using the provided tlsConfig, to support mTLS.
+func CustomHTTPClient(tlsConfig *tls.Config) *http.Client {
+	trans := baseTransport()
+	trans.TLSClientConfig = tlsConfig
+	return &http.Client{Transport: trans}
+}
+
+func baseTransport() *http.Transport {
```
codyoss, line 269: I feel like we have this code somewhere else as well. Could it be referenced instead.

auth/grpctransport/grpctransport.go
```diff
@@ -16,12 +16,13 @@ package grpctransport
...
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"net/http"
 
 	"cloud.google.com/go/auth"
-	"cloud.google.com/go/auth/credentials"
+	detect "cloud.google.com/go/auth/credentials"
```
codyoss, line 25: there seems to be no need to rename the import, I think we should prefer the normal name.

auth/credentials/detect.go
codyoss: looks like it is use in two package. I think this should be in internal.

auth/credentials/detect.go
andyrzhao: Done

auth/grpctransport/grpctransport.go
andyrzhao: Reverted to original, but note that "httptransport.go" does rename this to detect so I was simply trying to make the 2 consistent - you guys can figure it out later : )

auth/credentials/detect.go
andyrzhao: Moved to internal transport.go

auth/credentials/detect.go
andyrzhao: This is a port of the logic from the old library - it provides a fully configured based HTTP client to use with OAuth2 token exchange over mTLS - we don't have an existing implementation that fits the bill, but could consider relocating the impl in the future.

codyoss (approved):

PR #10405, opened 2024-06-21
feat(spanner): add support of checking row not found errors from ReadRow and ReadRowUsingIndex

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/10031

Customers can check row not found cases with below condition
```
if errors.Is(err, spanner.ErrRecordNotFound) {
```

- [x] Add test case for checking row not found with errors.Is
- [x] Ensure backward compatibility

### DIFF
--- a/spanner/errors.go
+++ b/spanner/errors.go
@@ -28,8 +28,8 @@ import (
 )
 
 var (
-	// ErrRowNotFound record not found error
-	ErrRowNotFound = errors.New("record not found")
+	// ErrRowNotFound row not found error
+	ErrRowNotFound = errors.New("row not found")
 )
 
 // Error is the structured error returned by Cloud Spanner client.
--- a/spanner/errors.go
+++ b/spanner/errors.go
@@ -18,12 +18,18 @@ package spanner
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/googleapis/gax-go/v2/apierror"
 	"google.golang.org/genproto/googleapis/rpc/errdetails"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+)
+
+var (
+	// ErrRecordNotFound record not found error
+	ErrRecordNotFound = errors.New("record not found")
 )
 
 // Error is the structured error returned by Cloud Spanner client.
--- a/spanner/errors.go
+++ b/spanner/errors.go
@@ -28,8 +28,8 @@ import (
 )
 
 var (
-	// ErrRecordNotFound record not found error
-	ErrRecordNotFound = errors.New("record not found")
+	// ErrRowNotFound record not found error
+	ErrRowNotFound = errors.New("record not found")
 )
 
 // Error is the structured error returned by Cloud Spanner client.
--- a/spanner/transaction.go
+++ b/spanner/transaction.go
@@ -327,14 +327,14 @@ func (t *txReadOnly) ReadWithOptions(ctx context.Context, table string, keys Key
 // key.
 func errRowNotFound(table string, key Key) error {
 	err := spannerErrorf(codes.NotFound, "row not found(Table: %v, PrimaryKey: %v)", table, key)
-	err.(*Error).err = ErrRecordNotFound
+	err.(*Error).err = ErrRowNotFound
 	return err
 }
 
 // errRowNotFoundByIndex returns error for not being able to read the row by index.
 func errRowNotFoundByIndex(table string, key Key, index string) error {
 	err := spannerErrorf(codes.NotFound, "row not found(Table: %v, IndexKey: %v, Index: %v)", table, key, index)
-	err.(*Error).err = ErrRecordNotFound
+	err.(*Error).err = ErrRowNotFound
 	return err
 }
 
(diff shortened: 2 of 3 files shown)


spanner/errors.go
```diff
@@ -26,6 +27,11 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+var (
```
olavloite, line 30: nit: Could this could be `const`?

spanner/errors.go
```diff
@@ -26,6 +27,11 @@ import (
...
+	// ErrRecordNotFound record not found error
+	ErrRecordNotFound = errors.New("record not found")
```
olavloite, line 32: `ErrRowNotFound` corresponds better with the original error message `row not found...`

```suggestion
	ErrRowNotFound = errors.New("record not found")
```

spanner/transaction.go
```diff
@@ -346,7 +350,7 @@ func errInlineBeginTransactionFailed() error {
 
 // ReadRow reads a single row from the database.
 //
-// If no row is present with the given key, then ReadRow returns an error where
+// If no row is present with the given key, then ReadRow returns an error(spanner.ErrRecordNotFound) where
 // spanner.ErrCode(err) is codes.NotFound.
 func (t *txReadOnly) ReadRow(ctx context.Context, table string, key Key, columns []string) (*Row, error) {
```
olavloite, line 355: Maybe add an example here for how users should check for the specific error.

spanner/transaction.go
```diff
@@ -373,7 +377,7 @@ func (t *txReadOnly) ReadRowWithOptions(ctx context.Context, table string, key K
 // ReadRowUsingIndex reads a single row from the database using an index.
 //
 // If no row is present with the given index, then ReadRowUsingIndex returns an
-// error where spanner.ErrCode(err) is codes.NotFound.
+// error(spanner.ErrRecordNotFound) where spanner.ErrCode(err) is codes.NotFound.
```
olavloite, line 380: Also here: Maybe add an example in the comments for how users should check for this specific error.

spanner/errors.go
rahul2393: No, const is reserved for string types only, here it is of error type.

spanner/transaction.go
rahul2393: Done

spanner/transaction.go
rahul2393: Done

spanner/errors.go
olavloite: ack

spanner/transaction.go
```diff
@@ -346,8 +350,14 @@ func errInlineBeginTransactionFailed() error {
 
 // ReadRow reads a single row from the database.
 //
-// If no row is present with the given key, then ReadRow returns an error where
+// If no row is present with the given key, then ReadRow returns an error(spanner.ErrRowNotFound) where
 // spanner.ErrCode(err) is codes.NotFound.
+//
+// To check if the error is spanner.ErrRecordNotFound:
+//
+//	if errors.Is(err, ErrRecordNotFound) {
```
olavloite, line 358:
```suggestion
//	if errors.Is(err, ErrRowNotFound) {
```

spanner/transaction.go
```diff
@@ -356,6 +366,12 @@ func (t *txReadOnly) ReadRow(ctx context.Context, table string, key Key, columns
 //
 // If no row is present with the given key, then ReadRowWithOptions returns an error where
 // spanner.ErrCode(err) is codes.NotFound.
+//
+// To check if the error is spanner.ErrRecordNotFound:
+//
+//	if errors.Is(err, ErrRecordNotFound) {
```
olavloite, line 372:
```suggestion
//	if errors.Is(err, ErrRowNotFound) {
```

spanner/transaction.go
```diff
@@ -373,7 +389,13 @@ func (t *txReadOnly) ReadRowWithOptions(ctx context.Context, table string, key K
 // ReadRowUsingIndex reads a single row from the database using an index.
 //
 // If no row is present with the given index, then ReadRowUsingIndex returns an
-// error where spanner.ErrCode(err) is codes.NotFound.
+// error(spanner.ErrRowNotFound) where spanner.ErrCode(err) is codes.NotFound.
+//
+// To check if the error is spanner.ErrRecordNotFound:
```
olavloite, line 394:
```suggestion
// To check if the error is spanner.ErrRowNotFound:
```

spanner/transaction.go
```diff
@@ -373,7 +389,13 @@ func (t *txReadOnly) ReadRowWithOptions(ctx context.Context, table string, key K
...
+//
+//	if errors.Is(err, ErrRecordNotFound) {
```
olavloite:
```suggestion
//	if errors.Is(err, ErrRowNotFound) {
```

olavloite (approved):

silas:
Easy enough to fix (switch to `errors.Is(err, spanner.ErrRowNotFound)`) and maybe something people shouldn't be relying on, but just wanted to let you know that if you were using `apierror.APIError` to detect this error previously, it no longer works (obviously still works if you're using the deprecated `spanner.Error`).

For example, I was doing the following previously to detect the row not being found:

```
var apiErr *apierror.APIError
if errors.As(err, &apiErr) {
    s := apiErr.GRPCStatus()
    return s.Code() == codes.NotFound && strings.HasPrefix(s.Message(), "row not found")
}
```

If you wanted to make this fully backwards compatible you could include `apierror.APIError` in the unwrap chain (I personally don't need it).

PR #10408, opened 2024-06-21
docs(compute/metadata): update OnGCE description

bgdnvk:
We are using the `metadata.OnGCE()` function to check if it runs on our Cloud Runs, this can be a helpful comment for others using the package.


compute/metadata/metadata.go
```diff
@@ -111,6 +111,7 @@ var (
 )
 
 // OnGCE reports whether this process is running on Google Compute Engine.
```
codyoss, line 113: maybe instead of calling out specific products we can just rephrase to "... Google Compute Platforms.". This way the docs can stay up-to-date over time.

bgdnvk: sure, the idea is to just give more context to the function for others

going to push the change now

codyoss (approved):

Issue #10414, opened 2024-06-21
auth: CachedTokenProviderOptions.DisableAutoRefresh returns nil token and nil err


quartzmo:
The logic in [cachedTokenProvider.Token](https://github.com/googleapis/google-cloud-go/blob/auth/v0.5.1/auth/auth.go#L257) does not check if the token is non-nil before returning it when `cachedTokenProvider.autoRefresh` is `false`.

**Code**

```go

	tp := NewCachedTokenProvider(&countingTestProvider{count: 1}, &CachedTokenProviderOptions{
		DisableAutoRefresh: true,
	})
	freshToken, err := tp.Token(context.Background())
	if err != nil {
		fmt.Println(err)
	}
	if freshToken == nil {
		fmt.Println("freshToken is nil")
	}
```

**Expected behavior**

The `CachedTokenProvider` should fetch a new token the first time `Token` is called, even when `DisableAutoRefresh` is `true`.

**Actual behavior**

The `CachedTokenProvider` returns `nil` token and `nil` err the first time `Token` is called when `DisableAutoRefresh` is `true`.

PR #10415, opened 2024-06-21
fix(auth): fetch initial token when CachedTokenProviderOptions.DisableAutoRefresh is true

quartzmo:
fixes: #10414


codyoss (approved):
good catch!

PR #10419, opened 2024-06-24
chore(spanner): add support of client side native metrics collection and export

rahul2393:
This PR adds built-in metrics to the Go spanner client library. This will introduce below metrics for the consumers of this library . 

This feature is not currently available for customers to use. Once GA'ed consumers of this library will start getting these metrics in the [MetricsExplorer](https://console.cloud.google.com/monitoring/metrics-explorer?_ga=2.218375641.1367355517.1723789359-1822529888.1723789359)

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-5283f587-7fff-966a-2e86-a09678ad4ca0"><br /><br /><div dir="ltr" style="margin-left:-117.75pt;" align="left">
Attempt Latency | The latency of an RPC attempt. | Useful for monitoring the general health of the applicationWith attributes like “Token Expiry”, “Status” etc, this metric can help us troubleshoot a wide range of issues.
-- | -- | --
Attempt Count | The count of RPC attempts | Useful for monitoring trafficUseful for deriving “Error Rate” with status attributes
Operation Latency | The total end-to-end latency across all request attempts, including retries. | Useful for monitoring the general health of the applicationThe combination of this metric and Attempt Latency can help us troubleshoot connectivity issues.
Operation Count | The count of operations | Useful for monitoring trafficUseful for deriving “Error Rate” with status labelThe combination of this metric and Attempt Count can be used for deriving Retry Count
</div></b>


odeke-em (changes requested):
Please separate non-standard library imports from standard library imports.

rahul2393:
@igorbernstein2 @codyoss 
Can you please help providing review here.
Thanks

Note: PR message is chore intentionally since we want this feature disabled(will be done in next commit here) but other things can be reviewed.

spanner/metrics.go
```diff
@@ -0,0 +1,469 @@
...
+)
+
+// These are effectively const, but for testing purposes they are mutable
+var (
+	// duration between two metric exports
+	defaultSamplePeriod = 5 * time.Minute
+
+	clientName = fmt.Sprintf("go-spanner v%v", internal.Version)
+
+	bucketBounds = []float64{0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 8.0, 10.0, 13.0, 16.0, 20.0, 25.0, 30.0, 40.0,
```
HailongWen, line 81: See internal discussion - please sync with Surbhi on the default buckets we use.

rahul2393:
[Update]
Integration tests passed: http://sponge2/b783a6e5-00ba-4e66-8990-3168fa92df9b

spanner/metrics.go
rahul2393: Took reference from Java PR https://github.com/googleapis/java-spanner/pull/3125/files#diff-be4eb8b7d32f038815df792bedaf4b93ee596806ed2aeaccec8afbcf89b2f342R100-R103

cc: @surbhigarg92

spanner/client.go
```diff
@@ -549,6 +571,43 @@ func allClientOpts(numChannels int, compression string, userOpts ...option.Clien
...
+		}
+		return streamer(ctx, desc, cc, method, opts...)
+	}
+}
+
+// AddUnaryPeerInterceptor intercepts unary requests and add PeerKey.
+func AddUnaryPeerInterceptor() grpc.UnaryClientInterceptor {
+	unaryInterceptors := []grpc.UnaryClientInterceptor{}
+	unaryInterceptors = append(unaryInterceptors, unaryPeerSetter())
+	return go_grpc_middleware.ChainUnaryClient(unaryInterceptors...)
```
zchee, line 601: grpc-go already have `WithChainUnaryInterceptor` and `WithChainStreamInterceptor`. No need `go_grpc_middleware` dependency.

- https://pkg.go.dev/google.golang.org/grpc@v1.65.0 #WithChainUnaryInterceptor
- https://pkg.go.dev/google.golang.org/grpc@v1.65.0#WithChainStreamInterceptor

spanner/grpc_client.go
```diff
@@ -0,0 +1,458 @@
...
+	"cloud.google.com/go/spanner/internal"
+)
+
+const (
+	directPathIPV6Prefix = "[2001:4860:8040"
+	directPathIPV4Prefix = "34.126"
+)
+
+// spannerClient is an interface that defines the methods available from Cloud Spanner API.
+type spannerClient interface {
```
hengfengli, line 44: Why do we need this new interface?

spanner/metrics.go
```diff
@@ -0,0 +1,474 @@
...
+	isDirectPathEnabled bool
+
+	// To be called on client close
+	shutdown func()
+
+	// attributes that are specific to a client instance and
+	// do not change across different function calls on client
+	clientAttributes []attribute.KeyValue
+
+	operationLatencies metric.Float64Histogram
```
hengfengli, line 144: So `operation` is the entire round-trip time from a client interface call which can include multiple `attempts`. `Attempt` is for each RPC call. Am I understanding it correctly? If we get two retryable errors and have two retries, does it count 1 or 2 `attempt`?

spanner/metrics.go
```diff
@@ -0,0 +1,474 @@
...
+	tracerFactory := &builtinMetricsTracerFactory{
+		enabled: false,
+		clientAttributes: []attribute.KeyValue{
+			attribute.String(monitoredResLabelKeyProject, project),
+			attribute.String(monitoredResLabelKeyInstance, instance),
+			attribute.String(metricLabelKeyDatabase, database),
+			attribute.String(metricLabelKeyClientUID, clientUID),
+			attribute.String(metricLabelKeyClientName, clientName),
+			// Skipping instance config until we have a way to get it
+			attribute.String(monitoredResLabelKeyInstanceConfig, "unknown"),
```
hengfengli, line 183: What does this mean to customers? Do they see this in somewhere?

spanner/metrics.go
```diff
@@ -0,0 +1,474 @@
...
+			return tracerFactory, err
+		}
+		meterProvider = sdkmetric.NewMeterProvider(mpOptions...)
+
+		tracerFactory.enabled = true
+		tracerFactory.shutdown = func() { meterProvider.Shutdown(ctx) }
+	} else {
+		switch metricsProvider.(type) {
+		case noop.MeterProvider:
+			tracerFactory.enabled = false
```
hengfengli, line 205: why not have `tracerFactory.enabled = false` before the switch?

spanner/metrics.go
```diff
@@ -0,0 +1,474 @@
...
+	tf.operationCount, err = meter.Int64Counter(
+		nativeMetricsPrefix+metricNameOperationCount,
+		metric.WithDescription("The count of database operations."),
+		metric.WithUnit(metricUnitCount),
+	)
+
+	// Create attempt_count
+	tf.attemptCount, err = meter.Int64Counter(
+		nativeMetricsPrefix+metricNameAttemptCount,
+		metric.WithDescription("The number of additional RPCs sent after the initial attempt."),
```
hengfengli, line 268: It's a bit confusing about the concept of `attempt`. Here, we exclude the initial attempt so the `initial attempt` is not actually a `attempt` here. I feel `attempt` more sounds like `how many retries we have done`.

spanner/metrics.go
```diff
@@ -0,0 +1,474 @@
...
+		metric.WithUnit(metricUnitCount),
+	)
+	return err
+}
+
+// builtinMetricsTracer is created one per operation
```
hengfengli, line 274: Can we format this comment again with correct `.`?

spanner/metrics.go
```diff
@@ -0,0 +1,474 @@
...
+	// Record attempt_count
+	attemptCntAttrs, _ := mt.toOtelMetricAttrs(metricNameAttemptCount)
+	if mt.currOp.attemptCount > 1 {
+		// Only record when retry count is greater than 0 so the retry
+		// graph will be less confusing
+		mt.instrumentAttemptCount.Add(mt.ctx, mt.currOp.attemptCount-1, metric.WithAttributes(attemptCntAttrs...))
+	}
+}
+func convertToMs(d time.Duration) float64 {
+	return float64(d.Nanoseconds()) / 1000000
```
hengfengli, line 473: Should we use `return float64(d.Nanoseconds()) / float64(time.Millisecond)`?

spanner/sessionclient.go
```diff
@@ -90,17 +89,18 @@ type sessionClient struct {
...
+	database             string
+	id                   string
+	userAgent            string
+	sessionLabels        map[string]string
+	databaseRole         string
+	md                   metadata.MD
+	batchTimeout         time.Duration
+	logger               *log.Logger
+	callOptions          *vkit.CallOptions
+	otConfig             *openTelemetryConfig
```
hengfengli, line 102: should we use `otel`?

spanner/sessionclient.go
```diff
@@ -404,21 +403,11 @@ func (sc *sessionClient) nextClient() (*vkit.Client, error) {
...
-		agentWithVersion := strings.SplitN(sc.userAgent, "/", 2)
-		if len(agentWithVersion) == 2 {
-			clientInfo = append(clientInfo, agentWithVersion[0], agentWithVersion[1])
-		}
-	}
-	client.SetGoogleClientInfo(clientInfo...)
-	if sc.callOptions != nil {
-		client.CallOptions = mergeCallOptions(client.CallOptions, sc.callOptions)
-	}
+
```
hengfengli, line 410: Where do we move this to? How can we make sure `gccl` is not missing? Do we have any test coverage for this?

spanner/metrics_monitoring_exporter.go
```diff
@@ -0,0 +1,344 @@
...
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const (
+	spannerResourceType = "spanner_instance_client"
+
+	// The number of timeserieses to send to GCM in a single request. This
+	// is a hard limit in the GCM API, so we never want to exceed 200.
+	sendBatchSize = 200
```
hengfengli, line 47: Why is 200?

spanner/metrics_monitoring_exporter.go
```diff
@@ -0,0 +1,344 @@
...
+	"google.golang.org/genproto/googleapis/api/distribution"
+	googlemetricpb "google.golang.org/genproto/googleapis/api/metric"
+	monitoredrespb "google.golang.org/genproto/googleapis/api/monitoredres"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const (
+	spannerResourceType = "spanner_instance_client"
+
+	// The number of timeserieses to send to GCM in a single request. This
```
hengfengli, line 45: What does `GCM` stand for?

spanner/read_test.go
```diff
@@ -1213,6 +1215,11 @@ func (sr *sReceiver) Recv() (*sppb.PartialResultSet, error) {
 	return sr.rpcReceiver.Recv()
 }
 
+func (sr *sReceiver) Context() context.Context {
+	return context.TODO()
```
hengfengli, line 1219: What does this TODO mean?

spanner/read.go
```diff
@@ -589,50 +611,57 @@ func (d *resumableStreamDecoder) next() bool {
...
 		}
-	} else {
-		delay, shouldRetry := retryer.Retry(d.err)
-		if !shouldRetry || d.state != queueingRetryable {
-			d.changeState(aborted)
-			return
+		if d.err == io.EOF {
+			d.err = nil
+			d.changeState(finished)
+			mt.currOp.attemptCount--
```
hengfengli, line 631: Why do we minus 1 here?

spanner/read.go
```diff
@@ -589,50 +611,57 @@ func (d *resumableStreamDecoder) next() bool {
...
-		// A 'Session not found' error occurred before we received a resume
-		// token and a replaceSessionFunc function is defined. Try to restart
-		// the stream on a new session.
-		if err := d.replaceSessionFunc(d.ctx); err != nil {
-			d.err = err
-			d.changeState(aborted)
-			return
+	// ignore error from gaxInvokeWithRecorder, as it is already handled in
+	// the retryer and resumableStreamDecoder.
+	_ = gaxInvokeWithRecorder(d.ctx, mt, d.method, func(ctx context.Context, settings gax.CallSettings) (context.Context, error) {
```
hengfengli, line 618: So we just do not need to do anything for the returned error?

hengfengli (commented):
Do we have any tests for metrics?

spanner/read.go
rahul2393: I have updated the code to record attempt latencies in stream interceptor, and yes we don;t want to surface any errors if error while recording native metrics.

spanner/read.go
rahul2393: Removed this
Thanks.

spanner/read_test.go
rahul2393: This receiver is used in few of th unit tests, returning context.TODO here meant there are those tests nothing.

spanner/metrics_monitoring_exporter.go
rahul2393: Updated comment.
`GCM => Google Cloud Monitoring`

spanner/metrics_monitoring_exporter.go
rahul2393: Updated comment to refer the limitation page https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.timeSeries/create

spanner/metrics.go
rahul2393: Thanks, moved.

spanner/metrics.go
rahul2393: Updated the comment, attempt will include initial attempt too.

spanner/metrics.go
rahul2393: Done

spanner/metrics.go
rahul2393: Done

spanner/sessionclient.go
rahul2393: Sorry are you suggesting renaming the variable here, or want to ask if we should otel, 
If it's latter then yes we now want to use otel going forwards since opencensus will deprecate soon.

spanner/metrics.go
rahul2393: Since getting the value for instance config involves admin RPC call and there is quota limit per project, we decided to keep the label but have unknown as value for now.
Java Reference: https://github.com/googleapis/java-spanner/pull/3125/files#diff-c896ca5e475a075f267b0294d92328dcd35db60bdd408600bf05354119650be1R85

spanner/metrics.go
rahul2393: It will count for 2 attempt.
So retry count can be derived from operationCount-attemptCount so decision was to not export retry.
https://docs.google.com/document/d/199avBkkj8ewCHsUYR2dtnnOVdDMTyWg0O-NtHuzWsiQ/edit?resourcekey=0-UydHkoiIv2akT3LNl5NM_w&tab=t.0#bookmark=id.e0pp687p4vm

spanner/grpc_client.go
rahul2393: We needed this interface to inject the metrics tracer object in context, later the object will be used to export operation and attempt metrics, it adds a very thin layer on top of autogenerated spanner_client thus making sure that all the current and future autogeneration changes applied to this interface implementation too.

spanner/client.go
```diff
@@ -988,54 +1051,68 @@ func (bwo BatchWriteOptions) merge(opts BatchWriteOptions) BatchWriteOptions {
...
 
 // Next returns the next result. Its second return value is iterator.Done if
 // there are no more results. Once Next returns Done, all subsequent calls
 // will return Done.
 func (r *BatchWriteResponseIterator) Next() (*sppb.BatchWriteResponse, error) {
-	for {
-		// Stream finished or in error state.
-		if r.err != nil {
-			return nil, r.err
+	mt := r.meterTracerFactory.createBuiltinMetricsTracer(r.ctx, true)
```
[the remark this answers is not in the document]
rahul2393: Removed, It was added earlier during debugging to test streamInterceptor.

spanner/client.go
```diff
@@ -549,6 +570,43 @@ func allClientOpts(numChannels int, compression string, userOpts ...option.Clien
 	return append(allDefaultOpts, userOpts...)
 }
 
+func unaryPeerSetter() grpc.UnaryClientInterceptor {
```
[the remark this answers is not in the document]
rahul2393: Removed this interceptor, instead added a new one with comment

rahul2393:
@hengfengli Added unit tests `TestNewBuiltinMetricsTracerFactory` which internally does both unary(CreateSession) and stream call(StreamingRead) to validate metrics captured works as expected.

spanner/sessionclient.go
rahul2393: In the existing tests we are asserting gccl and other related headers using grpcHeaderChecker, so it will ensure that new client also exports them.

spanner/client.go
```diff
@@ -24,15 +24,19 @@ import (
 	"os"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
+	"google.golang.org/grpc/status"
+
```
zchee, line 31: It would be better to merge the above import packages section.

spanner/client.go
```diff
@@ -42,6 +46,7 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/encoding/gzip"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
```
zchee, line 49: like this:

```suggestion
	"google.golang.org/grpc/peer"
	"google.golang.org/grpc/status"
```

zchee (commented):
@rahul2393  I'm not a Googler, so sorry for my external review.
I made a few comments about the code style, but it nits.

spanner/grpc_client.go
hengfengli: Thanks for the explanation.

spanner/sessionclient.go
hengfengli: I am suggesting to rename `otConfig` to `otelConfig`.

hengfengli (approved):
LGTM.

PR #10430, opened 2024-06-25
feat(storage/transfermanager): add DownloadDirectory 

BrennaEpp:


storage/transfermanager/downloader.go
```diff
@@ -69,6 +73,82 @@ func (d *Downloader) DownloadObject(ctx context.Context, input *DownloadObjectIn
...
+	if err := query.SetAttrSelection([]string{"Name"}); err != nil {
+		return fmt.Errorf("transfermanager: DownloadDirectory query.SetAttrSelection: %w", err)
+	}
+
+	it := d.client.Bucket(input.Bucket).Objects(ctx, query)
+
+	// TODO: Clean up any created directory structure on failure.
+
+	attrs, err := it.Next()
+	for err != iterator.Done {
```
tritone, line 113: I feel like this is a slightly confusing way to write this loop; is there a reason you do it this way vs. how we write it in the list objects sample?

storage/transfermanager/downloader.go
```diff
@@ -402,6 +499,40 @@ func (in *DownloadObjectInput) downloadShard(client *storage.Client, timeout tim
 	return
 }
 
+// DownloadDirectoryInput is the input for a directory to download.
+type DownloadDirectoryInput struct {
+	// Required fields.
+	Bucket         string
+	LocalDirectory string
```
tritone, line 506: This needs a doc

storage/transfermanager/downloader.go
```diff
@@ -69,6 +73,82 @@ func (d *Downloader) DownloadObject(ctx context.Context, input *DownloadObjectIn
...
+		// Make sure all directories in the object path exist.
+		objDirectory := filepath.Join(input.LocalDirectory, filepath.Dir(object))
+		err = os.MkdirAll(objDirectory, fs.ModeDir|fs.ModePerm)
+		if err != nil {
+			return fmt.Errorf("transfermanager: DownloadDirectory failed to make directory(%q): %w", objDirectory, err)
+		}
+
+		// Create file to download to.
+		filePath := filepath.Join(input.LocalDirectory, object)
+		f, fErr := os.Create(filePath)
```
tritone, line 131: Does this work to create subdirs if the object name has slashes?

tritone (commented):
A few minor things; generally looks good

storage/transfermanager/downloader.go
BrennaEpp: Honestly this seems more intuitive to me, but I can change it so it's consistent with the samples since it's usually how we do it.

storage/transfermanager/downloader.go
BrennaEpp: The subdirs should be created in L124 `os.MkdirAll`

storage/transfermanager/downloader.go
tritone: Yeah I would prefer that, it just seems odd to have the first call of Next happen before the loop.

storage/transfermanager/downloader.go
tritone: Gotcha, sounds good.

storage/transfermanager/downloader.go
```diff
@@ -31,6 +31,10 @@ import (
 	"google.golang.org/api/iterator"
 )
 
+// ErrDownloaderClosed indicates that the Downloader has already been closed and
+// must not be used for new operations.
+var ErrDownloaderClosed = errors.New("transfermanager: Downloader used after WaitAndClose was called")
```
tritone, line 36: I don't think this should be an exported sentinel error, it's not really something that people should need to check for programmatically.

storage/transfermanager/downloader.go
tritone: Looks good but I think we should add here the note about not modifying the directory while the download is in progress.

storage/transfermanager/downloader.go
```diff
@@ -402,6 +525,43 @@ func (in *DownloadObjectInput) downloadShard(client *storage.Client, timeout tim
...
+	// MatchGlob is a glob pattern used to filter results (for example, foo*bar). See
+	// https://cloud.google.com/storage/docs/json_api/v1/objects/list#list-object-glob
+	// for syntax details. Optional.
+	MatchGlob string
+
+	// Callback will run after each object in the directory as selected by the
+	// provided filters is finished downloading. It must be set if and only if
+	// the [WithCallbacks] option is set.
+	// WaitAndClose will wait for all callbacks to finish.
+	Callback func(*DownloadOutput)
```
tritone, line 562: One more thought about callbacks. Thinking about it some more, it seems like if someone is running in async mode (not calling WantAndClose) they would not have a way to know when the whole directory download is complete. Any thoughts about how to address this? Perhaps it makes more sense to have the callback take []DownloadOutput and only trigger once the whole directory is finished?

BrennaEpp: Hmmm... good point. I think it makes sense for the callback to be called on each file download. But I agree there should be a way of knowing when the directory is completely downloaded. 

What do you think of changing the current callbacks into a `ProgressFunc` (since keeping the code to trigger it would be pretty simple and it's a useful feature) and then have `Callback` be as you describe? I'll open another PR for this since it will require a bit more involved code added.

storage/transfermanager/downloader.go
```diff
@@ -546,11 +561,29 @@ type DownloadDirectoryInput struct {
...
-	// the [WithCallbacks] option is set.
-	// WaitAndClose will wait for all callbacks to finish.
-	Callback func(*DownloadOutput)
+	// Callback will run after all the objects in the directory as selected by
+	// the provided filters are finished downloading.
+	// It must be set if and only if the [WithCallbacks] option is set.
+	Callback func([]DownloadOutput)
+
+	// OnObjectDownload will run after every finished object download.
+	// It can only be set if the Downloader has the [WithCallbacks] option set.
```
tritone, line 570: I think we could probably loosen this requirement? A user could want to run in sync mode and still use this to track/log progress.

tritone (approved):
One more minor suggestion, otherwise looks good.

storage/transfermanager/downloader.go
BrennaEpp: Done.

tritone (approved):

PR #10436, opened 2024-06-25
feat(pubsub): add client ID to initial streaming pull request

hongalex:
Similar to https://github.com/googleapis/java-pubsub/pull/77, this will improve ordering key affinity and allow Pub/Sub servers to determine if streams are opened from the same subscriber client instance.

### DIFF
--- a/pubsub/subscription.go
+++ b/pubsub/subscription.go
@@ -27,6 +27,7 @@ import (
 	"cloud.google.com/go/internal/optional"
 	pb "cloud.google.com/go/pubsub/apiv1/pubsubpb"
 	"cloud.google.com/go/pubsub/internal/scheduler"
+	"github.com/google/uuid"
 	gax "github.com/googleapis/gax-go/v2"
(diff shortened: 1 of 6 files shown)


pubsub/subscription.go
```diff
@@ -50,6 +51,11 @@ type Subscription struct {
 
 	mu            sync.Mutex
 	receiveActive bool
+
+	// clientID to be used across all streaming pull connections that are created.
+	// This indicates to the server that any guarantees made for a stream that
+	// disconnected will be made for the stream that is created to replace it.
+	clientID string
```
shollyman, line 58: are there use cases for making this visible to users?

shollyman (approved):

pubsub/subscription.go
hongalex: In Java the field is marked private, so until that changes, I'm incline to leave as-is.

kamalaboulhosn (approved):

PR #10449, opened 2024-06-27
fix(spanner): fix negative values for max_in_use_sessions metrics

rahul2393:
Internal bug: b/343756862


spanner/session_test.go
```diff
@@ -1022,6 +1022,57 @@ func TestMinOpenedSessions(t *testing.T) {
...
+	}
+	sp.mu.Lock()
+	defer sp.mu.Unlock()
+	if int64(sp.numInUse) < 0 {
+		t.Fatal("numInUse must be >= 0")
+	}
+	// There should be still one session left in either the idle list or in one
+	// of the other opened states due to the min open sessions constraint.
+	if (sp.idleList.Len() +
+		int(sp.createReqs)) != 1 {
```
olavloite, line 1069: nit: Can this be on one line (I found it hard to read in the current form)

spanner/session_test.go
```diff
@@ -1022,6 +1022,57 @@ func TestMinOpenedSessions(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("failed to get session(%v): %v", i, err)
+		}
+		shs = append(shs, sh)
+	}
+	for _, sh := range shs {
+		sh.recycle()
+	}
+
+	for true {
```
olavloite, line 1052: nit: Can we add an escape hatch to the loop (e.g. stop after X time)? Now a future bug could cause this to loop for ever, which is harder to debug than a test failure after for example 2 seconds.

spanner/session.go
```diff
@@ -1152,8 +1152,10 @@ func (p *sessionPool) remove(s *session, isExpire bool) bool {
 	if s.invalidate() {
 		// Decrease the number of opened sessions.
 		p.numOpened--
-		// Decrease the number of sessions in use.
-		p.decNumInUseLocked(ctx)
+		// Decrease the number of sessions in use, only when not from idle list.
+		if !isExpire {
```
olavloite, line 1156: I think that it would be better to add an additional argument to the `remove(..)` method that explicitly says whether the session was in use or not. So something like this:

```go
func (p *sessionPool) remove(s *session, isExpire bool, wasInUse bool) {
  ...
  if wasInUse {
    p.decNumInUseLocked(ctx)
  }
}
```

In theory, it could be that this method is called in the future to remove sessions that have not expired, but that also were not in use at the time that they were being removed, and then we could re-introduce a similar bug as the one here. That is less likely with an explicit argument that clearly says what it is for.

spanner/session.go
```diff
@@ -907,7 +907,7 @@ func (p *sessionPool) close(ctx context.Context) {
 
 func deleteSession(ctx context.Context, s *session, wg *sync.WaitGroup) {
 	defer wg.Done()
-	s.destroyWithContext(ctx, false)
+	s.destroyWithContext(ctx, false, true)
```
olavloite, line 910: Is `deleteSession` only called for sessions that are in use at that moment?

Note that `inUse` means that the session was checked out of the pool at the moment that this method is being called. So in this case it would mean that we are calling `deleteSession(..)` for a session that was checked out.

spanner/session.go
```diff
@@ -197,7 +197,7 @@ func (sh *sessionHandle) destroy() {
 		p.trackedSessionHandles.Remove(tracked)
 		p.mu.Unlock()
 	}
-	s.destroy(false)
+	s.destroy(false, true)
```
olavloite, line 200: Is `destroy()` only called for sessions that are in use? I would expect that it could also be called for a session that is in the list of idle sessions, and in that case it was not in use.

rahul2393: sessionHandle is the wrapper which is created only for transactions(to be used) so we can assume any call for destroy using sessionHandle was in use

rahul2393: @olavloite added comment for the same

olavloite: Ah, that makes it clearer, thanks!

spanner/session.go
olavloite: But won't that then mean that the metric will drop to a negative value for a (very) short time when the application is shutting down? Assume that the situation is that:
1. The pool has 100 sessions.
2. 10 of them are in use.
3. The application shuts down and closes the client.
4. The client closes all sessions and calls this method for all 100 sesssions.
5. The inUseSessions metric drop from 10 to -90 for a very short time.

rahul2393: wasInUse was set to false in latest commit, hence negative values should not appear.

olavloite (approved):

PR #10241, opened 2024-05-21
feat(spanner): Add x-goog-spanner-end-to-end-tracing header for requests to Spanner

nareshz:
To enable tracing on Spanner layer, client can opt-in by following ways:

1. Set the `EnableEndToEndTracing` flag in ClientConfig to `true`. 
2. Set the `SPANNER_ENABLE_END_TO_END_TRACING` environment variable to `true`.

If enabled a new http header `x-goog-spanner-end-to-end-tracing` will be passed in Spanner requests.

**Note**: Spanner tracing also requires trace context propagation. This needs to be setup while setting OpenTelemetry.

Design doc: go/cs-tracing-mvp

PR #10450, opened 2024-06-27
feat(Spanner): Add changes in Spanner executor for testing end to end tracing

nareshz:
This PR adds changes in spanner executor code required for testing end to end tracing feature.


spanner/test/cloudexecutor/executor/internal/inputstream/handler.go
```diff
@@ -110,11 +140,20 @@ func (h *CloudStreamHandler) startHandlingRequest(ctx context.Context, req *exec
 	}
 
 	// Get a new action handler based on the input action.
-	actionHandler, err := h.newActionHandler(inputAction, outcomeSender)
+	actionType, actionHandler, err := h.newActionHandler(inputAction, outcomeSender)
```
harshachinta, line 143: `actionType` can be derived directly from `inputAction`. Passing it from newActionHandler is redundant and is hard to maintain.

My suggestion would be,
Create a new function and use it where required.
```
func getActionType(inputAction *executorpb.SpannerAction) string {
    // GET THE ACTION TYPE
    action := fmt.Sprintf("%T", inputAction.GetAction())
   // Here action will have output as `*executorpb.SpannerAction_Query`. Apply string processing to convert into required format
}

```

spanner/test/cloudexecutor/executor/internal/inputstream/handler.go
```diff
@@ -142,21 +181,68 @@ func (h *CloudStreamHandler) startHandlingRequest(ctx context.Context, req *exec
...
+		if !spannerLayerSpanPresent {
+			return fmt.Errorf("no internal span found for trace_id: %v", traceId)
+		}
+	}
+	return nil
+}
+
 // newActionHandler instantiates an actionHandler for executing the given action.
-func (h *CloudStreamHandler) newActionHandler(action *executorpb.SpannerAction, outcomeSender *outputstream.OutcomeSender) (cloudActionHandler, error) {
+func (h *CloudStreamHandler) newActionHandler(action *executorpb.SpannerAction, outcomeSender *outputstream.OutcomeSender) (string, cloudActionHandler, error) {
```
harshachinta, line 231: Lets not make this change and instead use the suggestion above.

spanner/test/cloudexecutor/executor/actions/batch.go
```diff
@@ -50,10 +52,16 @@ func (h *StartBatchTxnHandler) ExecuteAction(ctx context.Context) error {
 		return h.OutcomeSender.FinishWithError(spanner.ToSpannerError(status.Error(codes.InvalidArgument, "database path must be set for this action")))
 	}
 
-	client, err := spanner.NewClient(ctx, h.FlowContext.Database, h.Options...)
+	client, err := spanner.NewClientWithConfig(ctx, h.FlowContext.Database, spanner.ClientConfig{SessionPoolConfig: spanner.DefaultSessionPoolConfig, DisableRouteToLeader: false, EnableEndToEndTracing: true}, h.Options...)
+	if err != nil {
+		return h.OutcomeSender.FinishWithError(err)
+	}
+	// Create a trace client to read the traces from Cloud Trace.
+	traceClient, err := trace.NewClient(ctx, h.TraceClientOptions...)
```
harshachinta, line 60: Looks like the traceClient is independent of the type of action and can be initialized just once at the beginning of the `Execute()` method - [here](https://github.com/googleapis/google-cloud-go/blob/a187451a912835703078e5b6a339c514edebe5de/spanner/test/cloudexecutor/executor/internal/inputstream/handler.go#L62).
Remove it from `StartBatchTxnHandler` and `StartTxnHandler`.

spanner/test/cloudexecutor/executor/actions/batch.go
```diff
@@ -31,10 +32,11 @@ import (
...
 type StartBatchTxnHandler struct {
-	Action        *executorpb.StartBatchTransactionAction
-	FlowContext   *ExecutionFlowContext
-	OutcomeSender *outputstream.OutcomeSender
-	Options       []option.ClientOption
+	Action             *executorpb.StartBatchTransactionAction
+	FlowContext        *ExecutionFlowContext
+	OutcomeSender      *outputstream.OutcomeSender
+	Options            []option.ClientOption
+	TraceClientOptions []option.ClientOption
```
harshachinta, line 39: This is not needed and can be removed with https://github.com/googleapis/google-cloud-go/pull/10450/files#r1665520383

harshachinta (changes requested):
Reviewed couple of files and added comments. Will take a look at the remaining files after these changes are incorporated.

spanner/test/cloudexecutor/executor/internal/inputstream/handler.go
nareshz: Done.

spanner/test/cloudexecutor/executor/actions/batch.go
nareshz: Done.

spanner/test/cloudexecutor/executor/actions/batch.go
nareshz: Done.

spanner/test/cloudexecutor/executor/executor_proxy_server_impl.go
```diff
@@ -25,24 +25,38 @@ import (
...
+		ServerContext:          s.serverContext,
+		Options:                s.options,
+		TraceClientOptions:     s.traceClientOptions,
+		CloudTraceCheckAllowed: (s.cloudTraceCheckCount < MAX_CLOUD_TRACE_CHECK_LIMIT),
+	}
+	if err := handler.Execute(); err != nil {
+		return err
+	}
+	if handler.IsServerSideTraceCheckDone() {
+		s.cloudTraceCheckCount++
```
harshachinta, line 59: Here `cloudTraceCheckCount` can be read and written at the same time if the `ExecuteActionAsync` is executed parallely. To avoid race condition use mutex to access/write this thread safe.

spanner/client.go
```diff
@@ -331,6 +335,12 @@ type ClientConfig struct {
 	DirectedReadOptions *sppb.DirectedReadOptions
 
 	OpenTelemetryMeterProvider metric.MeterProvider
+
+	// EnableServerSideTracing indicates whether server side tracing is enabled or not.
+	// If it is enabled, trace spans will be created at Spanner layer.
+	//
+	// Default: false
+	EnableServerSideTracing bool
```
harshachinta, line 343: Do we need an ENV also here similar to what is being done in Java?
SPANNER_ENABLE_SERVER_SIDE_TRACING

spanner/client.go
```diff
@@ -331,6 +335,12 @@ type ClientConfig struct {
 	DirectedReadOptions *sppb.DirectedReadOptions
 
 	OpenTelemetryMeterProvider metric.MeterProvider
+
+	// EnableServerSideTracing indicates whether server side tracing is enabled or not.
```
harshachinta, line 339: Lets add same comment as that in java PR

spanner/test/cloudexecutor/executor/executor_proxy_server_impl.go
```diff
@@ -25,24 +25,38 @@ import (
...
+	traceClientOptions []option.ClientOption
+	// members below represent internal state
+	cloudTraceCheckCount int
 }
 
 // NewCloudProxyServer initializes and returns a new CloudProxyServer instance.
-func NewCloudProxyServer(ctx context.Context, opts []option.ClientOption) (*CloudProxyServer, error) {
-	return &CloudProxyServer{serverContext: ctx, options: opts}, nil
+func NewCloudProxyServer(ctx context.Context, opts []option.ClientOption, traceClientOpts []option.ClientOption) (*CloudProxyServer, error) {
+	return &CloudProxyServer{serverContext: ctx, options: opts, traceClientOptions: traceClientOpts, cloudTraceCheckCount: 0}, nil
```
harshachinta, line 42: cloudTraceCheckCount by default will be 0.

spanner/test/cloudexecutor/executor/internal/inputstream/handler.go
```diff
@@ -142,6 +209,48 @@ func (h *CloudStreamHandler) startHandlingRequest(ctx context.Context, req *exec
 	}
 }
 
+// verifyCloudTraceExportedTraces fetches the traces exported from client application using
+// Cloud Trace API to cross verify if end to end tracing is working or not.
+func (h *CloudStreamHandler) verifyCloudTraceExportedTraces(ctx context.Context, traceId string) error {
```
harshachinta, line 214: looks like this method is getting called only once after all the actions complete. If that is right, and TraceClient is getting used only at this place then there is no need to create it way before and store it in the execution flow context. You can create the trace client here and close it once the verification completes. 
WDYT?
Is the traceclient used somewhere else?

spanner/test/cloudexecutor/executor/executor_proxy_server_impl.go
nareshz: Done.

spanner/test/cloudexecutor/executor/actions/batch.go
```diff
@@ -50,7 +50,7 @@ func (h *StartBatchTxnHandler) ExecuteAction(ctx context.Context) error {
 		return h.OutcomeSender.FinishWithError(spanner.ToSpannerError(status.Error(codes.InvalidArgument, "database path must be set for this action")))
 	}
 
-	client, err := spanner.NewClient(ctx, h.FlowContext.Database, h.Options...)
+	client, err := spanner.NewClientWithConfig(ctx, h.FlowContext.Database, spanner.ClientConfig{SessionPoolConfig: spanner.DefaultSessionPoolConfig, DisableRouteToLeader: false, EnableServerSideTracing: true}, h.Options...)
```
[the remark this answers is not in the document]
nareshz: I think we will need to set it explicitly. Even `spanner.NewClient` method sets the default values explicitly as well in its implementation.

spanner/test/cloudexecutor/executor/executor_proxy_server_impl.go
nareshz: Removed. Thanks.

spanner/client.go
nareshz: No, we don't ENV here. Confirmed with Simil.

spanner/client.go
nareshz: Makes sense. Updated the opt-in header PR(https://github.com/googleapis/google-cloud-go/pull/10241) as well.

spanner/test/cloudexecutor/executor/internal/inputstream/handler.go
nareshz: No, trace client is only used at the end and doesn't need to be created at the start. thanks for pointing out.

spanner/client.go
harshachinta: Why do we have same changes in 2 different PRs. If https://github.com/googleapis/google-cloud-go/pull/10241/files is valid, then remove these changes from here and lets have the executor changes alone.

nareshz: This PR is created on top of the changes of https://github.com/googleapis/google-cloud-go/pull/10241. Executor changes also includes the usage of `EnableServerSideTracing` flag which is in https://github.com/googleapis/google-cloud-go/pull/10241. That's why copied the changes from that PR.

Considering this PR will be submitted after https://github.com/googleapis/google-cloud-go/pull/10241 is submitted, When we submit this PR, can't we take a pull from main branch?

harshachinta: It is better to not have changes from other PR here since it will double the review effort.
Since this PR will any how get submitted after #10241, when you rebase this PR with main you will have those changes applied automatically. However, till then you might see failing builds in this PR which is not a problem.

Lets have executor changes alone in this PR.

spanner/test/cloudexecutor/executor/executor_proxy_server_impl.go
```diff
@@ -19,30 +19,48 @@ package executor
...
 func (s *CloudProxyServer) ExecuteActionAsync(inputStream executorpb.SpannerExecutorProxy_ExecuteActionAsyncServer) error {
 	handler := &inputstream.CloudStreamHandler{
-		Stream:        inputStream,
-		ServerContext: s.serverContext,
-		Options:       s.options,
+		Stream:                 inputStream,
+		ServerContext:          s.serverContext,
+		Options:                s.options,
+		TraceClientOptions:     s.traceClientOptions,
+		CloudTraceCheckAllowed: (s.cloudTraceCheckCount < MAX_CLOUD_TRACE_CHECK_LIMIT),
```
harshachinta, line 55: Here we are reading a shared resource. take a lock before reading.

spanner/test/cloudexecutor/executor/internal/inputstream/handler.go
```diff
@@ -84,16 +104,57 @@ func (h *CloudStreamHandler) Execute() error {
...
+		}()
+
+		// Verify the end to end trace exported to Cloud Trace.
+		traceId := ottrace.SpanContextFromContext(ctx).TraceID().String()
+		if err = verifyCloudTraceExportedTraces(ctx, traceClient, traceId); err != nil {
+			log.Printf("Verification failed for exported traces: %v", err)
+			return err
+		}
+	}
 	log.Println("Done executing actions")
```
harshachinta, line 144: Lets move this line after `trace.EndSpan(ctx, nil) `,
This way we can know that action completed successfully and any error further might be due to e2e tracing

harshachinta:
LGTM with small comments.
@nareshz Can you share the systest run with this change?

spanner/test/cloudexecutor/executor/executor_proxy_server_impl.go
nareshz: Done.

spanner/test/cloudexecutor/executor/internal/inputstream/handler.go
nareshz: Okay. makes sense. Done

nareshz:
> LGTM with small comments. @nareshz Can you share the systest run with this change?

https://fusion2.corp.google.com/invocations/679341f2-7708-4795-98bd-4ea9ebe46768/targets

spanner/client.go
nareshz: No, for JAVA we are keeping to make consistency with other options. For Go, discussed with Simil earlier and we thought that we can add this incrementally if needed.

spanner/client.go
nareshz: Removed the opt-in header changes from this PR.

harshachinta (approved):

spanner/test/cloudexecutor/worker_proxy.go
```diff
@@ -28,6 +28,10 @@ import (
 
 	"cloud.google.com/go/spanner/executor/apiv1/executorpb"
 	"cloud.google.com/go/spanner/test/cloudexecutor/executor"
+	texporter "github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/trace"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
```
harshachinta, line 34: I am worried about this dep since we are adding a direct dependency of OTEL SDK package to the main package. This will download SDK dep to customer.

Can you check if we can have a seperate dependency file for executor which now relies on main package dependency. I have done something similar for opentelemetry tests since it also had SDK dependency.

https://github.com/googleapis/google-cloud-go/tree/main/spanner/test/opentelemetry/test

PR #10453, opened 2024-06-27
refactor(auth): unify how we handle requests in auth layer

codyoss:
Doing this uniformly will allow us to add things like logging and telemetry more easily in the future.


noahdietz (approved):
yay for context propagation

PR #10507, opened 2024-07-03
chore(storage/transfermanager): DownloadDirectory fails if file already exists

BrennaEpp:
- Also adds cleanup on error.


storage/transfermanager/downloader.go
```diff
@@ -118,12 +165,24 @@ func (d *Downloader) DownloadDirectory(ctx context.Context, input *DownloadDirec
 		// Make sure all directories in the object path exist.
 		err := os.MkdirAll(objDirectory, fs.ModeDir|fs.ModePerm)
 		if err != nil {
+			cleanFiles(inputs)
 			return fmt.Errorf("transfermanager: DownloadDirectory failed to make directory(%q): %w", objDirectory, err)
 		}
 
+		// Check if the file exists.
```
tritone, line 172: I'm wondering if we should just stat everything first in a loop before creating any files, assuming the user doesn't want SkipIfExists behavior? That way we would avoid having to do any cleanup steps. And in general I would expect that, if files already exist, it will be a mistake in terms of the provided path and so we'll encounter the issue fairly quickly.

BrennaEpp: It helps us avoid clean up steps unless there are failures in directory or file creation.

tritone (approved):

PR #10508, opened 2024-07-04
fix(spanner): fix negative values for max_in_use_sessions metrics #10449

rahul2393:
Internal bug: b/343756862

Previous work: https://github.com/googleapis/google-cloud-go/pull/10449


spanner/session.go
```diff
@@ -376,7 +376,7 @@ func (s *session) recycle() {
 		// session pool currently has enough open sessions.
 		s.pool.mu.Unlock()
 		s.destroy(false, true)
-		s.pool.mu.Lock()
+		return
```
olavloite, line 379: Can we have a test for this condition?

olavloite (approved):

PR #10509, opened 2024-07-04
chore(storage/transfermanager): non-positive partSize will turn off sharding

BrennaEpp:


storage/transfermanager/option.go
```diff
@@ -69,13 +69,23 @@ func (wpt withPerOpTimeout) apply(tm *transferManagerConfig) {
...
 
-// WithPartSize returns a TransferManagerOption that specifies the size of the
-// shards to transfer; that is, if the object is larger than this size, it will
-// be uploaded or downloaded in concurrent pieces.
-// The default is 32 MiB for downloads.
+// WithPartSize returns a TransferManagerOption that specifies the size in bytes
+// of the shards to transfer; that is, if the object is larger than partSize,
+// it will be uploaded or downloaded in concurrent pieces of size partSize.
+//
+// The default is 32 MiB for downloads. The minimum partSize allowed is 256kib;
```
tritone, line 76: Let's remove the min value to match other languages, as discussed in chat. We can keep 0 to turn off sharding however.

tritone (approved):

PR #10513, opened 2024-07-05
chore(storage/transfermanager): expand DownloadObjectInput docs

BrennaEpp:


storage/transfermanager/downloader.go
```diff
@@ -452,16 +452,34 @@ type DownloadRange struct {
...
-	// Required fields
-	Bucket      string
-	Object      string
+	// Bucket is the bucket in GCS to download from. Required.
+	Bucket string
+
+	// Object is the object in GCS to download. Required.
+	Object string
+
+	// Destination is the WriterAt to which write the bytes, such as an [os.File]
```
tritone, line 461: ...to which the Downloader will write the bytes? Seems like a word is missing. Also I would say "object data" over "bytes".

tritone (approved):
One wording nit, otherwise lgtm

storage/transfermanager/downloader.go
BrennaEpp: Thanks, that's better

PR #10514, opened 2024-07-06
chore(storage): remove x-goog-api-client header merging from invoke

BrennaEpp:
- go version must still be included since uploads do not always include the version otherwise (other operations will see the value repeated, however)
- xml still has to do the merging
- grpc had a bug where the ctx from invoke was not being used so the header wouldn't always be sent


storage/http_client.go
```diff
@@ -860,9 +860,21 @@ func (c *httpStorageClient) newRangeReaderXML(ctx context.Context, params *newRa
 			// Set custom headers passed in via the context. This is only required for XML;
 			// for gRPC & JSON this is handled in the GAPIC and Apiary layers respectively.
 			ctxHeaders := callctx.HeadersFromContext(ctx)
+
 			for k, vals := range ctxHeaders {
```
tritone, line 864: Can we add a test for this logic? Could either use a mock transport, or factor out the header logic into a separate func for easy testing.

tritone (commented):
LGTM, just needs a test.

tritone (approved):

PR #10521, opened 2024-07-08
feat(compute/metadata): add sys check for windows OnGCE

codyoss:
Tested the windows detection on a Windows VM on cloud.

Fixes: https://github.com/googleapis/google-api-go-client/issues/2117


compute/metadata/go.mod
```diff
@@ -1,3 +1,5 @@
 module cloud.google.com/go/compute/metadata
 
 go 1.20
+
+require golang.org/x/sys v0.22.0
```
noahdietz, line 5: the first dependency 🥲

noahdietz (approved):

compute/metadata/go.mod
liggitt: sad

liggitt: this module specifically is a dependency magnet (xref https://github.com/googleapis/google-cloud-go/issues/6311 https://github.com/kubernetes/kubernetes/issues/113366)

x/sys isn't the end of the world as a transitive dependency, but is there anything guarding against pulling in cloud dependencies for this?

codyoss: No. We will have a PR landing in the next month that will add one more dep to this package though that is a "cloud dep". But that should not pull in any more explicit deps. See https://github.com/googleapis/google-cloud-go/pull/11078

Issue #10517, opened 2024-07-06
"occurrences:vulnerabilitySummary" is missing in apiv1


ypwu1:
**Client**

 `GetVulnerabilitySummary` is only available in `containeranalysis/apiv1beta`.  However `apiv1beta` is marked as deprecated. 

[documentaton](https://cloud.google.com/artifact-analysis/docs/reference/rest/v1/projects.occurrences/getVulnerabilitySummary) do state that it's available  in v1. So what's status quo of this api?

**Environment**

e.g. Alpine Docker on GKE

**Go Environment**

Any version

**Expected behavior**
customer should be able to  send `GetVulnerabilitySummary` request provided by `containeranalysis/apiv1`

**Actual behavior**
Use deprecated api `containeranalysis/apiv1beta`

codyoss:
I will look into getting this client updated, thanks for the report.

PR #10533, opened 2024-07-10
feat(containeranalysis): add GetVulnerabilityOccurrencesSummary RPC

codyoss:
I hand patched this change in because this "generated" client is very hand tailored. I also took the opportunity to update the client in general to match other generated practices today.

Fixes: #10517


noahdietz (approved):
Just need to fix that import per vet check.

PR #10536, opened 2024-07-11
docs(logging): Documentation for automatic trace/span ID extraction

gkevinzheng:
Godoc preview: 
image


logging/doc.go
```diff
@@ -146,5 +146,32 @@ parent.Timestamp marks the end of the request.)
...
+	}
+
+When Entry with an [http.Request] is logged, its Trace, SpanID, and TraceSampled fields may be automatically populated as follows:
+
+ 1. If you are instrumenting your application with [OpenTelemetry] and your http.Request's Context contains an [OpenTelemetry span context],
+    the Entry's Trace, SpanID, and TraceSampled will be populated with information from that span context.
+ 2. Trace, SpanID, and TraceSampled fields will be populated from information from the http.Request's [W3C Traceparent]
+    or [X-Cloud-Trace-Context] headers, if those headers exist.
+
+Note that if Trace, SpanID, or TraceSampled are explicitly provided within an Entry object, those values will take precedence over values automatically
```
mtabasko, line 168: I'd say "object, then those values take precedence"  (a little more direct/less wordy than "will take", and then meaning is the same)

mtabasko (changes requested):

logging/doc.go
```diff
@@ -146,5 +146,32 @@ parent.Timestamp marks the end of the request.)
...
+
+	logging.Entry{
+		HTTPRequest: &logging.HTTPRequest{
+			Request: // Reference to your http.Request here
+		}
+	}
+
+When Entry with an [http.Request] is logged, its Trace, SpanID, and TraceSampled fields may be automatically populated as follows:
+
+ 1. If you are instrumenting your application with [OpenTelemetry] and your http.Request's Context contains an [OpenTelemetry span context],
```
cindy-peng, line 163: Hi Kevin, do we need to specify this "and your http.Request's Context contains an [OpenTelemetry span context]"? It is automatically added when customer's application is instrumented with OpenTelemetry right?

logging/doc.go
```diff
@@ -146,5 +146,32 @@ parent.Timestamp marks the end of the request.)
 You should observe the child log entries grouped under the parent on the console. The
 parent entry will not inherit the severity of its children; you must update the
 parent severity yourself.
+
+# Automatic Metadata Extraction
```
cindy-peng, line 150: Do we want to call this "Automatic Metadata Extraction" or "Automatic Trace/Span Id Extraction" as in https://github.com/googleapis/python-logging/pull/915?

gkevinzheng: I think it would be good to change it to be more specific in this instance.

logging/doc.go
```diff
@@ -146,5 +146,33 @@ parent.Timestamp marks the end of the request.)
...
+
+	logging.Entry{
+		HTTPRequest: &logging.HTTPRequest{
+			Request: // Reference to your http.Request here
+		}
+	}
+
+When Entry with an [http.Request] is logged, its Trace, SpanID, and TraceSampled fields may be automatically populated as follows:
+
+ 1. If you are instrumenting your application with [OpenTelemetry],
```
dashpole, line 163: It might be helpful to be more specific, and reference go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp, since it would be complicated for users to properly to use this without the instrumentation library itself.

dashpole (approved):

mtabasko (approved):

PR #10538, opened 2024-07-11
feat(storage): add full object checksum to reader.Attrs

BrennaEpp:
note - JSON does not currently support returning the checksum for range reads
-  GRPC does not support decompression


storage/reader.go
```diff
@@ -65,6 +65,14 @@ type ReaderObjectAttrs struct {
 	// meaningful in the context of a particular generation of a
 	// particular object.
 	Metageneration int64
+
+	// CRC32C is the CRC32 checksum of the entire object's content using the
+	// Castagnoli93 polynomial, if available.
+	CRC32C uint32
+
+	// Decompressed is true if the object was read decompressed and different
```
tritone, line 73: This seems a little vague, I would make something that more closely echoes the NewRangeReader docs about transcoding (and/or update those docs to clarify that this field will be True if decompressive transcoding occurs).

tritone (approved):
Overall looks good, one doc suggestion.

tritone (approved):

PR #10542, opened 2024-07-12
feat(bigquery): json support on managedwriter/adapt pkg

### LINKED ISSUE #10539: bigquery: support for JSON type
It would be possible to convert a bigquery.Schema to a storagepb.TableSchema using the adapt package functionality if the storagepb library supported JSON data types.

Given a bigquery.schema with a field Payload and a data type JSON, the below code:
`schema, err := adapt.BQSchemaToStorageTableSchema(schema)`
Throws the error:
`"could not convert field (Payload) due to unknown type value: JSON"`

Support for the JSON data type in the bigquery client libraries i.e. storagepb.TableSchema, this functionality exists for Java and Node libraries.

Using a STRING field which takes a stringified JSON instead of an actual JSON - but this is not ideal.

**Additional context**
N/A

alvarowolfx:
Resolves #10539


shollyman (approved):
Thanks for getting this together.

PR #10549, opened 2024-07-15
chore(firestore): use internal/detect

jba:
Use the internal/detect package to detect the project ID, instead
of a custom function.

Also, export the sentinel value from internal/detect.


jba:
Trying as two separate PRs.

Issue #10555, opened 2024-07-16
bigquery: job.wait  inner backoff configuration is to slow


idan3:
I'm using job.wait to know when the job is done. 
The inner configuration for the job.wait function is inside waitForQuery and is:
```
backoff := gax.Backoff{
		Initial:    1 * time.Second,
		Multiplier: 2,
		Max:        60 * time.Second,
	}
```

It is a big problem because the initial of 1 second is too slow if we do a lot of queries in a high rate.
Can you please fix that or give the option to inject this configuration?

Something like : 
```
backoff := gax.Backoff{
		Initial:    80 * time.Millisecond,
		Multiplier: 1.1,
		Max:        60 * time.Second,
	}
```

PR #10558, opened 2024-07-17
fix(bigquery): reduce default backoffs

shollyman:
fix(bigquery): reduce the initial backoff for job polling

This PR reduces initial backoffs and multipliers used during job polling.  Previously we used a 1s starting backoff, this has been reduced to 50ms.  In job execution most time is typically spent in server, waiting for the hanging GetQueryResults response.

This should in generally reduce wasted wait for jobs that poll for multiple iterations.

Fixes: https://github.com/googleapis/google-cloud-go/issues/10555

### DIFF
--- a/bigquery/bigquery.go
+++ b/bigquery/bigquery.go
@@ -234,9 +234,9 @@ func runWithRetry(ctx context.Context, call func() error) error {
 func runWithRetryExplicit(ctx context.Context, call func() error, allowedReasons []string) error {
 	// These parameters match the suggestions in https://cloud.google.com/bigquery/sla.
 	backoff := gax.Backoff{
-		Initial:    50 * time.Millisecond,
+		Initial:    1 * time.Second,
 		Max:        32 * time.Second,
-		Multiplier: 1.3,
+		Multiplier: 2,
 	}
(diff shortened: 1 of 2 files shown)


bigquery/bigquery.go
```diff
@@ -234,9 +234,9 @@ func runWithRetry(ctx context.Context, call func() error) error {
 func runWithRetryExplicit(ctx context.Context, call func() error, allowedReasons []string) error {
 	// These parameters match the suggestions in https://cloud.google.com/bigquery/sla.
```
alvarowolfx, line 235: Sounds like with those changes, it is not going to match the suggestions found in https://cloud.google.com/bigquery/sla

```
"Back-off Requirements" means, when an error occurs, the Customer Application is responsible for waiting for a period of time before issuing another request. This means that after the first error, there is a minimum back-off interval of 1 second and for each consecutive error, the back-off interval increases exponentially up to 32 seconds.
```

...**a minimum back-off interval of 1 second**...

shollyman: Ahh yes.  I'll back out the retry changes.

alvarowolfx (approved):

PR #10561, opened 2024-07-17
chore(firestore): use internal/detect

jba:


bhshkh (approved):

codyoss:
@jba it looks like the test here are failing. Is this change still wanted?

PR #10569, opened 2024-07-21
feat(storage/transfermanager): checksum full object downloads

BrennaEpp:


storage/transfermanager/downloader.go
```diff
@@ -31,6 +33,12 @@ import (
 	"google.golang.org/api/iterator"
 )
 
+// maxChecksumZeroArraySize is the maximum amount of memory to allocate for
+// updating the checksum. A larger size will occupy more memory but will require
+// more updates when computing the crc32c of a full object.
```
tritone, line 38: This seems wrong, I would imagine larger means fewer updates?

storage/transfermanager/downloader.go
```diff
@@ -784,3 +894,57 @@ func setUsageMetricHeader(ctx context.Context, method string) context.Context {
...
 }
+
+type crc32cPiece struct {
+	sum    uint32 // crc32c checksum of the piece
+	length int64  // number of bytes in this piece
+}
+
+// joinCRC32C pieces together the initial checksum with the orderedChecksums
+// provided to calculate the checksum of the whole.
+func joinCRC32C(initialChecksum uint32, orderedChecksums []crc32cPiece) uint32 {
```
tritone, line 905: It seems surprising that this isn't available in hash/crc32 or elsewhere in a standard library, but I assume you looked.

storage/transfermanager/downloader.go
```diff
@@ -366,6 +356,48 @@ func (d *Downloader) downloadWorker() {
 	d.workers.Done()
 }
 
+// startDownload downloads the first shard and schedules subsequent shards
+// if necessary.
+func (d *Downloader) startDownload(input *DownloadObjectInput) {
+	var out *DownloadOutput
+
+	// Special-case: full object read.
```
tritone, line 364: I wouldn't really call this a special case as it is the most common case. Also, only reading partSize bytes is only relevant to large objects which must be sharded.

storage/transfermanager/downloader.go
```diff
@@ -618,9 +634,63 @@ func (in *DownloadObjectInput) downloadShard(client *storage.Client, timeout tim
...
 
-	w := io.NewOffsetWriter(in.Destination, offset)
-	_, err = io.Copy(w, r)
+	var w io.Writer
+	w = io.NewOffsetWriter(in.Destination, offset)
+
+	var crcHash hash.Hash32
+	if in.checkCRC {
+		crcHash = crc32.New(crc32.MakeTable(crc32.Castagnoli))
+		w = io.MultiWriter(w, crcHash)
```
tritone, line 643: nice!

storage/transfermanager/downloader.go
```diff
@@ -784,3 +894,57 @@ func setUsageMetricHeader(ctx context.Context, method string) context.Context {
...
+	}
+	return base
+}
+
+func fullObjectRead(r *DownloadRange) bool {
+	return r == nil || (r.Offset == 0 && r.Length < 0)
+}
+
+func checksumObject(ctx context.Context, object *storage.ObjectHandle, got, want uint32) error {
+	// If we don't have a CRC32C from the reader, do a metadata call to get it.
```
tritone, line 936: Do we still need this? If so under what circumstances? Also, we should explicitly note that generation must be set on `object` for this to work.

tritone (commented):
In general looks really good, a few minor questions/comments. Also, could you add a note to the docs to indicate under what circumstances checksumming will be performed?

storage/transfermanager/downloader.go
BrennaEpp: Correct, I messed up that wording.

storage/transfermanager/downloader.go
BrennaEpp: True, I can remove this comment.

storage/transfermanager/downloader.go
BrennaEpp: Hmm, I guess this is just a fallback, but yeah I don't think it actually gets called - but then again I haven't done super extensive testing.

tritone: I think since Reader doesn't have a fallback if the header is unavailable, we shouldn't have one here either.

tritone (approved):

PR #10584, opened 2024-07-24
feat(bigquery): Add Clone() method to bigquery.Schema

IvanIsCoding:
This is mostly a convenience method. It comes in handy when creating Views from BigQuery, because we want to have a subset of the fields for the View but still want to keep the original FieldSchema entries of the cloned object intact.

There is nothing here that could not be replicated in clients because all the fields of the struct are exported, but it would be good if this was supported upstream.

### DIFF
--- a/bigquery/schema.go
+++ b/bigquery/schema.go
@@ -181,6 +181,7 @@ func (s Schema) Clone() Schema {
 			Scale:                  v.Scale,
 			DefaultValueExpression: v.DefaultValueExpression,
 			Collation:              v.Collation,
+			RoundingMode:           v.RoundingMode,
 		}
 		if v.RangeElementType != nil {
 			newV.RangeElementType = &RangeElementType{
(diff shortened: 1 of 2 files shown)


shollyman:
The PR itself looks straightforward, but I wanted to understand how you're leveraging this.  You mention views, but they're not typically defined via the output schema.  Can you elaborate a bit more?  Thanks!

IvanIsCoding:
> The PR itself looks straightforward, but I wanted to understand how you're leveraging this. You mention views, but they're not typically defined via the output schema. Can you elaborate a bit more? Thanks!

Sure, that is a good question. I mentioned BigQuery Views because my team uses views, but it could also apply to folks copying tables. We want to keep track of what is the schema for queries like:
```
SELECT A, B, C
FROM TABLE
``` 
For us, it is easier to clone the schema from `TABLE` and delete what is not in the `SELECT` statement, because our queries generally have lots of columns. The extreme is the case with `EXCEPT` clauses:
```
SELECT * EXCEPT D, E
FROM TABLE
```
where the clone method with deleting is the best way to keep track of the columns shown for the View.

codyoss:
@shollyman please see response above

PR #10602, opened 2024-07-26
feat(bigtable): expose protoToType

ron-gal:
protoToType needs to be reused by terraform, so we are exposing it.
The trigger for this change is we need in Terraform to accept a json string representing a protobuf of a column family type, so we can properly create and update column family types. We want to be flexible enough to not accept just hardcoded types, but also custom types, thus eliminating the need for client releases every time the server supports a new type.


bhshkh:
Please ignore conformance test failures. They were added in failed state

bhshkh:
Please add integration test

bhshkh (approved):

Issue #9295, opened 2024-01-24
firestore: let BulkWriter retry only when a retriable error occurs


KamikazeZirou:
**Is your feature request related to a problem? Please describe.**

BulkWriter will retry 10 times if it encounters an error that cannot be resolved by retrying. This slows down the process unnecessarily. If the error is not resolved by retrying, it is better to return an error immediately without retrying.

Here is an example.

Suppose you already have a Document on Firestore and call `BulkWriter.Create` on it. In that case, an AlreadyExists error occurs and `BulkWriter` retries 10 times to write in vain. In my environment, it took 10 seconds for all the retry to finish.

**Describe the solution you'd like**

BulkWriter doesn't retry write operations when errors occur that cannot be resolved by retrying.
As with Transaction, why not retry only when the error code is ABORTED?

**Describe alternatives you've considered**

When creating a BulkWriter, it would be good if the module user could specify the errors to retry as an option.

**Additional context**

BulkWriter's retry process is [here](https://github.com/googleapis/google-cloud-go/blob/092ee0ba59267b8fb4d3f4e7727ed3ccbf81e7e7/firestore/bulkwriter.go#L327).

I use BulkWriter's Create and Update as asynchronous processes. This asynchronous processing may be called multiple times and the order of processing is not guaranteed. That's why things like the first example I gave happen.

PR #10603, opened 2024-07-26
fix(firestore): Retry batchwrite only on retryable errors

bhshkh:
Java [retries](https://github.com/googleapis/java-firestore/blob/496498271b31a878910c17954350673beade2bef/google-cloud-firestore/src/main/java/com/google/cloud/firestore/v1/stub/FirestoreStubSettings.java#L849-L852) batchwrite requests only on [retryable error codes](https://github.com/googleapis/java-firestore/blob/496498271b31a878910c17954350673beade2bef/google-cloud-firestore/src/main/java/com/google/cloud/firestore/v1/stub/FirestoreStubSettings.java#L608-L614). Go should do the same.

Fixes: https://github.com/googleapis/google-cloud-go/issues/9295


firestore/bulkwriter.go
```diff
@@ -329,12 +337,13 @@ func (bw *BulkWriter) send(i interface{}) {
 				j.attempts++
 
 				// Do we need separate retry bundler?
-				if j.attempts < maxRetryAttempts {
+				_, isRetryable := batchWriteRetryCodes[codes.Code(s.Code)]
+				if j.attempts >= maxRetryAttempts || !isRetryable {
```
gkevinzheng, line 341: nit: For readability, I think it would be clearer to revert the `if/else` blocks back to original and have the conditional statement be `if j.attempts < maxRetryAttempts && isRetryable` so that someone reading the code won't have to mentally apply Demorgan's Laws to understand what's going on.

gkevinzheng (approved):

PR #10616, opened 2024-07-31
feat(bigtable): first_response_latencies metric

bhshkh:
This PR adds below metrics:
- [first_response_latencies](https://cloud.google.com/bigtable/docs/client-side-metrics-descriptions#first-response-latencies)


gkevinzheng (dismissed):

bigtable/bigtable.go
```diff
@@ -397,6 +400,7 @@ func (t *Table) readRows(ctx context.Context, arg RowSet, f func(Row) bool, mt *
 		*headerMD, _ = stream.Header()
 		for {
 			res, err := stream.Recv()
+			mt.currOp.setFirstRespTime(time.Now())
```
igorbernstein2, line 403: What happens when multiple responses are received?

what happens when there are no rows returned? 

is the behavior consistent with java?

Does this work across retries?

bigtable/bigtable.go
```diff
@@ -1615,9 +1627,15 @@ func recordAttemptCompletion(mt *builtinMetricsTracer) {
...
 
-	// Record server_latencies
+	// Record server_latencies and connectivity_error_count
+	connErrCountAttrs, _ := mt.toOtelMetricAttrs(metricNameConnErrCount)
 	serverLatAttrs, _ := mt.toOtelMetricAttrs(metricNameServerLatencies)
 	if mt.currOp.currAttempt.serverLatencyErr == nil {
 		mt.instrumentServerLatencies.Record(mt.ctx, mt.currOp.currAttempt.serverLatency, metric.WithAttributes(serverLatAttrs...))
 	}
+	if mt.currOp.currAttempt.serverLatencyErr == nil && mt.currOp.currAttempt.locationErr == nil {
+		mt.instrumentConnErrCount.Add(mt.ctx, 0, metric.WithAttributes(connErrCountAttrs...))
```
igorbernstein2, line 1637: do you need to add a zero?

igorbernstein2 (changes requested):

bigtable/bigtable.go
bhshkh: Adding it to be consistent with Java: https://github.com/googleapis/java-bigtable/blob/168293937cc7f438a3ec2dee46805aa8e12089c4/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracer.java#L359-L364

bigtable/bigtable.go
```diff
@@ -437,6 +440,9 @@ func (t *Table) readRows(ctx context.Context, arg RowSet, f func(Row) bool, mt *
 		for {
 			proto.Reset(res)
 			err := stream.RecvMsg(res)
+			if err == nil || err == io.EOF {
+				mt.currOp.setFirstRespTime(time.Now())
```
mutianf, line 444: This is inside of the for loop, would it get updated on every response? How difficult would it be to add a test: emulator sends back 2 responses, sleep before sending the second response, client verifies that the recorded value is <= operation time - sleep before the second response?

bhshkh: I'm using a sync.Once (firstRespTimeOnce) to ensure that the first response time is set only once. The setFirstRespTime function wraps the timestamp assignment in a Do call, so even though it's in the loop, the assignment will only happen on the first response. This is a standard and thread-safe way to handle one-time initializations in Go.
 
 https://pkg.go.dev/sync#Once.Do

bhshkh: Added test

mutianf: so on the subsequent iterations it's just going to be a noop?

bhshkh: yes

daniel-sanche (dismissed):

bigtable/metrics.go
```diff
@@ -347,6 +370,10 @@ type opTracer struct {
 
 	startTime time.Time
 
+	// Only for ReadRows. Time when the response headers are received in a streaming RPC.
+	firstRespTime     time.Time
+	firstRespTimeOnce sync.Once
```
igorbernstein2: I dont think this is the correct usage of this primitive. My understanding is that this carries overhead of a cache flush ("synchronizes before") which seems unecessary here since everything is happening in a single goroutine. It is also a bit confusing at the call site as mattie pointed out

bhshkh: Removed sync.Once

igorbernstein2 (approved):

PR #10619, opened 2024-07-31
fix(storage): add unknown host to retriable errors

BrennaEpp:


storage/invoke.go
```diff
@@ -124,6 +124,10 @@ func ShouldRetry(err error) bool {
 				return true
 			}
 		}
+	case *net.DNSError:
+		if strings.Contains(e.Error(), "no such host") {
```
frankyn, line 128: Could be retriable https://pkg.go.dev/net#DNSError.Temporary; is there more context?

BrennaEpp: Hm, I can't remember why I didn't use this to start off with, but it makes more sense. Updated. Context is just that we want to ensure to retry DNS errors as well.

PR #10620, opened 2024-07-31
fix(auth): disable automatic universe domain check for MDS

quartzmo:


auth/grpctransport/grpctransport.go
```diff
@@ -337,17 +337,19 @@ func (c *grpcCredentialsProvider) getClientUniverseDomain() string {
...
-		return nil, err
-	}
-	if err := transport.ValidateUniverseDomain(c.getClientUniverseDomain(), credentialsUniverseDomain); err != nil {
-		return nil, err
-	}
 	token, err := c.creds.Token(ctx)
 	if err != nil {
 		return nil, err
 	}
+	if source, _ := token.Metadata["auth.google.tokenSource"].(string); source != "compute-metadata" {
```
codyoss, line 344: These maps could be nil. Can you add a check.

quartzmo: 1. Would it be better to always initialize with an empty map?
2. Should I add the check to this existing usage as well? https://github.com/googleapis/google-cloud-go/blob/main/auth/grpctransport/directpath.go#L69-L74

codyoss: Users can create their own token types that we use, so best to be defensive.

quartzmo: Updated, PTAL.

auth/auth.go
```diff
@@ -101,6 +101,20 @@ func (t *Token) IsValid() bool {
 	return t.isValidWithEarlyExpiry(defaultExpiryDelta)
 }
 
+// MetadataString is a convenience method for accessing string values in the
+// token's metadata. Returns an empty string if the metadata is nil or the value
+// for the given key cannot be cast to a string.
+func (t *Token) MetadataString(k string) string {
```
codyoss, line 107: Great idea!

codyoss (approved):

PR #10634, opened 2024-08-02
feat(auth): auth library can talk to S2A over mTLS

xmenxk:
Talking to the secure session agent(s2a) requires mTLS on GCE/GKE, and also serverless in the future. This PR adds the functionality to load the mTLS credentials from disk and use that to talk to s2a.

Also removed the refresh logic for s2a configuration. We decided to only request it from MDS on first use and cache it afterwards.


quartzmo:
Please update your branch, thanks!

auth/internal/transport/cba.go
```diff
@@ -151,7 +169,18 @@ func GetHTTPTransportConfig(opts *Options) (cert.Provider, func(context.Context,
...
 	}
 
-	if config.s2aAddress == "" {
+	var s2aAddr string
+	var transportCredsForS2A credentials.TransportCredentials
+
+	if config.mtlsS2AAddress != "" {
+		s2aAddr = config.mtlsS2AAddress
+		transportCredsForS2A, err = loadMTLSMDSTransportCreds(mtlsMDSRoot, mtlsMDSKey)
+		if err != nil {
```
quartzmo, line 178: 1. Why is this `err` thrown away?
2. Why is the default (non-s2a) configuration being returned in case of an error from `loadMTLSMDSTransportCreds`, rather than falling through to set up `config.s2aAddress`?

xmenxk: 1. thanks for catching this, added a log line

2. in each environment GKE/GCE/Serverless, we either serve mtls_s2a_address or plaintext_s2a_address, not both (there's only one s2a gRPC server running in the infrastructure) 
so if `config.mtlsS2AAddress` is provided, there's no need to try the other `config.s2aAddress`.

quartzmo: Got it, that makes sense.

quartzmo (approved):

PR #10639, opened 2024-08-06
feat(storage): introduce gRPC client-side metrics

frankyn:
Introduce client-side metrics for gRPC client only.

Pending

- [x] Discuss user facing options and default behaviors with @tritone 
- [x] Upgrade gRPC version to 1.66 - https://github.com/googleapis/google-cloud-go/pull/10787
- [x] Run go work sync before merge
- [x] Upgrade OTel version to 1.28 (https://github.com/googleapis/google-cloud-go/pull/10706)
- [x] Error logs are surfaced when permissions aren't granted to the exporter and can get noisy. Java silenced them after the initial error to prevent noise but allows recovery if permissions are granted after release into production.


storage/grpc_client.go
```diff
@@ -136,6 +152,9 @@ func newGRPCStorageClient(ctx context.Context, opts ...storageOption) (storageCl
 }
 
 func (c *grpcStorageClient) Close() error {
+	if c.settings.meterProvider != nil {
```
tritone, line 155: Is it problematic if this never gets called? e.g. will some data not be flushed?

I just ask because calling client.Close() is basically optional; the program could just terminate first. But maybe that's fine.

storage/grpc_client.go
```diff
@@ -116,6 +117,21 @@ type grpcStorageClient struct {
 func newGRPCStorageClient(ctx context.Context, opts ...storageOption) (storageClient, error) {
 	s := initSettings(opts...)
 	s.clientOption = append(defaultGRPCOptions(), s.clientOption...)
+	// TODO: detect project id
```
tritone, line 120: I am assuming we'd have to add an option to enable/disable metrics as well?

storage/grpc_metrics.go
```diff
@@ -0,0 +1,161 @@
...
+	metricViews := getViewMasks(allMetrics)
+	provider := sdkmetric.NewMeterProvider(
+		sdkmetric.WithReader(sdkmetric.NewPeriodicReader(exporter)),
+		sdkmetric.WithResource(preparedResource),
+		sdkmetric.WithView(metricViews...),
+	)
+	return provider, nil
+}
+
+func togRPCDialOption(provider *sdkmetric.MeterProvider) option.ClientOption {
```
tritone, line 147: nit: name seems misleading, maybe `clientOptionFromMeterProvider`. Or just do this inline since the function is only called once.

storage/grpc_metrics.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func gcpAttributeExpectedDefaults() []attribute.KeyValue {
+	return []attribute.KeyValue{
+		{Key: "location", Value: attribute.StringValue(mr_location_label_default)},
+		{Key: "cloud_platform", Value: attribute.StringValue(mr_cloud_platform_label_default)},
+		{Key: "host_id", Value: attribute.StringValue(mr_host_id_label_default)}}
+}
+
+func getPreparedResourceUsingGCPDetector(ctx context.Context, project string) (*resource.Resource, error) {
```
tritone, line 91: nit: This func also seems superfluous and might be clearer in-line

storage/grpc_metrics.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	monitored_resource_name         = "storage.googleapis.com/Client"
+	mr_location_label_default       = "global"
+	mr_cloud_platform_label_default = "unknown"
+	mr_host_id_label_default        = "unknown"
+	metric_format_prefix            = "storage.googleapis.com/client/"
+	metric_lb_locality_label        = "grpc.lb.locality"
+)
+
+func getMetricsToEnable() []estats.Metric {
```
tritone, line 42: This and `getMetricsEnabledByDefault` could probably be local vars rather than functions given that they are just static data.

storage/grpc_metrics_test.go
```diff
@@ -0,0 +1,157 @@
...
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	"go.opentelemetry.io/otel/sdk/resource"
+	"google.golang.org/api/iterator"
+	estats "google.golang.org/grpc/experimental/stats"
+)
+
+func TestMetrics(t *testing.T) {
+	ctx := context.Background()
+	grpcClient, err := NewGRPCClient(ctx)
```
tritone, line 32: Does this work without auth? If not we should probably make it an integration test. Or try `option.WithoutAuthentication`

tritone (commented):
Overall looks pretty good. We should talk about default behavior and user facing options.

storage/grpc_metrics_test.go
frankyn: Requires auth; I need to update to not run with short tests.

storage/grpc_metrics.go
frankyn: I wrote it this way to write tests; should I be doing something different to make it testable?

storage/grpc_metrics.go
```diff
@@ -0,0 +1,295 @@
...
+	endpointToUse := determineMonitoringEndpoint(config.endpoint)
+	if endpointToUse != "" {
+		meOpts = append(meOpts, mexporter.WithMonitoringClientOptions(option.WithEndpoint(endpointToUse)))
+	}
+	exporter, err := mexporter.New(meOpts...)
+	if err != nil {
+		return nil, err
+	}
+	metricViews := []metric.View{
+		createHistogramView("grpc.client.attempt.duration", "A view of grpc.client.attempt.duration with histogram boundaries more appropriate for Google Cloud Storage RPCs", "s", latencyHistogramBoundaries()),
```
tritone, line 202: What are these description strings used for?

storage/grpc_metrics.go
```diff
@@ -0,0 +1,295 @@
...
+	})
+}
+
+func newGRPCMetricContext(ctx context.Context, config internalMetricsConfig) (*internalMetricsContext, error) {
+	preparedResource, err := newPreparedResource(ctx, config.project, []resource.Option{resource.WithDetectors(gcp.NewDetector())})
+	if err != nil {
+		return nil, err
+	}
+	if config.project != preparedResource.projectToUse {
+		log.Printf("The Project ID configured for metrics is %s, but the Project ID of the storage client is %s. Make sure that the service account in use has the required metric writing role (roles/monitoring.metricWriter) in the project projectIdToUse or metrics will not be written.", preparedResource.projectToUse, config.project)
```
tritone, line 186: How would this arise? Also this should probably return an error from this func?

storage/grpc_metrics.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	meOpts := []mexporter.Option{
+		mexporter.WithProjectID(preparedResource.projectToUse),
+		mexporter.WithMetricDescriptorTypeFormatter(metricFormatter),
+		mexporter.WithCreateServiceTimeSeries(),
+		mexporter.WithMonitoredResourceDescription(monitoredResourceName, []string{"project_id", "location", "cloud_platform", "host_id", "instance_id", "api"})}
```
tritone, line 192: Are these metric names supposed to align with metrics collected elsewhere? Just wondering what it would take to add a new attribute for example.

storage/grpc_metrics.go
```diff
@@ -0,0 +1,295 @@
...
+		endpoint:   endpointToUse,
+		clientOpts: []option.ClientOption{do},
+		provider:   provider,
+		close:      createShutdown(ctx, provider),
+	}
+	return context, nil
+}
+
+func enableClientMetrics(ctx context.Context, s *settings) (*internalMetricsContext, error) {
+	_, ep, err := htransport.NewClient(ctx, s.clientOption...)
```
tritone, line 237: It seems kind of backwards that we need to initialize a new http client just in order to capture the endpoint. Am I reading this correctly? Is there not another way to get this endpoint?

storage/grpc_metrics.go
```diff
@@ -0,0 +1,295 @@
...
+		resource.WithAttributes(updates...),
+	)
+	if err != nil {
+		return nil, err
+	}
+	preparedResource.resource = r
+	return preparedResource, nil
+}
+
+type internalMetricsConfig struct {
```
tritone, line 135: I don't think this needs to be a separate type, IIUC it's just being used to pass around these two values correct?

storage/grpc_metrics.go
```diff
@@ -0,0 +1,295 @@
...
+	return metricsContext, nil
+}
+
+func createShutdown(ctx context.Context, provider *metric.MeterProvider) func() {
+	return func() {
+		provider.Shutdown(ctx)
+	}
+}
+
+// Silences permission errors after initial error is emitted to prevent
```
tritone, line 263: Maybe explain that this implements the interface for `Exporter` as well? Otherwise it's confusing what's going on here.

storage/grpc_metrics_test.go
tritone: If it's an integration test, move to the file with the other integration tests.

storage/grpc_metrics_test.go
```diff
@@ -0,0 +1,208 @@
...
+	if testing.Short() && !replaying {
+		t.Skip("Integration tests skipped in short mode")
+	}
+	ctx := context.Background()
+	grpcClient, err := NewGRPCClient(ctx)
+	if err != nil {
+		log.Fatalf("Error setting up gRPC client: %v", err)
+	}
+	defer grpcClient.Close()
+	bucket := grpcClient.Bucket("anima-frank-gcs-grpc-team-test-central1")
```
tritone, line 40: Can a bucket be created in the test rather than using this magic one?

tritone (commented):
Sorry for the number of additional comments -- overall looks pretty good, could just use some more cleanup in a few places including docs.

storage/grpc_metrics.go
```diff
@@ -0,0 +1,295 @@
...
+		boundaries = append(boundaries, boundary)
+		if i != 0 && i%10 == 0 {
+			increment *= 2
+		}
+		boundary += increment
+	}
+	return boundaries
+}
+
+func sizeHistogramBoundaries() []float64 {
```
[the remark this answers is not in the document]
frankyn: Struct expects float64 https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric@v1.28.0#AggregationExplicitBucketHistogram

storage/grpc_metrics.go
frankyn: metric names are collected inside of gRPC Go client. I think we can extend the metrics collected to include others collected; checking with Timothy.

storage/grpc_metrics_test.go
frankyn: I removed this test as it was mainly a smoke test to make sure nothing failed when creating the gRPC client as I implemented the feature. Existing IT cover if metrics broke construction of NewGRPCClient unintentionally.

storage/grpc_metrics.go
```diff
@@ -0,0 +1,295 @@
...
+		boundary += increment
+		if boundary >= 4*mb {
+			increment *= 2
+		}
+	}
+	return boundaries
+}
+
+func metricFormatter(m metricdata.Metrics) string {
+	return "storage.googleapis.com/client/" + strings.ReplaceAll(string(m.Name), ".", "/")
```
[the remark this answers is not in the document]
frankyn: No, from tests, monitored resource is capital Client and metric prefix is lower client. Pulling it up into consts.

storage/grpc_metrics.go
frankyn: They're exposed to the user in Monitoring Metric Explorer to describe the metric. That said, it should be cached so clarifying if we really need to have this hardcoded client side.

storage/grpc_metrics.go
frankyn: We can extend them but to surface them to a customer they need to be allowlisted.

storage/grpc_metrics.go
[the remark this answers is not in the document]
frankyn: Monitred resource requires `storage.googleapis.com/Client` while the metric is `storage.googleapis.com/client`

storage/grpc_metrics.go
```diff
@@ -0,0 +1,291 @@
...
+type exporterLogSuppressor struct {
+	exporter       metric.Exporter
+	emittedFailure bool
+}
+
+// Implements OTel SDK metric.Exporter interface to prevent noisy logs from
+// lack of credentials after initial failure.
+// https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric@v1.28.0#Exporter
+func (e *exporterLogSuppressor) Export(ctx context.Context, rm *metricdata.ResourceMetrics) error {
+	if err := e.exporter.Export(ctx, rm); err != nil && !e.emittedFailure {
```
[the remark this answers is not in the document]
frankyn: The intention is that once it's deployed it can recover once appropriate permissions are provided otherwise a redeployment would be required.

storage/grpc_metrics.go
frankyn: I'm be confused in NewGRPCClient() there is no setting to track endpoint similar to [NewClient()](https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L185-L212). I can move in this block into NewGRPCClient() instead if that's what you're asking for. 

The endpoint is used to check if restricted.googleapis.com or private.googleapis.com is used for storage then monitoring API endpoint is changed to that.

storage/grpc_metrics.go
frankyn: Implementation uses OTel Detector to determine project-id to use and if that project is different from the project used by the storage client then this warning is emitted so it's clear which project is selected. If detector doesn't detect a project it will fallback to ADC. However, if no project is found then client-side metrics cannot start. I'll add a case for this.

storage/grpc_metrics.go
tritone: Discussed this offline; we should probably add a new func in google-api-go-client transport/http to just get the endpoint. @codyoss what do you think?

frankyn: Synced with @codyoss, he guided me through using manually initializing a ConnPool and passing it into the GAPIC client view grpc options. It does require duplicating default options from GAPIC client.

frankyn: Spoke too soon, I need the endpoint value before creating connPool; so will need to come up with a different solution. I'll file an issue for this; @tritone do you prefer one over the other in terms of less bad implementation? 

NewHTTPClient: https://github.com/googleapis/google-cloud-go/blob/1c537583857886881d3c3c87b8ad56d56cad22a1/storage/grpc_metrics.go#L233-L236
Double ConnPool: https://github.com/googleapis/google-cloud-go/pull/10639/commits/c0610c8cda5215cd93afda88b2ba1972b458122c

frankyn: Removed this entirely. no longer an issue.

danielduhh (approved):

PR #10646, opened 2024-08-07
feat(spanner): add support of multiplexed session support in writeAtleastOnce mutations

rahul2393:
* This PR add support of using Multiplexed session if enabled with `client.Apply(context.Background(), {mutations}, spanner.ApplyAtLeastOnce())`

- [X] Integration tests with multiplex session enabled pass http://sponge2/9fb45f05-a741-4efd-8252-85a1d94d8f0b

### DIFF
--- a/spanner/transaction.go
+++ b/spanner/transaction.go
@@ -1890,7 +1890,7 @@ func (t *writeOnlyTransaction) applyAtLeastOnce(ctx context.Context, ms ...*Muta
 		for {
 			if sh == nil || sh.getID() == "" || sh.getClient() == nil {
 				// No usable session for doing the commit, take one from pool.
-				sh, err = t.sp.take(ctx)
+				sh, err = t.sp.takeMultiplexed(ctx)
 				if err != nil {
 					// sessionPool.Take already retries for session
 					// creations/retrivals.
@@ -1912,6 +1912,7 @@ func (t *writeOnlyTransaction) applyAtLeastOnce(ctx context.Context, ms ...*Muta
 				RequestOptions: createRequestOptions(t.commitPriority, "", t.transactionTag),
 			})
 			if err != nil && !isAbortedErr(err) {
+				// should not be the case with multiplexed sessions
 				if isSessionNotFoundError(err) {
 					// Discard the bad session.
 					sh.destroy()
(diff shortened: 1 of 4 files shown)


spanner/transaction_test.go
```diff
@@ -163,13 +163,30 @@ func TestApply_Single(t *testing.T) {
...
-	if _, err := shouldHaveReceived(server.TestSpanner, []interface{}{
+	requests := drainRequestsFromServer(server.TestSpanner)
+	expectedReqs := []interface{}{
 		&sppb.BatchCreateSessionsRequest{},
 		&sppb.CommitRequest{},
-	}); err != nil {
+	}
+	if isMultiplexEnabled {
+		expectedReqs = []interface{}{
+			&sppb.CreateSessionRequest{},
```
harshachinta, line 173: When multiplexed session is enabled, does it not create regular sessions to be used for R/W transactions?
Should there not be a BatchCreateSessionRequest here?

rahul2393: No it does not create it, because MinOpened is set to 0 in test config.

spanner/transaction.go
```diff
@@ -1890,7 +1890,7 @@ func (t *writeOnlyTransaction) applyAtLeastOnce(ctx context.Context, ms ...*Muta
 		for {
 			if sh == nil || sh.getID() == "" || sh.getClient() == nil {
 				// No usable session for doing the commit, take one from pool.
-				sh, err = t.sp.take(ctx)
+				sh, err = t.sp.takeMultiplexed(ctx)
```
harshachinta, line 1893: I assume that it falls back to regular sessions in case there is an issue with multiplexed sessions?

rahul2393: Yes it will fallback to regular session, that logic is handled in `takeMultiplexed` because that behaviour is common

harshachinta:
LGTM.
One quick thought. I don't think we need a different environment variable to support multiplexed sessions in blind writes and instead when GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS is enabled it will automatically support blind writes also.

harshachinta (approved):

rahul2393:
> LGTM.
One quick thought. I don't think we need a different environment variable to support multiplexed sessions in blind writes and instead when GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS is enabled it will automatically support blind writes also.



We are not using different env var

PR #10658, opened 2024-08-08
fix(bigtable): Error logging for client side metrics

bhshkh:
1. Use Bigtable client options while creating Metrics client. For e.g. if service account key is getting used to create Bigtable client, same will be used for GCM client.
E.g.
```Go
        serviceAccountKeyFile := "key.json"
	credOpt := option.WithCredentialsFile(serviceAccountKeyFile)
	adminClient, err = bigtable.NewAdminClient(ctx, project, instance, credOpt)
```
2. Prefixed the errors with "bigtable-metrics: "
Unless the user has set their own error handler, error log will be of the format below:

```
2024/08/28 15:06:46 bigtable-metrics: rpc error: code = PermissionDenied desc = Permission monitoring.timeSeries.create denied (or the resource may not exist).
```

3. Moving tests for metrics_util.go into their own file.


bigtable/metrics.go
```diff
@@ -66,11 +67,20 @@ const (
...
 	defaultSamplePeriod = 5 * time.Minute
 
+	// Logger used to log errors seen while collecting and exporting metrics
+	// Log format is yyyy/MM/dd HH:mm:ss bigtable: .........
+	// Date and time are in local timezone
+	//
+	// Errors seen while collecting and exporting metrics should not fail user operation.
+	// So, not all errors are bubbled up to the user.
+	// These error logs will help in debugging any metrics related issues.
+	metricsLogger = log.New(os.Stderr, "bigtable: ", log.LstdFlags|log.Lmsgprefix)
```
gkevinzheng, line 82: nit: If this is only used for logging errors, could we call it `metricsErrorLogger`?

gkevinzheng (approved):

bigtable/metrics_monitoring_exporter.go
```diff
@@ -236,15 +248,15 @@ func (me *monitoringExporter) recordToTimeSeriesPb(m otelmetricdata.Metrics) ([]
 	default:
 		errs = append(errs, errUnexpectedAggregationKind{kind: reflect.TypeOf(m.Data).String()})
 	}
-	return tss, errors.Join(errs...)
+	return tss, me.logErrorAndReturn("recordToTimeSeriesPb", errors.Join(errs...))
 }
 
-func sumToTimeSeries[N int64 | float64](point otelmetricdata.DataPoint[N], metrics otelmetricdata.Metrics, mr *monitoredrespb.MonitoredResource) (*monitoringpb.TimeSeries, error) {
-	interval, err := toNonemptyTimeIntervalpb(point.StartTime, point.Time)
+func (me *monitoringExporter) sumToTimeSeries(point otelmetricdata.DataPoint[int64], metrics otelmetricdata.Metrics, mr *monitoredrespb.MonitoredResource) (*monitoringpb.TimeSeries, error) {
```
bhshkh, line 254: Need access to error logger. So, changing this function to a method on monitoringExporter.
Also, removing generics since we only use int64 and Go does not allow generic methods.

Below code would throw error `method must have no type parameters`
```
func (me *monitoringExporter) sumToTimeSeries[N int64 | float64](point otelmetricdata.DataPoint[int64], metrics otelmetricdata.Metrics, mr *monitoredrespb.MonitoredResource)
```

bigtable/metrics.go
```diff
@@ -144,10 +160,10 @@ type builtinMetricsTracerFactory struct {
 	retryCount         metric.Int64Counter
 }
 
-func newBuiltinMetricsTracerFactory(ctx context.Context, project, instance, appProfile string, metricsProvider MetricsProvider) (*builtinMetricsTracerFactory, error) {
+func newBuiltinMetricsTracerFactory(ctx context.Context, project, instance, appProfile string, metricsProvider MetricsProvider, opts ...option.ClientOption) (*builtinMetricsTracerFactory, error) {
 	clientUID, err := generateClientUID()
 	if err != nil {
-		log.Printf("built-in metrics: generateClientUID failed: %v. Using empty string in the %v metric atteribute", err, metricLabelKeyClientUID)
+		metricsErrorLogger.Printf("built-in metrics: generateClientUID failed: %v. Using empty string in the %v metric attribute", err, metricLabelKeyClientUID)
```
igorbernstein2, line 166: do you need the `built-in metrics:` prefix in the msg? I thought it would come from the logger config already?

bigtable/metrics_monitoring_exporter.go
```diff
@@ -86,31 +88,41 @@ func newMonitoringExporter(ctx context.Context, project string, opts ...option.C
...
+	if err != nil && me.errLogger != nil {
+		me.errLogger.Println(fmt.Errorf("%v: %w", src, err).Error())
+	}
+	return err
+}
+
 // ForceFlush does nothing, the exporter holds no state.
-func (e *monitoringExporter) ForceFlush(ctx context.Context) error { return ctx.Err() }
+func (me *monitoringExporter) ForceFlush(ctx context.Context) error {
+	return me.logErrorAndReturn("ForceFlush", ctx.Err())
```
igorbernstein2, line 104: generally logging + return errors is an anti pattern,  I think you should revert this to just return an error

bigtable/metrics_monitoring_exporter.go
```diff
@@ -86,31 +88,41 @@ func newMonitoringExporter(ctx context.Context, project string, opts ...option.C
 		client:    client,
 		shutdown:  make(chan struct{}),
 		projectID: project,
+		errLogger: errLogger,
 	}, nil
 }
 
+func (me *monitoringExporter) logErrorAndReturn(src string, err error) error {
```
igorbernstein2, line 95: same as below, logging + returning error is a an anti pattern. If you are returning an error, it should be bubbled up and logged in a central location. Does the otel api not log this?

bigtable/metrics_monitoring_exporter.go
```diff
@@ -86,31 +88,41 @@ func newMonitoringExporter(ctx context.Context, project string, opts ...option.C
...
-	return err
+	return me.logErrorAndReturn("Shutdown", err)
 }
 
 // Export exports OpenTelemetry Metrics to Google Cloud Monitoring.
 func (me *monitoringExporter) Export(ctx context.Context, rm *otelmetricdata.ResourceMetrics) error {
 	select {
 	case <-me.shutdown:
-		return errShutdown
+		return me.logErrorAndReturn("Export", errShutdown)
```
igorbernstein2, line 121: I dont think we need to log shutdown?

bigtable/metrics_monitoring_exporter.go
```diff
@@ -86,31 +88,41 @@ func newMonitoringExporter(ctx context.Context, project string, opts ...option.C
...
 	default:
 	}
 
-	return me.exportTimeSeries(ctx, rm)
+	return me.logErrorAndReturn("Export", me.exportTimeSeries(ctx, rm))
```
igorbernstein2, line 125: plz revert

igorbernstein2 (changes requested):

bhshkh:
@igorbernstein2 
As discussed, I have removed all the log writes and just prefixed the error messages with bigtable-metrics

bigtable/bigtable.go
```diff
@@ -1581,6 +1581,9 @@ func gaxInvokeWithRecorder(ctx context.Context, mt *builtinMetricsTracer, method
 
 		// f makes calls to CBT service
 		err := f(ctx, &attemptHeaderMD, &attempTrailerMD, callSettings)
+		if !mt.builtInEnabled {
+			return err
+		}
```
gkevinzheng, line 1586: Why do bigtable calls fail if client side metrics is disabled?

bigtable/metrics_test.go
```diff
@@ -115,14 +117,18 @@ func TestNewBuiltinMetricsTracerFactory(t *testing.T) {
...
-		option.WithGRPCDialOption(grpc.WithTransportCredentials(insecure.NewCredentials())),
+
+	// Override exporter options
+	origCreateExporterOptions := createExporterOptions
+	createExporterOptions = func(opts ...option.ClientOption) []option.ClientOption {
+		return []option.ClientOption{
+			option.WithEndpoint(monitoringServer.Endpoint), // Connect to mock
+			option.WithoutAuthentication(),
+			option.WithGRPCDialOption(grpc.WithTransportCredentials(insecure.NewCredentials())),
+		}
```
gkevinzheng, line 128: Manually overriding/reverting `createExporterOptions` could be very error prone in the future, especially since you can very easily forget to save the original function. Could you create a utility function in the tests to automatically save the original function then override it, like https://github.com/googleapis/google-cloud-go/blob/7e4fbc5612441c59bfaa1e5b9bbd06e3387b5c02/logging/logging_unexported_test.go#L413-L421?

bigtable/metrics.go
```diff
@@ -144,10 +150,17 @@ type builtinMetricsTracerFactory struct {
 	retryCount         metric.Int64Counter
 }
 
-func newBuiltinMetricsTracerFactory(ctx context.Context, project, instance, appProfile string, metricsProvider MetricsProvider) (*builtinMetricsTracerFactory, error) {
+func wrapMetricsError(err error) error {
+	if err == nil {
+		return err
+	}
```
gkevinzheng, line 156: nit: Is the `nil` case really that necessary? Most of the time this is being called is because `err` is non-null, and it's also a private function.

bigtable/metrics_monitoring_exporter.go
bhshkh: Reverted

bigtable/metrics_monitoring_exporter.go
bhshkh: Reverted logging

bigtable/metrics_monitoring_exporter.go
bhshkh: Only wrapping now

bigtable/metrics_monitoring_exporter.go
bhshkh: wrapping and returning error

bigtable/metrics.go
bhshkh: Removed

bigtable/metrics.go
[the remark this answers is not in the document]
bhshkh: Done

bigtable/metrics.go
bhshkh: Yes. This is needed.
In the exporter file, we don't always check whether the err is nil before calling wrap.

E.g. ForceFlush, exportTimeSeries, etc

bigtable/bigtable.go
bhshkh: They don't
This is sort of short cicuiting the code. The lines after this line record metrics. If built in metrics has been disabled, we don't want to record anything.

I have refactored the code to make the behavior more obvious.

bigtable/metrics_test.go
bhshkh: Done

igorbernstein2 (approved):

PR #10659, opened 2024-08-09
feat(bigtable): Remove deprecated Bytes from BigEndianBytesEncoding

ron-gal:


steveniemitz (approved):

steveniemitz:
This is technically an API breaking change, but given it only existed for ~a day in 1.28 I'm fine with this change.

PR #10660, opened 2024-08-09
fix(spanner): json null handling

egonelbre:
Improves performance and fixes a potential corner case with regards to mutations.


rahul2393:
Hello @egonelbre Can you please share what was the corner case with regards to mutation?

egonelbre:
@rahul2393 I added a test to show the main idea. Basically, when a user changes `MarshalJSON` result, they might accidentally change the underlying `jsonNullBytes` content, causing all following calls to return the wrong data.

egonelbre:
@rahul2393 in case it's not clear, I'm not able to merge any of my PR-s myself.

PR #10666, opened 2024-08-10
fix(spanner): encodeValue to use the Encoder interface before using generic proto serialization

### LINKED ISSUE #10662: spanner: encodeValue should prefer the Encoder interface before using generic proto serialization
**Client**

spanner

**Expected behavior**

When encodeValue encounters a proto message that implements the Encoder interface, EncodeSpanner should be used before generic proto encoding.

**Actual behavior**

encodeValue assumes that proto message types are being stored in a proto field type. This breaks backwards compatibility for custom types we've been using.

**Additional context**

e.g. Started after upgrading to v1.62.0

kai5263499:
Fixes #10662


harshachinta:
@kai5263499 
Encoder and Decoder are supported for custom types only. Here since the type is directly a proto message it does not fall under this category. For ex, you cannot add Encoder support for `int64` directly, and instead achieve by creating a `custom type` whose underlying type is `int64` ([Ref](https://github.com/googleapis/google-cloud-go/blob/1e21bcf262bca9781fd42e357cdddf0ed9ad9f1e/spanner/value_test.go#L247)).

You can achieve the same for proto by doing the below,
```
// SingerInfo is a proto message
type customSingerInfo SingerInfo

// Define methods on the custom type
func (v *customSingerInfo) EncodeSpanner() (interface{}, error) {
	return (*SingerInfo)(v).String(), nil
}
```

PR #10668, opened 2024-08-12
feat(spanner): support commit options in mutation operations.

rahul2393:


spanner/client.go
```diff
@@ -937,12 +946,17 @@ func (c *Client) Apply(ctx context.Context, ms []*Mutation, opts ...ApplyOption)
 	defer func() { trace.EndSpan(ctx, err) }()
 
 	if !ao.atLeastOnce {
+		var commitOptions CommitOptions
+		if ao.commitOptions != nil {
+			commitOptions = *ao.commitOptions
+		}
```
olavloite, line 952: I don't quite understand why this is different depending on whether `ao.atLeastOnce` is true or not.

spanner/transaction.go
```diff
@@ -1859,6 +1859,9 @@ type writeOnlyTransaction struct {
 	// current transaction from the allowed tracking change streams with DDL option
 	// allow_txn_exclusion=true.
 	excludeTxnFromChangeStreams bool
+	// maxCommitDelay is the maximum time that the commit will be delayed by the
+	// backend before it is acknowledged.
```
olavloite, line 1863: (The field here is not `maxCommitDelay`, but `commitOptions`)

```suggestion
	// commitOptions are applied to the Commit request for the writeOnlyTransaction.
```

spanner/integration_test.go
```diff
@@ -1498,7 +1498,11 @@ func TestIntegration_ReadWriteTransaction_StatementBased(t *testing.T) {
 		Insert("Accounts", []string{"AccountId", "Nickname", "Balance"}, []interface{}{int64(1), "Foo", int64(50)}),
 		Insert("Accounts", []string{"AccountId", "Nickname", "Balance"}, []interface{}{int64(2), "Bar", int64(1)}),
 	}
-	if _, err := client.Apply(ctx, accounts, ApplyAtLeastOnce()); err != nil {
+	duration, err := time.ParseDuration("100ms")
```
olavloite, line 1501: Can we also add mock server tests for both normal read/write transactions and writeAtLeastOnce transactions that verify that the maxCommitDelay that has been set is actually sent to Spanner in both cases? This integration test only shows that setting the option does not break anything, it is not able to determine whether the option was actually sent or not.

rahul2393: Added the tests

olavloite (approved):

Issue #10672, opened 2024-08-12
bigquery: ArrowIterator Next is not thread-safe and might be leaking goroutines.


alvarowolfx:
Per discussion with customer, we found points of issues:

* ArrowIterator.Next is not Thread-safe
    * init call needs locks
    * first `<-it.records` call can be potentially removed
    * if multiple errors lock the err channel, context is not checked

* consumeStream needs to check context done

PR #10673, opened 2024-08-12
refactor(bigquery): single ctx usage with storage read

alvarowolfx:
Towards #10672


shollyman (approved):
nit: consider expanding the PR description, most of the diffs here are related to the internal field rename rather than the ctx change.  Other than that, LGTM

alvarowolfx:
> nit: consider expanding the PR description, most of the diffs here are related to the internal field rename rather than the ctx change. Other than that, LGTM

updated description

PR #10675, opened 2024-08-12
fix(bigquery): processStream check ctx done when queuing non retryable err

alvarowolfx:
When service side returns non-retryable errors faster than they are being consumed by the user (via `Next` call), the `processStream` goroutine can become locked and leak, even when context is cancelled. With this PR, we wrap the call to `it.errs` call with `select` and check for context done. 

Towards #10672


shollyman (approved):
Thanks for consolidating the checks, definitely cleans up that process loop a bit.

PR #10679, opened 2024-08-13
feat(bigtable): Add MarshalJSON to allow clients to get a stringified version of the protobuf

ron-gal:


bigtable/type.go
```diff
@@ -16,12 +16,26 @@ limitations under the License.
...
+	btapb "cloud.google.com/go/bigtable/admin/apiv2/adminpb"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
 
 // Type wraps the protobuf representation of a type. See the protobuf definition
 // for more details on types.
 type Type interface {
 	proto() *btapb.Type
+	ToJSON() (string, error)
```
codyoss, line 29: Why is this needed in this library? This seems like code that could be written where it is needed as it does not access any private data to the interface.

ron-gal: It's accessing proto() that is private.
We opted to go with https://medium.com/@chaewonkong/go-and-json-a-comprehensive-guide-to-working-with-json-in-golang-143fa2dfa897#:~:text=of%20an%20integer.-,Custom%20types,-%3A instead though :)

bigtable/type.go
```diff
@@ -32,6 +36,14 @@ func (u unknown[T]) proto() *T {
 	return u.wrapped
 }
 
+// MarshalJSON returns the string representation of the protobuf.
+func (u unknown[T]) MarshalJSON() ([]byte, error) {
+	if t, ok := any(u.wrapped).(proto.Message); ok {
+		return protojson.MarshalOptions{}.Marshal(t)
```
codyoss, line 42: Consider using the same marshal settings we use in GAPICs. Example: https://github.com/googleapis/google-cloud-go/blob/60ad7f32d69ed3986fb1326841bba2c5dc36e508/secretmanager/apiv1/secret_manager_client.go#L950

ron-gal: Done

bigtable/type.go
```diff
@@ -32,6 +39,14 @@ func (u unknown[T]) proto() *T {
 	return u.wrapped
 }
 
+// MarshalJSON returns the string representation of the protobuf.
+func (u unknown[T]) MarshalJSON() ([]byte, error) {
+	if t, ok := any(u.wrapped).(proto.Message); ok {
```
steveniemitz, line 44: can we force wrapped to be a `proto.Message*` instead?

bigtable/type_test.go
```diff
@@ -55,38 +109,29 @@ func TestStringProto(t *testing.T) {
...
+	gotJSON, err := MarshalJSON(st)
+	if err != nil {
+		t.Fatalf("Error calling ToJSON: %v", err)
 	}
-
-	got := AggregateType{Input: Int64Type{}, Aggregator: SumAggregator{}}.proto()
-	if !proto.Equal(got, want) {
-		t.Errorf("got type %v, want: %v", got, want)
+	wantJSON := "{\"stringType\":{\"encoding\":{\"utf8Raw\":{}}}}"
+	if strings.ReplaceAll(string(gotJSON), " ", "") != wantJSON {
```
steveniemitz, line 123: rather than testing that the json equals a known string, I think we can instead just test that round tripping the object from json and back yield the "correct" thing?  Comparing json strings has the potential for false errors.

ron-gal: Done

bigtable/type.go
```diff
@@ -16,14 +16,40 @@ limitations under the License.
...
+)
 
 // Type wraps the protobuf representation of a type. See the protobuf definition
 // for more details on types.
 type Type interface {
 	proto() *btapb.Type
 }
 
+var marshalOptions = protojson.MarshalOptions{AllowPartial: true, UseEnumNumbers: true}
+var unmarshalOptions = protojson.UnmarshalOptions{AllowPartial: true, DiscardUnknown: true}
```
steveniemitz, line 32: I dont think we want DiscardUnknown? assuming that does what it sounds like.

bigtable/type_test.go
```diff
@@ -37,12 +39,54 @@ func aggregateProto() *btapb.Type {
...
+	it := Int64Type{Encoding: BigEndianBytesEncoding{}}
+	got := it.proto()
 	if !proto.Equal(got, want) {
 		t.Errorf("got type %v, want: %v", got, want)
 	}
+
+	gotJSON, err := MarshalJSON(it)
+	if err != nil {
+		t.Fatalf("Error calling MarshalJSON: %v", err)
+	}
```
steveniemitz, line 79: I think this could be pulled out into a helper method? it's basically the same for each type right?

bigtable/type.go
ron-gal: Done

PR #10683, opened 2024-08-14
feat(storage): read stall timeout

tritone:
Adds a RetryOption to set a timeout on stalled reads during downloads. If individual calls to body.Read() exceed this timeout, the connection will be closed and reopened from the last read offset.

This timeout will also be applied to gRPC reads and to the initial read request in future PRs.


BrennaEpp (commented):
LGTM, are we adding this to grpc as well?

tritone:
Decided to go with a min throughput option instead, see #10798

PR #10686, opened 2024-08-14
chore: upgrade to 1.21

codyoss:


.github/workflows/vet.yml
```diff
@@ -9,14 +9,17 @@ on:
...
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-go@v5
       with:
-        go-version: '1.20.x'
+        go-version: '1.21.x'
     - name: Install tools
       run: |
         go install golang.org/x/lint/golint@latest && \
```
quartzmo, line 25: Can you restore `go install honnef.co/go/tools/cmd/staticcheck@v0.4.7` to `go install honnef.co/go/tools/cmd/staticcheck@latest` on line 27, below? Thanks!

internal/generated/snippets/go.mod
```diff
@@ -1,6 +1,8 @@
 module cloud.google.com/go/internal/generated
 
-go 1.20
+go 1.21
+
+toolchain go1.21.13
```
quartzmo, line 5: Just curious, why does this go.mod get `toolchain`, when others do not?

spanner/test/opentelemetry/test/go.mod
```diff
@@ -1,6 +1,8 @@
 module cloud.google.com/go/spanner/test
 
-go 1.20
+go 1.21
+
+toolchain go1.21.13
```
quartzmo, line 5: Same question here also, why `toolchain`?

storage/internal/benchmarks/go.mod
```diff
@@ -1,6 +1,8 @@
 module main
 
-go 1.20
+go 1.21
+
+toolchain go1.21.13
```
quartzmo, line 5: Same question here also, why `toolchain`?

spanner/test/opentelemetry/test/go.mod
codyoss: same as above.

storage/internal/benchmarks/go.mod
codyoss: same as above.

.github/workflows/vet.yml
codyoss: The latest staticcheck requires 1.22: https://github.com/dominikh/go-tools/blob/56172d41b117cc2c2f99f65fe0a790c8d7d7ea66/go.mod#L3

One thing we could consider is only running vet on latest version. We do this in other repos today.

internal/generated/snippets/go.mod
codyoss: good catch, I think I used my wrong tidy alias here. tidied again and gone.

shollyman (approved):

.github/workflows/vet.yml
quartzmo: Until now, I thought it was good to run vet on the oldest version in order to keep `toolchain` from sneaking into go.mod files in our repo. But maybe we no longer have any oldest version concerns like that.

quartzmo: > The latest staticcheck requires 1.22

See you in 2025, staticcheck!

quartzmo (approved):

PR #10691, opened 2024-08-15
chore: update readme about version support

codyoss:
Starting in 2025 these libraries will support just two versions of Go. This is a similar policy used by the Go programming language and many of our dependencies with the addition of GOTOOLCHAIN which was added in 1.21.

This policy has also been announced here: https://cloud.google.com/go/getting-started/supported-go-versions

Also unfreeze some deps now that we support 1.21.


quartzmo (approved):
Fantastic to see so many deps un-ignored in renovate!

PR #10694, opened 2024-08-15
[POC] Retry on read stall

raj-prince:
- Adding logic to retry in case of stall while making read request.
- Changes to retry in case of stall while reading from http response. (tritone's changes: [link](https://github.com/tritone/google-cloud-go/tree/read-stall-timeout))


cpriti-os:
This looks obsolete, closing this.

Issue #10696, opened 2024-08-16
auth: ENOTDIR not handled when user's HOME is not a directory


dfinkel:
## Client

trace, but really the `auth` package.

## Environment

Alpine Docker on GKE when running as the `guest` user which has a home directory set to `/dev/null`.

## Code and Dependencies

Mostly irrelevant, but  the code that initializes the opencensus exporter/client is:
<details>

```go
import (
	"fmt"

	"cloud.google.com/go/compute/metadata"
	"contrib.go.opencensus.io/exporter/stackdriver"
	"go.opencensus.io/trace"
)

// SetupTracingExporter adds stackdriver to the list of exporters that will receive
// traces. defaultProjID is the GCP project ID that should be used to identify
// the Stackdriver project. defaultTraceAttributes will be appended to every
// span that is exported to Stackdriver Trace
func SetupTracingExporter(defaultProjID string, defaultTraceAttributes map[string]interface{}) (*stackdriver.Exporter, error) {
	sd, err := stackdriver.NewExporter(stackdriver.Options{
		ProjectID:              defaultedGCPProject(defaultProjID),
		DefaultTraceAttributes: defaultTraceAttributes,
	})

	if err != nil {
		return nil, fmt.Errorf("Failed to get stack driver exporter: %w", err)
	}

	trace.RegisterExporter(sd)
	return sd, nil
}

// defaultedGCPProject is a helper function to automatically get the project ID
// when running on GCE. If it's not on GCE, it will default to the default Project
// ID passed to the function
func defaultedGCPProject(defaultProjID string) string {
	if !metadata.OnGCE() {
		return defaultProjID
	}

	projectName, projErr := metadata.ProjectID()
	if projErr != nil {
		return defaultProjID
	}
	return projectName
}

```

</details>

## Expected behavior

The trace client (or any other GCP client) initializes without any client certificate. (using the GKE/GCE metadata server oauth credentials)

## Actual behavior

Setting up the opencensus stackdriver trace exporter fails with this error:
```
open /dev/null/.config/gcloud/certificate_config.json: not a directory
```

[...]

PR #10697, opened 2024-08-16
fix(auth): handle ENOTDIR when opening cert config

dfinkel:
Some users explicitly have their home directories set to `/dev/null`. When opening a file that's "under that directory", instead of getting the normal `ENOENT`, one should expect to get `ENOTDIR`.

Unfortunately, due to the variety of cases under which `ENOTDIR` is returned, the go team has declined to include ENOTDIR in `ErrNotExist`. (https://golang.org/issues/18974)

The most robust solution in this case (following [1]) is to treat any
read error as `errSourceUnavailable`. This has the mixed-benefit of also
covering EPERM. (hopefully this doesn't make it too hard to track down
permissions problems for secure connect users)
    
   
Resolves: #10696

 [1]: https://go-review.googlesource.com/c/oauth2/+/493695

### DIFF
--- a/auth/internal/transport/cert/secureconnect_cert.go
+++ b/auth/internal/transport/cert/secureconnect_cert.go
@@ -62,11 +62,11 @@ func NewSecureConnectProvider(configFilePath string) (Provider, error) {
 
 	file, err := os.ReadFile(configFilePath)
 	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			// Config file missing means Secure Connect is not supported.
-			return nil, errSourceUnavailable
-		}
-		return nil, err
+		// Config file missing means Secure Connect is not supported.
+		// There are non-os.ErrNotExist errors that may be returned.
+		// (e.g. if the home directory is /dev/null, *nix systems will
+		// return ENOTDIR instead of ENOENT)
+		return nil, errSourceUnavailable
 	}
 
 	var metadata secureConnectMetadata
--- a/auth/internal/transport/cert/secureconnect_cert_test.go
+++ b/auth/internal/transport/cert/secureconnect_cert_test.go
@@ -17,11 +17,41 @@ package cert
 import (
 	"bytes"
 	"errors"
+	"os"
+	"path/filepath"
 	"testing"
 )
 
 func TestSecureConnectSource_ConfigMissing(t *testing.T) {
 	source, err := NewSecureConnectProvider("missing.json")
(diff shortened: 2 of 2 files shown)


codyoss (changes requested):
Please see feedback on the linked issue.

codyoss (approved):
Thank you for the PR and the test coverage 🎆

codyoss:
Failing test:

```
=== RUN   TestSecureConnectSource_ConfigMissingPerms
    secureconnect_cert_test.go:53: got cert: could not parse JSON in "/tmp/TestSecureConnectSource_ConfigMissingPerms1547290171/001/unreadable.json": unexpected end of JSON input, want certificate source is unavailable
--- FAIL: TestSecureConnectSource_ConfigMissingPerms (0.00s)
```

auth/internal/transport/cert/secureconnect_cert.go
```diff
@@ -62,11 +62,11 @@ func NewSecureConnectProvider(configFilePath string) (Provider, error) {
...
-		if errors.Is(err, os.ErrNotExist) {
-			// Config file missing means Secure Connect is not supported.
-			return nil, errSourceUnavailable
-		}
-		return nil, err
+		// Config file missing means Secure Connect is not supported.
+		// There are non-os.ErrNotExist errors that may be returned.
+		// (e.g. if the home directory is /dev/null, *nix systems will
+		// return ENOTDIR instead of ENOENT)
+		return nil, errSourceUnavailable
```
jba, line 69: This could potentially break users who are looking for some other error, like permission denied.
It would be better if we could detect ENOTDIR somehow.

codyoss: I think this is okay for that case to fall back. This path happens in defaulting and I think it may be better to work in this case rather than now -- this sentinel is handled in a calling layer. I do understand though that hiding the details may make getting this feature working harder to detect. I think once we have logging in place in the near future details like this should be expressed in that way

dfinkel:
Oh, right!
That test won't work when running as root. :facepalm:.

I'll update it to skip that test when the UID is `0`. (it seems a bit too complicated to check for `CAP_FOWNER`)

dfinkel:
Ok, I've rebased and pushed a guard to skip the EPERM test when the UID is 0.
(I hope these tests sometimes run as something other than root -- it passes on my machine when running as a non-root user)

codyoss (approved):

PR #10699, opened 2024-08-16
feat(vertexai): add tokenizer package

eliben:


vertexai/genai/tokenizer/tokenizer.go
```diff
@@ -0,0 +1,157 @@
...
+	encoder, err := sentencepiece.NewEncoder(bytes.NewReader(data))
+	if err != nil {
+		return nil, fmt.Errorf("creating encoder: %w", err)
+	}
+
+	return &Tokenizer{encoder: encoder}, nil
+}
+
+// CountTokens counts the tokens in all the given parts and returns their
+// sum. Only [genai.Text] parts are suppored; an error will be returned if
```
jba, line 75: Then why not make the argument of type ...genai.Text?

jba (changes requested):

vertexai/genai/tokenizer/tokenizer.go
eliben: I was thinking it would be awkward for this to have a different type from https://pkg.go.dev/cloud.google.com/go/vertexai/genai#GenerativeModel.CountTokens

Consider wanting to count tokens with both. One would likely create a `[]Content`, because of the invariance issue we can't pass `[]Text` to `GenerativeModel.CountTokens`. But then to call `Tokenizer.CountTokens` one would then have to convert `[]Content` to `[]Text` which is awkward.

Also if in the future `Tokenizer.CountTokens` supports other parts, changing it to `[]Content` will be a breaking change.

jba (approved):

PR #10710, opened 2024-08-20
feat(storage): add update time in bucketAttrs

### LINKED ISSUE #9361: storage: bucketAttrs class in bucket.go does not have Updated field for last modification time for bucket
The "bucketAttrs" class in the "bucket.go" file (https://github.com/googleapis/google-cloud-go/blob/9ce3d5b63d2c41119b64e691118303be4ae1b204/storage/bucket.go#L341) lacks an updated field to capture the last modification time for the bucket. I am currently developing a proxy service that redirects requests from "gxs-hadoop-connector" to a proxy for GCS object storage. The GCS-Hadoop-connector requires the Modified time for the bucket, and it encounters issues because the proxy server is unable to provide this modified time, resulting in crashes.

How can I get last modified time of a bucket  in go client SDK. 

**Additional context**

shubham-diwakar:
Add update time field in bucketAttrs struct.

Fixes #9361


storage/bucket.go
```diff
@@ -416,6 +416,10 @@ type BucketAttrs struct {
 	// This field is read-only.
 	Created time.Time
 
+	// Time at which the bucket was last modified.
```
tritone, line 419: Nit: godocs should start with the name of the field; ie `Updated is the time at which the bucket was last modified`. The linter will complain about this.

storage/bucket.go
```diff
@@ -922,6 +928,7 @@ func (b *BucketAttrs) toRawBucket() *raw.Bucket {
 		Name:                  b.Name,
 		Location:              b.Location,
 		StorageClass:          b.StorageClass,
+		Updated:               b.Updated.Format(time.RFC3339),
```
tritone, line 931: Since the field is read-only, it doesn't have to be included in `toRawBucket` or `toProtoBucket` I believe.

storage/bucket_test.go
```diff
@@ -157,6 +158,7 @@ func TestBucketAttrsToRawBucket(t *testing.T) {
 			},
 			PublicAccessPrevention: "enforced",
 		},
+		Updated:    time.Now().Format("2006-01-02T15:04:05Z07:00"),
```
tritone, line 161: Does this actually work? Wouldn't time.Now() be later when the comparison is made?

storage/integration_test.go
```diff
@@ -589,6 +589,9 @@ func TestIntegration_BucketUpdate(t *testing.T) {
 		if !testutil.Equal(attrs.Labels, wantLabels) {
 			t.Fatalf("add labels: got %v, want %v", attrs.Labels, wantLabels)
 		}
+		if !attrs.Created.Before(attrs.Updated) {
+			t.Fatal("attrs.Updated should be newer than attrs.Created")
```
tritone, line 593: Nit: These checks should be `Errorf` rather than `Fatalf` because they don't need to stop the execution of the remainder of the test.

Also, include both timestamps in the error for debugging purposes, something like:
`got attrs.Updated %v before attrs.Created %v, want Attrs.Updated to be after`

storage/bucket.go
shubham-diwakar: Done

storage/bucket.go
shubham-diwakar: Done

storage/bucket_test.go
shubham-diwakar: Given that this was BucketAttrs to RawBucket Test hence i have removed this.
Although a nice catch, but it was working somehow. Will check why.

storage/integration_test.go
shubham-diwakar: Have changed up.

Reason behind picking Fatalf was that Bucket Attributes should update given that this was not done hence i thought tests should stop. Also all other checks where fatalf only.

tritone (approved):

PR #10732, opened 2024-08-21
fix(auth): Setting expireEarly to default when the value is 0

bhshkh:
**Issue**:
While running the Bigtable client on GCE instance, token is not expired early. 

**Cause**:
computeTokenProvider sets ExpireEarly to EarlyTokenRefresh but there is no way to pass this option in the gtransport.DialPool method used by bigtable.go

https://github.com/googleapis/google-cloud-go/blob/1afb9ee1ee9de9810722800018133304a0ca34d1/auth/credentials/compute.go#L40-L45

**Fix**:
Use the default expiry early delta if ExpireEarly is 0 or ctpo is nil


codyoss (approved):
Thanks!

Issue #10731, opened 2024-08-21
storage: implement dataflux fast listing


akansha1812:
To list large dataset in a GCS bucket sequential it takes a long time. If we can list objects in parallel, it will be much faster to complete listing.

Dataflux fast-listing will be used to list objects in a bucket in parallel using worksteal algorithm. It supports storage.Query to filter objects in a bucket and returns objects in batches. User can provide bucket, storage.Query and number of parallel worker and batch size.  

There are different implementation for worksteal algorithm done and after benchmarking those, dataflux implementation came out faster.

codyoss:
I am not sure what data flux is, is the related to storage? All List RPC today are compliant with https://google.aip.dev/158 and https://google.aip.dev/client-libraries/4233. This is all based on page_tokens that you need to do a fetch to get the next result.

tritone:
@codyoss this will be a sub-package for storage similar to transfer manager, but focused on a few new features for AI/ML workloads.

PR #10748, opened 2024-08-22
feat(storage/dataflux): add dataflux interface

akansha1812:
feat: add dataflux interface and helper functions to storage/dataflux.  
Dataflux fast-listing will be used to quickly list objects in a bucket in parallel. 


storage/dataflux/worksteal.go
```diff
@@ -0,0 +1,54 @@
...
+	rangesplitter *rangeSplitter
+	idleChannel   chan int
+	result        *listerResult
+	generation    int64
+}
+
+// workstealListing is the main entry point of the worksteal algorithm.
+// It performs worksteal to achieve highly dynamic object listing.
+func workstealListing(ctx context.Context, opts Lister) ([]*storage.ObjectAttrs, error) {
+	return nil, nil
```
MattIrv, line 53: Should this and all the other functions that aren't implemented in this PR return a not implemented error for now?

storage/dataflux/doc.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2024 Google LLC
```
MattIrv, line 1: Should we add a README to this directory or something indicating that it's under active development?

MattIrv (approved):

storage/dataflux/fast_list.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Input contains options for listing objects.
+type Input struct {
+	// Client is the storage client to list objects from.
+	Client *storage.Client
+	// BucketName is the name of the bucket to list objects from.
+	BucketName string
+	// Parallelism is number of parallel workers to use for listing.
+	Parallelism int
+	// pageSize is the number of objects to list.
```
tritone, line 43: BatchSize, not pageSize.

storage/dataflux/fast_list.go
```diff
@@ -0,0 +1,95 @@
...
+	BatchSize int
+	// query is the query to filter objects for listing.
```
tritone, line 45: Capitalize Query to match the param name.

storage/dataflux/fast_list.go
```diff
@@ -0,0 +1,95 @@
...
+	// batchSize is the number of objects to list.
+	batchSize int
+	// query is the query to filter objects for listing.
+	query storage.Query
+	// skipDirectoryObjects is to indicate whether to list directory objects.
+	skipDirectoryObjects bool
+}
+
+// CloseFunc is the function to close the range channel of a Lister.
+type CloseFunc func()
```
tritone, line 73: I don't think we need this since the user can just call Lister.Close directly.

storage/dataflux/fast_list.go
```diff
@@ -0,0 +1,95 @@
...
+
+// NewLister creates a new dataflux Lister to list objects in the give bucket.
+func NewLister(opts *Input) (*Lister, CloseFunc) {
+
+	lister := &Lister{}
+	return lister, func() { lister.Close() }
+}
+
+// NextBatch runs worksteal algorithm and sequential listing in parallel to quickly
```
tritone, line 82: I think this isn't completely clear on how one would call this. Maybe it's helpful to add an `example_test.go` file like we have in storage/ and transfermanager in order to make an example for the docs?

tritone: Also is there a way for the user to choose which algorithm is used, or do you always run both?

storage/dataflux/range_splitter.go
```diff
@@ -0,0 +1,43 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dataflux
+
+import (
+	"sync"
+)
+
+// rangeSplitter specifies the a list and a map of sorted alphabets.
```
tritone, line 21: rangeSplitting is only applicable for work stealing right? If so let's just put it in that file to keep things together.

storage/dataflux/sequential.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"context"
+
+	"cloud.google.com/go/storage"
+)
+
+// Listing performs a sequential listing on the given bucket.
+// It returns a list of objects and the next token to use to continue listing.
+// If the next token is empty, then listing is complete.
+func Listing(ctx context.Context, opts Lister) ([]*storage.ObjectAttrs, string, error) {
```
tritone, line 26: Is this intended to be exported? I don't see anything similar for work stealing.

storage/dataflux/fast_list.go
```diff
@@ -0,0 +1,95 @@
...
+	// open when any method can be used to list.
+	open listingMethod = iota
+	// sequential when the listing is done sequentially.
+	sequential
+	// worksteal when the listing is done using work stealing algorithm.
+	worksteal
+)
+
+// Input contains options for listing objects.
+type Input struct {
```
tritone, line 36: Let's call this ListerInput because I assume you will need a separate one for composite downloads?

storage/dataflux/fast_list.go
```diff
@@ -0,0 +1,95 @@
...
+	// Client is the storage client to list objects from.
+	Client *storage.Client
+	// BucketName is the name of the bucket to list objects from.
```
tritone, line 39: Note which fields are required vs. optional.

tritone (commented):
Some initial comments. Overall I'd like to see an MVP that works end-to-end (maybe just sequential list?) before we merge, as well as an example_test.go file and basic integration test. That will help validate that this interface is workable before we release anything. Obviously we can make breaking changes in the subpackage since we have the preview tag, but it's not ideal.

storage/dataflux/fast_list.go
akansha1812: Done.

storage/dataflux/fast_list.go
akansha1812: Done

storage/dataflux/sequential.go
akansha1812: It should not be exported.
Updated it.

storage/dataflux/fast_list.go
akansha1812: Added example_test.go. 


Users cannot choose the algorithm. When NextBatch is called for the first time by a lister, both the algorithm runs. Whichever algorithm completes first will be used for subsequent runs.

storage/dataflux/fast_list.go
akansha1812: Done.

storage/dataflux/fast_list.go
akansha1812: Done.

storage/dataflux/fast_list.go
```diff
@@ -0,0 +1,95 @@
...
+	// sequential when the listing is done sequentially.
+	sequential
+	// worksteal when the listing is done using work stealing algorithm.
+	worksteal
+)
+
+// Input contains options for listing objects.
+type Input struct {
+	// Client is the storage client to list objects from.
+	Client *storage.Client
```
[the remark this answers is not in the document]
akansha1812: Done.

storage/dataflux/doc.go
akansha1812: Added a README.md  with a description, usage example and some benchmarks.

storage/dataflux/worksteal.go
akansha1812: Done.

storage/dataflux/fast_list.go
```diff
@@ -73,18 +86,68 @@ type Lister struct {
...
-	return nil, nil
+	// countError tracks the number of failed listing methods.
+	countError := 0
+	var results []*storage.ObjectAttrs
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	g, childCtx := errgroup.WithContext(ctx)
+
+	// Run worksteal listing when method is Open or WorkSteal.
+	// Run sequential listing when method is Open or Sequential.
```
tritone, line 115: So the idea is that method is open to start, and then will be either sequential or workSteal subsequently after the first NextBatch call? If so maybe note that and add TODOs.

storage/dataflux/fast_list.go
```diff
@@ -73,18 +86,68 @@ type Lister struct {
...
+// sequential listing is expected to be faster. For larger dataset,
 // worksteal listing is expected to be faster.
 func (c *Lister) NextBatch(ctx context.Context) ([]*storage.ObjectAttrs, error) {
-	return nil, nil
+	// countError tracks the number of failed listing methods.
+	countError := 0
+	var results []*storage.ObjectAttrs
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	g, childCtx := errgroup.WithContext(ctx)
```
tritone, line 112: I'm not sure errgroup is what you want here. IIUC this means if one listing method fails, then the other will be cancelled as well. In addition, you probably don't want to wait on results from both methods, but rather take the first to succeed. I think you probably need to manually spin up go routines for each method and then have a channel to send results on, then wait on that channel.

I think this is fine to leave in for the MVP since you are only supporting one listing method so far, but you should add TODOs to fix this.

storage/dataflux/fast_list.go
```diff
@@ -53,18 +59,25 @@ type Input struct {
 type Lister struct {
 	// method indicates the listing method(open, sequential, worksteal) to be used for listing.
 	method listingMethod
+
 	// pageToken is the token to use for sequential listing.
 	pageToken string
+
 	// ranges is the channel to store the start and end ranges to be listed by the workers in worksteal listing.
 	ranges chan *listRange
```
tritone, line 67: Let's leave this out for now since it is unused. Some of the code where you're looking at the length of the channel in fast_list.go looks questionable to me.

storage/dataflux/worksteal.go
```diff
@@ -0,0 +1,95 @@
...
+	// doneListing indicates whether the lister has completed listing.
+	doneListing bool
+	// nextStartRange is the start offset of the next page of objects to be listed.
+	nextStartRange string
+	// generation is the generation number of the last object in the page.
+	generation int64
+}
+
+// newObjectLister creates a new ObjectLister using the given lister options.
+func newObjectLister(ctx context.Context, opts newObjectListerOpts) *nextPageResult {
```
tritone, line 86: This seems to be unused currently? Should it be removed?

storage/dataflux/sequential.go
```diff
@@ -0,0 +1,86 @@
...
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, "", fmt.Errorf("iterating through objects %w", err)
+		}
+		if !(opts.skipDirectoryObjects && strings.HasSuffix(attrs.Name, "/")) {
+			objects = append(objects, attrs)
+		}
+		if defaultPageSize != 0 && (pageInfo.Remaining() == 0) {
```
tritone, line 80: Why break after 1 page here? the storage client object iterator can do pagination, so I would think sequential listing should just use one iterator until batchSize is filled up. Otherwise you're duplicating pagination logic from the client.

storage/dataflux/README.md
```diff
@@ -0,0 +1,60 @@
+# Dataflux for Google Cloud Storage Go client library
+
+## Overview
+The purpose of this client is to quickly list data stored in GCS. The core functionalities of this client can be broken down into two key parts.
```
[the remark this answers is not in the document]
akansha1812: Makes sense. Remove that line from the readme.

storage/dataflux/README.md
```diff
@@ -0,0 +1,60 @@
...
+        }
+    if err != nil {
+        log.Fatal(err)
+        }
+}
+```
+
+### Fast List Benchmark Results
+
+|File Count|VM Core Count|List Time Without Dataflux  |List Time With Dataflux|
```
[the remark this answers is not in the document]
akansha1812: Done

storage/dataflux/README.md
```diff
@@ -0,0 +1,60 @@
+# Dataflux for Google Cloud Storage Go client library
+
+## Overview
+The purpose of this client is to quickly list data stored in GCS. The core functionalities of this client can be broken down into two key parts.
+
+## Fast List
+The fast list component of this client leverages GCS API to parallelize the listing of files within a GCS bucket. It does this by implementing a workstealing algorithm, where each worker in the list operation is able to steal work from its siblings once it has finished all currently stated listing work. This parallelization leads to a significant real world speed increase than sequential listing. Note that paralellization is limited by the machine on which the client runs. Benchmarking has demonstrated that the larger the object count, the better Dataflux performs when compared to a linear listing.
```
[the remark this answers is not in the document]
akansha1812: Added.

storage/dataflux/README.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+[snip]:# (storage-2)
+```go
+
+// storage.Query to filter objects that the user wants to list.
+query := storage.Query{}
+// Input for fast-listing.
+dfopts := dataflux.ListerInput{
+    BucketName:		bucketName,
```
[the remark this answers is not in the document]
akansha1812: Done.

storage/dataflux/README.md
```diff
@@ -0,0 +1,60 @@
...
+    objects, err := df.NextBatch(ctx)
+
+    if err == iterator.Done {
+        // No more objects in the bucket to list.
+        break
+        }
+    if err != nil {
+        log.Fatal(err)
+        }
+}
```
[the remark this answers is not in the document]
akansha1812: Done

storage/dataflux/fast_list.go
akansha1812: Done. 
Added explanation and TODO.

storage/dataflux/fast_list.go
akansha1812: >>  you probably don't want to wait on results from both methods, but rather take the first to succeed. 

Errgroup takes care of running both methods in parallel. I am cancelling context which stops the running method as soon as one of them is complete(fail/succeed). So we don't wait for both the methods to complete.

>>  if one listing method fails, then the other will be cancelled as well.

It is true, as soon as one of the method fails, the other method will fail. 
If you see sequential listing it will only fail due to some error while iterating objects. Worksteal listing uses similar iterator so there is a good chance that worksteal will fail too. 
If worksteal fails, I think it would make sense to show the error instead of silently falling back to sequential listing as user would not be able to take advantage of fast-listing then.

storage/dataflux/fast_list.go
akansha1812: Removed the line for now.

Listing is paused when we list given batch size objects. ranges is used to continue listing from where it was left off when NextBatch is called again.
listRange stores start and end range of each worker in worksteal algorithm. When NextBatch is called again, listRange is used by all the workers to find start and end ranges. 

I am checking length of the channel to make sure all the ranges have been listed.

storage/dataflux/worksteal.go
akansha1812: Removed.

storage/dataflux/sequential.go
akansha1812: We'll need to use pageInfo for token for NextBatch. To track tokens correctly I am setting pageSize.

storage/dataflux/example_test.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+		if err != nil {
+			// handle error
+		}
+		numOfObjects += len(objects)
+	}
+	log.Printf("listing %d objects in bucket %q is complete.", numOfObjects, in.BucketName)
+}
+
+func ExampleNextBatch_all() {
```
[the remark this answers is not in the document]
akansha1812: It is to show that when batchSize is 0, NextBatch will return all the objects at once. 
I see why this could be confusing and should be called as part of for loop. 
Updated it.

storage/dataflux/fast_list.go
akansha1812: removed it from return value.
Thanks Chris!

storage/dataflux/fast_list.go
tritone: Ahh, thanks for the explanation. That makes sense. Maybe just note the thing about not falling back to the other method in a comment.

storage/dataflux/sequential.go
akansha1812: In sequential listing instead of one page, now listing batchSize number of objects.

storage/dataflux/integration_test.go
```diff
@@ -0,0 +1,263 @@
...
+		t.Fatalf("NewClient: %v", err)
+	}
+	in := &ListerInput{
+		BucketName: httpTestBucket.bucket,
+	}
+
+	df, close := NewLister(c, in)
+	defer close()
+
+	objects, err := df.NextBatch(ctx)
```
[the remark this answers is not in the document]
akansha1812: It is expected batchSize will be quite large and fast-list returns approximately (could be more with worksteal, could be less with sequential). Fast-list returns at least defaultPageSize(5000) number of objects.

I think using a fixture bucket with large number of objects would make more sense here.

storage/dataflux/fast_list.go
```diff
@@ -32,19 +36,21 @@ const (
...
-	// Client is the storage client to list objects from.
-	Client *storage.Client
-	// BucketName is the name of the bucket to list objects from.
+// ListerInput contains options for listing objects.
+type ListerInput struct {
+	// BucketName is the name of the bucket to list objects from. Required.
 	BucketName string
-	// Parallelism is number of parallel workers to use for listing.
+
+	// Parallelism is number of parallel workers to use for listing. Optional.
```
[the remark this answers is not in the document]
akansha1812: Done

storage/dataflux/integration_test.go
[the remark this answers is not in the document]
akansha1812: Using fixture bucket fetch objects in batches.

storage/dataflux/sequential.go
```diff
@@ -26,47 +25,51 @@ import (
...
 // sequentialListing performs a sequential listing on the given bucket.
 // It returns a list of objects and the next token to use to continue listing.
 // If the next token is empty, then listing is complete.
-func sequentialListing(ctx context.Context, opts Lister) ([]*storage.ObjectAttrs, string, error) {
+func (c *Lister) sequentialListing(ctx context.Context) ([]*storage.ObjectAttrs, string, error) {
 	var result []*storage.ObjectAttrs
+	var objectsListed int
+	var lastToken string
+	objectIterator := c.bucket.Objects(ctx, &c.query)
+	objectIterator.PageInfo().Token = c.pageToken
```
tritone, line 39: Does this approach with pageToken work if BatchSize is not a multiple of the page size? I am a bit concerned that this could lead to duplication of objects between batches. Could you double check this?

If so, I think it's fine to require that batch size be a multiple of page size (and round if not)

storage/dataflux/sequential.go
```diff
@@ -26,47 +25,51 @@ import (
 
 const (
 	// defaultPageSize specifies the number of object results to include on a single page.
-	defaultPageSize = 5000
+	defaultPageSize = 2000
```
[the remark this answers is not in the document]
akansha1812: Previously I had noticed that the tokens change every 1000 objects even when PageInfo.MaxSize was 5000, now looks like it is changing every 2000 objects.

storage/dataflux/sequential.go
akansha1812: In a given batch, we are returning approximately (few more) objects than in batch size.
While returning objects, we are rounding it up to a multiple of  page size. 
This will not result in duplication.

storage/dataflux/integration_test.go
```diff
@@ -85,14 +84,55 @@ func TestIntegration_NextBatch(t *testing.T) {
...
 	if len(objects) != len(httpTestBucket.objects) {
 		t.Errorf("expected to receive %d results, got %d results", len(httpTestBucket.objects), len(objects))
 	}
 }
 
+func TestIntegration_NextBatch(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Integration tests skipped in short mode")
+	}
+	const landsatBucket = "gcp-public-data-landsat"
```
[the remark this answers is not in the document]
akansha1812: Done.

storage/dataflux/sequential.go
```diff
@@ -26,47 +25,51 @@ import (
...
+		if nextToken == "" || (c.batchSize > 0 && objectsListed >= c.batchSize) {
 			break
 		}
-		i++
+		c.pageToken = nextToken
+	}
+	return result, lastToken, nil
+}
 
+func doListing(objectIterator *storage.ObjectIterator, result *[]*storage.ObjectAttrs, skipDirectoryObjects bool, objectsListed *int) (string, error) {
```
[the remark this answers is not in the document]
akansha1812: Done.

storage/dataflux/sequential.go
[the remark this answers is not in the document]
tritone: I think token can only change once per page, so I'm confused whether something is not being set correctly, or GCS is not sending as many results as we ask for, or what exactly. Happy to discuss offline if you'd like to look into debugging this.

storage/dataflux/sequential.go
tritone: Sounds good. Let's just add a note to the docs for BatchSize to indicate that this will be rounded.

storage/dataflux/worksteal.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"context"
+	"sync"
+
+	"cloud.google.com/go/storage"
+)
+
+// WorkerStatus indicates the status of a worker.
+type WorkerStatus int
```
tritone, line 25: Missed this earlier -- can you make both this and the const values unexported? It looks like they are not in any public surfaces currently.

tritone (approved):
Two more small comments, otherwise LGTM -- I will approve now and let you fix.

storage/dataflux/sequential.go
```diff
@@ -0,0 +1,78 @@
...
+		lastToken = nextToken
+		if nextToken == "" || (c.batchSize > 0 && objectsListed >= c.batchSize) {
+			break
+		}
+		c.pageToken = nextToken
+	}
+	return result, lastToken, nil
+}
+
+func doListing(objectIterator *storage.ObjectIterator, skipDirectoryObjects bool, objectsListed *int) ([]*storage.ObjectAttrs, string, error) {
```
tritone, line 57: objectsListed can also be moved to a returned value rather than something passed by ptr.

storage/dataflux/sequential.go
akansha1812: Added to the doc and input interface so the user is updated while using the API that the number of objects will be more and rounded up.

aiplatform/apiv1beta1/aiplatformpb/vertex_rag_data.pb.go
```diff
@@ -2010,162 +2010,227 @@ var file_google_cloud_aiplatform_v1beta1_vertex_rag_data_proto_rawDesc = []byte{
 	0x72, 0x70, 0x75, 0x73, 0x5f, 0x73, 0x74, 0x61, 0x74, 0x75, 0x73, 0x18, 0x08, 0x20, 0x01, 0x28,
 	0x0b, 0x32, 0x2d, 0x2e, 0x67, 0x6f, 0x6f, 0x67, 0x6c, 0x65, 0x2e, 0x63, 0x6c, 0x6f, 0x75, 0x64,
 	0x2e, 0x61, 0x69, 0x70, 0x6c, 0x61, 0x74, 0x66, 0x6f, 0x72, 0x6d, 0x2e, 0x76, 0x31, 0x62, 0x65,
```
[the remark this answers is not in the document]
akansha1812: Removed this.

storage/dataflux/sequential.go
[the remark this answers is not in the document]
akansha1812: I'll leave default page size to 5000 for now as it is considered efficient. 
2000 objects will be returned per page and batchsize will be rounded up to multiple of 2000. 

As discussed this not a blocker for now.

PR #10749, opened 2024-08-22
chore(datastore): update genproto version

codyoss:
This references the version of genproto that has the old library aliased to this one.


codyoss:
cc @bhshkh

quartzmo (approved):

PR #10751, opened 2024-08-23
docs(auth): fix Go doc comment link

julieqiu:
Fix link in Go doc comment per https://go.dev/doc/comment#links, so that it renders properly on pkgsite.


codyoss (approved):
Thank you!

PR #10757, opened 2024-08-23
feat(storage): add direct google access side-effect imports by default

frankyn:
Add side-effect modules for gRPC DirectPath by default with an opt-out build tag: `disable_grpc_modules`

- [x] Update Go storage documentation to reflect this change.
- [x] Include automatic opt-in for `GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS=true`


storage/doc.go
```diff
@@ -349,15 +349,9 @@ To create a client which will use gRPC, use the alternate constructor:
...
-set the environment variable `GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS=true` and add
-the following side-effect imports to your application:
-
-	import (
-		_ "google.golang.org/grpc/balancer/rls"
-		_ "google.golang.org/grpc/xds/googledirectpath"
-	)
+If the application is running within GCP, users may get better performance with Direct Google Access (enabling requests to skip some proxy steps).
+
+Dependencies for Direct Google Access can increase the size of resulting binary by ~16MiB if not used by another dependency and you can opt-out of these dependencies using build tag `disable_grpc_modules`.
```
tritone, line 354: This seems a little ambiguous as to who wants to use this build tag. I'd say something like:

> Dependencies for the gRPC API may slightly increase the size of binaries for applications depending on this package. If you are not using gRPC, you can use the build tag `disable_grpc_modules` to opt out of these dependencies and reduce the binary size.

tritone (approved):
Just some docs suggestions, otherwise LGTM. Nice solution!

danielduhh (approved):

PR #10759, opened 2024-08-23
feat(bigtable): Add UpdateFamily to allow updating a family type

ron-gal:
It is currently only possible to update a column GC policy. We want to be able to also update the value_type of a column family.

### DIFF
--- a/bigtable/admin.go
+++ b/bigtable/admin.go
@@ -671,13 +671,13 @@ type updateFamilyOption struct {
 	ignoreWarnings bool
 }
 
-// UpdateFamilyOption is the interface to update family settings
-type UpdateFamilyOption interface {
+// GCPolicyOption is deprecated, kept for backwards compatibility, use UpdateFamilyOption in new code
+type GCPolicyOption interface {
 	apply(s *updateFamilyOption)
 }
 
-// GCPolicyOption is deprecated, kept for backwards compatibility, use UpdateFamilyOption in new code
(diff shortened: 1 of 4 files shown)


bhshkh:
Please resolve vet failures

bigtable/admin.go
```diff
@@ -667,60 +667,89 @@ func (ac *AdminClient) TableInfo(ctx context.Context, table string) (*TableInfo,
 	return ti, nil
 }
 
-type gcPolicySettings struct {
+type updateFamilyOption struct {
 	ignoreWarnings bool
 }
 
-// GCPolicyOption is the interface to change GC policy settings
-type GCPolicyOption interface {
```
bhshkh, line 675: Ideally, we try to avoid already exported and released types. This would break customer's code

https://pkg.go.dev/cloud.google.com/go/bigtable#GCPolicyOption

But I would like to know what any of the admins think

@codyoss @quartzmo

ron-gal:
> Please add description to the PR

done

ron-gal:
> Please resolve vet failures

Done

bigtable/admin.go
codyoss: I agree. I would suggest just updating the documentation.

ron-gal: Done.

bhshkh (approved):

PR #10763, opened 2024-08-25
test(firestore): Make test data independent

bhshkh:
TestIntegration_QueryDocuments_WhereEntity uses the same indexes as TestIntegration_AggregationQueries.
If index deletion fails in 1st test, index creation fails in second test. Each test should be independent. So, the field names are being changed in TestIntegration_AggregationQueries test to prevent such failures


gkevinzheng:
Why does index creation fail in subsequent tests if it fails the first time?

bhshkh:
> Why does index creation fail in subsequent tests if it fails the first time?
There is some weird sync issue.
The subsequent tests try to create index again. That operation fails with error IndexAlreadyExists. I have tried changing the subsequent tests to ignore IndexAlreadyExists errors and continue with the test run but these fail too.
I think this is what is happening in that case.


1) Test1 calls Create Index 
2) Test1 Run and succeeds
3) Test1 calls Delete index
4) Test2 calls create index
5) Test2 sees IndexAlreadyExists error
6) Test2 run
7) Index gets deleted
8) Test2 fails mid run


Delete index is a sync operation. Even though test1 fails for DeleteIndex call to finish, indexes are not getting completely cleaned up until later

gkevinzheng (approved):

PR #10769, opened 2024-08-26
feat(auth): consider sharing credential in gRPC pool

### LINKED ISSUE #10631: auth: consider sharing credential in gRPC pool
Today any time a pool of connections are created we create a credential object for each connection dialed.

Consider the case where a user specifies a pool of 10 and is using ADC compute credentials. Currently every hour when a credential expires all 10 credential objects need to re-auth to the metadata service. If we instead share just the credential object every hour only 1 re-auth call would be made. O(n) -> O(1).

julieqiu:
Fixes https://github.com/googleapis/google-cloud-go/issues/10631

### DIFF
--- a/auth/grpctransport/grpctransport.go
+++ b/auth/grpctransport/grpctransport.go
@@ -189,35 +189,72 @@ func Dial(ctx context.Context, secure bool, opts *Options) (GRPCClientConnPool, 
 	if err := opts.validate(); err != nil {
 		return nil, err
 	}
+
+	req := &dialRequest{
+		opts:   opts,
+		secure: secure,
+	}
 	if opts.PoolSize <= 1 {
-		conn, err := dial(ctx, secure, opts)
+		resp, err := dial(ctx, req)
 		if err != nil {
 			return nil, err
 		}
-		return &singleConnPool{conn}, nil
+		return &singleConnPool{resp.conn}, nil
 	}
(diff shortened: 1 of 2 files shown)


auth/grpctransport/grpctransport.go
```diff
@@ -277,21 +314,29 @@ func dial(ctx context.Context, secure bool, opts *Options) (*grpc.ClientConn, er
...
 
-	return grpc.DialContext(ctx, endpoint, grpcOpts...)
+	conn, err := grpc.DialContext(ctx, endpoint, grpcOpts...)
+	if err != nil {
+		return nil, err
+	}
+	return &dialResponse{
+		conn:     conn,
+		grpcOpts: grpcOpts,
+		endpoint: endpoint,
```
quartzmo, line 338: If you mutate `req` here, I believe you could simply return `conn` and not need the `dialResponse` type. Whether you mutate `req` here or in the loop in `Dial`, it seems to me that the level of confusion is about equal, but that's just imho.

```go
		// Set the request fields to the final configuration to share gRPC
		// auth instance for performance optimization and avoiding token
		// refreshes where possible.
		// See https://github.com/googleapis/google-cloud-go/issues/10631 for details.
		req.endpoint = endpoint
		req.grpcOpts = grpcOpts
```

quartzmo (changes requested):

PR #10784, opened 2024-08-28
vertexai(test): add corpora_test in tokenizer module

happy-qiao:


vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+	"cloud.google.com/go/vertexai/genai"
+	"golang.org/x/text/encoding"
+	"golang.org/x/text/encoding/charmap"
+	"golang.org/x/text/encoding/japanese"
+	"golang.org/x/text/encoding/simplifiedchinese"
+	"golang.org/x/text/transform"
+)
+
+// FileInfo holds the name and content  of a file in the zip archive
```
eliben, line 36: CorporaInfo?

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,274 @@
...
+type CorporaInfo struct {
+	Name    string
+	Content []byte
+}
+
+// fileGenerator is a generator function that returns a channel to iterate over files in the zip archive
```
eliben, line 42: here too, make sure the doc comment aligns with the function/type it documents:

```
// FooBar is ....
func FooBar
```

etc.

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,274 @@
...
+	return nil, false
+}
+
+// ShouldSkip checks if the file should be skipped
+func (ucr *UdhrCorpusReader) ShouldSkip(filename string) bool {
+	return ucr.Skip[filename]
+}
+
+// DecodeBytes converts a byte slice from a specified encoding to a UTF-8 string
+func DecodeBytes(enc encoding.Encoding, data []byte) (string, error) {
```
eliben, line 188: in general, all the symbols in this file should be unexported (start with a lowercase letter)

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	return string(decodedBytes), nil
+}
+
+const defaultModel = "gemini-1.0-pro"
+const defaultLocation = "us-central1"
+
+func TestCountTokensWithCorpora(t *testing.T) {
+	// projectID := os.Getenv("VERTEX_PROJECT_ID")
```
eliben, line 211: When we run live tests, we `t.Skip` them if these env vars are not set. I see the code is commented out, but maybe it shouldn't be?

See https://github.com/googleapis/google-cloud-go/blob/main/vertexai/genai/client_test.go#L36

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,274 @@
...
+	Name    string
+	Content []byte
+}
+
+// fileGenerator is a generator function that returns a channel to iterate over files in the zip archive
+func CorporaGenerator(url string) (<-chan CorporaInfo, <-chan error) {
+	out := make(chan CorporaInfo)
+	errCh := make(chan error, 1)
+
+	go func() {
```
eliben, line 47: I'm not sure why you need to run this in a separate goroutine?

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// EncodingPattern struct to hold regex pattern and corresponding encoding
+type EncodingPattern struct {
+	Pattern  *regexp.Regexp
+	Encoding encoding.Encoding
+}
+
+// NewUdhrCorpusReader initializes a new UdhrCorpusReader with encoding patterns and skip set
+func NewUdhrCorpusReader() *UdhrCorpusReader {
```
eliben, line 117: Nit: `Reader` has a certain meaning in Go (https://pkg.go.dev/io#Reader), so this naming can be confusing. AFAICS this type isn't a reader, it just describes the corpus?

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,274 @@
...
+				fileReader.Close()
+			}
+		}
+	}()
+
+	return out, errCh
+}
+
+// UdhrCorpusReader struct to hold encoding patterns and skip set
+type UdhrCorpusReader struct {
```
eliben, line 105: For people not familiar with the corpus, I think it'd be useful to explain what's going on here with a large comment. How do the files in the corpus look, what is their naming pattern, etc? Otherwise the logic with the regexps is unclear

eliben (commented):
Thanks!

Some initial comments here; I'll do a more deep read-through after these are addressed

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: The goroutine uses channels (out and errCh) to communicate with the main thread. It sends downloaded file information through out and any errors through errCh. This allows for asynchronous processing and error handling.

For my education, what does `a separate goroutine` mean? There is already a go routine?

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Removed the Reader suffix

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Added comment in next commit

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Changed to lowercase in next commit

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Done

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,305 @@
...
+	if enc == encoding.Nop {
+		return string(data), nil
+	}
+
+	if enc == nil {
+		return "", fmt.Errorf("unsupported encoding or custom handling required")
+	}
+
+	decoder := enc.NewDecoder()
+	reader := transform.NewReader(strings.NewReader(string(data)), decoder)
```
eliben, line 230: Could you use https://pkg.go.dev/golang.org/x/text/transform#Bytes here instead of a reader?

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	ctx := context.Background()
+	client, err := genai.NewClient(ctx, projectID, defaultLocation)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer client.Close()
+	model := client.GenerativeModel(defaultModel)
+
+	t.Run("RemoteAndLocalCountTokensTest", func(t *testing.T) {
```
eliben, line 259: Since this `TestCount...` function only runs this one test, you don't need `t.Run(...` (this is needed when you have sub-tests).

You can just place the test logic directly in the `TestCount...` function itself
Like here, for example: https://github.com/googleapis/google-cloud-go/blob/main/vertexai/genai/client_test.go#L357

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,305 @@
...
+				fmt.Printf("No encoding found for file: %s\n", fileInfo.Name)
+				continue
+			}
+
+			decodedContent, err := decodeBytes(enc, fileInfo.Content)
+			if err != nil {
+				log.Fatalf("Failed to decode bytes: %v", err)
+			}
+
+			tok, err := New("gemini-1.5-flash")
```
eliben, line 282: This model name is different from `defaultModel` -- shouldn't they be the same?

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Done

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Done

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Done

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Done

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,305 @@
...
+// encodingPattern struct to hold regex pattern and corresponding encoding
+type encodingPattern struct {
+	Pattern  *regexp.Regexp
+	Encoding encoding.Encoding
+}
+
+// newUdhrCorpus initializes a new udhrCorpus with encoding patterns and skip set
+func newUdhrCorpus() *udhrCorpus {
+	encodings := []encodingPattern{
+		{Pattern: regexp.MustCompile(".*-Latin1$"), Encoding: charmap.ISO8859_1},
```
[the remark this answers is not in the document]
happy-qiao: Changed to map instead of regex

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,305 @@
...
+		{Pattern: regexp.MustCompile(".*-Greek$"), Encoding: charmap.ISO8859_7},
+		{Pattern: regexp.MustCompile(".*-UTF8$"), Encoding: encoding.Nop}, // No transformation needed
+		{Pattern: regexp.MustCompile("Amahuaca"), Encoding: charmap.ISO8859_1},
+		{Pattern: regexp.MustCompile("Turkish_Turkce-Turkish"), Encoding: charmap.ISO8859_9},
+		{Pattern: regexp.MustCompile("Lithuanian_Lietuviskai-Baltic"), Encoding: charmap.ISO8859_4},
+		{Pattern: regexp.MustCompile("Japanese_Nihongo-EUC"), Encoding: japanese.EUCJP},
+		{Pattern: regexp.MustCompile(`Abkhaz\-Cyrillic\+Abkh`), Encoding: charmap.Windows1251},
+	}
+
+	skip := map[string]bool{
```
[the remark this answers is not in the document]
happy-qiao: The skip is from https://github.com/nltk/nltk/blob/f6567388b4399000b9aa2a6b0db713bff3fe332a/nltk/corpus/reader/udhr.py#L14

I have tried my best to reduce the skip list but there are still some cases cannot be supported in Go

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Done. Removed the go routinue

vertexai/genai/tokenizer/corpora_test.go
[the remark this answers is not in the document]
eliben: Got it. Can you add a comment saying this? That the skip list comes from the NLTK source code which says these are unsupported encodings, or in general encodings Go doesn't support

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"golang.org/x/text/transform"
+)
+
+// corporaInfo holds the name and content of a file in the zip archive
+type corporaInfo struct {
+	Name    string
+	Content []byte
+}
+
+// corporaGenerator is a generator function that returns a channel to iterate over files in the zip archive
```
eliben, line 42: update this comment now that there're no channels

eliben (approved):
This is LGTM now, thanks for your patience! Just a couple of small comment tweaks requested

You'll also have to see why the kokoro tests are failing

internal/generated/snippets/go.mod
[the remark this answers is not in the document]
happy-qiao: Done

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Done

happy-qiao:
@eliben Thanks for reviewing. The testing workflows are green.

Issue #10795, opened 2024-08-29
auth: The quota project is not set correctly, breaking clients


adamroyjones:
## Client

Cloud Build, but I expect this generalises to almost everything.

## Environment

```
$ go version
go version go1.22.1 linux/amd64
```

## Code and Dependencies

The projects are fictive.

```go
package main

import (
	"context"
	"fmt"

	cloudbuildv2 "cloud.google.com/go/cloudbuild/apiv1/v2"
	"cloud.google.com/go/cloudbuild/apiv1/v2/cloudbuildpb"
	"google.golang.org/api/iterator"
	"google.golang.org/api/option"
)

func main() {
	if err := run(); err != nil {
		panic(err)
	}
}

func run() error {
	const project = "A"
	const region = "europe-west1"

	ctx := context.Background()

	cli, err := cloudbuildv2.NewClient(ctx, option.WithQuotaProject(project))
	if err != nil {
		return err
	}

	it := cli.ListBuildTriggers(ctx, &cloudbuildpb.ListBuildTriggersRequest{
		Parent:    "projects/" + project + "/locations/" + region,
		ProjectId: project,
	})

	for {
		_, err := it.Next()
		if err == iterator.Done {
			return nil
		}
		if err != nil {
			return fmt.Errorf("could not list triggers: %w", err)
		}
	}
}
```

<details>
  <summary>go.mod</summary>

```text
module reprex

go 1.22.1

require (
	cloud.google.com/go/cloudbuild v1.17.0
	google.golang.org/api v0.195.0
)

require (
	cloud.google.com/go v0.115.1 // indirect
	cloud.google.com/go/auth v0.9.1 // indirect
	cloud.google.com/go/auth/oauth2adapt v0.2.4 // indirect
	cloud.google.com/go/compute/metadata v0.5.0 // indirect
	cloud.google.com/go/longrunning v0.5.12 // indirect
	github.com/felixge/httpsnoop v1.0.4 // indirect
	github.com/go-logr/logr v1.4.2 // indirect
	github.com/go-logr/stdr v1.2.2 // indirect
	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
	github.com/google/s2a-go v0.1.8 // indirect
	github.com/googleapis/enterprise-certificate-proxy v0.3.2 // indirect
	github.com/googleapis/gax-go/v2 v2.13.0 // indirect
	go.opencensus.io v0.24.0 // indirect
[...]

PR #10797, opened 2024-08-29
fix(auth): make sure quota option takes precedence over env/file

codyoss:
Also align header keys set from old library. Example: https://github.com/googleapis/google-api-go-client/blob/21f10edb1d9d6a3d3baa5b1edcef6588da99def8/transport/grpc/dial.go#L212

Eventually we should consider pushing all quota project resolution down the the credentials package, but for now this will work.

Fixes: #10795


adamroyjones:
I can confirm that this fixes the issue I raised (with the reproducing example I have).

quartzmo:
How does this fix work? Why does the capitalization matter?

quartzmo:
The reason I'm asking is that [header field names MUST be converted to lowercase prior to their encoding in HTTP/2](https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2). But the changed header key in this PR retains the uppercase `X`.

codyoss:
@quartzmo The capitalization only really matters here for the gRPC case, but I changed both for consistency. It matters so that their is a collision in the map sent by the bridging code I linked to in the PR desc.

auth/grpctransport/grpctransport.go
```diff
@@ -38,7 +38,7 @@ const (
 	// Check env to decide if using google-c2p resolver for DirectPath traffic.
 	enableDirectPathXdsEnvVar = "GOOGLE_CLOUD_ENABLE_DIRECT_PATH_XDS"
 
-	quotaProjectHeaderKey = "X-Goog-User-Project"
+	quotaProjectHeaderKey = "X-goog-user-project"
```
quartzmo, line 43: So it's OK to leave `X` capitalized here in grpc metadata?

codyoss: Yes, because that is what it is here: https://github.com/googleapis/google-api-go-client/blob/21f10edb1d9d6a3d3baa5b1edcef6588da99def8/transport/grpc/dial.go#L212

quartzmo (approved):

PR #10798, opened 2024-08-30
feat(storage): read min throughput option

tritone:
Adds a RetryOption to set a minimum throughput per second on reads. If Read was called and throughput is below this level, then the connection will be closed and reopened.

This timeout will also be applied to gRPC reads in a future PR.

Supercedes #10683


storage/http_client.go
```diff
@@ -1468,18 +1477,64 @@ func parseReadResponse(res *http.Response, params *newRangeReaderParams, reopen
...
+			var seen int64
+			var done bool
+			for !done {
+				select {
+				case <-time.After(time.Second):
+					// Only close here if a Read call is in progress or if there have been calls
+					// to Read in the past second. This avoids needlessly closing the stream in
+					// the case where the application is not making calls to Read and so
+					// throughput is low for that reason.
+					if (httpReader.readCalls-readCalls > 1 || httpReader.readCallOpen) && httpReader.seen-seen < s.retry.minReadThroughput {
```
BrennaEpp, line 1518: `httpReader.readCalls` is not thread safe, do we need to add a mutex around it?

storage/http_client.go
```diff
@@ -1468,18 +1477,64 @@ func parseReadResponse(res *http.Response, params *newRangeReaderParams, reopen
...
+					done = true
+				}
+			}
+		}
+	} else {
+		monitorThroughput = func() {
+			select {
+			case <-httpReader.closed:
+			case <-r.ctx.Done():
+			}
```
BrennaEpp, line 1535: What do this do?

storage/http_client.go
tritone: Good point -- I did this. There might be a cleaner way to accomplish this with a channel however.

storage/http_client.go
tritone: Was intended to make sure the closed channel was not blocked, but I was able to remove this by adding a default case to the select block in httpReader.Close()

PR #10800, opened 2024-08-30
fix(pubsub): add attributes before startSpan

jameshartig:
Attributes can be used for sampling and so they should be added before span creation, if possible. See [1].

We had an configuration AST that lets us set the sample rate based on the span attributes but it requires the span attributes to be sent to the create span so they can be used for sampling. In particular we wanted to increase the sample rate for `code.function: receive` so we could help debug the recent redelivery issues.

[1] https://opentelemetry.io/docs/concepts/signals/traces/#attributes

/cc @hongalex


hongalex (approved):

hongalex:
Thanks for the issue and the corresponding PR! This will be part of the next release

Issue #10804, opened 2024-09-02
auth: 'GOOGLE_CLOUD_QUOTA_PROJECT' env variable should override quota project setting from credential file


michalskalski:
## Client

Client which works with new auth library

## Environment

% go version
go version go1.22.6 darwin/arm64

## Code and Dependencies

```go
package main

import (
	"context"
	"os"

	pubsub "cloud.google.com/go/pubsub/apiv1"
	"cloud.google.com/go/pubsub/apiv1/pubsubpb"

	"google.golang.org/api/iterator"
)

func main() {
	// enable new auth lib
	os.Setenv("GOOGLE_API_GO_EXPERIMENTAL_ENABLE_NEW_AUTH_LIB", "true")

	projectID := "my-gcp-project"
	// override quota project
	os.Setenv("GOOGLE_CLOUD_QUOTA_PROJECT", projectID)

	ctx := context.Background()

	pubsubClient, err := pubsub.NewPublisherClient(ctx)
	if err != nil {
		panic(err)
	}

	it := pubsubClient.ListTopics(ctx, &pubsubpb.ListTopicsRequest{
		Project: "projects/" + projectID,
	})

	for {
		resp, err := it.Next()
		if err == iterator.Done {
			break
		}
		if err != nil {
			panic(err)
		}
		println(resp.Name)
	}

}
```

<details>
  <summary>go.mod</summary>

```text
module github.com/michalskalski/gcp-quota-project

go 1.22.6

require (
	cloud.google.com/go/pubsub v1.42.0
	google.golang.org/api v0.195.0
)

require (
	cloud.google.com/go/auth v0.9.2 // indirect
	cloud.google.com/go/auth/oauth2adapt v0.2.4 // indirect
	cloud.google.com/go/compute/metadata v0.5.0 // indirect
	cloud.google.com/go/iam v1.1.13 // indirect
	github.com/felixge/httpsnoop v1.0.4 // indirect
	github.com/go-logr/logr v1.4.2 // indirect
	github.com/go-logr/stdr v1.2.2 // indirect
	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
	github.com/google/s2a-go v0.1.8 // indirect
	github.com/googleapis/enterprise-certificate-proxy v0.3.3 // indirect
	github.com/googleapis/gax-go/v2 v2.13.0 // indirect
	go.opencensus.io v0.24.0 // indirect
[...]

PR #10805, opened 2024-09-02
Adjust priority of quota project sources

michalskalski:
'GOOGLE_CLOUD_QUOTA_PROJECT' env variable should override quota project setting from credentials file. This is a behavior of previous auth code: https://github.com/googleapis/google-api-go-client/blob/main/internal/creds.go#L242-L261

It is trying to solve issue described in https://github.com/googleapis/google-cloud-go/issues/10804

To activate it would required additional change in https://github.com/googleapis/google-api-go-client: https://github.com/googleapis/google-api-go-client/pull/2767

### DIFF
--- a/auth/credentials/detect.go
+++ b/auth/credentials/detect.go
@@ -158,6 +158,9 @@ type DetectOptions struct {
 	// The default value is "googleapis.com". This option is ignored for
 	// authentication flows that do not support universe domain. Optional.
 	UniverseDomain string
+	// QuotaProjectID allow overriding the project ID used for quota management.
+	// Optional.
+	QuotaProjectID string
 }
 
 func (o *DetectOptions) validate() error {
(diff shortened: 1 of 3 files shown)


codyoss:
Please see comments on the associated issue. I would like something like these changes in the library. There is just some other associated changes that would need to happen at the same time. If you would like to work on this I am happy to explain more in the new issue I created for this request, but if not that is fine to and we will get to it in the not too distant future. Thanks again.

michalskalski:
@codyoss if you have some general idea how you would like to see this being solved in long term, please share it in the issue.  I will try find time to implement. If you already know specifically how this should be implemented and it just matter of finding time, then I don't mind wait for your implementation. What is currently in repo already fixed my initial problem

codyoss:
@michalskalski We will get to it 😄 Just wanted to make sure I was not stealing an oppertunity away if you wanted to make a contribution. Thanks again for the report and glad the PR fixed the issue for you.

PR #10810, opened 2024-09-03
feat(auth): send auth metrics in token requests

quartzmo:


auth/internal/version.go
```diff
@@ -0,0 +1,18 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+// Version is the current tagged release of the library.
+const Version = "0.9.1"
```
codyoss: This value needs to be updated.

auth/threelegged_test.go
```diff
@@ -72,6 +74,10 @@ func Test3LO_StandardExchange(t *testing.T) {
 		if headerContentType != "application/x-www-form-urlencoded" {
 			t.Errorf("Unexpected Content-Type header %q", headerContentType)
 		}
+		headerApiClient := r.Header[http.CanonicalHeaderKey(header.GOOGLE_API_CLIENT_HEADER)][0]
```
codyoss: nit: user the getter

auth/credentials/internal/externalaccount/externalaccount.go
```diff
@@ -399,7 +400,7 @@ func newSubjectTokenProvider(o *Options) (subjectTokenProvider, error) {
 
 func getGoogHeaderValue(conf *Options, p subjectTokenProvider) string {
 	return fmt.Sprintf("gl-go/%s auth/%s google-byoid-sdk source/%s sa-impersonation/%t config-lifetime/%t",
-		goVersion(),
+		header.GoVersion(),
```
codyoss, line 403: the unknown value below this can be known now.

auth/credentials/internal/externalaccount/externalaccount.go
```diff
@@ -287,9 +288,9 @@ func (tp *tokenProvider) Token(ctx context.Context) (*auth.Token, error) {
 		SubjectToken:       subjectToken,
 		SubjectTokenType:   tp.opts.SubjectTokenType,
 	}
-	header := make(http.Header)
-	header.Set("Content-Type", "application/x-www-form-urlencoded")
-	header.Add("x-goog-api-client", getGoogHeaderValue(tp.opts, tp.stp))
+	hdr := make(http.Header)
+	hdr.Set("Content-Type", "application/x-www-form-urlencoded")
+	hdr.Add(header.GOOGLE_API_CLIENT_HEADER, getGoogHeaderValue(tp.opts, tp.stp))
```
codyoss, line 293: Do we need to send the new values here too?

auth/credentials/impersonate/user.go
```diff
@@ -153,6 +154,7 @@ func (u userTokenProvider) exchangeToken(ctx context.Context, signedJWT string)
 	if err != nil {
 		return nil, err
 	}
+	req.Header.Set(header.GOOGLE_API_CLIENT_HEADER, header.GetGoogHeaderToken(header.CredTypeImp, header.TokenTypeAccess))
```
codyoss, line 157: for normal user you had the type as none, but for an impersonated user it is access. Is that correct?

auth/credentials/idtoken/compute_test.go
```diff
@@ -26,6 +26,11 @@ const metadataHostEnv = "GCE_METADATA_HOST"
 
 func TestComputeCredentials(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// TODO(quartzmo): un-comment after compute/metadata packages sends x-goog-api-client
```
codyoss, line 29: want to file a tracking bug for this? Or are you working on this right away?

auth/credentials/idtoken/compute.go
```diff
@@ -65,6 +68,7 @@ func (c computeIDTokenProvider) Token(ctx context.Context) (*auth.Token, error)
 	if c.format == ComputeTokenFormatFullWithLicense {
 		v.Set("licenses", "TRUE")
 	}
+	callctx.SetHeaders(ctx, header.GOOGLE_API_CLIENT_HEADER, header.GetGoogHeaderToken(header.CredTypeMDS, header.TokenTypeID))
```
codyoss, line 71: Should this be commented out as well for now? Also, this returns a context who has that values set -- make sure to reassign.

codyoss: Happens elsewhere too.

codyoss (changes requested):

auth/credentials/idtoken/compute.go
quartzmo: Fixed, thanks!

auth/credentials/impersonate/user.go
quartzmo: @arithmetic1728 Should this [impersonate/user] flow be considered cred-type `u` or `imp`? I couldn't find a similar flow in googleapis/google-auth-library-python. Please link if you know of one.

auth/credentials/internal/externalaccount/externalaccount.go
quartzmo: @arithmetic1728 Should this external account STS exchange be given the headers, and if so, which values? I couldn't find a similar flow in googleapis/google-auth-library-python. Please link if you know of one.

auth/threelegged_test.go
quartzmo: Fixed, thanks!

auth/credentials/internal/externalaccount/externalaccount.go
quartzmo: Fixed, thanks!

auth/credentials/idtoken/compute_test.go
quartzmo: I'm going to make the change to `compute/metadata` in a follow-up PR right away.

auth/credentials/impersonate/user.go
sai-sunder-s: https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/impersonated_credentials.py#L425 is the python implementation for impersonation. I see `cred-type/imp` is used.

The token type needs to be `access` or `id` based on what token is being fetched/used. For normal user it could have been none because both the tokens are fetched at same time.

auth/credentials/internal/externalaccount/externalaccount.go
sai-sunder-s: I see that @aeitzman added metrics support for BYOID. Alex, is there some doc that details what headers to send?

https://github.com/googleapis/google-auth-library-python/pull/1332/files

sai-sunder-s: Ok looks like the code for sending x-goog-api-client headers for BYOID scenarios is already in there.
Sijun's design doesn't cover BYOID because Alex handled that separately. So no need to add the new values for BYOID files.

Issue #10684, opened 2024-08-14
datastore: Key.String() includes namespace


sgp:
**Client**

Datastore >= v1.14.0

**Environment**

Any

**Go Environment**

Any (but tested on `go version go1.21.11 darwin/arm64`)

**Code**

```go
package main

import (
	"cloud.google.com/go/datastore"
	"fmt"
)

func main() {
	k := datastore.NameKey("Foo", "Bar", nil)
	fmt.Printf("Key (no namespace)\t%s\n", k)

	k.Namespace = "Sensitive"
	fmt.Printf("Key (namespaced)\t%s\n", k)
}
```

**Expected behavior**

Calling `String()` on a `*datastore.Key` functions the way it always did, keeping namespace opaque.

```
Key (no namespace)      /Foo,Bar
Key (namespaced)        /Foo,Bar
```

**Actual behavior**

Namespace is leaked when calling `String()`.

```
Key (no namespace)      /Foo,Bar
Key (namespaced)        /Foo,Bar,Sensitive
```
 
**Additional context**

I understand that the change was made in #8363 to fix a real bug. However, the fix changed a stable API to achieve its goals, making it difficult to consume the latest library if there are tests that check for logs or other things containing the string value of a key. Furthermore, as we use namespaces for multi-tenancy, it's important that we don't accidentally leak namespace values outside of logs (via API or some other means). While it's possible to audit our code to _never_ use a `String()` representation of the key, or to ever present keys at all, it feels like the best course of action is to not include the namespace in the string representation at all.

Issue #7829, opened 2023-04-25
datastore: equal keys in different namespaces are deduped incorrectly


maros7:
**Client**

Datastore

**Environment**

Any...

**Go Environment**

$ go version
`go version go1.20.3 darwin/arm64`
$ go env

**Code**

https://github.com/googleapis/google-cloud-go/blob/e41aae26413fc0a31e2343c019b5c29a31efabc1/datastore/datastore.go#L459-L463

Let's say I have...
- Entity 1: `Kind: A, Name: B, Namespace: NS1`
- Entity 2: `Kind: A, Name: B, Namespace: NS2`

...then the second one will be deduped in the code above since namespace is not taken into consideration. Meaning that `pbKeys` is not complete.

**Expected behavior**

Both keys should be added to `pbKeys`.

**Actual behavior**

![image](https://user-images.githubusercontent.com/25764930/234247182-e8d6e390-470b-4b19-978f-fbf713ca0335.png)

![image](https://user-images.githubusercontent.com/25764930/234247592-951d97fc-6aae-4cfc-baf6-7d31343fdd50.png)

As one can see above. Index 1 and 2 in `keys` are just index 1 in `pbKeys` due to the fact that they share the same string representation.

**Additional context**

N/A

PR #10685, opened 2024-08-14
fix(datastore): remove namespace from Key.String() (#10684)

sgp:
Datastore namespaces may be sensitive, and it's best not to emit them.

Provide a `datastore.Key.StringWithNamespace()` as an alternative, and use this version internally, which preserves the fix from https://github.com/googleapis/google-cloud-go/issues/7829.

PR #10823, opened 2024-09-05
fix(datastore): remove namespace from Key.String() (#10684)

sgp:
Datastore namespaces may be sensitive, and it's best not to emit them. See #10684.

Provide a `datastore.Key.StringWithNamespace()` as an alternative, and use this version internally, which preserves the fix from https://github.com/googleapis/google-cloud-go/issues/7829.


sgp:
@bhshkh This is effectively the same as #10685 with the fixes you asked for already squashed. I had to reopen this PR on my on personal fork so that I could give Google permission to update the base branch, etc.

bhshkh (approved):

Issue #10793, opened 2024-08-29
datastore: RunInTransaction will backoff / sleep even if it's the last attempt


sgp:
## Client

Cloud Datastore

## Environment

All Environments

## Code and Dependencies

```go
package main

import (
	"context"
	"log"
	"sync"
	"time"

	"cloud.google.com/go/datastore"
)

var k = datastore.IDKey("TestFoo", 1, nil)

func updateIt(ctx context.Context, c *datastore.Client, i int) {
	toCtx, cancelFn := context.WithTimeout(ctx, 10*time.Second)
	defer cancelFn()
	start := time.Now()
	if _, err := c.RunInTransaction(toCtx, func(tx *datastore.Transaction) error {
		var e entity
		if err := tx.Get(k, &e); err != nil {
			return err
		}
		if e.Thing2 == "modified" {
			return nil
		}
		e.Thing2 = "modified"
		if _, err := tx.Put(k, &e); err != nil {
			return err
		}
		return nil
	}, datastore.MaxAttempts(1)); err != nil {
		log.Printf("[%03d %s] Error updating entity: %v", i, time.Since(start), err)
	} else {
		log.Printf("[%03d %s] Updated entity", i, time.Since(start))
	}
}

type entity struct {
	Thing1 string
	Thing2 string
}

func main() {
	ctx := context.Background()
	c, err := datastore.NewClient(ctx, "pendo-ionchef")
	if err != nil {
		log.Fatalf("datastore.NewClient: %v", err)
	}

	e := entity{
		Thing1: "One Fish",
		Thing2: "Two Fish",
	}

	if _, err := c.Put(ctx, k, &e); err != nil {
		log.Fatal(err)
	}

	var wg sync.WaitGroup
	for i := 0; i < 200; i++ {
		wg.Add(1)
		i := i
		go func() {
			defer wg.Done()
			updateIt(ctx, c, i)
		}()
	}

	wg.Wait()

	log.Println("Done.")
}
```

<details>
  <summary>go.mod</summary>

```text
module scratch

go 1.21

require cloud.google.com/go/datastore v1.17.1

require (
	cloud.google.com/go v0.114.0 // indirect
	cloud.google.com/go/auth v0.5.1 // indirect
	cloud.google.com/go/auth/oauth2adapt v0.2.2 // indirect
	cloud.google.com/go/compute/metadata v0.3.0 // indirect
	github.com/felixge/httpsnoop v1.0.4 // indirect
[...]

PR #10794, opened 2024-08-29
fix(datastore): do not delay on final transaction attempt

sgp:
Fix for #10793.

PR #10824, opened 2024-09-05
fix(datastore): do not delay on final transaction attempt

sgp:
Fix for #10793.


sgp:
@bhshkh This is the same as pull request #10794, but re-opened against a personal fork in order to give maintainers the ability to update the remote.

bhshkh (approved):

PR #10838, opened 2024-09-10
chore(storage): adding dynamic delay algorithm

raj-prince:
Adding go-routine safe dynamic delay algorithm that calculates the delay at a fixed percentile, based on delay samples.


storage/util/dynamic_percentile.go
```diff
@@ -0,0 +1,157 @@
+// Copyright 2024 Google LLC
```
tritone, line 1: Let's 1. move all this code to files in the top level storage package and 2. make all types unexported. I don't think it's necessary to add a separate utils package just for this, and I don't want external users depending on this code.

storage/util/dynamic_percentile.go
```diff
@@ -0,0 +1,157 @@
...
+package util
+
+import (
+	"fmt"
+	"math"
+	"sync"
+)
+
+// MetricType is the metric type DynamicPercentile algo expects.
+type MetricType int64
```
tritone, line 24: Do we need an aliased type for this? Seems like just leaving all values as an int64 would be fine. Also, what type does the internal version of this use for the values? Wondering if we might have issues with precision depending on what values we choose.

storage/util/dynamic_percentile.go
```diff
@@ -0,0 +1,157 @@
...
+
+// DynamicPercentile dynamically calculates the delay at a fixed percentile, based on
```
tritone, line 26: Nit: wouldn't call it "delay" if the value is not typed as time.

storage/util/dynamic_percentile.go
```diff
@@ -0,0 +1,157 @@
...
+func (dp *DynamicPercentile) Increase() {
+	dp.mu.Lock()
+	defer dp.mu.Unlock()
+
+	dp.increase()
+}
+
+func (dp *DynamicPercentile) decrease() {
+	v := MetricType(float64(dp.value) * dp.decreaseFactor)
+	if v < dp.min {
```
tritone, line 111: Perhaps this is trivial, but I was originally thinking that we would have the dynamic algorithm always return the true value for P99, and then the client retry logic would enforce min/max. Thoughts on this? Does the original version of this look the same?

storage/util/dynamic_percentile.go
tritone: Also, we could just leave this as a time.Duration and I could pipe through throughput as "the avg duration to transfer 1 MiB" for example (or 1 GiB).

tritone (commented):
Overall looks pretty good, I think the main change would be to move the pkg location and remove any exported types.

storage/util/dynamic_percentile.go
raj-prince: Yes, this is same as original version, where algorithm also takes min/max as input and handle it internally.

storage/util/dynamic_percentile.go
raj-prince: Internal version uses time.Duration only. 

Sure, reverting this to duration only. 
`the avg duration to transfer 1 MiB" for example (or 1 GiB).` - This will also work, the only thing we need to handle the stalled stream (0 byte transfer).

storage/util/dynamic_percentile.go
raj-prince: Not required, as changed the type.

tritone (approved):
LGTM, thanks Prince.

PR #10840, opened 2024-09-10
fix(compute/metadata): check error chain for retryable error

codyoss:
Thanks to @ianlancetaylor for pointing this out!


quartzmo (approved):
Good catch!

PR #10841, opened 2024-09-10
vertexai(test): Run corpora test in go coroutine to reduce test runtime

happy-qiao:


vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -241,47 +242,66 @@ func TestCountTokensWithCorpora(t *testing.T) {
...
 
-		enc, found := ucr.getEncoding(fileInfo.Name)
-		if !found {
-			fmt.Printf("No encoding found for file: %s\n", fileInfo.Name)
-			continue
-		}
+	worker := func() {
+		for corpora := range corporaChan {
+			if ucr.shouldSkip(corpora.Name) {
+				fmt.Printf("Skipping file: %s\n", corpora.Name)
```
eliben, line 263: `fmt` is not usually concurrency-safe (nothing bad will happen but output can get mixed up between goroutines).

Use `log.Printf` inside goroutines instead

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -241,47 +242,66 @@ func TestCountTokensWithCorpora(t *testing.T) {
...
+			}
+			remoteNtoks, err := model.CountTokens(ctx, genai.Text(decodedContent))
+			if err != nil {
+				log.Fatal(corpora.Name, err)
+			}
+			if localNtoks.TotalTokens != remoteNtoks.TotalTokens {
+				t.Errorf("expected %d(remote count-token results), but got %d(local count-token results)", remoteNtoks, localNtoks)
+			}
 		}
+		doneChan <- true
```
eliben, line 290: I think this code will work fine, but since your workers don't actually return any results, I recommend a slightly different pattern, using a `WorkGroup`.

This example shows how: https://gobyexample.com/waitgroups

Then you won't need the loop clearing the `doneChan` in the main goroutine; you'll just do a wait on the group.

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Done

vertexai/genai/tokenizer/corpora_test.go
happy-qiao: Done

vertexai/genai/tokenizer/corpora_test.go
```diff
@@ -241,47 +243,55 @@ func TestCountTokensWithCorpora(t *testing.T) {
...
-		if !found {
-			fmt.Printf("No encoding found for file: %s\n", fileInfo.Name)
-			continue
-		}
-
-		decodedContent, err := decodeBytes(enc, fileInfo.Content)
-		if err != nil {
-			log.Fatalf("Failed to decode bytes: %v", err)
-		}
+	workerCountChan := make(chan int, 10)
```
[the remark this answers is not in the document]
happy-qiao: I use this channel to manage up to 10 corpora runs simultaneously, preventing them from all launching at once.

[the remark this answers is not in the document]
eliben: I see, thanks for clarifying.

In this case the naming or a comment should be used to make this clearer. Also, for channels like this it's customary to have them carry no information, e.g. `chan struct{}` -- this further clarifies that the contents of the channel have no meaning.

The channel can be named something like `workLimiter` or `concurrencyLimiter`, or if you prefer to keep your channel name, just add a comment explaining what it's for.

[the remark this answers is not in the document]
happy-qiao: Done. Thanks for helping me improve it!

eliben (approved):

PR #10864, opened 2024-09-13
feat(compute/metadata): add logging

codyoss:
In order to add this correctly the default metadataClient needs to be initalized lazily. This is because calling clog.New has some logic that freezes logging configuration, so we need to ensure the user has a chance to set this config before global variables from a package are loaded.


codyoss:
This requires https://github.com/googleapis/gax-go/pull/372

PR #10870, opened 2024-09-14
fix(storage): disable grpc metrics using emulator

frankyn:


storage/grpc_client.go
```diff
@@ -95,6 +95,7 @@ func defaultGRPCOptions() []option.ClientOption {
 			option.WithEndpoint(host),
 			option.WithGRPCDialOption(grpc.WithInsecure()),
 			option.WithoutAuthentication(),
+			WithDisabledClientMetrics(),
```
BrennaEpp, line 98: Do we need to do the same for HTTP?

BrennaEpp (approved):

storage/grpc_client.go
frankyn: Not yet, metrics are only available in gRPC.

PR #10875, opened 2024-09-16
chore(storage): remove NotificationConfig implementation from gRPC

frankyn:


storage/grpc_client.go
```diff
@@ -1445,73 +1445,18 @@ func (c *grpcStorageClient) DeleteHMACKey(ctx context.Context, project string, a
...
-			}
-			req.PageToken = nextPageToken
-		}
-	}, s.retry, s.idempotent)
-	if err != nil {
-		return nil, err
-	}
-
-	return notificationsToMapFromProto(notifications), nil
+	return nil, fmt.Errorf("ListNotifications is not implemented")
```
tritone, line 1451: Use this error: https://github.com/googleapis/google-cloud-go/blob/4265accaab624984d7ea60025e24198a0a0db6d3/storage/storage.go#L69

storage/notifications.go
```diff
@@ -104,18 +104,6 @@ func toNotificationFromProto(pbn *storagepb.NotificationConfig) *Notification {
 	return n
 }
```
tritone, line 106: Add notes to the exported methods in this file indicating that gRPC is not supported.

tritone (approved):

PR #10888, opened 2024-09-19
perf(storage): gRPC zerocopy codec

tritone:
Replace the current custom codec for ReadObjectResponse with a CodecV2 that can handle data split across multiple buffers.

Unit tests pass but need to finish up some stuff to get the end-to-end working.


storage/grpc_client.go
```diff
@@ -1641,7 +1724,8 @@ func (r *gRPCReader) WriteTo(w io.Writer) (int64, error) {
 		// Attempt to receive the next message on the stream.
 		// Will terminate with io.EOF once data has all come through.
 		// recv() handles stream reopening and retry logic so no need for retries here.
-		msg, err := r.recv()
+		err := r.recv()
+		var msg []byte
```
frankyn, line 1728: How does msg get filled?

storage/grpc_client.go
```diff
@@ -1721,21 +1807,262 @@ const (
...
 	}
-	content, err := readProtoBytes(checksummedData, checksummedDataContentField)
+
+	return nil
+}
+
+// Consume a bytes field from the input. Returns offsets for the data in the buffer slices
+// and an error.
+func (d *readResponseDecoder) consumeBytes() (*bufferSliceOffsets, error) {
+	b := (*d.databufs)[d.currBuf].ReadOnlyData()
```
frankyn, line 1972: nit, this function only seems to require  `(*d.databufs)[d.currBuf].Len()`.

storage/grpc_client_test.go
```diff
@@ -85,8 +86,9 @@ func TestBytesCodec(t *testing.T) {
 	}
 
 	for _, test := range []struct {
-		desc string
-		resp *storagepb.ReadObjectResponse
+		desc    string
+		resp    *storagepb.ReadObjectResponse
+		offsets *bufferSliceOffsets
```
BrennaEpp, line 91: nit: wantOffsets

storage/grpc_client_test.go
```diff
@@ -23,11 +23,12 @@ import (
 
 	"cloud.google.com/go/storage/internal/apiv2/storagepb"
 	"github.com/google/go-cmp/cmp"
+	"google.golang.org/grpc/mem"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/testing/protocmp"
 )
 
-func TestBytesCodec(t *testing.T) {
+func TestBytesCodecV2(t *testing.T) {
```
BrennaEpp, line 31: We should also test with random splits in the encoded response.

BrennaEpp (commented):
Couple initial comments, looks like there is some commented out code as well.

storage/grpc_client_test.go
```diff
@@ -128,39 +128,67 @@ func TestBytesCodecV2(t *testing.T) {
...
+				},
+				{
+					desc:        "random split buffer",
+					splitPoints: []int{100, 500, 700},
+				},
+				{
+					desc:        "mid-tag split buffer",
+					splitPoints: []int{3, 1032},
+				},
+			} {
```
BrennaEpp, line 147: I think we should include more tests here, including an empty metadata struct (above). Maybe we could do something like fuzz tests? When I was testing I caught different errors on different split points (mid-tag on different tags, I think maybe end or start of a tag/data as well).

storage/grpc_client.go
```diff
@@ -1125,9 +1174,9 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
 			leftovers: msg.GetChecksummedData().GetContent(),
 			settings:  s,
 			zeroRange: params.length == 0,
-			databuf:   databuf,
-			wantCRC:   wantCRC,
-			checkCRC:  checkCRC,
+			// databuf:   databuf,
```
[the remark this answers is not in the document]
tritone: replaced this with the decoder.

storage/grpc_client_test.go
tritone: done, but I also think I'll add some fuzz testing later.

storage/grpc_client.go
tritone: irrelevant with the latest changes

storage/grpc_client.go
```diff
@@ -1052,7 +1060,8 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
 
 			// Receive the message into databuf as a wire-encoded message so we can
 			// use a custom decoder to avoid an extra copy at the protobuf layer.
-			err := stream.RecvMsg(&databuf)
+			databuf := &mem.BufferSlice{}
+			err := stream.RecvMsg(databuf)
```
neild, line 1064: A mem.BufferSlice is a []Buffer. In general, it's not necessary to pass around pointers to slices. I'd change this to:

```
var databuf mem.BufferSlice
err := stream.RecvMsg(&databuf)
```

...and then also change readResponseDecoder.databufs to be a non-pointer BufferSlice. The field doesn't need to be pointer valued, and making it a plain BufferSlice will slightly simplify a bit of the readResponseDecoder code below.

(RecvMsg gets passed &databuf, because in this case we're using it as an out parameter.)

storage/grpc_client.go
```diff
@@ -1721,21 +1728,323 @@ const (
...
+// data to the user application.
+type readResponseDecoder struct {
+	databufs *mem.BufferSlice // raw bytes of the message being processed
+	// Decoding offsets
+	off     uint64 // offset in the messsage relative to the data as a whole
+	currBuf int    // index of the current buffer being processed
+	currOff uint64 // offset in the current buffer
+	// Processed data
+	msg         *storagepb.ReadObjectResponse // processed response message with all fields other than object data populated
+	dataOffsets *bufferSliceOffsets           // offsets of the object data in the message.
```
neild, line 1742: Should dataOffsets be a pointer, or just a plain value? I think you might save an allocation by consistently passing around a non-pointer dataOffsets.

storage/grpc_client.go
```diff
@@ -1721,21 +1728,323 @@ const (
...
+	done        bool                          // true if the data has been completely read.
+}
+
+type bufferSliceOffsets struct {
+	startBuf, endBuf int    // inidices of start and end buffers of object data in the msg
```
neild, line 1747: typo: indices

storage/grpc_client.go
```diff
@@ -1721,21 +1728,323 @@ const (
...
+	// Check if the tag will fit in the current buffer. If not, copy the next 10
+	// bytes into a new buffer to ensure that we can read the tag correctly
+	// without it being divided between buffers.
+	tagBuf := b[d.currOff:]
+	remainingInBuf := len(tagBuf)
+	log.Printf("calling tag at offset %v", d.off)
+	log.Printf("tagbuf: %v, b: %v, d.off: %v\n", len(tagBuf), len(b), d.off)
+	if remainingInBuf < binary.MaxVarintLen64 {
+		log.Print("copying buf")
+		tagBuf = append(b[d.currOff:], ((*d.databufs)[d.currBuf+1].ReadOnlyData()[:binary.MaxVarintLen64-remainingInBuf])...)
```
neild, line 1771: This looks hinky: b comes from a ReadOnlyData method, but you're potentially modifying it.

I'd add a scratch buffer to readResponseDecoder:

```
type readResponseDecoder struct {
  tagbuf [binary.MaxVarintLen64]byte
}
```

```
if remainingInBuf  < binary.MaxVarintLen64 {
  nextBuf := d.databufs[d.currBuf+1].ReadOnlyData()
  copy(d.tagbuf[:], tagBuf)
  copy(d.tagbuf[len(tagBuf):], nextBuf)
  newBufLen := min(len(d.tagbuf), len(tagBuf)+len(nextBuf))
  tagBuf = d.tagBuf[:newBufLen]
}

Comment (neild) [inline storage/grpc_client.go:1771]: Cases that I'm not sure you're handling:

What if d.currBuf is the last buffer in d.databufs?
What if d.currBuf is the last buffer, but doesn't contain enough data to fill out 10 bytes?
What if d.currBuf isn't the last buffer, but doesn't contain enough data to fill out 10 bytes? (Probably can't happen if there's a minimum buffer size, so maybe not important.)

File: storage/grpc_client.go [hunk 11]
```diff
@@ -1721,21 +1728,323 @@ const (
...
+	}
+
+	d.off += uint64(tagLength)
+	return fieldNum, fieldType, nil
+}
+
+// Consume a varint that represents the length of a bytes field. Return the length of
+// the data, and advance the offsets by the length of the varint.
+func (d *readResponseDecoder) consumeVarint() (uint64, error) {
+	b := (*d.databufs)[d.currBuf].ReadOnlyData()
```
neild, line 1854: I wonder if it's be worth stashing the current buffer's []byte in readResponseDecoder, to avoid the need for repeated calls to ReadOnlyData.

Or perhaps even stash a [][]byte and materialize all the buffers once at the start of decoding. You could avoid an allocation in the common case by keeping a [N][]byte in the decoder:

```
type readResponseDecoder struct {
  bufs [][]byte
  staticBufs [4][]byte
}

func (d *readResponseDecoder) init() {
  d.bufs = d.staticBufs[:0]
  for _, b := range d.databufs {
    d.bufs = append(d.bufs, b.ReadOnlyData())
  }
}
```

storage/grpc_client.go
```diff
@@ -955,37 +958,44 @@ func (c *grpcStorageClient) RewriteObject(ctx context.Context, req *rewriteObjec
 	return r, nil
 }
 
-// bytesCodec is a grpc codec which permits receiving messages as either
-// protobuf messages, or as raw []bytes.
-type bytesCodec struct {
-	encoding.Codec
+type bytesCodecV2 struct {
+	encoding.CodecV2
```
dfawley, line 962: Why embed here?  You'll get nil panics if you forget to implement something.

If you're trying to "show" that you're implementing an interface, you can do:

```go
var _ encoding.CodecV2 = bytesCodecV2{}
```

storage/grpc_client.go
```diff
@@ -955,37 +958,44 @@ func (c *grpcStorageClient) RewriteObject(ctx context.Context, req *rewriteObjec
...
 }
 
-func (bytesCodec) Marshal(v any) ([]byte, error) {
+// Marshal is used to encode messages to send for bytesCodecV2. Since we are only
+// using this to send ReadObjectRequest messages we don't need to recycle buffers
+// here.
+func (bytesCodecV2) Marshal(v any) (mem.BufferSlice, error) {
```
dfawley, line 968: OR, embed `encoding.CodecV2` and populate it with `encoding.GetCodec("proto")`, and don't override this method at all.

storage/grpc_client.go
```diff
@@ -955,37 +958,44 @@ func (c *grpcStorageClient) RewriteObject(ctx context.Context, req *rewriteObjec
...
 	switch v := v.(type) {
-	case *[]byte:
-		// If gRPC could recycle the data []byte after unmarshaling (through
-		// buffer pools), we would need to make a copy here.
+	case *mem.BufferSlice:
 		*v = data
 		return nil
 	case proto.Message:
-		return proto.Unmarshal(data, v)
+		buf := data.MaterializeToBuffer(mem.DefaultBufferPool())
```
dfawley, line 990: This could instead also defer to the embedded proto codec's implementation if you go that route.

storage/grpc_client.go
```diff
@@ -1062,11 +1071,12 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
...
 			// Use a custom decoder that uses protobuf unmarshalling for all
-			// fields except the checksummed data.
-			// Subsequent receives in Read calls will skip all protobuf
-			// unmarshalling and directly read the content from the gRPC []byte
-			// response, since only the first call will contain other fields.
-			msg, err = readFullObjectResponse(databuf)
+			// fields except the object data. Object data is handled separately
+			// to avoid a copy.
+			decoder = &readResponseDecoder{
+				databufs: databuf,
```
dfawley, line 1077: Nit: make your local var & field name agree on plurality?

storage/grpc_client_test.go
dfawley: It might be interesting to split at every single place possible.  I.e.:

```go
for offs := 1; offs < len(buf); offs++ {
	unmarshal(buf[:offs], buf[offs:])
}
```

And/or a test case that literally puts exactly one byte per buffer.

storage/grpc_client.go
tritone: Good callout, I was running into panics for some of these cases. Should be fixed now -- if we are in the last buffer we just return what is left, which might be less than 10 bytes (but that's fine as long as the message is formatted correctly).

storage/grpc_client.go
tritone: I think this was something that @neild had in the original codec, but I can remove.

storage/grpc_client.go
tritone: Any particular reason to do this? or just to keep up with any changes to encoding.CodecV2's implementation?

storage/grpc_client_test.go
tritone: Ended up adding splits down to every byte which I think covers this pretty well. Will file a follow up issue to look into fuzzing.

storage/grpc_client.go
dfawley: Yeah, in theory we could make improvements over there, and this would pick those up.  Definitely an optional thing to do, though.

storage/grpc_client.go
tritone: Made this change.

storage/grpc_client.go
tritone: I'll leave this as is for now, can update this later.

storage/grpc_client.go
tritone: Actually I think this is no longer necessary with the changes I made here to call copyNextBytes.

storage/grpc_client.go
tritone: I don't fully understand this, why would staticBufs be of len 4?

I think I'll probably just leave this for now if that's okay, I don't see overhead from these calls.

storage/grpc_client.go
```diff
@@ -1562,21 +1577,297 @@ const (
...
+func (d *readResponseDecoder) copyNextBytes(n int) []byte {
+	remaining := n
+	if r := d.databufs.Len() - int(d.off); r < remaining {
+		remaining = r
+	}
+	currBuf := d.currBuf
+	currOff := d.currOff
+	var buf []byte
+	for remaining > 0 {
+		b := d.databufs[currBuf].ReadOnlyData()
```
neild, line 1635: Could simplify the code a tiny bit by making b start at the current offset:

```
b := b.databufs[currBuf].ReadOnlyData()[currOff:]
if len(b) < remaining {
  buf = append(buf, b...)
  remaining -= len(b)
  currBuf++
  currOff = 0
} else {
  buf = append(buf, b[:remaining]...)
  remaining = 0
}
```

storage/grpc_client.go
```diff
@@ -1562,21 +1577,297 @@ const (
...
+		remainingInCurr := uint64(d.databufs[d.currBuf].Len()) - d.currOff
+		if remainingInCurr <= remaining {
+			remaining -= remainingInCurr
+			d.currBuf++
+			d.currOff = 0
+		} else {
+			d.currOff += remaining
+			remaining = 0
+		}
+	}
```
neild, line 1664: Might be a bit simpler as:

```
d.currOff += n
for d.currBuf < len(d.databufs) && d.currOff >= d.databufs[d.currBuf].Len() {
  d.currOff -= d.databufs[d.currBuf].Len()
  d.currBuf++
}
```

neild (approved):
Looks good to me!

frankyn (approved):

PR #10893, opened 2024-09-20
feat(storage/transfermanager): add SkipIfExists option

BrennaEpp:


storage/transfermanager/integration_test.go
```diff
@@ -295,6 +296,75 @@ func TestIntegration_DownloadDirectoryAsync(t *testing.T) {
...
+			Bucket:         tb.bucket,
+			LocalDirectory: localDir,
+			OnObjectDownload: func(got *DownloadOutput) {
+				callbacks <- true
+
+				if got.Err != nil {
+					t.Errorf("result.Err: %v", got.Err)
+				}
+
+				if strings.EqualFold(got.Object, "dir/objA") {
```
frankyn, line 319: This only checks `dir/objA`, do you want to check the other included objects?

storage/transfermanager/integration_test.go
```diff
@@ -295,6 +296,75 @@ func TestIntegration_DownloadDirectoryAsync(t *testing.T) {
 				t.Errorf("unexpected file %q in dir", entry.Name())
 			}
 		}
+
+		// Now attempt to download the entire directory.
+		// The existing files should be skipped.
+		callbacks := make(chan bool)
```
frankyn, line 302: It would help to clarify that you're checking that SkipIfExists() should not attempt downloading included files from previous directory download;

```
// In lex order we have:
// "dir/nested/again/obj1", -- included
// "dir/nested/objA", -- skipped
// "dir/file" -- included
// "dir/objA", -- skipped
// "dir/objB", -- skipped
// "dir/objC", -- included
```

storage/transfermanager/integration_test.go
```diff
@@ -295,6 +296,75 @@ func TestIntegration_DownloadDirectoryAsync(t *testing.T) {
...
+
+		gotCallbacks := 0
+		done := make(chan bool)
+		go func() {
+			for {
+				select {
+				case <-done:
+					break
+				case <-callbacks:
+					gotCallbacks++
```
frankyn, line 355: For my learning, `gotCallbacks` is synchronized; so the directory download which uses a mutex could be rewritten this way. Is that correct?

storage/transfermanager/integration_test.go
```diff
@@ -295,6 +296,75 @@ func TestIntegration_DownloadDirectoryAsync(t *testing.T) {
...
+			}
+		}()
+
+		if _, err := d.WaitAndClose(); err != nil {
+			t.Errorf("d.WaitAndClose: %v", err)
+		}
+		done <- true
+
+		if want, got := len(tb.objects)-wantObjs, gotCallbacks; want != got {
+			t.Errorf("expected to receive %d callbacks, got %d", want, got)
```
frankyn, line 366: gotCallbacks could be better represented as `objectsDownloaded`; it's subtle otherwise.

storage/transfermanager/integration_test.go
BrennaEpp: Added.

storage/transfermanager/integration_test.go
BrennaEpp: I added a comment here and put those objects in a variable - does that seem sufficient here?

storage/transfermanager/integration_test.go
BrennaEpp: Correct. I re-wrote that logic, PTAL.

storage/transfermanager/integration_test.go
frankyn: Yes, thank you;

frankyn (approved):

Issue #10896, opened 2024-09-20
bigquery: need additional documentation for tags


septianhari-afk:
## Is your feature request related to a problem? Please describe.
The documentation on https://pkg.go.dev/cloud.google.com/go/bigquery seems doesn't tells about how to use tags on query result

## Describe the solution you'd like
I just want to add some line that we could use `bigquery` tags so the struct doesnt have to be exactly same with the query. E.g:
```
type Count struct {
    YearCount int `bigquery:"year"`
    NumCount  int `bigquery:"num"`
}
```

## Describe alternatives you've considered
N/A

## Additional context
N/A

PR #10897, opened 2024-09-20
Adding documentation about bigquery tags

septianhari-afk:
Based on my issue on https://github.com/googleapis/google-cloud-go/issues/10896, I create MR to add minimal documentation.


bigquery/doc.go
```diff
@@ -85,6 +85,24 @@ You can also use a struct whose exported fields match the query:
 	    fmt.Println(c)
 	}
 
+The struct also can tagged using `bigquery` tag if it's not match with the field name:
```
alvarowolfx, line 88: Maybe change to something like `Struct fields can also be tagged using the `bigquery` struct tag to match the source field name:`

alvarowolfx (changes requested):
Wording on new doc can be improved

PR #10901, opened 2024-09-20
chore(storage): remove HMAC and SA grpc support

frankyn:
Pending: #10875


storage/storage.go
```diff
@@ -2349,6 +2349,7 @@ func toProtoChecksums(sendCRC32C bool, attrs *ObjectAttrs) *storagepb.ObjectChec
 }
```
tritone, line 2349: Can you also update the docs for NewGRPCClient and the summary for gRPC in doc.go to indicate which methods are not supported?

tritone (approved):
LGTM, just update package docs as well.

PR #10913, opened 2024-09-25
feat(storage/dataflux): add worksteal algorithm to fast-listing

akansha1812:
feat: add worksteal algorithm to fast-listing
Dataflux fast-listing will be used to quickly list objects in a bucket in parallel leveraging worksteal algorithm. 

Worksteal algorithm splits a given namespace into multiple ranges for multiple workers(goroutines) to list objects in gcs bucket in parallel. 


storage/dataflux/fast_list.go
```diff
@@ -62,32 +68,53 @@ type Lister struct {
...
 	bucket := c.Bucket(in.BucketName)
+
+	// If parallelism is not given, set default value to 10x the number of available CPU.
+	if in.Parallelism == 0 {
+		in.Parallelism = runtime.NumCPU() * 10
+	}
+	// Initialize range channel with entire namespace of object for given prefix,
+	// startoffset and endoffset. For the default range to list is entire namespace,
+	// start and end will be empty.
+	rangeChannel := make(chan *listRange, in.Parallelism*2)
```
BrennaEpp, line 105: What is the reasoning behind `in.Parallelism*2` buffer size for the channel?

storage/dataflux/fast_list.go
```diff
@@ -150,6 +179,39 @@ func (c *Lister) NextBatch(ctx context.Context) ([]*storage.ObjectAttrs, error)
...
 
-	// TODO: Close range channel for worksteal lister.
+// updateStartEndOffset updates start and end offset based on prefix.
+// If a prefix is given, start and end should contain the value after prefix.
+// For example, if start is "prefix_a", then start should be "_a".
+func updateStartEndOffset(start, end, prefix string) (string, string) {
+	if prefix == "" {
+		return start, end
+	}
+	if start != "" && end != "" && start >= end {
```
BrennaEpp, line 194: Shouldn't this possibly modify `start` so that the prefix is not included if present?

storage/dataflux/fast_list.go
```diff
@@ -150,6 +179,39 @@ func (c *Lister) NextBatch(ctx context.Context) ([]*storage.ObjectAttrs, error)
...
 func (c *Lister) Close() {
+	if c.ranges != nil {
+		close(c.ranges)
+	}
+}
 
-	// TODO: Close range channel for worksteal lister.
+// updateStartEndOffset updates start and end offset based on prefix.
+// If a prefix is given, start and end should contain the value after prefix.
+// For example, if start is "prefix_a", then start should be "_a".
```
BrennaEpp, line 189: What happens if a prefix is given, ex "someprefix", but `start`/`end` does not contain that prefix? Looks like based on my understanding of the code below, if `start` is lexicographically smaller than "someprefix", it would return the empty string, otherwise it would just return `start`? 

Can you update this comment with the expected behaviour?
Also please add a unit test for this method.

storage/dataflux/fast_list.go
```diff
@@ -150,6 +179,39 @@ func (c *Lister) NextBatch(ctx context.Context) ([]*storage.ObjectAttrs, error)
...
+	if end != "" {
+		if len(end) > len(prefix) && strings.HasPrefix(end, prefix) {
+			end = end[len(prefix):]
+		} else if end > prefix {
+			end = ""
+		} else {
+			return end, end
+		}
+	}
+	return start, end
```
BrennaEpp, line 216: Is this case only reached when start and end == ""?

storage/dataflux/worksteal.go
```diff
@@ -45,16 +56,171 @@ type worker struct {
...
 // It performs worksteal to achieve highly dynamic object listing.
-func (c *Lister) workstealListing(ctx context.Context) []*storage.ObjectAttrs {
+// workstealListing creates multiple (parallelism) workers that simultaneosly lists
+// objects from the buckets.
+func (c *Lister) workstealListing(ctx context.Context) ([]*storage.ObjectAttrs, error) {
+	var workerErr []error
+	// Idle channel is used to track number of idle workers.
+	idleChannel := make(chan int, c.parallelism)
+	// Result channel is used to track the results from each worker.
+	result := &listerResult{
```
BrennaEpp, line 71: This is not a channel, right?

storage/dataflux/worksteal.go
```diff
@@ -16,9 +16,20 @@ package dataflux
...
 	"sync"
+	"time"
 
 	"cloud.google.com/go/storage"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	// defaultAlphabet used to initiliaze rangesplitter.
+	defaultAlphabet = "ab"
```
BrennaEpp, line 29: Curious what the reasoning behind this default is

storage/dataflux/worksteal.go
```diff
@@ -16,9 +16,20 @@ package dataflux
...
+	// sleepDurationWhenIdle is the milliseconds we want each worker to sleep before checking
+	// the next update if it is idle.
+	sleepDurationWhenIdle = 200
```
BrennaEpp, line 32: Suggestion: make this a time.Duration

storage/dataflux/worksteal.go
```diff
@@ -45,16 +56,171 @@ type worker struct {
...
+				// Worker is now idle.
+				w.status = idle
+			}
+			break
+		}
+
+		// If a worker is idle, sleep for a while before checking the next update.
+		// Worker is active when it finds work in range channel.
+		if w.status == idle {
+			if len(w.lister.ranges) == 0 {
```
BrennaEpp, line 140: Why not use a `select`?

```
case <-ctx.Done():
   return ctx.Err()
case newRange := <- w.lister.ranges:
   // do work
default:
   time.Sleep(...)   
```

The shutdown signal could be sent through a channel as well or checked in the default case.

storage/dataflux/worksteal.go
```diff
@@ -45,16 +56,171 @@ type worker struct {
...
+				time.Sleep(time.Millisecond * time.Duration(sleepDurationWhenIdle))
+				continue
+			} else {
+				newRange := <-w.lister.ranges
+				<-w.idleChannel
+				w.updateWorker(newRange.startRange, newRange.endRange, active)
+			}
+		}
+		// Active worker to list next page of objects within the range.
+		nextPageResult, err := objectLister(ctx, nextPageOpts{
```
BrennaEpp, line 150: Suggest splitting the code following this (active worker listing objects) out into a separate method.

storage/dataflux/worksteal.go
```diff
@@ -45,16 +56,171 @@ type worker struct {
...
+	// If all the workers are idle and range channel is empty, no more objects to list.
+	if len(w.idleChannel) == w.lister.parallelism && len(w.lister.ranges) == 0 {
+		return true
+	}
+	// If number of objects listed is equal to the given batchSize, then shutdown.
+	// If batch size is not given i.e. 0, then list until all objects have been listed.
+	if w.lister.batchSize > 0 && len(w.result.objects) >= w.lister.batchSize {
+		return true
+	}
+	return false
```
BrennaEpp, line 211: Suggest structuring this like the following (feel free to modify names etc):

```suggestion
	noMoreObjects := len(w.idleChannel) == w.lister.parallelism && len(w.lister.ranges) == 0 
	alreadyListedBatchSizeObjects := w.lister.batchSize > 0 && len(w.result.objects) >= w.lister.batchSize 
	return noMoreObjects || alreadyListedBatchSizeObjects
```

BrennaEpp (commented):
Some initial thoughts. This should be tested for race conditions as well but that can be as integration tests in the follow up.

storage/dataflux/fast_list.go
akansha1812: rangeChannel stores ranges for the workers.  This is to covers a scenario where all active workers will split work att the same time and push split work to rangeChannel.

storage/dataflux/fast_list.go
akansha1812: If start comes lexicographically after end  or equal to end, then no objects will be listed. Adding prefix or not,  will not make a difference.

storage/dataflux/fast_list.go
akansha1812: This is what will be  expected.  
If a `prefix` is given and `start` is smaller, that means listing will begin from the very first object with the given `prefix`. So the smallest possible `Startoffset` is the prefix. 

I have added an example.

storage/dataflux/fast_list.go
akansha1812: yes.

storage/dataflux/worksteal.go
akansha1812: Its not a channel. Thanks for catching this. 
Updated the comment.

storage/dataflux/worksteal.go
akansha1812: For this case, we'll have to get newRange, if the worker is already active, we'll have to put the range back to the channel and then start working. This  might increase wait time for idle workers waiting for work. 

 If we want to use select statement, we can do this I think:

case <-ctx.Done():
   return ctx.Err()
case w.shutdownSignal():
   // shutdown logic
case w.status == idle &&  len(w.lister.ranges) == 0:
   // sleep
default:
   // if idle, pick work. 
   // do work

storage/dataflux/worksteal.go
akansha1812: To understand this, are you suggesting a different function
```
doneListing, err : w.objectLister(ctx)
if err!=nil {
			// handle error
}
if donelisting{
			w.status = idle
			w.idleChannel <- 1
			w.generation = int64(0)
			continue
}
// split work
```

and w.objectLister to take care of the nextPageResult logic?

storage/dataflux/worksteal.go
akansha1812: In rangesplitter we dynamically add more alphabets so object names are not limited to english alphabet. 

For N-base to 10-base conversion there should be at least two default characters for rangesplitter `splitRange` method.
Example - If `defaultAlphabet` is `"a"` , and for `splitRange (start="a",  end="", numSplits=2)` , here end is pointing to the last name after `"a"`.  While calculating 10-base integer value, it will keep appending `"a"` to `start` and `end`.

storage/dataflux/worksteal.go
```diff
@@ -138,7 +138,7 @@ func (w *worker) doWorkstealListing(ctx context.Context) error {
 		// Worker is active when it finds work in range channel.
 		if w.status == idle {
 			if len(w.lister.ranges) == 0 {
-				time.Sleep(time.Millisecond * time.Duration(sleepDurationWhenIdle))
+				time.Sleep(time.Millisecond * sleepDurationWhenIdle)
```
BrennaEpp, line 141: Converting to milliseconds should be part of the definition of the `sleepDurationWhenIdle`.

storage/dataflux/worksteal.go
```diff
@@ -200,15 +183,15 @@ func (w *worker) doWorkstealListing(ctx context.Context) error {
...
-		return true
-	}
+	noMoreObjects := len(w.idleChannel) == w.lister.parallelism && len(w.lister.ranges) == 0
 	// If number of objects listed is equal to the given batchSize, then shutdown.
 	// If batch size is not given i.e. 0, then list until all objects have been listed.
-	if w.lister.batchSize > 0 && len(w.result.objects) >= w.lister.batchSize {
+	alreadyListedBatchSizeObjects := len(w.idleChannel) == w.lister.parallelism && len(w.lister.ranges) == 0
+	if len(w.idleChannel) == w.lister.parallelism && len(w.lister.ranges) == 0 {
 		return true
 	}
```
BrennaEpp, line 192: I don't think this is needed anymore

storage/dataflux/worksteal.go
BrennaEpp: However, this seems ok too. My suggestion if keeping as is would be to change the loop to have a terminating condition: 
`for !w.shutDownSignal()`

And then do everything currently in the `w.shutDownSignal()` conditional at the end of the function.

internal/testutil/context.go
```diff
@@ -37,7 +37,7 @@ const (
 // ProjID returns the project ID to use in integration tests, or the empty
 // string if none is configured.
 func ProjID() string {
-	return os.Getenv(envProjID)
+	return "zimbruplayground"
```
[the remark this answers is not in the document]
akansha1812: Thanks Brenna for catching this!

storage/dataflux/worksteal.go
akansha1812: Made shutdown signal as loop terminating condition.

storage/dataflux/worksteal.go
akansha1812: Yes, we'll need at least two characters, could be any character.

BrennaEpp: Please add that in a comment here. Thanks!

BrennaEpp (approved):
Thank you!

PR #10920, opened 2024-09-25
fix(auth): Make aws credentials provider retrieve fresh credentials each time a new subject token is requested

### LINKED ISSUE #10832: auth: get AWS security credentials constantly

We've been encountered following error in ExchangeToken function.

```
credentials: status code 400: {"error":"invalid_grant","error_description":"Received invalid AWS response of type ExpiredToken with error message: The security token included in the request is expired"}
```

I believe this cause is that awsSubjectProvider isn't getting security credentials constantly.

I think that awsSubjectProvider should always get security credentials as long as the securityCredentialsProvider is set.

```diff
--- a/auth/credentials/internal/externalaccount/aws_provider.go
+++ b/auth/credentials/internal/externalaccount/aws_provider.go
@@ -94,7 +94,7 @@ func (sp *awsSubjectProvider) subjectToken(ctx context.Context) (string, error)
        if sp.RegionalCredVerificationURL == "" {
                sp.RegionalCredVerificationURL = defaultRegionalCredentialVerificationURL
        }
-       if sp.requestSigner == nil {
+       if sp.requestSigner == nil || sp.securityCredentialsProvider != nil {
                headers := make(map[string]string)
                if sp.shouldUseMetadataServer() {
                        awsSessionToken, err := sp.getAWSSessionToken(ctx)
```

Because the security credentials tend to have an expiration date.

## Describe alternatives you've considered

I considered possibility that this change cause too many requests error to fetch AWS security credentials on some users environment. It varies depending on the implementation of securityCredentialsProvider for each user land.

Thus I've had one another approach that determining environment variable is set or not.

```diff
--- a/auth/credentials/internal/externalaccount/aws_provider.go
+++ b/auth/credentials/internal/externalaccount/aws_provider.go
@@ -69,6 +69,9 @@ const (
        awsSecretAccessKeyEnvVar = "AWS_SECRET_ACCESS_KEY"
        awsSessionTokenEnvVar    = "AWS_SESSION_TOKEN"

+       // Get AWS security credentials in every request.
+       getAwsSecurityCredentilasConstantly = "AWS_GET_SECURITY_CREDENTIALS_CONTANTLY"
+
        awsTimeFormatLong  = "20060102T150405Z"
        awsTimeFormatShort = "20060102"
        awsProviderType    = "aws"
@@ -94,7 +97,11 @@ func (sp *awsSubjectProvider) subjectToken(ctx context.Context) (string, error)
        if sp.RegionalCredVerificationURL == "" {
                sp.RegionalCredVerificationURL = defaultRegionalCredentialVerificationURL
        }
-       if sp.requestSigner == nil {
+       getSecurityCredentilasConstantly := false
+       if env := getenv(getAwsSecurityCredentilasConstantly); env != "" {
+               getSecurityCredentilasConstantly = true
+       }
+       if getSecurityCredentilasConstantly || sp.requestSigner == nil {
                headers := make(map[string]string)
                if sp.shouldUseMetadataServer() {
                        awsSessionToken, err := sp.getAWSSessionToken(ctx)
```

But I believe that awsSubjectProvider should not care about caching security credentials especially in case of securityCredentialsProvider was set. So I prefer the first solution that I described.

## Additional context

I'm using awsSecurityCredentialProvider with cloud.google.com/go/bigquery package.
I currently shared bigquery client object in each goroutine and I believe that this problem is not occured if I create bigquery client in each goroutine. But I don't want to do so.

Why I'm using awsSecurityCredentialProvider is that I'm running program on ECS Fargate container and I want to fetch AWS security credentials from container metadata endpoint on ECS Fargate.

aeitzman:
fixes #10832


auth/credentials/internal/externalaccount/aws_provider.go
```diff
@@ -94,32 +94,30 @@ func (sp *awsSubjectProvider) subjectToken(ctx context.Context) (string, error)
...
 
+	awsSecurityCredentials, err := sp.getSecurityCredentials(ctx, headers)
+	if err != nil {
+		return "", err
+	}
+	if sp.region, err = sp.getRegion(ctx, headers); err != nil {
+		return "", err
+	}
+	sp.requestSigner = &awsRequestSigner{
+		RegionName:             sp.region,
```
quartzmo, line 117: @aeitzman In the related issue, you wrote:

>  Instead of completely rebuilding the request signer, it might make more sense to retrieve a new security credential and modify requestSigner.AwsSecurityCredentials, that would better match the implementation in other languages.I can make a PR for that fix.

What happened to this plan? Is it not desirable or practical to cache the result of `getRegion`?

aeitzman: I was mistaken when I made the earlier comment, we actually do grab the region again in other languages isntead of caching it (see https://github.com/googleapis/google-auth-library-java/blob/afe4ff9f1f3c406c72ecb66fba50d7463f620467/oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java#L134C1-L135C1)

aeitzman: Additionally, now that we have the custom provider as an option for AWS, it would be confusing if we re-call the provider for the security credentials each time, but not the region. It's possible some implementations of the provider could handle multiple credentials/regions based on the options/context getting passed into it

quartzmo (approved):

PR #10580, opened 2024-07-23
feat(bigquery): add BigQuery Storage Read client options

alvarowolfx:
Still a work in progress on how to customize the ReadSession and internal parallelism configuration when using the BigQuery Storage Read API for fetching data. Methods are not final.

Example usage:

```go
client, err := bigquery.NewClient(ctx, "project-id")
if err != nil {
  // TODO: Handle error.
}
err = client.EnableStorageReadClient(ctx,
  bigquery.WithStorageReaderOptions(
    bigquery.WithMaxWorkerCount(runtime.GOMAXPROCS(0)),
    bigquery.WithMaxStreamCount(0),
  ),
)
```

PR #10924, opened 2024-09-26
feat(bigquery): Configurable table read session project

juanli16:
When creating a BigQuery table RowIterator with StorageReadAPI enabled, the read session is created by default in the table's project. We should be able to overwrite this, so that we can keep the storage of the table data and the permission/cost management of the read session separate.

To accomplish this, `TableReadOption` with `WithClientProject` option is added, and if set, it will create the session using the client's project, otherwise, it keeps the default behaviour.

### DIFF
--- a/bigquery/storage_client.go
+++ b/bigquery/storage_client.go
@@ -107,15 +107,8 @@ func (c *readClient) sessionForTable(ctx context.Context, table *Table, rsProjec
 		settings.maxStreamCount = 1
 	}
 
-	// configure where the read session is created
-	readSessionProjectID := table.ProjectID
-	if useClientProject {
-		readSessionProjectID = c.projectID
-	}
-
 	rs := &readSession{
 		ctx:                   ctx,
-		readSessionProjectID:  readSessionProjectID,
 		table:                 table,
 		tableID:               tableID,
 		projectID:             rsProjectID,
--- a/bigquery/storage_client.go
+++ b/bigquery/storage_client.go
@@ -107,9 +107,15 @@ func (c *readClient) sessionForTable(ctx context.Context, table *Table, rsProjec
 		settings.maxStreamCount = 1
 	}
 
+	// configure where the read session is created
+	readSessionProjectID := table.ProjectID
+	if useClientProject {
+		readSessionProjectID = c.projectID
+	}
+
 	rs := &readSession{
 		ctx:                   ctx,
-		projectID:             c.projectID,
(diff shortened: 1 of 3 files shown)


alvarowolfx:
hey @juanli16 thanks for the contribution. Some question on this improvement and your use case ? 

Do you have multiple billing project that can be target for different table reads or it's the same project ID that you want to configure the same as the project ID set up on on the BigQuery client ?  We make this configuration more global to the `ReadClient` and use the `projectID` already on the `bigquery.Client`. 

One concern that I'm having here is changing the signature of the `Table.Read` method to add client options, we thought on doing that in the past, but we always go back and forth on that idea. I'd rather prefer this for now to be an option on the `ReadCient`, like I started on PR https://github.com/googleapis/google-cloud-go/pull/10580. I can check to merge PR https://github.com/googleapis/google-cloud-go/pull/10580 and then you turn this improvement into a ReadClient level option, what do you think ?

And can we add integration tests ? Probably exercise reading a public dataset table and creating a session from the main project ID that the client is set up.

juanli16:
> Do you have multiple billing project that can be target for different table reads or it's the same project ID that you want to configure the same as the project ID set up on on the BigQuery client ?

Yes, in our use case, we have the same billing project that we want to use for reading tables stored in different projects.

> I'd rather prefer this for now to be an option on the ReadCient, like I started on PR https://github.com/googleapis/google-cloud-go/pull/10580. I can check to merge PR https://github.com/googleapis/google-cloud-go/pull/10580 and then you turn this improvement into a ReadClient level option, what do you think ?

This makes sense to me, it will make the setting more global, and we won't have to set it per table read.

And certainly, I can try to add integration test for it.

fsaintjacques:
It's more about having right to read from a table but without billing access to where it lives. In fact, in our use case, the service account does have the permissions to read on the table, but no other permissions including on the project where said table lives.

It's akin to controlling in which project `job.insert` is submitted, sometimes you simply don't have the choice (think public datasets).

juanli16:
Hi @alvarowolfx , I have rebased my PR on top of your change in https://github.com/googleapis/google-cloud-go/pull/10932 which has been merged. Now it only introduces the `TableReadOption` that allows to use different project for read sessions, with a fallback that uses the original bigquery client's PR. Let me know if this is still desired, or I can close it. As your PR already addressed my original concern :)

alvarowolfx:
superseded by #10932

PR #10939, opened 2024-10-01
feat(vertexai): Add supported models in tokenizer

happy-qiao:


eliben:
For future PRs, please create a fresh branch for each one.
Right now I see 25 commits in this PR, combining your other commits from older PRs, this is confusing for reviewing and I worry about accidental merging every time.

PR #10953, opened 2024-10-04
feat(auth): add universe domain support to credentials/impersonate

quartzmo:
* fix: return err if both Client and Credentials are set in CredentialsOptions
* fix: propagate Credentials universe domain in httptransport.AddAuthorizationMiddleware


auth/credentials/impersonate/impersonate.go
```diff
@@ -87,30 +90,32 @@ func NewCredentials(opts *CredentialsOptions) (*auth.Credentials, error) {
 		client = opts.Client
 	}
 
+	universeDomainProvider := resolveUniverseDomainProvider(opts, creds)
 	// If a subject is specified a domain-wide delegation auth-flow is initiated
 	// to impersonate as the provided subject (user).
 	if opts.Subject != "" {
-		if !opts.isUniverseDomainGDU() {
+		gdu, err := isUniverseDomainGDU(universeDomainProvider)
```
codyoss, line 97: Can this check be moved to the impl for the token provider? Having this code here puts this check at construction time vs usage-time.

auth/credentials/impersonate/impersonate.go
```diff
@@ -125,16 +130,43 @@ func NewCredentials(opts *CredentialsOptions) (*auth.Credentials, error) {
...
+// universe, with the following precedence:
+//
+// 1. A non-empty CredentialsOptions.UniverseDomain.
+// 2. If non-nil creds, then creds.UniverseDomain.
+// 3. The default value "googleapis.com".
+//
+// This is the universe domain configured for the credentials, which will be
+// used in endpoint(s), and compared to the universe domain that is separately
+// configured for the client.
+func resolveUniverseDomainProvider(opts *CredentialsOptions, creds *auth.Credentials) auth.CredentialsPropertyProvider {
```
codyoss, line 149: This seems like common logic that could be broadened and used consistently all over, similar to what we do for quotaproject. Could we do something similar here: https://github.com/googleapis/google-cloud-go/blob/main/auth/internal/internal.go#L100

auth/httptransport/httptransport.go
```diff
@@ -170,10 +171,14 @@ func AddAuthorizationMiddleware(client *http.Client, creds *auth.Credentials) er
 			base = http.DefaultTransport
 		}
 	}
+	clientUniverseDomain, err := creds.UniverseDomain(context.Background())
```
codyoss, line 174: This now potentially can hard pull MDS, or at least could in the future. Can this be delayed?

auth/credentials/impersonate/impersonate.go
quartzmo: Fixed, thanks!

auth/httptransport/httptransport.go
quartzmo: Fixed, thanks for catching this.

auth/credentials/impersonate/impersonate.go
quartzmo: If we move as-is this to internal, it will import `CredentialsOptions` and `Credentials`. Maybe we can wait until we see this logic duplicated elsewhere?

auth/httptransport/httptransport.go
```diff
@@ -171,14 +170,10 @@ func AddAuthorizationMiddleware(client *http.Client, creds *auth.Credentials) er
...
 	}
-	clientUniverseDomain, err := creds.UniverseDomain(context.Background())
-	if err != nil {
-		return err
-	}
 	client.Transport = &authTransport{
 		creds:                creds,
 		base:                 base,
-		clientUniverseDomain: clientUniverseDomain,
+		clientUniverseDomain: creds.UniverseDomainPropertyProvider(),
```
codyoss, line 176: Just call this on the creds obj that is already in the middleware. If you call from the middleware it is already lazy and this does not need to be exposed.

quartzmo: Removed, thanks!

auth/auth.go
```diff
@@ -196,6 +196,14 @@ func (c *Credentials) UniverseDomain(ctx context.Context) (string, error) {
 	return v, err
 }
 
+// UniverseDomainPropertyProvider returns a [CredentialsPropertyProvider] for
+// the default service domain for a given Cloud universe.
+//
+// To immediately access the universe domain, use [Credentials.UniverseDomain].
+func (c *Credentials) UniverseDomainPropertyProvider() CredentialsPropertyProvider {
```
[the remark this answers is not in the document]
quartzmo: Removed, thanks!

auth/credentials/impersonate/impersonate.go
```diff
@@ -62,55 +65,49 @@ func NewCredentials(opts *CredentialsOptions) (*auth.Credentials, error) {
...
-		creds, err = credentials.DetectDefault(&credentials.DetectOptions{
-			Scopes:           []string{defaultScope},
-			UseSelfSignedJWT: true,
-		})
-		if err != nil {
-			return nil, err
+		if opts.Credentials == nil {
+			creds, err = credentials.DetectDefault(&credentials.DetectOptions{
+				Scopes:           []string{defaultScope},
+				UseSelfSignedJWT: true,
```
codyoss, line 73: Pass through UD here too.

auth/credentials/impersonate/user.go
```diff
@@ -84,21 +85,44 @@ type exchangeTokenResponse struct {
...
 	signedJWT, err := u.signJWT(ctx)
 	if err != nil {
 		return nil, err
 	}
 	return u.exchangeToken(ctx, signedJWT)
 }
 
+// isUniverseDomainGDU returns true if the universe domain is the default Google
+// universe or if it is empty.
+func isUniverseDomainGDU(universeDomainProvider auth.CredentialsPropertyProvider) (bool, error) {
```
codyoss, line 117: Take a context

auth/credentials/impersonate/impersonate.go
quartzmo: I thought about doing this, but I don't think we should, since the value provided would override a `universe_domain` value in the detected credentials (see `resolveUniverseDomain` in `filetypes.go`).

The problem is that the UD provided here is the *client* UD, which must be compared against the *credentials* UD. Users should not normally need to provide a *credentials* UD in code (override) to `DetectDefault`. But if they do, I think they should then call `DetectDefault` externally with the provided UD, then pass the returned credentials to the impersonate call.

quartzmo: TL;DR: The *client* UD available here in `impersonate.CredentialsOptions` should never be used as a *credentials* UD.

auth/credentials/impersonate/user.go
quartzmo: Fixed, thanks!

codyoss (approved):

PR #10956, opened 2024-10-07
feat(spanner): add ResetForRetry method for stmt-based transactions

olavloite:
Read/write transactions that are aborted should preferably be retried using the same session as the original attempt. For this, statement-based transactions should have a ResetForRetry function. This was missing in the Go client library.

This change adds this method, and re-uses the session when possible. If the aborted error happens during the Commit RPC, the session handle was already cleaned up by the original implementation. We will not change that now, as that could lead to breakage in existing code that depends on this. When the Go client is switched to multiplexed sessions for read/write transactions, then this implementation should be re-visited, and it should be made sure that ResetForRetry optimizes the retry attempt for an actual retry.

Updates https://github.com/googleapis/go-sql-spanner/issues/300


spanner/transaction.go
```diff
@@ -1839,7 +1854,22 @@ func (t *ReadWriteStmtBasedTransaction) Rollback(ctx context.Context) {
...
+}
+
+// ResetForRetry resets the transaction before a retry. This should be
+// called if the transaction was aborted by Spanner and the application
+// wants to retry the transaction.
+// It is recommended to use this method above creating a new transaction,
+// as this method will give the transaction a higher priority and thus a
+// smaller probability of being aborted again by Spanner.
+func (t *ReadWriteStmtBasedTransaction) ResetForRetry(ctx context.Context) (*ReadWriteStmtBasedTransaction, error) {
+	if t.state == txNew || t.state == txInit {
```
harshachinta, line 1868: Should we add a check whether the previous attempt failed with ABORTED, since for all the other errors (ex: session not found error) using the same session might not help during retry?

harshachinta (approved):

harshachinta (approved):

PR #10958, opened 2024-10-07
feat(storage): dynamic read request stall timeout

raj-prince:
- Adding a new storage client option to  enable pro-actively retry for stalled request based on dynamic timeout.
- Added emulator test to validated the changes.


storage/option.go
```diff
@@ -108,3 +111,59 @@ func WithDisabledClientMetrics() option.ClientOption {
...
+// Grpc read-operation will be supported soon.
+//
+// Here, the input parameter decides the value of dynamic-timeout.
+// targetPercentile is the desired percentile of the observed latencies.
+// increaseRate determines the rate, timeout is adjusted. High means slow increase in
+// timeout and low means rapid increase.
+// initialTimeout decides the initial timeout.
+// minTimeout, maxTimeout is the lower & upper bound of the timeout.
+//
+// TODO (raj-prince): To keep separate dynamicDelay instance for different BucketHandle.
```
tritone, line 130: Let's move this TODO to the function body, I think it will show up in the godoc if you have it here.

storage/client_test.go
```diff
@@ -1448,6 +1495,10 @@ func createRetryTest(t *testing.T, project, bucket string, client storageClient,
 		t.Fatalf("creating bucket: %v", err)
 	}
 
+	return plantRetryInstructions(t, client, instructions)
+}
+
+func plantRetryInstructions(t *testing.T, client storageClient, instructions map[string][]string) string {
 	// Need the HTTP hostname to set up a retry test, as well as knowledge of
```
tritone, line 1502: This should also have t.Helper(). Actually you could probably just move the client.CreateBucket request to the main test body for anywhere else this is used, and not spin off a separate function for `plantRetryInstructions`

storage/option.go
```diff
@@ -108,3 +111,59 @@ func WithDisabledClientMetrics() option.ClientOption {
...
+// BucketHandle, which is not not the ideal state. As latency depends on location of VM
+// and Bucket, and read latency of different buckets may lie in different range.
+// Hence hence having a separate dynamicTimeout instance at BucketHandle level will
+// be better
+func WithDynamicReadReqStallTimeout(
+	targetPercentile float64,
+	increaseRate float64,
+	initialTimeout time.Duration,
+	minTimeout time.Duration,
+	maxTimeout time.Duration) option.ClientOption {
```
tritone, line 141: I think we could reduce the surface area here. Only targetPercentile and minTimeout are really crucial. increaseRate can be set internally (not sure people would do this level of tweaking), initialTimeout could be equal to minTimeout, and maxTimeout should probably be infinite (or very high) for safety reasons.

tritone: Let's also change the args to a config struct so that we can add stuff in the future if it does end up being needed.

storage/option.go
```diff
@@ -108,3 +111,59 @@ func WithDisabledClientMetrics() option.ClientOption {
...
+// initialTimeout decides the initial timeout.
+// minTimeout, maxTimeout is the lower & upper bound of the timeout.
+//
+// TODO (raj-prince): To keep separate dynamicDelay instance for different BucketHandle.
+// Currently, dynamicTimeout is kept at the client and hence shared across all the
+// BucketHandle, which is not not the ideal state. As latency depends on location of VM
+// and Bucket, and read latency of different buckets may lie in different range.
+// Hence hence having a separate dynamicTimeout instance at BucketHandle level will
+// be better
+func WithDynamicReadReqStallTimeout(
```
tritone, line 136: As discussed, we can move this to a storage/experimental subpackage. @frankyn will follow up on this.

tritone (commented):
Looking good overall, a few more minor points.

storage/option.go
raj-prince: We can hardcore the value of maxTimeout. 

For increaseRate - I am not very sure if we should hardcore the value of increaseRate. Although less likely but we might need to change the value of increaseRate based on testing.  Is it possible to expose these as part of experimental flag and later remove it if not required.

For initialTimeout - I was thinking to keep the initial value slightly higher than minTimeout. This will give some time to the dynamicDelay algorithm to converge.

>> Let's also change the args to a config struct so that we can add stuff in the future if it does end up being needed. 
- Yeah, make sense.

raj-prince: As discussed, we will keep large maxTimeout. increaseRate and initialTimeout (not discussed, but still feel like we might need to change this) will be exposed via environment variable.

tritone (commented):
Few more docs and defaults suggestions. Also, please follow https://github.com/googleapis/google-cloud-go/pull/10984 to start an experimental package for the option.

storage/option.go
```diff
@@ -119,51 +158,33 @@ func (w *withDisabledClientMetrics) ApplyStorageOpt(c *storageConfig) {
...
-// and Bucket, and read latency of different buckets may lie in different range.
-// Hence hence having a separate dynamicTimeout instance at BucketHandle level will
-// be better
-func WithDynamicReadReqStallTimeout(
-	targetPercentile float64,
-	increaseRate float64,
-	initialTimeout time.Duration,
-	minTimeout time.Duration,
-	maxTimeout time.Duration) option.ClientOption {
+func WithDynamicReadReqStallTimeout(drrst *DynamicReadReqStallTimeoutConfig) option.ClientOption {
```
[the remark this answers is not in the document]
raj-prince: Also, since we are using a struct now rather than args, we should choose sensible default values. I suggest 500 ms & 0.99.
> Yes, 500ms and 0.99 make sense. But this user has to set these value while using this option. Right?

Or do we want to expose the field as pointer so that we can set the right default if it's not set by user?

storage/option.go
raj-prince: Done, thanks @frankyn for the  storage/experimental setup support (PR: https://github.com/googleapis/google-cloud-go/pull/10984).

storage/option.go
```diff
@@ -63,10 +70,10 @@ func getDynamicReadReqInitialTimeoutSecFromEnv(defaultVal time.Duration) time.Du
...
 // set through storageClientOptions.
 type storageConfig struct {
-	useJSONforReads                  bool
-	readAPIWasSet                    bool
-	disableClientMetrics             bool
-	DynamicReadReqStallTimeoutConfig *DynamicReadReqStallTimeoutConfig
+	useJSONforReads        bool
+	readAPIWasSet          bool
+	disableClientMetrics   bool
+	ReadStallTimeoutConfig *experimental.ReadStallTimeoutConfig
```
tritone, line 76: This should be unexported probably?

storage/option.go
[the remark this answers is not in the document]
tritone: I think we can leave it as values rather than pointers, and set to default values if zero values are detected for either parameter. Zero would not be a valid value for either param.

tritone (approved):
LGTM!

PR #10959, opened 2024-10-07
feat(pubsub): add support for cloud storage ingestion topics

hongalex:
Adds support for Cloud Storage ingestion in topic create/update.

### DIFF
--- a/pubsub/topic.go
+++ b/pubsub/topic.go
@@ -44,6 +44,7 @@ import (
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/durationpb"
 	fmpb "google.golang.org/protobuf/types/known/fieldmaskpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 const (
@@ -350,6 +351,9 @@ type TopicConfigToUpdate struct {
 	// IngestionDataSourceSettings are settings for ingestion from a
 	// data source into this topic.
 	//
+	// When changing this value, the entire data source settings object must be applied,
+	// rather than just the differences.
+	//
 	// Use the zero value &IngestionDataSourceSettings{} to remove the ingestion settings from the topic.
 	IngestionDataSourceSettings *IngestionDataSourceSettings
 }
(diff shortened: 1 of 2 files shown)


pubsub/topic.go
```diff
@@ -509,6 +600,25 @@ func protoToIngestionDataSourceSettings(pbs *pb.IngestionDataSourceSettings) *In
...
+				Delimiter: *t.TextFormat.Delimiter,
+			}
+		case *pb.IngestionDataSourceSettings_CloudStorage_AvroFormat_:
+			format = &IngestionDataSourceCloudStorageAvroFormat{}
+		case *pb.IngestionDataSourceSettings_CloudStorage_PubsubAvroFormat:
+			format = &IngestionDataSourceCloudStoragePubSubAvroFormat{}
+		}
+		s.Source = &IngestionDataSourceCloudStorage{
+			State:                   CloudStorageIngestionState(cs.State),
+			Bucket:                  cs.Bucket,
```
michaelpri10, line 617: Is there a reason for just using `cs.Bucket` (and similarly elsewhere for this PR), instead of using getters like is done for Kinesis? I'm not too familiar with using protobufs with Go, so unsure if there's any difference between the two.

pubsub/topic_test.go
```diff
@@ -165,6 +165,65 @@ func TestTopic_IngestionKinesis(t *testing.T) {
...
+	}
+	config2, err := topic.Update(ctx, TopicConfigToUpdate{IngestionDataSourceSettings: settings})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !testutil.Equal(config2.IngestionDataSourceSettings, settings, opt) {
+		t.Errorf("\ngot  %+v\nwant %+v", config2.IngestionDataSourceSettings, settings)
+	}
+
+	// Clear schema settings.
```
michaelpri10, line 216: "Clear ingestion settings."

pubsub/topic.go
```diff
@@ -534,6 +644,48 @@ func (i *IngestionDataSourceSettings) toProto() *pb.IngestionDataSourceSettings
 				},
 			}
 		}
+		if cs, ok := out.(*IngestionDataSourceCloudStorage); ok {
+			switch format := cs.InputFormat.(type) {
+			case *IngestionDataSourceCloudStorageTextFormat:
```
michaelpri10, line 649: Could a similar method to what is used in `protoToIngestionDataSourceSettings` be used for this function, so that you create an InputFormat proto object within the switch statement and then declare the full proto once at the end? There may be some proto limitations that I'm unaware of that make this difficult/impossible.

pubsub/topic.go
hongalex: Accessing via getters is better, will switch over.

pubsub/topic.go
hongalex: Yeah this looks messy/redundant because there is no single "InputFormat" proto object that I can assign to, since it's not a public interface but [private](https://github.com/googleapis/google-cloud-go/blob/3377a3c709b66e14ff02e564014fedc231f5d25a/pubsub/apiv1/pubsubpb/pubsub.pb.go#L4450).

I could define a custom interface, but given that we're moving to GAPIC admin clients anyway, I think this is an ok compromise.

pubsub/topic_test.go
hongalex: Fixed

michaelpri10 (approved):

michaelpri10:
LGTM

michaelpri10 (approved):

PR #10966, opened 2024-10-08
feat(storage/dataflux): run worksteal listing parallel to sequential listing

akansha1812:
feat: run worksteal listing parallel to sequential listing
Dataflux fast-listing leverages worksteal algorithm to quickly list objects in a bucket by running several parallel processes. Worksteal algorithm splits a given namespace into multiple ranges for multiple workers(goroutines) to list objects in gcs bucket in parallel. 

Adding logic to list next page in worksteal listing and basic unit tests that uses storage/emulator.


storage/dataflux/workstea_test.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2024 Google LLC
```
BrennaEpp, line 1: nit: typo in file name

storage/dataflux/workstea_test.go
```diff
@@ -0,0 +1,55 @@
...
+			Parallelism: 3,
+		}
+		c := NewLister(client, in)
+		c.method = worksteal
+		objects, err := c.workstealListing(ctx)
+		if err != nil {
+			t.Fatalf("failed to call workstealListing() : %v", err)
+		}
+		if len(objects) != numObjects {
+			t.Errorf("doSeqListing() expected to receive  %d results, got %d results", numObjects, len(objects))
```
BrennaEpp, line 52: nit: error references sequential listing

storage/dataflux/workstea_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"cloud.google.com/go/storage"
+)
+
+func TestWorkstealListingEmulated(t *testing.T) {
+	transportClientTest(context.Background(), t, func(t *testing.T, ctx context.Context, project, bucket string, client *storage.Client) {
+
+		attrs := &storage.BucketAttrs{
+			Name: bucket,
+			Logging: &storage.BucketLogging{
+				LogBucket: bucket,
```
BrennaEpp, line 30: Why are we setting a LogBucket here?

storage/dataflux/fast_list_test.go
```diff
@@ -192,3 +197,91 @@ func TestNewLister(t *testing.T) {
...
+}
+
+// isEmulatorEnvironmentSet checks if the emulator environment variables are set.
+func isEmulatorEnvironmentSet() bool {
+	return os.Getenv("STORAGE_EMULATOR_HOST_GRPC") != "" && os.Getenv("STORAGE_EMULATOR_HOST") != ""
+}
+
+// createObject creates an object in the emulator and returns its name, generation, and
+// metageneration.
+func createObject(ctx context.Context, bucket *storage.BucketHandle, numObjects int) error {
```
BrennaEpp, line 273: Looks like the function name and description are outdated

storage/dataflux/fast_list_test.go
```diff
@@ -192,3 +197,91 @@ func TestNewLister(t *testing.T) {
...
+
+	for i := 0; i < numObjects; i++ {
+		// Generate a unique object name using UUIDs
+		objectName := fmt.Sprintf("object%d", i)
```
BrennaEpp, line 277: May want to use the uuid library here; also, may be better to distribute the names (that is, not have a prefix, the object name would be just the uuid).

storage/dataflux/fast_list_test.go
```diff
@@ -192,3 +197,91 @@ func TestNewLister(t *testing.T) {
...
+		// Create a writer for the object
+		wc := bucket.Object(objectName).NewWriter(ctx)
```
BrennaEpp, line 279: nit: what does the C stand for in the var name `wc`?

storage/dataflux/sequential_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+func TestDoSeqListingEmulated(t *testing.T) {
+	transportClientTest(context.Background(), t, func(t *testing.T, ctx context.Context, project, bucket string, client *storage.Client) {
+
+		attrs := &storage.BucketAttrs{
+			Name: bucket,
+			Logging: &storage.BucketLogging{
+				LogBucket: bucket,
+			},
+			VersioningEnabled: true,
```
BrennaEpp, line 32: Same here, is there a reason for setting these?

storage/dataflux/sequential_test.go
```diff
@@ -0,0 +1,97 @@
...
+		objects, nextToken, err := c.sequentialListing(ctx)
+
+		if err != nil {
+			t.Fatalf("failed to call doSeqListing() : %v", err)
+		}
+		// Even if the batchSize is smaller , more objects are listed. Sequential listing
+		// lists an entire page (defaultPageSize) which is bigger than given batchSize.
+		if len(objects) != wantObjects {
+			t.Errorf("sequentialListing() expected to receive %d results, got %d results", len(objects), wantObjects)
+		}
```
BrennaEpp, line 92: How does batch size and default page size play into this code?

storage/dataflux/sequential.go
```diff
@@ -33,37 +33,37 @@ const (
...
+func doSeqListing(objectIterator *storage.ObjectIterator, skipDirectoryObjects bool) (result []*storage.ObjectAttrs, token string, pageSize int, err error) {
 
 	for {
 		attrs, errObjectIterator := objectIterator.Next()
-		objectsListed++
 		// Stop listing when all the requested objects have been listed.
 		if errObjectIterator == iterator.Done {
 			break
 		}
+		pageSize++
```
BrennaEpp, line 66: Nice catch! Shouldn't this go even later, after the err.. != nil check?

storage/dataflux/next_page.go
```diff
@@ -0,0 +1,170 @@
...
+		generation = items[itemIndex].Generation
+		items = items[:len(items)-1]
+	} else if indexLexLast >= 0 {
+		generation = items[indexLexLast].Generation
+		items = append(items[:indexLexLast], items[indexLexLast+1:]...)
+	}
+
+	// Check if is versions is false, generation is not required.
+	if !opts.query.Versions {
+		generation = int64(0)
```
BrennaEpp, line 154: Why is generation not required in that case?

storage/dataflux/next_page.go
```diff
@@ -0,0 +1,170 @@
...
+	// doneListing indicates whether the lister has completed listing.
+	doneListing bool
+	// nextStartRange is the start offset of the next page of objects to be listed.
+	nextStartRange string
+	// generation is the generation number of the last object in the page.
+	generation int64
+}
+
+// nextPage lists objects using the given lister options.
+func nextPage(ctx context.Context, opts nextPageOpts) (*nextPageResult, error) {
```
BrennaEpp, line 56: This method seems a little convoluted; do you think there is a way to simplify it or make it a little easier to read?

storage/dataflux/workstea_test.go
akansha1812: Removed it. Thanks Brenna for catching this!

storage/dataflux/next_page.go
akansha1812: Since version is disabled in query, only the latest version of the object will be fetch by gcs. So generation  is not required is this case.

storage/dataflux/sequential.go
akansha1812: Done!

storage/dataflux/sequential_test.go
akansha1812: It was an outdated comment.

storage/dataflux/sequential_test.go
akansha1812: No use. Removed it.

storage/dataflux/next_page.go
akansha1812: Update nextPage. Let me know if its more readable now.

storage/dataflux/fast_list.go
```diff
@@ -131,31 +131,53 @@ func (c *Lister) NextBatch(ctx context.Context) ([]*storage.ObjectAttrs, error)
...
 	// in parallel. The method which completes first is used for all subsequent runs.
 
-	// TODO: Run worksteal listing when method is Open or WorkSteal.
+	// Run worksteal listing when method is Open or WorkSteal.
+	if c.method != sequential {
+
+		g.Go(func() error {
+			objects, err := c.workstealListing(childCtx)
+			if err != nil {
+				countError++
```
BrennaEpp, line 154: Won't this be a race condition when c.method is open and both sequential and worksteal are running at the same time? Can we hit this case in a emulator test?

storage/dataflux/integration_test.go
```diff
@@ -115,22 +120,24 @@ func TestIntegration_NextBatch(t *testing.T) {
...
 		}
+		totalObjects += len(objects)
 	}
-	if totalObjects != wantObjects {
-		t.Errorf("expected to receive %d objects in results, got %d objects in results", wantObjects, totalObjects)
-
+	if totalObjects != numObjectsPrefix {
+		t.Errorf("expected to receive %d objects in results, got %d objects in results", numObjectsPrefix, totalObjects)
+	}
+	if counter <= 1 {
```
BrennaEpp, line 139: I think this condition will never be hit because the counter is updated at the start of the for loop.

storage/dataflux/fast_list.go
akansha1812: Added unit test to catch this in the race conditions.

storage/dataflux/integration_test.go
BrennaEpp: Sorry, it will be hit because it's greater than or equal to. So intstead I'll change this to a nit: easier to read this code if the counter updates after each loop rather than at the beginning

BrennaEpp (commented):
Looks like there is a test failing. Can you look into it?

storage/dataflux/fast_list.go
```diff
@@ -127,35 +132,58 @@ func NewLister(c *storage.Client, in *ListerInput) *Lister {
...
 	// Errgroup takes care of running both methods in parallel. As soon as one of
 	// the method is complete, the running method also stops.
 	g, childCtx := errgroup.WithContext(ctx)
 
 	// To start listing method is Open and runs both worksteal and sequential listing
 	// in parallel. The method which completes first is used for all subsequent runs.
 
-	// TODO: Run worksteal listing when method is Open or WorkSteal.
+	// Run worksteal listing when method is Open or WorkSteal.
+	if c.method != sequential {
```
BrennaEpp, line 155: I think this method would be easier to understand with a little repetition, using a switch on `c.method`. That way we also don't have to create child contexts + errgroups every time this runs, and the common/simple case (where we already know what kind of algorithm we're running) is easy to parse quickly.

Here's an outline of what I am thinking:

```go
	switch c.method {
	case sequential:
		objects, token, err := c.sequentialListing(ctx)
		if err != nil {
			return nil, fmt.Errorf("error in running sequential listing: %w", err)
		}
		c.pageToken = token
		
		// do checks and return here, or save objects in the results var to do the
		// checks at the end
		if c.pageToken == "" {
		   return objects, iterator.Done
	        }
		
		return objects, nil // or save results = objects
	case worksteal:
		objects, err := c.workstealListing(ctx)
		if err != nil {
			return nil, fmt.Errorf("error in running worksteal_lister: %w", err)
		}
		return objects, nil // add check too, or same as above, save results
	case open:
                // Do both here:
                // We can move all vars that are only relevant to this case here
                	ctx, cancel := context.WithCancel(ctx)
		defer cancel()
		wsCompletedfirst := false
		seqCompletedfirst := false
		var wsObjects []*storage.ObjectAttrs
		var seqObjects []*storage.ObjectAttrs
		var nextToken string
	
		g, childCtx := errgroup.WithContext(ctx)
		g.Go(func() error {
			objects, err := c.workstealListing(childCtx)
			// and so on...
		})

		g.Go(func() error {
			objects, token, err := c.sequentialListing(childCtx)
			// etc
		})

		// Close all functions if either sequential listing or worksteal listing is complete.
		err := g.Wait()
		
		// Now we only need to do these checks when the method type is open
		if err != nil && (!strings.Contains(err.Error(), context.Canceled.Error()) || cc.counter > 1) {
			return nil, fmt.Errorf("failed waiting for sequential and work steal lister : %w", err)
		}
		if wsCompletedfirst {
			//...
		} else if seqCompletedfirst {
			//...
		}
	}
```

storage/dataflux/fast_list.go
```diff
@@ -188,6 +230,12 @@ func (c *Lister) Close() {
 	}
 }
 
+func (cc *contextErr) increment() {
+	cc.mu.Lock()
+	defer cc.mu.Unlock()
+	cc.counter++
+}
+
 // updateStartEndOffset updates start and end offset based on prefix.
```
BrennaEpp, line 239: Reviewing this again, I think we should rename this function. It's returning new offsets, it's not updating them directly. It's also not really clear when or why to use this in the name nor in the comment.

What it seems to be doing is giving prefix-removed offsets. So it removes the prefix or if the offset is outside of the prefix range it returns empty to indicate the prefix is sufficient to ensure we stay within the offset, or if it disallows the prefix, it returns the same start and end offset which will not list any objects. Let me know if this understanding is incorrect.

So really, what it is doing is returning adjusted offsets that already account for the prefix, so that the original offsets can essentially be ignored. Maybe we can rename this to `prefixAdjustedOffsets`.
```go
// prefixAdjustedOffsets returns a start and end offset adjusted from the given offsets based on the prefix, stripping the prefix.
// These offsets can be used by adding back the prefix, so that the original offsets do not need to be checked.
// This means that if the given offsets are out of range of the prefix (for example, offsets {start:"a", end: "b"}, with prefix "c" which is lexicographically outside of "a" to "b"), the returned offsets will ensure no strings fall in their range. 
// Otherwise, if the offset is too permissive given the prefix, it returns an empty string to indicate there is no offset and all objects starting from or ending at the prefix should be listed. Here the example you already gave is perfect to add.
// Otherwise, it just strips the prefix from the offset as shown by the end offset in the example above.
```

You can re-work what I wrote above a bit, but I think that should capture the gist of what's happening, when to use, and what it's doing - it's unfortunately not easy to sum up quickly.

storage/dataflux/fast_list.go
```diff
@@ -127,35 +132,58 @@ func NewLister(c *storage.Client, in *ListerInput) *Lister {
 // worksteal listing is expected to be faster.
 func (c *Lister) NextBatch(ctx context.Context) ([]*storage.ObjectAttrs, error) {
 	// countError tracks the number of failed listing methods.
-	countError := 0
+	cc := &contextErr{counter: 0}
```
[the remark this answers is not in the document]
akansha1812: We are using context cancelation as one of the listing method completes. So after first run there will be a context cancelled error. If both the methods fail, only then we return error.
I have added this where we check countError so future maintainers know what it is used for.

storage/dataflux/fast_list.go
```diff
@@ -92,6 +92,11 @@ type Lister struct {
 	skipDirectoryObjects bool
 }
 
+type contextErr struct {
```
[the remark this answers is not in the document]
akansha1812: Its keeping a count of context error. I'll change the name.

storage/dataflux/fast_list.go
akansha1812: This sounds correct. Updated it.

storage/dataflux/fast_list.go
```diff
@@ -127,35 +132,58 @@ func NewLister(c *storage.Client, in *ListerInput) *Lister {
 // worksteal listing is expected to be faster.
```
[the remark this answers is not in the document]
akansha1812: 1. Done.
2. Incase if failure, it immediately returns the error and same lister shouldn't be used. We can look into this as a future improvement. Updated the comments to show this. 
3. Worksteal algorithm is explained in in worksteal.go and sequential in sequential.go. 
4. Updated it to show both methods run only for the first run.

storage/dataflux/fast_list.go
akansha1812: Done.

storage/dataflux/next_page.go
```diff
@@ -0,0 +1,184 @@
+// Copyright 2024 Google LLC
```
BrennaEpp, line 1: Can you indicate in the file name that this is next page only for worksteal listing (`worksteal_next_page` maybe)?

storage/dataflux/next_page.go
```diff
@@ -0,0 +1,184 @@
...
+	nextStartRange string
+	// generation is the generation number of the last object in the page.
+	generation int64
+}
+
+// nextPage retrieves a single page of objects from GCS using the provided
+// listing options (nextPageOpts). It returns a nextPageResult containing the
+// list of objects, a flag indicating if the listing is complete, the starting
+// point for the next page, and the generation of the last object in the page.
+// In case multiple versions of objects needs to be listed, then
```
BrennaEpp, line 64: This comment was cut off?

storage/dataflux/next_page.go
```diff
@@ -0,0 +1,184 @@
...
+				doneListing:    true,
+				nextStartRange: "",
+				generation:     int64(0),
+			}, nil
+		} else if err != nil {
+			return nil, fmt.Errorf("iterating through objects: %w", err)
+		}
+
+		// Skip object versions already processed in the previous page to prevent duplicates.
+		if opts.query.Versions && opts.query.StartOffset == attrs.Name && attrs.Generation < opts.generation {
```
BrennaEpp, line 99: Let's make sure we test versions against prod.  

Why are we checking if `attrs.Generation < opts.generation`?  I don't think we can assume objects listed will be listed in increasing generation order.

storage/dataflux/next_page.go
```diff
@@ -0,0 +1,184 @@
...
+// nextPageResult represents the results of fetching a single page of objects
+// from a GCS listing operation and information for remaining objects to be listed.
+type nextPageResult struct {
+	// items is the list of objects listed.
+	items []*storage.ObjectAttrs
+	// doneListing indicates whether the lister has completed listing.
+	doneListing bool
+	// nextStartRange is the start offset of the next page of objects to be listed.
+	nextStartRange string
+	// generation is the generation number of the last object in the page.
```
BrennaEpp, line 56: This is the generation number of the last object in the previous page, right?

storage/dataflux/next_page.go
```diff
@@ -0,0 +1,184 @@
...
+	// thus we need to manually loop to construct a page of results using the iterator.
+	for i := 0; i < wsDefaultPageSize; i++ {
+		attrs, err := objectIterator.Next()
+
+		// If the lister has listed the last item for the assigned range,
+		// then set doneListing to true and return.
+		if err == iterator.Done {
+			return &nextPageResult{
+				items:          items,
+				doneListing:    true,
```
BrennaEpp, line 90: I'm confused. It looks like we are discarding results we already got in some cases - if we reach the end of this function without encountering iterator.Done, then we are returning without exhausting objects from `objectIterator`, which may still have items we can get without another call to GCS?

storage/dataflux/range_splitter.go
```diff
@@ -216,6 +216,7 @@ func (rs *rangeSplitter) addCharsToAlphabet(characters []rune) {
 		if _, exists := rs.alphabetMap[char]; !exists {
 			allAlphabet = append(allAlphabet, char)
 			newChars = true
+			rs.alphabetMap[char] = 0
```
BrennaEpp, line 219: I thought we previously removed this line, was it accidentally added back in?

storage/dataflux/sequential.go
```diff
@@ -24,42 +24,44 @@ import (
...
 		c.pageToken = nextToken
 	}
 	return result, lastToken, nil
 }
 
-func doSeqListing(objectIterator *storage.ObjectIterator, skipDirectoryObjects bool) (result []*storage.ObjectAttrs, token string, objectsListed int, err error) {
+// doSeqListing leverages the GCS API's pagination capabilities to
+// list [seqDefaultPageSize] numbe of objects, token to list next page of obhects and
+// number of objects iterated(even if not in results).
+func doSeqListing(objectIterator *storage.ObjectIterator, skipDirectoryObjects bool) (result []*storage.ObjectAttrs, token string, pageSize int, err error) {
```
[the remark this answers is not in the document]
akansha1812: - changed doSeqListing -> listNextPageSequentially
- result -> results
- With pageSize I want to show the objects iterated even if some of them were not added to the results. If skipDirectoryObjects is true, directory objects are skipped and not added to the results.

storage/dataflux/next_page.go
akansha1812: nextPageResult stores result of the current page listed.
items are the objects fetched from gcs. We list 1000 objects and discard the last one. The generation number of the last object(discarded) is used for next call.

storage/dataflux/next_page.go
akansha1812: In versions parameter it shows objects will be listed in order of increasing generation number.

https://cloud.google.com/storage/docs/json_api/v1/objects/list#parameters

storage/dataflux/range_splitter.go
akansha1812: I ran a few tests and without this line, range_splitter was returning unexpected results.

storage/dataflux/fast_list.go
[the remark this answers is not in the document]
akansha1812: Added brief description of worksteal algorithm.

storage/dataflux/sequential.go
```diff
@@ -24,42 +24,44 @@ import (
...
 	var result []*storage.ObjectAttrs
-	var objectsListed int
+	var objectsIterated int
 	var lastToken string
 	objectIterator := c.bucket.Objects(ctx, &c.query)
 	objectIterator.PageInfo().Token = c.pageToken
-	objectIterator.PageInfo().MaxSize = defaultPageSize
+	objectIterator.PageInfo().MaxSize = seqDefaultPageSize
 
 	for {
```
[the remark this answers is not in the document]
akansha1812: We want the list to be a multiple of number_of_objects_gcs_api_fetches  in single call to make sure all the objects for a given token is listed.  Also when we set MaxSize to 5000, when request is asking for ACLs only 2000 objects are returned. 
The reason we are making doSeqListing return just one page is that the api call can fetch either 2000 or 5000 objects and still fetch all objects linked to the given token.
Updated the comment. 

Let me know what you think here.

storage/dataflux/worksteal.go
```diff
@@ -94,8 +95,7 @@ func (c *Lister) workstealListing(ctx context.Context) ([]*storage.ObjectAttrs,
 		idleChannel <- 1
 		g.Go(func() error {
 			if err := idleWorker.doWorkstealListing(ctx); err != nil {
-				workerErr = append(workerErr, err)
-				return fmt.Errorf("listing worker ID %q: %w", i, err)
+				return fmt.Errorf("listing worker ID %d: %w", idleWorker.id, err)
```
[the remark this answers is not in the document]
akansha1812: Removed redundant error statements.

storage/dataflux/worksteal.go
```diff
@@ -154,15 +153,15 @@ func (w *worker) doWorkstealListing(ctx context.Context) error {
 
 		// If listing not complete and idle workers are available, split the range
 		// and give half of work to idle worker.
-		if len(w.idleChannel)-len(w.lister.ranges) > 0 && ctx.Err() == nil {
+		for len(w.idleChannel)-len(w.lister.ranges) > 0 && ctx.Err() == nil {
```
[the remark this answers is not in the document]
akansha1812: - The example looks perfect.  
-    The distribution of work affects the performance by large for real world data. There is a higher chance of next object names being closer to start of the range. And if its far, one of the passed split will get to it. It is much faster splitting work in 1/2, 1/4,1/8... manner.

- We don't want to split and add work to the queue if there is pending work in range channel and the idle workers are in sleep at the point. This will check number of workers that can pick the work and work that is already pending. Similar case would be when we call NextBatch for the second time, the range channel would be full due to remaining work in the previous run and all the workers may be idle too. This create more work when all the workers have not picked the work yet.

storage/dataflux/sequential.go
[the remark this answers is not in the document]
BrennaEpp: Ah I see, that makes sense.

storage/dataflux/sequential.go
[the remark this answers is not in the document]
BrennaEpp: > sending in batchSize as a minSize to the doSeqListing and stopping there

By sending in a minimum size we can check here 
```
   if objectIterator.PageInfo().Remaining() == 0 { // check if we have reached at least the amount of objects we want here
			...
	}
```
I don't think we have public documentation on the page size, and it seems like an internal detail we shouldn't rely on, in fact, I don't think GCS guarantees anywhere publicly that each page will be of equal size. Please correct me if you know otherwise. This code doesn't seem to be relying on that as is, so it's okay to leave as is, but can we more specifically say in the comment that `listNextPageSequentially` will make at most one network call to GCS and will exhaust all objects currently held in the iterator?

In this vein, can we also add to the docs for `ListerInput.SkipDirectoryObjects` that directory objects count in the `batchSize`? Otherwise users may assume that we are filtering them server-side. We probably should mention it in `ListerInput.BatchSize` as well.

Also, in `ListerInput.BatchSize`, I see it mentions `The number of objects returned will be rounded up to a multiple of gcs page size.`"gcs page size" could be confusing to users, especially since as mentioned above, there is no exact value for this concept. Instead, can we rephrase it, maybe:

```
	// BatchSize is the minimum number of objects to list in each batch. 
	// The number of objects returned in a batch will be rounded up to 
        // include all the objects received in the last request to GCS.
	// By default, the Lister returns all objects in one batch. 
	// Optional.
```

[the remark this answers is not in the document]
akansha1812: Done.

storage/dataflux/fast_list.go
```diff
@@ -92,116 +125,156 @@ type Lister struct {
...
+
+		// If the error is not context.Canceled, then return error instead of falling back
+		// to the other method. This is so that the error can be fixed and user can take
+		// advantage of fast-listing.
+		// As one of the listing method completes, it is expected to cancel context and
+		// return context canceled error for the other method. Since context canceled is expected, it
+		// will not be considered an error. If both sequential and worksteal listing fail due
+		// to context canceled, then return error.
+		if err != nil && (!errors.Is(err, context.Canceled) || countErr.counter > 1) {
+			return nil, fmt.Errorf("dataflux : %w", err)
```
BrennaEpp, line 214: nit
```suggestion
			return nil, fmt.Errorf("dataflux: %w", err)
```

storage/dataflux/worksteal.go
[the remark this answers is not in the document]
BrennaEpp: I see, that makes a lot more sense now. Thanks Akansha!

storage/dataflux/next_page.go
BrennaEpp: Ah, my bad, TIL. In that case this looks good 👍

storage/dataflux/next_page.go
BrennaEpp: So I think this is a case where we are relying on undocumented GCS behaviour - unless you know for sure that GCS fetches 1000 objects every time and that will never change? If not, this code should not assume that GCS will always fetch the same amount of objects.

storage/dataflux/range_splitter.go
BrennaEpp: What tests did you run? I ran the ones in `range_splitter_test` in this PR and they pass fine without this line.

BrennaEpp (commented):
Looking good, thanks for the changes

storage/dataflux/next_page.go
akansha1812: You are right, 100 objects is not documented. It was something that I noticed. 
Its not relying on GCS fetching 1000 objects, but then we might end up discarding some of the objects that are already fetched. I can set maxSize to 1000, that way it will fetch 1000 objects only and we won't end up discarding fetched objects. 
Thanks Brenna.

storage/dataflux/range_splitter.go
akansha1812: So if I remove this line, and `characters` contains same new alphabet multiple times,  then it will be appended to  `allAlphabet` multiple times as the check to append is `if exists := rs.alphabetMap[char]; !exists{}`. So we will have repeated characters in allAlphabet and alphabetMap will not be as expected. 

Added a unit to cover this.

BrennaEpp: This feels like the wrong place to check for this.
If it's an error for the same character to be added several times then we should return an error.

At the very least add a comment to this line please to indicate that it's there to avoid repeating the same character more than once.

Please fix the comment of the method too `// addCharsToAlphabet adds the given chars to the known alphabet. Repeated chars are ignored [or it's an error to repeat chars]`

akansha1812: The `alphabet` has only unique characters (similar to english alphabets have unique characters) and the main goal of this method is to add new characters to `alphabet`. 
Updated newRangeSplitter comment to show that alphabets are unique chars. 
Update comments to explain the reason behind adding the char to the map.

BrennaEpp (approved):
Thank you!

Issue #10962, opened 2024-10-08
auth: OpenTelemetry handling missing from grpctransport.go and httptransport/transport.go


quartzmo:
- [ ] The OpenTelemetry support added in https://github.com/googleapis/google-api-go-client/commit/46421d477a0b93e0bf5894fc9cc3599b06f9f85c to `/google-api-go-client/transport/grpc/dial.go` was never copied to [auth/grpctransport/grpctransport.go](https://github.com/googleapis/google-cloud-go/blob/auth/v0.9.7/auth/grpctransport/grpctransport.go#L294-L297). It should be placed next to the call to `addOCStatsHandler`.
- [ ] The OpenTelemetry support added in https://github.com/googleapis/google-api-go-client/commit/46421d477a0b93e0bf5894fc9cc3599b06f9f85c to `/google-api-go-client/transport/http/dial.go` was never copied to [auth/httptransport/transport.go](https://github.com/googleapis/google-cloud-go/blob/auth/v0.9.7/auth/httptransport/transport.go#L45). It should be placed next to the call to `addOCTransport`.

PR #10968, opened 2024-10-08
fix(auth): restore OpenTelemetry handling in transports

quartzmo:
* Based on https://github.com/googleapis/google-api-go-client/pull/2127

fixes: #10962


codyoss (approved):
Thanks!

PR #10969, opened 2024-10-09
feat(pubsub): add support for ingestion platform logging settings

hongalex:
Adds support for platform logging and clarifies that updating a topic to use ingestion requires setting this and source settings in its entirety each time.


pubsub/topic_test.go
```diff
@@ -122,6 +122,9 @@ func TestTopic_IngestionKinesis(t *testing.T) {
 				AWSRoleARN:        "fake-aws-role-arn",
 				GCPServiceAccount: "fake-gcp-sa",
 			},
+			PlatformLogsSettings: &PlatformLogsSettings{
```
michaelpri10, line 125: nit: Users are allowed to set Platform Logs settings on Kinesis ingestion topics but we don't actually do any platform logging for Kinesis ingestion. Fine with keeping it or removing it.

michaelpri10 (approved):

michaelpri10:
LGTM, thanks for quick turnaround!

PR #10979, opened 2024-10-09
fix(logging): Fixed input validation for X-Cloud-Trace-Context; encoded spanID from XCTC header into hex string.

### LINKED ISSUE #10910: logging: incorrect span ID logged when using "X-Cloud-Trace-Context"
For Googlers: motivating details / context can be found in Buganizer issue # 367762417

# Steps to reproduce

 1. Execute in an environment in which tracing context has been supplied via X-Cloud-Trace-Context.
 
# Observed

The LogEntry that is created contains a span_id that is incorrect.

Note that LogEntry.span_id should be the hexadecimal encoding of the span ID, whereas X-Cloud-Trace-Context (unlike traceparent) encodes the span_id in decimal format.

# Expected

The LogEntry.span_id field should be populated with the canonical (hexadecimal) representation of the span ID.

# Additional Notes

Likely fix involves changing `deconstructXCloudTraceContext` in `logging/logging.go` to convert the span ID to hex.

# Related Issues

There is a similar bug in the Python implementation. See:

https://github.com/googleapis/python-logging/issues/942

gkevinzheng:
Changes:

1. Link to format for `X-Cloud-Trace-Context` has moved, so that comment has been updated.
2. Encoded `spanID` in the return value of `deconstructXCloudTraceContext` into 16-bit hexadecimal string instead of decimal string.

Fixes #10910


bhshkh (approved):

logging/logging.go
```diff
@@ -884,15 +884,15 @@ var validXCloudTraceContext = regexp.MustCompile(
...
 func deconstructXCloudTraceContext(s string) (traceID, spanID string, traceSampled bool) {
-	// As per the format described at https://cloud.google.com/trace/docs/setup#force-trace
-	//    "X-Cloud-Trace-Context: TRACE_ID/SPAN_ID;o=TRACE_TRUE"
+	// As per the format described at https://cloud.google.com/trace/docs/trace-context#legacy-http-header
+	//    "X-Cloud-Trace-Context: TRACE_ID[/SPAN_ID][;o=TRACE_TRUE]"
 	// for example:
-	//    "X-Cloud-Trace-Context: 105445aa7843bc8bf206b120001000/1;o=1"
+	//    "X-Cloud-Trace-Context: 105445aa7843bc8bf206b12000100000/1;o=1"
 	//
 	// We expect:
```
michaelsafyan, line 892: It might be easier to understand if you phrase the comment as:

```
  // Parameters:
  //   s - ...
  //
  // Returns:
  //     ...
```

With respect to "We expect", it is unclear whether this is referring to the input or the output.

If it is intended to describe the input, then shouldn't the documentation for `spanID` say it is a decimal string?

logging/logging.go
```diff
@@ -903,6 +903,15 @@ func deconstructXCloudTraceContext(s string) (traceID, spanID string, traceSampl
...
+	if spanID != "" {
+		// Convert to 16 byte unsigned hex string
+		intSpanID, err := strconv.ParseUint(spanID, 10, 64)
+		if err != nil || intSpanID == 0 {
+			spanID = ""
+		} else {
+			spanID = fmt.Sprintf("%016x", intSpanID)
+		}
+	}
 	return
```
michaelsafyan, line 915: Optional readability suggestion: consider returning the values explicitly rather than relying on named return values. It was a little confusing seeing this return without the list of variables that are returned.

logging/logging.go
```diff
@@ -903,6 +903,15 @@ func deconstructXCloudTraceContext(s string) (traceID, spanID string, traceSampl
 		spanID = ""
```
michaelsafyan, line 903: You may want to consider similar validation/conversion on the trace ID (to ensure that it isn't all-zero and to ensure that it is zero-padded if smaller than the W3C spec length).

logging/logging.go
```diff
@@ -884,15 +884,15 @@ var validXCloudTraceContext = regexp.MustCompile(
...
+	//    "X-Cloud-Trace-Context: TRACE_ID[/SPAN_ID][;o=TRACE_TRUE]"
 	// for example:
-	//    "X-Cloud-Trace-Context: 105445aa7843bc8bf206b120001000/1;o=1"
+	//    "X-Cloud-Trace-Context: 105445aa7843bc8bf206b12000100000/1;o=1"
 	//
 	// We expect:
-	//   * traceID (optional): 			"105445aa7843bc8bf206b120001000"
-	//   * spanID (optional):       	"1"
-	//   * traceSampled (optional): 	true
+	//   * traceID (optional, 32-bit hex string):  "105445aa7843bc8bf206b12000100000"
```
michaelsafyan, line 893: The trace ID is 128-bit, not 32-bit; it is represented as a 16-byte (32-char) array. This comment and the next one may be confusing the number of symbols in the string with the number of bits in the numeric value.

https://www.w3.org/TR/trace-context/#trace-id

logging/logging_test.go
```diff
@@ -327,28 +327,56 @@ func TestToLogEntry(t *testing.T) {
...
+					},
+				},
+			},
+			want: &logpb.LogEntry{
+				Trace:        "projects/P/traces/105445aa7843bc8bf206b12000100000",
+				SpanId:       "0000000000000010",
 				TraceSampled: true,
 			},
+		}, {
+			name: "X-Trace-Context with Span ID too large",
```
michaelsafyan, line 355: Is there any logging or other debug information that can be output in such a case?

logging/logging.go
```diff
@@ -903,6 +903,15 @@ func deconstructXCloudTraceContext(s string) (traceID, spanID string, traceSampl
 		spanID = ""
 	}
 
+	if spanID != "" {
+		// Convert to 16 byte unsigned hex string
+		intSpanID, err := strconv.ParseUint(spanID, 10, 64)
+		if err != nil || intSpanID == 0 {
+			spanID = ""
```
michaelsafyan, line 910: Is it possible to leave some clues when this happens?

logging/logging.go
michaelsafyan, line 903: Is it possible to leave some clues when this happens?

michaelsafyan (commented):
Thanks for taking care of this! Per your request over Google Chat, I've left some feedback on this PR. I will defer, however, to your judgement and that of the other reviewer concerning what subset of this feedback to act on. Thanks!

PR #10983, opened 2024-10-11
chore(spanner): generate client_hash metric attribute using FNV

rahul2393:
Related PR: https://github.com/googleapis/java-spanner/pull/3357

### DIFF
--- a/spanner/metrics_test.go
+++ b/spanner/metrics_test.go
@@ -244,20 +244,23 @@ func TestGenerateClientHash(t *testing.T) {
 	tests := []struct {
 		name             string
 		clientUID        string
+		expectedValue    string
 		expectedLength   int
 		expectedMaxValue int64
 	}{
-		{"Simple UID", "exampleUID", 6, 0x3FF},
-		{"Empty UID", "", 6, 0x3FF},
-		{"Special Characters", "!@#$%^&*()", 6, 0x3FF},
-		{"Very Long UID", "aVeryLongUniqueIdentifierThatExceedsNormalLength", 6, 0x3FF},
-		{"Numeric UID", "1234567890", 6, 0x3FF},
+		{"Simple UID", "exampleUID", "00006b", 6, 0x3FF},
+		{"Empty UID", "", "000000", 6, 0x3FF},
+		{"Special Characters", "!@#$%^&*()", "000389", 6, 0x3FF},
+		{"Very Long UID", "aVeryLongUniqueIdentifierThatExceedsNormalLength", "000125", 6, 0x3FF},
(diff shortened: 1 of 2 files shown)


spanner/metrics_test.go
```diff
@@ -41,21 +42,19 @@ func TestNewBuiltinMetricsTracerFactory(t *testing.T) {
...
 	wantClientAttributes := []attribute.KeyValue{
-		attribute.String(monitoredResLabelKeyProject, project),
-		attribute.String(monitoredResLabelKeyInstance, instance),
+		attribute.String(monitoredResLabelKeyProject, "[PROJECT]"),
+		attribute.String(monitoredResLabelKeyInstance, "[INSTANCE]"),
 		attribute.String(metricLabelKeyDatabase, "[DATABASE]"),
 		attribute.String(metricLabelKeyClientUID, clientUID),
 		attribute.String(metricLabelKeyClientName, clientName),
-		attribute.String(monitoredResLabelKeyClientHash, "cloud_spanner_client_raw_metrics"),
+		attribute.String(monitoredResLabelKeyClientHash, "0000ed"),
```
olavloite, line 57: I find this a bit suspicious. Why is the hash for this random test client exactly `0000ed`? It can be coincidence, but have we verified that the hash function is effectively using the entire result space, and not always outputting hashes that start with `0000`?

spanner/metrics_test.go
```diff
@@ -235,3 +238,46 @@ func TestNewBuiltinMetricsTracerFactory(t *testing.T) {
...
 	}
 }
+
+// TestGenerateClientHash tests the generateClientHash function.
+func TestGenerateClientHash(t *testing.T) {
+	tests := []struct {
+		name             string
+		clientUID        string
+		expectedLength   int
+		expectedMaxValue int64
```
olavloite, line 248: Could we maybe add the expected hash value to this test? I know that it is an implementation detail, but seeing 5 different hash values here would make me more comfortable when it comes to knowing that the hash function actually generates a reasonable variety in output. It would also ensure that the hash function is stable.

spanner/metrics_test.go
rahul2393: It's not random, generateClientUID is mocked to return always `test-uid` for which client hash is `0000ed `

spanner/metrics_test.go
rahul2393: Added

spanner/metrics_test.go
olavloite: Sorry, I missed the part that it only uses the 10 most significant bits, which means that the generated hashes all do start with 000.

olavloite (approved):

PR #10984, opened 2024-10-11
feat(storage): add grpc metrics experimental options

frankyn:


storage/experimental/experimental.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"time"
+
+	"cloud.google.com/go/storage/internal"
+	"go.opentelemetry.io/otel/sdk/metric"
+	"google.golang.org/api/option"
+)
+
+// WithMetricInterval how often to emit metrics when using NewPeriodicReader
```
tritone, line 27: Make it clear that these options are supposed to be used with storage.NewClient.

tritone (commented):
Looks great overall, thanks for figuring this out Frank!

tritone:
Also, looks like your unit test fails... can you investigate?

storage/experimental/experimental.go
```diff
@@ -25,9 +25,27 @@ import (
 	"time"
 
 	"cloud.google.com/go/storage/internal"
+	"go.opentelemetry.io/otel/sdk/metric"
 	"google.golang.org/api/option"
 )
 
+// WithMetricInterval provides a [ClientOption] that may be passed to [storage.NewGrpcClient].
+// It sets how often to emit metrics when using NewPeriodicReader and must be
+// greater than 1 minute.
```
tritone, line 34: Why is this the minimum? Not sure why this requires validation, users might have their own systems with higher rate limits.

storage/experimental/experimental.go
```diff
@@ -25,9 +25,27 @@ import (
 	"time"
 
 	"cloud.google.com/go/storage/internal"
+	"go.opentelemetry.io/otel/sdk/metric"
 	"google.golang.org/api/option"
 )
 
+// WithMetricInterval provides a [ClientOption] that may be passed to [storage.NewGrpcClient].
```
tritone, line 32: typo: should be storage.NewGRPCClient (here and below)

storage/experimental/experimental.go
```diff
@@ -25,9 +25,27 @@ import (
...
+// It sets how often to emit metrics when using NewPeriodicReader and must be
+// greater than 1 minute.
+// https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#NewPeriodicReader
```
tritone, line 35: leave an empty line (with a `//` only) between links if you want them to not wrap in the godoc.

storage/experimental/experimental.go
```diff
@@ -25,9 +25,27 @@ import (
...
+// https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#WithInterval
+func WithMetricInterval(metricInterval time.Duration) option.ClientOption {
+	return internal.WithMetricInterval.(func(time.Duration) option.ClientOption)(metricInterval)
+}
+
+// WithMetricExporter provides a [ClientOption] that may be passed to [storage.NewGrpcClient].
+// Set an alternate client-side metric Exporter to emit metrics through.
+// Must implement interface metric.Exporter:
```
tritone, line 43: Just do [metric.Exporter] and the godoc should autolink.

storage/option.go
```diff
@@ -15,12 +15,14 @@
 package storage
 
 import (
+	"time"
```
tritone, line 18: Not sure why this got moved but it will probably break formatting presubmits

tritone (commented):
Couple minor nits/questions, overall looks good

storage/experimental/experimental.go
BrennaEpp, line 32: nit: what does ClientOption link to?

storage/experimental/experimental.go
frankyn: Good point; could pass along a value if it's non-zero.

storage/experimental/experimental.go
```diff
@@ -25,10 +25,25 @@ import (
...
 
 	"cloud.google.com/go/storage/internal"
+	"go.opentelemetry.io/otel/sdk/metric"
 	"google.golang.org/api/option"
 )
 
-// WithReadStallTimeout provides a [ClientOption] that may be passed to [storage.NewClient].
+// WithMetricInterval provides a [option.ClientOption] that may be passed to [storage.NewGRPCClient].
+// It sets how often to emit metrics [metric.WithInterval] when using
+// [metric.NewPeriodicReader]
```
tritone, line 34: I think removing the minimum setting is good. Maybe suggest time.Minute or higher as an interval in the godoc if exporting to cloud monitoring.

tritone (approved):
One more doc comment, otherwise LGTM

Jenniferab5 (commented):
Review

PR #10992, opened 2024-10-14
chore(storage): fix ReadStallTimeout test

tritone:
Fix this test to use the emulator test harness so it will run in the presubmit emulator run and not with the integration tests. Also reduce timings for quicker execution.


storage/client_test.go
```diff
@@ -1458,51 +1458,50 @@ func TestRetryDeadlineExceedeEmulated(t *testing.T) {
...
-			gen:    defaultGen,
-			offset: 0,
-			length: -1,
-		}, idempotent(true))
-		if err != nil {
-			t.Fatalf("NewRangeReader: %v", err)
-		}
-		defer r.Close()
+	// Plant stall at start for 1s.
+	instructions := map[string][]string{"storage.objects.get": {"stall-for-1s-after-0K"}}
```
raj-prince, line 1488: We should set the delay to longer than 5 seconds. If a retry occurs, the process will work as expected; otherwise, it will fail with a deadline exceeded error.

Also, I have fixed a bug in the storage-test-bench. We need to take this change before this PR. P

Storage-test-bench fix: https://github.com/googleapis/storage-testbench/pull/690

raj-prince (approved):
LGTM. Added a comment.

storage/client_test.go
tritone: Good catch. I released a new emulator version and updated the test to have a longer stall.

frankyn (approved):

PR #10994, opened 2024-10-15
test(storage): fix ReadStallTimeout test

tritone:
Ensure that this is skipped in the continuous run and only run with the emulator.


frankyn (approved):

tritone:
I'm going to merge this to fix the continuous build, but the test is still being skipped in the presubmit and I am still working on figuring out how to make it succeed with the emulator.

PR #11021, opened 2024-10-23
fix: Fix default service account tests on GCE.

cjc25:
Tests running on GCE VMs automatically fall back to using the local metadata service to detect the local default service account. In those cases, tests for malformed creds won't fail, but they should return a VM-specific default service account email.


tritone (approved):
Awesome, thanks for fixing this!

PR #11032, opened 2024-10-24
perf(bigtable): reduce latency

bhshkh:
Notice the timestamps in the logs.
I couldn't simulate a noticeable server side delay while sending trailers but microseconds delay can be observed.

```
# Below is printed when response is received by the end user
024-10-24 22:22:20.996658405 +0000 UTC m=+5.020035733  Successfully read row with key  row-00004

# Below is printed when client library records metrics
2024-10-24 22:22:20.996639015 +0000 UTC m=+5.020016332  recordAttemptCompletion Recording attempt_latencies

# Below is printed when client library records metrics
2024-10-24 22:22:20.996805456 +0000 UTC m=+5.020182782  recordAttemptCompletion Recording server_latencies

# Below is printed when client library records metrics
2024-10-24 22:22:20.996929846 +0000 UTC m=+5.020307163  recordOperationCompletion Recording operation_latencies

# Below is printed when client library records metrics
2024-10-24 22:22:20.997028566 +0000 UTC m=+5.020405893  recordOperationCompletion Recording retry_count

```


bhshkh:
Will reopen when this work is actively picked up

PR #11039, opened 2024-10-28
fix(spanner): attempt latency for streaming call should capture the total latency till decoding of protos

rahul2393:


spanner/read.go
```diff
@@ -138,10 +142,32 @@ var _ rowIterator = (*RowIterator)(nil)
...
-	for len(r.rows) == 0 && r.streamd.next() {
+	// Start new attempt if we don't have one
+	if mt.currOp.currAttempt == nil {
+		mt.currOp.incrementAttemptCount()
+		mt.currOp.currAttempt = &attemptTracer{
+			startTime: time.Now(),
+		}
+	}
+	defer func() {
+		if mt.method != "" {
```
olavloite, line 157: This could use a comment. Does `mt.method != ""` mean that the operation completed? Why?

spanner/read.go
```diff
@@ -138,10 +142,32 @@ var _ rowIterator = (*RowIterator)(nil)
...
+			// record the attempt completion
+			if mt.currOp.currAttempt != nil {
```
olavloite, line 159: When is `mt.currOp.currAttempt == nil`? On line 152 above, mt.currOp.currAttempt is set to something that is not nil. Apparently, something else can set it back to nil before we get here. What is that?

spanner/read.go
```diff
@@ -138,10 +142,32 @@ var _ rowIterator = (*RowIterator)(nil)
...
+				mt.currOp.currAttempt.setStatus(status.Code(r.err).String())
+				recordAttemptCompletion(&mt)
```
olavloite, line 161: This seems to indicate that we can have a situation where the operation completed, but the attempt did not complete (at least not at the same time). What is that situation?

spanner/read.go
```diff
@@ -138,10 +142,32 @@ var _ rowIterator = (*RowIterator)(nil)
...
+			}
+			statusCode, _ := convertToGrpcStatusErr(r.err)
+			mt.currOp.setStatus(statusCode.String())
+			recordOperationCompletion(&mt)
```
olavloite, line 165: This seems to indicate that we will record the latency of the last attempt that finished as the operation latency. That means that an operation that is retried will not contain the entire latency, but only the latency of the last attempt. Is that intentional? If so, could we document that in code to clarify?

spanner/read.go
```diff
@@ -503,32 +526,30 @@ var (
...
 				d.changeState(aborted)
 				continue
 			}
 			trace.TracePrintf(d.ctx, nil, "Backing off stream read for %s", delay)
 			if err := gax.Sleep(d.ctx, delay); err == nil {
+				// record the attempt completion
+				mt.currOp.currAttempt.setStatus(status.Code(d.err).String())
+				recordAttemptCompletion(mt)
+				mt.currOp.incrementAttemptCount()
+				mt.currOp.currAttempt = &attemptTracer{}
```
olavloite, line 552: Shouldn't this be:

```go
		mt.currOp.currAttempt = &attemptTracer{
			startTime: time.Now(),
		}
```

spanner/read.go
```diff
@@ -629,7 +654,11 @@ func (d *resumableStreamDecoder) tryRecv(retryer gax.Retryer) {
 			d.changeState(aborted)
 			return
 		}
+		mt.currOp.incrementAttemptCount()
+		mt.currOp.currAttempt = &attemptTracer{}
```
olavloite, line 658: Here also: should we not set the start time of the attempt?

spanner/read.go
```diff
@@ -640,6 +669,8 @@ func (d *resumableStreamDecoder) tryRecv(retryer gax.Retryer) {
...
```
olavloite, line 673: And here: should we not set the attempt start time?

olavloite (commented):
Do we have any way to test this? There are a lot of corner cases where we are retrying the query stream etc., and it is very hard to tell from just looking at the code that we actually cover all corner cases, and that it actually records the latency that we want it to record.

spanner/read.go
rahul2393: Actually the currOp start time is initalised in first line of Next() and in recordOperationCompletion we capture to total time including multiple attempts latencies

spanner/read.go
rahul2393: Added a comment

spanner/read.go
olavloite: Ack. I see now that there is a separate field for the operation start time.

olavloite (approved):

PR #11048, opened 2024-10-29
feat(spanner): implement generation and propagation of "x-goog-spanner-request-id" Header

### LINKED ISSUE #11073: spanner: propagate x-goog-spanner-request-id header on every call and increment it appropriately per retry
This feature was requested for implementation by the Google Cloud Spanner internal engineering teams, to aid in much better debugging consistently and non-sporadically without sampling, we should send over a request header "x-goog-spanner-request-id".
Trying to use the OpenTelemetry TraceID is a non-starter because firstly traces are usually sampled and very expensive to always have on given that sampled tracing generates spans with more fields and exporting those to a backend can consume bandwidth and high costs for customers; also trying to retrieve a traceID retroactively is almost impossible and very difficult. Instead, the Cloud Spanner internal engineering teams are ready to accept x-goog-spanner-request-id that won't incur any extra costs for customers and can directly allow correlation of RPCs and help quickly debug.

## Structure
`<process_id>.<client_id>.<request_number>.<channel_id>.<rpc_number>`
where
* process_id: a randomly generated 32-bit unsigned integer, created at startup time once per process and shared across all Spanner clients
* client_id: a monotonically increasing (within process) value of clients created within a process
* request_number: the nth request per grpc_spanner_client
* channel_id: the gRPC channel ID
* rpc_number: within retries per method; rpc_number is monotonically increasing if request_number is the same but otherwise starts from 1 for every other independent call

## Task expectations
- [x] Generating x-goog-spanner-request-id according to the rubric above
- [x] Propagating x-goog-spanner-request-id in every call and every retry accordingly and tested
- [x] On any error returned from an RPC, augment that error with the value of x-spanner-request-id so that customers can report it to Google Customer
- [ ] Benchmarking to show the implications of this header being generated for client libraries aka latency impact and RAM impact to generate the ID and have it plumbed in these libraries
- [x] For performance benefits, make sure that we use the most efficient lock-free structures and mechanisms

I have already began on this task with a working sample with a bunch of tests at PR #11048 

Kindly /cc-ing @tharoldD @willpoint @olavloite

odeke-em:
In tandem this specification

https://orijtech.notion.site/x-goog-spanner-request-id-always-on-gRPC-header-to-aid-in-quick-debugging-of-errors-14aba6bc91348091a58fca7a505c9827

This change adds sending over the "x-goog-spanner-request-id" header
for every unary and streaming call, in the form:

    <version>.<processId>.<clientId>.<channelId>.<requestCountForClient>.<rpcCountForRequest>

where:
* version is the version of the specification
* processId is a randomly generated uint64 singleton for the lifetime of a process
* clientId is the monotonically increasing id/number of gRPC Spanner clients created
* requestCountForClient is the monotonically increasing number of requests made by the client
* channelId currently at 1 is the Id of the client for Go
* rpcCountForRequest is the number of RPCs/retries within a specific request

This header is to be sent on both unary and streaming calls and it'll
help debug latencies for customers. On an error, customers can assert against
.Error and retrieve the associated .RequestID and log it, or even better
it'll be printed out whenever errors are logged.

Importantly making randIdForProcess to be a uint6 which is 64bits and not
a UUID4 which is 128bits which surely massively reduces the possibility of collisions
to ensure that high QPS applications can function and accept bursts of traffic
without worry, as the prior design used uint32 aka 32 bits for
which just 50,000 new processes being created could get the probability
of collisions to 25%, with this new change a company would have to
create 82 million QPS every second for 1,000 years for a 1% collision
with 2.6e18 for which the collision would be 1%.
Using 64-bits still provides really good protection whereby for a 1% chance of collision,
we would need 810 million objects, so we have good protection.
However, Google Cloud Spanner's backend has to store every one of the always on
headers for a desired retention period hence 64-bits is a great balance between collision
protection vs storage.

Fixes #11073

### DIFF
--- a/spanner/errors.go
+++ b/spanner/errors.go
@@ -204,9 +204,6 @@ func toSpannerErrorWithCommitInfo(err error, errorDuringCommit bool) error {
 
 // ErrCode extracts the canonical error code from a Go error.
 func ErrCode(err error) codes.Code {
-	if err == nil {
-		return codes.OK
-	}
 	s, ok := status.FromError(err)
 	if !ok {
 		return codes.Unknown
--- a/spanner/errors.go
+++ b/spanner/errors.go
@@ -89,13 +89,17 @@ func (e *Error) Error() string {
 		return "spanner: OK"
 	}
 	code := ErrCode(e)
+
+	var s string
 	if e.additionalInformation == "" {
-		return fmt.Sprintf("spanner: code = %q, desc = %q", code, e.Desc)
+		s = fmt.Sprintf("spanner: code = %q, desc = %q", code, e.Desc)
+	} else {
+		s = fmt.Sprintf("spanner: code = %q, desc = %q, additional information = %s", code, e.Desc, e.additionalInformation)
 	}
-	if e.RequestID == "" {
-		return fmt.Sprintf("spanner: code = %q, desc = %q, additional information = %s", code, e.Desc, e.additionalInformation)
+	if e.RequestID != "" {
+		s = fmt.Sprintf("%s, requestID = %q", s, e.RequestID)
 	}
-	return fmt.Sprintf("spanner: code = %q, desc = %q, additional information = %s; requestID = %q", code, e.Desc, e.additionalInformation, e.RequestID)
+	return s
 }
 
 // Unwrap returns the wrapped error (if any).
@@ -202,9 +206,6 @@ func toSpannerErrorWithCommitInfo(err error, errorDuringCommit bool) error {
 func ErrCode(err error) codes.Code {
 	if err == nil {
 		return codes.OK
-	}
-	if spe, ok := err.(*Error); ok {
-		return spe.Code
 	}
 	s, ok := status.FromError(err)
 	if !ok {
--- a/spanner/errors.go
+++ b/spanner/errors.go
@@ -58,6 +58,10 @@ type Error struct {
 	// additionalInformation optionally contains any additional information
 	// about the error.
 	additionalInformation string
+
+	// RequestID is the associated ID that was sent to Google Cloud Spanner's
+	// backend, as the value in the "x-goog-spanner-request-id" gRPC header.
+	RequestID string
 }
 
 // TransactionOutcomeUnknownError is wrapped in a Spanner error when the error
@@ -88,7 +92,10 @@ func (e *Error) Error() string {
 	if e.additionalInformation == "" {
 		return fmt.Sprintf("spanner: code = %q, desc = %q", code, e.Desc)
 	}
-	return fmt.Sprintf("spanner: code = %q, desc = %q, additional information = %s", code, e.Desc, e.additionalInformation)
+	if e.RequestID == "" {
+		return fmt.Sprintf("spanner: code = %q, desc = %q, additional information = %s", code, e.Desc, e.additionalInformation)
+	}
+	return fmt.Sprintf("spanner: code = %q, desc = %q, additional information = %s; requestID = %q", code, e.Desc, e.additionalInformation, e.RequestID)
 }
 
 // Unwrap returns the wrapped error (if any).
@@ -123,6 +130,10 @@ func (e *Error) decorate(info string) {
 // APIError error having given error code as its status.
 func spannerErrorf(code codes.Code, format string, args ...interface{}) error {
 	msg := fmt.Sprintf(format, args...)
+	return spannerError(code, msg)
+}
+
+func spannerError(code codes.Code, msg string) error {
 	wrapped, _ := apierror.FromError(status.Error(code, msg))
 	return &Error{
 		Code: code,
@@ -172,9 +183,9 @@ func toSpannerErrorWithCommitInfo(err error, errorDuringCommit bool) error {
 			desc = fmt.Sprintf("%s, %s", desc, transactionOutcomeUnknownMsg)
 			wrapped = &TransactionOutcomeUnknownError{err: wrapped}
 		}
-		return &Error{status.FromContextError(err).Code(), toAPIError(wrapped), desc, ""}
+		return &Error{status.FromContextError(err).Code(), toAPIError(wrapped), desc, "", ""}
 	case status.Code(err) == codes.Unknown:
-		return &Error{codes.Unknown, toAPIError(err), err.Error(), ""}
+		return &Error{codes.Unknown, toAPIError(err), err.Error(), "", ""}
 	default:
 		statusErr := status.Convert(err)
 		code, desc := statusErr.Code(), statusErr.Message()
@@ -183,12 +194,18 @@ func toSpannerErrorWithCommitInfo(err error, errorDuringCommit bool) error {
 			desc = fmt.Sprintf("%s, %s", desc, transactionOutcomeUnknownMsg)
 			wrapped = &TransactionOutcomeUnknownError{err: wrapped}
 		}
-		return &Error{code, toAPIError(wrapped), desc, ""}
+		return &Error{code, toAPIError(wrapped), desc, "", ""}
 	}
 }
 
 // ErrCode extracts the canonical error code from a Go error.
 func ErrCode(err error) codes.Code {
+	if err == nil {
+		return codes.OK
+	}
+	if spe, ok := err.(*Error); ok {
+		return spe.Code
+	}
 	s, ok := status.FromError(err)
 	if !ok {
 		return codes.Unknown
--- a/spanner/grpc_client.go
+++ b/spanner/grpc_client.go
@@ -75,7 +75,7 @@ type grpcSpannerClient struct {
 	channelID uint64
 	// id is derived from the SpannerClient.
 	id int
-	// nthRequest shall always be incremented on every fresh request.
+	// nthRequest is incremented for each new request (but not for retries of requests).
 	nthRequest *atomic.Uint32
 }
 
@@ -182,8 +182,8 @@ func (g *grpcSpannerClient) ExecuteSql(ctx context.Context, req *spannerpb.Execu
 }
 
 func (g *grpcSpannerClient) ExecuteStreamingSql(ctx context.Context, req *spannerpb.ExecuteSqlRequest, opts ...gax.CallOption) (spannerpb.Spanner_ExecuteStreamingSqlClient, error) {
-	// Note: Please don't add g.optsWithNextRequestID to inject x-goog-spanner-request-id
-	//as it is already manually added in when creating Stream iterators for ExecuteStreamingSql.
+	// Note: This method does not add g.optsWithNextRequestID to inject x-goog-spanner-request-id
+	// as it is already manually added when creating Stream iterators for ExecuteStreamingSql.
 	return g.raw.ExecuteStreamingSql(peer.NewContext(ctx, &peer.Peer{}), req, opts...)
 }
 
@@ -208,8 +208,8 @@ func (g *grpcSpannerClient) Read(ctx context.Context, req *spannerpb.ReadRequest
 }
 
 func (g *grpcSpannerClient) StreamingRead(ctx context.Context, req *spannerpb.ReadRequest, opts ...gax.CallOption) (spannerpb.Spanner_StreamingReadClient, error) {
-	// Note: Please don't add g.optsWithNextRequestID as it is already manually
-	// added in when creating Stream iterators for ExecuteStreamingSql.
+	// Note: This method does not add g.optsWithNextRequestID, as it is already
+	// manually added when creating Stream iterators for StreamingRead.
 	return g.raw.StreamingRead(peer.NewContext(ctx, &peer.Peer{}), req, opts...)
 }
 
--- a/spanner/grpc_client.go
+++ b/spanner/grpc_client.go
@@ -86,22 +86,10 @@ var (
 
 // newGRPCSpannerClient initializes a new spannerClient that uses the gRPC
 // Spanner API.
-func newGRPCSpannerClient(ctx context.Context, sc *sessionClient, opts ...option.ClientOption) (spannerClient, error) {
+func newGRPCSpannerClient(ctx context.Context, sc *sessionClient, channelID uint64, opts ...option.ClientOption) (spannerClient, error) {
 	raw, err := vkit.NewClient(ctx, opts...)
 	if err != nil {
 		return nil, err
-	}
-
-	// Retrieve the channelID for each spannerClient.
-	// It is assumed that this method is invoked
-	// under a lock already.
-	channelID, ok := sc.channelIDMap[raw]
-	if !ok {
-		if sc.channelIDMap == nil {
-			sc.channelIDMap = make(map[*vkit.Client]uint64)
-		}
-		channelID = uint64(len(sc.channelIDMap)) + 1
-		sc.channelIDMap[raw] = channelID
 	}
 
 	g := &grpcSpannerClient{raw: raw, metricsTracerFactory: sc.metricsTracerFactory}
--- a/spanner/grpc_client.go
+++ b/spanner/grpc_client.go
@@ -19,6 +19,7 @@ package spanner
 import (
 	"context"
 	"strings"
+	"sync/atomic"
 
 	vkit "cloud.google.com/go/spanner/apiv1"
 	"cloud.google.com/go/spanner/apiv1/spannerpb"
@@ -67,6 +68,15 @@ type spannerClient interface {
 type grpcSpannerClient struct {
 	raw                  *vkit.Client
 	metricsTracerFactory *builtinMetricsTracerFactory
+
+	// These fields are used to uniquely track x-goog-spanner-request-id where:
+	// raw(*vkit.Client) is the channel, and channelID is derived from the ordinal
+	// count of unique *vkit.Client as retrieved from the session pool.
+	channelID uint64
+	// id is derived from the SpannerClient.
+	id int
+	// nthRequest shall always be incremented on every fresh request.
+	nthRequest *atomic.Uint32
 }
 
 var (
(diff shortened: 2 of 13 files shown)


odeke-em:
Kindly cc-ing @olavloite @tharoldD @willpoint

odeke-em:
Kindly cc-ing you @harshachinta.

rahul2393:
@odeke-em can you please fix the vet and tests issues here?

spanner/grpc_client.go
```diff
@@ -65,10 +70,43 @@ type spannerClient interface {
 // grpcSpannerClient is the gRPC API implementation of the transport-agnostic
 // spannerClient interface.
 type grpcSpannerClient struct {
+	id                   uint64
 	raw                  *vkit.Client
 	metricsTracerFactory *builtinMetricsTracerFactory
+
+	// These fields are used to uniquely track x-goog-spanner-request-id
+	// grpc.ClientConn is presumed to be the channel, hence channelID
```
olavloite, line 78: The `raw *vkit.Client` is the channel, so in that sense, this could be said to be redundant. But that property does not have a simple number or other simple string representation, which means that it is probably better/easier to just use the channel pool index that was used to fetch the channel as the channel ID here. Which again means that this property is not redundant and should be assigned a value.

spanner/sessionclient.go
```diff
@@ -274,6 +274,8 @@ func (sc *sessionClient) executeBatchCreateSessions(client spannerClient, create
 			break
 		}
 		var mdForGFELatency metadata.MD
+		// Each invocation of client.BatchCreateSessions is not automatically retried
```
olavloite, line 277: Same question here as for PartitionQuery; I don't quite understand what you mean with 'not automatically retried' in these cases. Could you elaborate a bit on that?

spanner/transaction.go
```diff
@@ -785,22 +785,36 @@ func (t *ReadOnlyTransaction) begin(ctx context.Context) error {
...
 		t.setSessionEligibilityForLongRunning(sh)
 		sh.updateLastUseTime()
 		var md metadata.MD
-		res, err = sh.getClient().BeginTransaction(contextWithOutgoingMetadata(ctx, sh.getMetadata(), t.disableRouteToLeader), &sppb.BeginTransactionRequest{
+
+		client := sh.getClient()
+		// Firstly set the number of retries as the RPCID.
+		gcl, ok := client.(*grpcSpannerClient)
+		if ok {
+			gcl.setRPCID(nRPCs)
```
olavloite, line 804: This seems to confuse two different types of retries:
1. Single RPCs are retried if they receive a retryable error (e.g. `UNAVAILABLE`). This is handled by the gRPC libraries and transparent to the Spanner client. This attempt number should be based on the number of retries that the RPC is being retried by the gRPC libraries.
2. Read/write transactions are retried if Spanner returns an `ABORTED` error. That will cause the entire transaction to be retried. That should not affect the attempt number that is used for an RPC.

spanner/grpc_client.go
odeke-em: Yeah I revised the design entirely and with a much better and simpler functionality whereby each channelID is retrieved from the pool, each clientID is retrieved from the globally incremented value so this should no longer be a problem.

spanner/transaction.go
odeke-em: Yeah I went with an entirely and more direct approach.

spanner/request_id_header_test.go
```diff
@@ -0,0 +1,1278 @@
...
+	return streamer(ctx, desc, cc, method, opts...)
+}
+
+func newInterceptorTracker() *interceptorTracker {
+	return &interceptorTracker{
+		nUnaryClientCalls:  new(atomic.Uint64),
+		nStreamClientCalls: new(atomic.Uint64),
+	}
+}
+
```
olavloite, line 449: The tests that I would have expected and that are missing here are:

### Base Scenario
Verify that we see exactly the request IDs that we expect when we have a scenario that can be 100% predicted. That would be for example:
1. A Client that has a SessionPoolConfig with Min/Max: 1 (this prevents concurrent BatchCreateSessions calls from making things unpredictable)
2. The application waits until the session pool has been filled with 1 session.
3. The client then executes a read-only transaction that executes 1 query.

Then verify exactly the request-IDs that we see (so not only that they are monotonically increasing)

### Base Unary Retry
Same as above, but let one of the unary RPCs be retried by simulating an UNAVAILABLE error.

### Base Streaming Retry
Same as above, but let the streaming ExecuteStreamingSql be retried by simulating an UNAVAILABLE error.

### Session Pool Growth
I think that there's a bug in the current assignment of channel IDs, as the session pool creates a new vkit.Client when it grows the pool. That does not create new gRPC channels, but my impression is that this event leads to a new channel-ID being assigned to the new session(s).

spanner/errors.go
```diff
@@ -172,9 +183,9 @@ func toSpannerErrorWithCommitInfo(err error, errorDuringCommit bool) error {
 			desc = fmt.Sprintf("%s, %s", desc, transactionOutcomeUnknownMsg)
 			wrapped = &TransactionOutcomeUnknownError{err: wrapped}
 		}
-		return &Error{status.FromContextError(err).Code(), toAPIError(wrapped), desc, ""}
+		return &Error{status.FromContextError(err).Code(), toAPIError(wrapped), desc, "", ""}
 	case status.Code(err) == codes.Unknown:
-		return &Error{codes.Unknown, toAPIError(err), err.Error(), ""}
+		return &Error{codes.Unknown, toAPIError(err), err.Error(), "", ""}
```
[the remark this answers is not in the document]
odeke-em: At this stage we don't have the requestID as these are errors being transformed directly from the wire.

spanner/errors.go
```diff
@@ -183,12 +194,18 @@ func toSpannerErrorWithCommitInfo(err error, errorDuringCommit bool) error {
 			desc = fmt.Sprintf("%s, %s", desc, transactionOutcomeUnknownMsg)
 			wrapped = &TransactionOutcomeUnknownError{err: wrapped}
 		}
-		return &Error{code, toAPIError(wrapped), desc, ""}
+		return &Error{code, toAPIError(wrapped), desc, "", ""}
```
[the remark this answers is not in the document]
odeke-em: At this stage we don't yet have the requestID.

spanner/grpc_client.go
```diff
@@ -82,7 +92,22 @@ func newGRPCSpannerClient(ctx context.Context, sc *sessionClient, opts ...option
 		return nil, err
 	}
 
+	// Retrieve the channelID for each spannerClient.
+	// It is assumed that this method is invoked
+	// under a lock already.
+	channelID, ok := sc.channelIDMap[raw]
```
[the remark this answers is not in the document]
odeke-em: We do use it to retain the mapping of vkit.Client to channelIDs; we do need it because between multiple sessionClients there are different vkit.Client objects created from each sessionClient. Is my reasoning and understanding correct?

spanner/sessionclient.go
```diff
@@ -101,15 +106,19 @@ type sessionClient struct {
...
 // newSessionClient creates a session client to use for a database.
 func newSessionClient(connPool gtransport.ConnPool, database, userAgent string, sessionLabels map[string]string, databaseRole string, disableRouteToLeader bool, md metadata.MD, batchTimeout time.Duration, logger *log.Logger, callOptions *vkit.CallOptions) *sessionClient {
+	clientID, nthClient := cidGen.nextClientIDAndOrdinal(database)
 	return &sessionClient{
 		connPool:             connPool,
 		database:             database,
 		userAgent:            userAgent,
-		id:                   cidGen.nextID(database),
+		id:                   clientID,
+		nthClient:            nthClient,
```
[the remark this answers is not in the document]
odeke-em: Well, id is a mangled and concatenated combo of "client-{nthClient}"; it made much more sense to not have to parse it on every invocation and instead have it specialized.

spanner/request_id_header.go
```diff
@@ -0,0 +1,227 @@
...
+	err := rew.ClientStream.SendMsg(msg)
+	return rew.processFromOutgoingContext(err)
+}
+
+func (rew *requestIDErrWrappingClientStream) RecvMsg(msg any) error {
+	err := rew.ClientStream.RecvMsg(msg)
+	return rew.processFromOutgoingContext(err)
+}
+
+var _ grpc.ClientStream = (*requestIDErrWrappingClientStream)(nil)
```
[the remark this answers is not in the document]
odeke-em: Compile time type assertions are super useful to catch failing code immediately instead of deep debugging when things fails, they also aid in readability even more.

spanner/request_id_header_test.go
```diff
@@ -0,0 +1,1278 @@
...
+func ensureMonotonicityOfRequestIDs(requestIDs []*requestIDSegments) error {
+	if false {
+		blob, err := json.MarshalIndent(requestIDs, "", " ")
+		if err != nil {
+			return err
+		}
+		println(string(blob))
+	}
```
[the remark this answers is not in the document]
odeke-em: roger that, thank you, those were to debug.

spanner/request_id_header_test.go
```diff
@@ -0,0 +1,1278 @@
...
+	server, sc, tearDown := setupMockedTestServerWithConfigAndClientOptions(t, clientConfig, clientOpts)
+	t.Cleanup(tearDown)
+	defer sc.Close()
+
+	// First commit will fail, and the retry will begin a new transaction.
+	server.TestSpanner.PutExecutionTime(testutil.MethodCommitTransaction,
+		testutil.SimulatedExecutionTime{
+			Errors: []error{newAbortedErrorWithMinimalRetryDelay()},
+		})
```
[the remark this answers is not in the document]
odeke-em: So my strategy was to go through all the tests that I could find for Transaction and get them tested with the requestID and ensure that each unique scenario returned the appropriate results. I definitely do have down tests that check that the number of RPCs is increased on retries. Would you like me to delete this one?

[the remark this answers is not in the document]
olavloite: I don't mind this one being here. What I would like to see is the same type of test, but then with mocked errors with error code UNAVAILABLE for both unary and streaming RPCs, combined with verifications that the attempt number actually increases in for those RPCs when Gax retries them.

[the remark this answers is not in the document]
odeke-em: Hmm, thanks for raising the manual retry of ExecuteStreamingSql manually.

spanner/request_id_header_test.go
odeke-em: When I made updates to the change to infer the channelID directly from grpc.ClientConn from ConnPool, that eliminated that possibility but indeed, let me add a test to grow the session pool and test out channelID doling out.

olavloite:
@odeke-em I added a number of new tests in https://github.com/googleapis/google-cloud-go/commit/b7ce43e5c465db82f059aa295f2053d13245d56e

As far as I can see, there are two issues that show up in those tests:
1. The `Attempt` number does not appear to increase when `ExecuteStreamingSql` is being retried due to an `UNAVAILABLE` error. Instead, the RPC field is increased.
2. `Context deadline exceeded` errors do not appear to include the RequestID. These are some of the most important errors to have these request IDs, as when this error happens, we want to know for sure whether the request reached Spanner (and search for traces on the Spanner servers).

odeke-em:
Awesome and thank you @olavloite for the tests! Yeah you did mention to me that ExecuteStreamingSql is manually retried. Yeah context related errors aren't wrapped up because they are sentinel errors essentially which trigger stopping of iterations and historically are compared to directly with `err == context.Deadline` and also for grpc.Codes.Deadline and all. But let me take a look more and figure out the appropriate wrapping. I am firstly going to tackle the manual ExecuteStreamingSql.

odeke-em:
@olavloite I have added the fix for the manual retries and those tests pass now for ExecuteStreamingSql all good except one small test but kindly help me run the bots.

odeke-em:
@olavloite kindly help me with the re-run, sorry I had to add a test assertion for an existent .RequestID :-)

olavloite:
The last build failed with:

```
=== NAME  TestRequestIDHeader_SingleUseReadOnly_ExecuteStreamingSql_ContextDeadlineExceeded
    request_id_header_test.go:1628: missing RequestID on error
--- FAIL: TestRequestIDHeader_SingleUseReadOnly_ExecuteStreamingSql_ContextDeadlineExceeded (0.05s)
```

odeke-em:
Thanks @olavloite, let me quickly take a look at it and see the fix.

odeke-em:
Hmm, I just ran it locally and no failures
```shell
$ go test -run=TestRequestIDHeader_SingleUseReadOnly_ExecuteStreamingSql_ContextDeadlineExceeded -short
2024/12/16 01:32:16 Integration tests skipped in -short mode.
PASS
ok  	cloud.google.com/go/spanner	1.096s
Emmanuels-MacBook-Pro-2:spanner emmanuelodeke$ go test -run=. -short
2024/12/16 01:33:04 Integration tests skipped in -short mode.
PASS
ok  	cloud.google.com/go/spanner	3.902s
```

I have just pushed up the rebased version, kindly help me re-run kokoro bots @olavloite

odeke-em:
@olavloite all tests pass, kindly help me take a look again.

spanner/read.go
```diff
@@ -533,11 +539,17 @@ var (
 
 func (d *resumableStreamDecoder) next(mt *builtinMetricsTracer) bool {
 	retryer := onCodes(d.backoff, codes.Unavailable, codes.ResourceExhausted, codes.Internal)
+
+	// Setup and track x-goog-request-id in the manual retries for ExecuteStreamingSql.
+	attempt := uint32(0)
+	riw := d.gsc.generateRequestIDHeaderInjector()
```
olavloite, line 545: This should only be initialized once per stream, and not once for each call to `next`. A stream can break halfway, and will be retried from the point where it broke. Now, that will cause a new request number to be generated, instead of registering it as a retry.

See the test in https://github.com/googleapis/google-cloud-go/commit/b5a70e4104fd862ab42eba45ce43f2c3cc7fb5d8

spanner/request_id_header.go
```diff
@@ -0,0 +1,265 @@
...
+// into every RPC, and it appropriately increments the RPC's ordinal number per retry.
+type retryerWithRequestID struct {
+	gsc *grpcSpannerClient
+}
+
+var _ gax.CallOption = (*retryerWithRequestID)(nil)
+
+func (g *grpcSpannerClient) appendRequestIDToGRPCOptions(priors []grpc.CallOption, nthRequest, attempt uint32) []grpc.CallOption {
+	// Google Engineering has requested that each value be added in Decimal unpadded.
+	// Should we have a standardized endianness: Little Endian or Big Endian?
```
olavloite, line 79:
```suggestion
	// Each value should be added in Decimal unpadded.
```

spanner/request_id_header.go
```diff
@@ -0,0 +1,265 @@
...
+	return append(priors, grpc.Header(&md))
+}
+
+type requestID string
+
+// augmentErrorWithRequestID introspects error converting it to an *.Error and
+// attaching the subject requestID, unless it is one of the following:
+// * nil
+// * context.Canceled
+// * context.DeadlineExceeded
```
olavloite, line 91: Remove `context.DeadlineExceeded` from this list

spanner/request_id_header.go
```diff
@@ -0,0 +1,265 @@
...
+// * io.EOF
+// * iterator.Done
+// of which in this case, the original error will be attached as it, since those
```
olavloite, line 94: nit:

```suggestion
// of which in this case, the original error will be attached as is, since those
```

spanner/request_id_header.go
```diff
@@ -0,0 +1,265 @@
...
+	cs.Retry = newRetryer
+}
+
+func (wr *retryerWithRequestID) generateAndInsertRequestID(md *metadata.MD, nthRequest, attempt uint32) {
+	wr.gsc.generateAndInsertRequestID(md, nthRequest, attempt)
+}
+
+func (gsc *grpcSpannerClient) generateAndInsertRequestID(md *metadata.MD, nthRequest, attempt uint32) {
+	// Google Engineering has requested that each value be added in Decimal unpadded.
+	// Should we have a standardized endianness: Little Endian or Big Endian?
```
olavloite, line 229:
```suggestion
        // Each value should be added in Decimal unpadded.
```

spanner/sessionclient.go
```diff
@@ -396,14 +412,30 @@ func (sc *sessionClient) sessionWithID(id string) (*session, error) {
 // optimal usage of server side caches.
 func (sc *sessionClient) nextClient() (spannerClient, error) {
 	var clientOpt option.ClientOption
+	var channelID uint64
 	if _, ok := sc.connPool.(*gmeWrapper); ok {
 		// Pass GCPMultiEndpoint as a pool.
 		clientOpt = gtransport.WithConnPool(sc.connPool)
```
olavloite, line 418: I don't know whether anything has been agreed upon regarding channel number when using GCPMultiEndpoint, but in the current implementation, the channel number will always be zero. Is that intentional?

spanner/read.go
odeke-em: Awesome and thank you @olavloite. I added your test and also sent the update so we now get the expected retries and attempt sequences.

spanner/request_id_header.go
odeke-em: Done, thank you!

spanner/request_id_header.go
odeke-em: Done, thank you!

spanner/request_id_header.go
odeke-em: Done, thank you!

spanner/sessionclient.go
odeke-em: Nothing indeed has yet been agreed for GCPMultiEndpoint. I was thinking that that wouldn't be commonly used. Is it a show stopper for the initial code in?

olavloite (approved):

PR #11051, opened 2024-10-29
chore(storage): introduce storage trace package

cojenco:
- Introduce storage trace package
- Test the new storage trace `startSpan` accepts StartSpanOption and can setAttributes
- Test Cloud Trace adoption attributes are appended with the new storage trace package

### DIFF
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -83,7 +83,7 @@ func (c *Client) Bucket(name string) *BucketHandle {
 // If attrs is nil the API defaults will be used.
 func (b *BucketHandle) Create(ctx context.Context, projectID string, attrs *BucketAttrs) (err error) {
 	if isOTelTracingDevEnabled() {
-		ctx, _ = startSpan(ctx, "cloud.google.com/go/storage.Bucket.Create")
+		ctx, _ = startSpan(ctx, "storage.Bucket.Create")
 		defer func() { endSpan(ctx, err) }()
 	} else {
 		ctx = trace.StartSpan(ctx, "cloud.google.com/go/storage.Bucket.Create")
@@ -101,7 +101,7 @@ func (b *BucketHandle) Create(ctx context.Context, projectID string, attrs *Buck
 // Delete deletes the Bucket.
 func (b *BucketHandle) Delete(ctx context.Context) (err error) {
 	if isOTelTracingDevEnabled() {
-		ctx, _ = startSpan(ctx, "cloud.google.com/go/storage.Bucket.Delete")
+		ctx, _ = startSpan(ctx, "storage.Bucket.Delete")
 		defer func() { endSpan(ctx, err) }()
 	} else {
 		ctx = trace.StartSpan(ctx, "cloud.google.com/go/storage.Bucket.Delete")
@@ -161,7 +161,7 @@ func (b *BucketHandle) Object(name string) *ObjectHandle {
 // Attrs returns the metadata for the bucket.
 func (b *BucketHandle) Attrs(ctx context.Context) (attrs *BucketAttrs, err error) {
 	if isOTelTracingDevEnabled() {
-		ctx, _ = startSpan(ctx, "cloud.google.com/go/storage.Bucket.Attrs")
+		ctx, _ = startSpan(ctx, "storage.Bucket.Attrs")
 		defer func() { endSpan(ctx, err) }()
 	} else {
 		ctx = trace.StartSpan(ctx, "cloud.google.com/go/storage.Bucket.Attrs")
@@ -175,7 +175,7 @@ func (b *BucketHandle) Attrs(ctx context.Context) (attrs *BucketAttrs, err error
 // Update updates a bucket's attributes.
 func (b *BucketHandle) Update(ctx context.Context, uattrs BucketAttrsToUpdate) (attrs *BucketAttrs, err error) {
 	if isOTelTracingDevEnabled() {
-		ctx, _ = startSpan(ctx, "cloud.google.com/go/storage.Bucket.Update")
+		ctx, _ = startSpan(ctx, "storage.Bucket.Update")
 		defer func() { endSpan(ctx, err) }()
 	} else {
 		ctx = trace.StartSpan(ctx, "cloud.google.com/go/storage.Bucket.Update")
--- a/storage/bucket.go
+++ b/storage/bucket.go
@@ -82,13 +82,8 @@ func (c *Client) Bucket(name string) *BucketHandle {
 // Create creates the Bucket in the project.
 // If attrs is nil the API defaults will be used.
 func (b *BucketHandle) Create(ctx context.Context, projectID string, attrs *BucketAttrs) (err error) {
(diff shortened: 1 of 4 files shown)


storage/trace.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"cloud.google.com/go/storage/internal"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	otelcodes "go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const (
+	OpenTelemetryTracingExpVar = "GO_STORAGE_EXPERIMENTAL_OTEL_TRACING"
```
BrennaEpp, line 29: I don't think we want to export this since we'd ideally remove it after the migration is done.

storage/trace.go
```diff
@@ -0,0 +1,70 @@
...
+// isOTelTracingDevEnabled checks the development flag until experimental feature is launched.
+func isOTelTracingDevEnabled() bool {
+	return os.Getenv(OpenTelemetryTracingExpVar) == "true"
+}
+
+func tracer() trace.Tracer {
+	return otel.Tracer(defaultTracerName, trace.WithInstrumentationVersion(internal.Version))
+}
+
+// startSpan accepts SpanStartOption and is used to replace internal/trace/StartSpan.
```
BrennaEpp, line 44: This comment is fine for drafting, but we should change it to something else before merging

storage/bucket.go
```diff
@@ -82,8 +82,13 @@ func (c *Client) Bucket(name string) *BucketHandle {
 // Create creates the Bucket in the project.
 // If attrs is nil the API defaults will be used.
 func (b *BucketHandle) Create(ctx context.Context, projectID string, attrs *BucketAttrs) (err error) {
-	ctx = trace.StartSpan(ctx, "cloud.google.com/go/storage.Bucket.Create")
-	defer func() { trace.EndSpan(ctx, err) }()
+	if isOTelTracingDevEnabled() {
+		ctx, _ = startSpan(ctx, "cloud.google.com/go/storage.Bucket.Create")
```
BrennaEpp, line 86: Since `startSpan` is completely internal to Storage now, can we factor out `cloud.google.com/go/storage`?

BrennaEpp (commented):
Quick review, looking good!

storage/trace.go
cojenco: Updated, thanks! If fuse needs it anytime sooner, we can change it to an experimental env var.

storage/bucket.go
cojenco: Absolutely, I've changed these to `storage.Bucket.Create` so we're still following the semantic conventions of $package.$service.$method

BrennaEpp: Oh I meant, just to move `cloud.google.com/go/` to the `startSpan` method so that we don't have to repeat it in the name of each span (and to help with consistency).

storage/trace.go
BrennaEpp: Not seeing any changes here yet

storage/trace.go
```diff
@@ -0,0 +1,70 @@
...
+
+const (
+	storageOtelTracingDevVar = "GO_STORAGE_DEV_OTEL_TRACING"
+	defaultTracerName        = "cloud.google.com/go/storage"
+	gcpClientRepo            = "googleapis/google-cloud-go"
+	gcpClientArtifact        = "storage"
+)
+
+// isOTelTracingDevEnabled checks the development flag until experimental feature is launched.
+func isOTelTracingDevEnabled() bool {
```
tritone, line 36: This func seems unused? What is the actual purpose of the env var?

storage/trace_test.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, span := range spans {
+		if diff := testutil.Diff(span, wantAttributes, opts...); diff != "" {
+			t.Errorf("diff: -got, +want:\n%s\n", diff)
+		}
+	}
+}
+
+func spanAttributesComparer(a, b tracetest.SpanStub) bool {
+	if a.Name != b.Name {
+		fmt.Printf("name mismatch: a.Name: %v, b.Name: %v\n", a.Name, b.Name)
```
BrennaEpp, line 85: Shouldn't this be printed in the diff when comparing? What does the output on failure look like?

storage/trace.go
BrennaEpp: It will be used during development to enable the new stuff only if that var is set. It will be removed once launched

tritone: Yes, but it doesn't actually seem used anywhere currently, in which case it should be removed from this PR.

storage/trace.go
```diff
@@ -0,0 +1,70 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	otelcodes "go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const (
+	storageOtelTracingDevVar = "GO_STORAGE_DEV_OTEL_TRACING"
+	defaultTracerName        = "cloud.google.com/go/storage"
+	gcpClientRepo            = "googleapis/google-cloud-go"
+	gcpClientArtifact        = "storage"
```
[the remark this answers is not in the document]
cojenco: `gcp.client.artifact` IIUC refers to the alternative targets, such as `storage` vs `control/apiv2` vs `dataflux` etc. Therefore I've set it to "storage" for now. Open to suggestions to make it more clear.

storage/trace.go
```diff
@@ -0,0 +1,70 @@
...
+		span.RecordError(err)
+	}
+	span.End()
+}
+
+// getCommonTraceOptions includes the common attributes used for Cloud Trace adoption tracking.
+func getCommonTraceOptions() []trace.SpanStartOption {
+	opts := []trace.SpanStartOption{
+		trace.WithAttributes(
+			attribute.String("gcp.client.version", internal.Version),
```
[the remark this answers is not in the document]
cojenco: `gcp.client.service` would be `storage` or `google.storage.v2.Storage` depending on the transport, so we'd annotate this in a different place

storage/trace_test.go
cojenco: Updated, yep this is printed in the diff.

storage/trace.go
cojenco: `TestTraceStorageTraceStartEndSpan` uses this development flag env var.  

It verifies the storage trace package works only when this development flag is enabled (so we're not breaking existing usage). It also tests that the new startSpan accepts StartSpanOption and returns a span so attributes can be easily set through a library call.

storage/trace.go
cojenco: Updated PTAL, thanks!

storage/trace.go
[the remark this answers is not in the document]
tritone: cloud.google.com/go/storage is preferred because it's the package name in Go. But, we can update it later.

tritone (approved):

PR #11052, opened 2024-10-29
fix(bigquery): parse negative NUMERIC from arrow

alvarowolfx:
The `big.Int.SetBytes` method interprets input as an unsigned number, so we lose sign information when parsing using. So Arrow decimal parsing code is simplified by just relying on `big.Rat.SetString` method for parsing. 

Towards internal b/376149769


alvarowolfx:
cc @tfinlay-lightspeed

shollyman (approved):

PR #11059, opened 2024-10-29
feat(auth): add universe domain support to idtoken

quartzmo:
# Summary

IdToken flow can be done in two ways: 1) OAuth2 server calls or 2) IAM service calls. The majority of implementations went with the OAuth2 server calls. In non-GDU environments, however, there is no OAuth2 server, so the IAM implementation must be used. This is enabled as follows:

* Move `impersonatedIDTokenProvider.Token` in `auth/credentials/impersonate/idtoken.go` to `IDTokenIAMOptions.Token` in `auth/credentials/internal/impersonate/idtoken.go` to enable reuse in `auth/credentials/idtoken/iam.go`.
* Move `formatIAMServiceAccountName` in `auth/credentials/impersonate/impersonate.go` to `FormatIAMServiceAccountResource ` in `auth/internal/internal.go`, for reuse.
* In `auth/credentials/idtoken/file.go` introduce a conditional (`resolveUniverseDomain(f) == internal.DefaultUniverseDomain`) which if true sustains the existing behavior (`auth.New2LOTokenProvider`) and if false introduces the new non-GDU behavior in `IDTokenIAMOptions.Token`

Note that to use the IAM endpoint in a non-GDU environment, the service account must have the `iam.serviceAccountTokenCreator` role because the service account is not only the target service account, but also the caller; the Oauth2 endpoint doesn’t need this.


auth/credentials/idtoken/iam.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// iamIDTokenProvider performs an authenticated RPC with the IAM service to
+// obtain an ID token. The provided client must be fully authenticated and
+// authorized with the iam.serviceAccountTokenCreator role.
+//
+// This TokenProvider is primarily intended for use in non-GDU universes, which
+// do not have access to the oauth2.googleapis.com/token endpoint, and thus must
+// use IAM generateIdToken instead.
+type iamIDTokenProvider struct {
```
codyoss, line 51: reduce duplication in internal folder or something if possible.

quartzmo: @codyoss I extracted the duplicated code to `internal`. PTAL at the final commit and let me know if you think it's worth it, otherwise easy to revert: https://github.com/googleapis/google-cloud-go/pull/11059/commits/00a963eeb7ffd7d068c5119463a8456460de40b9

auth/internal/internal.go
```diff
@@ -141,6 +142,24 @@ func GetProjectID(b []byte, override string) string {
 	return v.Project
 }
 
+// DoJSONRequest sends the provided JSON bytes with the client. It reads the response
+// body, checks the status code for errors, and returns the body.
+func DoJSONRequest(ctx context.Context, c *http.Client, url string, bodyBytes []byte, pkg string) ([]byte, error) {
+	req, err := http.NewRequestWithContext(ctx, "POST", url, bytes.NewReader(bodyBytes))
```
codyoss, line 148: I think either the name of this function and should be changed, or the signature . This hardcodes that requests flowing through here have to be a POST.

auth/credentials/idtoken/file.go
```diff
@@ -50,38 +52,23 @@ func credsFromBytes(b []byte, opts *Options) (*auth.Credentials, error) {
...
 			return nil, err
 		}
-		opts2LO := &auth.Options2LO{
-			Email:        f.ClientEmail,
-			PrivateKey:   []byte(f.PrivateKey),
-			PrivateKeyID: f.PrivateKeyID,
-			TokenURL:     f.TokenURL,
-			UseIDToken:   true,
+		var tp auth.TokenProvider
+		if resolveUniverseDomain(f) != internal.DefaultUniverseDomain {
```
codyoss, line 56: nit: I usually try to keep the first case positive, `==`, when possible.

auth/credentials/idtoken/file.go
```diff
@@ -50,38 +52,23 @@ func credsFromBytes(b []byte, opts *Options) (*auth.Credentials, error) {
...
-		customClaims["target_audience"] = opts.Audience
-
-		opts2LO.PrivateClaims = customClaims
-		tp, err := auth.New2LOTokenProvider(opts2LO)
 		if err != nil {
 			return nil, err
 		}
 		tp = auth.NewCachedTokenProvider(tp, nil)
-		return auth.NewCredentials(&auth.CredentialsOptions{
+		creds := auth.NewCredentials(&auth.CredentialsOptions{
```
codyoss, line 65: nit: this variable is immediately returned. please remove.

auth/credentials/idtoken/file.go
```diff
@@ -125,3 +112,77 @@ func credsFromBytes(b []byte, opts *Options) (*auth.Credentials, error) {
...
+		})
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		client = opts.Client
+	}
+	its := iamIDTokenProvider{
+		client:         client,
+		universeDomain: resolveUniverseDomain(f),
```
codyoss, line 173: I don't think `f` is needed in this context. In this case f == creds, right?

Also, this is being used as a client universe domain to manipulate the endpoint. Should this come from opts? If yes, can we add a test for this. If no, can you write a comment here why we are using the creds UD.

auth/credentials/idtoken/file.go
```diff
@@ -125,3 +112,77 @@ func credsFromBytes(b []byte, opts *Options) (*auth.Credentials, error) {
...
+		signerEmail:    f.ClientEmail,
+		audience:       opts.Audience,
+	}
+	return its, nil
+}
+
+// resolveUniverseDomain returns the default service domain for a given
+// Cloud universe. This is the universe domain configured for the credentials,
+// which will be used in endpoint.
+func resolveUniverseDomain(f *credsfile.ServiceAccountFile) string {
```
codyoss, line 183: nit: based on my comment above I think this can be removed.

auth/credentials/idtoken/file.go
```diff
@@ -125,3 +112,77 @@ func credsFromBytes(b []byte, opts *Options) (*auth.Credentials, error) {
...
+			Credentials:    creds,
+			UniverseDomain: opts.UniverseDomain,
+		})
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		client = opts.Client
+	}
+	its := iamIDTokenProvider{
```
codyoss, line 171: Is there a reason not to just call the impersonate package instead of duplicating this code? Or could they share code in a local internal?

auth/credentials/impersonate/idtoken.go
```diff
@@ -85,49 +90,45 @@ func NewIDTokenCredentials(opts *IDTokenOptions) (*auth.Credentials, error) {
...
+		if opts.Credentials == nil {
+			creds, err = credentials.DetectDefault(&credentials.DetectOptions{
+				Scopes:           []string{defaultScope},
+				UseSelfSignedJWT: true,
+			})
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			creds = opts.Credentials
```
codyoss, line 104: nit: you can set this value before the if and then check if it is nil after.

auth/credentials/impersonate/idtoken.go
```diff
@@ -85,49 +90,45 @@ func NewIDTokenCredentials(opts *IDTokenOptions) (*auth.Credentials, error) {
...
 	}
 
+	universeDomainProvider := resolveUniverseDomainProvider(creds)
 	itp := impersonatedIDTokenProvider{
-		client:          client,
-		targetPrincipal: opts.TargetPrincipal,
-		audience:        opts.Audience,
-		includeEmail:    opts.IncludeEmail,
+		client:                 client,
+		universeDomainProvider: universeDomainProvider,
```
codyoss, line 120: kinda similar to an earlier comment. Should this be coming from opts.UniverseDomain now? If not, add a comment.

auth/credentials/impersonate/user.go
```diff
@@ -29,6 +29,10 @@ import (
 	"cloud.google.com/go/auth/internal"
 )
 
+var (
+	iamCredentialsEndpoint = "https://iamcredentials.googleapis.com"
```
codyoss, line 33: Should this endpoint be universe aware?

quartzmo: No, user authentication is not supported.

auth/internal/internal.go
quartzmo: Added param `method string` to signature. Please suggest a different name if you want it. The reason for the current name is that it sets Content-Type to application/json.

auth/credentials/idtoken/file.go
quartzmo: > I don't think f is needed in this context. In this case f == creds, right?

If authenticated `client` is provided, the creds won't be accessible. Also `Credentials` does not expose the universe domain string from the source file, just the provider.

> write a comment here why we are using the creds UD.

Done, thanks.

auth/credentials/idtoken/file.go
quartzmo: This is needed because we are obtaining the universe domain from the credentials source file, which may result in a blank value.

auth/credentials/impersonate/idtoken.go
quartzmo: Done, thanks.

auth/credentials/impersonate/idtoken.go
quartzmo: Great suggestion, fixed in two files, thanks!

auth/credentials/idtoken/file.go
quartzmo: I think you're right, I'll try again to refactor out duplicated code, thanks.

auth/credentials/idtoken/file.go
quartzmo: Fixed, thanks!

auth/credentials/idtoken/file.go
quartzmo: Fixed, thanks!

auth/credentials/idtoken/iam.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// iamIDTokenProvider performs an authenticated RPC with the IAM service to
+// obtain an ID token. The provided client must be fully authenticated and
+// authorized with the iam.serviceAccountTokenCreator role.
+//
+// This TokenProvider is primarily intended for use in non-GDU universes, which
+// do not have access to the oauth2.googleapis.com/token endpoint, and thus must
+// use IAM generateIdToken instead.
+type iamIDTokenProvider struct {
```
codyoss, line 33: I think this struct can go away completely. should now be able to use impersonate.IDTokenIAMOptions directly.

auth/credentials/impersonate/idtoken.go
```diff
@@ -108,48 +108,39 @@ func NewIDTokenCredentials(opts *IDTokenOptions) (*auth.Credentials, error) {
...
-}
-
-type generateIDTokenResponse struct {
-	Token string `json:"token"`
-}
-
 type impersonatedIDTokenProvider struct {
-	client *http.Client
-	logger *slog.Logger
+	client                 *http.Client
```
codyoss, line 141: Same with my other comment, this structure should go away in favor of the common code

codyoss (approved):
LGTM to me overall. Just refactor a little to remove some now unnecessary structs.

auth/credentials/idtoken/iam.go
quartzmo: Good catch, can't believe I missed this. Thanks!

PR #11062, opened 2024-10-30
fix(auth/oauth2adapt): convert token metadata where possible

codyoss:
This was missed in the orginal implmentation. The code is still not prefect as converting from an oauth2.Token to auth.Token is lossy because there is no way to extract out the whole metadata context. Converting from auth.Token -> oauth2.Token is not lossy though.

This bug was discovered because some direct path checks look for theses metadata values. Just to be safe we will preserve both forms of these values in the final token metadata, if present.

Internal Bug: 376281562

### DIFF
--- a/auth/oauth2adapt/oauth2adapt.go
+++ b/auth/oauth2adapt/oauth2adapt.go
@@ -26,6 +26,13 @@ import (
 	"golang.org/x/oauth2/google"
 )
 
+const (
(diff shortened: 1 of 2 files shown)


quartzmo (approved):

tritone (approved):
Thanks for the quick fix Cody!

PR #11066, opened 2024-10-31
feat(bigtable): Async refresh dry run in parallel with sync refresh

bhshkh:
Async refresh token and record any errors to debug_tags metrics with tag field set to "async_refresh_dryn_run"
Corresponding google-api-go-client PR: https://github.com/googleapis/google-api-go-client/pull/2849 

Fixes: b/372244283

This PR also updates conformance test Go version. Currently, conformance tests fail to run with error:
```
go: ../go.mod requires go >= 1.22.7 (running go 1.22.5; GOTOOLCHAIN=local)
```


bhshkh:
Facing https://github.com/googleapis/google-cloud-go/issues/11161 when running multiple ReadRow operations in parallel. Will work on fixing it

bigtable/bigtable.go
```diff
@@ -112,21 +125,24 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
...
+	asyncRefreshMetricAttrs := metricsTracerFactory.clientAttributes
+	asyncRefreshMetricAttrs = append(asyncRefreshMetricAttrs,
+		attribute.String(metricLabelKeyTag, "async_refresh_dry_run"),
 
-	// Create a OpenTelemetry metrics configuration
-	metricsTracerFactory, err := newBuiltinMetricsTracerFactory(ctx, project, instance, config.AppProfile, metricsProvider, opts...)
+		// Table, cluster and zone are unknown at this point
+		// Use default values
+		attribute.String(monitoredResLabelKeyTable, ""),
+		attribute.String(monitoredResLabelKeyCluster, ""),
```
mutianf, line 136: unknown cluster should be `<unspecified>`

bigtable/metrics.go
```diff
@@ -253,6 +263,16 @@ func (tf *builtinMetricsTracerFactory) createInstruments(meter metric.Meter) err
...
 		metric.WithUnit(metricUnitCount),
 	)
+	if err != nil {
+		return err
+	}
+
+	// Create debug_tags
+	tf.debugTags, err = meter.Int64Counter(
+		metricNameDebugTags,
+		metric.WithDescription("The number of additional RPCs sent after the initial attempt."),
```
mutianf, line 273: description should be: "A counter of internal client events used for debugging."

bigtable/bigtable.go
bhshkh: Updated

bigtable/metrics.go

bhshkh:
> Facing #11161 when running multiple ReadRow operations in parallel. Will work on fixing it

This has been fixed in oauth2adapt 0.2.6 . oauth2adapt is consumed in google.golang.org/api 0.210.0
So, I have updated go.mod to use google.golang.org/api 0.210.0

bigtable/metrics_util.go
```diff
@@ -30,6 +30,7 @@ import (
 const (
 	defaultCluster = "unspecified"
 	defaultZone    = "global"
+	defaultTable   = "unspecified"
```
mutianf, line 33:
```suggestion
	defaultTable   = "<unspecified>"
```

bigtable/metrics_util.go
```diff
@@ -30,6 +30,7 @@ import (
 const (
 	defaultCluster = "unspecified"
```
mutianf, line 31:
```suggestion
	defaultCluster = "<unspecified>"
```

bigtable/metrics.go
```diff
@@ -109,6 +111,12 @@ var (
 			},
 			recordedPerAttempt: true,
 		},
+		metricNameDebugTags: {
+			additionalAttrs: []string{
+				metricLabelKeyTag,
+			},
+			recordedPerAttempt: true,
```
mutianf, line 118: Where is this metric updated? I don't think we want to record a value per attempt, we only want to increment the value when there's an error with the async refresh.

mutianf (changes requested):

bigtable/metrics_util.go
bhshkh: Java has "unspecified": https://github.com/googleapis/java-bigtable/blob/168293937cc7f438a3ec2dee46805aa8e12089c4/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracer.java#L78-L80

bigtable/metrics_util.go
bhshkh: Java has "unspecified": https://github.com/googleapis/java-bigtable/blob/168293937cc7f438a3ec2dee46805aa8e12089c4/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracer.java#L78-L80

bigtable/metrics.go
bhshkh: That field was not getting used for `debug_tags` metric. I have removed it now.

**More detail**:
What internally happens in the Go client library when user creates a client and invokes `ReadRows` method :
1. Create client
    1.1. Create tracerFactory
    1.2. Create attributes to be added to the `debug_tags` metric
    1.3. Create callback function to run when async refresh fails. This callback function records a change to the counter with the attributes created in 1.2
    1.4. Pass the callback function in the `EnableAsyncRefreshDryRun` client option while creating the GRPC dial pool
2. ReadRows
    2.1. Create metrics tracer
    2.2. Obtain Token. Run the callback function if any errors (not done by Bigtable client but by the auth package)
    2.3. Make ReadRows RPC call
    2.4. Receive Response
    2.5. After ReadRows operation completes, invoke `func (mt *builtinMetricsTracer) toOtelMetricAttrs(metricName string) ([]attribute.KeyValue, error)` 4 times, once with each metric name: operation_latencies, attempt_latencies, server_latencies and retry_count
    2.6. `toOtelMetricAttrs` uses the data saved in metrics tracer to create attributes for the metrics. If the metric needs to be recorded per attempt, include attempt status, else include operation status

As can be seen from above flow, ` toOtelMetricAttrs` is never called with `debug_tags` metric name. So, `recordedPerAttempt` was not getting used. Since, it was also incorrectly set to true, I have removed it now.

Also wanted to highlight that since the `debug_tags` metric is getting recorded long before the response is received, its table, cluster and zone attributes will be <unspecified>

bigtable/metrics_util.go
bhshkh: Updated to <unspecified>

bigtable/metrics_util.go
bhshkh: Updated to <unspecified>

bigtable/metrics_util.go
mutianf: Yeah, sorry we updated it recenlty to be in sync with the server metrics.

mutianf (approved):

bigtable/bigtable.go
```diff
@@ -112,21 +125,24 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
...
-		// Do not emit metrics when emulator is being used
-		metricsProvider = NoopMetricsProvider{}
-	}
+	asyncRefreshMetricAttrs := metricsTracerFactory.clientAttributes
+	asyncRefreshMetricAttrs = append(asyncRefreshMetricAttrs,
+		attribute.String(metricLabelKeyTag, "async_refresh_dry_run"),
 
-	// Create a OpenTelemetry metrics configuration
-	metricsTracerFactory, err := newBuiltinMetricsTracerFactory(ctx, project, instance, config.AppProfile, metricsProvider, opts...)
+		// Table, cluster and zone are unknown at this point
```
gkevinzheng, line 133: Add the same TODO for deleting the async refresh dry run metric/internal option once the internal option is removed from the other repo after the bugfix.

PR #11067, opened 2024-10-31
chore(storage): bump auth dependency

tritone:
Picks up fix to directPath logic.


BrennaEpp (commented):
I think we need to bump `auth/oauth2adapt` - see #11063

BrennaEpp (approved):

PR #11078, opened 2024-11-02
feat(compute/metadata): add debug logging

codyoss:


codyoss:
Requires https://github.com/googleapis/gax-go/pull/380

compute/metadata/metadata.go
```diff
@@ -24,13 +24,16 @@ import (
...
+	"log/slog"
 	"net"
 	"net/http"
 	"net/url"
 	"os"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/googleapis/gax-go/v2/internallog"
```
liggitt, line 36: is there *any* way to avoid this? compute/metadata is *impossible* to avoid as a dependency (google.golang.org/grpc pulls it in, etc)

we *really* should not be forcing transitive deps on cloud modules like this

liggitt: compute/metadata specifically is required by almost everything in the go ecosystem, keeping its dependency graph small and avoiding transitive deps to other cloud libraries is much more important for this module than anything else in this repo.

using stdlib slog or a locally-defined interface and pushing callers to wire any specific cloud instrumentation they want is much more in line with what I would expect here.

I'd also like to see a presubmit that will catch / guard against unconsidered introduction of new dependencies to this module in particular

compute/metadata/metadata.go
```diff
@@ -408,17 +412,41 @@ func strsContains(ss []string, s string) bool {
...
 // A Client provides metadata.
 type Client struct {
-	hc *http.Client
+	hc     *http.Client
+	logger *slog.Logger
+}
+
+// Options for configuring a [Client].
+type Options struct {
+	// Client is the HTTP used to make requests. Optional.
```
quartzmo, line 421: Typo: "the HTTP used"

compute/metadata/metadata.go
```diff
@@ -408,17 +412,41 @@ func strsContains(ss []string, s string) bool {
...
+	Client *http.Client
+	// Logger is used to log information about outgoing requests. Optional.
```
quartzmo, line 423: "to log information about outgoing requests" isn't accurate, because responses are also logged. Maybe: "to log HTTP requests and responses".

It would probably also be good to disclose what is logged: "The structured log statements for requests will include method, URL, headers and payload. The structured log statements for responses will include status, headers and payload. 

And we should add a reassurance here about the default behavior. Maybe: "If not provided, requests and responses will not be logged."

quartzmo (changes requested):
I think we should copy/adapt https://github.com/googleapis/gax-go/blob/main/v2/internallog/internallog_test.go here as well, to cover `log.go`.

compute/metadata/metadata.go
codyoss: done.

compute/metadata/metadata.go
codyoss: fixed up a bit. Did not give low level details of what is logged or the format as this is an impl details that can change over time and/or by the logger that is passed in.

quartzmo (approved):

PR #11079, opened 2024-11-02
feat(auth): add logging support

codyoss:
- add logging to all outgoing requests
- accept a logger from public constructors
- left a TODO to cleanup in a followup metadata package PR


codyoss:
Requires https://github.com/googleapis/gax-go/pull/380

auth/credentials/impersonate/idtoken.go
```diff
@@ -86,20 +93,21 @@ func NewIDTokenCredentials(opts *IDTokenOptions) (*auth.Credentials, error) {
...
-			// TODO: test not signed jwt more
-			creds, err = credentials.DetectDefault(&credentials.DetectOptions{
-				Scopes:           []string{defaultScope},
-				UseSelfSignedJWT: true,
-			})
-			if err != nil {
-				return nil, err
-			}
+		// TODO: test not signed jwt more
+		creds, err = credentials.DetectDefault(&credentials.DetectOptions{
```
quartzmo, line 100: Will this overwrite the value from `opts.Credentials` (if present) set above on line 95?

quartzmo: Maybe `opts.Credentials` could be passed in here using `DetectOptions`, and line 95 deleted?

auth/auth.go
```diff
@@ -560,10 +568,12 @@ func (tp tokenProvider2LO) Token(ctx context.Context) (*Token, error) {
 		return nil, err
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	tp.logger.DebugContext(ctx, "2LO token fetch", "request", internallog.HTTPRequest(req, []byte(v.Encode())))
```
quartzmo, line 571: Bikeshed: "request" instead of "fetch" in all of these request logs. (Logic: request/response are the standard pair, fetch/response seems unusual.)

quartzmo (changes requested):

auth/auth.go
codyoss: Done.

auth/credentials/impersonate/idtoken.go
codyoss: This was just a bad merge on my part when I rebased. Fixed it up.

quartzmo (approved):

PR #11083, opened 2024-11-04
fix(auth): restore Application Default Credentials support to idtoken

quartzmo:


auth/credentials/idtoken/idtoken.go
```diff
@@ -113,13 +113,20 @@ func NewCredentials(opts *Options) (*auth.Credentials, error) {
...
+	b := opts.jsonBytes()
+	if b == nil && metadata.OnGCE() {
 		return computeCredentials(opts)
+	} else if b == nil {
+		creds, err := credentials.DetectDefault(&credentials.DetectOptions{
+			Audience: opts.Audience,
+		})
+		if err != nil {
+			return nil, err
+		}
```
quartzmo, line 125: Note that in the case of ADC and impersonated or external account credential types, the `creds` created here are discarded, then re-created in a new call to `DetectDefault` (with additional `DetectOptions` args) in `credsFromBytes`.

auth/credentials/idtoken/idtoken.go
```diff
@@ -113,13 +113,20 @@ func NewCredentials(opts *Options) (*auth.Credentials, error) {
...
-	if b := opts.jsonBytes(); b != nil {
-		return credsFromBytes(b, opts)
-	}
-	if metadata.OnGCE() {
+	b := opts.jsonBytes()
+	if b == nil && metadata.OnGCE() {
 		return computeCredentials(opts)
+	} else if b == nil {
+		creds, err := credentials.DetectDefault(&credentials.DetectOptions{
+			Audience: opts.Audience,
```
codyoss, line 121: instead of calling the jsonBytes method should we just call DetectDefault and pass in the opts here.

codyoss, line 121: This aud is for the idtoken, not the creds. This should be some default scopes

auth/credentials/idtoken/idtoken.go
codyoss: I think those opts could just be all passed above.

codyoss (changes requested):

auth/credentials/idtoken/idtoken.go
quartzmo: `jsonBytes` is called in order to determine if we should skip `DetectDefault` and go immediately into the `idtoken.computeCredentials` path. Thus the precedence is as follows. Let me know if it's wrong.

1. CredentialsFile or CredentialsJSON is present
2. OnGCE == true
3. DetectDefault.

quartzmo: Question: If we just call DetectDefault and pass in the opts here, what do we do with the resulting `creds`? If we simply call `credsFromBytes(creds.JSON(), opts)` with it, we are mostly throwing away the opts passed here.

auth/credentials/idtoken/idtoken.go
quartzmo: See questions above.

quartzmo: I refactored to use `DetectDefault` in all non-compute cases, which simplifies the code a bit.

auth/credentials/idtoken/idtoken.go
quartzmo: Fixed, thanks!

quartzmo: I refactored to use `DetectDefault` in all non-compute cases, which simplifies the code a bit.

auth/credentials/idtoken/idtoken.go
```diff
@@ -113,13 +113,18 @@ func NewCredentials(opts *Options) (*auth.Credentials, error) {
...
 	}
-	return nil, fmt.Errorf("idtoken: couldn't find any credentials")
+	creds, err := credentials.DetectDefault(&credentials.DetectOptions{
+		Scopes:          defaultScopes,
+		CredentialsJSON: b,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return credsFromBytes(creds.JSON(), opts)
```
codyoss, line 127: Can this be refactored a little more to take the whole creds now, that was you don't have to call detect default again later for one of the flows.

quartzmo: Updated, PTAL!

codyoss (approved):

PR #11100, opened 2024-11-07
feat: introduce dp detector based on grpc metrics

frankyn:


storage/storage.go
```diff
@@ -233,6 +239,73 @@ func NewGRPCClient(ctx context.Context, opts ...option.ClientOption) (*Client, e
...
 
+// CheckDirectConnectivitySupported checks if Direct Connectivity is available
+// for a specific bucket.
+//
+// Implementation currently uses client-side metrics for gRPC to detect
+// Direct Connectivity using grpc.lb.locality label.
+//
+// You may configure the client by passing in options from the [google.golang.org/api/option]
+// package.
+func CheckDirectConnectivitySupported(ctx context.Context, bucket string, opts ...option.ClientOption) (bool, error) {
```
tritone, line 250: I wonder if we should just make this return an error, and nil error means DP worked?

I think we should wrap and return any errors from the metrics stuff to help debug if things go wrong.

tritone (commented):
Few comments, thanks for your work on this!

storage/integration_test.go
```diff
@@ -324,6 +324,16 @@ var readCases = []readCase{
 	},
 }
 
+func TestIntegration_DetectDirectConnectivity(t *testing.T) {
+	ctx := skipHTTP("direct connectivity isn't available for json")
+	multiTransportTest(ctx, t, func(t *testing.T, ctx context.Context, bucket string, _ string, _ *Client) {
+		_, err := CheckDirectConnectivitySupported(ctx, bucket)
```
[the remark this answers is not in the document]
frankyn: Yea, we can do ResourceDetection and if in GCE we expect direct path for a co-located bucket. Worth trying and seeing what happens.

storage/storage.go
frankyn: I like this idea; instead of returning a boolean emit an err when DP isn't available otherwise no error means DP is available.

storage/integration_test.go
```diff
@@ -324,6 +326,34 @@ var readCases = []readCase{
...
+		// Using Resoource Detector to detect if test is being ran inside GCE
+		// if so, the test expects Direct Connectivity to be detected.
+		// Otherwise, it will only validate that Direct Connectivity was not
+		// detected.
+		detectedAttrs, err := resource.New(ctx, resource.WithDetectors(gcp.NewDetector()))
+		if err != nil {
+			t.Fatalf("resource.New: %v", err)
+		}
+		if v, present := detectedAttrs.Set().Value("cloud.platform"); present && v.AsString() == "gcp_compute_engine" {
+			err := CheckDirectConnectivitySupported(ctx, bucket)
```
tritone, line 341: I think we'd have to create the bucket in the test to make sure it is in the same region as the VM.

Does the resource detector allow parsing the region?

storage/storage.go
```diff
@@ -233,6 +238,58 @@ func NewGRPCClient(ctx context.Context, opts ...option.ClientOption) (*Client, e
 	return &Client{tc: tc}, nil
 }
 
+// CheckDirectConnectivitySupported checks if gRPC direct connectivity
+// is available for a specific bucket from the environment where the client
+// is running. A `nil` error represents Direct Connectivity was detected.
+//
+// You can pass in [option.ClientOption] you plan on passing to [NewGRPCClient]
```
tritone, line 245: Let's also add a line like:

> // Direct connectivity is expected to be available when running from inside GCP and connecting to a bucket in the same region.

storage/integration_test.go
frankyn: good call addressed.

storage/integration_test.go
```diff
@@ -324,6 +326,46 @@ var readCases = []readCase{
...
+			t.Fatalf("resource.New: %v", err)
+		}
+		attrs := detectedAttrs.Set()
+		if v, exists := attrs.Value("cloud.platform"); exists && v.AsString() == "gcp_compute_engine" {
+			v, exists = attrs.Value("cloud.region")
+			if !exists {
+				t.Fatalf("CheckDirectConnectivitySupported: region not detected")
+			}
+			region := v.AsString()
+			newBucketName := "go-integration-dc-" + region + "-" + uidSpace.New()
```
tritone, line 349: Use the prefix arg from multiTransportTest to name the bucket and insure it gets cleaned up.

storage/storage.go
```diff
@@ -233,6 +238,60 @@ func NewGRPCClient(ctx context.Context, opts ...option.ClientOption) (*Client, e
...
+	defer provider.Shutdown(ctx)
+	mo := opentelemetry.MetricsOptions{
+		MeterProvider:  provider,
+		Metrics:        stats.NewMetrics("grpc.client.attempt.duration"),
+		OptionalLabels: []string{"grpc.lb.locality"},
+	}
+	combinedOpts := append(opts, WithDisabledClientMetrics(), option.WithGRPCDialOption(opentelemetry.DialOption(opentelemetry.Options{MetricsOptions: mo})))
+	client, err := NewGRPCClient(ctx, combinedOpts...)
+	if err != nil {
+		return fmt.Errorf("checkDirectConnectivitySupported: %v", err)
```
tritone, line 268: These errors should reference the function being called for stack traces purposes, and use the %w verb for error wrapping (e.g. `fmt.Errorf("storage.NewGRPCClient: %w", err)`). Same below.

tritone (approved):
Two more minor nits, otherwise LGTM. Approving now once you fix.

storage/storage.go
frankyn: @tritone i introduced an error named `ErrDirectConnectivityNotDetected` in the latest commit; IIUC the client has errors to make it easier to check rather than string checking the error.

https://github.com/googleapis/google-cloud-go/pull/11100/files#diff-2a93e444b612bd9853c32889fb82c4041760536f84356bb0db04738c19b62ddeR74

WDYT?

PR #11107, opened 2024-11-08
chore(storage): Instrument existing metadata ops with storage trace package

cojenco:
Update: supersedes #11051

- Adds a development flag GO_STORAGE_DEV_OTEL_TRACING that will be removed upon experimental feature launch
- Replaces existing metadata ops instrumentation with storage trace package ([tracking sheet](https://docs.google.com/spreadsheets/d/1IYUimVewLLfUdQW7d5w4BFRWdMubtXc3vofJJRdpst0/edit?gid=0#gid=0))
- Adds emulated trace tests


storage/acl.go
```diff
@@ -77,8 +77,13 @@ type ACLHandle struct {
 
 // Delete permanently deletes the ACL entry for the given entity.
 func (a *ACLHandle) Delete(ctx context.Context, entity ACLEntity) (err error) {
-	ctx = trace.StartSpan(ctx, "cloud.google.com/go/storage.ACL.Delete")
-	defer func() { trace.EndSpan(ctx, err) }()
+	if isOTelTracingDevEnabled() {
```
tritone, line 80: Can we handle this check in startSpan rather than at each individual callsite? It would reduce a lot of boilerplate.

cojenco: Absolutely, this makes much more sense and is easier for cleanup. Updated PTAL, thanks!

cojenco:
@BrennaEpp could you PTAL and help check the CLA, thanks!

tritone (commented):
Nice start on this!

storage/trace_test.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	for _, span := range spans {
+		if diff := testutil.Diff(span, wantSpan, opts...); diff != "" {
+			t.Errorf("diff: -got, +want:\n%s\n", diff)
+		}
+	}
+	e.Reset()
+}
+
+func TestTraceSpansEmulated(t *testing.T) {
```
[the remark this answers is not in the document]
BrennaEpp: If we are grouping by emulator tests I'd suggest renaming the file where we keep them

storage/trace.go
```diff
@@ -0,0 +1,87 @@
...
+// If the context.Context provided in `ctx` contains a span then the newly-created
+// span will be a child of that span, otherwise it will be a root span.
+func startSpan(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
+	// TODO: Remove internalTrace upon experimental launch.
+	if !isOTelTracingDevEnabled() {
+		ctx = internalTrace.StartSpan(ctx, name)
+		return ctx, nil
+	}
+	opts = append(opts, getCommonTraceOptions()...)
+	ctx, span := tracer().Start(ctx, name, opts...)
```
BrennaEpp, line 56: Just to be clear, we are leaving out `cloud.google.com/go/` from the trace name?

I'd suggest at least adding it back to the internalTrace implementation.

cojenco: I've added the package name back to the internalTrace implementation with `appendPackageName()`. I originally removed this to follow the span name pattern of $package.$service.$method (c++ example `storage::Client::WriteObject/CreateResumableUpload`) but am open to adding it back. 

Pro is to have the official package name; Con is longer span name and repeated info with added attribute. Let me know what you think!

storage/trace_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// Create non-wrapped client to use for setup steps.
+	ctx := context.Background()
+	client, err := NewClient(ctx)
+	if err != nil {
+		t.Fatalf("storage.NewClient: %v", err)
+	}
+
+	skippedTraceMethods := []string{"storage.Bucket.AddNotification", "storage.Bucket.Notifications", "storage.Bucket.DeleteNotification"}
+	for spanName, fn := range traceMethods {
```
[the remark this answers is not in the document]
cojenco: Yeah I was hoping to do that too. I actually updated to use the `multiTransportTest` harness from the integration_test. 

I couldn't leverage the client_test `transportClientTest` mainly because we're testing trace instrumentation on the transport-agnostic layer, and for testing purposes I also need to configure `NewGRPCClient` and `NewClient` by passing in ClientOption.

storage/trace_test.go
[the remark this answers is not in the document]
cojenco: Continued discussion offline... So I guess we could have the trace tests and non-client emulator tests in a separate test file, maybe call it `emulator_test.go`? And the relevant client tests would remain in client_test.  

I won't move anything in the client_test just yet. We can do that after December as discussed :)

storage/acl.go
```diff
@@ -91,8 +90,8 @@ func (a *ACLHandle) Delete(ctx context.Context, entity ACLEntity) (err error) {
 
 // Set sets the role for the given entity.
 func (a *ACLHandle) Set(ctx context.Context, entity ACLEntity, role ACLRole) (err error) {
-	ctx = trace.StartSpan(ctx, "cloud.google.com/go/storage.ACL.Set")
-	defer func() { trace.EndSpan(ctx, err) }()
+	ctx, _ = startSpan(ctx, "storage.ACL.Set")
```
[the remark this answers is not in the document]
cojenco: Yes, we will return the span and add more context pieces wherever applicable. The ability to return a span is one of the main changes for the new trace package.

Error recording is also handled in `endSpan` rather than at each individual callsite. I think this will help reduce the boilerplate as well.

storage/trace_test.go
```diff
@@ -0,0 +1,302 @@
...
+			}
+			et.resources.notification = n
+		}
+	}
+}
+
+// traceMethods are library methods that have trace instrumentation. This is a map whose keys are
+// a string describing the spanName (e.g. storage.Bucket.Attrs) and values are functions which
+// wrap library methods that implement the API calls.
+var traceMethods = map[string]traceFunc{
```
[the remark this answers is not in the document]
cojenco: Updated, thanks!

storage/trace_test.go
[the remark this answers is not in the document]
tritone: That sounds good!

storage/trace_test.go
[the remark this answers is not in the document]
tritone: Ahh got it.

storage/trace.go
tritone: I think  keeping `cloud.google.com/go/storage` in the span name is preferable.

tritone (approved):
one minor nit, otherwise looks good to me at this point.

storage/trace_test.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func TestTraceSpansMultiEmulated(t *testing.T) {
+	ctx := skipJSONReads(context.Background(), "no reads in test")
+	// To only test storage library layer instrumentation,
+	// we disable transport layer traces for testing purposes.
+	opts := []option.ClientOption{
+		option.WithTelemetryDisabled(),
+	}
+	multiTransportTest(ctx, t, func(t *testing.T, ctx context.Context, bucket string, prefix string, client *Client) {
```
BrennaEpp, line 87: Should this be a `transportClientTest` instead?

BrennaEpp (commented):
Thanks Cathy!

storage/trace_test.go
BrennaEpp: `transportClientTest` has a check to not run if the emulator env var is not set https://github.com/googleapis/google-cloud-go/blob/48022acb5550244ffdff9a87f2792731eec9b005/storage/client_test.go#L2175

We should modify `transportClientTest` to allow configuration of client opts similar to `multiTransportTest`. Or add a check to this test specifically so it only runs on the emulator

BrennaEpp: I believe this caused this issue #11480

Issue #11080, opened 2024-11-04
spanner: panic on native metrics


apstndb:
I have found the client built in metrics(https://github.com/googleapis/google-cloud-go/pull/10419, https://github.com/googleapis/google-cloud-go/pull/10998) can panic in some condition.

## Client

Spanner

## Environment

go version go1.23.2 darwin/arm64

## Code and Dependencies

I can't find minimum reproducing code, but I can reproduce it in [cloudspannerecosystem/spanner-cli](https://github.com/cloudspannerecosystem/spanner-cli).

### Reproducing branch

Only do `go get cloud.google.com/go/spanner@v1.71.0`

https://github.com/apstndb/spanner-cli/pull/1

```
$ go test -v -run TestRequestPriority
=== RUN   TestRequestPriority
=== RUN   TestRequestPriority/use_default_MEDIUM_priority
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x28 pc=0x10316ef04]

goroutine 94 [running]:
cloud.google.com/go/spanner.(*builtinMetricsTracer).toOtelMetricAttrs(0x14000384090, {0x1032c539b, 0xf})
        /Users/apstndb/go/pkg/mod/cloud.google.com/go/spanner@v1.71.0/metrics.go:424 +0x3f4
cloud.google.com/go/spanner.recordOperationCompletion(0x14000384090)
        /Users/apstndb/go/pkg/mod/cloud.google.com/go/spanner@v1.71.0/metrics.go:490 +0x58
cloud.google.com/go/spanner.(*grpcSpannerClient).BatchCreateSessions(0x1400058a870, {0x1038a1980, 0x14000414d80}, 0x14000470000, {0x1400038a0c0, 0x1, 0x1})
        /Users/apstndb/go/pkg/mod/cloud.google.com/go/spanner@v1.71.0/grpc_client.go:134 +0x1c4
cloud.google.com/go/spanner.(*sessionClient).executeBatchCreateSessions(0x14000528160, {0x1038ae1b8, 0x1400058a870}, 0x1, 0x14000584690, 0x140005846c0, {0x10389aed8, 0x14000270200})
        /Users/apstndb/go/pkg/mod/cloud.google.com/go/spanner@v1.71.0/sessionclient.go:277 +0x430
created by cloud.google.com/go/spanner.(*sessionClient).batchCreateSessions in goroutine 50
[...]

PR #11109, opened 2024-11-09
fix(spanner): use spanner options when initializing monitoring exporter

rahul2393:
Issue: https://github.com/googleapis/google-cloud-go/issues/11080#issuecomment-2466057506


spanner/metrics.go
```diff
@@ -167,7 +167,12 @@ var (
 		return "global"
 	}
 
-	exporterOpts = []option.ClientOption{}
+	// GCM exporter should use the same options as Bigtable client
```
olavloite, line 170: nit: 
```suggestion
	// GCM exporter should use the same options as the Spanner client
```

olavloite (commented):
The change LGTM, but can we add a test that fails without this change, and succeeds with it? Or is that for some reason not possible?

spanner/metrics_monitoring_exporter.go
```diff
@@ -74,21 +77,22 @@ func (e errUnexpectedAggregationKind) Error() string {
...
 // Default exporter for built-in metrics
 type monitoringExporter struct {
-	shutdown     chan struct{}
-	client       *monitoring.MetricClient
-	shutdownOnce sync.Once
-	projectID    string
+	shutdown               chan struct{}
+	client                 *monitoring.MetricClient
+	shutdownOnce           sync.Once
+	projectID, compression string
```
olavloite, line 83: nit: the formatting seems a bit strange here. Could we have `compression` on a separate line?

olavloite (approved):

PR #11111, opened 2024-11-11
feat(storage): add Writer.ChunkTransferTimeout

Tulsishah:
Expose the ChunkTransferTimeout MediaOption through the manual
client layer. This allows users to set a longer timeout for
chunk retries in case of stalls in resumable uploads if desired.

Added emulator based unit tests to cover all the scenarios.


storage/client_test.go
```diff
@@ -1506,6 +1506,226 @@ func TestRetryReadStallEmulated(t *testing.T) {
...
+			name         string
+			instructions map[string][]string
+			wantDuration time.Duration
+		}{
+			{
+				name: "stall-on-first-chunk-with-chunk-transfer-timeout-zero",
+				instructions: map[string][]string{
+					"storage.objects.insert": {"stall-for-10s-after-1024K"},
+				},
+				wantDuration: 10 * time.Second,
```
tritone, line 1530: So it looks like this test will take >30s to run, which is too much for an emulator test IMO.

Could you maybe rewrite to have the stalls last for 1s, and set a context timeout of 500ms or less on the request? We would expect the requests to fail in this case.

storage/client_test.go
```diff
@@ -1506,6 +1506,226 @@ func TestRetryReadStallEmulated(t *testing.T) {
...
+			instructions         map[string][]string
+			chunkTransferTimeout time.Duration
+			wantDuration         time.Duration
+		}{
+			{
+				name: "stall-on-first-chunk-with-chunk-transfer-timeout-nonzero",
+				instructions: map[string][]string{
+					"storage.objects.insert": {"stall-for-10s-after-1024K"},
+				},
+				chunkTransferTimeout: 2 * time.Second,
```
tritone, line 1639: I would set the chunk transfer timeout shorter, maybe 100ms. We are working with a local emulator so latency should be lower than for the real service.

storage/client_test.go
```diff
@@ -1506,6 +1506,226 @@ func TestRetryReadStallEmulated(t *testing.T) {
...
+				wantDuration:         3 * time.Second,
```
tritone, line 1640: I don't think we should set a wantDuration. We can just set a context timeout that's shorter than the stall. That way, if the retry didn't work, we expect the request to fail with context deadline exceeded.

storage/client_test.go
```diff
@@ -1506,6 +1506,226 @@ func TestRetryReadStallEmulated(t *testing.T) {
...
+				}
+			})
+		}
+	})
+}
+
+// In resumable uploads, the chunk encounter a stall. The chunkTransferTimeout will then cancel the request and resend it.
+// The second request is expected to succeed. This allows the upload to finish earlier than if it had waited for the entire stall
+// duration, as the chunkTransferTimeout prevents unnecessary delays.
+func TestRetryWriteReqStallWithNonZeroChunkTransferTimeoutEmulated(t *testing.T) {
```
tritone, line 1619: nit: I'd just call this: TestWriterChunkTransferTimeoutEmulated.

You can also combine both tests; just separate out into cases which are expected to succeed and those which are expected to fail.

storage/writer.go
```diff
@@ -88,6 +88,15 @@ type Writer struct {
 	// cancellation.
 	ChunkRetryDeadline time.Duration
 
+	// ChunkTransferTimeOut sets a per-chunk retry transfer timeout for multi-chunk
```
tritone, line 91: I'd rephrase a bit; how about:

```
ChunkTransferTimeout sets a per-chunk request timeout for resumable uploads.

For resumable uploads, the Writer will terminate the request and attempt a retry
if the request to upload a particular chunk stalls for longer than this duration. Retries
may continue until the ChunkRetryDeadline is reached.

The default value is no timeout.
```

storage/writer.go
```diff
@@ -88,6 +88,15 @@ type Writer struct {
...
+	// resumable uploads.
+	//
+	// For uploads of larger files, the Writer will attempt to retry if the
+	// request to upload a particular chunk stalls for some time.
+	// If a single chunk has been attempting to upload for longer than this
+	// timeout, then it will be retried. The default value no timeout.
+	ChunkTransferTimeOut time.Duration
```
tritone, line 98: Let's keep capitalization consistent with the other repo; use `ChunkTransferTimeout`

tritone (commented):
Some suggestions on testing/docs

storage/client_test.go
Tulsishah: Done

storage/client_test.go
Tulsishah: Done

storage/client_test.go
Tulsishah: Done

tritone (commented):
LGTM, we can merge once the apiary release happens and dependency is picked up.

tritone (approved):

PR #10780, opened 2024-08-28
chore(auth): fix staticcheck errors

julieqiu:
Fix these issues:

```
credentials/detect.go:102:12: metadata.ProjectID is deprecated: Please use the context aware variant [ProjectIDWithContext].  (SA1019)
credentials/idtoken/compute.go:47:11: metadata.ProjectID is deprecated: Please use the context aware variant [ProjectIDWithContext].  (SA1019)
grpctransport/grpctransport.go:294:9: grpc.DialContext is deprecated: use NewClient instead.  Will be supported throughout 1.x.  (SA1019)
grpctransport/pool_test.go:71:16: grpc.Dial is deprecated: use NewClient instead.  Will be supported throughout 1.x.  (SA1019)
internal/transport/s2a.go:87:9: metadata.Get is deprecated: Please use the context aware variant [GetWithContext].  (SA1019)
```

For #10779

PR #11118, opened 2024-11-12
fix(auth): restore use of grpc.Dial

codyoss:
There have been a couple of reports lately around some subtle changes in behaviour of using the new recommend method. There is also a known bug, see https://github.com/grpc/grpc-go/issues/7556. Until the bug is fixed and we have a way forward we will revert to calling the deprecated Dial function.

Fixes: #7556


quartzmo:
Note that this reverts changes in https://github.com/googleapis/google-cloud-go/pull/10780.

quartzmo (approved):

iwinux:
@quartzmo Previously `auth/grpctransport/grpctransport.go` was using `grpc.DialContext`. Why did you change it to `grpc.Dial`?

codyoss:
@iwinux See the linked grpc bug in the OP for details.

iwinux:
I see no discussion related to `Dial` (the "reverted" call) v.s. `DialContext` (the original call).

codyoss:
Ah looking closer into the gRPC code this was correct at the time to avoid that linked bug. DialContext calls NewClient and had this bug. This does appear to be fixed now though, so still proceeding.

PR #11123, opened 2024-11-12
feat(bigquery): support IAM conditions in datasets

shollyman:
This PR adds support for IAM conditions via the existing dataset access mechanism.  To do so, the following changes are necessary:

* add the `Expr` type for expressing conditions, and wire it into the existing DatasetAccessEntry.
* add an option pattern to the Dataset-related RPC methods
* Add a new WithAccessPolicyVersion option for setting access policies

To expose the new functionality, this PR adds CreateWithOptions, UpdateWithOptions, MetadataWithOptions methods on Dataset that accept the new option.


bigquery/dataset.go
```diff
@@ -811,6 +892,50 @@ type AccessEntry struct {
 	View       *Table              // The view granted access (EntityType must be ViewEntity)
 	Routine    *Routine            // The routine granted access (only UDF currently supported)
 	Dataset    *DatasetAccessEntry // The resources within a dataset granted access.
+	Condition  *Expr               // Condition for the access binding.
+}
+
+// Expr represents the conditional information related to dataset access policies.
+type Expr struct {
+	//Textual representation of an expression in Common Expression Language syntax.
```
alvarowolfx, line 900: nit: missing space here after //

bigquery/dataset_integration_test.go
```diff
@@ -435,6 +436,103 @@ func TestIntegration_DatasetUpdateAccess(t *testing.T) {
...
+	}
+	ctx := context.Background()
+	// Use our test dataset for a base access policy.
+	md, err := dataset.Metadata(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	wantEntry := &AccessEntry{
+
```
alvarowolfx, line 453: nit: extra line break

alvarowolfx (commented):
I wonder if we can check inside the `Create/Update/Metadata` methods if the dataset has any `AccessEntry` with a non `nil` `Condition` and automatically add the `AccessPolicyVersion = 3`, without requiring to add the new `*WithOptions` methods and `DatasetOption` for now.  

One thing that I'm concerned is if users starts to use IAM conditions and have a mixed set of `Datasets` with and without conditions, they have to change all of their code to adopt the new `*WithOptions` methods and check by themselves if the dataset has conditions or not to add the proper `WithAccessPolicyVersion` version. Calling the `Metadata` call might be the most common one that would cause more confusions.

alvarowolfx (approved):
per internal discussion, it's better for users to be explicit about integrating with IAM conditions, so makes sense to have the `*WIthOptions` methods. When they start adopting this feature, they can use Policy version 3 for all of they dataset access usage.

bigquery/dataset.go
shollyman: Resolved.

bigquery/dataset_integration_test.go
shollyman: Resolved.

Issue #11126, opened 2024-11-13
spanner: Fix a possible channel blocking issue


openrefactorymunawar:
## Client

Spanner

## Environment

Result of a static analysis. No runtime information.

## Code and Dependencies

In file: [session.go](https://github.com/googleapis/google-cloud-go/blob/2456b943b7b8aaabd4d8bfb7572c0f477ae0db45/spanner/session.go#L909), there is a possible channel blocking case.

In line [909](https://github.com/googleapis/google-cloud-go/blob/2456b943b7b8aaabd4d8bfb7572c0f477ae0db45/spanner/session.go#L909), in function sessionPool:createMultiplexedSession, there is a send operation but no matching receive operation is found on that channel from any other goroutine.

The sequence starts from function `createMultiplexedSession` being invoked in separate goroutine.

https://github.com/googleapis/google-cloud-go/blob/79556c445ab35c74d52ea80cdd32013f245c880c/spanner/session.go#L752

Inside `createMultiplexedSession`, the code includes the following channel operation.

```go
p.mayGetMultiplexedSession <- true
```

This line sends true to the `mayGetMultiplexedSession` channel to signal that the session request is processed. The intent seems to be to notify other parts of the code that it can now proceed to either handle the session (or manage the error).

The `p.mayGetMultiplexedSession` is an unbuffered channel. 

https://github.com/googleapis/google-cloud-go/blob/79556c445ab35c74d52ea80cdd32013f245c880c/spanner/session.go#L710

So, the send operation will block until another goroutine receives from it. In the case of this code, there appears to be no guarantee that a corresponding receiver will be available to consume the message.

In the following code:

https://github.com/googleapis/google-cloud-go/blob/79556c445ab35c74d52ea80cdd32013f245c880c/spanner/session.go#L762

`pool.mayGetMultiplexedSession` is received. But it is inside a select statement. So, if the `ctx.Done()` becomes ready before the `pool.mayGetMultiplexedSession`, the function would return and the receive operation would not execute. As a a result the send operation would get blocked.

It is observed that all the other send operations on `p.mayGetMultiplexedSession` is done inside a select statement to check for the `context`.
* https://github.com/googleapis/google-cloud-go/blob/79556c445ab35c74d52ea80cdd32013f245c880c/spanner/session.go#L916
* https://github.com/googleapis/google-cloud-go/blob/79556c445ab35c74d52ea80cdd32013f245c880c/spanner/session.go#L938
[...]

PR #11131, opened 2024-11-14
fix(spanner): add safecheck to avoid deadlock when creating multiplex session

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/11126

### DIFF
--- a/spanner/session.go
+++ b/spanner/session.go
@@ -906,7 +906,10 @@ func (p *sessionPool) createMultiplexedSession() {
 				p.mu.Lock()
 				p.multiplexedSessionCreationError = err
 				p.mu.Unlock()
-				p.mayGetMultiplexedSession <- true
+				select {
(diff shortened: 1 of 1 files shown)


spanner/session.go
```diff
@@ -906,7 +906,11 @@ func (p *sessionPool) createMultiplexedSession() {
 				p.mu.Lock()
 				p.multiplexedSessionCreationError = err
 				p.mu.Unlock()
-				p.mayGetMultiplexedSession <- true
+				select {
+				case p.mayGetMultiplexedSession <- true:
+				case <-c.ctx.Done():
+					continue
```
olavloite, line 912: nit: Do we need this line at all? The first line after this block is also just `continue`

olavloite (approved):

spanner/session.go
rahul2393: Not needed, removed

PR #11137, opened 2024-11-14
feat(storage): allow non default service account

frankyn:
Since Go Storage enables DirectPath by default, the client needs to enable the flag to allow non-default service accounts as well. 

In talking with @mohanli-ml, there's an approval process that GCS has already done as a service and we only need to enable it client side. Mohan can i get your approval on this PR to ack this context?

Fixes: #11128
Triggered a continuous job with this PR:
image


mohanli-ml (approved):
Thanks Frank!

tritone (approved):
We should also likely disable this test for cloudtop as well though since it seems to fail there...

frankyn:
Cloudtop works you just need to set environment variable:
```
GOOGLE_CLOUD_DISABLE_DIRECT_PATH=false
```

Direct Connectivity is disabled by default on cloudtop.

PR #11151, opened 2024-11-19
fix(spanner): do not mutate incoming options in createGCPMultiEndpoint

nimf:


rahul2393:
@nimf What is the issue this PR is fixing, can you also please add a failing test it will help me in understanding changes better.

rahul2393:
Thanks, is it possible to add failing test with current code?

nimf:
> Thanks, is it possible to add failing test with current code?

Yes, let me add that.

nimf:
@rahul2393 can you trigger the kokoro run, please?

nimf:
This is another test failing, not related to the chages. It passes locally, but fails in the kokoro env. I will investigate.

PR #11152, opened 2024-11-19
fix(storage): correct direct connectivity check

frankyn:
Raised in: https://github.com/raj-prince/custom-go-client-benchmark/pull/31
Condition didn't check for empty `{}` which failed for multi-region and InterContinental GCE / bucket.
Added an additional integration test to check for this case.


storage/integration_test.go
```diff
@@ -348,12 +348,20 @@ func TestIntegration_DetectDirectConnectivity(t *testing.T) {
...
 				t.Fatalf("CheckDirectConnectivitySupported: %v", err)
 			}
+			newMRBucketName := prefix + "-mr-" + uidSpace.New()
+			newMRBucket := client.Bucket(newMRBucketName)
+			h.mustCreate(newMRBucket, testutil.ProjID(), &BucketAttrs{Location: "US"})
+			defer h.mustDeleteBucket(newMRBucket)
+			err = CheckDirectConnectivitySupported(ctx, newMRBucketName)
+			if err == nil {
+				t.Fatalf("CheckDirectConnectivitySupported: error expected, got nil")
+			}
```
tritone, line 364: So this covers the multi-regional case. Should we add a few more cases? I think we should cover:

1. single region, same region as VM (expect success)
2. single region, diff region from VM (fail)

Same thing with dual region and multiregion.

We could probably refactor this as a table driven test to make it clearer.

frankyn: make sense! i'll work that next. thanks

storage/integration_test.go
```diff
@@ -340,40 +341,81 @@ func TestIntegration_DetectDirectConnectivity(t *testing.T) {
...
+				name:  "multi-region bucket",
+				attrs: &BucketAttrs{Location: "US"},
+			},
+			{
+				name:  "dual-region bucket",
+				attrs: &BucketAttrs{Location: "NAM4"},
+			},
+			{
+				name:  "dual-region bucket",
+				attrs: &BucketAttrs{Location: "NAM4"},
```
tritone: Why is this case repeated?

storage/integration_test.go
```diff
@@ -340,32 +341,78 @@ func TestIntegration_DetectDirectConnectivity(t *testing.T) {
...
+				attrs: &BucketAttrs{Location: otherRegion},
+			},
+			{
+				name:  "multi-region bucket",
+				attrs: &BucketAttrs{Location: "US"},
+			},
+			{
+				name:                     "dual-region bucket",
+				attrs:                    &BucketAttrs{Location: "NAM4"},
+				expectDirectConnectivity: true,
```
frankyn, line 383: I didn't expect NAM4 to return true.

storage/integration_test.go
frankyn: Typo :)

storage/integration_test.go
```diff
@@ -340,40 +341,81 @@ func TestIntegration_DetectDirectConnectivity(t *testing.T) {
...
-				t.Fatalf("CheckDirectConnectivitySupported: error expected, got nil")
-			}
-		} else {
-			err = CheckDirectConnectivitySupported(ctx, bucket)
-			if err == nil {
-				t.Fatal("CheckDirectConnectivitySupported: expected error but none returned")
-			}
-			if err != nil && !strings.Contains(err.Error(), "direct connectivity not detected") {
-				t.Fatalf("CheckDirectConnectivitySupported: failed on a different error %v", err)
+		if v, exists := attrs.Value("cloud.platform"); !exists || v.AsString() == "gcp_compute_engine" {
```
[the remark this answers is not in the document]
frankyn: I'll trigger a job in kokoro; i was testing this in GCE first

frankyn:
Kokoro failed on 
image
Looks concerning.

raj-prince (approved):

storage/integration_test.go
```diff
@@ -340,32 +351,67 @@ func TestIntegration_DetectDirectConnectivity(t *testing.T) {
...
+			name                     string
+			attrs                    *BucketAttrs
+			expectDirectConnectivity bool
+		}{
+			{
+				name:                     "co-located-bucket",
+				attrs:                    &BucketAttrs{Location: "us-west1"},
+				expectDirectConnectivity: true,
+			},
+			{
```
tritone, line 370: Maybe have a TODO to add NAM4 with true as well?

tritone (approved):

storage/integration_test.go
frankyn: `NAM4` wouldn't work in this case since it's a pairing of us-central1 and us-east1.
image


Need to create a new pairing with customPlacementConfig; I'll add a todo for customPlacement though.

tritone: Ahh got it. Sounds good.

PR #11159, opened 2024-11-20
feat(auth): add universe domain support to mTLS

quartzmo:
* Remove runtime error for mTLS usage with non-GDU universe domain
* Add UNIVERSE_DOMAIN placeholder support for Options.DefaultMTLSEndpoint

refs: googleapis/google-api-go-client#2880

### DIFF
--- a/auth/grpctransport/grpctransport.go
+++ b/auth/grpctransport/grpctransport.go
@@ -197,8 +197,7 @@ type InternalOptions struct {
 	// DefaultEndpointTemplate combined with UniverseDomain specifies
 	// the default endpoint.
 	DefaultEndpointTemplate string
-	// DefaultMTLSEndpoint combined with UniverseDomain specifies the
-	// default mTLS endpoint.
+	// DefaultMTLSEndpoint specifies the default mTLS endpoint.
 	DefaultMTLSEndpoint string
 	// DefaultScopes specifies the default OAuth2 scopes to be used for a
 	// service.
--- a/auth/grpctransport/grpctransport.go
+++ b/auth/grpctransport/grpctransport.go
@@ -245,7 +245,7 @@ func dial(ctx context.Context, secure bool, opts *Options) (*grpc.ClientConn, er
 	}
 	if io := opts.InternalOptions; io != nil {
 		tOpts.DefaultEndpointTemplate = io.DefaultEndpointTemplate
-		tOpts.DefaultMTLSEndpointTemplate = io.DefaultMTLSEndpoint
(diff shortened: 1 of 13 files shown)


auth/httptransport/httptransport.go
```diff
@@ -200,7 +201,7 @@ func NewClient(opts *Options) (*http.Client, error) {
 	}
 	if io := opts.InternalOptions; io != nil {
 		tOpts.DefaultEndpointTemplate = io.DefaultEndpointTemplate
-		tOpts.DefaultMTLSEndpoint = io.DefaultMTLSEndpoint
+		tOpts.DefaultMTLSEndpointTemplate = io.DefaultMTLSEndpoint
```
codyoss, line 204: I don't think this is safe to move without us making a generator change first and landing that.

codyoss: nvm

auth/grpctransport/grpctransport.go
```diff
@@ -244,7 +245,7 @@ func dial(ctx context.Context, secure bool, opts *Options) (*grpc.ClientConn, er
 	}
 	if io := opts.InternalOptions; io != nil {
 		tOpts.DefaultEndpointTemplate = io.DefaultEndpointTemplate
-		tOpts.DefaultMTLSEndpoint = io.DefaultMTLSEndpoint
+		tOpts.DefaultMTLSEndpointTemplate = io.DefaultMTLSEndpoint
```
quartzmo, line 248: @codyoss Because the new internal options "template" is set here from the old external options "endpoint", this means the `internaloption.WithDefaultMTLSEndpointTemplate` field added in https://github.com/googleapis/google-api-go-client/pull/2880 will not be used. Instead, the workaround will be to set a templatized value (something like `https://test.mtls.UNIVERSE_DOMAIN/`) in `internaloption.WithDefaultMTLSEndpoint`. So should I close that PR?

quartzmo: Also, to keep things consistent, I could revert `tOpts.DefaultMTLSEndpointTemplate` to `tOpts.DefaultMTLSEndpoint`. The name of the internal option does not matter right now.

codyoss: SGTM

quartzmo: Both done, thanks.

codyoss (approved):

viacheslav-rostovtsev (commented):
LGTM

Issue #11161, opened 2024-11-20
auth/oauth2adapt: fatal error: concurrent map writes


bhshkh:
## Client

auth/oauth2adapt

## Environment
```
$ go version
go version go1.21.13 linux/amd64
```
## Code and Dependencies
auth/oauth2adapt 0.2.5

## Expected behavior

Token call succeeds

## Actual behavior

Seeing fatal error: concurrent map writes

```
fatal error: concurrent map writes

goroutine 235 [running]:
cloud.google.com/go/auth/oauth2adapt.(*tokenSourceAdapter).Token(0x416025?)
	/******/go/pkg/mod/cloud.google.com/go/auth/oauth2adapt@v0.2.5/oauth2adapt.go:107 +0x1a7
google.golang.org/api/transport/grpc.dryRunAsync.Token({{0x13b3b60, 0xc0002aa430}, {0x13b79a0, 0xc00064ee40}, 0xc0003ac360})
	/******/google-api-go-client/transport/grpc/dial.go:290 +0x37
google.golang.org/grpc/credentials/oauth.TokenSource.GetRequestMetadata({{0x13b3ae0?, 0xc00064ee70?}}, {0x13d0118, 0xc0ebe8c750}, {0x0?, 0xc001938130?, 0x410485?})
	/******/go/pkg/mod/google.golang.org/grpc@v1.67.1/credentials/oauth/oauth.go:42 +0x38
google.golang.org/api/transport/grpc.grpcTokenSource.GetRequestMetadata({{{0x13b3ae0, 0xc00064ee70}}, {0x0, 0x0}, {0x0, 0x0}}, {0x13d0118?, 0xc0ebe8c750?}, {0xc000792590, 0x1, ...})
	/******/google-api-go-client/transport/grpc/dial.go:423 +0x76
google.golang.org/grpc/internal/transport.(*http2Client).getTrAuthData(0xc000034480, {0x13d0118, 0xc0ebe8c750}, {0xc00005a780, 0x3b})
	/******/go/pkg/mod/google.golang.org/grpc@v1.67.1/internal/transport/http2_client.go:662 +0x14e
google.golang.org/grpc/internal/transport.(*http2Client).createHeaderFields(0xc000034480, {0x13d0118, 0xc0ebe8c720}, 0xc001938b08)
	/******/go/pkg/mod/google.golang.org/grpc@v1.67.1/internal/transport/http2_client.go:544 +0x125
[...]

PR #11164, opened 2024-11-21
fix(auth/oauth2adapt): copy map in tokenSourceAdapter.Token

quartzmo:
fixes: #11161

### DIFF
--- a/auth/oauth2adapt/oauth2adapt.go
+++ b/auth/oauth2adapt/oauth2adapt.go
@@ -100,8 +100,13 @@ func (ts *tokenSourceAdapter) Token() (*oauth2.Token, error) {
 		Expiry:      tok.Expiry,
 	}
 	// Preserve token metadata.
-	metadata := tok.Metadata
-	if metadata != nil {
+	m := tok.Metadata
+	if m != nil {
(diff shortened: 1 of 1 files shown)


auth/oauth2adapt/oauth2adapt.go
```diff
@@ -100,8 +100,13 @@ func (ts *tokenSourceAdapter) Token() (*oauth2.Token, error) {
 		Expiry:      tok.Expiry,
 	}
 	// Preserve token metadata.
-	metadata := tok.Metadata
-	if metadata != nil {
+	m := tok.Metadata
+	if m != nil {
+		// Copy map to avoid concurrent map writes error (#11161).
+		metadata := make(map[string]interface{})
```
codyoss, line 106: nit: we know at least the current capacity of this map. Maybe we should use that information and allocate +2 incase we add both keys?

codyoss (approved):

auth/oauth2adapt/oauth2adapt.go
quartzmo: Fixed, thanks!

codyoss (approved):

PR #11175, opened 2024-11-22
fix(pubsub): fix defer call in for loop

zchee:
SSIA. Avoid the `defer` function call in for loop.


zchee:
@hongalex gentle ping

hongalex:
Apologies for missing this the first time around. Could you explain in more detail why this is necessary? Calling `defer` in a for loop can lead to issues with freeing memory, but in this case, I believe the behavior is correct and wrapping it in anonymous function makes it harder to read.

zchee:
@hongalex Example playground is here: https://go.dev/play/p/tM6Gt11reCo

I thought you assumed `deferAnonFunc` behavior, but actually, When calling `defer` within a for loop, added functions to the `defer` stack are called in reverse order. (`deferForLoop` behavior
If we want to keep the order of `bmsgs`, we should enclose it in an anonymous function.

pubsub/topic.go
```diff
@@ -1083,15 +1083,22 @@ func (t *Topic) initBundler() {
...
-				defer m.createSpan.End()
-				defer m.createSpan.AddEvent(eventPublishEnd)
+				fns[i] = func() {
+					m.createSpan.AddEvent(eventPublishEnd)
+					m.createSpan.End()
+				}
 			}
 		}
 		t.publishMessageBundle(ctx, bmsgs)
+		for _, fn := range fns {
```
zchee, line 1099: If `t.enableTracing` is false, `fns` is nil. So this for loop is not called.

hongalex:
Yeah I definitely overlooked the order of when the message spans are ended when using defer. In my mind, it's not a super pressing bug since the events are batched anyway on the exporter so the timing is fairly tight, but it would be nice to get correct.

If that's the case, I would prefer iterating over `bmsgs` rather than allocating space for the function calls. wdyt?
```go
t.publishMessageBundle(ctx, bmsgs)
if t.enableTracing {
	for i, m := range bmsgs {
		m.createSpan.End()
		m.createSpan.AddEvent(eventPublishEnd)
        }
}
```

zchee:
@hongalex I agreed your opinion.

But just for confirm, you means calling `AddEvent` before `End` instead of?

```go
t.publishMessageBundle(ctx, bmsgs)
if t.enableTracing {
	for i, m := range bmsgs {
		m.createSpan.AddEvent(eventPublishEnd)
		m.createSpan.End()
        }
}
```

hongalex:
> But just for confirm, you means calling `AddEvent` before `End` instead of?

Sorry, no I didn't. I copy and pasted the order from the defer side (because the opposite order is correct) but in this case we should be calling AddEvent and then End.
Thanks!

hongalex (approved):

PR #11184, opened 2024-11-25
feat(storage/dataflux): increase test coverage for fast-list

akansha1812:
BEGIN_COMMIT_OVERRIDE
test(storage/dataflux): increase test coverage for fast-list
END_COMMIT_OVERRIDE

feat:increase test coverage for fast-list. 
Dataflux fast-listing leverages worksteal algorithm to quickly list objects in a bucket by running several parallel processes.

Adding unit tests that uses storage/emulator.sh ( [storage-testbench](https://github.com/googleapis/storage-testbench)) to improve test coverage.
Follow up PR to : https://github.com/googleapis/google-cloud-go/pull/10966

### DIFF
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -452,6 +452,9 @@ func (c *grpcStorageClient) ListObjects(ctx context.Context, bucket string, q *Q
 		ctx = setUserProjectMetadata(ctx, s.userProject)
 	}
 	fetch := func(pageSize int, pageToken string) (token string, err error) {
+		// Add trace span around List API call within the fetch.
+		ctx, _ = startSpan(ctx, "grpcStorageClient.ObjectsListCall")
+		defer func() { endSpan(ctx, err) }()
 		var objects []*storagepb.Object
 		var gitr *gapic.ObjectIterator
 		err = run(it.ctx, func(ctx context.Context) error {
--- a/storage/http_client.go
+++ b/storage/http_client.go
@@ -344,6 +344,10 @@ func (c *httpStorageClient) ListObjects(ctx context.Context, bucket string, q *Q
 		it.query = *q
 	}
 	fetch := func(pageSize int, pageToken string) (string, error) {
+		var err error
+		// Add trace span around List API call within the fetch.
+		ctx, _ = startSpan(ctx, "httpStorageClient.ObjectsListCall")
+		defer func() { endSpan(ctx, err) }()
 		req := c.raw.Objects.List(bucket)
 		if it.query.SoftDeleted {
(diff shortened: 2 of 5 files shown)


storage/dataflux/fast_list_test.go
```diff
@@ -192,3 +200,249 @@ func TestNewLister(t *testing.T) {
...
+		}
+		if err := createObject(ctx, bucketHandle, 2, ""); err != nil {
+			t.Fatalf("unable to create objects: %v", err)
+		}
+		c := NewLister(client, &ListerInput{BucketName: bucket})
+		defer c.Close()
+		childCtx, cancel := context.WithCancel(ctx)
+		cancel()
+		result, err := c.NextBatch(childCtx)
+		if err != nil && !strings.Contains(err.Error(), context.Canceled.Error()) {
```
akansha1812, line 221: @BrennaEpp  for comment "This error check should use errors.Is"

GRPC has context.Canceled.Error() in desc and not as error. So this succeeds for http but fails for GRPC.

storage/dataflux/integration_test.go
```diff
@@ -89,48 +95,65 @@ func TestIntegration_NextBatch_All(t *testing.T) {
...
+		counter++
+		totalObjects += len(objects)
 	}
-	if totalObjects != wantObjects {
-		t.Errorf("expected to receive %d objects in results, got %d objects in results", wantObjects, totalObjects)
-
+	if totalObjects != numObjectsPrefix {
+		t.Errorf("expected to receive %d objects in results, got %d objects in results", numObjectsPrefix, totalObjects)
+	}
+	if df.method != worksteal {
```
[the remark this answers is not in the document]
BrennaEpp: I mean flaky in the sense that we expect that worksteal will always finish faster than sequential, 100% of the time, in this scenario. I guess it is a lot of objects and we would expect that - anyway we can leave it in and if we get errors deal with it then.

storage/dataflux/fast_list_test.go
tritone: You can use the status package to check the error without string-matching, for example like this test here: https://github.com/googleapis/google-cloud-go/blob/main/storage/client_test.go#L1428

akansha1812: Thanks! Updated it.

akansha1812:
The vet failure is in the main branch https://github.com/googleapis/google-cloud-go/pull/11195

bhshkh:
> The vet failure is in the main branch #11195

Fix is in https://github.com/googleapis/google-cloud-go/pull/11202

PR #11121, opened 2024-11-12
chore(storage): store credentials as auth.Credentials

BrennaEpp:
Migrate from google.Credentials to auth.Credentials where posible

PR #11191, opened 2024-11-25
fix(storage): migrate oauth2/google usages to cloud.google.com/go/auth

quartzmo:
This PR is intended as a replacement for #11121. It uses a new [internaloption.AuthCreds](https://pkg.go.dev/google.golang.org/api@v0.228.0/option/internaloption#AuthCreds) helper func added to the `google.golang.org/api/option` package to support sourcing credentials from both old and new auth options. You can see that helper here: https://github.com/googleapis/google-api-go-client/pull/2891.


storage/storage.go
```diff
@@ -117,13 +116,22 @@ type Client struct {
...
 	tc storageClient
 }
 
+// credsJSON returns the raw JSON of the Client's creds, or an empty slice
+// if no credentials JSON is available.
+func (c Client) credsJSON() []byte {
+	if c.creds != nil && len(c.creds.JSON()) > 0 {
+		return c.creds.JSON()
+	}
+	return []byte{}
```
BrennaEpp, line 132: I think this would be nicer returning the creds and a bool so that callers can check `creds, ok := credsJSON()`
```suggestion
func (c Client) credsJSON() ([]byte, bool) {
	if c.creds != nil && len(c.creds.JSON()) > 0 {
		return c.creds.JSON(), true
	}
	return []byte{}, false
```

storage/integration_test.go
```diff
@@ -5728,7 +5751,7 @@ func TestIntegration_PostPolicyV4_WithCreds(t *testing.T) {
 				}
 			})
 		}
-	}, option.WithCredentials(creds))
+	}, option.WithAuthCredentials(creds))
```
BrennaEpp, line 5754: Did you try running the integration tests without changes? I would like to make sure they pass even using the old auth before switching them over.

Signed URL stuff in particular, do you think it would be useful to test with both the old and new auth paths?

storage/bucket_test.go
```diff
@@ -1291,11 +1291,16 @@ func TestDetectDefaultGoogleAccessID(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
+			jsonBytes := []byte(tc.creds(tc.serviceAccount))
+			creds, err := credentials.DetectDefault(&credentials.DetectOptions{
+				CredentialsJSON: jsonBytes,
+			})
```
BrennaEpp, line 1297: Is it possible to get auth credentials here without this function?

BrennaEpp (commented):
In general looks good, thank you for the changes. Just a note on making sure tests pass with old auth creds sent in by users.

storage/storage.go
```diff
@@ -154,10 +162,10 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 
 		// Don't error out here. The user may have passed in their own HTTP
 		// client which does not auth with ADC or other common conventions.
-		c, err := transport.Creds(ctx, opts...)
+		c, err := internaloption.AuthCreds(ctx, opts)
```
quartzmo, line 165: Does `internaloption.EnableNewAuthLibrary()` need to be added here (or elsewhere)?

storage/bucket_test.go
quartzmo: No, `DetectDefault` is the right public function for getting credentials from JSON bytes (or filepath to JSON bytes, etc). The name may seem wrong, but with the `CredentialsJSON` option it does not attempt to find ADC.

storage/integration_test.go
quartzmo: This PR is intended as guidance only. Think of it as a suggestion for how you can transition to the new Auth lib.

storage/storage.go
quartzmo: SGTM!

storage/integration_test.go
quartzmo: Please disregard previous response about guidance only. I'm updating this PR to provide an acceptable solution. Restoring old auth path in addition to adding new auth path in affected tests.

storage/storage.go
BrennaEpp: I don't think so, `internaloption.AuthCreds(ctx, opts)` will call into the new auth library no matter what it seems. What maybe should be added is a check for the env var to disable new auth if we still need that in case we need to debug?

BrennaEpp:
Just adding a note here, integration tests pass on the latest commit to this PR.

BrennaEpp (approved):
Thank you!! Sorry for the delay

PR #11197, opened 2024-11-26
fix(storage): monitored resource detection

frankyn:
Summary of changes
- Fixed resource attribute names detected by GCP resource detector by using `_` instead of `.`, e.g. "host_id" instead of "host.id" resulting in storageClient always using default attribute values.
- Prioritize project ID from ADC before using resource detector project
- Refactored preparedResource into storageMonitoredResource to separate responsibilities
- Explicitly name gRPC metrics enabled instead of relying on `opentelemetry.DefaultMetrics()`  
- Introduced integration tests for gRPC Metrics are collected through a secondary Reader 
- Used struct embeddings to reduce boiler plate on log suppression exporter (thanks @BrennaEpp)


frankyn:
Spot checked resulting monitored resource values to double check: 
image

storage/grpc_client.go
```diff
@@ -118,6 +119,24 @@ type grpcStorageClient struct {
 	settings *settings
 }
 
+func enableClientMetrics(ctx context.Context, s *settings, config storageConfig) (*metricsContext, error) {
+	var project string
+	// TODO: use new auth client
```
tritone, line 124: do we have a bug for this?

storage/grpc_metrics.go
```diff
@@ -39,98 +39,76 @@ const (
...
-		}
+func (smr *storageMonitoredResource) exporter() (metric.Exporter, error) {
+	exporter, err := mexporter.New(
+		mexporter.WithProjectID(smr.project),
+		mexporter.WithMetricDescriptorTypeFormatter(metricFormatter),
+		mexporter.WithCreateServiceTimeSeries(),
+		mexporter.WithMonitoredResourceDescription(monitoredResourceName, []string{"project_id", "location", "cloud_platform", "host_id", "instance_id", "api"}),
+	)
+	if err != nil {
+		return nil, err
```
tritone, line 61: Let's wrap this with a string like "storage: creating metrics exporter: %w"

storage/grpc_metrics.go
```diff
@@ -39,98 +39,76 @@ const (
...
+	smr := &storageMonitoredResource{
+		instance: uuid.New().String(),
+		api:      api,
+		project:  project,
+	}
 	s := detectedAttrs.Set()
-	p, present := s.Value("cloud.account.id")
-	if present {
-		preparedResource.projectToUse = p.AsString()
+	if p, present := s.Value("cloud.account.id"); present && smr.project == "" {
```
tritone, line 77: Maybe add a comment about what this is trying to do?

storage/grpc_metrics_test.go
```diff
@@ -34,15 +34,19 @@ func TestMetricFormatter(t *testing.T) {
...
 		desc               string
+		project            string
+		api                string
 		detectedAttributes []attribute.KeyValue
 		wantAttributes     attribute.Set
 	}{
 		{
-			desc: "default values set when GCP attributes are not detected",
+			desc:    "default values set when GCP attributes are not detected",
+			project: "project-id",
```
tritone, line 48: project and api seem the same in all test cases, should they just be removed as fields here? Or did you want to test other values?

storage/integration_test.go
```diff
@@ -413,6 +415,114 @@ func TestIntegration_DoNotDetectDirectConnectivityWhenDisabled(t *testing.T) {
...
+		}
+		for k, v := range metricCheck {
+			if !v {
+				t.Errorf("metric %v not found", k)
+			}
+		}
+	}, withTestMetricReader(mr))
+}
+
+func TestIntegration_MetricsEnablementInGCE(t *testing.T) {
```
tritone, line 451: Do we expect this test to run without skipping in Kokoro?

storage/integration_test.go
```diff
@@ -413,6 +415,114 @@ func TestIntegration_DoNotDetectDirectConnectivityWhenDisabled(t *testing.T) {
...
+		metricCheck := map[string]bool{
+			"grpc.client.attempt.started":                            false,
+			"grpc.client.attempt.duration":                           false,
+			"grpc.client.attempt.sent_total_compressed_message_size": false,
+			"grpc.client.attempt.rcvd_total_compressed_message_size": false,
+			"grpc.client.call.duration":                              false,
+			"grpc.lb.rls.cache_entries":                              false,
+			"grpc.lb.rls.cache_size":                                 false,
+			"grpc.lb.rls.default_target_picks":                       false,
+			// TODO: determine a way to force these metrics to be collected
```
tritone, line 505: Remove commented values here; please file an issue if there is other stuff we need to add later.

tritone (commented):
Few minor questions/nits, otherwise looks good.

storage/integration_test.go
tritone: No, looks fine to me, just wanted to confirm.

tritone (approved):

PR #11200, opened 2024-11-27
fix(storage): add backoff to gRPC write retries

BrennaEpp:
- Integration tests pass
- Emulated/conformance tests pass except for the following which I assume is unrelated:
```
--- FAIL: TestRetryReadStallEmulated (5.52s)
    client_test.go:1496: NewReader: context deadline exceeded
```


storage/grpc_client.go
```diff
@@ -2370,3 +2382,78 @@ func checkCanceled(err error) error {
...
+			w.backoff.Initial = w.settings.retry.backoff.Initial
+			w.backoff.Max = w.settings.retry.backoff.Max
+		}
+	}
+}
+
+// shouldRetry determines if a retry is necessary and if so waits the appropriate
+// amount of time. It returns true if the error is retryable or the error to be
+// surfaced to the user if not.
+func (w *gRPCWriter) shouldRetry(ctx context.Context, err error) (bool, error) {
```
tritone, line 2411: Can we name this something else? Just confusing to conflate this with the ShouldRetry that only takes an err and returns a bool, especially since the backoff pause happens inside this func.

storage/grpc_client.go
```diff
@@ -2370,3 +2382,78 @@ func checkCanceled(err error) error {
...
+		retryable = false
+
+		if w.lastErr != nil {
+			return false, fmt.Errorf("retry failed with %v; last error: %w", ctxErr, w.lastErr)
+		}
+		return false, ctxErr
+	}
+
+	if retryable {
+		p := w.backoff.Pause()
```
tritone, line 2448: As discussed, maybe we can add a mock test in this PR to ensure that backoff.Pause is actually called?

tritone (commented):
Few minor comments but overall the approach looks good.

tritone:
Also, are you sure your testbench is up to date? That would be a reason that the ReadStallTimeout test might not work locally. Looks like it runs fine in Kokoro so I am not concenred.

storage/grpc_client.go
BrennaEpp: Yeah, that's fair. I'll think of something better.

storage/grpc_client.go
```diff
@@ -2370,3 +2382,78 @@ func checkCanceled(err error) error {
 
 	return err
 }
+
+func (w *gRPCWriter) initializeRetryConfig() {
+	if w.attempts == 0 {
```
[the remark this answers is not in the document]
BrennaEpp: Changed this

BrennaEpp:
> Also, are you sure your testbench is up to date? That would be a reason that the ReadStallTimeout test might not work locally. Looks like it runs fine in Kokoro so I am not concenred.

It is up to date. It's possible it just doesn't run fast enough or something like that locally; I'm also not concerned if they are passing in kokoro.

storage/grpc_client.go
BrennaEpp: I made a mock for the backoff and had to change more than anticipated to be able to inject it, PTAL

danielduhh (commented):
Can we run an e2e test to verify it fixes b/379925581?

storage/grpc_client.go
```diff
@@ -2383,77 +2379,88 @@ func checkCanceled(err error) error {
...
 	}
 }
 
-// shouldRetry determines if a retry is necessary and if so waits the appropriate
-// amount of time. It returns true if the error is retryable or the error to be
-// surfaced to the user if not.
-func (w *gRPCWriter) shouldRetry(ctx context.Context, err error) (bool, error) {
+// retriable determines if a retry is necessary and if so returns a nil error;
+// otherwise it returns the error to be surfaced to the user.
+func (retry *uploadBufferRetryConfig) retriable(ctx context.Context, err error) error {
```
tritone, line 2417: It seems a little odd to have this return the error to be resurfaced. Maybe just have it return a bool, and you can use retry.lastErr to set the error to return if need be?

Also not a big deal if you want to leave this as-is for now, since we'll be overwriting with the refactored version anyway.

storage/grpc_client_test.go
```diff
@@ -218,3 +225,162 @@ func TestBytesCodecV2(t *testing.T) {
...
+			}
+
+			if got, want := backoff.pauseCallsCount, test.expectedRetries; got != want {
+				t.Errorf("backoff.Pause was called %d times, expected %d", got, want)
+			}
+		})
+	}
+}
+
+type MockWriteStream struct {
```
tritone, line 320: This looks useful to test other behavior for writer; we should file a ticket to add.

tritone (approved):
Looks good, one minor suggestion but I think maybe just for future reference.

storage/grpc_client.go
BrennaEpp: That's fair; I think I may have been too deep in this. Although, this returns a formatted error for when max attempts are reached, which you wouldn't know outside this method unless you returned two bools... but I guess adding the number of attempts to the error may be useful for all cases anyway. I'll leave as-is and consider this feedback if relevant to the other version.

PR #11208, opened 2024-12-02
skip Errorf for a frequent error that is never read

sduchesneau:
While hunting down memory allocations, we found lots of calls to Errorf() that were useless (because the error is never actually read, it is only checked for nil condition)
Reusing the same error prevents all these allocations, while keeping a sane behavior for future use.


tritone (commented):
Code change seems fine to me. I am more curious about how you are seeing this error occur internally to the library so frequently. Could you file an issue and describe what kind of data you are downloading, whether you are using storage.NewReader or NewRangeReader, whether the object has a checksum present when you look at it in the console? I would expect this failure case not to be very common in general.

cpriti-os:
@sduchesneau I am cleaning up old PRs, will close this in another week if there is no response.

cpriti-os:
Closing this. (non-responsive)

Issue #11211, opened 2024-12-03
storage: return file metadata on read


jshore1296:
## Is your feature request related to a problem? Please describe.
Currently, it requires two API calls to open a file and retrieve its metadata.  It is cheaper at scale to return metadata when opening the file.  This is supported at the API level by both xml and gRPC, but is not exposed by this library.

## Describe the solution you'd like
Add a field to the `ReaderObjectAttrs` that stores the Metadata on the file being read, and expose this to the caller similar to how `Size` and other such options are exposed.

tritone:
#11212 has been merged and will go out tomorrow in storage/v1.49.0. Thanks for the contribution!

PR #11212, opened 2024-12-03
feat(storage): return file metadata on read

jshore1296:
implements #11211

This isn't in a mergeable state quite yet, but I wanted to start the discussion.  This is only supported by the XML and gRPC apis.  I'm unsure of how to handle the JSON api.  Should I just document that the JSON api does not support metadata on read?  Or is there a more elegant way to block the user from trying to read it and being surprised by it being empty?


tritone:
> implements #11211
> 
> This isn't in a mergeable state quite yet, but I wanted to start the discussion. This is only supported by the XML and gRPC apis. I'm unsure of how to handle the JSON api. Should I just document that the JSON api does not support metadata on read? Or is there a more elegant way to block the user from trying to read it and being surprised by it being empty?

I'd say we can just document that it is unavailable via the JSON API. There are other fields in ReaderObjectAttrs which are only populated sometimes.

I need to look at the rest of the code more deeply; just let me know when it's ready for review.

jshore1296:
Alright, this is ready for review.

jshore1296:
@tritone sending a ping in case the previous comment didn't notify you

storage/client_test.go
```diff
@@ -795,6 +795,62 @@ func TestOpenReaderEmulated(t *testing.T) {
 	})
 }
 
+func TestOpenReaderEmulated_Metadata(t *testing.T) {
+	transportClientTest(skipHTTP("metadata on read not supported by JSON api"), t, func(t *testing.T, ctx context.Context, project, bucket string, client storageClient) {
```
tritone, line 799: I believe the HTTP client should use XML by default, does the test not pass with a default HTTP client?

storage/client_test.go
```diff
@@ -795,6 +795,62 @@ func TestOpenReaderEmulated(t *testing.T) {
...
+		}
+		w := veneerClient.Bucket(bucket).Object(want.Name).NewWriter(ctx)
+		if _, err := w.Write(randomBytesToWrite); err != nil {
+			t.Fatalf("failed to populate test data: %v", err)
+		}
+		if err := w.Close(); err != nil {
+			t.Fatalf("closing object: %v", err)
+		}
+		if _, err := veneerClient.Bucket(bucket).Object(want.Name).Update(ctx, ObjectAttrsToUpdate{
+			Metadata: map[string]string{
```
tritone, line 820: Let's do at least 2 keys in here just to make sure the decoding logic works correctly.

storage/grpc_client.go
```diff
@@ -1133,6 +1133,7 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
 			CacheControl:    obj.GetCacheControl(),
 			LastModified:    obj.GetUpdateTime().AsTime(),
 			Metageneration:  obj.GetMetageneration(),
+			Metadata:        obj.GetMetadata(),
```
tritone, line 1136: I was concerned this wouldn't be sufficient because we didn't have it in the unmarshaler for gRPC, but it looks like we did that part already so this should be fine.

storage/integration_test.go
```diff
@@ -204,10 +205,10 @@ func initIntegrationTest() func() error {
 		if err != nil {
 			log.Fatalf("NewStorageControlClient: %v", err)
 		}
-		if err := client.Bucket(bucketName).Create(ctx, testutil.ProjID(), nil); err != nil {
+		if err := client.Bucket(bucketName).Create(ctx, testutil.ProjID(), &BucketAttrs{SoftDeletePolicy: &SoftDeletePolicy{RetentionDuration: 0}}); err != nil {
```
tritone, line 208: I think this should maybe be a separate change in our test harness? Was it impossible to delete the bucket if this wasn't included?

storage/integration_test.go
```diff
@@ -5041,7 +5042,58 @@ func TestIntegration_ReaderAttrs(t *testing.T) {
 			Metageneration:  attrs.Metageneration,
 			CRC32C:          crc32c(c),
 		}
-		if got != want {
+		if !reflect.DeepEqual(got, want) {
```
tritone, line 5045: Maybe we can use https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff as elsewhere in these tests?

storage/integration_test.go
```diff
@@ -5041,7 +5042,58 @@ func TestIntegration_ReaderAttrs(t *testing.T) {
...
+		}
+	})
+}
+
+func TestIntegration_ReaderAttrs_Metadata(t *testing.T) {
+	multiTransportTest(skipJSONReads(context.Background(), "metadata on read not supported on JSON api"), t, func(t *testing.T, ctx context.Context, bucket, _ string, client *Client) {
+		bkt := client.Bucket(bucket)
+
+		const defaultType = "text/plain"
+		o := bkt.Object("reader-attrs-obj")
```
tritone, line 5056: Let's use a different object name; repeatedly overwriting the same object can be problematic.

storage/integration_test.go
```diff
@@ -5041,7 +5042,58 @@ func TestIntegration_ReaderAttrs(t *testing.T) {
...
+		c := randomContents()
+		if err := writeObject(ctx, o, defaultType, c); err != nil {
+			t.Errorf("Write for %v failed with %v", o.ObjectName(), err)
+		}
+		defer func() {
```
tritone, line 5061: Use t.Cleanup to ensure this is run.

storage/integration_test.go
```diff
@@ -5041,7 +5042,58 @@ func TestIntegration_ReaderAttrs(t *testing.T) {
...
+			if err := o.Delete(ctx); err != nil {
+				log.Printf("failed to delete test object: %v", err)
+			}
+		}()
+
+		oa, err := o.Update(ctx, ObjectAttrsToUpdate{Metadata: map[string]string{"Custom-Key": "custom-value"}})
```
tritone, line 5067: Again, let's use at least 2 keys here.

storage/integration_test.go
```diff
@@ -5041,7 +5042,58 @@ func TestIntegration_ReaderAttrs(t *testing.T) {
...
+			ContentType:     attrs.ContentType,
+			ContentEncoding: attrs.ContentEncoding,
+			CacheControl:    got.CacheControl, // ignored, tested separately
+			LastModified:    got.LastModified, // ignored, tested separately
+			Generation:      attrs.Generation,
+			Metadata:        map[string]string{"Custom-Key": "custom-value"},
+			Metageneration:  attrs.Metageneration,
+			CRC32C:          crc32c(c),
+		}
+		if !reflect.DeepEqual(got, want) {
```
tritone, line 5096: Use cmp.Diff and let's only check the Metadata field in this test.

storage/reader.go
```diff
@@ -59,6 +59,10 @@ type ReaderObjectAttrs struct {
 	// Generation is the generation number of the object's content.
 	Generation int64
 
+	// Metadata represents user-provided metadata, in key/value pairs.
+	// Not supported by the JSON api.  Use ObjectHandle.Attrs instead.
```
tritone, line 63: I would rephrase this as follows:

```
// Metadata represents user-provided metadata, in key/value pairs.
//
// It can be nil if no metadata is present, or if the client uses the JSON
// API for downloads. Only the XML and gRPC APIs support getting
// custom metadata via the Reader; for JSON make a separate call to 
// ObjectHandle.Attrs.
```

storage/retry_conformance_test.go
```diff
@@ -586,6 +587,9 @@ func TestRetryConformance(t *testing.T) {
 	if host == "" {
 		t.Skip("This test must use the testbench emulator; set STORAGE_EMULATOR_HOST to run.")
 	}
+	if runtime.GOOS == "darwin" {
```
tritone, line 590: Let's remove this from the PR; seems non-relevant to this particular issue.

tritone (commented):
A few comments; overall looks good. Thanks for adding tests!

storage/client_test.go
jshore1296: right you are.

storage/client_test.go
jshore1296: 👍

storage/grpc_client.go
jshore1296: Yeah, I was surprised that was already in there!

storage/integration_test.go
jshore1296: Oops, this is leftover.  My org has a policy set that made this necessary for me to be able to run the tests.  I'll remove this before merge.

storage/integration_test.go
jshore1296: done.

storage/integration_test.go
jshore1296: Done.

storage/integration_test.go
jshore1296: Done.

storage/integration_test.go
jshore1296: Done.

storage/integration_test.go
jshore1296: Done.

storage/reader.go
jshore1296: Done.

storage/retry_conformance_test.go
jshore1296: 👍 there was handling for this in the test script so I moved it in here, good to remove though.

storage/reader_test.go
```diff
@@ -470,6 +472,72 @@ func TestContentEncodingGzipWithReader(t *testing.T) {
...
+		rd, err := obj.NewReader(ctx)
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer rd.Close()
+
+		expectedMetadata := map[string]string{
+			"Custom-Metadata-Key": "custom-metadata-value",
+		}
+		if !reflect.DeepEqual(rd.Attrs.Metadata, expectedMetadata) {
```
tritone, line 527: One more place to switch to cmp.Diff

tritone (approved):
One more minor nit; otherwise looks good. Thanks for your contribution!

storage/reader_test.go
jshore1296: Old habits die hard, haha.

tritone:
Okay so looks like there are some presubmit issues:

1. Check for formatting; you can fix this with gofmt.
2. Technically this is a breaking change to Reader because, with an exported map field, ReaderObjectAttrs becomes a non-comparable type.

(2) is kind of an annoying check (who is comparing storage.Reader using comparison operators?) but I think we should fix it. Perhaps we can make a GetMetadata() method on Reader rather than making it a field of ReaderObjectAttrs?

Sorry for not flagging that sooner and LMK if that is sufficient clarity on what needs to change...

jshore1296:
makes sense @tritone .  Pushed the updates, hopefully I caught all the formatting stuff.

storage/reader.go
```diff
@@ -222,7 +222,9 @@ var emptyBody = ioutil.NopCloser(strings.NewReader(""))
 // the stored CRC, returning an error from Read if there is a mismatch. This integrity check
 // is skipped if transcoding occurs. See https://cloud.google.com/storage/docs/transcoding.
 type Reader struct {
-	Attrs              ReaderObjectAttrs
+	Attrs          ReaderObjectAttrs
+	objectMetadata map[string]string
```
tritone, line 226: Looks like even a non-exported map here makes the comparable check unhappy.

Let's make this a pointer, and you can keep the method to return the map.

PR #11214, opened 2024-12-03
fix(pubsub): convert stream ack deadline seconds from duration

### LINKED ISSUE #10557: pubsub: updating ack deadline seconds in streaming pull behavior is wrong
**Client**

Pub/Sub 

**Expected behavior**

When exactly once is toggled on/off, the ack deadline is updated on the stream (when keep alives happen) and stream gets new ack deadline based on messages.

**Actual behavior**

The new ack deadline is measured in time.Duration converted to an int directly, instead of converting to seconds first

hongalex:
Fixes #10557

### DIFF
--- a/pubsub/iterator.go
+++ b/pubsub/iterator.go
@@ -893,7 +893,7 @@ func (it *messageIterator) pingStream() {
 	spr := &pb.StreamingPullRequest{}
 	it.eoMu.RLock()
 	if it.sendNewAckDeadline {
-		spr.StreamAckDeadlineSeconds = int32(it.ackDeadline())
+		spr.StreamAckDeadlineSeconds = int32(it.ackDeadline()) / int32(time.Second)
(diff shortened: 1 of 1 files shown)


pubsub/iterator.go
```diff
@@ -893,7 +893,7 @@ func (it *messageIterator) pingStream() {
 	spr := &pb.StreamingPullRequest{}
 	it.eoMu.RLock()
 	if it.sendNewAckDeadline {
-		spr.StreamAckDeadlineSeconds = int32(it.ackDeadline())
+		spr.StreamAckDeadlineSeconds = int32(it.ackDeadline()) / int32(time.Second)
```
shollyman, line 896: Might be more readable to use https://pkg.go.dev/time#Duration.Seconds and cast from there, rather than doing unit division.

shollyman (approved):

PR #11215, opened 2024-12-03
feat(bigtable): Hot backups

bhshkh:
Fixes: b/365774730

More about hot backups: https://cloud.google.com/bigtable/docs/backups#hot-backups

Generated protos have 2 new fields in the Backup struct. These fields are being added to various methods in this PR:  https://github.com/googleapis/google-cloud-go/blob/649c075d5310e2fac64a0b65ec445e7caef42cb0/bigtable/admin/apiv2/adminpb/table.pb.go#L1390-L1400

------------------------------------------

To create a hot backup that expires after 4 days:
```
hotToStdTime := 
adminClient.CreateBackupWithOptions(ctx, "TableID", "Cluster", "BkpName",
		time.Now().Add(24 * 4 * time.Hour), HotToStandardBackup(time.Now().Add(2 * 24 * time.Hour)))
```
or without specifying HotToStandardTime i.e. hot backup does not get converted to standard backup
```
hotToStdTime := time.Now()
adminClient.CreateBackupWithOptions(ctx, "TableID", "Cluster", "BkpName",
		time.Now().Add(24 * 4 * time.Hour), HotBackup())
```

-----------------------------------------

To update hot_to_standard_time of hot backup: 
```
adminClient.UpdateBackupHotToStandardTime(ctx, "Cluster", "BkpName", &hotToStdTime)
```

To remove hot_to_standard_time of hot backup:
```
adminClient.UpdateBackupHotToStandardTime(ctx, "Cluster", "BkpName", nil)
```


bigtable/admin.go
```diff
@@ -2150,9 +2150,60 @@ func (ac *AdminClient) RestoreTableFrom(ctx context.Context, sourceInstance, tab
...
+
+// HotBackup can be used to set backup type to [BackupTypeHot] and optionally, specify the
+// time at which the hot backup will be converted to a standard backup.
+// Once the `hot_to_standard_time` has passed, Cloud Bigtable will convert the
+// hot backup to a standard backup. This value, if provided, must be greater than the backup
+// creation time by:
+// - At least 24 hours
+func HotBackup(hotToStandardTime *time.Time) BackupOption {
+	return hotBackupOption{htsTime: hotToStandardTime}
+}
```
gkevinzheng, line 2193: Why is this option provided as a pointer instead of passing in a value then converting it into a reference in the library?

bhshkh: hotToStandardTime is optional here same as the protos:

https://github.com/googleapis/google-cloud-go/blob/649c075d5310e2fac64a0b65ec445e7caef42cb0/bigtable/admin/apiv2/adminpb/table.pb.go#L1390-L1400

gkevinzheng: I've been trying to also come up with ways to represent an optional `Time` parameter here, even looking at https://stackoverflow.com/questions/30716354/how-do-i-do-a-literal-int64-in-go for help. You could do something like `HotBackup(time.Time)` then another backup option like `HotBackupNil()` or something like that. Not sure if it's the cleanest solution to this problem, but would also like to hear from others.

bhshkh: Changed the options to `HotToStandardBackup` and `HotBackup`
Open to other naming suggestions

gkevinzheng: I'd leave the final decision for the names to the Bigtable team, but setting `hot_to_standard_time` in the proto means setting the time at which it gets converted, and if doing a hot backup without having a provided value means it doesn't get converted, right?

In that case I'd rename the two to something like `HotBackupWithConversion(hotToStandardTime time.Time)` and `HotBackupWithoutConversion()`, as it's more clear to the user what those options are doing.

bigtable/admin.go
```diff
@@ -2371,6 +2474,29 @@ func (ac *AdminClient) UpdateBackup(ctx context.Context, cluster, backup string,
 	return err
 }
 
+// UpdateBackupHotToStandardTime updates the HotToStandardTime of a hot backup.
+func (ac *AdminClient) UpdateBackupHotToStandardTime(ctx context.Context, cluster, backup string, hotToStandardTime *time.Time) error {
```
gkevinzheng, line 2478: Could we use `HotBackupOptions` here to eliminate another `*time.Time` parameter

bigtable/admin.go
```diff
@@ -2150,9 +2150,60 @@ func (ac *AdminClient) RestoreTableFrom(ctx context.Context, sourceInstance, tab
...
+	apply(o *backupOptions)
+}
+
+// setHotToStandardTime can be used to set backup type to [BackupTypeHot] and the
+// time at which the hot backup will
+// be converted to a standard backup.
+// Once the `hot_to_standard_time` has passed, Cloud Bigtable will convert the
+// hot backup to a standard backup. This value must be greater than the backup
+// creation time by at least 24 hours.
+type hotBackupOption struct {
```
DerekLeeCS, line 2169: do we need these `hotBackupOption` lines and functions?

I'm wondering how useful these will be, as opposed to a user creating their own `backupOptions` to create a hot backup

bigtable/admin.go
```diff
@@ -2150,9 +2150,60 @@ func (ac *AdminClient) RestoreTableFrom(ctx context.Context, sourceInstance, tab
...
 func (ac *AdminClient) CreateBackup(ctx context.Context, table, cluster, backup string, expireTime time.Time) error {
+	return ac.createBackup(ctx, table, cluster, backup, expireTime, nil)
+}
+
+// CreateBackupWithOptions is similar to CreateBackup but lets the user specify additional options.
+func (ac *AdminClient) CreateBackupWithOptions(ctx context.Context, table, cluster, backup string, expireTime time.Time, opts ...BackupOption) error {
+	return ac.createBackup(ctx, table, cluster, backup, expireTime, opts...)
+}
+
+func (ac *AdminClient) createBackup(ctx context.Context, table, cluster, backup string, expireTime time.Time, opts ...BackupOption) error {
```
DerekLeeCS, line 2206: if we go with adding `expireTime` to `backupOptions`, then we can have the original `CreateBackup` (that takes `expireTime` as an arg) create its own `backupOptions` and pass it to this function

and then we won't need `expireTime` as an arg to this function

DerekLeeCS (changes requested):
overall logic looks good, just have a few suggestions for the implementation

bigtable/admin.go
```diff
@@ -2150,9 +2150,60 @@ func (ac *AdminClient) RestoreTableFrom(ctx context.Context, sourceInstance, tab
 	return longrunning.InternalNewOperation(ac.lroClient, op).Wait(ctx, &resp)
 }
 
+type backupOptions struct {
```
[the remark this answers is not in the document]
bhshkh: ExpireTime is a required field and not optional. Adding it to backupOptions indicates otherwise
https://github.com/googleapis/google-cloud-go/blob/649c075d5310e2fac64a0b65ec445e7caef42cb0/bigtable/admin/apiv2/adminpb/table.pb.go#L1367-L1374

bigtable/admin.go
bhshkh: hotBackupOption is a private struct and not visible to the user. Having this simplifies the user experience. To create a hot backup, user can just do below instead of having to create their own struct. 
```
hotToStdTime := time.Now()
adminClient.CreateBackupWithOptions(ctx, "TableID", "Cluster", "BkpName",
		time.Now().Add(24 * 4 * time.Hour), HotBackup())
```

This is the common pattern followed across libraries in google-cloud-go to create optional parameters e.g.

https://github.com/googleapis/google-cloud-go/blob/649c075d5310e2fac64a0b65ec445e7caef42cb0/storage/storage.go#L2147-L2221

bigtable/admin.go
bhshkh: Having the same options for create and update can complicate this later. 
Updated the PR to add 2 separate methods similar to other implementation:

https://github.com/googleapis/google-cloud-go/blob/8109157cb2bfb700fde04361e0fa7c1345608fce/bigtable/admin.go#L487-L507

bigtable/admin.go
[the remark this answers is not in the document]
DerekLeeCS: yeah, we can rename `backupOptions` if that's the issue

my suggestion was to have all configurable fields for a backup in a single struct

bigtable/admin.go
DerekLeeCS: alright, I'll defer to the readability reviewer

bigtable/admin.go
```diff
@@ -2150,9 +2150,53 @@ func (ac *AdminClient) RestoreTableFrom(ctx context.Context, sourceInstance, tab
...
+	backupType        *BackupType
+	hotToStandardTime *time.Time
+}
+
+// BackupOption can be used to specify parameters for backup operations.
+type BackupOption interface {
```
shollyman, line 2159: I think my main question here for the option style here is how complex you expect the options to become.  I see only a single option with two behavior fields, so struct-based options that satisfy the apply() contract may be more than you need.

By way of comparison, if the contract is simplified to the core backupOptions and some funcs that can modify it:
```
type BackupOption func(*backupOptions)

func WithHotBackupOption(t time.Time) BackupOption {
  return func(bo *backupOptions) {
     bo.backupType = blah
     bo.hotToStandardTime = blah
  }
}

func WithExpiration(t time.Time) BackupOption {
  return func(bo *backupOptions) {
     bo.expireTime = blah
  }
}
```

bigtable/admin.go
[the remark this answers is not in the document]
shollyman: It's not unreasonable to subsume expiration time into the set of variadic options if you explicitly validate the backupOptions has any necessary fields in CreateWithOptions.

The other way forward might would be if there's a reasonable default expiration (7d, etc).  However in that case I'd probably recommend that the service makes those behavioral choices via an optional field, and documents the defaults. Otherwise you end up with a potential consistency problem across client implementations.

bigtable/admin.go
bhshkh: Thanks for the review. I have updated the code.

bhshkh:
@DerekLeeCS , please review. I have updated the code as per review comments

bhshkh:
@shollyman @gkevinzheng, please review. I have updated the code as per review comments

bigtable/integration_test.go
```diff
@@ -3850,23 +3890,108 @@ func TestIntegration_AdminBackup(t *testing.T) {
...
+			}
+			// Check that updated backup has the correct hot_to_standard_time
+			updatedBackup, err := adminClient.BackupInfo(ctx, testEnv.Config().Cluster, bkpName)
+			if err != nil {
+				t.Fatalf("BackupInfo: %v", err)
+			}
+			gotHtsTime := updatedBackup.HotToStandardTime
+			if (test.wantHtsTime != nil && !test.wantHtsTime.Equal(*gotHtsTime)) ||
+				(test.wantHtsTime == nil && gotHtsTime != nil) ||
+				(test.wantHtsTime != nil && gotHtsTime == nil) {
```
DerekLeeCS, line 3988: do we need this 3rd conditional check?
`(test.wantHtsTime != nil && gotHtsTime == nil)`

I'm wondering if the 1st conditional check is enough (and safe given that `gotHtsTime` could be `nil`)
`(test.wantHtsTime != nil && !test.wantHtsTime.Equal(*gotHtsTime))`

DerekLeeCS (approved):
thanks for the changes!

bigtable/admin.go
```diff
@@ -2150,13 +2152,72 @@ func (ac *AdminClient) RestoreTableFrom(ctx context.Context, sourceInstance, tab
...
+	return func(bo *backupOptions) {
+		btHot := BackupTypeHot
+		bo.backupType = &btHot
+	}
+}
+
 // CreateBackup creates a new backup in the specified cluster from the
 // specified source table with the user-provided expire time.
 func (ac *AdminClient) CreateBackup(ctx context.Context, table, cluster, backup string, expireTime time.Time) error {
+	return ac.createBackup(ctx, table, cluster, backup, WithExpiry(expireTime))
```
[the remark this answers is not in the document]

bigtable/integration_test.go
bhshkh: 1st conditional check is not enough.
Modified the order of checks. 
For e.g.:
```go
package main

import "fmt"

type foo struct {
	bar string
}

func main() {
	var fooPtr *foo
	fmt.Println(*fooPtr)
}
```
Above code throws error:
```
$ go run .
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x48f776]

goroutine 1 [running]:
main.main()
        /usr/local/google/home/bahaaiman/Documents/cfdb-workspace-01/playground/main.go:11 +0x16
exit status 2
```

gkevinzheng (approved):

PR #11224, opened 2024-12-04
fix(bigtable): Resolve discrepancy between server and client qps

bhshkh:
Details in b/358175516


bigtable/metrics.go
```diff
@@ -125,6 +125,7 @@ var (
 	// createExporterOptions takes Bigtable client options and returns exporter options
 	// Overwritten in tests
 	createExporterOptions = func(btOpts ...option.ClientOption) []option.ClientOption {
+		btOpts = append(btOpts, option.WithGRPCConnectionPool(4))
```
mutianf, line 128: do we need 4 channels for this? I think maximum 2 is fine cause there's only 1 RPC per minute.

igorbernstein2: why do we need a channel pool for a client that will do 1 rpc per minute?

bhshkh: Removed

mutianf (approved):

Issue #11199, opened 2024-11-27
storage: SoftDeletePolicy is not getting disabled on bucket update


vsharma-code:
## Client

Storage v1.47.0

## Environment

go version go1.23.3 linux/amd64

## Code and Dependencies

```
package main

import (
	"cloud.google.com/go/storage"
	"context"
	"fmt"
	_ "fmt"
	"github.com/googleapis/gax-go/v2"
	"google.golang.org/api/option"
	"time"
)

type GCPConfig struct {
	ProjectID string
	Region    string
	Options   []option.ClientOption
}

func main() {
	bgCtx := context.Background()
	gcpConf := &GCPConfig{
		ProjectID: "something",
		Region:    "us-east4",
	}
	gcs, err := storage.NewClient(bgCtx, gcpConf.Options...)
	if err != nil {
		fmt.Printf("Failed to create GCS client: %v", err)
		fmt.Println(err)
		return
	}
	gcs.SetRetry(storage.WithPolicy(storage.RetryAlways),
		storage.WithBackoff(gax.Backoff{
			// Set the initial retry delay to a maximum of 2 seconds. The length of
			// pauses between retries is subject to random jitter.
			Initial: 2 * time.Second,
			// Set the maximum retry delay to 60 seconds.
			Max: 60 * time.Second,
			// Set the backoff multiplier to 3.0.
			Multiplier: 3,
		}))
	bucket := gcs.Bucket("A bucket")
	_, err = bucket.Update(context.Background(), storage.BucketAttrsToUpdate{
		SoftDeletePolicy: &storage.SoftDeletePolicy{
			RetentionDuration: 0,
		},

		VersioningEnabled: true,
	})
}
```

<details>
  <summary>go.mod</summary>

```text
module my-go-project

go 1.23.3

require cloud.google.com/go/storage v1.47.0

require (
	cel.dev/expr v0.16.1 // indirect
	cloud.google.com/go v0.116.0 // indirect
	cloud.google.com/go/auth v0.10.2 // indirect
	cloud.google.com/go/auth/oauth2adapt v0.2.5 // indirect
	cloud.google.com/go/compute/metadata v0.5.2 // indirect
	cloud.google.com/go/iam v1.2.1 // indirect
	cloud.google.com/go/monitoring v1.21.1 // indirect
	github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp v1.24.1 // indirect
[...]

PR #11226, opened 2024-12-04
fix(storage): disable soft delete policy using 0 duration

frankyn:
I found the API doesn't allow disabling Soft delete policy sending a 'nil' for raw.BucketSoftDeletePolicy field. Updated implementation to disable soft delete using 0 seconds on retention duration. 

Fixes: #11199
Updates #10761


tritone (approved):
LGTM. Should this close out the integration test failure issue as well?

PR #11231, opened 2024-12-05
test(bigquery): migrate trace testing from opencensus to otel

shollyman:


alvarowolfx:
@shollyman  I think we should migrate this to otel instead of just deleting the trace integration test. We recently solved a issue with PR https://github.com/googleapis/google-cloud-go/issues/11119 that added more test scenario to that file.

shollyman:
> @shollyman I think we should migrate this to otel instead of just deleting the trace integration test. We recently solved a issue with PR #11119 that added more test scenario to that file.

Fair.  My intent was to kill opencensus in this PR and introduce opentelemetry in another PR, but I can do the migration in one shot.

alvarowolfx (approved):

PR #11244, opened 2024-12-07
ci(spanner): emulator test should run only when env is set already

rahul2393:
Alternate PR:  https://github.com/googleapis/google-cloud-go/pull/11243/


spanner/metrics_test.go
```diff
@@ -148,9 +148,8 @@ func TestNewBuiltinMetricsTracerFactory(t *testing.T) {
 	}
 	for _, test := range tests {
 		t.Run(test.desc, func(t *testing.T) {
-			if test.setEmulator {
```
sakthivelmanii, line 151: can you use `runOnlyInEmulator` and set the value to `isEmulatorEnvSet()` so that you can avoid having two checks.

harshachinta (approved):

spanner/metrics_test.go
```diff
@@ -141,16 +141,16 @@ func TestNewBuiltinMetricsTracerFactory(t *testing.T) {
 			},
 		},
 		{
-			desc:        "should not create instruments when SPANNER_EMULATOR_HOST is set",
-			config:      ClientConfig{},
-			setEmulator: true,
+			desc:               "should not create instruments when SPANNER_EMULATOR_HOST is set",
+			config:             ClientConfig{},
+			runOnlyInEmulator:  true,
```
sakthivelmanii, line 146: `runOnlyInEmulator: isEmulatorEnvSet()` can't we use it like this?

PR #11255, opened 2024-12-10
fix(auth): check compute cred type before non-default flag for DP

codyoss:
We need to first ensure a token is a compute type before we enforce the option to allow for a non-default service account. This was leading to a strange XDS error if say a user credential was supplied and this option was enabled.


codyoss:
@adonovan I think this is a new behavior. I wonder if it is related to https://github.com/golang/exp/commit/14434422244cefca3b64b0c2a68634e5e8c406ee

quartzmo (approved):

PR #11278, opened 2024-12-12
feat(secretmanager): upgrade protobuf generation with go_grpc_library

quartzmo:
This PR was produced for demo purposes only using the following commands on my Cloudtop. The real PR for this change will be delivered through OwlBot.

```
gbazelisk clean --expunge
gbazelisk build //google/cloud/secretmanager/v1:gapi-cloud-secretmanager-v1-go
```

The Protobuf gRPC generation produces an empty file (`apiv1/secretmanagerpb/resources_grpc.pb.go`) which breaks CI in this PR and which must be removed by OwlBot post-processing.


jskeet:
@quartzmo I suspect this can be closed now?

PR #11288, opened 2024-12-13
feat(auth): pass logger from auth layer to metadata package

codyoss:


quartzmo (commented):
LGTM, but there are other `metadata` requests in auth that could potentially be logged. Do these need TODOs?

```
auth/internal/internal.go:
  204  	defer cancel()
  205: 	return metadata.GetWithContext(ctx, "universe/universe-domain")
  206  }

auth/internal/transport/s2a.go:
  87  var httpGetMetadataMTLSConfig = func() (string, error) {
  88: 	return metadata.GetWithContext(context.Background(), configEndpointSuffix)
  89  }

auth/credentials/detect.go:
  101  			ProjectIDProvider: auth.CredentialsPropertyFunc(func(ctx context.Context) (string, error) {
  102: 				return metadata.ProjectIDWithContext(ctx)
  103  			}),

auth/credentials/idtoken/compute.go:
  46  		ProjectIDProvider: auth.CredentialsPropertyFunc(func(ctx context.Context) (string, error) {
  47: 			return metadata.ProjectIDWithContext(ctx)
  48  		}),
```

codyoss:
@quartzmo thanks for the callout. This ended up needing quite a few more change to plumb to those lower layers, but I think it is good now :)

quartzmo (approved):

PR #11302, opened 2024-12-17
feat(storage): add ObjectHandle.Move method

tritone:
Support the new MoveObject API in both gRPC and JSON.


storage/http_client.go
```diff
@@ -593,6 +593,31 @@ func (c *httpStorageClient) RestoreObject(ctx context.Context, params *restoreOb
 	return newObject(obj), err
 }
 
+func (c *httpStorageClient) MoveObject(ctx context.Context, params *moveObjectParams, opts ...storageOption) (*ObjectAttrs, error) {
```
Tulsishah, line 596: What happen if we call this same API for flat bucket, I mean what error it will return?

storage/client.go
```diff
@@ -313,6 +314,13 @@ type restoreObjectParams struct {
 	copySourceACL  bool
 }
 
+type moveObjectParams struct {
+	bucket, srcObject, dstObject string
+	srcConds                     *Conditions
+	dstConds                     *Conditions
+	encryptionKey                []byte
```
Tulsishah, line 321: Do we require predefinedACL, sendCRC32C in moveObject?

storage/http_client.go
```diff
@@ -593,6 +593,31 @@ func (c *httpStorageClient) RestoreObject(ctx context.Context, params *restoreOb
...
+	}
+	if err := setEncryptionHeaders(req.Header(), params.encryptionKey, false); err != nil {
+		return nil, err
+	}
+	var obj *raw.Object
+	var err error
+	err = run(ctx, func(ctx context.Context) error { obj, err = req.Context(ctx).Do(); return err }, s.retry, s.idempotent)
+	var e *googleapi.Error
+	if ok := errors.As(err, &e); ok && e.Code == http.StatusNotFound {
+		return nil, ErrObjectNotExist
```
Tulsishah, line 616: Is there any reason why we are converting NoFound Http error to storage errror?

storage/client.go
tritone: No -- MoveObject does not change the ACL or need checksumming (b/c object data does not change, only the name)

storage/http_client.go
tritone: That's what other methods in this library do; but really we should fix this to wrap the HTTP error as well.

storage/grpc_client.go
```diff
@@ -662,6 +662,36 @@ func (c *grpcStorageClient) RestoreObject(ctx context.Context, params *restoreOb
...
+	s := callSettings(c.settings, opts...)
+	req := &storagepb.MoveObjectRequest{
+		Bucket:            bucketResourceName(globalProjectAlias, params.bucket),
+		SourceObject:      params.srcObject,
+		DestinationObject: params.dstObject,
+	}
+	if err := applyCondsProto("MoveObjectDestination", defaultGen, params.dstConds, req); err != nil {
+		return nil, err
+	}
+	if err := applySourceCondsProto("MoveObjectSource", defaultGen, params.srcConds, req); err != nil {
```
BrennaEpp, line 675: Why is this set to `defaultGen`?

storage/http_client.go
BrennaEpp: +1 We should wrap the error as well

storage/integration_test.go
```diff
@@ -4603,6 +4603,80 @@ func TestIntegration_SoftDelete(t *testing.T) {
...
+		h := testHelper{t}
+		srcObj := "move-src-obj"
+		dstObj := "move-dst-obj"
+
+		// Create bucket with HNS enabled
+		bkt := client.Bucket(prefix + uidSpace.New())
+		attrs := &BucketAttrs{
+			HierarchicalNamespace:    &HierarchicalNamespace{Enabled: true},
+			UniformBucketLevelAccess: UniformBucketLevelAccess{Enabled: true},
+			SoftDeletePolicy:         &SoftDeletePolicy{RetentionDuration: 0},
```
BrennaEpp, line 4617: Tangential thought here, should we modify the default bucket/bucket creation to set this policy as well?

storage/integration_test.go
```diff
@@ -4603,6 +4603,80 @@ func TestIntegration_SoftDelete(t *testing.T) {
...
+		// Bad source generation should cause 412.
+		_, err = obj2.If(Conditions{
+			GenerationMatch: 123,
+		}).Move(ctx, MoveObjectDestination{Object: dstObj2})
+		if err == nil {
+			t.Error("ObjectHandle.Move: got nil, want 412 error")
+		}
+		if !(status.Code(err) == codes.FailedPrecondition || extractErrCode(err) == http.StatusPreconditionFailed) {
+			t.Errorf("ObjectHandle.Move: got err %v, want failed precondition (412)", err)
+		}
```
BrennaEpp, line 4658: nit, I think it's cleaner to have these two combined
```suggestion
		if err == nil || !(status.Code(err) == codes.FailedPrecondition || extractErrCode(err) == http.StatusPreconditionFailed) {
			t.Errorf("ObjectHandle.Move: got err %v, want failed precondition (412)", err)
		}
```

storage/integration_test.go
```diff
@@ -4603,6 +4603,80 @@ func TestIntegration_SoftDelete(t *testing.T) {
...
+		// Correctly applied preconditions should work.
+		_, err = obj2.If(Conditions{
+			GenerationMatch:     w2.Attrs().Generation,
+			MetagenerationMatch: w2.Attrs().Metageneration,
+		}).Move(ctx, MoveObjectDestination{Object: dstObj2, Conditions: &Conditions{DoesNotExist: true}})
+		if err != nil {
+			t.Fatalf("ObjectHandle.Move: %v", err)
+		}
+	})
+}
```
rainmark11, line 4678: optional: do we want to check if the src obj is deleted?

storage/integration_test.go
tritone: Not default for the library, but yeah we can change the test defaults.

storage/integration_test.go
tritone: Yup, I'll fix that.

storage/grpc_client.go
tritone: Generation is not a field on the request. In general it doesn't look like you can specify a generation to move while leaving behind others at the source.

storage/integration_test.go
tritone: Good idea, will add.

storage/http_client.go
tritone: It's not the best error -- I am filing a bug with an internal team.

GCSFuse should avoid using MoveObject unless you know it's an HNS bucket.

storage/http_client.go
tritone: Will do this as a follow up.

BrennaEpp (approved):

PR #11303, opened 2024-12-17
fix(storage/dataflux): address deadlock when reading from ranges

jdnurme:
This change addresses an intermittent deadlock issue when reading from the range channel. The fix introduces a 200ms timeout on all reads off of range, opting to re-check for available work if the timeout is triggered. Since this timeout is only applied to reads off of the channel it should not introduce any new blocking behavior.

### DIFF
--- a/storage/dataflux/range_splitter.go
+++ b/storage/dataflux/range_splitter.go
@@ -328,6 +328,8 @@ func (rs *rangeSplitter) convertStringRangeToMinimalIntRange(
 
 // charPosition returns the index of the character in the alphabet set.
 func (rs *rangeSplitter) charPosition(ch rune) (int, error) {
+	rs.mu.Lock()         // Acquire the lock
+	defer rs.mu.Unlock() // Release the lock when the function exits
 	if idx, ok := rs.alphabetMap[ch]; ok {
 		return idx, nil
 	}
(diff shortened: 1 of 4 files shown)


storage/dataflux/worksteal.go
```diff
@@ -130,9 +130,13 @@ func (w *worker) doWorkstealListing(ctx context.Context) error {
...
 				continue
 			} else {
-				newRange := <-w.lister.ranges
-				<-w.idleChannel
-				w.updateWorker(newRange.startRange, newRange.endRange, active)
+				select {
+				case newRange := <-w.lister.ranges:
+					<-w.idleChannel
+					w.updateWorker(newRange.startRange, newRange.endRange, active)
+				case <-time.After(sleepDurationWhenIdle):
```
tritone, line 137: This seems like kind of a stopgap?

Really I think this and the if case above at L129 should be a single for/select loop potentially. Could you try writing the code this way and let me know if it works?

I see why this was causing deadlocks now, because lister.ranges could be non-empty but then become empty before L133 is reached if other goroutines are waiting.

In general we should recv from a channel to check if it is ready, and use select to block on that or give an out if it is not. Checking and then receiving separately is non-atomic so seems reasonable that these kinds of issues will arise.

jdnurme: Yes the code above can be fully deleted, just forgot to remove it in this push from my fork.

storage/dataflux/range_splitter.go
```diff
@@ -328,6 +328,8 @@ func (rs *rangeSplitter) convertStringRangeToMinimalIntRange(
 
 // charPosition returns the index of the character in the alphabet set.
 func (rs *rangeSplitter) charPosition(ch rune) (int, error) {
+	rs.mu.Lock()         // Acquire the lock
```
BrennaEpp, line 331: Nice, thanks for fixing this as well.

BrennaEpp (approved):

tritone (approved):

Issue #11305, opened 2024-12-18
storage: Misleading Documentation


blagoySimandov:
https://github.com/googleapis/google-cloud-go/blob/1a04bafa83c27342b9308d785645e1e5423ea10d/storage/storage.go#L425

In here the ContentType is marked as "Optional", when in fact, if left blank the only content type accepted for the signed url is the empty string (""). To me, optional sounds a bit more like "if you omit it anything goes.", rather than "only empty strings are accepted".

tritone:
I think there might be a bug (in our code or yours) based on how I am reading the docs. Left a comment on the PR with details.

tritone:
Closing this due to lack of response.

PR #11306, opened 2024-12-18
doc(storage): updated the ContentType documentation for the signedUrl in the storage package

blagoySimandov:
Updated the documentation for the ContentType field in the SignedURLOptions struct in the storage package, to correctly 
state that the field is required.

Motivation: Issue #11305


storage/storage.go
```diff
@@ -475,7 +475,7 @@ type SignedURLOptions struct {
 
 	// ContentType is the content type header the client must provide
 	// to use the generated signed URL.
-	// Optional.
+	// Required.
```
tritone, line 478: I think this is not really accurate. It's valid to generate a URL without a Content-Type header (e.g. for a GET request), so the field is definitely not required in SignedURLOptions.

Moreover from the docs (https://cloud.google.com/storage/docs/authentication/canonical-requests#about-headers) it sounds like content-type is not a required canonical header, so you should be able to add it to the request after signing.

Maybe there is a server side issue here? Let me know if you think I'm mistaken, or you can file an issue with support to share details about the error.

blagoySimandov: I just believe it's misleading... I imagined that omitting the ContentType would make it so it accepts any kind of content type, this in fact was not the case and the only accepted content type was that of an empty string... Not to mention the fact that in a client-side/browser setting if you omit the content-type from the header the browser will automatically put in a "guess" for the content-type of the file that you are uploading, which would then be rejected since the only accepted content type would be ""... Results in a really nasty bug..

blagoySimandov: Here is something i wrote quite a while ago that might help you understand my frustration. [blog](https://simandoff.blog/posts/gcp-signed-urls-browser-oddity/)

tritone: Okay so I investigated this further and it looks like you are using v2 signing, while the docs I linked to are for v4.

We recommend v4 signing generally and it does something more like what you want (at least according to the docs). Can you try that out by setting https://pkg.go.dev/cloud.google.com/go/storage#SignedURLOptions.Scheme ?

I think we could improve the docs here for v2, but I would definitely not want to mark this field as required in any case since it is not required for most signed URLs.

PR #11342, opened 2024-12-23
chore(spanner): support mutation only operation for read-write mux

rahul2393:
The following describes the case when a read-write transaction is executed on a multiplexed session and this PR handles this scenario:

If a read-write transaction contains only mutations, a random mutation is selected from the mutation list and sent with the BeginTransactionRequest. The resulting Transaction response includes a precommit token, which is tracked by the client library and used in the subsequent CommitRequest


spanner/transaction_test.go
```diff
@@ -413,19 +413,31 @@ func TestReadWriteTransaction_PrecommitToken(t *testing.T) {
...
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			_, err := client.ReadWriteTransaction(ctx, func(ctx context.Context, tx *ReadWriteTransaction) error {
+				if tc.mutationsOnly {
+					ms := []*Mutation{
+						Insert("t_foo", []string{"col1", "col2"}, []interface{}{int64(1), int64(2)}),
+						Update("t_foo", []string{"col1", "col2"}, []interface{}{"one", []byte(nil)}),
+					}
+					if err := tx.BufferWrite(ms); err != nil {
```
harshachinta, line 436: can we verify that the BeginTransactionRequest has a mutation key set and not null?

spanner/transaction_test.go
```diff
@@ -413,19 +413,31 @@ func TestReadWriteTransaction_PrecommitToken(t *testing.T) {
...
 	}
 
 	testCases := []testCase{
-		{"Only Query", true, false, false, "PartialResultSetPrecommitToken", 3}, //since mock server is returning 3 rows
-		{"Query and Update", true, true, false, "ResultSetPrecommitToken", 4},
-		{"Query, Update, and Batch Update", true, true, true, "ExecuteBatchDmlResponsePrecommitToken", 5},
+		{"Only Query", true, false, false, false, "PartialResultSetPrecommitToken", 3},
+		{"Query and Update", true, true, false, false, "ResultSetPrecommitToken", 4},
+		{"Query, Update, and Batch Update", true, true, true, false, "ExecuteBatchDmlResponsePrecommitToken", 5},
+		{"Only Mutations", false, false, false, true, "BeginTransactionPrecommitToken", 1},
```
harshachinta, line 425: Can you add a negative test scenario also?
i.e when a  Commit RPC returns aborted then the retry should set mutation key and the commit request should have latest precommit token
https://github.com/harshachinta/java-spanner/blob/01c03144134febfb3c83e1bc80a7533f53f067ba/google-cloud-spanner/src/test/java/com/google/cloud/spanner/MultiplexedSessionDatabaseClientMockServerTest.java#L1140

spanner/transaction_test.go
```diff
@@ -413,19 +413,31 @@ func TestReadWriteTransaction_PrecommitToken(t *testing.T) {
 		query                  bool
 		update                 bool
 		batchUpdate            bool
+		mutationsOnly          bool
```
harshachinta, line 416: Can you add one test case with client.Apply and verify the same behavior?
https://cloud.google.com/spanner/docs/modify-mutation-api#go

spanner/mutation.go
```diff
@@ -427,24 +429,50 @@ func (m Mutation) proto() (*sppb.Mutation, error) {
 
 // mutationsProto turns a spanner.Mutation array into a sppb.Mutation array,
 // it is convenient for sending batch mutations to Cloud Spanner.
-func mutationsProto(ms []*Mutation) ([]*sppb.Mutation, error) {
+func mutationsProto(ms []*Mutation) ([]*sppb.Mutation, *sppb.Mutation, error) {
+	var selectedMutation *Mutation
+	var nonInsertMutations []*Mutation
+	maxValues := -1
```
harshachinta, line 435: Why do we need to store seperate maxValues variable?
This will always be equal to selectedMutation.values() if selectedMutation is not null ?

harshachinta (commented):
Changes LGTM. Requested for couple of tests to verify negative scenarios

harshachinta (approved):

PR #11359, opened 2025-01-01
feat(firestore): add UsesEmulator field to Client

raymonstah:
Adds a boolean flag to determine if the client is connected to the emulator. Checking the `UsesEmulator` boolean is a safeguard where there may be multiple instantiated clients.


bhshkh (commented):
nit: Rename `UsingEmulator` to `UsesEmulator`

bhshkh (approved):

raymonstah:
@bhshkh are there any teams we want to tag to get the approvals we need to merge?

PR #11378, opened 2025-01-07
chore: update remaining usages of Go 1.21 to 1.22

fasmat:
I found a few places in the codebase where Go 1.21 is still used after support for it has officially been dropped. This updates those to Go 1.22.


codyoss (approved):

codyoss:
Thanks again!

PR #11382, opened 2025-01-07
Fix CVE-2024-45338.

twixthehero:
* Upgrade github.com/google/s2a-go 0.1.8 => 0.1.9.


codyoss:
Hey, thanks for the PR. Would you mind opening an issue for discussion. It is not clear to me this is an issue for this library. Looking at your other PR in s2a the problem here was in x/net which we have already updated in this repository.

twixthehero:
> Hey, thanks for the PR. Would you mind opening an issue for discussion. It is not clear to me this is an issue for this library. Looking at your other PR in s2a the problem here was in x/net which we have already updated in this repository.

Created #11426. Likely I'll need to include updates to other dependencies for a complete fix. Lmk what you think

quartzmo:
Closing per commentary on https://github.com/googleapis/google-cloud-go/issues/11426

PR #11387, opened 2025-01-07
test(storage): fix emulated tests for metadata on read

cojenco:


storage/client_test.go
```diff
@@ -849,8 +849,13 @@ func TestOpenReaderEmulated_Metadata(t *testing.T) {
...
 			"Some-Other-Key": "some-other-value",
 		}
-		if diff := cmp.Diff(r.Metadata(), expectedMetadata); diff != "" {
-			t.Fatalf("Object Metadata: got(-),want(+):\n%s", diff)
+		gotMetaData := r.Metadata()
+		// Testbench specific metadata is included for testing purposes.
+		for key, expectedValue := range expectedMetadata {
+			actualValue, ok := gotMetaData[key]
+			if !ok || actualValue != expectedValue {
+				t.Fatalf("Object Metadata: Expected key %q with value %q, but got %q", key, expectedValue, actualValue)
```
BrennaEpp, line 857: I assume this fails if one of the expected keys is in the metadata (`!ok` would return true?), but can you/did you double check?

danielduhh (approved):

storage/client_test.go
BrennaEpp: Nice thanks

PR #11398, opened 2025-01-08
test(storage): parallelize concurrency test for MRD

BrennaEpp:
Original change from @shubham-diwakar

This test ensure that Adds can be called in parallel


danielduhh (commented):
Do we need to make any changes to make sure these tests run with  the -race option?

BrennaEpp:
> Do we need to make any changes to make sure these tests run with the -race option?

I verified manually that it passes with that flag. CI runs integration tests with that flag post-submit.

danielduhh (approved):
Do we need to make any changes to make sure these tests run with  the -race option?

Issue #11368, opened 2025-01-03
pubsub: release v2 library


hongalex:
The Pub/Sub Go library is in progress to receive a v2 update. Details will be provided in the migration guide of the library but the key differences are

1. The existing resource management API (admin operations) will be deprecated, and replaced with the gRPC autogenerated surface that the other packages in this repository use. This affects topic, subscription, and schema management and related API.
2. The `Topic` and `Subscription` structs will be renamed to `Publisher` and `Subscriber` respectively
3. Some settings fields will be removed (ReceiveSettings.Synchronous, PublishSettings.BufferedByteLimit, ReceiveSettings.UseLegacyFlowControl)
4. Some settings fields will be renamed
5. Some default settings will be changed (ReceiveSettings.NumGoroutines will now default to 1)

We plan to support the `cloud.google.com/go/pubsub` (henceforth known as the v1 library) for 1 year from when the v2 library is launched, to give developers proper time to migrate to the v2. We will prioritize new features in v2, but providing security and bug fixes in the v1.

The following steps will be done, and will roughly correspond to 1 PR each, with changes being staged in the `pubsub-v2-trunk` branch.
- [x] Create new v2 directory and copy existing v1 surface over
- [x] Rewrite the imports of v2 directory to point to `cloud.google.com/go/pubsub/v2`, including GAPIC code #11292
- [ ] Remove existing handwritten admin API and associated tests from v2 library #11367
- [ ] Settings refactoring (removal, renaming, default settings changes) 
- [ ] Expose autogenerated library (to be used for admin operations) in the Pub/Sub client
- [ ] Rename Topic/Subscription to Publisher/Subscriber
- [ ] Refactor tests that use removed handwritten API to use the new admin operations
- [ ] Mark v1 package as deprecated, and all operations that will be removed
- [ ] Configure GAPIC generator to start generating new v2 library

PR #11429, opened 2025-01-09
refactor(pubsub/v2): make tests use new GAPIC library

hongalex:
One step in https://github.com/googleapis/google-cloud-go/issues/11368

What this PR does:
1. TopicAdminClient and SubscriptionAdminClient now exposed via the uber Pub/Sub `Client`
2. Makes tests pass again by making the remaining v2 library tests use the above clients
  - Tests passes locally, but CI still failing because of weirdness with new package
3. Start part of the process to rename topic -> publisher, subscription -> subscriber
4. Remove some old test helper methods that are not needed because they're admin specific


pubsub/v2/iterator_test.go
```diff
@@ -34,14 +34,15 @@ import (
...
 
 var (
 	projName                = "P"
-	topicName               = "some-topic"
-	subName                 = "some-sub"
-	fullyQualifiedTopicName = fmt.Sprintf("projects/%s/topics/%s", projName, topicName)
-	fullyQualifiedSubName   = fmt.Sprintf("projects/%s/subscriptions/%s", projName, subName)
+	topicID                 = "some-topic"
+	subID                   = "some-sub"
+	fullyQualifiedTopicName = fmt.Sprintf("projects/%s/topics/%s", projName, topicID)
```
shollyman, line 44: to consider: there's a lot of these string builders in the test code.  Did you want to switch over to using the generated path functions from the generator?  e.g. https://github.com/googleapis/google-cloud-go/blob/main/pubsub/apiv1/path_funcs.go

shollyman (approved):
One overall comment about all the resource URL constructions, but otherwise LGTM

PR #11432, opened 2025-01-10
fix(storage): fix stream termination in MRD.

shubham-diwakar:
1. Make CloseSend() call before releasing resource.
2. Drain inbound response from the stream.


storage/grpc_client.go
```diff
@@ -1373,8 +1375,6 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
 }
 
 func getActiveRange(r *gRPCBidiReader) []rangeSpec {
-	r.mu.Lock()
-	defer r.mu.Unlock()
 	var activeRange []rangeSpec
```
BrennaEpp, line 1378: What's the guarantee here? That if `getActiveRange` is running, no other goroutine will be running as well? I don't think we can guarantee that since `Wait()` might be running, right?

storage/grpc_client.go
```diff
@@ -1468,6 +1468,10 @@ func (mr *gRPCBidiReader) wait() {
 
 // Close will notify stream manager goroutine that the reader has been closed, if it's still running.
 func (mr *gRPCBidiReader) close() error {
+	// Before release of resource we close the client->server connection.
+	if err := mr.stream.CloseSend(); err != nil {
+		return err
+	}
```
BrennaEpp, line 1474: Should we also drain the stream (as we do in `drainInboundStream()`, receiving from stream until we get a non-nil error) to make sure its resources are released? See https://github.com/grpc/grpc-go/commit/365770fcbd7dfb9d921cb44827ede770f33be44f

storage/grpc_client.go
BrennaEpp: You can trigger this situation running `go test -v -short -run="TestRetryConformance/grpc-8-\[return-broken-stream\]-storage.objects.download-3" -race`

shubham-diwakar: Yes makes sense, thanks.

storage/grpc_client.go
shubham-diwakar: Actually i am not sure that will be get any outputs here if we call stream.Recv post CloseSend(). What if all the responses where consumed by streamReceiver go routine?

Although there are some cases when we close the stream even with requests added i have drained responses there.

LMK your thoughts.

cjc25: I think Recv() continually returns the err (io.EOF or something else) once the stream is done? This should be easy to check with a toy program.

However, multiple concurrent calls to Recv() are not allowed, so if streamReceiver goroutine may be calling Recv(), you have to be careful not to call Recv() elsewhere until streamReceiver is done. I think it's probably easiest for all Recv() calls to live on that one goroutine.

It's maybe easiest to call CloseSend() on the same goroutine which calls Send()? Then you have one goroutine for Send/CloseSend, and another for Recv, and user code can cancel the context then call Close() to trigger a cancellation.

shubham-diwakar: Thanks Chris for the suggestion.
That kind of simplifies structure too.

Now one go routine has all Recv() calls and other has all Send()/CloseSend() calls.

storage/grpc_client.go
```diff
@@ -1470,18 +1487,32 @@ func (mr *gRPCBidiReader) wait() {
 
 // Close will notify stream manager goroutine that the reader has been closed, if it's still running.
 func (mr *gRPCBidiReader) close() error {
-	if mr.cancel != nil {
-		mr.cancel()
-	}
+	mr.closeManager <- true
+	mr.closeReceiver <- true
```
arjan-bal, line 1491: There seem to be buffered channels. So it's not guaranteed that the receivers would have read the message when execution resumes. Even if thee were unbuffered channels, the sender would get unblocked as soon as the receiver gets the message. The sender and receiver would race to lock the mutex and the sender could win. This could result in the context getting cancelled before the stream is drained.

shubham-diwakar:
Made sure that we receive EOF error or any permanent error from server on inbound stream(recv) and on outbound stream made sure to call closesend before calling the cancel().

Seeing ok response now on server side dashboards, Ref: https://screenshot.googleplex.com/8qAqijdXnohNeDi.

BrennaEpp (approved):

tritone:
This required some modifications on merge with the commit that changed callback semantics to bytes written.

storage/grpc_client.go
```diff
@@ -1483,18 +1514,39 @@ func (mr *gRPCBidiReader) wait() {
...
 	mr.mu.Lock()
+	for key := range mr.mp {
+		mr.mp[key].callback(mr.mp[key].offset, rr.mp[key].totalBytesWritten, fmt.Errorf("stream closed early"))
```
tritone, line 1521: ``
```suggestion
		mr.mp[key].callback(mr.mp[key].offset, mr.mp[key].totalBytesWritten, fmt.Errorf("stream closed early"))
```

Issue #11012, opened 2024-10-18
datastore: snapshot reads for GetAll queries


mrdomino:
## Is your feature request related to a problem? Please describe.

When you make a client with `WithReadOptions(datastore.ReadTime())`, this read time is not respected for aggregate queries. It looks like this just never got added in #6755, which only added snapshot reads for the individual accessors.

## Describe the solution you'd like

It'd be great to be able to do a GetAll at a specific time that shows all of a certain entity type at that time.

mrdomino:
Awesome, thank you!

PR #11447, opened 2025-01-14
feat(datastore): Add read time from client while running queries

bhshkh:
Fixes: https://github.com/googleapis/google-cloud-go/issues/11012

User can create a client with read time option. This option wasn't getting passed in RunQueryRequest and RunAggregationQueryRequest. This PR adds this option.

https://github.com/googleapis/google-cloud-go/blob/93e313d90a2345c1ead1c1fe9149643d310bfa17/datastore/apiv1/datastorepb/datastore.pb.go#L464-L465

https://github.com/googleapis/google-cloud-go/blob/93e313d90a2345c1ead1c1fe9149643d310bfa17/datastore/apiv1/datastorepb/datastore.pb.go#L701-L702

https://github.com/googleapis/google-cloud-go/blob/93e313d90a2345c1ead1c1fe9149643d310bfa17/datastore/apiv1/datastorepb/datastore.pb.go#L2184-L2205

As can be seen in above proto, the consistency type field in ReadOptions can be one of ReadOptions_ReadConsistency_, ReadOptions_Transaction, ReadOptions_NewTransaction or ReadOptions_ReadTime.  Same validation has been added on client


datastore/query.go
```diff
@@ -976,22 +977,32 @@ func (c *Client) RunAggregationQueryWithOptions(ctx context.Context, aq *Aggrega
 	return ar, nil
 }
 
-func validateReadOptions(eventual bool, t *Transaction) error {
-	if t == nil {
+func validateReadOptions(eventual bool, t *Transaction, clientReadSettings *readSettings) error {
+	crtExists := clientReadSettings != nil && !clientReadSettings.readTime.IsZero()
```
gkevinzheng, line 981: I see this line `clientReadSettings != nil && !clientReadSettings.readTime.IsZero()` being used twice, and it could be reused in the future. Could you turn `clientReadSettingsExists` into a function?

bhshkh: Moved to a function. Thanks!

gkevinzheng (approved):

PR #11457, opened 2025-01-16
fix(bigtable): Allow nil condition in conditional mutation

bhshkh:
When the test proxy receives a conditional mutation request with nil condition, it uses the PassAllFilter to call Apply method of Go library. Thus, the request sent by Go library to Cloud Bigtable service contains the field pass all filter.
[TestCheckAndMutateRow_NoRetry_TrueMutations](https://github.com/googleapis/cloud-bigtable-clients-test/blob/v0.0.3/tests/checkandmutaterow_test.go#L109-L110) conformance test fails with error 

```
=== RUN   TestCheckAndMutateRow_NoRetry_TrueMutations
    checkandmutaterow_test.go:138: diff found (-want +got):
          (*bigtablepb.CheckAndMutateRowRequest)(Inverse(protocmp.Transform, protocmp.Message{
          	"@type":            s"google.bigtable.v2.CheckAndMutateRowRequest",
        + 	"predicate_filter": s"{pass_all_filter:true}",
          	"row_key":          []uint8("row-01"),
          	"table_name":       string("projects/project/instances/instance/tables/table"),
          	"true_mutations":   []protocmp.Message{{"@type": s"google.bigtable.v2.Mutation", ...}, {"@type": s"google.bigtable.v2.Mutation", ...}},
          }))
--- FAIL: TestCheckAndMutateRow_NoRetry_TrueMutations (0.00s)
```

The Go library assumes that the field `cond` in `Mutation` being non-nil denotes conditional mutation but as per the above test we want to allow conditional mutation with no condition. So, a new variable `isConditional` is being added to `Mutation` to distinguish between conditional and non-conditional mutations.

After fix:
```
=== RUN   TestCheckAndMutateRow_NoRetry_TrueMutations
--- PASS: TestCheckAndMutateRow_NoRetry_TrueMutations (0.00s)
```


gkevinzheng (commented):
Where is `m.isConditional` ever set to `false`?

gkevinzheng (approved):

andre-sampaio (approved):

PR #11462, opened 2025-01-16
docs(auth): add warning about externally-provided credentials

quartzmo:


auth/credentials/idtoken/idtoken.go
```diff
@@ -78,9 +78,25 @@ type Options struct {
...
 	// CredentialsJSON sources a JSON credential file from the provided bytes.
 	// If provided, do not provide CredentialsJSON. Optional.
+	//
+	// Important: If you accept a credential configuration (credential
+	// JSON/File/Stream) from an external source for authentication to Google
+	// Cloud Platform, you must validate it before providing it to any Google
+	// API or library. Providing an unvalidated credential configuration to
+	// Google APIs can compromise the security of your systems and data. For
+	// more information, refer to [Validate credential configurations from
+	// external sources](https://cloud.google.com/docs/authentication/external/externally-sourced-credentials).
```
sai-sunder-s, line 99: Does impersonated cred also need a similar update then?

quartzmo: I'm not sure. In this PR I did not update configuration fields that accept the general type `auth.Credentials`. PTAL at the `impersonate` package surface and let me know if you see any configuration fields (or functions) that need the warning, thank you. https://pkg.go.dev/cloud.google.com/go/auth@v0.14.0/credentials/impersonate

quartzmo: I don't believe the `impersonate` package is used with external accounts. I think impersonation use cases with external accounts are supported by the `externalaccount` package, which is updated in this PR in `auth/credentials/externalaccount/externalaccount.go`.

sai-sunder-s (approved):

auth/credentials/externalaccount/externalaccount.go
```diff
@@ -109,18 +109,50 @@ type CredentialSource struct {
 	// File is the location for file sourced credentials.
 	// One field amongst File, URL, Executable, or EnvironmentID should be
 	// provided, depending on the kind of credential in question.
+	//
+	// Important: If you accept a credential configuration (credential
```
codyoss, line 113: At least for this package we have a security considerations section in the godoc -- perhaps this should be added there. This actually sounds pretty similar to that disclaimer. Wonder if we should just have that sort of message in each of the credential packages instead of putting this on the field docs.

codyoss (approved):
Approving, but consider the comment.

auth/credentials/externalaccount/externalaccount.go
quartzmo: @sai-sunder-s PTAL at Cody's comment above.

sai-sunder-s: We mainly need it for the method that loads any arbitrary json. This is what we are doing in other languages. 

@codyoss what do you mean by "each credential package"? I thought go doesn't expose per credential type methods?

sai-sunder-s: The existing warning is in external account specific section. Do you mean put the warning in the documentation of each cred?

If so, that is not of much use for the issue we are dealing with now. The issue is specifically when a developer accepts cred config from external sources and builds a cred out of it. They expect only certain type like Service Account but they end up accepting other cred types.

So the warning needs to be in a central place like the method used to load arbitrary creds. Maybe it can go at the top of the detect method as well?

codyoss: I think what I was aiming for was having a central section to put such information. I don't love having to maintain the same docs in N places. Would prefer to have it in a single place. I was wondering if having a security section in the package docs would suffice.

sai-sunder-s: I think it needs to be there for the detect method. Additionally can be put in central place as well. We can remove other places added in this PR. 
Does that sound ok?

auth/credentials/detect.go
```diff
@@ -149,10 +149,26 @@ type DetectOptions struct {
 	// CredentialsFile overrides detection logic and sources a credential file
 	// from the provided filepath. If provided, CredentialsJSON must not be.
 	// Optional.
+	//
```
sai-sunder-s, line 152: lets keep updates in this file

auth/credentials/externalaccount/externalaccount.go
sai-sunder-s: we can remove the warnings added to this file

quartzmo: > We can remove other places added in this PR.

Please directly note on my PR which warnings to keep and which to remove.

auth/credentials/idtoken/idtoken.go
```diff
@@ -78,9 +78,25 @@ type Options struct {
...
 	// CredentialsFile sources a JSON credential file from the provided
 	// filepath. If provided, do not provide CredentialsJSON. Optional.
+	//
+	// Important: If you accept a credential configuration (credential
+	// JSON/File/Stream) from an external source for authentication to Google
+	// Cloud Platform, you must validate it before providing it to any Google
+	// API or library. Providing an unvalidated credential configuration to
+	// Google APIs can compromise the security of your systems and data. For
+	// more information, refer to [Validate credential configurations from
+	// external sources](https://cloud.google.com/docs/authentication/external/externally-sourced-credentials).
```
sai-sunder-s, line 88: lets keep the updates in this file.

Initially did not realize that this is also like detect.go but for id token.

sai-sunder-s (approved):

Issue #11460, opened 2025-01-16
bigquery/storage/managedwriter: connection improvements


shollyman:
Land improvments to the managewriter to improve connection management logic.  This is impacting a particular customer workload.

Relevant internal issue: b/388522702

In consideration:

* reduce reconnects on schema changes
* improve draining for queued responses on a connection that's being recycled

shollyman:
Deferring on the reconnection work for the time being.  While we could potentially reduce the number of disconnect/reconnect events, one of the things the explicit reconnect is giving us is more consistency around the consistency of the schema in the backend.

shollyman:
Closing this for the time being, as we're still not prioritizing refactoring how connection handling in managedwriter functions.

PR #11463, opened 2025-01-17
feat(bigquery/storage/managedwriter): graceful connection drains

shollyman:
This PR enables a more graceful shutdown when we're doing client initiated reconnects, typically in response to things like schema changes.

With this PR, rather than immediately cancelling the previous connection context, we instead use a goroutine to defer this by a fixed interval, which gives our recv goroutine more time to process any incoming notifications from the service.

This can yield reductions in duplicate rows for some scenarios, particularly when EnableWriteRetries is enabled and we see sequential disconnects (e.g. a multiplex connection observing a number of schema updates).

It does, however, yield possible short term increases in resource usage, as we're now potentially retaining goroutines/channels/connections for a longer interval during these reconnect events.  This PR updates testing that looks for leaking resources with this new understanding accordingly.

Towards: https://github.com/googleapis/google-cloud-go/issues/11460

### DIFF
--- a/bigquery/storage/managedwriter/connection.go
+++ b/bigquery/storage/managedwriter/connection.go
@@ -37,6 +37,8 @@ const (
 
 var (
 	errNoRouterForPool = errors.New("no router for connection pool")
+	// TODO: consider if this should be user configurable
+	gracefulReconnectDuration = 20 * time.Second
 )
 
 // connectionPool represents a pooled set of connections.
@@ -491,10 +493,9 @@ func (co *connection) getStream(arc *storagepb.BigQueryWrite_AppendRowsClient, f
 	}
 	if co.cancel != nil {
 		// Delay cancellation to give queued writes a chance to drain normally.
-		// TODO: Revisit this to see if this should be user configurable.
 		oldCancel := co.cancel
 		go func() {
-			time.Sleep(20 * time.Second)
+			time.Sleep(gracefulReconnectDuration)
 			oldCancel()
 		}()
 		co.ctx, co.cancel = context.WithCancel(co.pool.ctx)
--- a/bigquery/storage/managedwriter/connection.go
+++ b/bigquery/storage/managedwriter/connection.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"sync"
+	"time"
 
 	"cloud.google.com/go/bigquery/storage/apiv1/storagepb"
 	"github.com/googleapis/gax-go/v2"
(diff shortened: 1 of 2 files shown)


bigquery/storage/managedwriter/connection.go
```diff
@@ -36,6 +37,8 @@ const (
 
 var (
 	errNoRouterForPool = errors.New("no router for connection pool")
+	// TODO: consider if this should be user configurable
```
leahecole, line 40: Make sure there's a bug or an issue open for this todo

As for if it should be user configurable, maybe, but it depends on how it would impact other timeouts configured by gax or at the service level

bigquery/storage/managedwriter/managed_stream_test.go
```diff
@@ -539,12 +541,24 @@ func TestManagedStream_LeakingGoroutinesReconnect(t *testing.T) {
 		if testArc.openCount != i+2 {
 			t.Errorf("should trigger a reconnect, but found openCount %d", testArc.openCount)
 		}
+		// Bump our threshold each append.  We add 2 goroutines each reconnect due to a new processor
+		// and a graceful cancellation, but this should recover over time.
```
leahecole, line 545: "should" doing a lot of heavy lifting here 😁  Are there conditions in which this would not recover over time?

bigquery/storage/managedwriter/managed_stream_test.go
```diff
@@ -539,12 +541,24 @@ func TestManagedStream_LeakingGoroutinesReconnect(t *testing.T) {
...
+		threshold = threshold + 2
+
 		if i%10 == 0 {
 			if current := runtime.NumGoroutine(); current > threshold {
 				t.Errorf("potential goroutine leak, append %d: current %d, threshold %d", i, current, threshold)
```
leahecole, line 550: Is it possible for this error to occur if we're constantly increasing the threshold?

bigquery/storage/managedwriter/managed_stream_test.go
```diff
@@ -539,12 +541,24 @@ func TestManagedStream_LeakingGoroutinesReconnect(t *testing.T) {
...
 			}
 		}
 	}
+	// Verify goroutine count drops after graceful shutdowns should have concluded.
+	time.Sleep(gracefulReconnectDuration)
+	threshold = threshold - (2 * appendCount)
+
+	if current := runtime.NumGoroutine(); current > threshold {
+		t.Errorf("potential goroutine leak after recovery: current %d, threshold %d", current, threshold)
```
leahecole, line 559: ah okay I feel less angsty after seeing this check

leahecole (commented):
Leaving lots of questions and one half opinion

bigquery/storage/managedwriter/connection.go
shollyman: Done.

bigquery/storage/managedwriter/managed_stream_test.go
shollyman: We're in a situation where we're intentionally increasing the resources by a known amount (2 goroutines per reconnect); we're looking for increases beyond that expectation.

leahecole (approved):

jameshartig:
@shollyman this is much appreciated! We've run into this before and are looking forward to the improvements.

PR #11472, opened 2025-01-20
chore(spanner): handle commit retry protocol extension for mux rw

rahul2393:
In a read-write transaction using a multiplexed session with (read/query + mutation) operations, the CommitResponse from the backend during the commit RPC may include a MultiplexedSessionRetry field (indicated by a precommit token). This field signals that the commit RPC should be retried once using the new precommit token. During this retry, mutations should not be resent, as they were already buffered in spanFE during the initial commit RPC call.


spanner/transaction.go
```diff
@@ -1703,17 +1703,31 @@ func (t *ReadWriteTransaction) commit(ctx context.Context, options CommitOptions
...
+		}
+		return client.Commit(contextWithOutgoingMetadata(ctx, t.sh.getMetadata(), t.disableRouteToLeader), req, gax.WithGRPCOptions(grpc.Header(&md)))
+	}
+	// Initial commit attempt with mutations
+	res, err := performCommit(precommitToken, true)
+	if err != nil {
+		return resp, t.txReadOnly.updateTxState(toSpannerErrorWithCommitInfo(err, true))
+	}
+	// Retry if MultiplexedSessionRetry is present, without mutations
+	if res.GetMultiplexedSessionRetry() != nil {
```
harshachinta, line 1728: nit: Here can we get the precommit token, pass it to setPrecommitToken method and then below use t.precommittoken?
I understand that the precommittoken reveived in res should be latest but here we will need to rely on sequence number to determine that the precommit token is indeed latest. This will save us in future in cases where the precommit token received in res could be outdated.

spanner/transaction_test.go
```diff
@@ -504,6 +504,101 @@ func TestReadWriteTransaction_PrecommitToken(t *testing.T) {
...
+			if err == iterator.Done {
+				break
+			}
+			if err != nil {
+				return err
+			}
+		}
+
+		if _, err := tx.Update(ctx, Statement{SQL: UpdateBarSetFoo}); err != nil {
+			return err
```
harshachinta, line 560: Commit retry only happens in cases of DQL+Mutation. Lets remove DML to avoid confusion

harshachinta (approved):

PR #11475, opened 2025-01-21
fix(storage): propagate ctx from invoke to grpc upload reqs

BrennaEpp:
The context from `run` (see `invoke.go`) is not being used for gRPC uploads.

Rather than keep a source of truth in a writer var (except for the original context), pass through contexts. 

The other option is to keep as is and instead do the following. This may be "safer" (at least in terms of, more likely to keep current behaviour) as it is hard to follow exactly how these contexts are behaving currently.                  
```
                                 uploadBuff := func(ctx context.Context) error {
					saveCtx := gw.ctx
					gw.ctx = ctx
					obj, err := gw.uploadBuffer(ctx, recvd, offset, doneReading)
					o = obj
					gw.ctx = saveCtx
					return err
				}
```

- Note: `bucketContext` also changed to be called once for ALL grpc upload types.

### DIFF
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -1725,12 +1725,12 @@ func (c *grpcStorageClient) OpenWriter(params *openWriterParams, opts ...storage
 
 				var o *storagepb.Object
 				uploadBuff := func(ctx context.Context) error {
-					obj, err := gw.uploadBuffer(recvd, offset, doneReading)
+					obj, err := gw.uploadBuffer(ctx, recvd, offset, doneReading)
 					o = obj
 					return err
 				}
 
-				err = run(gw.ctx, uploadBuff, gw.settings.retry, s.idempotent)
+				err = run(bucketContext(gw.ctx, gw.bucket), uploadBuff, gw.settings.retry, s.idempotent)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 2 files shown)


tritone (approved):
@cjc25 I think this looks like a good change but just FYI.

cjc25:
> @cjc25 I think this looks like a good change but just FYI.

LGTM. The stashed context was mirroring the implementation as of Nov 2024, which also stashed a context (and also didn't use run): https://github.com/googleapis/google-cloud-go/blob/19f01c3c48ed1272c8fc0af9e5f69646cb662808/storage/grpc_client.go#L2030. I think @BrennaEpp is right that the prior behavior didn't handle retry/backoff correctly. I did not merge #11200 correctly, which is how we lost it.

tritone:
> > @cjc25 I think this looks like a good change but just FYI.
> 
> LGTM. The stashed context was mirroring the implementation as of Nov 2024, which also stashed a context (and also didn't use run):
> 
> https://github.com/googleapis/google-cloud-go/blob/19f01c3c48ed1272c8fc0af9e5f69646cb662808/storage/grpc_client.go#L2030
> 
> . I think @BrennaEpp is right that the prior behavior didn't handle retry/backoff correctly. I did not merge #11200 correctly, which is how we lost it.

Re: #11200 , that was expected, @BrennaEpp re-implemented the retry/backoff behavior in the private repo given the large changes in uploads overall.

BrennaEpp:
> > > @cjc25 I think this looks like a good change but just FYI.
> > 
> > 
> > LGTM. The stashed context was mirroring the implementation as of Nov 2024, which also stashed a context (and also didn't use run):
> > https://github.com/googleapis/google-cloud-go/blob/19f01c3c48ed1272c8fc0af9e5f69646cb662808/storage/grpc_client.go#L2030
> > 
> > . I think @BrennaEpp is right that the prior behavior didn't handle retry/backoff correctly. I did not merge #11200 correctly, which is how we lost it.
> 
> Re: #11200 , that was expected, @BrennaEpp re-implemented the retry/backoff behavior in the private repo given the large changes in uploads overall.

+1, when I introduced run to get the backoff I didn't think about the ctx from run being lost; really, it has never used that ctx and this change adds it in.

PR #11476, opened 2025-01-21
feat(storage): implement RetryChunkDeadline for grpc writes

BrennaEpp:
Branched from #11475 

Adds a timer to the upload buffer. This is equivalent to the http implementation. 

Changed from implementing directly in the upload code to implementing in `invoke.go`. This makes it extensible (if we ever want to add a similar deadline to other calls), but also more prone to affecting other ops (just to be clear, I don't think it will given testing/how it is written). Reason for changing it is because we need to be able to stop the retries while still wrapping the error. If we do it a level above, if we return a "deadline done, here is last error" wrapping the error, it will continue to retry if the last error is retriable.


storage/storage.go
```diff
@@ -2420,6 +2420,7 @@ type retryConfig struct {
 	policy      RetryPolicy
 	shouldRetry func(err error) bool
 	maxAttempts *int
+	maxDuration time.Duration
```
tritone, line 2423: Also let's call it `maxRetryDuration` since it doesn't actually limit the total duration of calls.

storage/integration_test.go
```diff
@@ -6962,11 +6961,17 @@ func extractErrCode(err error) int {
...
+//   - apierror.APIError with grpcStatusCode, or
+//   - grpc/status.Status error with grpcStatusCode.
 func errorIsStatusCode(err error, httpStatusCode int, grpcStatusCode codes.Code) bool {
 	var httpErr *googleapi.Error
 	var grpcErr *apierror.APIError
+
 	return (errors.As(err, &httpErr) && httpErr.Code == httpStatusCode) ||
-		(errors.As(err, &grpcErr) && grpcErr.GRPCStatus().Code() == grpcStatusCode)
+		(errors.As(err, &grpcErr) && grpcErr.GRPCStatus().Code() == grpcStatusCode) ||
+		status.Code(err) == grpcStatusCode
```
tritone, line 6974: Why does this need 3 cases?

storage/invoke.go
```diff
@@ -67,22 +68,40 @@ func run(ctx context.Context, call func(ctx context.Context) error, retry *retry
...
+
+	var lastErr error
 	return internal.Retry(ctx, bo, func() (stop bool, err error) {
+		if retry.maxDuration != 0 {
+			select {
+			case <-quitAfterTimer.C:
+				if lastErr == nil {
+					return true, errors.New("storage: request not sent, choose a larger value for the retry deadline")
+				}
+				return true, fmt.Errorf("storage: retry timed out after %v attempts; last error: %w", attempts, lastErr)
```
tritone, line 85: This err should include the value of the timeout as well.

tritone (commented):
Overall looks good, a few minor things

storage/integration_test.go
BrennaEpp: Testbench errors are Status errors but not `apierror.APIError`s, so the added test fails when checking the status code without this change

storage/storage.go
BrennaEpp: Done.
It stutters a bit (`retryConfig.maxRetryDuration`) but that was my initial instinct as well. It's clearer anyway since `retryConfig` can have any name.

storage/invoke.go
BrennaEpp: Done

storage/integration_test.go
tritone: Hmm. That seems like an issue with wrapping at some layer. But in any case L6974 should work for all grpc errors (both real service and testbench)

tritone (approved):

PR #11487, opened 2025-01-22
chore(internal): remove go.opencensus.io pattern from leakcheck and thirdpartycheck

quartzmo:


internal/actions/cmd/thirdpartycheck/main.go
```diff
@@ -41,7 +41,6 @@ var (
 		"golang.org/",
 
 		// Third party deps (allowed).
-		"go.opencensus.io",
```
codyoss, line 44: I see this as a required dep in some of the veneers, I don't think this can be removed yet until it is completely out of the tree.

codyoss (changes requested):

internal/actions/cmd/thirdpartycheck/main.go
quartzmo: Yes, I see that too now, thanks.

quartzmo: Restored, thanks.

codyoss (approved):

PR #11488, opened 2025-01-22
chore: bump dependencies

codyoss:


quartzmo:
Kokoro:

```
malformed go.sum:
/tmpfs/src/google-cloud-go/go.work.sum:85: wrong number of fields 2
```

codyoss:
@quartzmo should be fixed now, bad merge

quartzmo (approved):

PR #11499, opened 2025-01-24
fix(bigtable): Allow GC updates on emulated aggregate column family

### LINKED ISSUE #11498: bigtable: Emulator does not allow GC policy changes on existing aggregate column family
## Client

Bigtable emulator (`bigtable/bttest`)

## Environment

```
go version go1.23.3 darwin/arm64
```

## Code and Dependencies

Add the following test to `bigtable/bttest/inmem_test.go`:

```go
func TestUpdateGCPolicyOnAggregateColumnFamily(t *testing.T) {
	ctx := context.Background()
	s := &server{
		tables: make(map[string]*table),
	}
	tblInfo, err := s.CreateTable(ctx, &btapb.CreateTableRequest{
		Parent:  "cluster",
		TableId: "t",
		Table: &btapb.Table{
			ColumnFamilies: map[string]*btapb.ColumnFamily{
				"sum": {
					ValueType: &btapb.Type{
						Kind: &btapb.Type_AggregateType{
							AggregateType: &btapb.Type_Aggregate{
								InputType: &btapb.Type{
									Kind: &btapb.Type_Int64Type{},
								},
								Aggregator: &btapb.Type_Aggregate_Sum_{
									Sum: &btapb.Type_Aggregate_Sum{},
								},
							},
						},
					},
				},
			},
		},
	})
	if err != nil {
		t.Fatal(err)
	}
	_, err = s.ModifyColumnFamilies(ctx, &btapb.ModifyColumnFamiliesRequest{
		Name: tblInfo.Name,
		Modifications: []*btapb.ModifyColumnFamiliesRequest_Modification{
			{
				Id: "sum",
				Mod: &btapb.ModifyColumnFamiliesRequest_Modification_Update{
					Update: &btapb.ColumnFamily{
						GcRule: &btapb.GcRule{Rule: &btapb.GcRule_MaxNumVersions{MaxNumVersions: 1}},
					},
				},
			},
		},
	})
	if err != nil {
		t.Fatal(err)
	}
}
```

This test fails with the following error:

```
=== RUN   TestUpdateGCPolicyOnAggregateColumnFamily
    inmem_test.go:2759: rpc error: code = InvalidArgument desc = Immutable fields 'value_type.aggregate_type' cannot be updated
--- FAIL: TestUpdateGCPolicyOnAggregateColumnFamily (0.00s)
```

## Expected behavior

In the above example, we are merely trying to override the GC policy of the aggregate column family in the Bigtable emulator. We aren't in any way attempting to mutate the _type_ of the column family. Therefore, I would expect this test to succeed.

## Actual behavior

The Bigtable emulator does not seem to allow for a partial modification of an aggregate column family, and it mistakenly believes that the GC policy update also contains a (`nil`) modification to the aggregate column family type. This causes the above test to fail.

wesnel:
Closes #11498:

- https://github.com/googleapis/google-cloud-go/issues/11498


bigtable/bttest/inmem.go
```diff
@@ -338,15 +338,22 @@ func (s *server) ModifyColumnFamilies(ctx context.Context, req *btapb.ModifyColu
 			if !ok {
 				return nil, fmt.Errorf("no such family %q", mod.Id)
 			}
-			if cf.valueType != nil {
-				_, isOldAggregateType := cf.valueType.Kind.(*btapb.Type_AggregateType)
-				if isOldAggregateType && cf.valueType != newcf.valueType {
-					return nil, status.Errorf(codes.InvalidArgument, "Immutable fields 'value_type.aggregate_type' cannot be updated")
-				}
+
+			if newcf.valueType == nil {
```
bhshkh, line 342: This condition needs to be changed.

Right now, column family has only two fields that can be updated. So, it makes sense to assume that when a field is nil, it is not getting updated and the other one is. This is not future compatible in the scenario where more fields get added to column family.

https://github.com/googleapis/google-cloud-go/blob/a2a926df08fa129ce3ea0513f23acf23ef12f6ac/bigtable/admin/apiv2/adminpb/table.pb.go#L953-L976


Instead, use the UpdateMask field. As documented [here](https://github.com/googleapis/google-cloud-go/blob/a2a926df08fa129ce3ea0513f23acf23ef12f6ac/bigtable/admin/apiv2/adminpb/bigtable_table_admin.pb.go#L3271-L3274), read the updatemask from the req. Then, check if it is nil. If it is nil, assume it is updating gcrule.

wesnel: Thank you, this was extremely helpful! I've just pushed some changes that will allow for partial updates of the column family based on the `UpdateMask`, as well as some additional tests to validate the expected behavior of the `UpdateMask` in this case. Please let me know if there's anything else I need to change.

bhshkh (approved):

PR #11501, opened 2025-01-24
feat(storage): Benchmark with experimental MRD.

cjc25:
Allow benchmarks to use the experimental MultiRangeDownloader for normal range reads.

Note that this does not use a single MultiRangeDownloader across multiple reads, which is supported but would require a more invasive change to the benchmark.


storage/internal/benchmarks/client_pool.go
```diff
@@ -247,6 +250,9 @@ func initializeGRPCClient(ctx context.Context, config clientConfig) (*storage.Cl
 	if config.readBufferSize != useDefault {
 		opts = append(opts, option.WithGRPCDialOption(grpc.WithReadBufferSize(config.readBufferSize)))
 	}
+	if config.useMultiRangeDownloader {
```
tritone, line 253: I think this would only use bidi reads via storage.Reader, not actually use MultiRangeDownloader.

I think this is fine but maybe we should rename the opt; or just have Bidi reads be a transport option instead?

cjc25: useMultiRangeDownloaderForSingleRangeReads?

It gets a little unwieldy :) But I agree that users might be a little misled by the current name. I'm open to suggestions.

BrennaEpp: +1 to Chris, I would go with having the option for bidi reads

cjc25: Done, renamed to useGRPCBidiReads to match the experimental option

BrennaEpp (approved):

tritone (approved):

PR #11519, opened 2025-01-29
feat(storage): wrap NotFound errors for buckets and objects

### LINKED ISSUE #2587: storage: disambiguate errors when accessing object attributes
I wanted to differentiate errors when using `client.Bucket(bktName).Object(objName).Attrs(context.Background())`. 
Under this method, the following cases all have the same `ErrObjectNotExist` error:
1. non-existent object, OR
2. bucket does not exist / user does not have access to the bucket

However, the underlying status code and error message in the original `googleapi.Error` is actually important to differentiate the two cases above so that we can handle these two errors properly.

A way to differentiate the above cases. After some exploration, I was able to differentiate the two cases above by inspecting the raw `googleapi.Error` error message. 

The 1st case (non-existent object) has a `404` statusCode and `No such object: .+` error message. 
The 2nd case (no access to the bucket) has a `404` statusCode and `Not Found` error message.

This problem can be solved by one of these solutions: 
1. Reduce error abstraction by returning the raw errors instead of `ErrObjectNotExist` or `ErrBucketNotExist`
In this solution, I expect that I can extract the actual status code and error message, and perform the necessary error handling mechanism according to the different type of errors.

2. Do the error message checks inside Attrs and return the appropriate error based on the actual error message.
In this solution, I expect that Attrs is able to disambiguate between different types of errors and returns the error as either `ErrObjectNotExist` or `ErrBucketNotExist` according to the actual cause of the error.

BrennaEpp:
Code that checks directly against ErrBucketNotExist and ErrObjectNotExist will break; use `errors.Is()` instead.

This change allows users to inspect the underlying error.
Fixes #2587


tritone (approved):
Looks good, nice changes!

Maybe add to the docs for the sentinel errors that they should be checked using errors.Is?

Issue #11466, opened 2025-01-17
spanner: spansql.ParseDLL error by parsing text search features in cloudspanner


geerttouquet:
## Client
go
## Code and Dependencies

try to parse example inhttps://cloud.google.com/spanner/docs/full-text-search/tokenization, gives an error in 
 spansql.ParseDLL

example
CREATE TABLE Albums (
  AlbumId STRING(MAX) NOT NULL,
  AlbumTitle STRING(MAX),
  Rating FLOAT64,
  AlbumTitle_Tokens TOKENLIST AS (TOKENIZE_FULLTEXT(AlbumTitle)) HIDDEN,
  Rating_Tokens TOKENLIST AS (TOKENIZE_NUMBER(Rating)) HIDDEN
) PRIMARY KEY(AlbumId);

spansql.ParseDDL("filename", test.sql)

gives error : filename:5: got "(", want ")" or ","

PR #11522, opened 2025-01-29
feat(spanner/spansql): Add support for tokenlist and create search index

rasviitanen:
Adds support for parsing `tokenlist` and `create search index` in spanner/spansql.

Fixes https://github.com/googleapis/google-cloud-go/issues/11466


spanner/spansql/keywords.go
```diff
@@ -234,6 +236,11 @@ var funcNames = []string{
 	"TRIM",
 	"UPPER",
 
+	// Token functions.
```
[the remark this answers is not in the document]
rasviitanen: We are missing `TOKENIZE_BOOL`, `TOKENIZE_NGRAMS` too in that case. I can look at adding it and make sure we can parse arguments for each.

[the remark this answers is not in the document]
rasviitanen: Pushed some changes now to allow this, I believe it could be more  enforced since the options for each function are limited and known up front. I've tried to mimic what's already implemented here and opted to allow both the key and value for the "arguments" to be quite generic when parsing (even though spanner only accepts `language_tag`, `content_type`, `token_category` for the `TOKENIZE_FULLTEXT` function for example).

It didn't look like the lib supports constraints like that for other functions, see for example [FLOAT64](https://cloud.google.com/spanner/docs/reference/standard-sql/json_functions#double_for_json), which has a `wide_number_mode`, which this lib doesn't seem to support. So could be good to make sure such functions are parsed correctly too.

spanner/spansql/sql.go
```diff
@@ -98,6 +98,63 @@ func (ci CreateIndex) SQL() string {
 	return str
 }
 
+func (csi CreateSearchIndex) SQL() string {
+	str := "CREATE SEARCH INDEX"
```
neophenix, line 102: I believe you need a space at the end of this like CreateIndex has.  I was getting DDL errors because my
```
CREATE SEARCH INDEX foo ON ...
```
Was becoming
```
CREATE SEARCH INDEXfoo ON ...
```

spanner/spansql/parser.go
```diff
@@ -1361,6 +1382,132 @@ func (p *parser) parseCreateIndex() (*CreateIndex, *parseError) {
...
+
+	if p.eat("STORING") {
+		ci.Storing, err = p.parseColumnNameList()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if p.eat("PARTITION", "BY") {
+		ci.PartitionBy, err = p.parseColumnNameList()
```
neophenix, line 1461: This should be p.parseIDList() since the format is `PARTITION BY col1 ...` and not `PARTITION BY (col1,...)`

spanner/spansql/sql.go
```diff
@@ -98,6 +98,63 @@ func (ci CreateIndex) SQL() string {
...
+		}
+		str += c.SQL()
+	}
+	str += ")"
+	if len(csi.Storing) > 0 {
+		str += " STORING (" + idList(csi.Storing, ", ") + ")"
+	}
+
+	if len(csi.PartitionBy) > 0 {
+		str += " PARTITION BY (" + idList(csi.PartitionBy, ", ") + ")"
```
neophenix, line 117: PARTITION BY shouldn't have () around the identifiers

rasviitanen:
Thanks for the review, I have now fixed your comments.
I expanded the tests a bit and also fixed some issues with `ORDER BY` and `OPTIONS`.

neophenix (approved):

neophenix:
Awesome thanks!  I did not do an exhaustive review, these were just things I ran into when I pulled your branch and was running some local tests, but the changes all seem to work for me, so fingers crossed :)

karunacybozu:
@rasviitanen it's not working if we combine with other function. 
For example I need `NORMALIZE_AND_CASEFOLD()` function to normalize Katakana to Hiragana for searching: https://cloud.google.com/spanner/docs/reference/standard-sql/string_functions#normalize_and_casefold

Example would be:

```
CREATE TABLE `Posts` (
    `TenantId` STRING(8) NOT NULL,
    `UserId` STRING(26) NOT NULL,
    `Id` STRING(26) NOT NULL,
    `Title` STRING(256) NOT NULL,
    `Body` STRING(1024) NOT NULL,
    `Optional` STRING(256),
    `Title_Tokens` TOKENLIST AS (TOKENIZE_FULLTEXT(Title)) HIDDEN,
    `Body_Tokens` TOKENLIST AS (TOKENIZE_FULLTEXT(Body)) HIDDEN,
    `Combined_Tokens` TOKENLIST AS (TOKENLIST_CONCAT([Title_Tokens, Body_Tokens])) HIDDEN,
    `Title_SubStrTokens` TOKENLIST AS (TOKENIZE_SUBSTRING(Title)) HIDDEN,
    `Body_SubStrTokens` TOKENLIST AS (TOKENIZE_SUBSTRING(Body)) HIDDEN,
    `Combined_SubStrTokens` TOKENLIST AS (TOKENLIST_CONCAT([Title_SubStrTokens, Body_SubStrTokens])) HIDDEN,
    `Title_NFKC_SubStrTokens` TOKENLIST AS (TOKENIZE_SUBSTRING(NORMALIZE_AND_CASEFOLD(Title, NFKC))) HIDDEN,
    `Body_NFKC_SubStrTokens` TOKENLIST AS (TOKENIZE_SUBSTRING(NORMALIZE_AND_CASEFOLD(Body, NFKC))) HIDDEN,
    `Combined_NFKC_SubStrTokens` TOKENLIST AS (TOKENLIST_CONCAT([Title_NFKC_SubStrTokens, Body_NFKC_SubStrTokens])) HIDDEN,
) PRIMARY KEY (`TenantId`, `UserId`, `Id`)
```

It's working just fine in spanner studio.

But in Go code:

`panic: 01-Posts.sql:14: got "(", want ")" or ","`

rasviitanen:
@karunacybozu Thanks for your comments!
The issue here is not the parsing of the `TOKENIZE_*` function arugments in this PR. The parsing I added here actually supports nested functions and is correct. The problem is rather that `NORMALIZE_AND_CASEFOLD` isn't recognized as a known string function. I would prefer to not add support for it in this PR, as it isn't required by `TOKENLIST`/`TOKENIZE_`/`SEARCH INDEX`.

I would suggest to submit a PR of your own to add additional string functions such as `NORMALIZE` and `NORMALIZE_AND_CASEFOLD` to `spanner/spansql/keywords.go`. That should be all that's necessary.

One could argue that the error message from the parser is a bit cryptic though, maybe we could improve it in a follow-up.
It looks like the lib has the same problem for all the other argument parsers, and I don't want to add any more to this already quite big PR. It's probably better to solve error handling for everything at once and in that case we are better off keeping the scope of this PR limited.
Hope that works, thanks!🙏

rasviitanen:
@rahul2393 It would be greatly appreciated if this could be reviewed, not sure who to ping 🙏

rahul2393 (approved):

rasviitanen:
@rahul2393 looks like my branch was outdated, had to merge main again. Could you readd `kokoro:force-run`? Thanks!

PR #11527, opened 2025-01-29
fix(storage): wait for XML read req to finish to avoid data races

BrennaEpp:
Fixes TestRetryReadStallEmulated failing with `-race`


tritone (approved):
Nice catch on this one; thanks @BrennaEpp

PR #11537, opened 2025-01-30
feat(pubsub): support new forms of topic ingestion

hongalex:
This PR supports topic ingestion for AWS MSK, Azure Event Hub, and Confluent Cloud Managed Kafka.


pubsub/topic.go
```diff
@@ -592,6 +592,189 @@ func (i *IngestionDataSourceCloudStoragePubSubAvroFormat) isCloudStorageIngestio
 	return true
 }
 
+// EventHubsState denotes the possible states for ingestion from Event Hubs.
+type EventHubsState int
```
michaelpri10, line 596: Can we use AzureEventHubs to match the proto definition more closely?

pubsub/topic.go
```diff
@@ -592,6 +592,189 @@ func (i *IngestionDataSourceCloudStoragePubSubAvroFormat) isCloudStorageIngestio
...
+	MSKPublishPermissionDenied
+
+	// MSKClusterNotFound indicates the provided Msk cluster wasn't found.
+	MSKClusterNotFound
+
+	// MSKTopicNotFound indicates the provided topic wasn't found.
+	MSKTopicNotFound
+)
+
+// IngestionDataSourceAmazonMSK are ingestion settings for Amazon MSK.
```
michaelpri10, line 685: AWSMSK would more closely match our proto, but does look a bit strange. I would be fine keeping AmazonMSK or maybe AwsMsk.

pubsub/topic.go
```diff
@@ -592,6 +592,189 @@ func (i *IngestionDataSourceCloudStoragePubSubAvroFormat) isCloudStorageIngestio
...
+}
+
+var _ IngestionDataSource = (*IngestionDataSourceAzureEventHubs)(nil)
+
+func (i *IngestionDataSourceAzureEventHubs) isIngestionDataSource() bool {
+	return true
+}
+
+// MSKState denotes the possible states for ingestion from Amazon MSK.
+type MSKState int
```
michaelpri10, line 663: AmazonMSKState or AwsMsk to match the struct below.

michaelpri10 (changes requested):

pubsub/topic.go
hongalex: I shortened this (and other State consts) just because it was getting a bit lengthy and customers are only ever going to be checking for this output anyway, rather than setting it. Also considering we aren't supporting this surface (the v2 will be using a different admin surface), I felt being brief was ok.

I don't feel that strongly about it, so if you think it's best to match proto definition, I'm ok with that.

pubsub/topic.go
hongalex: Go style guide says initialisms must keep the same case, so unfortunately `AwsMsk` isn't correct. I think AmazonMSK is clearer, and matches the official product name ("Amazon Managed Streaming for Apache Kafka") anyway

pubsub/topic.go
michaelpri10: I'm okay with keeping just EventHubsState then.

pubsub/topic.go
michaelpri10: AmazonMSK is good with me.

pubsub/topic.go
hongalex: I'll switch to AmazonMSKState

michaelpri10 (approved):

PR #11546, opened 2025-01-31
feat(spanner): make randIDForProcess hexadecimal in x-goog-spanner-request-id

### LINKED ISSUE #11545: spanner: x-goog-spanner-request-id's randIDForProcess should be printed in hexadecimal (%016x) and not decimal %d
## Spanner

randIDForProcess should be in hexadecimal as has been requested offline.

odeke-em:
Implements a new requirement that randIDForProcess should be a hexadecimal value with the format specifier "%016x".

Fixes #11545

### DIFF
--- a/spanner/request_id_header.go
+++ b/spanner/request_id_header.go
@@ -46,7 +46,7 @@ func init() {
 	if err != nil {
 		panic(err)
 	}
-	randIDForProcess = r64.String()
+	randIDForProcess = fmt.Sprintf("%016x", r64.Uint64())
 }
 
 // Please bump this version whenever this implementation
--- a/spanner/request_id_header_test.go
+++ b/spanner/request_id_header_test.go
@@ -16,6 +16,7 @@ package spanner
 
 import (
 	"context"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"math"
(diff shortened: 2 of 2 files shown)


rahul2393 (approved):

spanner/request_id_header.go
```diff
@@ -38,15 +38,15 @@ import (
 var randIDForProcess string
 
 func init() {
-	bigMaxInt64, _ := new(big.Int).SetString(fmt.Sprintf("%d", uint64(math.MaxUint64)), 10)
-	if g, w := bigMaxInt64.Uint64(), uint64(math.MaxUint64); g != w {
+	bigMaxUint64, _ := new(big.Int).SetString(fmt.Sprintf("%d", uint64(math.MaxUint64)), 10)
+	if g, w := bigMaxUint64.Uint64(), uint64(math.MaxUint64); g != w {
```
rahul2393, line 42: @odeke-em What are the cases when this fail?

odeke-em: I really just changed the name to match actually uint64 but if that is distracting I can change it back

rahul2393: Sorry I understand we are changing just names, but I want to understand in what case panic will happen.
https://go.dev/wiki/CodeReviewComments#dont-panic

odeke-em: It is really a guard and sanity check in case someone makes a change in the future. Also that advice doesn't apply to this case as this is an init function that runs at the very beginning of the program vs when a program is actually running, so if we don't parse a uint64, we have an incorrect logical setup of what is needed to function correctly.

olavloite (approved):

PR #11563, opened 2025-02-05
fix(internal/godocfx): don't save timestamps until modules are successfully processed

tbpg:
I refactored the index client a bit to use a URL that can be faked out for testing.

I also confirmed this properly queries the index by running it locally.


codyoss (approved):
Thanks!

Issue #8219, opened 2023-07-06
iam: mistmatch in exported functions in IAMClient and proto client


antgamdia:
**Client**

Importing the project and using the grpc client for making requests.

**Environment**

Local development, N/A

**Go Environment**

$ go version: go version go1.20.5 linux/amd64
$ go env N/A

**Code**

We found out some IAM admin operations are not available in the exposed IAMClient client but they are in the proto client. Perhaps it is the intended behavior, I don't know, but reporting it just in case.

**Expected behavior**

- Example of an OK correspondence between the exposed IAMClient and the internal proto client:

https://github.com/googleapis/google-cloud-go/blob/54218e963bb5a6d47411a490985b54053825064f/iam/admin/apiv1/iam_client.go#L185

https://github.com/googleapis/google-cloud-go/blob/54218e963bb5a6d47411a490985b54053825064f/iam/admin/apiv1/adminpb/iam.pb.go#L5661

**Actual behavior**

- Example of a NOK correspondence between the clients:

https://github.com/googleapis/google-cloud-go/blob/54218e963bb5a6d47411a490985b54053825064f/iam/admin/apiv1/adminpb/iam.pb.go#L5724

**Additional context**

Initially reported at https://github.com/googleapis/google-cloud-go/issues/8064, but moved to a separate issue instead.

noahdietz:
So it seems that the `cloud.google.com/go/iam/admin/apiv1` client hasn't been regenerated in a long long time. I'm not sure of the history there, but I will take a look at getting it updated and back in the regen pipeline.

noahdietz:
I see why it hasn't been regenerated. In the same package, there is manually authored code that conflicts directly with the generated code, and the generated code was manually edited to allow for this. This renders this package impossible to regenerate automatically without a breaking change.

noahdietz:
As I try to manually rectify the breaking & uncompilable changes introduced by regeneration, I wonder if taking a major version would just be the best thing. This will only continue to be a problem (needing to manually update this one client) whenever we make generator changes/a new API feature is added. We could potentially add some manual code on the side that won't conflict with generated code, but keeps some of the convenience methods for working with the manual type `iam.Policy`...

List of changes that breaks compilation:
- `google-cloud-go/iam/admin/apiv1/policy_methods.go:28:21: method IamClient.GetIamPolicy already declared at google-cloud-go/iam/admin/apiv1/iam_client.go:504:21`
[...]

PR #11570, opened 2025-02-07
feat(iam/admin): regenerate client

codyoss:
This client has some handwritten code in it that causes failures when generating it normally. I generated the client and hand patched around handwritten code as to not create a breaking change. Changes made:

- Keep old variants of RPCs that did not generate an iterator helper to avoid breaking changes: ListRoles, QueryGrantableRoles, and QueryTestablePermissions
- Add variants of above RPCs that do return an iterator: ListRolesIter, QueryGrantableRolesIter, and QueryTestablePermissionsIter
- Preserve API for GetIamPolicy and SetIamPolicy.

Fixes: #8219


quartzmo (approved):
> I generated the client and hand patched around handwritten code

Would it be possible to write a detailed step-set for regularly reproducing what you did? This seems to be a semi-permanent situation...

codyoss:
@quartzmo I will write a small doc internally. I think what Noah suggested in the linked issue may be the correct long term solution here -- we make a v2 and fix this. What I did was not hard per-say, but it is not super clean or maintainable to do regularly. 

Or when we implement the selective generation feature, maybe that could help here.

PR #11571, opened 2025-02-07
fix: panic when explicitly beginning prepared transaction

rahul2393:


spanner/client.go
```diff
@@ -1005,7 +1005,12 @@ func (c *Client) prepareTransaction(ctx context.Context, options TransactionOpti
 	t := &ReadWriteTransaction{
 		txReadyOrClosed: make(chan struct{}),
 	}
+	t.txReadOnly.sp = c.idleSessions
 	t.txReadOnly.sh = sh
+	t.txReadOnly.txReadEnv = t
+	t.txReadOnly.qo = c.qo
+	t.txReadOnly.ro = c.ro
+	t.txReadOnly.disableRouteToLeader = c.disableRouteToLeader
```
rahul2393, line 1013: This is added because it's used in begin call 
https://github.com/googleapis/google-cloud-go/blob/spanner/v1.66.0/spanner/transaction.go#L786

qo, ro are not needed but I added it anyways

spanner/client_test.go
```diff
@@ -1833,6 +1833,55 @@ func TestClient_PreparedTransaction(t *testing.T) {
 	}
 }
 
+func TestClient_PreparedTransaction_SessionNotFoundOnBeginTransaction(t *testing.T) {
```
rahul2393, line 1836: Screenshot 2025-02-07 at 10 42 09 PM

olavloite (approved):

PR #11580, opened 2025-02-11
feat(firestore): Add String() method to DocumentChangeKind type

### LINKED ISSUE #11578: firestore: Add .String() method to DocumentChangeKind

I'm adding metrics to my database watch client. I want to know how many of each type of event we are seeing.

DocumentChangeKind is an integer so if we include that in our metric, we will see a number instead of a readable kind.

Add a .String() method to `type DocumentChangeKind int` so that we can see the name of the event kind instead of a number.

https://github.com/googleapis/google-cloud-go/blob/90140b17da6378fa87d4bec0d404c18a78d6b02a/firestore/watch.go#L40-L49

I'd like to make a PR for this, if it is approved. Here's the code I would add:

```
func (k firestore.DocumentChangeKind() string {
	switch k {
	case firestore.DocumentAdded:
		return "Added"
	case firestore.DocumentModified:
		return "Modified"
	case firestore.DocumentRemoved:
		return "Removed"
	default:
		return "Unknown"
	}
}
```

## Describe alternatives you've considered

Maintaining my own would be annoying and problematic if Firestore adds more event kinds.

nathanperkins:
The DocumentChangeKind.String() method returns a string representation of the kind:

* DocumentAdded: "DocumentAdded"
* DocumentModified: "DocumentModified"
* DocumentRemoved: "DocumentRemoved"
* Invalid value: "Unknown"

resolves #11578


firestore/watch.go
```diff
@@ -48,6 +48,20 @@ const (
 	DocumentModified
 )
 
+// String returns the string representation of the DocumentChangeKind.
+func (k DocumentChangeKind) String() string {
+	switch k {
+	case DocumentAdded:
+		return "Added"
```
bhshkh, line 55: DocumentAdded instead of Added
same for below

nathanperkins: Done

bhshkh (approved):

PR #11583, opened 2025-02-11
chore(spanner): support multiplexed session for Partitioned operations

rahul2393:
* Added support of using multiplexed session for all partitioned read, query and dml operations.
* Handled server side kill switch for multiplexed sessions with Partitioned Ops.


spanner/pdml.go
```diff
@@ -51,14 +51,19 @@ func (c *Client) partitionedUpdate(ctx context.Context, statement Statement, opt
...
+	}
 	if err != nil {
 		return 0, ToSpannerError(err)
 	}
-	if sh != nil {
-		defer sh.recycle()
+	if sh == nil {
+		return 0, spannerErrorf(codes.Internal, "no session available")
 	}
+	defer sh.recycle()
```
harshachinta, line 66: What does this mean for multiplexed session?

spanner/pdml.go
```diff
@@ -88,6 +93,23 @@ func (c *Client) partitionedUpdate(ctx context.Context, statement Statement, opt
 			if err == nil {
 				return count, nil
 			}
+			if isUnimplementedErrorForMultiplexedPartitionedDML(err) && sh.session.pool.isMultiplexedSessionForPartitionedOpsEnabled() {
+				sh.session.pool.disableMultiplexedSessionForPartitionedOps()
```
harshachinta, line 97: Are we trying to pick a regular session during failure for the transaction to proceed?
I guess this will complicate client lib logic. We should instead fail this transaction and all subsequent transactions using multiplexed sessions can fallback to regular.
Atleast this was how it was implemented for read-write to avoid addition client complexity.

spanner/client.go
```diff
@@ -867,13 +867,20 @@ func (c *Client) BatchReadOnlyTransaction(ctx context.Context, tb TimestampBound
 		err error
 	)
 
-	// Create session.
-	s, err = c.sc.createSession(ctx)
-	if err != nil {
-		return nil, err
+	if c.idleSessions.isMultiplexedSessionForPartitionedOpsEnabled() {
```
harshachinta, line 870: Where are we validating that the env for pops and global env for mux is set to true?

spanner/client.go
```diff
@@ -867,13 +867,20 @@ func (c *Client) BatchReadOnlyTransaction(ctx context.Context, tb TimestampBound
...
+		sh, err = c.idleSessions.takeMultiplexed(ctx)
```
harshachinta, line 871: In case of partitioned read/query methods how is it getting multiplexed sessions? I don't see any changes in this PR for them

spanner/pdml.go
rahul2393: @pratickchokhani please help clarify this.

pratickchokhani: For PDML, this is the case as the complication is trival.

pratickchokhani (approved):

spanner/session.go
```diff
@@ -1972,6 +1987,21 @@ func isUnimplementedErrorForMultiplexedRW(err error) bool {
...
+		return false
+	}
+	return ErrCode(err) == codes.Unimplemented && strings.Contains(err.Error(), "Transaction type partitioned_dml not supported with multiplexed sessions")
+}
+
+func isUnimplementedErrorForMultiplexedPartitionReads(err error) bool {
+	if err == nil {
+		return false
+	}
+	return ErrCode(err) == codes.InvalidArgument && strings.Contains(err.Error(), "Partitioned operations are not supported with multiplexed sessions")
```
pratickchokhani, line 2002: Please remove the InvalidArgument check here.

spanner/client.go
rahul2393: For partition read/query we are using BatchReadOnlyTransaction object which already has transaction created using multiplex session so they are just reusing it.

spanner/client.go
rahul2393: https://github.com/googleapis/google-cloud-go/pull/11583/files#diff-a7eef968d026e320839e33486c1ce4e449d69d9f6c86052edeae4528b60b5579R505

spanner/pdml.go
rahul2393: It will decrement the multiplex session usage metric https://github.com/googleapis/google-cloud-go/blob/main/spanner/session.go#L389

spanner/pdml.go
rahul2393: Added warning log for this fallback as discussed offline.

spanner/client.go
harshachinta: Got it.

PR #11584, opened 2025-02-11
fix(storage): fix adding multiple range on stream with same read id

shubham-diwakar:
Issue:
1. If we get a read add call we put it in map `mr.mp[curentID] = spec`
2. And if at that same point we receive an error from server we go into retry loop and close the stream manager go-routine.
3. During retry we see this new entry in map which is also retried.
4. Basically this creates two instances of curentID which we send to server. One due to add call one due to retry.

Fix:
1. Remove adding the current id in map in add call. As that can create duplicates as seen above.
2. Ignore the values from server in case entry is not found in map, given user will be notified if an entry is not in map by the help of callback.

### DIFF
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -1287,6 +1287,11 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params 
 					for _, val := range arr {
 						id := val.GetReadRange().GetReadId()
 						rr.mu.Lock()
+						_, ok := rr.mp[id]
+						if !ok {
+							// it's ok to ignore responses for read_id not in map as user would have been notified by callback.
+							continue
+						}
 						_, err = rr.mp[id].writer.Write(val.GetChecksummedData().GetContent())
 						if err != nil {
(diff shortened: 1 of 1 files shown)


storage/grpc_client.go
```diff
@@ -1458,12 +1465,13 @@ func (mr *gRPCBidiReader) add(output io.Writer, offset, limit int64, callback fu
 
 func (mr *gRPCBidiReader) wait() {
 	mr.mu.Lock()
-	keepWaiting := len(mr.mp) != 0 && mr.activeTask != 0
+	// we should wait until there is active task or an entry in the map.
+	keepWaiting := len(mr.mp) != 0 || mr.activeTask != 0
```
tritone, line 1469: Maybe explain that this is necessary b/c map can be empty during stream recreation.

shubham-diwakar: Done

shubham-diwakar:
Thanks for taking a look!!

tritone (approved):

Issue #11569, opened 2025-02-07
pstest: support listening on 0.0.0.0


larryanz:
## Is your feature request related to a problem? Please describe.
Currently, the server in `pstest` listens on `localhost` . This limits its accessibility to the local machine only, making it difficult to access the server from outside the container or other machines. In a Dockerized environment, this prevents external clients or other containers from communicating with the server. It would be easier to test and interact with the app if the server could be configured to listen on 0.0.0.0, allowing it to be accessed externally.

## Describe the solution you'd like
I would like to see an option (perhaps via a configuration flag or environment variable) to make the server listen on 0.0.0.0 instead of localhost. This would allow the app to be accessible from outside the container, enabling easier integration in containerized environments and network access for testing purposes.

For example:
```go
host := "0.0.0.0:3000"
srv := pstest.NewServerWithHost(host)
```

## Describe alternatives you've considered
You may wonder why I need to containerize it? It's because the pubsub-emulator works not stably in our case.

hongalex:
The code that configures what address to listen on is actually in `NewServerWithPort`, defined in `internal/testutil/server.go`. Since this affects more than just Pub/Sub, @codyoss or @quartzmo, wdyt about expanding that test package to support listening on custom addresses?

codyoss:
@hongalex I think we could accommodate that if needed

larryanz:
If we decided to support this, do we have any direction? Add a second method for the testutil?
I'm ready to help @hongalex

hongalex:
Yeah if you want to create a PR for this change, I'd be happy to review it. Adding the method to `testutil` and wrapping in pubsub fake seems correct.

PR #11586, opened 2025-02-12
feat(internal/testutil): support listening on custom host (testutil)

larryanz:
Regarding #11569, this PR adds support of pstest test server to listening on custom hosts rather than only localhost.

Due to CI running issue, this is the first part of the changes. Which only includes changes of testutil/server.
Part 2 is #11606


larryanz:
Hi @hongalex, please help to take a look

internal/testutil/server.go
```diff
@@ -52,16 +52,24 @@ type Server struct {
...
+// Addr field, without TLS.
 func NewServerWithPort(port int, opts ...grpc.ServerOption) (*Server, error) {
-	l, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", port))
+	return NewServerWithHost(fmt.Sprintf("localhost:%d", port))
+}
+
+// NewServerWithHost creates a new Server with a specific host and port. The
+// Server will be listening for gRPC connections at the address named by the
+// Addr field, without TLS.
+func NewServerWithHost(host string, opts ...grpc.ServerOption) (*Server, error) {
```
hongalex, line 71: Currently this function only takes in the address, and not a host. I recommend either renaming this to be `NewServerWithAddress` if you want to keep it as is, or have a separate option to specify the port if you want to keep the function name.

Same applies for the functions in `fake.go`

internal/testutil/server_test.go
```diff
@@ -29,13 +31,38 @@ func TestNewServer(t *testing.T) {
...
 	srv.Start()
-	conn, err := grpc.Dial(srv.Addr, grpc.WithInsecure())
+	conn, err := grpc.NewClient(srv.Addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
 		t.Fatal(err)
 	}
 	defer conn.Close()
 }
 
+func TestNewServerWithHost(t *testing.T) {
```
hongalex, line 41: nit: same as above, technically this tests valid addresses

hongalex (commented):
Looks good overall, but I have some concerns about naming

pubsub/pstest/fake.go
```diff
@@ -112,18 +112,26 @@ func NewServer(opts ...ServerReactorOption) *Server {
...
+// at the specified address (host and port).
+func NewServerWithAddress(address string, opts ...ServerReactorOption) *Server {
+	return NewServerWithCallback(address, func(*grpc.Server) { /* empty */ }, opts...)
+}
+
+// NewServerWithCallback creates new fake server running in the current process
+// at the specified address (host and port). Before starting the server, the
+// provided callback is called to allow caller to register additional fakes into
+// grpc server.
+func NewServerWithCallback(address string, callback func(*grpc.Server), opts ...ServerReactorOption) *Server {
```
hongalex, line 131: Changing the signature of `NewServerWithCallback` is a breaking change, and although we do allow for breaking changes within the fake, we should probably avoid it when it is easy to do so. Can you revert this change?

larryanz: updated, thanks!

larryanz:
hi @hongalex , seems the CI fails, could you please try it again?

hongalex (approved):

larryanz:
hi @hongalex , could you please try kokoro again? seems It cant be automatically merged

hongalex:
Yeah so the CI is failing because the pubsub fake is grabbing the head version of `cloud.google.com/go/internal` and not seeing the new method. Can you make this PR just the `testutil` changes and then make another PR with the pubsub fake changes?

larryanz:
seems something wrong with the CI due to fork/origin repo rebasing, if this PR cant be merge, please look at #11608 (@hongalex), thanks

quartzmo (approved):

Issue #11592, opened 2025-02-13
pubsub/pstest: message ordering fails


larryanz:
## Client

pstest

## Environment

- go version go1.23.4 darwin/arm64

## Code and Dependencies

Not any changes. But we can use this command to test the functionality without cached:
```bash
go test -timeout 30s -run ^TestSubscriptionMessageOrdering$ cloud.google.com/go/pubsub/pstest -count=1
```

## Expected behavior

Message ordering test must pass.

## Actual behavior

Message ordering unit test fails. Not always but sometimes.

## Screenshots

<img width="1276" alt="Image" src="https://github.com/user-attachments/assets/80c518f4-0c43-48c7-acf9-6268594c6e40" />

## Additional context
I printed the time of published messages and the received messages, and there's an issue with the time accuracy.
Publish Time of published messages are `time.Time`, but for received messages, It's protobuf` *timestamppb.Timestamp`

```bash
My Mac :: Documents/google-cloud-go/pubsub ‹main*› » go test -timeout 30s -run ^TestSubscriptionMessageOrdering$ cloud.google.com/go/pubsub/pstest -count=1
received_msg: m0 2025-02-18 14:53:41.862569 +0700 +07 m=+0.007351459 2025-02-18 14:53:41.862569000 0
....
received_msg: m18 2025-02-18 14:53:41.862595 +0700 +07 m=+0.007377293 2025-02-18 14:53:41.862595000 0
received_msg: m19 2025-02-18 14:53:41.862596 +0700 +07 m=+0.007378168 2025-02-18 14:53:41.862596000 0
received_msg: m20 2025-02-18 14:53:41.862596 +0700 +07 m=+0.007379126 2025-02-18 14:53:41.862596000 0
received_msg: m21 2025-02-18 14:53:41.862597 +0700 +07 m=+0.007380001 2025-02-18 14:53:41.862597000 0
received_msg: m22 2025-02-18 14:53:41.862598 +0700 +07 m=+0.007380918 2025-02-18 14:53:41.862598000 0
...
received_msg: m49 2025-02-18 14:53:41.862679 +0700 +07 m=+0.007461834 2025-02-18 14:53:41.862679000 0
msgs [ack_id:"m0"  message:{data:"hello"  message_id:"m0"  publish_time:{seconds:1739865221  nanos:862569000}  ordering_key:"ordering-key"}]
[...]

PR #11603, opened 2025-02-18
fix(pubsub/pstest): message ordering issue

larryanz:
As explained in #11592, currently the pstest's message ordering can not be ensured because of the accuracy of using comparison by `*timestamppb.Timestamp` instead of standard `time.Time`.

The solution is changing the comparison logic from proto time back to message published time.

## How to test/ re-produce the error:
Since this error happens quite randomly, we can use this script to do a fuzz test. The script will fail on main branch and succeed on this fix branch.
```bash
#!/bin/bash

# Default values for flags
print_output=false
count_tests=true
iterations=100

# Parse command-line flags
while getopts "pc:i:" opt; do
  case $opt in
    p) print_output=true ;;  # Enable printing output
    c) count_tests=true ;;   # Enable counting
    i) iterations=$OPTARG ;; # Set the number of iterations (default 1000)
    \?) echo "Usage: $0 [-p] [-c] [-i iterations]"
        exit 1 ;;
  esac
done

# Variables to count passes and fails
pass_count=0
fail_count=0

# Loop through test runs
for (( i=1; i<=iterations; i++ )); do
    output=$(go test -timeout 2s -run ^TestSubscriptionMessageOrdering$ cloud.google.com/go/pubsub/pstest -count=1)

    # Print the output if the flag is set
    if $print_output; then
        echo "$output"
    fi

    # Count pass/fail if the flag is set
    if $count_tests; then
        if echo "$output" | grep -q "ok\s*cloud.google.com/go/pubsub/pstest"; then
            ((pass_count++))
        elif echo "$output" | grep -q "FAIL\s*cloud.google.com/go/pubsub/pstest"; then
            ((fail_count++))
        fi
    fi

    # Print progress after each iteration
    if $count_tests; then
        echo "Iteration $i/$iterations - Passed: $pass_count, Failed: $fail_count"
    fi
done

# Print final count results if the flag is set
if $count_tests; then
    echo "Final Results: Passed tests: $pass_count, Failed tests: $fail_count"
fi

```


larryanz:
hi @hongalex, please help to review this PR

pubsub/pstest/fuzz_message_ordering.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
hongalex, line 1: Please do not include this file in the fix. Create a gist instead and add it to the PR description.

larryanz: hi @hongalex , I updated

hongalex (approved):

PR #11610, opened 2025-02-19
feat(storage): open subsequent streams in MRD in exceesive load condition

shubham-diwakar:
1. Open Multiple Streams in MRD in case of excessive load.
2. We have created a new flag limitPerStream to decide if we want to open multiple stream or not in case load is more than that.
3. If the value of flag is -1 it means we don't want to open a new stream.
4. For now limitPerStream flag is set to -1 for retry tests, emulator tests and integration tests.
5. For some of the integration tests turned on the flag with different values.
6. More through testing for this feature needs to happen, which will be picked up in follow-up PR.


cpriti-os:
@tritone this looks obsolete, Should we close this?

cpriti-os:
We would have to open a new PR for this feature, closing this.

PR #11614, opened 2025-02-20
chore(storage): Support a cookie header in calls.

cjc25:
This allows callers to run storage clients with an environment variable that propagates tracing cookies.


storage/invoke.go
```diff
@@ -39,8 +41,16 @@ var defaultRetry *retryConfig = &retryConfig{}
...
-	idempotencyHeaderKey = "x-goog-gcs-idempotency-token"
+	xGoogHeaderKey            = "x-goog-api-client"
+	idempotencyHeaderKey      = "x-goog-gcs-idempotency-token"
+	cookieHeaderKey           = "cookie"
+	directpathCookieHeaderKey = "x-directpath-tracing-cookie"
+)
+
+var (
+	cookieHeader = sync.OnceValue(func() string {
+		return os.Getenv("STORAGE_TRACING_COOKIE")
```
tritone, line 52: Let's do `GOOGLE_SDK_GO_TRACING_COOKIE` (matches prefix of some other env vars in the library).

tritone (approved):
Looks good to me. Also, changed PR type to "chore" since it's not an officially documented API so we don't want this in release notes.

storage/invoke.go
cjc25: Done, thanks

PR #11633, opened 2025-02-21
feat(spanner): report unclosed tracked sessions

egonelbre:
Any session that hasn't been returned to the pool when the pool is being closed is most likely a leak. Report all such sessions so that tests can capture the issue and fail as necessary.


egonelbre:
This helped to track down a bug in go-sql-spanner https://github.com/googleapis/go-sql-spanner/issues/380.

PR #11636, opened 2025-02-22
feat(storage): enable storage gRPC client to use DirectPath bound token if possible

rockspore:


rockspore:
@yamandabbagh JFYI

rockspore:
I've tested this on GCE and GKE and both worked. I can't assign reviewers. @tritone @quartzmo Could you help take a look?

tritone:
This seems fine, but let's wait until after the next (patch) release goes out to merge.

rockspore:
@tritone Looks like the patch has been released. Could you help merge this when you have a chance?

PR #11638, opened 2025-02-22
feat(spanner): add option for LastStatement in transaction

olavloite:
Adds an option to indicate that a statement is the last statement in a transaction. This allows Spanner to skip some validations during the execution of the statement, and instead rely on the validations during the Commit.

This option will also be used by the database/sql driver for statements that are executed directly on a connection instead of in a transaction.

This change also fixes a number of test cases so that the test time is reduced from 2 minutes to approx 5 seconds.


spanner/internal/testutil/inmem_spanner_server.go
```diff
@@ -681,48 +681,57 @@ func (s *inMemSpannerServer) getStatementResult(sql string) (*StatementResult, e
 
 func (s *inMemSpannerServer) simulateExecutionTime(method string, req interface{}) (interface{}, error) {
 	s.mu.Lock()
-	defer s.mu.Unlock()
```
olavloite, line 684: Keeping this lock here prevented the server from being shut down if there was a request still waiting for an execution time to finish (e.g. if an execution time of 1 minute had been set for an RPC, then keeping this lock would prevent the tests from stopping the server for 1 minute, which again would mean that the test would take 1 minute).

rahul2393 (approved):

spanner/transaction_test.go
```diff
@@ -1342,6 +1342,191 @@ func TestPriorityInQueryOptions(t *testing.T) {
...
+	requests := drainRequestsFromServer(server.TestSpanner)
+	executeRequests := requestsOfType(requests, reflect.TypeOf(&sppb.ExecuteSqlRequest{}))
+	if g, w := len(executeRequests), 2; g != w {
+		t.Fatalf("num requests mismatch\n Got: %v\nWant: %v", g, w)
+	}
+	for i := 0; i < len(executeRequests); i++ {
+		if g, w := executeRequests[i].(*sppb.ExecuteSqlRequest).LastStatement, i == len(executeRequests)-1; g != w {
+			t.Fatalf("%d: last statement mismatch\n Got: %v\nWant: %v", i, g, w)
+		}
+	}
```
olavloite, line 1371: This test shows how to use the `LastStatement` option in the client library. Setting the `LastStatement: true` option causes the client library to include the option in the request, which again allows Spanner to optimize the transaction execution.

PR #11647, opened 2025-02-24
feat(storage): Unfinalized appendable objects.

cjc25:
Allow callers to close an upload to an appendable object without finalizing it.


storage/writer.go
```diff
@@ -113,6 +113,13 @@ type Writer struct {
 	// yet available for general use.
 	Append bool
 
+	// FinalizeOnClose indicates whether the writer should finalize an object when
+	// closing the write stream. This only applies to writers where Append is
+	// true, since append semantics allow a prefix of the object to be durable and
+	// readable. For other writers, the correct way to abandon an upload is to
+	// cancel its context.
+	FinalizeOnClose bool
```
tritone, line 121: So IIUC this would mean that Appendable objects are never finalized by default. Does this square with how we expect appends to typically be used?

I was originally thinking we should make the param `CloseWithoutFinalize` to make finalization the default and ensure that people explicitly opt in to leaving open objects, but either way would be fine ultimately.

storage/writer.go
```diff
@@ -113,6 +113,13 @@ type Writer struct {
 	// yet available for general use.
 	Append bool
 
+	// FinalizeOnClose indicates whether the writer should finalize an object when
+	// closing the write stream. This only applies to writers where Append is
+	// true, since append semantics allow a prefix of the object to be durable and
+	// readable. For other writers, the correct way to abandon an upload is to
+	// cancel its context.
```
tritone, line 120: Add a preview tag to this doc as well.

storage/writer.go
tritone: Ah, I see that you make it true by default in NewWriter. I am not a fan of this because people expect default value of false for a bool in Go. If we want to default to finalization, let's do what I suggested as far as the param name.

tritone (approved):

storage/writer.go
```diff
@@ -107,12 +107,25 @@ type Writer struct {
...
+	// and can be appended to later.
 	//
 	// Append is only supported for gRPC. This feature is in preview and is not
 	// yet available for general use.
 	Append bool
 
+	// FinalizeOnClose indicates whether the writer should finalize an object when
+	// closing the write stream. This only applies to Writers where Append is
+	// true, since append semantics allow a prefix of the object to be durable and
+	// readable. By default, objects written with Append semantics will not be
```
BrennaEpp, line 121: nit: not sure the audience for this, but I would specify here something to the effect that the "prefix of the object" is the first N bytes or bytes uploaded so far; at least personally I'd never seen this term used this way outside of this and figure it might be easily confusable for object name prefix.

BrennaEpp (approved):

2FaceS-bit (approved):

PR #11650, opened 2025-02-24
test(storage): skip selected attrs subtest

tritone:
This is broken because of a service side issue.

Updates #11648


storage/integration_test.go
```diff
@@ -2540,35 +2540,38 @@ func testObjectsIterateSelectedAttrs(t *testing.T, bkt *BucketHandle, objects []
...
-	}
-	query.SetAttrSelection(selectedAttrs)
-
-	count := 0
-	it := bkt.Objects(context.Background(), query)
-	for {
-		_, err := it.Next()
-		if err == iterator.Done {
-			break
+	t.Run("testObjectsIterateAllSelectedAttrs", func(t *testing.T) {
```
BrennaEpp, line 2543: Should we do this for the other subtests as well?

PR #11653, opened 2025-02-25
test(storage): avoid inspecting creds in -short test

BrennaEpp:
If we don't provide the `GoogleAccessID` it attempts to detect it by inspecting the creds on the client/metadata; -short tests should not require real credentials and we don't need to test this in this case.


tritone (approved):
Thanks for this!

PR #11659, opened 2025-02-25
fix(spanner): avoid desructive context augmentation that dropped all headers

### LINKED ISSUE #11656: Spanner: x-goog-spanner-request-id related changes have caused other headers not to be forwarded
Reported offline in a bug report by Simil and team, after the reported changes, a bunch of headers aren’t being forwarded. This is an issue for myself to diagnose the problem and also to write tests for other headers so that in the future breakages shall be noticed.

odeke-em:
This change fixes a destructive context operation that dropped all prior headers, metadata.NewOutgoingContext sadly doesn't document that it doesn't reuse the input conext negating the expectation in Go that building from a parent context brings along the prior keys and metadata. While here, this chhange adds a regression test to ensure that in the future, any dropped or lost metadata will be reported during development.

Fixes #11656

### DIFF
--- a/spanner/regression_test.go
+++ b/spanner/regression_test.go
@@ -1,0 +1,214 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package spanner
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"maps"
+	"slices"
+	"sort"
+	"testing"
+
+	sppb "cloud.google.com/go/spanner/apiv1/spannerpb"
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/api/iterator"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/protobuf/types/known/structpb"
+
+	"cloud.google.com/go/spanner/internal/testutil"
+)
+
+type methodAndMetadata struct {
+	method string
+	md     metadata.MD
+}
+
+type ourInterceptor struct {
+	unaryHeaders  []*methodAndMetadata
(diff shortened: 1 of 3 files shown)


spanner/regression_test.go
```diff
@@ -0,0 +1,214 @@
...
+		t.Fatal(err)
+	}
+	defer sc.Close()
+
+	// 2. Perform a simple "SELECT 1" to trigger both unary and streaming gRPC calls.
+	sqlSELECT1 := "SELECT 1"
+	resultSet := &sppb.ResultSet{
+		Rows: []*structpb.ListValue{
+			{Values: []*structpb.Value{
+				{Kind: &structpb.Value_NumberValue{NumberValue: 1}},
```
olavloite, line 104: nit: not directly relevant to this test, as it does not read the value, but this encoding is not correct. INT64 values are encoded as string values. NumberValue is only used for FLOAT32 and FLOAT64. But it would be good to fix that here, as others might copy-paste this test later and run into unexpected errors when they try to read the data.
(If you want to verify that it fails: Try to do a `row.Scan(..)` in the loop below that iterates over the rows.)

spanner/regression_test.go
```diff
@@ -0,0 +1,214 @@
...
+	oi.unaryHeaders = append(oi.unaryHeaders, &methodAndMetadata{usi.FullMethod, md})
+	return handler(ctx, req)
+}
+
+// This is a regression test to assert that all the expected headers are propagated
+// along to the final gRPC server avoiding scenarios where headers got dropped from a
+// destructive context augmentation call.
+// Please see https://github.com/googleapis/google-cloud-go/issues/11656
+func TestAllHeadersForwardedAppropriately(t *testing.T) {
+	if isMultiplexEnabled {
```
olavloite, line 69: Instead of skipping if multiplexed sessions are enabled, can we disable multiplexed sessions during this test, and re-enable after the test?
Or alternatively only check for the requests that we know will be present regardless whether multiplexed sessions are being used or not (meaning: Only check for the presence of `BeginTransaction` and `ExecuteSql`, as those requests will be there in both cases.) The latter solution would also allow this test to be parallel.

olavloite (approved):
LGTM, with some nits on the test. We can pick that up in a follow-up PR as well if we want to ship this ASAP.

spanner/regression_test.go
odeke-em: @olavloite done, thanks and please take a look.

spanner/regression_test.go
odeke-em: Gotcha and thanks, done! Please take a look.

odeke-em:
@olavloite kindly help me re-run the bots!

PR #11674, opened 2025-02-28
chore(spanner): enable spanner gRPC client to use DirectPath bound token if possible

rockspore:
This enables spanner gRPC client to use DirectPath bound token if possible


rockspore:
Needs #11665 to be effective.

rockspore:
@olavloite @quartzmo Could you help PTAL? Even though it's ineffective without #11665, it doesn't do any harm. And I will make sure to verify the flow end to end before both this PR and #11665 are merged.

Thanks.

olavloite (approved):

PR #11687, opened 2025-03-05
fix(storage): modify the callback of mrd to return length of data read instead of limit.

shubham-diwakar:
Currently MRD takes a callback in each Add call.
Current call back returns the offset, length and error on that add call making the callback function look like, func(int64,int64,error) .
In some scenario data read can be different from limit set.
For example: We just have 50 bytes to read and ask for 100 bytes(length/limit) in this scenario we will return just 50 bytes.

Hence changed the callback to return amount of data read instead of data limit set.

Added the go doc for callback.


tritone (approved):

storage/integration_test.go
```diff
@@ -402,6 +402,55 @@ func TestIntegration_MultiRangeDownloader(t *testing.T) {
...
+		defer func() {
+			if err := obj.Delete(ctx); err != nil {
+				log.Printf("failed to delete test object: %v", err)
+			}
+		}()
+		reader, err := obj.NewMultiRangeDownloader(ctx)
+		if err != nil {
+			t.Fatalf("NewMultiRangeDownloader: %v", err)
+		}
+		res := make([]multiRangeDownloaderOutput, 3)
```
BrennaEpp, line 427: Why is this an array? Looks like we are only using the first var from it.

shubham-diwakar: Thanks, updated.

PR #11688, opened 2025-03-05
feat(storage): add CurrentState function to determine state of stream in MRD

shubham-diwakar:
Add a new method to MRD which tells about status of stream. 
GCSFuse will need this method to identify the status of stream so that they can work on recreation of MRD.

There are certain scenario they would like to re-create the stream. 

Example: storage.Reader (of bidi-conns) gets stuck for 15mins due to which MRD connection is also broken as it is stuck and server tears up the connection.
Example: https://b.corp.google.com/issues/392495592

Any other Add() call after that will result into error("stream is closed") hence it is beneficial to provide the status of stream.


storage/reader.go
```diff
@@ -439,3 +440,9 @@ func (mrd *MultiRangeDownloader) Wait() {
 func (mrd *MultiRangeDownloader) GetHandle() []byte {
 	return mrd.reader.getHandle()
 }
+
+// CurrentState returns the current state of the MultiRangeDownloader.
+// It returns true if the stream is still open and false if it is closed.
+func (mrd *MultiRangeDownloader) CurrentState() bool {
```
tritone, line 446: Maybe this should return an error rather than a bool?

I would expect:
* nil error = still open
* sentinel error (maybe call it `ErrStreamClosed`) if there was a graceful closure with all work completed.
* error that was returned by `MultiRangeDownloader.Close` otherwise.

That seems like it would give the most context to the caller.

BrennaEpp (commented):
Thank you!
+1 to Chris's comments

storage/reader.go
BrennaEpp: Actually, it looks like `close` always returns a nil error. I am also confused as to why we are returning only the `MultiRangeDownloader.Close` (are there no errors elsewhere [maybe the callback] that would be relevant? 

I am also unsure of how to detect a graceful closure (if there is no `activeTask` maybe?). @shubham-diwakar you probably know how to do that better than me, what do you think?

storage/reader.go
```diff
@@ -443,3 +444,9 @@ func (mrd *MultiRangeDownloader) Wait() {
 func (mrd *MultiRangeDownloader) GetHandle() []byte {
 	return mrd.reader.getHandle()
 }
+
+// ValidateOpen validates that the MultiRangeDownloader is open and can be used.
+// It returns an error if the stream is permanently closed.
+func (mrd *MultiRangeDownloader) ValidateOpen() error {
```
tritone, line 450: I still am not crazy about this name... Maybe we should just call this Error()? This will become more accurate once we make this get a meaningful error to clarify why there was a termination.

tritone (approved):

2FaceS-bit (approved):

PR #11754, opened 2025-03-05
feat(vertexai): support generate content based on request and not parts

amalucelli:
This change introduces two new methods that allow users to work with raw requests instead of `Part` objects:

- `GenerateContentFromRequest`
- `GenerateContentStreamFromRequest`

This will be useful when it's required to add history to the conversation mapping the right roles in a multi-turn conversation. At the moment, everything is identified as `user` generated content.


codyoss:
@amalucelli Before we review PRs we generally ask for issues open first for discussion to make sure this is something we want to accept. Would you mind opening and issue explaining the want here.

amalucelli:
I raised https://github.com/googleapis/google-cloud-go/issues/11756 for discussions around this, thank you for letting me know.

PR #11758, opened 2025-03-06
chore: fix failing vet checks

codyoss:
Followup from #11522


rahul2393 (approved):

rahul2393:
@codyoss Why was it not captured in presubmits?

PR #11791, opened 2025-03-10
feat(storage): Add Flush for appendable writes

tritone:
Adds Writer.Flush, which flushes any data in the internal buffer to GCS.

Only supported for appendable writes. Private preview feature.


storage/client_test.go
```diff
@@ -1052,6 +1053,94 @@ func TestOpenAppendableWriterMultipleFlushesEmulated(t *testing.T) {
...
+		}
+		if n != 0 {
+			t.Errorf("flushing at 0: got %v offset, want 0", n)
+		}
+
+		// Write first 2 MiB data, expect progress every 1 MiB.
+		_, err = w.Write(randomBytes3MiB[0 : 2*MiB])
+		if err != nil {
+			t.Fatalf("writing test data: got %v; want ok", err)
+		}
```
BrennaEpp, line 1092: Maybe try another edge case 0-byte flush here after writing the 2MIB?

storage/client_test.go
```diff
@@ -1052,6 +1053,94 @@ func TestOpenAppendableWriterMultipleFlushesEmulated(t *testing.T) {
...
+			}
+			if off != end {
+				t.Errorf("flushing 1k data: got %v bytes committed, want %v", off, end)
+			}
+		}
+
+		// Write remainder of data and close the writer.
+		if _, err := w.Write(randomBytes3MiB[2*MiB+3072:]); err != nil {
+			t.Fatalf("writing remaining data: got %v; want ok", err)
+		}
```
BrennaEpp, line 1117: I would also add a Flush here to check that `doneReading` etc works correctly if flushing after all bytes are written.

storage/grpc_client.go
```diff
@@ -3021,6 +3037,29 @@ func (w *gRPCWriter) read() (int, bool, error) {
...
+// flush flushes the current buffer regardless of whether it is full or not.
+// It's the implementation for Writer.Flush.
+func (w *gRPCWriter) flush() (int64, error) {
+	log.Printf("calling flush")
+	if !w.append {
+		return 0, errors.New("Flush is supported only if Writer.Append is set to true")
+	}
+
+	// Close PipeWriter to trigger EOF on read side of the stream.
+	w.flushInProgress = true
```
BrennaEpp, line 3049: I think this would be nicer as a bool chan - we only need to consume it in one place, right?
Also, I don't think it's getting set back to false anywhere (which wouldn't be necessary if it were a chan)?

tritone: This should be fixed now?

storage/client_test.go
tritone: Done.

storage/client_test.go
tritone: I wanted to have Close without a flush immediately preceding it; however I could add a separate test for this.

storage/writer.go
```diff
@@ -172,6 +173,43 @@ func (w *Writer) Write(p []byte) (n int, err error) {
...
+//
+// Do not call Flush concurrently with Write or Close. A single Writer is not
+// safe for unsynchronized use across threads.
+//
+// Flush is supported only on gRPC clients where [Writer.Append] is set
+// to true. This feature is in preview and is not yet available for general use.
+func (w *Writer) Flush() (int64, error) {
+	// Return error if Append is not true.
+	if !w.Append {
+		return 0, errors.New("Flush not supported unless client uses gRPC and Append is set to true")
```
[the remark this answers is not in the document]
tritone: Done.

storage/client_test.go
```diff
@@ -1052,6 +1053,120 @@ func TestOpenAppendableWriterMultipleFlushesEmulated(t *testing.T) {
...
+			}
+		}
+
+		// Do one more Flush that would require multiple messages (over 2 MiB).
+		n, err := w.Write(randomBytes9MiB[4*MiB+3072 : 6*MiB+4096])
+		if err != nil {
+			t.Fatalf("writing 2MiB + 1k data: got %v; want ok", err)
+		}
+		if n != 2*MiB+1024 {
+			t.Errorf("writing 2 MiB + 1k data: got %v bytes written, want %v", n, 1024)
```
ashmeenkaur, line 1121:
```suggestion
			t.Errorf("writing 2 MiB + 1k data: got %v bytes written, want %v", n, 2*MiB+1024)
```

storage/grpc_client.go
```diff
@@ -3014,13 +3033,38 @@ func (w *gRPCWriter) read() (int, bool, error) {
...
 	}
 	var done bool
-	if err == io.EOF {
+	if err == io.EOF && !w.flushInProgress {
+		err = nil
 		done = true
+	}
+
+	if w.flushInProgress {
 		err = nil
```
cjc25, line 3042: Does setting `err = nil` assume that `err == io.EOF` here? before we start to do the `w.flushInProgress` checks, should we return if `err != io.EOF`?

cjc25 (approved):

storage/client_test.go
```diff
@@ -1052,6 +1053,120 @@ func TestOpenAppendableWriterMultipleFlushesEmulated(t *testing.T) {
...
+			t.Errorf("flushing 0b data: got %v bytes committed, want %v", off, 6*MiB+4096)
+		}
+
+		// Write remainder of data and close the writer.
+		if _, err := w.Write(randomBytes9MiB[6*MiB+4096:]); err != nil {
+			t.Fatalf("writing remaining data: got %v; want ok", err)
+		}
+		if err := w.Close(); err != nil {
+			t.Fatalf("closing writer: %v", err)
+		}
```
ashmeenkaur, line 1146: shall we check w.flush fails after w.close?

storage/grpc_client.go
tritone: Good point, will do this.

storage/grpc_client.go
```diff
@@ -3014,13 +3033,38 @@ func (w *gRPCWriter) read() (int, bool, error) {
...
+		// Reset pipe
+		pr, pw := io.Pipe()
```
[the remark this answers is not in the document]
tritone: It's a good question. I tried a bunch of approaches before settling on this one. Unfortunately the call to Read on io.PipeReader cannot be interrupted other than by sending additional data through the pipe, or by closing the Pipe. Therefore even if we synchronize via channels, the Read call will still be hanging and waiting to receive data on the wrong part of the buffer.

Having a lock on Writer.Flush wouldn't affect things either way.

@cjc25 suggested as a follow up that we can replace io.PipeWriter here with our own pipe implementation that can be interrupted by a flush (and also have some other improvements that will allow for better pipelining for writes). We will plan on doing this as a follow up but it's a larger refactor.

storage/client_test.go
BrennaEpp: That's a good point; yeah I think we should include both.

storage/writer.go
```diff
@@ -172,6 +173,43 @@ func (w *Writer) Write(p []byte) (n int, err error) {
...
+	}
+	// If Flush called before any bytes written, it should start the upload
+	// at zero bytes. This will make the object visible with zero length data.
+	if !w.opened {
+		err := w.openWriter()
+		w.progress(0)
+		return 0, err
+	}
+
+	return w.flush()
```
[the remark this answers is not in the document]
tritone: Added this.

ashmeenkaur (approved):

BrennaEpp (approved):

PR #11793, opened 2025-03-10
feat(bigtable): Add support for materialized views for admin client

ron-gal:


bigtable/admin.go
```diff
@@ -2883,3 +2887,151 @@ func (iac *InstanceAdminClient) DeleteLogicalView(ctx context.Context, instanceI
...
+		default:
+			break
+		}
+	}
+	req := &btapb.CreateMaterializedViewRequest{
+		Parent:             "projects/" + iac.project + "/instances/" + instanceID,
+		MaterializedViewId: conf.MaterializedViewID,
+		MaterializedView:   mv,
+	}
+	_, err := iac.iClient.CreateMaterializedView(ctx, req)
```
bhshkh, line 2918: Why don't we wait for the lro to complete similar to CreateTableFromSnapshot ? Saw the same implementation in CreateAuthorizedView and CreateLogicalView as well
Is this intentional?

ron-gal: Good catch! Fixed for both MVs and LVs (not touching AVs, they have been live for a long time)

bhshkh (approved):

PR #11799, opened 2025-03-10
fix(spanner): revert the ALTS bound token enablement

rockspore:
This is to revert #11674.


rockspore:
@olavloite I'm very sorry for the inconvenience. Could you help merge this rollback asap? I will do a similar one in Java.

There is nothing wrong with the PRs but we found a bug at the server side so it will be safer if we don't enable it right now. I think fortunately #11674 is not in a released version yet.

olavloite (approved):

PR #11808, opened 2025-03-12
feat(civil): implement database/sql.Scanner|Valuer (#1145)

### LINKED ISSUE #1145: civil: support sql.Scanner and driver.Valuer
It would be great if e.g. civil.Date was directly usable in SQL queries.

derekperkins:
Fixes #1145. I wasn't sure how to test these in the context of this repo.

Questions:
- I tried to be comprehensive in the Scan functions, I'm not sure exactly what types of values to expect
- For the Valuer, I assumed that most/all databases should accept the ISO string values, but that may not be globally true


quartzmo:
@derekperkins Thanks for opening this PR. I'm trying to direct it to engineers directly involved with database products. Greetings from SLC, UT btw.

quartzmo:
cc @rahul2393

derekperkins:
> Greetings from SLC, UT btw.

Thanks, you too! You should come to the monthly meetup some time, we usually have ~20-30 people https://www.meetup.com/utahgophers/

quartzmo:
Thanks for the link! Just joined [utahgophers/](https://www.meetup.com/utahgophers/) using my personal acct.

rahul2393 (approved):

quartzmo:
@derekperkins Looks like your PR got approval! 🎉  Please click "Update branch" and I'll restart the CI, then merge when green.

quartzmo:
@derekperkins Oops, sorry but it looks like I need to approve this too. So, that means a harder look and I saw your question about testing. Are you able to run `civil/civil_test.go` and add coverage for your Scanner|Valuer funcs?

quartzmo:
@shollyman PTAL:

> Questions:

> * I tried to be comprehensive in the Scan functions, I'm not sure exactly what types of values to expect
> * For the Valuer, I assumed that most/all databases should accept the ISO string values, but that may not be globally true

derekperkins:
@quartzmo I went ahead and added some tests. They call the functions directly, vs routing through database/sql, but that seems sufficient for here.

civil/civil.go
```diff
@@ -158,6 +159,44 @@ func (d *Date) UnmarshalText(data []byte) error {
...
+		var err error
+		*d, err = ParseDate(string(vt))
+		return err
+	case *[]byte:
+		var err error
+		if vt != nil {
+			*d, err = ParseDate(string(*vt))
+		}
+		return err
+	}
```
quartzmo, line 196: From: 

```
	// Scan assigns a value from a database driver.
	//
	// The src value will be of one of the following types:
	//
	//    int64
	//    float64
	//    bool
	//    []byte
	//    string
	//    time.Time
	//    nil - for NULL values
	//
	// An error should be returned if the value cannot be stored
	// without loss of information.
```

Two questions:

1. Should int64 values be converted to Date?
2. How should float64 and bool values be handled? Does the error rule above apply here?

derekperkins: That's my bad, I should've had a default case that returned an error for other types. There's a case to be made for treating `int64` as a unix timestamp, and we could use some heuristics to guess whether it's seconds, milliseconds, etc., but that feels unnecessary.

derekperkins: I don't think so. People store seconds, milliseconds, microseconds, and nanoseconds. If they're doing that, they've already opted out of having date/time conveniences, and I lean towards just erroring on int64, same as we'd error for float64, or bool

quartzmo: Hmmm, OK, I suppose we can wait for someone to make a feature request for int64 support if they need it.

derekperkins: I added that default error for other scan types, and added additional testing for bad inputs.

quartzmo (approved):

civil/civil.go
derekperkins: > The time.Time type accepts a "format" value which may either be a Go identifier for one of the format constants (e.g., "RFC3339") or the format string itself to use with time.Time.Format or time.Parse (https://github.com/golang/go/issues/21990). It can also be "unix", "unixmilli", "unixmicro", or "unixnano" to be represented as a decimal number reporting the number of seconds (or milliseconds, etc.) since the Unix epoch.
https://github.com/golang/go/issues/71497#:~:text=The%20time.Time,the%20Unix%20epoch.

The new json/v2 proposal adds some tag level formatting that could be useful for this in the future. Assuming it gets merged as is, this would be a good route forward, to add some custom json support, and these db functions could use similar tagging to opt into specific formats

PR #11828, opened 2025-03-13
fix(auth): allow non-default SA credentials for DP

### LINKED ISSUE #11821: auth: EnableNonDefaultSAForDirectPath is ineffective
As far as I know, GCE MDS only supports fetching tokens for the default SA, which is the SA attached to the VM. So #11255 basically shadows the `InternalOptions.EnableNonDefaultSAForDirectPath` check with `tok.MetadataString("auth.google.tokenSource") != "compute-metadata"`.

I'm not sure what kind of xDS error that was seen before #11255, but I don't think that's the right fix. Right now, if I do `gcloud auth application-default login` on a GCE VM and use my user credentials to make gRPC requests, DirectPath won't be used even when it's enabled.

rockspore:
This effectively reverts #11255 with the slight difference in that I refactored out the compute engine token provider check as an individual func, which I will utilize in pending #11812. Because the bound token credentials should only be configured if users didn't provide credentials explicitly or the given one is compute engine type.

Fixes #11821.


rockspore:
Looking at the commit history of #11255, the auth package was using [gRPC 1.67.1](https://github.com/googleapis/google-cloud-go/blob/4347ca141892be8ae813399b4b437662a103bc90/auth/go.mod#L13) at that time. That was before I [fixed ](https://github.com/grpc/grpc-go/pull/7845) the issue in gRPC that caused "strange xDS error" with DirectPath using user credentials. The fix has been released since gRPC 1.69 and as of now 1.71 is used. cc @codyoss 

@quartzmo Could you PTAL? Thank you.

quartzmo:
Please update the commit message to match the PR title, thanks!

rockspore:
> Please update the commit message to match the PR title, thanks!

Done. Thanks for the tip!

auth/grpctransport/directpath.go
```diff
@@ -69,15 +69,22 @@ func isTokenProviderDirectPathCompatible(tp auth.TokenProvider, o *Options) bool
 	if tok.MetadataString("auth.google.tokenSource") != "compute-metadata" {
 		return false
 	}
-	if o.InternalOptions != nil && o.InternalOptions.EnableNonDefaultSAForDirectPath {
-		return true
-	}
 	if tok.MetadataString("auth.google.serviceAccount") != "default" {
```
quartzmo, line 72: What is the flow or use case in which "auth.google.tokenSource" == "compute-metadata" but "auth.google.serviceAccount" != "default"? Is there a better name for this func than `isTokenProviderComputeEngine` that would somehow express this possibility?

auth/grpctransport/directpath_test.go
```diff
@@ -41,28 +41,14 @@ func TestIsTokenProviderDirectPathCompatible(t *testing.T) {
...
-				}),
-			},
+			tp:   &staticTP{tok: &auth.Token{Value: "fakeToken"}},
 			opts: &Options{
 				InternalOptions: &InternalOptions{
 					EnableNonDefaultSAForDirectPath: true,
 				},
 			},
 			want: true,
 		},
```
quartzmo, line 51: Is there a test case in which "auth.google.tokenSource" == "compute-metadata" but "auth.google.serviceAccount" != "default"?

auth/grpctransport/directpath.go
rockspore: TBH, I'm not aware of any use case. My understanding is the metadata server can only issue access tokens for the default service account. This is understandable since it cannot impersonate any arbitrary SA as requested. The only public source I can find is https://cloud.google.com/compute/docs/access/authenticate-workloads#applications.

Therefore, I believe "auth.google.tokenSource" being "compute-metadata" indicates "auth.google.serviceAccount" has to be "default". It's technically redundant to check both. I am fairly new to this code base, and my guess is we are just making our best effort here to match all of them. Because all these APIs being public means it's theoretically possible for people to create a custom token provider that carries these metadata strings without actually getting the token from the MDS.

So I think `isTokenProviderComputeEngine` is correct. But if you think we only need to check "auth.google.tokenSource", that seems fine too. WDYT?

auth/grpctransport/directpath_test.go
rockspore: I will add a test case. But whether it ends up necessary depends on the outcome of the other comment.

auth/grpctransport/directpath.go
quartzmo: SGTM. Thanks for the well-considered explanation.

quartzmo (approved):

rockspore:
@quartzmo Thanks for the review. Could you help merge it? I don't have the permission.

PR #11854, opened 2025-03-17
perf(spanner): Skip gRPC trailers for StreamingRead & ExecuteStreamingSql

sakthivelmanii:
Fixes: b/394475581

**Description:**
gRPC trailers are headers which is usually sent after all the messages are sent by the server in the stream. Trailers usually causes additional latency of 250-300 μs after last message is sent in the stream. From the server, we will receiving last field which tells us about last PartialResultSet and we no longer will be waiting for trailers to assume that all the messages are received.


spanner/read_test.go
```diff
@@ -1672,6 +1672,74 @@ func setupStatementResult(t *testing.T, server *MockedSpannerInMemTestServer, st
...
+		noOfRows++
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if row.ColumnValue(0).GetStringValue() != "1" {
+			t.Fatalf("ID Mismatch in the result")
+		}
+	}
+	if noOfRows != 1 {
```
olavloite, line 1704: nit: can we make at least one of these tests a bit more interesting by having more than 1 row?

spanner/internal/testutil/inmem_spanner_server.go
```diff
@@ -144,20 +145,23 @@ func (s *StatementResult) ToPartialResultSets(resumeToken []byte) (result []*spa
 			} else {
 				rt = s.ResumeTokens[startIndex]
 			}
+			startIndex += rowCount
 			result = append(result, &spannerpb.PartialResultSet{
 				Metadata:    s.ResultSet.Metadata,
 				Values:      values,
 				ResumeToken: rt,
+				// set the last flag only for last PartialResultSet
+				Last: s.Last && startIndex == totalRows,
```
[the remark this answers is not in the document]
sakthivelmanii: `rowCount := min(totalRows-startIndex, uint64(MaxRowsPerPartialResultSet))`

As per row count logic, it looks like it works. Can you please check?

spanner/internal/testutil/inmem_spanner_server.go
```diff
@@ -102,6 +102,7 @@ type StatementResult struct {
 	ResultSet    *spannerpb.ResultSet
 	UpdateCount  int64
 	ResumeTokens [][]byte
+	Last         bool
```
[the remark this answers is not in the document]
sakthivelmanii: Done

spanner/read.go
```diff
@@ -50,13 +50,13 @@ func errEarlyReadEnd() error {
...
 // Spanner.
 func stream(
-	ctx context.Context,
-	logger *log.Logger,
-	meterTracerFactory *builtinMetricsTracerFactory,
-	rpc func(ct context.Context, resumeToken []byte, opts ...gax.CallOption) (streamingReceiver, error),
-	setTimestamp func(time.Time),
-	release func(error),
-	gsc *grpcSpannerClient,
+		ctx context.Context,
```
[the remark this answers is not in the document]
sakthivelmanii: Done

spanner/read_test.go
sakthivelmanii: Done

spanner/read_test.go
```diff
@@ -1672,6 +1672,75 @@ func setupStatementResult(t *testing.T, server *MockedSpannerInMemTestServer, st
...
+			t.Fatal(err)
+		}
+
+		if row.ColumnValue(0).GetStringValue() == "0" {
+			t.Fatalf("ID Mismatch in the result")
+		}
+	}
+	if noOfRows != 4 {
+		t.Fatalf("no of rows mismatch, Expected %v Got %v", 1, noOfRows)
+	}
```
olavloite, line 1706: nit, the best is to always use a construct like the following for this type of verification, as it makes sure that you print the same value as that you tested against:

```
if g, w := noOfRows, 4; g != w {
    t.Fatalf("num rows mismatch\n Got: %v\nWant: %v", g, w)
}
```

Using line feeds also helps readability of the error that is printed.

spanner/read_test.go
```diff
@@ -1672,6 +1672,75 @@ func setupStatementResult(t *testing.T, server *MockedSpannerInMemTestServer, st
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if row.ColumnValue(0).GetStringValue() == "0" {
+			t.Fatalf("ID Mismatch in the result")
+		}
+	}
+	if noOfRows != 5 {
+		t.Fatalf("no of rows mismatch, Expected %v Got %v", 1, noOfRows)
```
olavloite, line 1740: Same here as above; rewrite to a `if g, w := ...` style check, and use `g` and `w` in the message (the current message will print an illogical string, as you forgot to change the `1` to `5` when you changed the expected value)

spanner/read_test.go
```diff
@@ -1672,6 +1672,75 @@ func setupStatementResult(t *testing.T, server *MockedSpannerInMemTestServer, st
...
+		row, err := iter.Next()
+		if errors.Is(err, iterator.Done) {
+			break
+		}
+		noOfRows++
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if row.ColumnValue(0).GetStringValue() == "0" {
```
olavloite, line 1735: I think this would be a more relevant check:

```suggestion
		if row.ColumnValue(0).GetStringValue() != fmt.Sprintf("%v", noOfRows) {
```

olavloite: (Same for the other test function)

spanner/internal/testutil/inmem_spanner_server.go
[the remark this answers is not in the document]
olavloite: Sorry, you're right, that works

olavloite (approved):
LGTM, with some nits on the test cases

spanner/read_test.go
sakthivelmanii: done

spanner/read_test.go
sakthivelmanii: done. thanks

spanner/read_test.go
sakthivelmanii: done

yordis:
Hey team, I am chasing an issue with Spanner library, where we get "context canceled" in the traces, and I tracked the situation to v1.80.0

It only happens for `ExecuteStreamingSql`

Any thoughts?

rahul2393:
@yordis Can you please try https://github.com/googleapis/google-cloud-go/releases/tag/spanner%2Fv1.81.1, it has fix for this case.

yordis:
I will try `v1.81.1` but I was in `v1.83.0` and it was failing, which is why I went back and did 1 release at a time.

rahul2393:
If it's happening in 1.83.0 then it means that issue is still happening.

Only possible case I can think of is, we cancel the context when doing iter.Stop()
https://github.com/googleapis/google-cloud-go/blob/c7413916ad6ea7305f5842698748aa48773b30f3/spanner/read.go#L293
and then if we try to actually receive the message here
 https://github.com/googleapis/google-cloud-go/blob/c7413916ad6ea7305f5842698748aa48773b30f3/spanner/read.go#L681 it will raise context cancel exception, basically it should depend on timing of when and what thread executes first.
 
 @yordis in theory it should be happening for few ExecuteStreamingSql and not all the calls right?

yordis:
https://github.com/googleapis/google-cloud-go/pull/12630 I just need help with the testing situation, I made it work and prove the brokeness, but still, I may need your help

yordis:
> @yordis in theory it should be happening for few ExecuteStreamingSql and not all the calls right?

Yes, only `ExecuteStreamingSql`

> If it's happening in 1.83.0 then it means that issue is still happening.

Yes

yordis:
> Only possible case I can think of is, we cancel the context when doing iter.Stop()

I think we are doing `defer iter.Stop()` in the codebase, still, it was not reporting issues before this version, so the team lead decided to rollback

Issue #11855, opened 2025-03-17
vertexai: goroutine leak when closing a client


suhailpatel:
## Client

VertexAI 

## Environment

go version go1.23.6 darwin/arm64 but this should be replicable in all environments

I am using the following release of `cloud.google.com/go/vertexai` (which I believe to be the latest)
```
	cloud.google.com/go/vertexai => cloud.google.com/go/vertexai v0.13.3
```

## Code and Dependencies

When a `vertexai/genai` client is created and then closed, a gRPC callback serialiser instance is still left behind
```
3 @ 0x104f135b8 0x104eef1fc 0x1052d9414 0x104f1beb4
#	0x1052d9413	google.golang.org/grpc/internal/grpcsync.(*CallbackSerializer).run+0xc3	/Users/Suhail/Development/Go/src/github.com/suhailpatel/experimental/vendor/google.golang.org/grpc/internal/grpcsync/callback_serializer.go:88
```

This can be reproduced with this minimal program and inspecting the goroutine profile
```go
package main

import (
	"context"
	"fmt"
	"log"
	"net"
	"net/http"
	_ "net/http/pprof"
	"os"
	"os/signal"
	"syscall"

	"cloud.google.com/go/vertexai/genai"
)

func main() {
	go func() {
		listener, err := net.Listen("tcp", ":0")
		if err != nil {
			log.Fatalf("failed to listen: %v", err)
		}

		log.Println(fmt.Sprintf("Starting pprof server at %v", listener.Addr()))
		log.Println(http.Serve(listener, nil))
	}()

	ctx := context.Background()
	client, err := genai.NewClient(ctx, "suhailpatel-experimental", "us-central1")
	if err != nil {
		log.Fatalf("Error creating client: %v", err)
	}
	log.Println("Client created")

	client.Close()

	log.Println("Client closed")

	done := make(chan os.Signal, 1)
	signal.Notify(done, syscall.SIGINT, syscall.SIGTERM)
	<-done
}

```

## Expected behavior

No goroutines should remain when a client is created and then closed

## Actual behavior

There is a goroutine leak because we have a gRPC callback serialiser left behind

suhailpatel:
@codyoss I was wondering if you had a chance to review the patch I submitted to fix this here with some additional context https://github.com/googleapis/google-cloud-go/pull/11856. 

In the VertexAI code, we are constructing two gRPC clients, a prediction client and a cache client. 
[...]

PR #11856, opened 2025-03-17
fix(vertexai): close cache client to avoid leaking gRPC goroutines

suhailpatel:
Fixes issue #11855 

In the VertexAI client, we are creating a prediction client and a cache client. 
https://github.com/googleapis/google-cloud-go/blob/528d9ddc548c5f05237cf5a0cc4b762ca5f1dd31/vertexai/genai/client.go#L64-L76

In the `Close` method, we close the prediction client (`c.pc`) but do not close the cache client (`c.cc`)
https://github.com/googleapis/google-cloud-go/blob/528d9ddc548c5f05237cf5a0cc4b762ca5f1dd31/vertexai/genai/client.go#L100-L103

The `cacheClient` is a `aiplatform.GenAiCacheClient` as defined here and mentions that we should `Close` the client when not needed anymore

https://github.com/googleapis/google-cloud-go/blob/528d9ddc548c5f05237cf5a0cc4b762ca5f1dd31/vertexai/genai/caching.go#L31

https://github.com/googleapis/google-cloud-go/blob/528d9ddc548c5f05237cf5a0cc4b762ca5f1dd31/aiplatform/apiv1beta1/gen_ai_cache_client.go#L288-L292


codyoss (approved):
Thank you!

suhailpatel:
@codyoss I can't seem to trigger the Kokoro or Vet workflows, to get this automerged

> This workflow requires approval from a maintainer. [Learn more about approving workflows.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

Can you trigger these please and add the automerge label back when you have a chance 🙏

codyoss:
Thank you for you contribution! 🎆

PR #11866, opened 2025-03-18
feat(bigtable): adding node scaling factor

shantstepanian:


bigtable/admin.go
```diff
@@ -1507,6 +1511,31 @@ func (a *AutoscalingConfig) proto() *btapb.Cluster_ClusterAutoscalingConfig {
 	}
 }
 
+// NodeScalingFactor is the type of storage used for all tables in an instance
+type NodeScalingFactor int
```
bhshkh, line 1515:
```suggestion
type NodeScalingFactor int32
```

bigtable/admin.go
```diff
@@ -1238,6 +1238,9 @@ type InstanceConf struct {
 	// AutoscalingConfig configures the autoscaling properties on the cluster
 	// created with the instance. It is optional.
 	AutoscalingConfig *AutoscalingConfig
+
+	// Insert comment here
```
bhshkh, line 1242: Add proper comments everywhere

bigtable/admin.go
```diff
@@ -1507,6 +1511,31 @@ func (a *AutoscalingConfig) proto() *btapb.Cluster_ClusterAutoscalingConfig {
 	}
 }
 
+// NodeScalingFactor is the type of storage used for all tables in an instance
```
bhshkh, line 1514: Incorrect comment

bigtable/admin.go
```diff
@@ -1507,6 +1511,31 @@ func (a *AutoscalingConfig) proto() *btapb.Cluster_ClusterAutoscalingConfig {
...
+type NodeScalingFactor int
+
+// 1x better comment here.
+// 2x better comment there.
+const (
+	NODE_SCALING_FACTOR_1X NodeScalingFactor = iota
+	NODE_SCALING_FACTOR_2X
+)
```
bhshkh, line 1522:
```suggestion
const (
    // Proper comment here
    NodeScalingFactorUnspecified = iota
    // Proper comment here
    NodeScalingFactor1X
    // Proper comment here
    NodeScalingFactor2X
}
```

bigtable/admin.go
```diff
@@ -1507,6 +1511,31 @@ func (a *AutoscalingConfig) proto() *btapb.Cluster_ClusterAutoscalingConfig {
...
+	return btapb.Cluster_NODE_SCALING_FACTOR_1X
+}
+
+func nodeScalingFactorFromProto(st btapb.Cluster_NodeScalingFactor) NodeScalingFactor {
+	if st == btapb.Cluster_NODE_SCALING_FACTOR_2X {
+		return NODE_SCALING_FACTOR_2X
+	}
+
+	return NODE_SCALING_FACTOR_1X
+}
```
bhshkh, line 1537: remove this. Its used only once

bigtable/admin.go
```diff
@@ -1740,12 +1776,13 @@ func (iac *InstanceAdminClient) GetCluster(ctx context.Context, instanceID, clus
...
-		State:       c.State.String(),
-		StorageType: storageTypeFromProto(c.DefaultStorageType),
-		KMSKeyName:  kmsKeyName,
+		Name:              nameParts[len(nameParts)-1],
+		Zone:              locParts[len(locParts)-1],
+		ServeNodes:        int(c.ServeNodes),
+		State:             c.State.String(),
+		StorageType:       storageTypeFromProto(c.DefaultStorageType),
+		KMSKeyName:        kmsKeyName,
+		NodeScalingFactor: nodeScalingFactorFromProto(c.NodeScalingFactor),
```
bhshkh, line 1785:
```suggestion
		NodeScalingFactor: NodeScalingFactor(c.NodeScalingFactor.Number()),
```

go.work.sum
```diff
@@ -3,6 +3,7 @@ cloud.google.com/go/auth/oauth2adapt v0.2.0/go.mod h1:AfqujpDAlTfLfeCIl/HJZZlIxD
 cloud.google.com/go/dataproc v1.12.0 h1:W47qHL3W4BPkAIbk4SWmIERwsWBaNnWm0P2sdx3YgGU=
 cloud.google.com/go/gaming v1.9.0 h1:7vEhFnZmd931Mo7sZ6pJy7uQPDxF7m7v8xtBheG08tc=
 cloud.google.com/go/gaming v1.10.1 h1:5qZmZEWzMf8GEFgm9NeC3bjFRpt7x4S6U7oLbxaf7N8=
+cloud.google.com/go/resourcesettings v1.8.3 h1:13HOFU7v4cEvIHXSAQbinF4wp2Baybbq7q9FMctg1Ek=
```
bhshkh, line 6: Remove this file from diff

bigtable/integration_test.go
```diff
@@ -3630,6 +3630,65 @@ func TestIntegration_InstanceAdminClient_AppProfile(t *testing.T) {
...
+	}
+
+	cluster, err := iAdminClient.GetCluster(ctx, instanceToCreate, clusterID)
+	if err != nil {
+		t.Fatalf("GetCluster: %v", err)
+	}
+
+	wantNodeScalingFactor := NODE_SCALING_FACTOR_2X
+	if gotNodeScalingFactor := cluster.NodeScalingFactor; gotNodeScalingFactor != wantNodeScalingFactor {
+		t.Fatalf("want node scaling factor = %v, got = %v", wantNodeScalingFactor, gotNodeScalingFactor)
```
bhshkh, line 3688: Similar to rest of the errors, this should be
```suggestion
		t.Fatalf("NodeScalingFactor: got: %v, want: %v", gotNodeScalingFactor, wantNodeScalingFactor)
```

bigtable/integration_test.go
```diff
@@ -3630,6 +3630,65 @@ func TestIntegration_InstanceAdminClient_AppProfile(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("CreateInstanceWithClusters: %v", err)
+	}
+
+	cluster, err := iAdminClient.GetCluster(ctx, instanceToCreate, clusterID)
+	if err != nil {
+		t.Fatalf("GetCluster: %v", err)
+	}
+
+	wantNodeScalingFactor := NODE_SCALING_FACTOR_2X
```
bhshkh, line 3686: Move this to line 3660 and use `wantNodeScalingFactor` in line 3669. This way we can avoid the scenario where updating only one of the values (line 3669, line 3686) could cause the test to fail

bigtable/admin.go
```diff
@@ -1548,6 +1577,9 @@ type ClusterConfig struct {
 	// AutoscalingConfig configures the autoscaling properties on a cluster.
 	// One of NumNodes or AutoscalingConfig is required.
 	AutoscalingConfig *AutoscalingConfig
+
+	// Unit of scaling that allows more performance.
```
bhshkh, line 1581: Similar to rest of the comments, this comment should be
```suggestion
	// NodeScalingFactor specifies unit of scaling that allows more performance.
```

bigtable/integration_test.go
```diff
@@ -3630,6 +3630,65 @@ func TestIntegration_InstanceAdminClient_AppProfile(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("IntegrationEnv: %v", err)
+	}
+	defer testEnv.Close()
+
+	if !testEnv.Config().UseProd {
+		t.Skip("emulator doesn't support instance creation")
+	}
+
+	timeout := 5 * time.Minute
```
bhshkh, line 3649: nit: increase this to 7 or 10 min similar to some of the other tests

bigtable/admin.go
```diff
@@ -1507,6 +1511,31 @@ func (a *AutoscalingConfig) proto() *btapb.Cluster_ClusterAutoscalingConfig {
...
+	NODE_SCALING_FACTOR_1X NodeScalingFactor = iota
+	NODE_SCALING_FACTOR_2X
+)
+
+func (st NodeScalingFactor) proto() btapb.Cluster_NodeScalingFactor {
+	if st == NODE_SCALING_FACTOR_2X {
+		return btapb.Cluster_NODE_SCALING_FACTOR_2X
+	}
+	return btapb.Cluster_NODE_SCALING_FACTOR_1X
+}
```
bhshkh, line 1529: This function is not needed

bigtable/admin.go
```diff
@@ -1558,6 +1590,7 @@ func (cc *ClusterConfig) proto(project string) *btapb.Cluster {
 		EncryptionConfig: &btapb.Cluster_EncryptionConfig{
 			KmsKeyName: cc.KMSKeyName,
 		},
+		NodeScalingFactor: cc.NodeScalingFactor.proto(),
```
bhshkh, line 1593:
```suggestion
		NodeScalingFactor: btapb.Cluster_NodeScalingFactor(cc.NodeScalingFactor),
```

bigtable/integration_test.go
shantstepanian: Done

bigtable/admin.go
shantstepanian: I modeled these conversion proto methods off other enums in this file, e.g. StorageType and storageTypeFromProto on lines 1183-1196

Re: using `NodeScalingFactor(c.NodeScalingFactor.Number())` and `NodeScalingFactor: btapb.Cluster_NodeScalingFactor(cc.NodeScalingFactor)` for converting, what happens if we were to add another enum later in the proto without updating this client, and a request comes in w/ the new enum? This seems rare to happen, but I usually haven't converted between two separate enums like this before, so I want to double-check

bhshkh: Good point. Lets keep the existing implementation then

bigtable/admin.go
shantstepanian: Done

bigtable/admin.go
shantstepanian: SG - though I will make some changes on it based on your suggestion to add the "unspecified" enum

bigtable/admin.go
shantstepanian: Looking again - I actually found another place where this was used (both Clusters() and GetCluster()), so I think we can keep it. I modeled this after the existing storageTypeFromProto() method. It's helpful to see the conversion methods (to / from the proto) next to each other

bigtable/admin.go
shantstepanian: Updated now, and including the unspecified enums

bigtable/admin.go
shantstepanian: Per above comment,  keeping this as we use it in a couple places, and it helps to see it next to the other conversion method too

bigtable/admin.go
shantstepanian: Keeping the proto method per above comments

bhshkh (approved):

bhshkh:
LGTM

PR #11871, opened 2025-03-19
fix(storage): force first message on next sendBuffer when nothing sent on current

BrennaEpp:
If nothing is sent but the stream is reconnected, next send will not recognize the stream as reconnected without this change.

Also, update the conformance tests to catch this.


tritone (commented):
This looks good to me; we can wait to merge pending diagnosis of the other error you saw in your test repro with this change.

tritone (approved):

Issue #11873, opened 2025-03-19
bigquery: total rows has incorrect value


zecke:
## Client

bigquery

## Environment

Go 1.22

## Code and Dependencies

```go
package main

func main() {
    // ...
    var values []bigquery.Value
    err := it.Next(&values)
    if err == iterator.Done {
        return
    }
    if err != nil {
        // TODO: Handle error.
    }
    fmt.Println(it.TotalRows)
}
```

## Expected behavior

it.TotalRows() should have the correct value after the first call to `RowIterator.Next(context.Context)`

## Actual behavior

When the first call of `it.Next(ctx)` hits the paged cache the length of the cache is returned and not the total number of rows.

https://github.com/googleapis/google-cloud-go/blob/c2ee207621b2bb5fad8e6821892eae0041f469cd/bigquery/iterator.go#L409

PR #11874, opened 2025-03-19
fix(bigquery): totalRows returned by fetchCachedPage

zecke:
The RowIterator.TotalRows is wrong when hitting the cached page on the fast path (e.g. not changing start index or updating the page info).

fixes: #11873

### DIFF
--- a/bigquery/iterator.go
+++ b/bigquery/iterator.go
@@ -229,6 +229,7 @@ type rowSource struct {
 	cachedRows      []*bq.TableRow
 	cachedSchema    *bq.TableSchema
 	cachedNextToken string
+	cachedTotalRows uint64
 }
 
 // fetchPageResult represents a page of rows returned from the backend.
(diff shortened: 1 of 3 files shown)


alvarowolfx (changes requested):
Overall looks good, but can you add an integration test that shows that issue ? It happens when running two queries in sequence ?

zecke:
> Overall looks good, but can you add an integration test that shows that issue ? It happens when running two queries in sequence ?

Sure. Do you have any indication of the cost of running the BigQuery `integration_test.go`?

alvarowolfx:
> > Overall looks good, but can you add an integration test that shows that issue ? It happens when running two queries in sequence ?
> 
> Sure. Do you have any indication of the cost of running the BigQuery `integration_test.go`?

I don't recommended running the full suite, you can run just one test with a command like: `go test -run ^TestIntegration_StorageReadClientProject -v`. Not sure about the costs of running the full suite, but running just a few tests should be under the free tier. 

I can also take the work over and write the tests for it, so I can run the integration tests with my corp account here. Let me know what works best for you.

zecke:
> I can also take the work over and write the tests for it, so I can run the integration tests with my corp account here. Let me know what works best for you.

Thanks for the offer. Please take this over and write the test/make changes as needed.

alvarowolfx:
superseded by https://github.com/googleapis/google-cloud-go/pull/12230

PR #11881, opened 2025-03-19
fix(auth): Clone detectopts before use

yamandabbagh:
Clone detectopts before using it to make sure the value doesn't persist from previous calls of dial(). This issue was noticed in https://github.com/googleapis/google-cloud-go/pull/11665#issuecomment-2712018017


auth/grpctransport/grpctransport.go
```diff
@@ -304,6 +304,7 @@ func dial(ctx context.Context, secure bool, opts *Options) (*grpc.ClientConn, er
 			// This condition is only met for non-DirectPath clients because
 			// TransportTypeMTLSS2A is used only when InternalOptions.EnableDirectPath
 			// is false.
+			opts.DetectOpts.TokenBindingType = credentials.NoBinding
```
codyoss, line 307: I don't think we can do this always. I don't have the full context of this change as I did not review the prior PR, but this value can be user specified. If they set something we should honor it.

codyoss (changes requested):

auth/grpctransport/grpctransport.go
yamandabbagh: I see your point!

An alternative solution is to reset the field after its use in line: 318: creds, err = credentials.DetectDefault(opts.resolveDetectOptions())
This way if the user sets this field before calling Dial that value is honored. Something similar to here: https://github.com/googleapis/google-cloud-go/blob/10ad0a8c21cad5f7ea0a41d74e0fe143df3d18f5/auth/grpctransport/directpath.go#L121

WDYT?

codyoss: Yeah that code looks buggy to me. We should not be setting anything on the detect options directly. The resolve options func returns a clone for similar reasons. I think that bit of logic needs to be refactored.

yamandabbagh: we can have the resolveDetectOptions() set the  opts.DetectOpts.TokenBindingType on the clone, but we would need to pass the transportType to it (or set transportType in new option in opts which resolveDetectOptions() takes as a receiver).

yamandabbagh: @quartzmo can you take a look as well since you were involved in previous PRs.

quartzmo: @yamandabbagh In this PR's description, you wrote:

> to make sure the value doesn't persist from previous calls of Dial.

This is incorrect. The public func `Dial` is only called once per lifecycle of `opts`. The private func `dial` (small "d") is potentially called multiple times. Please update the description for accuracy.

quartzmo: @yamandabbagh @rockspore,

Regarding changing the value of `opts.DetectOpts.TokenBindingType` in `dial` (and in `configureDirectPath` called by `dial`), can you instead mutate the return value (clone) from `resolveDetectOptions` before it is passed to `DetectDefault`?

```
-                       opts.DetectOpts.TokenBindingType = credentials.ALTSHardBinding
-                       altsCreds, err := credentials.DetectDefault(opts.resolveDetectOptions())
-                       // Revert it back since the same opts will be used in subsequent dial() calls.
-                       opts.DetectOpts.TokenBindingType = credentials.NoBinding
+                       optsClone := opts.resolveDetectOptions()
+                       optsClone.TokenBindingType = credentials.ALTSHardBinding
+                       altsCreds, err := credentials.DetectDefault(optsClone)
```

rockspore: Yeah that's a good point. The goal is simply to avoid mutating the original `DetectOpts` and mutating the clone is much cleaner. Thanks.

yamandabbagh: +1 thank you for the suggestion. Thank you!

yamandabbagh: @quartzmo added the suggested fix to both the MTLSHardBinding and ALTSHardBinding. PTAL

yamandabbagh (commented):
fixed

quartzmo (approved):

codyoss (approved):

PR #11862, opened 2025-03-18
feat(bigtable): Add Bind function to bind query params

bhshkh:
To execute a SQL query, user needs to 
1) **Prepare** query
The "Prepare" step takes a SQL query text and some additional information and returns some state and an execution plan. Essentially, it processes the query and sets up what is needed to run it, before the actual execution of the query occurs.
2) **Bind** parameters
3) **Execute** the query

This PR adds support for the second step.

More context:
- Related protos are here: [request](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L1927-L1964), [response](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L2050-L2066), [API](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L2050-L2066)
- Design: go/cfdb-cbt-go-googlesql
- Previous BTQL PRs:
   https://github.com/googleapis/google-cloud-go/pull/11816
https://github.com/googleapis/google-cloud-go/pull/11829

PR #11829, opened 2025-03-13
feat(bigtable): Add PrepareStatement to run GoogleSQL queries

bhshkh:
To execute a SQL query, user needs to 
1) **Prepare** query
The "Prepare" step takes a SQL query text and some additional information and returns some state and an execution plan. Essentially, it processes the query and sets up what is needed to run it, before the actual execution of the query occurs.
2) **Bind** parameters
3) **Execute** the query

This PR adds support for the first step.

More context:
- Related protos are here: [request](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L1927-L1964), [response](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L2050-L2066), [API](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L2050-L2066)
- Design: go/cfdb-cbt-go-googlesql
- Previous BTQL PRs:
   https://github.com/googleapis/google-cloud-go/pull/11816

PR #11816, opened 2025-03-12
feat(bigtable): Add datatypes for GoogleSQL

bhshkh:
Adding SQL datatypes that can be used while preparing a query. These new datatypes are all wrappers over data API datatypes listed [here](https://github.com/googleapis/google-cloud-go/blob/3f23a9176f29a0a69b9d57b16f44b72eb3096d0c/bigtable/admin/apiv2/adminpb/types.pb.go#L70-L81).

Bigtable does not yet support Map and Struct as parameter types in SQL queries
 
Design: go/cfdb-cbt-go-googlesql

PR #11883, opened 2025-03-20
feat(bigtable): Add Execute method to execute GoogleSQL queries

bhshkh:
To execute a SQL query, user needs to 
1) **Prepare** query
The "Prepare" step takes a SQL query text and some additional information and returns some state and an execution plan. Essentially, it processes the query and sets up what is needed to run it, before the actual execution of the query occurs.
2) **Bind** parameters
3) **Execute** the query

This PR adds support for the third step.
Most of the implementation of `func (bs *BoundStatement) execute` is similar to [readrows](https://github.com/googleapis/google-cloud-go/blob/24c5ae0ab119b2b7ba54ba0ea9c5432cdc5ec28f/bigtable/bigtable.go#L557-L698)
Data() method on ResultRow to access the results will be added in next PR: https://github.com/bhshkh/google-cloud-go/pull/5 .

More context:
- Related protos are here: [request](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L1927-L1964), [response](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L2050-L2066), [API](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L2050-L2066)
- Design: go/cfdb-cbt-go-googlesql
- Previous BTQL PRs:
   https://github.com/googleapis/google-cloud-go/pull/11816
   https://github.com/googleapis/google-cloud-go/pull/11829
   https://github.com/googleapis/google-cloud-go/pull/11862
   https://github.com/googleapis/google-cloud-go/pull/11892


bigtable/bigtable.go
```diff
@@ -507,6 +546,221 @@ func (ps *PreparedStatement) Bind(values map[string]any) (*BoundStatement, error
...
+	}
+	ps.refreshMutex.Lock()
+	defer ps.refreshMutex.Unlock()
+
+	// Recheck whether validity changed while acquiring lock
+	if ps.valid() {
+		return nil
+	}
+
+	return ps.refresh(ctx)
```
jackdingilian, line 561: in Java we trigger a background refresh 1 second ahead of plan expiry, but still use the current plan until it completes. Can we do that here as well? So that it doesn't add latency every time the plan expires.

bigtable/bigtable.go
```diff
@@ -507,6 +546,221 @@ func (ps *PreparedStatement) Bind(values map[string]any) (*BoundStatement, error
...
+			var batchData []byte
+			if partialResultSet.GetProtoRowsBatch() != nil {
+				batchData = partialResultSet.GetProtoRowsBatch().GetBatchData()
+				ongoingResultBatch.Write(batchData)
+			}
+
+			// If `resume_token` is non-empty and any data has been received since the
+			// last one, BatchChecksum is guaranteed to be non-empty. In other words, a batch will
+			// never cross a `resume_token` boundary. It is an error otherwise
+			if partialResultSet.GetResumeToken() != nil &&
```
jackdingilian, line 679: I think this can be made clearer if it's moved below, so that the logic is:
- If batchChecksum -> process batch
- If resumeToken -> check onGoingResultBatch is empty and yield rows

bigtable/bigtable.go
```diff
@@ -507,6 +546,221 @@ func (ps *PreparedStatement) Bind(values map[string]any) (*BoundStatement, error
...
+			if partialResultSet.GetResumeToken() != nil {
+				// Values can be yielded to the caller
+
+				// Save ResumeToken for subsequent requests
+				resumeToken = partialResultSet.GetResumeToken()
+
+				if bs.ps.metadata == nil || bs.ps.metadata.GetProtoSchema() == nil {
+					return errors.New("bigtable: metadata missing")
+				}
+				cols := bs.ps.metadata.GetProtoSchema().GetColumns()
```
jackdingilian, line 715: metadata should be stored separately at the start of the request and should only be changed when the plan is refreshed with the logic discussed above. The schema can change on refresh (for SELECT * when the column families change).  So we can't use the latest metadata from the prepared statement like this

bigtable/bigtable.go
```diff
@@ -507,6 +546,221 @@ func (ps *PreparedStatement) Bind(values map[string]any) (*BoundStatement, error
...
+				numCols := len(cols)
+
+				// Parse rows
+				for len(valuesBuffer) != 0 {
+					var completeRowValues []*btpb.Value
+
+					// Pop first 'numCols' values to create a row
+					completeRowValues, valuesBuffer = valuesBuffer[0:numCols], valuesBuffer[numCols:]
```
jackdingilian, line 723: Can we check for a size mismatch here and throw a better error than index out of bounds? It should never happen but we want to make sure we throw an exception in case of a bug.

jackdingilian (changes requested):

bigtable/bigtable.go
```diff
@@ -186,11 +194,37 @@ var (
...
+				Initial:    100 * time.Millisecond,
+				Max:        2 * time.Second,
+				Multiplier: 1.2,
+			}
+			return &bigtableRetryer{
+				additionalRetryCondition: func(err error) bool {
+					apiErr, ok := apierror.FromError(err)
+					if ok && apiErr != nil && apiErr.Details().PreconditionFailure != nil {
+						for _, violation := range apiErr.Details().PreconditionFailure.GetViolations() {
+							if violation != nil && violation.GetType() == queryExpiredViolationType {
```
jackdingilian, line 210: this needs to trigger the preparedStatement refresh. It's possible in the future we will return this error before time based expiry

bigtable/bigtable.go
bhshkh: I did not understand.
When the plan is refreshed, the metadata in ps gets updated too. So, why can't the latest metadata from ps be used?

bigtable/bigtable.go
```diff
@@ -507,6 +546,221 @@ func (ps *PreparedStatement) Bind(values map[string]any) (*BoundStatement, error
...
+	var resumeToken []byte
+
+	err := gaxInvokeWithRecorder(ctx, mt, "ExecuteQuery", func(ctx context.Context, headerMD, trailerMD *metadata.MD, _ gax.CallSettings) error {
+		ctx, cancel := context.WithCancel(ctx) // for aborting the stream
+		defer cancel()
+
+		var err error
+		err = bs.ps.refreshIfInvalid(ctx)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
bhshkh: Yes, the complete code block in gaxInvokeWithRecorder is rerun if this err is retryable.

bigtable/bigtable.go
jackdingilian: For example: 
- I start the request with ps at t1
- A column family is added to the table
- some other request triggers refresh of ps
- I get the metadata here from the refreshed ps at t2
- Metadata contains the new column family, the responses do not (because the request used the plan from t1)

bigtable/bigtable.go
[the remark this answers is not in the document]
jackdingilian: It's a bit of a weird case but even if there is a non-retriable error, we want to make sure the prepared statement is still usable. We know that the PrepareRequest is valid, so we don't want an error here to fail all execute requests.

As long as subsequent execute requests retry (which they do) I guess this is probably ok

bigtable/bigtable.go
```diff
@@ -205,7 +239,8 @@ func containsAny(str string, substrs []string) bool {
 }
 
 func (r *bigtableRetryer) Retry(err error) (time.Duration, bool) {
-	if status.Code(err) == codes.Internal && containsAny(err.Error(), retryableInternalErrMsgs) {
+	if (status.Code(err) == codes.Internal && containsAny(err.Error(), retryableInternalErrMsgs)) ||
+		(r.additionalRetryCondition != nil && r.additionalRetryCondition(err)) {
```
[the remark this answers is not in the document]
bhshkh: Renamed

bigtable/bigtable.go
bhshkh: Moved

bigtable/bigtable.go
bhshkh: Stored metadata separately

bigtable/bigtable.go
bhshkh: Added check

bigtable/bigtable.go
bhshkh: Triggered

bigtable/bigtable.go

bigtable/bigtable.go
```diff
@@ -507,6 +553,296 @@ func (ps *PreparedStatement) Bind(values map[string]any) (*BoundStatement, error
...
+				// never cross a `resume_token` boundary. It is an error otherwise
+				if ongoingResultBatch.Len() != 0 &&
+					partialResultSet.BatchChecksum == nil {
+					prevError = errors.New("bigtable: received resume_token with buffered data and no batch_checksum")
+					return prevError
+				}
+
+				if !receivedResumeToken {
+					// first ResumeToken received
+					finalizedMetadata = bs.ps.metadata
```
jackdingilian, line 780: this needs to be synchronized with the preparedQuery we use. It's still possible that metadata here has changed since we got preparedQuery from ps above. We should probably make it so they are returned together, so that we can guarantee they are from the same prepare response

bigtable/bigtable_test.go
```diff
@@ -1681,3 +1688,418 @@ func TestPreparedStatement_Bind(t *testing.T) {
...
+				nil,
+				io.EOF,
+			},
+			wantExecErr: errors.New("bigtable: server stream ended without sending a resume token"),
+		},
+		{
+			desc: "retry on expired query FailedPrecondition error",
+			prepQueryResps: []*btpb.PrepareQueryResponse{
+				getPrepareQueryResponse(), // PrepareStatement
+				getPrepareQueryResponse(), // Execute
```
jackdingilian, line 1784: Can we assert that it uses the preparedQuery and metadata from the second response (and return different values)?

jackdingilian: Oh relatedly can we also add a case for something along these lines:
- prepareResponse(preparedQuery: a, md: [colA])
- executeQuery(batch of data, not token)
- unavailable
- executeQuery(from start)
- plan refresh error
- prepareResponse(preparedQuery: b, md: [colA, colB])
- successful execute stream with the new schema, make sure we discard the first batch

bigtable/bigtable.go
```diff
@@ -777,18 +788,21 @@ func (bs *BoundStatement) execute(ctx context.Context, f func(ResultRow) bool, m
 
 				if !receivedResumeToken {
 					// first ResumeToken received
-					finalizedMetadata = bs.ps.metadata
+					fs = &finalizedStatement{
+						metadata:      bs.ps.metadata,
+						preparedQuery: bs.ps.preparedQuery,
```
jackdingilian, line 793: sorry this isn't exactly what I meant, the metadata still could have changed since `execPreparedQuery` was set above.

I think PreparedStatement should return a struct like this (maybe we a different name in that context). And then we pull metadata and preparedQuery from there and update the full struct on refresh only

bigtable/bigtable_test.go
bhshkh: Added

bigtable/bigtable.go
bhshkh: Updated the code

bigtable/bigtable.go
bhshkh: doned

bigtable/bigtable_test.go
```diff
@@ -1681,3 +1688,582 @@ func TestPreparedStatement_Bind(t *testing.T) {
...
+				{err: io.EOF},
+			},
+			wantExecReqPrepQuerys: [][]byte{
+				[]byte(preparedQuery1),
+				[]byte(preparedQuery1),
+			},
+			wantResultRowValues: [][]*btpb.Value{newProtoRowValuesWithKey(colFamAddress)},
+		},
+		{
+			/*
```
jackdingilian, line 1903: Sorry one more test case we should have that I left out before:

- PrepareQuery
- ExecuteQuery
- receive resume token, no data
- receive resume token, no data
- EOF

Expect this to work ok (should work fine w current implementation)

jackdingilian (changes requested):
Mostly Looking good, just want to resolve comment about potential race condition on refresh

bigtable/bigtable_test.go
bhshkh: Added

bigtable/bigtable.go
```diff
@@ -507,6 +555,332 @@ func (ps *PreparedStatement) Bind(values map[string]any) (*BoundStatement, error
...
+// if the prepared query is valid and has not reached the early expiration threshold.
+func (ps *PreparedStatement) valid() (valid bool, validEarly bool) {
+	nowTime := time.Now().UTC()
+	expireTime := ps.validUntil.AsTime()
+	return nowTime.Before(expireTime), nowTime.Add(preparedQueryExpireEarlyDuration).Before(expireTime)
+}
+
+func (ps *PreparedStatement) refresh(ctx context.Context) error {
+	newPs, err := ps.c.prepareStatementWithMetadata(ctx, ps.query, ps.paramTypes, ps.opts...)
+	ps.metadata = newPs.metadata
```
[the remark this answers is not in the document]
bhshkh: Good point. Updated the code

gkevinzheng (approved):

bigtable/bigtable.go
```diff
@@ -659,17 +658,12 @@ func (bs *BoundStatement) Execute(ctx context.Context, f func(ResultRow) bool, o
...
-}
-
-func newFinalizedStatement(metadata *btpb.ResultSetMetadata, query []byte) *finalizedStatement {
-	return &finalizedStatement{
+func newPreparedQueryData(ps *PreparedStatement) *preparedQueryData {
+	return &preparedQueryData{
 		// Make a deep copy. Even if preparedStatement is refreshed, this wouldn't change
-		metadata:      proto.Clone(metadata).(*btpb.ResultSetMetadata),
-		preparedQuery: query,
+		metadata:      proto.Clone(ps.data.metadata).(*btpb.ResultSetMetadata),
```
jackdingilian, line 664: I think there is still a race here. If ps.data is updated between accessing the metadata and the preparedQuery.

We need to do 
data = ps.data

return {
  data.preparedQuery,
  data.metadata,
... 
}

bhshkh: fixed

jackdingilian (approved):

PR #11884, opened 2025-03-20
ci(spanner): Create Spanner Benchmarking Script

sakthivelmanii:


spanner/benchmarks/benchmarks.go
```diff
@@ -0,0 +1,156 @@
...
+		if time.Now().Local().After(endTime) {
+			break
+		}
+		duration, err := executeQuery(ctx, client)
+		if err != nil {
+			fmt.Println(err)
+			return make([]int64, 0), err
+		}
+		durations = append(durations, duration)
+		time.Sleep(time.Millisecond * time.Duration(waitBetweenRequests))
```
olavloite, line 114: maybe randomize the sleep time in the range [0, 2*sleep]

spanner/benchmarks/benchmarks.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	return durations, nil
+}
+
+func executeQuery(ctx context.Context, client *spanner.Client) (int64, error) {
+	start := time.Now()
+
+	iter := client.Single().Query(ctx, spanner.Statement{SQL: selectQuery, Params: map[string]interface{}{
```
olavloite, line 123: Should we use a stale query here?

spanner/benchmarks/benchmarks.go
```diff
@@ -0,0 +1,156 @@
...
+		"p1": generateUniqueID(),
+	}})
+	for {
+		_, err := iter.Next()
+		if errors.Is(err, iterator.Done) {
+			break
+		}
+		if err != nil {
```
olavloite, line 131: Note that this means that you are never decoding the results from protobuf objects to Go objects. That is OK, but should be taken into account when compared to for example the Java results, if the latter does decode (which it does by default)

spanner/benchmarks/benchmarks.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+}
+
+func percentiles(percentile int, latencies []int64) any {
+	rank := ((percentile / 100) * (len(latencies) - 1)) + 1
+	return latencies[rank]
+}
+
+func generateUniqueID() int64 {
+	return rand.Int64N(totalRecords)
```
olavloite, line 155: nit: should this be

```suggestion
	return rand.Int64N(totalRecords) + 1
```

spanner/benchmarks/benchmarks.go
sakthivelmanii: Eventually we will support both. For now I am creating it for query. I will raise a PR to do it for read as well

olavloite: I meant add __staleness__ for the query. IIRC, we wanted to add staleness to make sure that the query is can be executed as quickly as possible by Spanner.

spanner/benchmarks/benchmarks.go
```diff
@@ -124,13 +126,17 @@ func executeQuery(ctx context.Context, client *spanner.Client) (int64, error) {
...
 		if errors.Is(err, iterator.Done) {
 			break
 		}
 		if err != nil {
 			return time.Duration(0).Microseconds(), err
 		}
+
+		if row.ColumnValue(0).GetStringValue() == "" {
+			return time.Duration(0).Microseconds(), errors.New("empty ID in the table")
+		}
```
olavloite, line 139: This does not decode the value, this just gets the protobuf string value. What you should do is this:

```
var id int64
if err := row.Columns(&id); err != nil {
	return err
}
```

That will check what type the column is, see that it is an `INT64`, and then convert the underlying string value to an int64.

spanner/benchmarks/benchmarks.go
```diff
@@ -152,5 +158,9 @@ func percentiles(percentile int, latencies []int64) any {
 }
 
 func generateUniqueID() int64 {
-	return rand.Int64N(totalRecords)
+	return rand.Int64N(totalRecords) + 1
+}
+
+func getRandomWaitTime(waitTime int64) time.Duration {
+	return time.Duration(rand.Int64N(waitTime-1) + 1)
```
olavloite, line 165: I think that this should be:

```suggestion
	return time.Duration(rand.Int64N(2 * waitTime-1) + 1)
```

that will ensure that when you specify for example 10ms as the wait time, then the average wait time will be around 10ms.

spanner/benchmarks/benchmarks.go
```diff
@@ -0,0 +1,156 @@
...
+		fmt.Println("Please set warm up time, execution time and wait between requests in the command line")
+		return
+	}
+
+	warmupTime, _ := strconv.ParseInt(os.Args[1], 10, 8)          // in minutes
+	executionTime, _ := strconv.ParseInt(os.Args[2], 10, 8)       // in minutes
+	waitBetweenRequests, _ := strconv.ParseInt(os.Args[3], 10, 8) // in milliseconds
+
+	db := fmt.Sprintf("projects/%v/instances/%v/databases/%v", project, instance, database)
+
```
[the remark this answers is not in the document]
sakthivelmanii: Done

spanner/benchmarks/benchmarks.go
sakthivelmanii: got it. Adding that also.

olavloite (approved):

PR #11898, opened 2025-03-24
feat(spanner): allow string values for Scan functions

olavloite:
Support scanning string values for the various
spanner.NullTypes.

Use the Scan implementation of the underlying data type for spanner.NullDate, as civil.Date now implements the sql.Scanner interface.

The Scan implementations are used by the database/sql driver and applications that depend on the database/sql driver.


derekperkins:
related issue:
- #1145

PR #11921, opened 2025-03-28
test(bigtable): Correct the partial batch testcase

bhshkh:
if complete batch data is abcd and it's split into 2 chunks, result sets should be 
```
partialResultSet {
    protoRowsBatch{ 
        batchData: 'ab'
    }
    reset = true
}


partialResultSet {   
    protoRowsBatch{       
        batchData: 'cd'   
    }   
    reset = false
    batch_checksum = crc32c(abcd)
} 

```

Currently, newExecQueryRespPartialBatchFirstHalf returns a resp with batchCheckSum. This means it is actually a complete batch and not just first half of the batch. i.e 

```
partialResultSet {
    protoRowsBatch{ 
        batchData: 'ab'
    }
    reset = true
    batch_checksum = crc32c(ab)
}


partialResultSet {   
    protoRowsBatch{       
        batchData: 'cd'   
    }   
    reset = false
    batch_checksum = crc32c(cd)
} 

```
Updating the test to fix this


jackdingilian (approved):
Can you add a test case that tests partial batch reset?

- newExecQueryRespPartialBatchFirstHalf(reset=true, data="foo")
- unavailable
- newExecQueryRespPartialBatchFirstHalf(reset=true, data="bar")
- newExecQueryRespPartialBatchSecondHalf(reset=false, data="baz")
- EOF

And we should only have bar,baz as the results

bhshkh:
> Can you add a test case that tests partial batch reset?
> 
> * newExecQueryRespPartialBatchFirstHalf(reset=true, data="foo")
> * unavailable
> * newExecQueryRespPartialBatchFirstHalf(reset=true, data="bar")
> * newExecQueryRespPartialBatchSecondHalf(reset=false, data="baz")
> * EOF
> 
> And we should only have bar,baz as the results

Added

gkevinzheng (approved):

jackdingilian (approved):

PR #11924, opened 2025-03-28
feat(bigtable): Add methods to read ResultRow

bhshkh:
To execute a SQL query, user needs to 
1) **Prepare** query
The "Prepare" step takes a SQL query text and some additional information and returns some state and an execution plan. Essentially, it processes the query and sets up what is needed to run it, before the actual execution of the query occurs.
2) **Bind** parameters
3) **Execute** the query

This PR adds support to access the results in ResultRow. The surface is pretty minimal for initial release. More methods will be added in upcoming releases.
While retrieving data for a column, the value types are converted as follows:
- Timestamp is returned as time.Time
- Date is returned as civil.Date (to denote that the date does not contain time or time zone information) instead of using time.Time

   bigtablepb Date_Value

   https://github.com/googleapis/google-cloud-go/blob/dd0d1d7b41884c9fc9b5fe808139cccd29e1e486/bigtable/apiv2/bigtablepb/data.pb.go#L501-L504

   date.Date

   https://github.com/googleapis/go-genproto/blob/b45e905df463ff5f9668293553ec0e2bda3b4ce4/googleapis/type/date/date.pb.go#L38-L51

Note: In go/cfdb-cbt-go-googlesql, `Data()` was finalized to be in the initial implementation but it introduces a lot of complexity. So, will be included in later releases.

More context:
- Related protos are here: [request](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L1927-L1964), [response](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L2050-L2066), [API](https://github.com/googleapis/google-cloud-go/blob/b323f888e349168b4e30a9edb4630dfb6060ed4d/bigtable/apiv2/bigtablepb/bigtable.pb.go#L2050-L2066)
- Design: go/cfdb-cbt-go-googlesql
- Previous BTQL PRs:
   https://github.com/googleapis/google-cloud-go/pull/11816
   https://github.com/googleapis/google-cloud-go/pull/11829
   https://github.com/googleapis/google-cloud-go/pull/11862
   https://github.com/googleapis/google-cloud-go/pull/11892
   https://github.com/googleapis/google-cloud-go/pull/11883
   https://github.com/googleapis/google-cloud-go/pull/11921


bigtable/result_row.go
```diff
@@ -0,0 +1,799 @@
...
+		return nil, fmt.Errorf("bigtable: error converting column %d (%q): %w", index, rr.Metadata.Columns[index].Name, err)
+	}
+	return goVal, nil
+}
+
+// GetByName returns the value of the first column found with the specified name.
+// The returned value will be of the corresponding Go type.
+// Returns an error if no column with the specified name is found.
+// Column name matching is case-sensitive.
+func (rr *ResultRow) GetByName(name string) (any, error) {
```
jackdingilian, line 122: I thought the signature here was going to be `GetByName(name string, dst any) error`

(caveat that Im not a golang person) That seems preferable because users wont need to cast any to the appropriate type every call

bigtable/result_row.go
```diff
@@ -0,0 +1,799 @@
...
+			}
+			fieldSQLType, err := pbTypeToSQLType(fieldPbType)
+			if err != nil {
+				return nil, fmt.Errorf("invalid struct field %q type: %w", f.GetFieldName(), err)
+			}
+			structFields[i] = StructSQLField{Name: f.GetFieldName(), Type: fieldSQLType}
+		}
+		return StructSQLType{Fields: structFields}, nil
+	default:
+		return nil, fmt.Errorf("unsupported protobuf type kind: %T", k)
```
jackdingilian, line 464: Nit: this might happen when users query new (future) data types with old clients. Can we make the error message something a little more user friendly, like "Unrecognized response type: %T. You might need to upgrade your client"

jackdingilian: Also, it works correctly right now but we want to make sure the schema is always checked before we return a value to users, in case of this happening, so we don't mistake the value as being null

bigtable/result_row.go
```diff
@@ -0,0 +1,799 @@
...
+		arrValProto, ok := pbVal.Kind.(*btpb.Value_ArrayValue)
+		if !ok {
+			return nil, fmt.Errorf("type mismatch: expected ArrayValue for ArrayType, got %T", pbVal.Kind)
+		}
+		elemPbType := k.ArrayType.GetElementType()
+		if elemPbType == nil {
+			return nil, errors.New("array element type is nil")
+		}
+
+		if arrValProto.ArrayValue == nil || len(arrValProto.ArrayValue.Values) == 0 {
```
jackdingilian, line 619: Ther's a difference between Null array and empty array, these should be separate cases

bigtable/result_row.go
```diff
@@ -0,0 +1,799 @@
...
+	return rr.GetByIndex(indices[0])
+}
+
+// GetAllByName returns a slice containing the values of *all* columns matching the
+// specified name, in the order they appear in the result set.
+// If no columns match the name, it returns (nil, nil).
+// The values in the returned slice will be of their corresponding Go types.
+// Returns an error if any value conversion fails.
+// Column name matching is case-sensitive.
+func (rr *ResultRow) GetAllByName(name string) ([]any, error) {
```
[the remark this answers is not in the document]
bhshkh: removed and returned error from GetByName

bigtable/result_row.go
```diff
@@ -0,0 +1,799 @@
...
+		pbValues:   values,
+		pbMetadata: metadata,
+		Metadata:   *rrMetadata,
+	}, nil
+}
+
+// newResultRowMetadata returns the schema of the result row, describing the name and type of each column.
+// The order of columns matches the order of values returned by [ResultRow.Scan].
+func newResultRowMetadata(metadata *btpb.ResultSetMetadata) (*ResultRowMetadata, error) {
+	protoSchema := metadata.GetProtoSchema()
```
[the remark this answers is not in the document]
bhshkh: Added check

bigtable/result_row.go
bhshkh: Modified error message

bigtable/result_row.go
bhshkh: https://github.com/googleapis/google-cloud-go/pull/11924/files#diff-09f386e918f9faeac5778d669637d232d22fdcf61ae8a23d6e7e5fe76ae27ae8R290-R293

Does Kind being nil not denote nil array ?

bigtable/bigtable.go
```diff
@@ -737,6 +738,9 @@ func (bs *BoundStatement) execute(ctx context.Context, f func(ResultRow) bool, m
 		for {
 			proto.Reset(eqResp)
 			err := stream.RecvMsg(eqResp)
+
+			bytes, _ := protojson.MarshalOptions{AllowPartial: true, UseEnumNumbers: true, Multiline: true}.Marshal(eqResp)
+			fmt.Println("eqResp: " + string(bytes))
```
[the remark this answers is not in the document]
bhshkh: Removed

bigtable/result_row.go
bhshkh: Taking dst as input needs a lot of reflect usage similar to the code I had written in Scan and DataTo. This keeps the surface minimal for initial release

bigtable/result_row.go
```diff
@@ -0,0 +1,799 @@
...
+}
+
+// ResultRowMetadata provides information about the schema of the ResultRow
+type ResultRowMetadata struct {
+	// the order of values returned by [ResultRow.Scan].
+	Columns []ColumnMetadata
+}
+
+func newResultRow(values []*btpb.Value, metadata *btpb.ResultSetMetadata) (*ResultRow, error) {
+	rrMetadata, err := newResultRowMetadata(metadata)
```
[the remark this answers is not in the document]
bhshkh: Done

bigtable/result_row.go
bhshkh: Done

bigtable/result_row.go
bhshkh: separated

bigtable/result_row.go
```diff
@@ -0,0 +1,801 @@
...
+func pbTypeToGoReflectType(pbType *btpb.Type) (reflect.Type, error) {
+	return pbTypeToGoReflectTypeInternal(pbType, true)
+}
+
+// pbValueToGoValue converts a protobuf Value to a standard Go value (any).
+// Base types -> T (e.g. int64, string), []byte -> []byte,
+// Arrays -> []*T (e.g. []*int64), Maps -> map[K]*V, Structs -> map[string]any.
+// errors returned should be wrapped before returning to the end user.
+func pbValueToGoValue(pbVal *btpb.Value, pbType *btpb.Type) (any, error) {
+	if pbVal == nil || pbVal.Kind == nil {
```
jackdingilian, line 358: pbVal itself should never be nil, that case should probably error out

bigtable/result_row.go
```diff
@@ -0,0 +1,801 @@
...
+		// Represent SQL NULL as Go's nil interface value.
+		return nil, nil
+	}
+	if pbType == nil {
```
jackdingilian, line 362: Can we move this above the valid nil check? This should result in an error no matter what

bigtable/result_row.go
```diff
@@ -0,0 +1,801 @@
...
+	}
+
+	// Special case: []byte <-> string
+	// If dest is string, and src is []byte
+	if dest.Kind() == reflect.String && srcVal.Kind() == reflect.Slice && srcVal.Type().Elem().Kind() == reflect.Uint8 {
+		dest.SetString(string(srcVal.Bytes()))
+		return nil
+	}
+	// If dest is []byte, and src is string
+	if dest.Kind() == reflect.Slice && dest.Type().Elem().Kind() == reflect.Uint8 && srcVal.Kind() == reflect.String {
```
jackdingilian, line 795: Do we need this one? I'd prefer to disallow accidental implicit conversions if we don't need it to handle go map keys

bigtable/result_row.go
```diff
@@ -0,0 +1,801 @@
...
+					return fmt.Errorf("error assigning map value type %s to interface{}: %w", srcValue.Type(), err)
+				}
+				dest.SetMapIndex(destKey, destValue)
+			}
+			return nil
+		}
+	}
+
+	// Common numeric conversions (allow assigning int64 to int, float64 to float32 etc.)
+	if srcVal.CanConvert(dest.Type()) {
```
jackdingilian, line 783: I would vote against allowing implicit conversions like this. I think we should make users pass the exact types and they can convert the output as they wish.

bigtable/result_row.go
```diff
@@ -0,0 +1,801 @@
...
+		} // Return empty map
+		pbEntries := mapArrProto.ArrayValue.Values
+		for i, pbEntry := range pbEntries {
+			kvPairProto, ok := pbEntry.Kind.(*btpb.Value_ArrayValue)
+			if !ok || kvPairProto.ArrayValue == nil || len(kvPairProto.ArrayValue.Values) != 2 {
+				return nil, fmt.Errorf("invalid map entry format at index %d", i)
+			}
+			pbKey := kvPairProto.ArrayValue.Values[0]
+			pbValue := kvPairProto.ArrayValue.Values[1]
+			goKey, errK := pbValueToGoValue(pbKey, keyPbType)
```
jackdingilian, line 494: Can you leave a comment explaining how this handles []byte. My understanding is we have a map[string]any but this returns []byte, and that is converted within assignValue below.

bigtable/result_row.go
bhshkh: pbValueToGoValue is recursively called on the array elements. Can any of the array elements be nil ?

bigtable/result_row.go
```diff
@@ -0,0 +1,801 @@
...
+
+	if src == nil {
+		// Assigning nil. Check if dest is nillable.
+		switch dest.Kind() {
+		case reflect.Interface, reflect.Ptr, reflect.Map, reflect.Slice, reflect.Chan, reflect.Func:
+			// Assign typed nil (zero value of the destination type)
+			dest.Set(reflect.Zero(dest.Type()))
+			return nil
+		default:
+			// Cannot assign nil to non-nillable types like int, string, bool, struct.
```
jackdingilian, line 570: Maybe add something to this message saying you should use a pointer type if your value is nullable

bigtable/result_row_test.go
```diff
@@ -0,0 +1,424 @@
...
+	typeFloat64        = &btpb.Type{Kind: &btpb.Type_Float64Type{Float64Type: &btpb.Type_Float64{}}}
+	typeBool           = &btpb.Type{Kind: &btpb.Type_BoolType{BoolType: &btpb.Type_Bool{}}}
+	typeTimestamp      = &btpb.Type{Kind: &btpb.Type_TimestampType{TimestampType: &btpb.Type_Timestamp{}}}
+	typeDate           = &btpb.Type{Kind: &btpb.Type_DateType{DateType: &btpb.Type_Date{}}}
+	typeArrayString    = &btpb.Type{Kind: &btpb.Type_ArrayType{ArrayType: &btpb.Type_Array{ElementType: typeString}}}
+	typeMapStringBytes = &btpb.Type{Kind: &btpb.Type_MapType{MapType: &btpb.Type_Map{KeyType: typeString, ValueType: typeBytes}}}
+	typeMapStringInt64 = &btpb.Type{Kind: &btpb.Type_MapType{MapType: &btpb.Type_Map{KeyType: typeString, ValueType: typeInt64}}}
+	typeStructSimple   = &btpb.Type{Kind: &btpb.Type_StructType{StructType: &btpb.Type_Struct{Fields: []*btpb.Type_Struct_Field{
+		{FieldName: "name", Type: typeString}, {FieldName: "count", Type: typeInt64},
+	}}}}
```
jackdingilian, line 90: here and elsewhere can we test the with_history map type. Map<Bytes, Array<Struct<Timestamp timestamp, Bytes value>>>

bigtable/result_row.go
jackdingilian: no, null is represented as a Value message with no kind set. It should never be null

bigtable/integration_test.go
```diff
@@ -5135,13 +5304,97 @@ func TestIntegration_Execute(t *testing.T) {
...
+					},
+					{
+						name:     "boolCol",
+						gotDest:  false,
+						wantDest: true,
+					},
+					{
+						name:     "tsCol",
+						gotDest:  time.Time{},
+						wantDest: time.Now(),
```
jackdingilian, line 5350: This isn't using a fake clock is it? I would expect this to fail if not, since time should equal the time from above

bigtable/integration_test.go
```diff
@@ -5135,13 +5304,97 @@ func TestIntegration_Execute(t *testing.T) {
...
+					},
+					{
+						name:     "dateCol",
+						gotDest:  civil.Date{},
+						wantDest: civil.DateOf(time.Now()),
```
jackdingilian, line 5355: This will also fail if it's run on either side of a date boundary I guess

bigtable/result_row.go
```diff
@@ -0,0 +1,801 @@
...
+// It performs basic type conversions. It converts columns to the following Go types where possible:
+//   - string
+//   - []byte
+//   - int64 (and other integer types like int, int32, uint64 etc.)
+//   - float32, float64
+//   - bool
+//   - time.Time (for TIMESTAMP)
+//   - civil.Date (for DATE)
+//   - Slice types (e.g., []string, []int64) for ARRAY
+//   - Map types (e.g., map[string]any) for MAP or STRUCT
```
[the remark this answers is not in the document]
bhshkh: Added special type Struct

bigtable/result_row.go
bhshkh: Moved

bigtable/result_row.go
bhshkh: removed

bigtable/result_row.go
bhshkh: Added

bigtable/result_row.go
```diff
@@ -91,6 +92,123 @@ func newResultRowMetadata(metadata *btpb.ResultSetMetadata) (*ResultRowMetadata,
...
+}
+
+// Len returns the number of fields in the Struct.
+func (s Struct) Len() int {
+	return len(s.fields)
+}
+
+// Field returns the name and value of the field at the specified zero-based index.
+// Returns an error if the index is out of bounds.
+func (s Struct) Field(index int) (name string, value any, err error) {
```
jackdingilian, line 129: Do we need this? Id vote to just support GetByIndex for now for simplicity

jackdingilian: I think we can also remove Fields. Users can use the struct schema to iterate fields for now, and we can decide what syntax sugar we want for this in the future

bigtable/result_row_test.go
```diff
@@ -0,0 +1,424 @@
...
+		{"any slice to int slice fail nil", []any{int64(1), nil}, func() any { var v []int64; return &v }, true, nil},
+		{"any slice to ptr int slice ok", []any{int64(1), nil, int64(3)}, func() any { var v []*int64; return &v }, false, []*int64{int64Ptr1, nil, int64Ptr3}},
+		{"any slice to ptr int slice fail type", []any{int64(1), "nil"}, func() any { var v []*int64; return &v }, true, nil},
+
+		// map Conversions
+		{"map ptr val to map val ok", map[string]*int64{"a": int64Ptr1}, func() any { var v map[string]int64; return &v }, false, map[string]int64{"a": 1}},
+		{"map ptr val to map val fail nil", map[string]*int64{"a": nil}, func() any { var v map[string]int64; return &v }, true, nil},
+		{"map val to map ptr val", map[string]int64{"a": 1}, func() any { var v map[string]*int64; return &v }, false, map[string]*int64{"a": int64Ptr1}},
+		{"map to map stringany", map[string]*int64{"a": int64Ptr1, "b": nil}, func() any { var v map[string]any; return &v }, false, map[string]any{"a": int64Ptr1, "b": (*int64)(nil)}},
+
```
gkevinzheng, line 415: Could you add some test cases for more complex map/slice conversions for this test and for `TestResultRow_GetByIndex`? The conversion logic for more map/slice datatypes could get very complex, and you want to test that the code correctly assigns something like 

`struct(string, array<map<string, array<*int64>>>, map<string, map<string, any>>)` to a `struct(string, array<map<string, array<int64>>>, map<string, map<string, array<array<array<array<array<*int64>>>>>>`

gkevinzheng (approved):

bigtable/result_row_test.go
bhshkh: Added

bigtable/integration_test.go
```diff
@@ -5055,22 +5056,125 @@ func TestIntegration_Execute(t *testing.T) {
...
+						name:     "sum",
+						gotDest:  map[string]int64{},
+						wantDest: map[string]int64{},
+					},
+				},
+			},
 		},
 		{
 			desc:    "WITH_HISTORY",
 			psQuery: "SELECT _key, " + colFam + "['state'] AS state FROM `" + table.table + "`(WITH_HISTORY=>TRUE) LIMIT 5",
```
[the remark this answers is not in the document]
bhshkh: Added

bigtable/integration_test.go
bhshkh: Not a fake clock. I was ignoring seconds  precision.  Changed it to ignore millisecond

bigtable/integration_test.go
bhshkh: Updated

bigtable/result_row.go
bhshkh: Removed

bigtable/bigtable.go
```diff
@@ -524,17 +538,20 @@ func (c *Client) prepareStatement(ctx context.Context, mt *builtinMetricsTracer,
 // Allowed parameter value types are []byte, string, int64, float32, float64, bool,
 // time.Time, civil.Date, array, slice and nil
 func (ps *PreparedStatement) Bind(values map[string]any) (*BoundStatement, error) {
-	bs := BoundStatement{
-		ps:     ps,
-		params: map[string]*btpb.Value{},
+	// check that every parameter is bound
+	for paramName := range ps.paramTypes {
```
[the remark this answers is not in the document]
bhshkh: Added check for ps.
The code won't panic even if ps.paramTypes is nil

bigtable/result_row.go
bhshkh: Added

bigtable/result_row_test.go
bhshkh: When a struct's pointer is passed in the fields of existing struct are overwritten. I can add a testcase for struct with the fields struct(string, array<map<string, array<*int64>>>, map<string, map<string, any>>)

bhshkh: Added

jackdingilian (approved):

PR #11931, opened 2025-03-31
feat(bigquery): support managed iceberg tables

shollyman:
This CL plumbs in the new configuration object to configured managed tables for apache iceberg.


bigquery/table.go
```diff
@@ -157,13 +157,16 @@ type TableMetadata struct {
 	MaxStaleness *IntervalValue
 
 	// The tags associated with this table. Tag
-	// keys are globally unique. See additional information on tags
+	// keys are globally unique. See additional information owwn tags
```
alvarowolfx, line 160: typo here

bigquery/table.go
```diff
@@ -380,6 +383,66 @@ func bqToMaterializedViewDefinition(q *bq.MaterializedViewDefinition) *Materiali
 	}
 }
 
+// Supported
+type BigLakeFileFormat string
```
alvarowolfx, line 387: I wonder if those two new types should be declared separately and have their own comment to avoid the vet issues.

shollyman: Done.

alvarowolfx (approved):

PR #11934, opened 2025-03-31
fix(storage): fix panic when Flush called early

tritone:
Writer.Flush was panicking due to blocking content sniffing logic if Writer.Flush was called before 512 bytes.

Re-ordering calls here allows Flush to signal EOF and unblock the content sniffer if a flush happens that early.

Add a test with multiple small Flushes to validate that this works as expected.


storage/grpc_client.go
```diff
@@ -1722,24 +1722,23 @@ func (c *grpcStorageClient) OpenWriter(params *openWriterParams, opts ...storage
...
 			}
-
 			// Set Flush func for use by exported Writer.Flush.
 			setFlush(func() (int64, error) {
 				return gw.flush()
 			})
 
+			// Unless the user told us the content type, we have to determine it from
+			// the first read.
+			var r io.Reader = pr
```
BrennaEpp, line 1737: Can we remove `r` and just use `gw.reader`?

storage/grpc_client.go
```diff
@@ -1722,24 +1722,23 @@ func (c *grpcStorageClient) OpenWriter(params *openWriterParams, opts ...storage
...
+			if params.attrs.ContentType == "" && !params.forceEmptyContentType {
+				gw.reader, params.attrs.ContentType = gax.DetermineContentType(r)
```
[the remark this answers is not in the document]
tritone: Whoops, I fixed this locally and forgot to add it to the PR. Added now.

storage/grpc_client.go
```diff
@@ -1722,24 +1722,23 @@ func (c *grpcStorageClient) OpenWriter(params *openWriterParams, opts ...storage
...
-			gw, err := newGRPCWriter(c, s, params, r, pw, params.setPipeWriter)
+			gw, err := newGRPCWriter(c, s, params, pr, pw, params.setPipeWriter)
 			if err != nil {
 				return err
 			}
-
 			// Set Flush func for use by exported Writer.Flush.
 			setFlush(func() (int64, error) {
 				return gw.flush()
 			})
```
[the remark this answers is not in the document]
tritone: Done

storage/grpc_client.go
tritone: Done.

storage/grpc_client.go
```diff
@@ -1718,25 +1718,26 @@ func (c *grpcStorageClient) OpenWriter(params *openWriterParams, opts ...storage
...
+	})
+
 	// This function reads the data sent to the pipe and sends sets of messages
 	// on the gRPC client-stream as the buffer is filled.
 	go func() {
 		err := func() error {
-			var gw *gRPCWriter
-			gw, err := newGRPCWriter(c, s, params, pr, pw, params.setPipeWriter)
+			var err error
+			gw, err = newGRPCWriter(c, s, params, pr, pw, params.setPipeWriter)
```
BrennaEpp, line 1732: I think we should move this out of the go func as well

storage/grpc_client.go
```diff
@@ -1718,25 +1718,26 @@ func (c *grpcStorageClient) OpenWriter(params *openWriterParams, opts ...storage
...
-			setFlush(func() (int64, error) {
-				return gw.flush()
-			})
 
 			// Unless the user told us the content type, we have to determine it from
 			// the first read.
-			var r io.Reader = pr
 			if params.attrs.ContentType == "" && !params.forceEmptyContentType {
-				gw.reader, params.attrs.ContentType = gax.DetermineContentType(r)
+				gw.reader, gw.spec.Resource.ContentType = gax.DetermineContentType(r)
```
BrennaEpp, line 1740:
```suggestion
				gw.reader, gw.spec.Resource.ContentType = gax.DetermineContentType(gw.reader)
```

storage/client_test.go
```diff
@@ -1235,6 +1235,94 @@ func TestWriterFlushAtCloseEmulated(t *testing.T) {
...
+		wantLen := 3 * MiB
+		got, err := io.ReadAll(r)
+		if n := len(got); n != wantLen {
+			t.Fatalf("expected to read %d bytes, but got %d (%v)", wantLen, n, err)
+		}
+		if diff := cmp.Diff(got, randomBytes3MiB); diff != "" {
+			t.Errorf("checking written content: got(-), want(+):\n%s", diff)
+		}
+		// Expect application/octet-stream as the content type.
+		if got, want := r.Attrs.ContentType, "application/octet-stream"; got != want {
```
BrennaEpp, line 1320: As a test, can you try writing the first 10ish bytes as plain text? 
Noting this because the following sentence seems to be enough for the algorithm to determine the content type in our integration tests: "It was the best of times, it was the worst of times." 

Just noting that it is possible this may be flaky since the contents are random.

tritone: I tried this out locally and was able to get plain text to work. I rewrote the test to deterministically force application/octet-stream by rewriting the initial bytes

storage/grpc_client.go
tritone: Addressed this while keeping the error handling the same...

BrennaEpp (approved):

PR #11935, opened 2025-04-01
ci(spanner): Support tracing in benchmarking and optional command line arguments

sakthivelmanii:
1. Made read as default 
2. Made command line argument optional
3. Support tracing
4. Provide an option to disable native metrics


spanner/benchmarks/benchmarks.go
```diff
@@ -70,6 +76,25 @@ var monitoringHosts = map[cloudEnvironment]string{
...
 
+var cloudTracingHosts = map[cloudEnvironment]string{
+	production: "cloudtrace.googleapis.com:443",
+	devel:      "staging-cloudtrace.sandbox.googleapis.com:443",
+}
+
+type benchmarkingConfiguration struct {
+	warmUpTime            int8            // in minutes, default 7 minutes
+	executionTime         int8            // in minutes, default 30 minutes
+	waitBetweenRequests   int8            // in ms, 		 default 5 ms
```
olavloite, line 87: nit:
```suggestion
	waitBetweenRequests   int8            // in ms, default 5 ms
```

spanner/benchmarks/benchmarks.go
```diff
@@ -229,6 +248,32 @@ func executeRead(ctx context.Context, client *spanner.Client, staleness int64) (
...
+	// Create a new cloud trace exporter
+	exporter, err := traceExporter.New(traceExporter.WithProjectID(projectID), traceExporter.WithTraceClientOptions([]option.ClientOption{option.WithEndpoint(cloudTracingHost)}))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	tracerProvider := sdktrace.NewTracerProvider(
+		sdktrace.WithResource(res),
+		sdktrace.WithBatcher(exporter),
+		sdktrace.WithSampler(sdktrace.TraceIDRatioBased(0.1)),
```
olavloite, line 271: 10% is a relatively high trace ratio. You might want to make this configurable and/or reduce the default value

spanner/benchmarks/benchmarks.go
```diff
@@ -277,6 +386,6 @@ func generateUniqueID() int64 {
 	return rand.Int64N(totalRecords) + 1
 }
 
-func getRandomWaitTime(waitTime int64) time.Duration {
-	return time.Duration(rand.Int64N(2*waitTime-1) + 1)
+func getRandomWaitTime(waitTime int8) time.Duration {
+	return time.Duration(rand.Int64N(int64(2*waitTime-1)) + 1)
```
olavloite, line 390: (This bug was already there, but maybe good to fix): If `waitTime` is zero, then I think this fails.

olavloite (approved):
LGTM, but as there are no tests, I have to assume that you tested it manually :-)

spanner/benchmarks/benchmarks.go
sakthivelmanii: done

go.work
```diff
@@ -157,6 +157,7 @@ use (
 	./shell
 	./shopping
 	./spanner
+	./spanner/benchmarks
```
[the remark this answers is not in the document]
sakthivelmanii: yes. all the modules needs to be added into go workspace file.

spanner/benchmarks/benchmarks.go
sakthivelmanii: Fixed it.

PR #11955, opened 2025-04-04
feat(bigtable): Use delay from RetryInfo

bhshkh:
b/368356092

1) Used information from RetryInfo, if available, to retry requests. When server stop sending RetryInfo, use initial backoff. This is similar to grpc design 

https://github.com/grpc/proposal/blob/90462b53883a0455157f9a0fd0a76b36fa567acc/A6-client-retries.md?plain=1#L209

2) User can disable the usage of RetryInfo using DISABLE_RETRY_INFO environment variable


bigtable/bigtable.go
```diff
@@ -87,6 +90,11 @@ type ClientConfig struct {
 	//
 	// TODO: support user provided meter provider
 	MetricsProvider MetricsProvider
+
+	// If true, library uses default retry options.
+	// If false, library bases retry decision and back off time on
+	// server returned RetryInfo value.
+	DisableRetryInfo bool
```
igorbernstein2, line 97: drive-by comment - I dont think we want to commit to the ability to disable DisableRetryInfo as part of the public api forever. Maybe use the same approach as auth and have a hidden env var to disable it in case of emergency.

bigtable/admin.go
```diff
@@ -217,7 +217,7 @@ func (ac *AdminClient) Tables(ctx context.Context) ([]string, error) {
 		var err error
 		res, err = ac.tClient.ListTables(ctx, req)
 		return err
-	}, retryOptions...)
+	}, noRetryInfoRetryOptions...)
```
igorbernstein2, line 220: nit - to minimize the noise, consider using regular retryOptions for admin ops and then adding a dataRetryOptions that uses RetryInfo...this way all of the diffs point to the methods that are getting affected and the methods that dont get affected dont appear in the diff

bigtable/bigtable.go
```diff
@@ -176,44 +195,90 @@ func (c *Client) Close() error {
...
+func retryInfoRetry(backoff *gax.Backoff, err error) (time.Duration, bool) {
+	fmt.Println("Entering retryInfoRetry")
+	defer fmt.Println("Exiting retryInfoRetry")
+	apiErr, ok := apierror.FromError(err)
+	if ok && apiErr != nil && apiErr.Details().RetryInfo != nil {
+		// If there is RetryInfo, retry the error irrespective of the error code
+
+		// get pause duration from backoff and discard it.
+		// When server stops sending retry info, the backoff won't start from Initial duration
+		fmt.Printf("Retrying.... %v\n", apiErr.Details().RetryInfo.GetRetryDelay().AsDuration())
```
igorbernstein2, line 233: before committing please make sure to change this & all of the other Printfs to a log

bigtable/bigtable_test.go
```diff
@@ -1723,7 +1723,7 @@ func TestExecuteQuery(t *testing.T) {
 	}
 	defer client.Close()
 
-	stPcf, _ := status.New(codes.InvalidArgument, "invalid argument").WithDetails(&errdetails.PreconditionFailure{
+	stPcf, _ := status.New(codes.FailedPrecondition, "invalid argument").WithDetails(&errdetails.PreconditionFailure{
```
igorbernstein2, line 1726: 🧐

igorbernstein2 (commented):
just a few drive by comments

bigtable/bigtable_test.go
bhshkh: Moved this change to separate pull request: https://github.com/googleapis/google-cloud-go/pull/11984

bigtable/bigtable.go
bhshkh: All the Prints were for developer debugging. Removed them now

bigtable/admin.go

bigtable/bigtable.go
bhshkh: Done

bigtable/bigtable.go
```diff
@@ -176,35 +193,60 @@ func (c *Client) Close() error {
...
-	retryOptions = []gax.CallOption{
-		gax.WithRetry(func() gax.Retryer {
-			return &bigtableRetryer{
-				Backoff: defaultBackoff,
-			}
-		}),
+	clientOnlyRetryOption             = newRetryOption(clientOnlyRetry, true /* disableRetryInfo */)
+	clientOnlyExecuteQueryRetryOption = newRetryOption(clientOnlyExecuteQueryRetry, true /* disableRetryInfo */)
+	defaultRetryOption                = newRetryOption(clientOnlyRetry, false /* disableRetryInfo */)
+	defaultExecuteQueryRetryOption    = newRetryOption(clientOnlyExecuteQueryRetry, false /* disableRetryInfo */)
```
gkevinzheng, line 209: It looks like `disableRetryInfo` is false here. Does this mean these two comments are incorrect?

bigtable/bigtable.go
```diff
@@ -157,13 +160,27 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
 		return nil, err
 	}
 
+	disableRetryInfo := false
+
+	// If DISABLE_RETRY_INFO=1, library does not base retry decision and back off time on server returned RetryInfo value.
+	disableRetryInfoEnv := os.Getenv("DISABLE_RETRY_INFO")
```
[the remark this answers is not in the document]
bhshkh: This is a new environment variable we are introducing to allow the user to go back to using client only retries just in case they run into any issues with the new default retry logic of using server retry infos.
We don't _want_ the user to disable the default retries but still have a way to do it in case of escalations.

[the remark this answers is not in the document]
gkevinzheng: Is there a way to document this somewhere?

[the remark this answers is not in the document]
bhshkh: We don't want to publicly document it. Will add it to internal bug.

gkevinzheng (approved):

PR #11963, opened 2025-04-08
feat(pubsub): rename lease settings, change byte limit default

hongalex:
Contributes to #11368

Some presubmit checks will fail and this is expected. This should be resolved when making the final merge from v2-trunk to main.

### DIFF
--- a/pubsub/v2/subscription.go
+++ b/pubsub/v2/subscription.go
@@ -102,24 +102,24 @@ type ReceiveSettings struct {
 	// duration less than 0.
 	MaxExtension time.Duration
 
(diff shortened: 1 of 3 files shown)


shollyman (approved):
Is there a precedent already for using MaxInt to represent unbounded in the library?  That's the only thing that looks a bit odd.

hongalex:
> Is there a precedent already for using MaxInt to represent unbounded in the library? That's the only thing that looks a bit odd.

It was previously using MaxInt defined [manually](https://github.com/googleapis/google-cloud-go/blob/74559fc6a1e94004c880055a8b73d11fc7bc38ae/pubsub/topic.go#L64). I removed this but decided to add it back:

PR #11964, opened 2025-04-09
chore(storage): cleanup MRD code

BrennaEpp:
Some initial cleanup of MRD code.

- Rename some variables.
- Rephrase errors and use errors.New
- Move some code around (an attempt to organize the grpc struct a tiny bit and move methods underneath their struct).
- Make activeRange function a method instead.

Follows:
- Move grpcBidiReader to a new file.
- Add a readIDgenerator to avoid the need for the mutex on it and make it clearer the purpose of the variable in the struct. 

Eventually it would be good to move most if not all vars in the struct to use their own mutexes, or likely, structures that don't need explicit coordination when accessed.

There are possibly more name changes to be done - this is not a full audit of the code but an attempt to make the code easier to read.


storage/grpc_reader_bidi.go
```diff
@@ -0,0 +1,531 @@
+// Copyright 2025 Google LLC
```
tritone, line 1: Could you skip the move to a new file for this PR? It makes it hard to understand the diff.

Also I was thinking we could just centralize all download logic in grpc_reader.go.

BrennaEpp: I'll move it back for this PR. 

I think we should keep more separation - MRD seems to be a unit of code that doesn't interact with the other read code too much.

BrennaEpp: Ok I also removed the readIDGenerator piece, I will open a separate pr for that. I know it's a bunch of little changes together which makes it hard to review! Let me know if you want me to piece this out even more.

storage/grpc_client.go
```diff
@@ -1083,30 +1073,30 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
 
 	b := bucketResourceName(globalProjectAlias, params.bucket)
 	object := params.object
-	r := &storagepb.BidiReadObjectSpec{
+	bidiObject := &storagepb.BidiReadObjectSpec{
```
tritone, line 1076: Maybe call this `readSpec` or `bidiReadSpec`? Since the proto also has object as a data type it's clearer if you indicate that it's a spec (to me anyway).

storage/grpc_client.go
```diff
@@ -1381,16 +1371,43 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
...
 	}, nil
 }
 
-func getActiveRange(r *gRPCBidiReader) []rangeSpec {
-	r.mu.Lock()
-	defer r.mu.Unlock()
-	var activeRange []rangeSpec
-	for k, v := range r.mp {
-		activeRange = append(activeRange, rangeSpec{
+type gRPCBidiReader struct {
```
tritone, line 1378: IMO this could also be renamed to something related to MRD given that it's not used by the actual bidi storage.Reader implementation.

storage/grpc_client.go
```diff
@@ -1168,58 +1158,58 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
...
-				rr.mu.Lock()
-				if len(rr.mp) != 0 {
-					for key := range rr.mp {
-						rr.mp[key].callback(rr.mp[key].offset, rr.mp[key].totalBytesWritten, fmt.Errorf("stream closed early"))
-						delete(rr.mp, key)
+			case <-mrd.closeSender:
+				mrd.mu.Lock()
+				if len(mrd.activeRanges) != 0 {
+					for key := range mrd.activeRanges {
+						mrd.activeRanges[key].callback(mrd.activeRanges[key].offset, mrd.activeRanges[key].totalBytesWritten, fmt.Errorf("stream closed early"))
```
tritone, line 1197: this could also be an errors.New I guess

tritone (approved):
LGTM, just a few nits

storage/grpc_client.go
BrennaEpp: Agreed

storage/grpc_client.go
BrennaEpp: Yes, `readSpec` is good. Was avoiding colliding with the one defined in our own code when I made this change but I renamed that one.

PR #11969, opened 2025-04-10
chore(storage): use a generator for the read ids

BrennaEpp:
This makes it less likely for the readID value to cause concurrency issues and makes it clearer what it is used for.

Adding to new file. Follow up will move the rest of the MRD code to that file.


MalutkaKarteczka (dismissed):
Leave

tritone (approved):

PR #11972, opened 2025-04-11
Pool grpc writer buffers for small objects

fpetkovski:
The gRPCWriter allocates a new byte slice for each new object, which causes a lot of GC pressure for use cases where many small objects are uploaded to a bucket.

This commit adds a bytes pool so that allocations can be amortized for objects smaller than 256K, which is the minimum chunk size.


bwplotka (approved):
Thanks! 👍🏽  from my side, but ofc I don't maintain this library.

cc @tritone

tritone (commented):
I think this is a good idea, but some of the details could use some discussion
1. Using 256k chunk size leads to larger uploads being slower. Should we think about pooling for larger buffers as well?
2. Perhaps a more flexible approach would be to allow the user to pass in a buffer to the Writer? That way the end user could handle the pooling based on their workload profile.
3. We'd also need to run benchmarking against the real service to see how this does in practice.

PR #11977, opened 2025-04-14
feat(storage): takeover appendable object

tritone:
Adds a new method `NewWriterFromAppendableObject` which allows the caller to start writing to an appendable-but-unfinalized object in GCS.

Conformance/emulator tests waiting on full support in the testbench for takeover. Integration tests pass locally, but cannot run in kokoro since zonal buckets are not yet available in the region.

Also fixes a bug where Writer.Attrs is not set appropriately for unfinalized uploads in some cases.

### DIFF
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -3248,10 +3248,11 @@ func TestIntegration_WriterAppendTakeover(t *testing.T) {
 		defer h.mustDeleteBucket(bkt)
 
 		testCases := []struct {
-			name           string
-			content        []byte
-			takeoverOffset int64
-			opts           *AppendableWriterOpts
+			name                 string
+			content              []byte
+			takeoverOffset       int64
+			opts                 *AppendableWriterOpts
+			checkProgressOffsets []int64
 		}{
 			{
 				name:           "first message takeover",
@@ -3260,12 +3261,13 @@ func TestIntegration_WriterAppendTakeover(t *testing.T) {
 				opts:           nil,
 			},
 			{
-				name:           "first chunk takeover",
+				name:           "first chunk takeover, progressfunc",
 				content:        randomBytes9MiB,
 				takeoverOffset: 3 * MiB,
 				opts: &AppendableWriterOpts{
 					ChunkSize: 4 * MiB,
 				},
+				checkProgressOffsets: []int64{7 * MiB},
 			},
 			{
 				name:           "middle chunk takeover",
@@ -3276,12 +3278,11 @@ func TestIntegration_WriterAppendTakeover(t *testing.T) {
 				},
 			},
 			{
-				name:           "final chunk takeover, progressfunc",
+				name:           "final chunk takeover",
 				content:        randomBytes9MiB,
 				takeoverOffset: 8*MiB + 100,
 				opts: &AppendableWriterOpts{
-					ChunkSize:    4 * MiB,
-					ProgressFunc: func(_ int64) {},
+					ChunkSize: 4 * MiB,
 				},
 			},
 			{
@@ -3297,15 +3298,16 @@ func TestIntegration_WriterAppendTakeover(t *testing.T) {
 				name:           "finalize object, default chunkSize",
 				content:        randomBytes9MiB,
 				takeoverOffset: MiB,
+				opts:           nil,
+			},
+			{
+				name:           "0 byte takeover, progressfunc",
+				content:        randomBytes9MiB,
+				takeoverOffset: 0,
 				opts: &AppendableWriterOpts{
 					ChunkSize: 4 * MiB,
 				},
-			},
-			{
-				name:           "0 byte takeover",
-				content:        randomBytes9MiB,
-				takeoverOffset: 0,
-				opts:           nil,
+				checkProgressOffsets: []int64{4 * MiB, 8 * MiB},
 			},
 			{
 				name:           "last byte takeover",
@@ -3342,6 +3344,12 @@ func TestIntegration_WriterAppendTakeover(t *testing.T) {
 				// Takeover to create new Writer.
 				gen := w.Attrs().Generation
 				opts := tc.opts
+				gotOffsets := []int64{}
+				if tc.checkProgressOffsets != nil {
+					opts.ProgressFunc = func(n int64) {
+						gotOffsets = append(gotOffsets, n)
+					}
+				}
 				w2, off, err := obj.Generation(gen).NewWriterFromAppendableObject(ctx, opts)
 				if err != nil {
 					t.Fatalf("NewWriterFromAppendableObject: %v", err)
@@ -3383,7 +3391,12 @@ func TestIntegration_WriterAppendTakeover(t *testing.T) {
 				if (opts == nil || !opts.FinalizeOnClose) && !attrs.Finalized.IsZero() {
 					t.Errorf("got object finalized at %v, want unfinalized", attrs.Finalized)
 				}
-
+				// Check ProgressFunc was called if applicable
+				if tc.checkProgressOffsets != nil {
+					if !slices.Equal(gotOffsets, tc.checkProgressOffsets) {
+						t.Errorf("progressFunc calls: got %v, want %v", gotOffsets, tc.checkProgressOffsets)
+					}
+				}
 			})
 		}
 	})


madisonhall38 (approved):

storage/integration_test.go
```diff
@@ -3233,7 +3215,100 @@ func TestIntegration_WriterAppend(t *testing.T) {
...
+// Writer test for append takeover of unfinalized object.
+func TestIntegration_WriterAppendTakeover(t *testing.T) {
+	t.Skip("b/402283880")
+	ctx := skipAllButBidi(context.Background(), "ZB test")
+	multiTransportTest(ctx, t, func(t *testing.T, ctx context.Context, _, prefix string, client *Client) {
+		h := testHelper{t}
+		bucketName := prefix + uidSpace.New()
+		bkt := client.Bucket(bucketName)
+
+		h.mustCreateZonalBucket(bkt, testutil.ProjID())
```
BrennaEpp, line 3247: nit: I'd suggest `mustCreateZonalBucket` also handle the deletion of the bucket using `t.Cleanup()`

storage/integration_test.go
```diff
@@ -3233,7 +3215,100 @@ func TestIntegration_WriterAppend(t *testing.T) {
...
+		defer h.mustDeleteBucket(bkt)
+
+		testCases := []struct {
+			name           string
+			finalize       bool
```
BrennaEpp, line 3252: Looks like this is unused

storage/integration_test.go
```diff
@@ -3233,7 +3215,100 @@ func TestIntegration_WriterAppend(t *testing.T) {
...
+			},
+			{
+				name:           "middle chunk takeover",
+				finalize:       false,
+				content:        randomBytes9MiB,
+				chunkSize:      4 * MiB,
+				takeoverOffset: 6 * MiB,
+			},
+			{
+				name:           "final chunk takeover",
```
BrennaEpp, line 3279: Is it possible to do a zero byte write with FinalizeOnClose= true  to finalize an object? If so, maybe it would be good to test?

storage/integration_test.go
```diff
@@ -6637,6 +6712,30 @@ func (h testHelper) mustCreate(b *BucketHandle, projID string, attrs *BucketAttr
 	}
 }
 
+func (h testHelper) mustCreateZonalBucket(b *BucketHandle, projID string) {
+	h.t.Helper()
+
+	// Create a bucket in the same zone as the test VM.
+	zone, err := metadata.ZoneWithContext(context.Background())
+	if err != nil {
+		h.t.Fatalf("could not determine VM zone: %v", err)
```
BrennaEpp, line 6721: This will cause integration tests to fail locally unless on a VM, correct? Should we add a skip/flag to avoid tests that require this similar to `-short`?

storage/integration_test.go
BrennaEpp: Also would be good to test an object with takeoverOffset at zero bytes.

storage/integration_test.go
```diff
@@ -3233,7 +3215,100 @@ func TestIntegration_WriterAppend(t *testing.T) {
...
+			t.Run(tc.name, func(t *testing.T) {
+				// Create non-finalized appendable writer and upload first part of content.
+				obj := bkt.Object(tc.name + uidSpace.New()).Retryer(WithPolicy(RetryAlways))
+				defer h.mustDeleteObject(obj)
+				w := obj.If(Conditions{DoesNotExist: true}).NewWriter(ctx)
+				w.Append = true
+				w.FinalizeOnClose = false
+				w.ChunkSize = tc.chunkSize
+
+				h.mustWrite(w, tc.content[:tc.takeoverOffset])
```
BrennaEpp, line 3296: What happens if the first writer isn't closed (but is appropriately flushed to GCS through Flush or regular buffer flow), and we try to take over? I assume it will work? But if we try to use this first writer again after taking over, what error do we get?

storage/storage.go
```diff
@@ -1241,6 +1241,48 @@ func (o *ObjectHandle) NewWriter(ctx context.Context) *Writer {
...
 }
 
+// NewWriterFromAppendableObject opens a new Writer to an object which has been
+// partially flushed to GCS, but not finalized. It returns the Writer as well
+// as the current end offset of the object. All bytes written will be appended
+// continuing from the offset.
+//
+// Generation must be set on the ObjectHandle or an error will be returned.
+//
+// Writer attributes such as ChunkSize or ChunkRetryDuration can be set on the
```
BrennaEpp, line 1251: What about crc32c?

storage/integration_test.go
```diff
@@ -3233,7 +3215,100 @@ func TestIntegration_WriterAppend(t *testing.T) {
...
+				// Takeover and write remainder of content.
+				gen := h.mustObjectAttrs(obj).Generation
+				w2, off, err := obj.Generation(gen).NewWriterFromAppendableObject(ctx)
+				if err != nil {
+					t.Fatalf("NewWriterFromAppendableObject: %v", err)
+				}
+				if off != tc.takeoverOffset {
+					t.Errorf("takeover offset: got %v, want %v", off, tc.takeoverOffset)
+				}
+				w2.FinalizeOnClose = true
```
BrennaEpp, line 3307: Should we test setting params on the writer?

BrennaEpp, line 3307: Should we also test doubly-taken over objects (`w2.FinalizeOnClose = false` here and add a second takeover)? Not sure how much value that would add.

BrennaEpp (commented):
Overall LGTM, mostly some non-blocking nits.

storage/grpc_writer.go
```diff
@@ -245,6 +278,7 @@ func (s *gRPCAppendBidiWriteBufferSender) sendOnConnectedStream(buf []byte, offs
 		proto.Merge(req, s.firstMessage)
 	}
 
+	// log.Printf("sending msg: send first msg %v, offset %v, flush %v, finishwrite %v, finalize: %v", sendFirstMessage, offset, flush, finishWrite, finalizeObject)
```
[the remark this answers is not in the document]
tritone: Done

storage/integration_test.go
tritone: Added a test case using this.

storage/integration_test.go
tritone: I think this should cause an error in the first writer on Close; I can test it out.

storage/integration_test.go
tritone: Discussed offline, yes we should.

tritone: I don't think it should make a difference, this seems like more of a server functionality test?

storage/integration_test.go
tritone: added first byte and last byte cases as well as takeover just to finalize.

storage/integration_test.go
tritone: Added some tests with a bit more of this.

storage/integration_test.go
tritone: This mirrors how h.mustCreate works currently; we can refactor later if desired.

storage/storage.go
```diff
@@ -1273,16 +1274,49 @@ func (o *ObjectHandle) NewWriterFromAppendableObject(ctx context.Context) (*Writ
...
+}
+
+func (opts *AppendableWriterOpts) apply(w *Writer) {
+	if opts == nil {
+		return
+	}
+	w.ChunkRetryDeadline = opts.ChunkRetryDeadline
+	w.ProgressFunc = opts.ProgressFunc
+	w.ChunkSize = opts.ChunkSize
+	w.FinalizeOnClose = opts.FinalizeOnClose
```
BrennaEpp, line 1317: Are the defaults for these set elsewhere? I am concerned a user might set a zero value here on some (but not all) of the fields expecting the default behaviour to occur.

storage/integration_test.go
BrennaEpp: Did you test this?

BrennaEpp:
I think we are still missing tests for ProgressFunc and Chunksize?

storage/storage.go
tritone: Default for ChunkRetryDeadline is set in OpenWriter. But, I can add checks.

BrennaEpp (approved):
Thanks Chris
I think it would be good to actually check that the progress func gets called. This is good enough for now.

storage/integration_test.go
tritone: Yes, we get a precondition failed error: 

```close write object "first message takeover-20250415-2531366975931-0004" to bucket "golang-grpc-test-20250415-2531366975931-0003": rpc error: code = FailedPrecondition desc = A different writer has become the exclusive writer of this object.```

tritone:
> Thanks Chris I think it would be good to actually check that the progress func gets called. This is good enough for now.

I added this check as well.

2FaceS-bit (approved):

PR #11980, opened 2025-04-14
feat(bigtable): Allow configuring multicluster routing and isolation

bhshkh:
b/410568914

RoutingPolicy and Isolation fields in AppProfile
https://github.com/googleapis/google-cloud-go/blob/e8a253a351416269afd6ba86847be3a9da3455a1/bigtable/admin/apiv2/adminpb/instance.pb.go#L792-L807

MultiClusterRoutingUseAny and SingleClusterRouting
https://github.com/googleapis/google-cloud-go/blob/e8a253a351416269afd6ba86847be3a9da3455a1/bigtable/admin/apiv2/adminpb/instance.pb.go#L911-L927

- Earlier fields RoutingPolicy, ClusterID and AllowTransactionalWrites need to be deprecated since it did not allow specifying clusterIDs for multi-cluster routing
- Adding isolation fields


bigtable/admin.go
```diff
@@ -1848,28 +1852,118 @@ const (
 	// another cluster in the event of transient errors or delays. Choosing
 	// this option sacrifices read-your-writes consistency to improve
 	// availability.
+	// Deprecated: Use MultiClusterRoutingUseAnyConfig instead.
```
andre-sampaio, line 1855: NIT: Move the deprecation notice as first item in the comment

bigtable/admin.go
```diff
@@ -1848,28 +1852,118 @@ const (
...
+// ProfileConf contains the information necessary to create a profile
 type ProfileConf struct {
-	Name                     string
-	ProfileID                string
-	InstanceID               string
-	Etag                     string
-	Description              string
-	RoutingPolicy            string
-	ClusterID                string
+	Name        string
```
andre-sampaio, line 1866: NIT: formatting (the rest of the file uses all the types starting on the same column)

bigtable/admin.go
```diff
@@ -1848,28 +1852,118 @@ const (
...
 	// SingleClusterRouting is a policy that unconditionally routes all
 	// read/write requests to a specific cluster. This option preserves
 	// read-your-writes consistency, but does not improve availability.
+	// Deprecated: Use SingleClusterRoutingConfig instead.
 	SingleClusterRouting = "single_cluster_routing"
 )
 
-// ProfileConf contains the information necessary to create an profile
+// ProfileConf contains the information necessary to create a profile
 type ProfileConf struct {
```
andre-sampaio, line 1865: Isn't this a breaking change? (adding new members to a struct we expose + changing order of members)

bigtable/admin.go
```diff
@@ -1848,28 +1852,118 @@ const (
...
+	Name        string
+	ProfileID   string
+	InstanceID  string
+	Etag        string
+	Description string
+
+	RoutingConfig RoutingPolicyConfig
+	Isolation     AppProfileIsolation
+
+	// Deprecated: Use RoutingConfig instead.
```
andre-sampaio, line 1875: NIT: add that it will be overwritten by `RoutingConfig` if that's set (same for other members)

bigtable/admin.go
bhshkh: `the rest of the file uses all the types starting on the same column`
Not necessarily. Please see ClusterConfig for refrerence.

I can make few of them start on the same column by removing line breaks but I feel that comproises readability.
Screenshot 2025-04-14 at 1 03 26 PM

The formatting is done by gofmt tool. Any line breaks or comments puts the fields into separate group and each group has the same formatting

bigtable/admin.go
bhshkh: Not a breaking change

bigtable/admin.go
bhshkh: moved

bigtable/admin.go
andre-sampaio: ahhhh ok! thanks for the info

andre-sampaio (approved):

hongalex (approved):

PR #11989, opened 2025-04-15
fix(storage): fix Attrs for append takeover

tritone:
Writer.Attrs were not being piped through for takeovers. This fixes the issue. Also updates Attrs.Size to reflect the actual persisted size when the Writer is closed.

Also improve the note about thread safety of appends.


storage/grpc_writer.go
```diff
@@ -291,6 +293,9 @@ func (s *gRPCAppendBidiWriteBufferSender) sendOnConnectedStream(buf []byte, offs
 			if resp.GetResource() != nil {
 				obj = resp.GetResource()
 			}
+			if s.obj != nil && resp.GetPersistedSize() > 0 {
+				s.obj.Size = resp.GetPersistedSize()
+			}
```
BrennaEpp, line 298: So s.obj is only set by the first response, except for `size`? And is only used by the constructor? I am confused by this flow a bit, maybe comments would help

tritone: I renamed and added some comments. Basically the object resource is returned 1. on the initiation of the first write. 2. on takeover. 3. on finalization. In previous write flows we were relying (1) or (3) to populate the object; but for takeover + unfinalized we need to populate it on takeover and also update the size at the end to reflect the actual persisted size.

BrennaEpp (approved):

PR #12009, opened 2025-04-17
feat(spanner): add interval type support

rahul2393:


spanner/value.go
```diff
@@ -5065,3 +5214,340 @@ func trimDoubleQuotes(payload []byte) ([]byte, error) {
...
+	n.Valid = true
+	return nil
+}
+
+// ParseInterval parses an ISO8601 duration format string into an Interval.
+func ParseInterval(s string) (Interval, error) {
+	var interval Interval
+	interval.Nanos = new(big.Int)
+
+	pattern := `^P(-?\d+Y)?(-?\d+M)?(-?\d+D)?(T(-?\d+H)?(-?\d+M)?(-?(\d+(\.\d{1,9})?|\.\d{1,9})S)?)?$`
```
sagarwaal, line 5380: decimal can be represented through both `,` and '.'

spanner/integration_test.go
```diff
@@ -858,6 +880,195 @@ func TestIntegration_SingleUse_WithQueryOptions(t *testing.T) {
...
+
+	for i := range intervals {
+		if intervals[i].Valid != wantIntervals[i].Valid || intervals[i].Interval.Months != wantIntervals[i].Interval.Months ||
+			intervals[i].Interval.Days != wantIntervals[i].Interval.Days ||
+			(intervals[i].Interval.Nanos != nil && wantIntervals[i].Interval.Nanos != nil && intervals[i].Interval.Nanos.Cmp(wantIntervals[i].Interval.Nanos) != 0) {
+			t.Errorf("interval %d: got %+v, want %+v", i, intervals[i], wantIntervals[i])
+		}
+	}
+}
+
```
sagarwaal, line 1071: 1. Can we also have tests for passing parameters of type Interval and Interval Array (both NULL and non NULL params).
2. Can we also have tests for reading NULL Interval and NULL Interval Array.

spanner/value.go
```diff
@@ -5065,3 +5214,345 @@ func trimDoubleQuotes(payload []byte) ([]byte, error) {
...
+
+	if i.Nanos != nil && i.Nanos.Sign() != 0 {
+		result.WriteString("T")
+		nanos := new(big.Int).Set(i.Nanos)
+		isNegative := nanos.Sign() < 0
+		if isNegative {
+			nanos.Neg(nanos)
+		}
+
+		nanosPerHour := new(big.Int).SetInt64(3600000000000)
```
sagarwaal, line 5276: should we use constants, same at the other places?

spanner/value.go
```diff
@@ -5065,3 +5214,345 @@ func trimDoubleQuotes(payload []byte) ([]byte, error) {
...
+
+			secondsNanos := new(big.Int).SetInt64(secondsInt * 1000000000)
+			interval.Nanos.Add(interval.Nanos, secondsNanos)
+
+			if fractionStr != "" {
+				for len(fractionStr) < 9 {
+					fractionStr += "0"
+				}
+				if len(fractionStr) > 9 {
+					fractionStr = fractionStr[:9]
```
sagarwaal, line 5469: we should never reach this, the regex parsing will fail for more than 9 digits after decimal.

sagarwaal (approved):

harshachinta (approved):

PR #12012, opened 2025-04-18
fix(storage): fix zero-byte flush

tritone:
Fix logic to ensure that a zero byte flush actually syncs data to GCS. Add integration test cases for this.


ddelgrosso1 (approved):
I am but a C++ programmer but LGTM 😆

PR #12020, opened 2025-04-21
feat(storage): add IpFilter to bucketAttrs

madisonhall38:


storage/bucket.go
```diff
@@ -517,6 +517,11 @@ type BucketAttrs struct {
 
 	// OwnerEntity contains entity information in the form "project-owner-projectId".
 	OwnerEntity string
+
+	// IpFilter specifies the network sources that are allowed to access
+	// operations on the bucket, as well as its underlying objects.
+	// Only enforced when mode is set to 'Enabled'.
+	IPFilter *raw.BucketIpFilter
```
tritone, line 524: In this library we typically wrap the raw Apiary and proto library types. You should create a wrapper type for this library and converter methods to Apiary and proto. See SoftDeletePolicy or HierarchicalNamespace as examples.

storage/bucket.go
```diff
@@ -868,6 +873,7 @@ func newBucket(b *raw.Bucket) (*BucketAttrs, error) {
 		SoftDeletePolicy:         toSoftDeletePolicyFromRaw(b.SoftDeletePolicy),
 		HierarchicalNamespace:    toHierarchicalNamespaceFromRaw(b.HierarchicalNamespace),
 		OwnerEntity:              ownerEntityFromRaw(b.Owner),
+		IPFilter:                 b.IpFilter,
```
tritone, line 876: Looks like this only works for JSON, not gRPC. You need to add it to the proto converter as well.

tritone (commented):
Please add an integration and/or emulator test for the feature as well.

storage/bucket.go
madisonhall38: Done. Proto will be a done in a follow up once cl/753186465 is in

storage/http_client.go
```diff
@@ -309,6 +309,9 @@ func (c *httpStorageClient) UpdateBucket(ctx context.Context, bucket string, uat
 	if uattrs != nil && uattrs.PredefinedDefaultObjectACL != "" {
 		req.PredefinedDefaultObjectAcl(uattrs.PredefinedDefaultObjectACL)
 	}
+	if uattrs != nil && uattrs.IPFilter != nil {
+		// req.Header().Set("X-Goog-IP-Filter", uattrs.IPFilter)
```
tritone, line 313: What is this for?

storage/bucket.go
```diff
@@ -827,6 +832,21 @@ type HierarchicalNamespace struct {
...
+// IPFilter specifies the network sources that are allowed to access
+// operations on the bucket, as well as its underlying objects.
+// Only enforced when mode is set to 'Enabled'.
+// Wrapper type for raw.BucketIpFilter.
+// See https://cloud.google.com/storage/docs/ip-filtering-overview.
+type IPFilter struct {
+	// Mode: The mode of the IP filter. Valid values are 'Enabled' and 'Disabled'.
+	Mode string
+	// PublicNetworkSource: The public network source of the bucket's IP filter.
+	PublicNetworkSource *raw.BucketIpFilterPublicNetworkSource
```
tritone, line 844: Again, these types need to be wrapped; we don't expose Apiary types via this library.

storage/bucket.go
```diff
@@ -1243,6 +1265,10 @@ type BucketAttrsToUpdate struct {
 	// Library users should use ACLHandle methods directly.
 	defaultObjectACL []ACLRule
 
+	// If set, updates the IP filtering configuration of the bucket.
+	// Only enforced when mode is set to 'Enabled'.
```
tritone, line 1269: This is a bit confusing... can you update it to disabled? It might help to be more explicit.

tritone (commented):
You should have gRPC return an error if someone tries to create or update a bucket with this field set.

storage/http_client.go
madisonhall38: Debugging. Removed

storage/bucket.go
madisonhall38: Updated the wording of the comment

storage/bucket.go
madisonhall38: Done

storage/bucket.go
```diff
@@ -827,6 +832,62 @@ type HierarchicalNamespace struct {
...
+type PublicNetworkSource struct {
+	// AllowedIpCidrRanges: The list of public IPv4, IPv6 cidr ranges that are
+	// allowed to access the bucket.
+	AllowedIPCidrRanges []string
+	// ForceSendFields is a list of field names (e.g. "AllowedIpCidrRanges") to
+	// unconditionally include in API requests. By default, fields with empty or
+	// default values are omitted from API requests. See
+	// https://pkg.go.dev/google.golang.org/api#hdr-ForceSendFields for more
+	// details.
+	ForceSendFields []string
```
tritone, line 846: This should not be exposed publicly as it is an Apiary library construct; same with NullFields (they are equivalent to fieldmask in proto/grpc). The handwritten library should decide whether to add these based on the request. See how it's used by other types e.g. here https://github.com/googleapis/google-cloud-go/blob/672bd793fc782be6e28d783f9dde678612edfd50/storage/bucket.go#L1268

storage/bucket.go
```diff
@@ -827,6 +832,62 @@ type HierarchicalNamespace struct {
...
+}
+
+// VpcNetworkSources contains the VPC network source of the
+// bucket's IP filter. Wrapper type for raw.BucketIpFilterVpcNetworkSources.
+type VpcNetworkSources struct {
+	// AllowedIpCidrRanges: The list of IPv4, IPv6 cidr ranges subnetworks that are
+	// allowed to access the bucket.
+	AllowedIPCidrRanges []string
+	// Network: Name of the network. Format:
+	// projects/{PROJECT_ID}/global/networks/{NETWORK_NAME}
```
tritone, line 861: Are these using a specific project ID or is it a global one like for GCS buckets?

storage/bucket.go
```diff
@@ -827,6 +832,62 @@ type HierarchicalNamespace struct {
...
+	Network string
+	// ForceSendFields is a list of field names (e.g. "AllowedIpCidrRanges") to
+	// unconditionally include in API requests. By default, fields with empty or
+	// default values are omitted from API requests. See
+	// https://pkg.go.dev/google.golang.org/api#hdr-ForceSendFields for more
+	// details.
+	ForceSendFields []string
```
tritone, line 868: Again, omit forceSendFields and NullFields.

storage/bucket.go
```diff
@@ -827,6 +832,62 @@ type HierarchicalNamespace struct {
...
+// Wrapper type for raw.BucketIpFilter.
+// See https://cloud.google.com/storage/docs/ip-filtering-overview.
+type IPFilter struct {
+	// Mode: The mode of the IP filter. Valid values are 'Enabled' and 'Disabled'.
+	Mode string
+	// PublicNetworkSource: The public network source of the bucket's IP filter.
+	PublicNetworkSource *PublicNetworkSource
+	// VpcNetworkSources: The list of VPC network
+	// (https://cloud.google.com/vpc/docs/vpc) sources of the bucket's IP filter.
+	VpcNetworkSources []*VpcNetworkSources
```
tritone, line 888: Having a slice of pointers is unnecessary, you can just do a slice of the type.

storage/bucket.go
```diff
@@ -827,6 +832,62 @@ type HierarchicalNamespace struct {
...
+	// NullFields is a list of field names (e.g. "AllowedIpCidrRanges") to include
+	// in API requests with the JSON null value. By default, fields with empty
+	// values are omitted from API requests. See
+	// https://pkg.go.dev/google.golang.org/api#hdr-NullFields for more details.
+	NullFields []string
+}
+
+// VpcNetworkSources contains the VPC network source of the
+// bucket's IP filter. Wrapper type for raw.BucketIpFilterVpcNetworkSources.
+type VpcNetworkSources struct {
```
tritone, line 856: With Go capitalization conventions this should be VPCNetworkSources. Or probably VPCNetworkSource as it is just one.

madisonhall38: Done

storage/bucket.go
madisonhall38: Done

storage/bucket.go
madisonhall38: Done

storage/bucket.go
madisonhall38: Done.

storage/bucket.go
```diff
@@ -827,6 +832,40 @@ type HierarchicalNamespace struct {
...
+	AllowedIPCidrRanges []string
+	// Network: Name of the network. Format:
+	// projects/{PROJECT_ID}/global/networks/{NETWORK_NAME}
+	Network string
+}
+
+// IPFilter specifies the network sources that are allowed to access
+// operations on the bucket, as well as its underlying objects.
+// Only enforced when mode is set to 'Enabled'.
+// Wrapper type for raw.BucketIpFilter.
```
tritone, line 857: This line can be removed, not necessary in public docs.

storage/bucket_test.go
```diff
@@ -252,6 +253,7 @@ func TestBucketAttrsToRawBucket(t *testing.T) {
 				},
 			},
 		},
+		IpFilter: &raw.BucketIpFilter{Mode: "disabled"},
```
tritone, line 256: Maybe this should include PublicNetworkSource/VPCNetworkSource fields to test the nested converters for these?

storage/integration_test.go
```diff
@@ -5315,6 +5315,53 @@ func TestIntegration_ObjectRetention(t *testing.T) {
...
+
+		ua := BucketAttrsToUpdate{
+			IPFilter: want,
+		}
+		// Verify updated configuration.
+		attrs = h.mustUpdateBucket(bucket, ua, attrs.MetaGeneration)
+		if !testutil.Equal(attrs.IPFilter, want) {
+			t.Errorf("got bucket IPFilter %+v, want %+v", attrs.IPFilter, want)
+		}
+	})
```
tritone, line 5362: Can you also try a case that clears VPCNetworkSource and/or PublicNetworkSource fields? That can be tricky to make sure it works correctly.

storage/bucket.go
```diff
@@ -827,6 +832,40 @@ type HierarchicalNamespace struct {
...
+// See https://cloud.google.com/storage/docs/ip-filtering-overview.
```
tritone, line 858: Can you note that it's not supported for gRPC currently?

tritone (commented):
Couple more minor nits, otherwise looks good to me.

storage/integration_test.go
madisonhall38: Done

storage/bucket.go
```diff
@@ -2229,6 +2280,176 @@ func toHierarchicalNamespaceFromRaw(r *raw.BucketHierarchicalNamespace) *Hierarc
...
+		return nil
+	}
+	return &IPFilter{
+		Mode:                r.Mode,
+		PublicNetworkSource: publicNetworkSourceFromRaw(r.PublicNetworkSource),
+		VPCNetworkSource:    vpcNetworkSourcesFromRaw(r.VpcNetworkSources),
+	}
+}
+
+// func toRawIPFilter(p *IPFilter) *raw.BucketIpFilter {
```
BrennaEpp: Do these lines need to be removed?

storage/bucket_test.go
```diff
@@ -406,6 +430,18 @@ func TestBucketAttrsToUpdateToRawBucket(t *testing.T) {
...
+			Mode: "disabled",
+			// VPCNetworkSource: []VPCNetworkSource{
+			// 	{
+			// 		Network:             fmt.Sprintf("projects/%s/global/networks/default", testutil.ProjID()),
+			// 		AllowedIPCidrRanges: []string{"0.0.0.0/0"},
+			// 	},
+			// },
+			// PublicNetworkSource: &PublicNetworkSource{
+			// 	AllowedIPCidrRanges: []string{"1.2.3.4/32"},
+			// },
```
BrennaEpp: Why are these commented out?

storage/integration_test.go
```diff
@@ -5315,6 +5315,74 @@ func TestIntegration_ObjectRetention(t *testing.T) {
...
+
+		// Now clear VPCNetworkSource and PublicNetworkSource.
+		want = &IPFilter{
+			Mode: "Disabled",
+			// VPCNetworkSource and PublicNetworkSource omitted/zeroed to clear them.
+		}
+		ua = BucketAttrsToUpdate{
+			IPFilter: want,
+		}
+		attrs = h.mustUpdateBucket(bucket, ua, attrs.MetaGeneration)
```
BrennaEpp: This is a metadata update iiuc; which means that the below lines will be flaky on the update. Can you preemptively add a retry loop around them to avoid flakiness?

BrennaEpp (commented):
Couple nits

storage/bucket_test.go
madisonhall38: oops, was debugging, added back.

storage/bucket.go
madisonhall38: yes thx

storage/integration_test.go
madisonhall38: Done

storage/integration_test.go
```diff
@@ -5366,23 +5364,32 @@ func TestIntegration_BucketIPFilter(t *testing.T) {
...
 		attrs = h.mustUpdateBucket(bucket, ua, attrs.MetaGeneration)
 		if attrs.IPFilter == nil {
 			t.Errorf("got nil IPFilter, want non-nil")
-		}
-		if len(attrs.IPFilter.VPCNetworkSource) != 0 {
-			t.Errorf("expected VPCNetworkSource to be cleared, got %+v", attrs.IPFilter.VPCNetworkSource)
-		}
-		if attrs.IPFilter.PublicNetworkSource != nil && len(attrs.IPFilter.PublicNetworkSource.AllowedIPCidrRanges) != 0 {
-			t.Errorf("expected PublicNetworkSource to be cleared, got %+v", attrs.IPFilter.PublicNetworkSource)
+		} else {
```
BrennaEpp: We can probably just fail above (`t.Fatalf`) and remove this else

storage/integration_test.go
```diff
@@ -5366,23 +5364,32 @@ func TestIntegration_BucketIPFilter(t *testing.T) {
...
+			// Retry in case of metadata propagation delay.
+			err := retry(ctx, func() error {
+				return nil
```
BrennaEpp: I think we need to get bucket attrs here (you may need to do the update and get separately without using mustUpdateBucket). Otherwise attrs will never update

storage/integration_test.go
```diff
@@ -5366,23 +5364,32 @@ func TestIntegration_BucketIPFilter(t *testing.T) {
...
+				if len(attrs.IPFilter.VPCNetworkSource) != 0 {
+					return fmt.Errorf("expected VPCNetworkSource to be cleared, got %+v", attrs.IPFilter.VPCNetworkSource)
+				}
+				if attrs.IPFilter.PublicNetworkSource != nil && len(attrs.IPFilter.PublicNetworkSource.AllowedIPCidrRanges) != 0 {
+					return fmt.Errorf("expected PublicNetworkSource to be cleared, got %+v", attrs.IPFilter.PublicNetworkSource)
+				}
+				return nil
+			})
+			if err != nil {
+				t.Error(err)
```
BrennaEpp: Instead of calling t.Error from here I think we need to return the error and check the error returned from `retry`.

storage/integration_test.go
```diff
@@ -5368,28 +5368,29 @@ func TestIntegration_BucketIPFilter(t *testing.T) {
 		want = &IPFilter{
 			Mode: "Disabled",
 		}
-		ua = BucketAttrsToUpdate{
-			IPFilter: want,
+		attrs, err := bucket.Attrs(ctx)
+		if err != nil {
+			t.Fatalf("b.Attrs(%q): %v", bucket.name, err)
```
BrennaEpp, line 5374:
```suggestion
		var attrs *BucketAttrs
```

PR #12031, opened 2025-04-22
doc(storage): update gRPC docs

tritone:
Clarify auth and other requirements for direct connectivity. Reduce metrics docs and rely on link to cloud docs instead.


storage/doc.go
```diff
@@ -349,43 +350,29 @@ To create a client which will use gRPC, use the alternate constructor:
...
-[Direct Connectivity] (enabling requests to skip some proxy steps and reducing
-response latency). A warning is emmitted if gRPC is not used within GCP to
-warn that Direct Connectivity could not be initialized. Direct Connectivity
-is not required to access the gRPC API.
+One major advantage of the gRPC API is that it can use [Direct Connectivity],
+enabling requests to skip some proxy steps and reducing responce latency.
+Requirements to use Direct Connectivity include:
+
+  - Your application must be running inside Google Cloud.
+  - Your Cloud Storage bucket must be located in the same region as your VM or
```
thomasmaclean, line 358: This is now slightly more nuanced. I've had to summarize this in a bunch of other places, so hopefully this makes sense:

- Your Cloud Storage `[bucket location]` must overlap with your VM or compute environment zone. For example, if your VM is in `us-east1a`, you bucket can be located in any of the following:
  - `us-east1`, a region containing all zones starting with `us-east1`
  - `us`, a multi-region containing all US data centers
  - `nam4`, a dual-region containing `us-east1` and `us-central1`

`[bucket location]: https://cloud.google.com/storage/docs/locations`

storage/doc.go
```diff
@@ -349,43 +350,29 @@ To create a client which will use gRPC, use the alternate constructor:
...
 	// Use client as usual.
 
-Using the gRPC API inside GCP with a bucket in the same region can allow for
-[Direct Connectivity] (enabling requests to skip some proxy steps and reducing
-response latency). A warning is emmitted if gRPC is not used within GCP to
-warn that Direct Connectivity could not be initialized. Direct Connectivity
-is not required to access the gRPC API.
+One major advantage of the gRPC API is that it can use [Direct Connectivity],
+enabling requests to skip some proxy steps and reducing responce latency.
+Requirements to use Direct Connectivity include:
```
thomasmaclean, line 355: Nit: should Direct Connectivity be in brackets? Maybe that's too many links...

thomasmaclean (changes requested):
Huge improvement! There's some new nuance to include in the point about region, but otherwise looks great.

storage/doc.go
tritone: Sure, I can add this detail -- obviously it will be better when this is in the cloud.google.com docs so I can just link to a reference.

storage/doc.go
tritone: I just linked to it the first time I mentioned it in the paragraph, I think that is sufficient...

storage/doc.go
tritone: okay, added a (lightly edited) version of this.

thomasmaclean (approved):

BenWhitehead (approved):

PR #12033, opened 2025-04-23
fix(storage): validate Bidi option for MRD

tritone:
MultiRangeDownloader should require the experimental option WithGRPCBidiReads to work.

Also makes integration tests for MultiRangeDownloader only run when this option is set.


storage/client_test.go
```diff
@@ -2702,3 +2701,14 @@ func checkEmulatorEnvironment(t *testing.T) {
...
+// setBidiReads sets the WithGRPCBidiReads option on the client for the duration
+// of the test if applicable.
+func setBidiReads(t *testing.T, client storageClient) {
+	if c, ok := client.(*grpcStorageClient); ok {
+		c.config.grpcBidiReads = true
+		t.Cleanup(func() {
+			c.config.grpcBidiReads = false
+		})
+	}
+}
```
BrennaEpp, line 2714: Nice

BrennaEpp (approved):

PR #12034, opened 2025-04-23
fix(spanner): retry INTERNAL retriable auth error

tomo241:
The INTERNAL error "Authentication backend internal server error. Please retry" is a transient error and retryable.


spanner/retry.go
```diff
@@ -65,7 +65,9 @@ func (r *spannerRetryer) Retry(err error) (time.Duration, bool) {
 		!strings.Contains(err.Error(), "HTTP/2 error code: INTERNAL_ERROR") &&
 		// See b/27794742.
 		!strings.Contains(err.Error(), "Connection closed with unknown cause") &&
-		!strings.Contains(err.Error(), "Received unexpected EOS on DATA frame from server") {
+		!strings.Contains(err.Error(), "Received unexpected EOS on DATA frame from server") &&
+		// See b/331865203.
```
olavloite, line 69: nit: please remove internal links

```suggestion
```

spanner/retry_test.go
```diff
@@ -140,3 +140,37 @@ func TestRetryerRespectsServerDelayResourceExhausted(t *testing.T) {
...
+		}
+		return nil
+	}
+
+	// Use a very short backoff for testing purposes to speed it up.
+	originalBackoff := DefaultRetryBackoff
+	DefaultRetryBackoff = gax.Backoff{
+		Initial:    1 * time.Millisecond,
+		Max:        10 * time.Millisecond,
+		Multiplier: 1.1,
```
olavloite, line 163: nit: Set the initial and max backoff to 1 nanosecond instead of 1 millisecond, and the multiplier to 1.0

This test is not parallel, and so every millisecond that this adds, is added to the total build time (and yes, for one test, that does not make much of a difference, but people often copy-paste tests, so it's preferable to have this as low as possible)

olavloite (approved):

spanner/retry_test.go
tomo241: Done. Thanks for the suggestion.

PR #12067, opened 2025-04-28
feat(spanner): add support of AFE and GRPC metrics in client-side metrics

rahul2393:
This PR adds GFE/AFE/GRPC Metrics available for Spanner Engineers.

GRPC metrics are disabled as default currently. To enable set `SPANNER_DISABLE_DIRECT_ACCESS_GRPC_BUILTIN_METRICS=false`

- grpc.lb.rls.default_target_picks
- grpc.lb.rls.target_picks
- grpc.xds_client.server_failure
- grpc.xds_client.resource_updates_invalid
- grpc.xds_client.resource_updates_valid

AFE metrics will be enabled by default if direct path is used `GOOGLE_SPANNER_ENABLE_DIRECT_ACCESS =true`


spanner/client.go
```diff
@@ -713,6 +732,19 @@ func (w *wrappedStream) RecvMsg(m any) error {
 	if mt.currOp.currAttempt != nil {
 		mt.currOp.currAttempt.setDirectPathUsed(isDirectPathUsed)
 	}
+	if header, err := w.ClientStream.Header(); err == nil {
```
olavloite, line 735: Do we receive this for every message? Or only at the start of the stream? In the latter case, this should probably move to the initialization function for the stream.

spanner/metrics.go
```diff
@@ -228,18 +277,31 @@ func newBuiltinMetricsTracerFactory(ctx context.Context, dbpath string, metricsP
...
+		if isEnableGRPCBuiltInMetrics {
+			mo := opentelemetry.MetricsOptions{
+				MeterProvider: meterProvider,
+				Metrics:       stats.NewMetrics(grpcMetricsToEnable...),
+			}
+			tracerFactory.clientOpts = []option.ClientOption{
+				option.WithGRPCDialOption(
+					opentelemetry.DialOption(opentelemetry.Options{MetricsOptions: mo})),
+				option.WithGRPCDialOption(
+					grpc.WithDefaultCallOptions(grpc.StaticMethodCallOption{})),
```
olavloite, line 302: What does this do?

spanner/metrics.go
```diff
@@ -228,18 +277,31 @@ func newBuiltinMetricsTracerFactory(ctx context.Context, dbpath string, metricsP
 		},
 		shutdown: func(ctx context.Context) {},
 	}
-
+	tracerFactory.isAFEMetricsEnabled = strings.EqualFold("false", os.Getenv("SPANNER_DISABLE_AFE_SERVER_TIMING"))
```
olavloite, line 280: Is the reason that we have named the env var `SPANNER_DISABLE_AFE_SERVER_TIMING` (instead of `ENABLE`) that it will be enabled by default later? The current requirement to enable it by setting `DISABLE=false` sounds weird, but I understand it if it later will be enabled by default.

olavloite (commented):
What is the

spanner/client.go
```diff
@@ -84,7 +89,8 @@ const (
 )
 
 var (
-	validDBPattern = regexp.MustCompile("^projects/(?P<project>[^/]+)/instances/(?P<instance>[^/]+)/databases/(?P<database>[^/]+)$")
+	validDBPattern      = regexp.MustCompile("^projects/(?P<project>[^/]+)/instances/(?P<instance>[^/]+)/databases/(?P<database>[^/]+)$")
+	serverTimingPattern = regexp.MustCompile(`([a-zA-Z0-9_-]+);\s*dur=(\d+)`)
```
HailongWen, line 93: The duration number could be float numbers. `dur=(\d+)` may not handle this well (I think Java has this problem as well).

spanner/client.go
HailongWen: This will only be sent in the initial metadata so only need to process once.

spanner/client.go
```diff
@@ -1439,3 +1471,31 @@ func logf(logger *log.Logger, format string, v ...interface{}) {
...
+	if len(serverTiming) == 0 {
+		return metrics
+	}
+
+	for _, timing := range serverTiming {
+		matches := serverTimingPattern.FindAllStringSubmatch(timing, -1)
+		for _, match := range matches {
+			if len(match) == 3 { // full match + 2 capture groups
+				metricName := match[1]
+				duration, err := strconv.ParseInt(match[2], 10, 64)
```
HailongWen, line 1492: Note: the duration could be float.

spanner/client.go
```diff
@@ -1439,3 +1471,31 @@ func logf(logger *log.Logger, format string, v ...interface{}) {
...
+	if md == nil {
+		return metrics
+	}
+
+	serverTiming := md.Get(serverTimingHeaderKey)
+	if len(serverTiming) == 0 {
+		return metrics
+	}
+
+	for _, timing := range serverTiming {
```
HailongWen, line 1487: Interesting - does golang automatically split the header value using ","?

spanner/metrics.go
```diff
@@ -189,18 +231,25 @@ type metricInfo struct {
...
 
 	// Metrics instruments
 	operationLatencies metric.Float64Histogram // Histogram for operation latencies.
 	attemptLatencies   metric.Float64Histogram // Histogram for attempt latencies.
-	operationCount     metric.Int64Counter     // Counter for the number of operations.
-	attemptCount       metric.Int64Counter     // Counter for the number of attempts.
+	gfeLatencies       metric.Float64Histogram
+	afeLatencies       metric.Float64Histogram
+	gfeErrorCount      metric.Int64Counter
+	afeErrorCount      metric.Int64Counter
```
HailongWen, line 250: Add comment to these metrics.

spanner/metrics.go
HailongWen: Yes. These new metrics will be enabled by default to make sure stats are always available when we need them.

spanner/client.go
rahul2393: Done, added unit test `TestParseServerTimingHeader`

spanner/client.go
rahul2393: Done, added unit test TestParseServerTimingHeader

spanner/metrics.go
rahul2393: Added a comment in code copying here as well 
```
Configure gRPC dial options to enable gRPC metrics collection and static method call option.
The static method call option ensures consistent method names in metrics by preventing gRPC from
automatically adding service prefixes to method names. This helps maintain consistent metric
```

spanner/client.go
rahul2393: Thanks, moved this out from RecvMessage, added the logic to set server-timing for streaming request in grcp_client.go and upon recordAttemptCompletion only we will record GFE/AFE metrics

spanner/grpc_client.go
```diff
@@ -184,7 +184,16 @@ func (g *grpcSpannerClient) ExecuteSql(ctx context.Context, req *spannerpb.Execu
 func (g *grpcSpannerClient) ExecuteStreamingSql(ctx context.Context, req *spannerpb.ExecuteSqlRequest, opts ...gax.CallOption) (spannerpb.Spanner_ExecuteStreamingSqlClient, error) {
 	// Note: This method does not add g.optsWithNextRequestID to inject x-goog-spanner-request-id
 	// as it is already manually added when creating Stream iterators for ExecuteStreamingSql.
-	return g.raw.ExecuteStreamingSql(peer.NewContext(ctx, &peer.Peer{}), req, opts...)
+	mt, ok := ctx.Value(metricsTracerKey).(*builtinMetricsTracer)
+	if !ok {
+		return g.raw.ExecuteStreamingSql(peer.NewContext(ctx, &peer.Peer{}), req, opts...)
+	}
+	client, err := g.raw.ExecuteStreamingSql(peer.NewContext(ctx, &peer.Peer{}), req, opts...)
```
olavloite, line 191: nit: could we restructure this a bit like this:

```suggestion
	client, err := g.raw.ExecuteStreamingSql(peer.NewContext(ctx, &peer.Peer{}), req, opts...)
	mt, ok := ctx.Value(metricsTracerKey).(*builtinMetricsTracer)
	if !ok {
		return client, err
	}
```

That way, we have the call to `ExecuteStreamingSql` only once, which reduces the probability that if we ever need to change it in the future, we forget to update one of them.

spanner/grpc_client.go
```diff
@@ -210,7 +219,16 @@ func (g *grpcSpannerClient) Read(ctx context.Context, req *spannerpb.ReadRequest
 func (g *grpcSpannerClient) StreamingRead(ctx context.Context, req *spannerpb.ReadRequest, opts ...gax.CallOption) (spannerpb.Spanner_StreamingReadClient, error) {
 	// Note: This method does not add g.optsWithNextRequestID, as it is already
 	// manually added when creating Stream iterators for StreamingRead.
-	return g.raw.StreamingRead(peer.NewContext(ctx, &peer.Peer{}), req, opts...)
+	mt, ok := ctx.Value(metricsTracerKey).(*builtinMetricsTracer)
+	if !ok {
+		return g.raw.StreamingRead(peer.NewContext(ctx, &peer.Peer{}), req, opts...)
+	}
+	client, err := g.raw.StreamingRead(peer.NewContext(ctx, &peer.Peer{}), req, opts...)
```
olavloite, line 226: nit: same here as above, could we restructure it a little bit?

spanner/grpc_client.go
```diff
@@ -264,5 +282,14 @@ func (g *grpcSpannerClient) PartitionRead(ctx context.Context, req *spannerpb.Pa
 }
 
 func (g *grpcSpannerClient) BatchWrite(ctx context.Context, req *spannerpb.BatchWriteRequest, opts ...gax.CallOption) (spannerpb.Spanner_BatchWriteClient, error) {
-	return g.raw.BatchWrite(peer.NewContext(ctx, &peer.Peer{}), req, g.optsWithNextRequestID(opts)...)
+	mt, ok := ctx.Value(metricsTracerKey).(*builtinMetricsTracer)
+	if !ok {
+		return g.raw.BatchWrite(peer.NewContext(ctx, &peer.Peer{}), req, g.optsWithNextRequestID(opts)...)
+	}
+	client, err := g.raw.BatchWrite(peer.NewContext(ctx, &peer.Peer{}), req, g.optsWithNextRequestID(opts)...)
```
olavloite, line 289: here also, maybe have only one call to BatchWrite

olavloite (approved):
Is there any way that we could add some more tests for this? E.g. would it be possible to have the mock server return these headers, and then verify that these are correctly being collected by the client?

spanner/client.go
```diff
@@ -84,7 +89,8 @@ const (
 )
 
 var (
-	validDBPattern = regexp.MustCompile("^projects/(?P<project>[^/]+)/instances/(?P<instance>[^/]+)/databases/(?P<database>[^/]+)$")
+	validDBPattern      = regexp.MustCompile("^projects/(?P<project>[^/]+)/instances/(?P<instance>[^/]+)/databases/(?P<database>[^/]+)$")
+	serverTimingPattern = regexp.MustCompile(`([a-zA-Z0-9_-]+);\s*dur=(\d*\.?\d+)`)
```
HailongWen, line 93: nit: seems that ".5" will also match in this regex. That is probably fine.

rahul2393:
@olavloite I have updated the unit test to validate the gfe header from mock StreamingRead is captured and should be equal to set value for success response and for failed attempt gfe_error_count is exported. `TestNewBuiltinMetricsTracerFactory`

spanner/test/opentelemetry/test/ot_metrics_test.go
```diff
@@ -321,7 +321,7 @@ func TestOTMetrics_GFELatency(t *testing.T) {
 	}); err != nil {
 		t.Fatalf("could not add result: %v", err)
 	}
-	iter := client.Single().Read(context.Background(), "Users", spanner.AllKeys(), []string{"email"})
+	iter := client.Single().Query(ctx, spanner.NewStatement("SELECT email FROM Users"))
```
rahul2393, line 324: this is done because StreamingRead mock result not sets the gfe latency header

PR #12074, opened 2025-04-28
fix(storage): fix append edge cases

tritone:
* Fix the validation for generation being included. 
* Fix a case where the error was not being picked up during append takeover. 
* Fix a hang that can occur when Flush returns an error.
* Add an integration test for these issues


storage/integration_test.go
```diff
@@ -3524,6 +3524,68 @@ func TestIntegration_WriterAppendTakeover(t *testing.T) {
 	})
 }
 
+func TestIntegration_WriterAppendEdgeCases(t *testing.T) {
```
meet2mky, line 3527: Hi @tritone, Just confirming about a test scenario if it is covered in this test.

- Write is going on for an unfinalized Object in ZB using writer A
- Some different writer B deleted this object
- On next chunk write we get failed per-condition error on writer A

tritone: Writer B would have to first takeover and finalize the object before deleting; you can't delete unfinalized objects currently.

I think this is similar to the case in 3573 and following? But I can test this explicitly as well.

storage/storage.go
```diff
@@ -1267,7 +1267,7 @@ func (o *ObjectHandle) NewWriter(ctx context.Context) *Writer {
 // This feature is in preview and is not yet available for general use.
 func (o *ObjectHandle) NewWriterFromAppendableObject(ctx context.Context, opts *AppendableWriterOpts) (*Writer, int64, error) {
 	ctx = trace.StartSpan(ctx, "cloud.google.com/go/storage.Object.Writer")
-	if o.gen == 0 {
+	if o.gen < 0 {
```
BrennaEpp, line 1270: gen 0 is valid?

tritone: Yes -- object handle has -1 by default unless something is explicitly set.

storage/integration_test.go
tritone: Added this case to the test.

BrennaEpp (approved):

PR #12106, opened 2025-05-02
chore(pubsub): module and workspace updates

hongalex:


hongalex:
Merging preemptively since no code changes and this is just to make CI pass finally.

PR #12108, opened 2025-05-02
chore(pubsub): rename data plane clients

hongalex:
This is generic library-wide replacement of the following:

Topic -> Publisher
Subscription -> Subscriber

This includes both docs, the API surface (via exposed Errors) and helper methods/library internals.

This is part of #11368


shollyman (approved):
I contemplated whether aliasing the old error types would be useful, but the renames here are all consistent enough that migrating consumers should be able to work out the new exported names.

PR #12201, opened 2025-05-05
fix(storage): fix append writer hang

tritone:
After a Flush call, we weren't correctly tracking the pipe reader to close to signal to Write that an error had happened. Fixes that issue and adds another flush to the test, which triggers the hang without this fix.

### DIFF
--- a/storage/grpc_writer.go
+++ b/storage/grpc_writer.go
@@ -102,7 +102,7 @@ func (c *grpcStorageClient) OpenWriter(params *openWriterParams, opts ...storage
 	setFlush(func() (int64, error) {
 		return gw.flush()
 	})
-	gw, err := newGRPCWriter(c, s, params, pr, pw, params.setPipeWriter)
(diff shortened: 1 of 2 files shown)


storage/grpc_writer.go
```diff
@@ -102,7 +102,7 @@ func (c *grpcStorageClient) OpenWriter(params *openWriterParams, opts ...storage
 	setFlush(func() (int64, error) {
 		return gw.flush()
 	})
-	gw, err := newGRPCWriter(c, s, params, pr, pw, params.setPipeWriter)
+	gw, err := newGRPCWriter(c, s, params, pr, pr, pw, params.setPipeWriter)
```
BrennaEpp, line 105: Feels redundant to send this in twice

tritone: I agree but unfortunately gax.DetermineContentType returns a reader and therefore it's not possible to just switch to Io.PipeReader throughout (I tried it...)

BrennaEpp (approved):

PR #12206, opened 2025-05-06
fix(spanner): explicitly disable gRPC new pickfirst.

HailongWen:
As seen in b/415354418, this new experimental feature introduced by https://github.com/grpc/grpc-go/pull/8126 is not yet stable and may cause intermittent CloudPath fallbacks.

As a user of the OSS gRPC library, CS library should explicitly disable it until the underlying issue is fixed.

### DIFF
--- a/spanner/client.go
+++ b/spanner/client.go
@@ -408,6 +408,15 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
 	ctx = trace.StartSpan(ctx, "cloud.google.com/go/spanner.NewClient")
 	defer func() { trace.EndSpan(ctx, err) }()
 
+	// Explicitly disable some gRPC experiments as they are not stable yet.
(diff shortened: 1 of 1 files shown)


spanner/client.go
```diff
@@ -408,6 +408,15 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
 	ctx = trace.StartSpan(ctx, "cloud.google.com/go/spanner.NewClient")
 	defer func() { trace.EndSpan(ctx, err) }()
 
+	// Explicitly disable some gRPC experiments as they are not stable yet.
+	gRPCPickFirstEnvVarName := "GRPC_EXPERIMENTAL_ENABLE_NEW_PICK_FIRST"
```
HailongWen, line 412: Note: this env var only applies to DirectPath and will be a no-op to existing CloudPath users.

rahul2393 (approved):

PR #11607, opened 2025-02-19
feat(bigquery): Add new fields to JobStatistics

klondikedragon:
Add ReservationId, TotalSlotMs, FinalExecutionDurationMs fields to JobStatistics. These are already populated in the auto-generated struct receiving the API response, but were not being propagated to this struct.

Also update the comment for ReservationUsage based on latest API docs.

PR #12212, opened 2025-05-07
feat(bigquery): Add some missing fields to BQ stats 

ash-ddog:
Supersedes https://github.com/googleapis/google-cloud-go/pull/11607 (i.e. we can close that one out in favor of this)

This PR does a couple of things:
- reorders fields on `JobStatistics` in the same order they appear in the docs: https://cloud.google.com/bigquery/docs/reference/rest/v2/Job
- adds missing comments for fields from the docs
- maps new fields: `TotalSlotDuration, ReservationID, FinalExecutionDuration, Edition`

Please let me know if you have any questions! 🙏


alvarowolfx (approved):
LGTM

bigquery/job.go
```diff
@@ -398,14 +421,19 @@ type JobStatistics struct {
 	// a script.
 	ScriptStatistics *ScriptStatistics
 
-	// ReservationUsage attributes slot consumption to reservations.
-	ReservationUsage []*ReservationUsage
-
```
alvarowolfx, line 403: ops, better to leave this field and update the comment about deprecation, because that would be a breaking change it we remove it right away.

alvarowolfx (changes requested):
added comment on the ReservationUsage field, better to keep it and add the comment for now, as we need a deprecation notice to remove a field like that

bigquery/job.go
alvarowolfx: ah, was just reordered, sorry

alvarowolfx (approved):
nevermind, the field still there, just moved.

shollyman (approved):

PR #12217, opened 2025-05-07
feat(firestore): add pipeline queries

bhshkh:
### What are pipeline queries?

A Pipeline Query is constructed by defining a series of stages that are executed in order

**Stages**: A pipeline may consist of one or more stages. Logically, these represent the series of steps (or stages) taken to execute the query. Note: In practice, stages may be executed out of order to improve performance. However, this does not modify the intent or correctness of the query.

**Expressions**: Stages will often accept an expression allowing you to express more complex queries. Expression may be simple and consist of a single function like eq("a", 1). You can also express more complex expressions by nesting expressions like and(eq("a", 1), eq("b", 2)).

Screenshot 2025-05-07 at 2 37 14 PM

More info about pipeline queries can be found here: https://cloud.google.com/firestore/docs/pipeline/overview

### Changes in this PR
This is the initial PR for pipeline queries to have basic components in place before adding more stages and functions. 
This PR is for feature branch and not main. Once the [protos](https://github.com/googleapis/googleapis-gen/tree/preview/google/firestore/v1) are public, I will create a new PR to merge feature branch into main.

- The new structs introduced in this PR are inline with [Java](https://github.com/googleapis/java-firestore/compare/main...tomandersen/pipelineOptions#diff-ba3aea44ca47b6d53d5d1f966c85e86cffad43ed9548f372fb493fc1bdd04e70R403-R408) and [Node](https://github.com/googleapis/nodejs-firestore/pull/2094 ) implementations

- PipelineResult.Data() and PipelineResult.DataTo implementations are similar to existing [DocumentSnapshot.Data](https://github.com/googleapis/google-cloud-go/blob/2671c4fdc6e5d9c6e357222bba09b095689274d1/firestore/document.go#L60-L77) and DataTo

- streamPipelineResultIterator is similar to [queryDocumentIterator](https://github.com/googleapis/google-cloud-go/blob/2671c4fdc6e5d9c6e357222bba09b095689274d1/firestore/query.go#L1362-L1442)

image

### DIFF
--- a/firestore/pipeline.go
+++ b/firestore/pipeline.go
@@ -1,0 +1,100 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package firestore
+
+import (
+	"context"
+	"fmt"
+
+	pb "cloud.google.com/go/firestore/apiv1/firestorepb"
+)
+
+// Pipeline class provides a flexible and expressive framework for building complex data
+// transformation and query pipelines for Firestore.
+
+// A pipeline takes data sources, such as Firestore collections or collection groups, and applies
+// a series of stages that are chained together. Each stage takes the output from the previous stage
+// (or the data source) and produces an output for the next stage (or as the final output of the
+// pipeline).
+
+// Expressions can be used within
+// each stages to filter and transform data through the stage.
+
+// NOTE: The chained stages do not prescribe exactly how Firestore will execute the pipeline.
+// Instead, Firestore only guarantees that the result is the same as if the chained stages were
+// executed in order.
+type Pipeline struct {
+	c      *Client
+	stages []pipelineStage
+	err    error
+}
+
+func newPipeline(client *Client, initialStage pipelineStage) *Pipeline {
+	return &Pipeline{
+		c:      client,
+		stages: []pipelineStage{initialStage},
+	}
+}
+
+// Execute executes the pipeline and returns an iterator for streaming the results.
+// TODO: Accept PipelineOptions
+func (p *Pipeline) Execute(ctx context.Context) *PipelineResultIterator {
+	return &PipelineResultIterator{
+		iter: newStreamPipelineResultIterator(ctx, p),
+	}
+}
+
+func (p *Pipeline) toExecutePipelineRequest() (*pb.ExecutePipelineRequest, error) {
+	protoStages := make([]*pb.Pipeline_Stage, len(p.stages))
+	for i, s := range p.stages {
+		ps, err := s.toProto()
+		if err != nil {
+			return nil, fmt.Errorf("firestore: error converting stage %q to proto: %w", s.name(), err)
+		}
+		protoStages[i] = ps
+	}
+
+	req := &pb.ExecutePipelineRequest{
+		Database: p.c.path(),
+		PipelineType: &pb.ExecutePipelineRequest_StructuredPipeline{
+			StructuredPipeline: &pb.StructuredPipeline{
+				Pipeline: &pb.Pipeline{
+					Stages: protoStages,
+				},
+			},
+		},
+		// TODO: Add consistencyselector
+	}
+	return req, nil
+}
+
+// append creates a new Pipeline by adding a stage to the current one.
+func (p *Pipeline) append(s pipelineStage) *Pipeline {
+	if p.err != nil {
+		return p
+	}
+	newP := &Pipeline{
+		c:      p.c,
+		stages: make([]pipelineStage, len(p.stages)+1),
+	}
+	copy(newP.stages, p.stages)
+	newP.stages[len(p.stages)] = s
+	return newP
+}
+
+// Limit limits the maximum number of documents returned by previous stages.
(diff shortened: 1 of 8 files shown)


firestore/pipeline_result.go
```diff
@@ -0,0 +1,275 @@
...
+		return nil
+	}
+	m, err := createMapFromValueMap(p.proto.Fields, p.c)
+	// Any error here is a bug in the client.
+	if err != nil {
+		panic(fmt.Sprintf("firestore: %v", err))
+	}
+	return m
+}
+
```
daniel-sanche, line 111: is PipelineResult.Get() coming later?

firestore/pipeline_stage.go
```diff
@@ -0,0 +1,63 @@
...
+	return &inputStageCollection{path: path}
+}
+func (s *inputStageCollection) name() string { return "collection" }
+func (s *inputStageCollection) toProto() (*pb.Pipeline_Stage, error) {
+	arg := &pb.Value{ValueType: &pb.Value_ReferenceValue{ReferenceValue: s.path}}
+	return &pb.Pipeline_Stage{
+		Name: s.name(),
+		Args: []*pb.Value{arg},
+	}, nil
+}
```
daniel-sanche, line 47: Is there any way to simplify how stages are represented, to make it easier to scale?

Like maybe each the only actual struct is a base like this:

```
type PipelineStage struct {
  string
  values
  options
}
```

And each sub-type just builds one of those?

I'm not super familiar with go, so let me know if the answer is "no". Just want to make sure scalability is being considered, because we're going to have a lot more of these

firestore/pipeline_result.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	return pr, nil
+}
+
+// Exists reports whether the PipelineResult represents a document.
+// Even if Exists returns false, the ExecutionTime field of the PipelineResult
+// is valid.
+func (p *PipelineResult) Exists() bool {
+	return p.proto != nil
+}
```
daniel-sanche, line 91: Are you sure `exists` is the right word for a PipelineResult that doesn't have a document? I don't remember seeing that in the other languages

I think the document will be empty if it's an aggregation, but the aggregation result still exists.

firestore/pipeline_result.go
bhshkh: PipelineResult is similar to DocumentSnapshot. So, there won't be a Get. 

DocumentRef field inside PipelineResult and DocumentSnapshot already have a Get


The other methods that I would add are DataAt and DataAtPath
 https://github.com/googleapis/google-cloud-go/blob/2671c4fdc6e5d9c6e357222bba09b095689274d1/firestore/document.go#L120-L151

firestore/pipeline_stage.go
bhshkh: In Go, the approach we're currently using (an interface like pipelineStage with specific structs like limitStage, collectionStage implementing it) is generally the more idiomatic and preferred way to handle this kind of polymorphism while maintaining strong type safety.

While a generic PipelineStage struct could make the Pipeline.stages slice look uniform, it would shift complexity and reduce type safety:

- Type Safety: We'd lose compile-time checks on the arguments specific to each stage. For example, limitStage clearly takes an int. With a generic struct, we'd use interface{} for arguments and rely on runtime type assertions within the logic that processes these stages (like converting to protobuf), which is more error-prone.
- Clarity: Specific structs make it very clear what parameters each stage type accepts.
- Encapsulation: Each stage struct can manage its own logic for converting to its protobuf representation in its toProto() method. With a generic struct, this logic would likely end up in a large, centralized switch statement.

The interface approach scales well because adding a new stage type involves defining its specific struct, implementing the interface, and adding a corresponding builder method to Pipeline. The core pipeline processing logic that iterates over []pipelineStage and calls toProto() doesn't need to be modified for each new stage type.

This pattern is common in Go for these reasons.

firestore/pipeline_result.go
bhshkh: Removed

firestore/pipeline.go
```diff
@@ -0,0 +1,100 @@
...
+		return p
+	}
+	newP := &Pipeline{
+		c:      p.c,
+		stages: make([]pipelineStage, len(p.stages)+1),
+	}
+	copy(newP.stages, p.stages)
+	newP.stages[len(p.stages)] = s
+	return newP
+}
```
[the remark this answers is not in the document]
bhshkh: This is a deliberate choice to make the Pipeline builder immutable. It's a common pattern for fluent, chainable APIs.
1. This allows branching e.g.
```go
base := client.Pipeline().Collection("events")
pA := base.Where(Field("type").Eq("A"))
pB := base.Where(Field("type").Eq("B"))
// 'base' is still just Collection("events")
// pA and pB are distinct and don't interfere.
```
2. Thread Safety (for reading):
Immutable objects are inherently safe to share across goroutines for reading purposes without requiring locks, as their state never changes once created.

firestore/pipeline_source.go
```diff
@@ -0,0 +1,26 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package firestore
+
+// PipelineSource is a factory for creating Pipeline instances.
+// It is obtained by calling [Client.Pipeline()].
+type PipelineSource struct {
```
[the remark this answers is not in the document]
bhshkh: This is to have uniformity across clients. Java, Node and Python use the same terminology.

"Source" is domain-appropriate. While it is a factory, naming it PipelineSource effectively describes its role. Changing to PipelineFactory might make it sound more like a GoF pattern and less like a domain concept. 

Changing Client.Pipeline() to something like Client.GetPipelineFactory() or Client.NewPipelineBuilder() is more verbose.

daniel-sanche (approved):
LGTM

wu-hui (commented):
LGTM, FWIW.

PR #12220, opened 2025-05-08
chore(spanner): always enable AFE native metrics if direct path is used unless explicitly disabled

rahul2393:
- AFE and GRPC metrics will be disabled by default.
- If direct path is enabled AFE and GRPC metrics will be enabled unless explicitly disabled by user via env `SPANNER_DISABLE_AFE_SERVER_TIMING`, `SPANNER_DISABLE_DIRECT_ACCESS_GRPC_BUILTIN_METRICS`

### DIFF
--- a/spanner/client.go
+++ b/spanner/client.go
@@ -71,6 +71,10 @@ const (
 	// has opted-in for the creation of trace spans on the Spanner layer.
 	endToEndTracingHeader = "x-goog-spanner-end-to-end-tracing"
 
+	// afeMetricHeader is the name of the metadata header if client
+	// has opted-in for the receiving Spanner API Frontend server timing metrics.
+	afeMetricHeader = "x-goog-spanner-enable-afe-server-timing"
+
 	// numChannels is the default value for NumChannels of client.
 	numChannels = 4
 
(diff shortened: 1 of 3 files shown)


spanner/client.go
```diff
@@ -454,8 +458,23 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
 		// Do not emit native metrics when DisableNativeMetrics is set
 		metricsProvider = noop.NewMeterProvider()
 	}
+	isAFEBuiltInMetricEnabled := strings.EqualFold("false", os.Getenv("SPANNER_DISABLE_AFE_SERVER_TIMING"))
```
sakthivelmanii, line 461: It should be isAFEBuiltInMetricDisabled right? We are checking for false.

spanner/client.go
```diff
@@ -454,8 +458,23 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
...
+	isGRPCBuiltInMetricsEnabled := strings.EqualFold("false", os.Getenv("SPANNER_DISABLE_DIRECT_ACCESS_GRPC_BUILTIN_METRICS"))
```
sakthivelmanii, line 462: Same as above

spanner/client.go
rahul2393: No it's isAFEBuiltInMetricEnabled only by default it will be disabled means if env is unset then it's disabled
https://github.com/googleapis/java-spanner/pull/3852/files#diff-2c37e3e5595beb5b0ff59563b34415c48dc5bb5fd9c58e877f49bb14a189fc7aR679

spanner/client.go
rahul2393: https://github.com/googleapis/java-spanner/blob/23b985c9a04837b0b38f2cfc5d96469e1d664d67/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java#L925-L928

sakthivelmanii (approved):

PR #12222, opened 2025-05-08
fix(spanner): row mismatch in SelectAll using custom type

### LINKED ISSUE #11101: spanner: possible row mismatch in SelectAll using custom type
## Client

Spanner

## Environment

```
go version go1.23.2 darwin/arm64
```

## Code and Dependencies

```go
package main

func main() {
	// spannerClient created per best practices
	rowIter := spannerClient.Single().Query(
		context.Background(),
		spanner.Statement{SQL: `
		SELECT
			Id,
			OtherId,
		FROM Items
		ORDER BY OtherId DESC
		`},
	)
	defer rowIter.Stop()

	var dest []struct {
		Id        string
		OtherId Inner[string]
	}

	if err := spanner.SelectAll(rowIter, &dest); err != nil {
		t.Fatal(err)
	}
	for _, d := range dest {
		fmt.Println(d)
	}
}


type Inner[T any] struct {
	val T
}

func (n *Inner[T]) DecodeSpanner(input any) error {
	switch val := input.(type) {
	case T:
		n.val = val
		return nil
	case *T:
		if val == nil {
			return nil
		}
		n.val = *val
		return nil
	}
	panic("n/a")
}

```

<details>
  <summary>go.mod</summary>

```text
module modname

go 1.23

require (
	cloud.google.com/go/spanner v1.70.0
	google.golang.org/api v0.201.0
	google.golang.org/grpc v1.67.1
	…
)
```

</details>

## Expected behavior

Given the following data in the table `Items`:

```
Id        OtherId
715e8f8d  None
5865d2cd  foobar
```

with the schema

```
CREATE TABLE IF NOT EXISTS Items (
    Id STRING(36) DEFAULT (GENERATE_UUID()),
    OtherId STRING(MAX),
) PRIMARY KEY (Id);
```

and running the above code, I would expect 

```
{5865d2cd {foobar}}
{715e8f8d {}}
```

## Actual behavior

I see a row mismatch, where the row `715e8f8d` somehow inherits the value `foobar` from `5865d2cd`.

```
{5865d2cd {foobar}}
{715e8f8d {foobar}}
```

(guids truncated for pith)

## Additional context

Is not an issue when iterating and decoding using `ToStruct()`.

sakthivelmanii:
Fixes #11101 

**Why is it an issue?**
We are creating a new pointers for each columns only once while decoding the first row and reusing it in subsequent rows. In the decoding, if customer didn't set certain values, it's picking up values from previous column since pointers always same when DecodeSpanner is called. When we try to call DecodeSpanner, we need to make sure we have to use new memory instead of reusing it.

### DIFF
--- a/spanner/row.go
+++ b/spanner/row.go
@@ -464,7 +464,7 @@ func SelectAll(rows rowIterator, destination interface{}, options ...DecodeOptio
 	var err error
 	return rows.Do(func(row *Row) error {
 		sliceItem := reflect.New(itemType)
-		if isFirstRow && !isPrimitive {
+		if !isPrimitive {
 			defer func() {
 				isFirstRow = false
 			}()
--- a/spanner/row.go
+++ b/spanner/row.go
@@ -465,13 +465,24 @@ func SelectAll(rows rowIterator, destination interface{}, options ...DecodeOptio
 	return rows.Do(func(row *Row) error {
 		sliceItem := reflect.New(itemType)
 		if !isPrimitive {
-			defer func() {
-				isFirstRow = false
-			}()
-			if pointers, err = structPointers(sliceItem.Elem(), row.fields, s.Lenient); err != nil {
-				return err
+			if isFirstRow {
+				defer func() {
+					isFirstRow = false
+				}()
+				if pointers, err = structPointers(sliceItem.Elem(), row.fields, s.Lenient); err != nil {
+					return err
+				}
 			}
-		} else {
+			defer func() {
+				for _, ptr := range pointers {
+					v := reflect.ValueOf(ptr)
+					if v.IsValid() && !(v.IsNil() || v.IsZero()) {
+						e := v.Elem()
+						e.Set(reflect.Zero(e.Type()))
+					}
+				}
+			}()
+		} else if isPrimitive {
 			if len(row.fields) > 1 && !s.Lenient {
 				return errTooManyColumns()
 			}
--- a/spanner/row.go
+++ b/spanner/row.go
@@ -471,7 +471,7 @@ func SelectAll(rows rowIterator, destination interface{}, options ...DecodeOptio
 			if pointers, err = structPointers(sliceItem.Elem(), row.fields, s.Lenient); err != nil {
 				return err
 			}
-		} else if isPrimitive {
+		} else {
 			if len(row.fields) > 1 && !s.Lenient {
 				return errTooManyColumns()
 			}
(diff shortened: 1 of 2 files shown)


spanner/row.go
```diff
@@ -464,14 +464,14 @@ func SelectAll(rows rowIterator, destination interface{}, options ...DecodeOptio
 	var err error
 	return rows.Do(func(row *Row) error {
 		sliceItem := reflect.New(itemType)
-		if isFirstRow && !isPrimitive {
+		if !isPrimitive {
```
olavloite, line 467: Do I understand it correctly that we could, instead of the current change, also do the following:
1. Keep the if statements as is.
2. Add a `defer func() { ... set all pointers to nil ...}` to this function.

That would mean that we are re-using the slice of pointers, but they are reset to nil for each new row. Would that be a slightly more efficient solution?

If not, could we otherwise just move the entire declaration of `var pointers` into this function, as I dont' think we are using it outside of the function.

rahul2393: I think re-using the slice of pointers will be more optimal here and that was why we initially added this

rahul2393: Can we do benchmark test also like here
https://github.com/googleapis/google-cloud-go/pull/9206

To confirm we are not adding regression

sakthivelmanii:
@rahul2393 @olavloite I changed the approach and set the value using reflection instead of re-initializing memory.

Benchmark results

```
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
                             │ 10_runs_bench_using_select_all_main.txt │ 10_runs_bench_using_select_all.txt  │
                             │                 sec/op                  │   sec/op     vs base                │
Scan100RowsUsingSelectAll-12                               11.72µ ± 1%   14.78µ ± 0%  +26.06% (p=0.000 n=20)

                             │ 10_runs_bench_using_select_all_main.txt │ 10_runs_bench_using_select_all.txt │
                             │                  B/op                   │      B/op       vs base            │
Scan100RowsUsingSelectAll-12                              12.30Ki ± 0%     12.30Ki ± 0%  ~ (p=1.000 n=20) ¹
¹ all samples are equal

                             │ 10_runs_bench_using_select_all_main.txt │ 10_runs_bench_using_select_all.txt │
                             │                allocs/op                │   allocs/op     vs base            │
Scan100RowsUsingSelectAll-12                                222.0 ± 0%       222.0 ± 0%  ~ (p=1.000 n=20) ¹
¹ all samples are equal
```

rahul2393 (approved):

PR #12231, opened 2025-05-09
chore(internal/postprocessor): don't format pb.go files

codyoss:
Today we currently run goimports in our post processor over all files in the repo. Because of this the purposed changes from owlbot always has a large diff from what is landed in our repo. This in turn causes a lot of churn in the owlbot PRs as well as everyone gettings added as reviewers onto the PR. With this change we will no longer format the files that we don't contorl the generation of -- the raw proto gen. The files are still valid go code and the only difference is that imports won't be sorted in the style goimports likes. The upside is it fixes the other issues outlined above.

gofmt was removed as goimports is a super-set of features.

### DIFF
--- a/internal/postprocessor/execv/gocmd/gocmd.go
+++ b/internal/postprocessor/execv/gocmd/gocmd.go
@@ -105,18 +105,21 @@ func Build(dir string) error {
 	return nil
 }
 
-// Vet runs linters on all .go files recursively from the given directory.
-func Vet(dir string) error {
-	log.Println("vetting generated code")
+// Format runs goimports on all .go files recursively from the given directory.
+func Format(dir string) error {
+	log.Println("formatting generated code")
 	c := execv.Command("goimports", "-w", ".")
 	c.Dir = dir
 	if err := c.Run(); err != nil {
 		return err
 	}
-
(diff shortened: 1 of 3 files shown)


internal/postprocessor/execv/gocmd/gocmd.go
```diff
@@ -105,18 +105,21 @@ func Build(dir string) error {
...
+// Format runs goimports on all .go files recursively from the given directory.
+func Format(dir string) error {
+	log.Println("formatting generated code")
 	c := execv.Command("goimports", "-w", ".")
 	c.Dir = dir
 	if err := c.Run(); err != nil {
 		return err
 	}
-
-	c = execv.Command("gofmt", "-s", "-d", "-w", "-l", ".")
```
quartzmo, line 117: Can you add a bullet point in the commit explaining why gofmt is no longer desired, just for future reference?

quartzmo (approved):

internal/postprocessor/execv/gocmd/gocmd.go
codyoss: added a note to the commit, good idea!

Issue #12221, opened 2025-05-08
spanner: "iter.Do" always sets ExecuteStreamingSql span Status to "error"


AlexB-mgtc:
## Client

Spanner

## Environment

Alpine Docker on GKE

## Code and Dependencies

```go
if err := iter.Do(func(r *spanner.Row) error {
	var version domain.GameVersion
	if err := r.ToStructLenient(&version); err != nil {
		return err
	}
	// Do something with it
	return nil
}); err != nil {
	return nil, err
}
```

<details>
  <summary>go.mod</summary>

```text
go 1.24.0

require (
	cloud.google.com/go/spanner v1.80.0
        go.opentelemetry.io/otel v1.35.0
	go.opentelemetry.io/otel/trace v1.35.0
)
```

</details>

## Expected behavior
Spanner spans are all with `Status: unset`.

## Actual behavior
The span called `google.spanner.v1.Spanner/ExecuteStreamingSql` has:
```
    Status: error
    Status Message: context canceled
    Library Name: go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc
    Library Version: 0.60.0
```

Note that NO error is returned from the `iter.Do` func call and the context is not canceled while executing it.

## Screenshots

![Image](https://github.com/user-attachments/assets/603f4ebd-273b-4609-bc2c-68aeafe61d01)

PR #12235, opened 2025-05-12
fix(spanner): fix invalid trace in case of skipping trailers

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/12221

Internal Reference: go/golang-skip-trailer


spanner/stream_cleanup.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"io"
+	"sync"
+)
+
+// unfinishedStream represents a stream that needs cleanup
+type unfinishedStream struct {
+	stream streamingReceiver
+	mt     *builtinMetricsTracer
```
sakthivelmanii, line 25: You can remove this

spanner/stream_cleanup.go
```diff
@@ -0,0 +1,75 @@
...
+	finisherChannel chan unfinishedStream
+	wg              sync.WaitGroup
+}
+
+// newStreamCleanupWorker creates a new stream cleanup worker
+func newStreamCleanupWorker() *streamCleanupWorker {
+	w := &streamCleanupWorker{
+		finisherChannel: make(chan unfinishedStream, 100), // Buffer size of 100
+	}
+	w.start()
```
sakthivelmanii, line 38: Can we read some environment variable and create more worker goroutine? Can we have atleast 10 workers? goroutines are usually light weight. 10 goroutine should per client should be okay to configure.

spanner/stream_cleanup.go
```diff
@@ -0,0 +1,75 @@
...
+// streamCleanupWorker handles cleanup of unfinished streams
+type streamCleanupWorker struct {
+	finisherChannel chan unfinishedStream
+	wg              sync.WaitGroup
+}
+
+// newStreamCleanupWorker creates a new stream cleanup worker
+func newStreamCleanupWorker() *streamCleanupWorker {
+	w := &streamCleanupWorker{
+		finisherChannel: make(chan unfinishedStream, 100), // Buffer size of 100
```
sakthivelmanii, line 36: What happens if size is more than 100?

spanner/stream_cleanup.go
sakthivelmanii: Two things.

1. atleast 10 default go routines
2. make it configurable.

spanner/stream_cleanup.go
```diff
@@ -0,0 +1,73 @@
...
+// scheduleCleanup schedules a stream for cleanup
+func (w *streamCleanupWorker) scheduleCleanup(s streamingReceiver, mt *builtinMetricsTracer) {
+	w.finisherChannel <- unfinishedStream{
+		stream: s,
+	}
+}
+
+// stop stops the worker and waits for all pending cleanups to complete
+func (w *streamCleanupWorker) stop() {
+	close(w.finisherChannel)
```
sakthivelmanii, line 68: Does closing the channel make sure it process all the pending streams in the channel?

spanner/stream_cleanup.go
```diff
@@ -0,0 +1,73 @@
...
+// start starts the worker goroutine
+func (w *streamCleanupWorker) start() {
+	w.wg.Add(1)
+	go func() {
+		defer w.wg.Done()
+		for stream := range w.finisherChannel {
+			// Process remaining messages until EOF
+			for {
+				_, err := stream.stream.Recv()
+				if err != io.EOF {
```
sakthivelmanii, line 51: There's no break here. it will run infinitely. 

```
if err == io.EOF {
  break
}
```

should work

spanner/stream_cleanup.go
```diff
@@ -0,0 +1,73 @@
...
+					continue
+				}
+			}
+		}
+	}()
+}
+
+// scheduleCleanup schedules a stream for cleanup
+func (w *streamCleanupWorker) scheduleCleanup(s streamingReceiver, mt *builtinMetricsTracer) {
```
sakthivelmanii, line 60: Remove `builtinMetricsTracer` here

spanner/stream_cleanup.go
rahul2393: Yes, we are using waitGroup to make sure it process all

spanner/stream_cleanup.go
rahul2393: It will be queued

spanner/stream_cleanup.go
```diff
@@ -0,0 +1,72 @@
...
+		go func() {
+			defer w.wg.Done()
+			for s := range w.finisherChannel {
+				// Drain until EOF
+				for {
+					_, err := s.Recv()
+					if err == io.EOF {
+						break
+					}
+					if err != nil {
```
sakthivelmanii, line 48: Even without this condition, it continues right?

spanner/stream_cleanup.go
```diff
@@ -0,0 +1,72 @@
...
+					// otherwise drop the message and keep draining
+				}
+			}
+		}()
+	}
+	return w
+}
+
+// globalStreamCleanupWorker is a singleton using NumCPU() workers.
+var globalStreamCleanupWorker = newStreamCleanupWorker(runtime.NumCPU())
```
sakthivelmanii, line 61: I don't think we should have this. Can we use a standard number and environment variable so that it can be configured if needed.

rahul2393: This is the best practice to have this instead of looking for some standard number we don't know what environment this code will be running in,  @codyoss to help here.

spanner/stream_cleanup.go
rahul2393: Right, I wanted to log some error here, will do later in PR once I get review on other aspects

spanner/client.go
```diff
@@ -819,6 +819,7 @@ func getQueryOptions(opts QueryOptions) QueryOptions {
 
 // Close closes the client.
 func (c *Client) Close() {
+	stopCleanup()
```
sakthivelmanii, line 822: `stopStreamCleanup` would make more sense

spanner/read.go
```diff
@@ -672,6 +672,8 @@ func (d *resumableStreamDecoder) tryRecv(mt *builtinMetricsTracer, retryer gax.R
 	if d.err == nil {
 		d.q.push(res)
 		if res.GetLast() {
+			// Schedule cleanup of remaining stream data
+			//scheduleCleanup(d.stream)
```
sakthivelmanii, line 676: why commented here?

spanner/stream_cleanup.go
```diff
@@ -0,0 +1,75 @@
...
+var stopCleanupOnce sync.Once
+
+// scheduleCleanup enqueues a stream to be drained by the pool.
+func scheduleCleanup(s streamingReceiver) {
```
sakthivelmanii, line 65: nit: similary scheduleStreamCleanup be more readable and understandable.

spanner/read.go
rahul2393: I am waiting for the Kokoro to fail

spanner/read.go
```diff
@@ -646,7 +647,10 @@ func (d *resumableStreamDecoder) next(mt *builtinMetricsTracer) bool {
 			// to caller.
 			d.q.clear()
 			return false
-		case finished:
+		case finished, ended:
+			if d.state == ended {
```
sakthivelmanii, line 651: when will we reach this condition?

spanner/read.go
```diff
@@ -559,7 +560,7 @@ var (
 	maxBytesBetweenResumeTokens = int32(128 * 1024 * 1024)
 )
 
-func (d *resumableStreamDecoder) next(mt *builtinMetricsTracer) bool {
+func (d *resumableStreamDecoder) next(mt *builtinMetricsTracer, cancel func()) bool {
```
sakthivelmanii, line 563: are you using cancel function?

rahul2393: Yes I am setting it to nil when trailers is skipped

spanner/read.go
rahul2393: Screenshot 2025-05-13 at 12 20 11 PM

When we know that result is last

spanner/read.go
```diff
@@ -384,6 +384,7 @@ const (
 	queueingUnretryable                                    // 2
 	aborted                                                // 3
 	finished                                               // 4
+	ended                                                  // 5
```
olavloite, line 387: I don't understand why we would need a separate status for this, and can't just use finished.

spanner/read.go
```diff
@@ -646,7 +647,10 @@ func (d *resumableStreamDecoder) next(mt *builtinMetricsTracer) bool {
 			// to caller.
 			d.q.clear()
 			return false
-		case finished:
+		case finished, ended:
+			if d.state == ended {
+				cancel = nil
```
olavloite, line 652: Setting cancel to nil here does not have any effect. This only sets the local variable to nil. It does not stop `iterator.Stop` from calling `cancel()`, as that reference is not set to nil by this.

I just check this locally, and the current implementation has a random race condition:
1. `iterator.Stop` will always call `cancel()`, because the reference on the iterator is not set to nil.
2. The call to `cancel()` sometimes happen before the goroutine that fetches the trailers has executed (marking the stream as cancelled).
3. The call to `cancel()` can however also be executed after the goroutine that fetches the trailers has executed. Then the stream is marked as ended successfully.

spanner/read.go
```diff
@@ -672,7 +676,19 @@ func (d *resumableStreamDecoder) tryRecv(mt *builtinMetricsTracer, retryer gax.R
...
+				for {
+					_, err := s.Recv()
+					if err == io.EOF {
+						break
+					}
+					if err != nil {
+						break
+					}
+				}
+
```
olavloite, line 690: I would suggest that we simplify this to only `_, _ = s.Recv()`. The reason is that:
1. The contract is that we only reach this if `Last` has been set. That means that calling `s.Recv()` once should be enough to get the trailers and end the stream.
2. If the server for whatever reason does not comply with the contract and returns `Last=true` before the end of the stream, we are still stopping early. However, this would then at least cause the stream to marked as cancelled instead of ended normally.

spanner/read.go
```diff
@@ -672,7 +676,19 @@ func (d *resumableStreamDecoder) tryRecv(mt *builtinMetricsTracer, retryer gax.R
...
+			}(d.stream)
```
olavloite, line 691: We should also call the `cancel()` method here (assuming that we set the reference to `nil` before this, so that `iterator.Stop` does not call it). That ensures that the context is marked as done.

spanner/read.go
```diff
@@ -672,7 +676,19 @@ func (d *resumableStreamDecoder) tryRecv(mt *builtinMetricsTracer, retryer gax.R
 	if d.err == nil {
 		d.q.push(res)
 		if res.GetLast() {
-			d.changeState(finished)
+			d.changeState(ended)
+			go func(s streamingReceiver) {
+				for {
```
sakthivelmanii, line 681: nit: do we need a for loop here? @olavloite

olavloite (changes requested):

spanner/read.go
olavloite: No, I don't think we should add a loop here. See my comment below. The TLDR is: This should always return io.EOF on the first call, and if it doesn't, then the stream should be marked as cancelled, which would happen if we don't have a loop here.

spanner/read.go
rahul2393: I have updated the code to not set cancel nil, but checking while stopping iter does not call cancel, and is it not always the case that context is marked as done after receiving io.EOF in GRPC?

spanner/read.go
rahul2393: I added this to make sure we don't call cancel() while doing iter.Stop in case we saw the last message already

olavloite (approved):

PR #12292, opened 2025-05-16
feat(pubsub/v2): add back fake and internal packages

hongalex:
Turns out a new v2 verison of the fake (package `pstest`) is necessary after all. This is mostly because the fake takes in certain cloud.google.com/go/pubsub/apiv1/pubsubpb types as parameters which wouldn't be compatible with the pubsub/v2/apiv1/pubsubpb types. Since we plan on deprecating that first package, it's best to remove that dependency early and just release a fake v2.

Also includes the internal packages since otherwise the v2 package will take a dependency on cloud.google.com/go/pubsub package, which is breaks the proto namespace (including schema.proto from different packages).

Replaces #12291


shollyman (approved):
May be worth amending the description here to be clearer that `pstest` is the referenced fake, otherwise LGTM to re-introduce.

PR #12301, opened 2025-05-20
docs(storage): Improve error inspection documentation

### LINKED ISSUE #1865: storage: better documentation of error values and types
I don't think the client library is wrapping oauth errors into a `googleapi.Error`.

**Client**

Storage

**Environment**

ECS Service on AWS

**Code**

```go
package main

func main() {
        bucket := "..."
        path := "..."
        r := bytes.NewReader("...")

	w := s.Bucket(bucket).Object(path).NewWriter(ctx)
	w.ContentType = "application/gzip"
        = io.Copy(w, in.Body)

	if err = w.Close(); err != nil {
		var gErr *googleapi.Error
		if errors.As(err, &gErr) {
			if gErr.Code >= 400 && gErr.Code <= 499 {
				fmt.Println("client error") 
			} else {
 				panic(err)
                        }
		} else {
                        panic(err)
                }
	}
}
```

**Expected behavior**

I think ideally this should be wrapped into a `googleapi.Error`. Or at least, the other potential error types could be better documented.

https://pkg.go.dev/cloud.google.com/go/storage?tab=doc

```
Errors returned by this client are often of the type [`googleapi.Error`](https://godoc.org/google.golang.org/api/googleapi#Error). These errors can be introspected for more information by type asserting to the richer `googleapi.Error` type. For example:

if e, ok := err.(*googleapi.Error); ok {
	  if e.Code == 409 { ... }
}
```

**Actual behavior**

The error returned is not wrapped into a `googleapi.Error`.

```
oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"invalid_grant","error_description":"Invalid JWT: iss field missing."}
```

**Screenshots**

N/A

**Additional context**

N/A

frq-dev:
closes #1865


storage/doc.go
```diff
@@ -274,15 +274,43 @@ To generate the signature, you must have:
 
 # Errors
 
-Errors returned by this client are often of the type [googleapi.Error].
-These errors can be introspected for more information by using [errors.As]
-with the richer [googleapi.Error] type. For example:
+Errors returned by this client are often of the type [googleapi.Error] or
+[github.com/googleapis/gax-go/v2/apierror.APIError].
```
BrennaEpp, line 278: Since we mention `googleapi.Error` below in reference to HTTP, we can leave that out here. 
```suggestion
Errors returned by this client are often of the type [github.com/googleapis/gax-go/v2/apierror.APIError].
```

Note that this was accurate to out client before we supported gRPC so now we should mention both as errors unwrapped from the "higher" level apierror.APIError.

storage/doc.go
```diff
@@ -274,15 +274,43 @@ To generate the signature, you must have:
...
+The [apierror.APIError] type can wrap a [google.golang.org/grpc/status.Status]
+if gRPC was used, or a [google.golang.org/api/googleapi.Error] if HTTP/REST was used.
+You might also encounter `[googleapi.Error]` directly from HTTP operations.
```
BrennaEpp, line 281: Where did you see this?

storage/doc.go
```diff
@@ -274,15 +274,43 @@ To generate the signature, you must have:
...
+These two types of errors can be inspected for more information by using [errors.As]
```
BrennaEpp, line 282:
```suggestion
These types of errors can be inspected for more information by using [errors.As]
```

storage/doc.go
```diff
@@ -274,15 +274,43 @@ To generate the signature, you must have:
...
+		log.Printf("HTTP Code: %d, Message: %s", e.Code, e.Message)
+
+		if e.HTTPCode() == 409 {
+			// ... handle conflict ...
+		}
 	}
 
+	var ae *apierror.APIError
+	if ok := errors.As(err, &ae); ok {
+		// ae can wrap a googleapi.Error or a status.Status (from gRPC)
```
BrennaEpp, line 301:
```suggestion
		// APIErrors often wrap either a googleapi.Error (for JSON and XML calls) or a status.Status (for gRPC calls)
```

storage/doc.go
```diff
@@ -274,15 +274,43 @@ To generate the signature, you must have:
...
+
+		// To get the underlying gRPC status directly:
+		if s, ok := status.FromError(ae.Unwrap()); ok {
+			if s.Code() == codes.Unavailable {
+				// ... handle gRPC unavailable ...
+			}
+		}
+	}
+
+See GRPC Status Codes at https://pkg.go.dev/google.golang.org/grpc/codes.
```
BrennaEpp, line 312: Thinking about it again, I don't think we need this link; you can find this on several of the functions in the example code.

storage/doc.go
```diff
@@ -274,15 +274,43 @@ To generate the signature, you must have:
...
+		}
 	}
 
+	var ae *apierror.APIError
+	if ok := errors.As(err, &ae); ok {
+		// ae can wrap a googleapi.Error or a status.Status (from gRPC)
+		log.Printf("APIError: HTTPCode: %d, GRPCStatusCode: %s", ae.HTTPCode(), ae.GRPCStatus().Code())
+
+		// To get the underlying gRPC status directly:
+		if s, ok := status.FromError(ae.Unwrap()); ok {
```
BrennaEpp, line 305: Instead of `ae.Unwrap()`, I think this can be done directly using the top level error `err`.

storage/doc.go
```diff
@@ -274,15 +274,43 @@ To generate the signature, you must have:
...
+		// e.Code is the HTTP status code.
+		// e.Message is the error message.
+		// e.Body is the raw response body.
+		// e.Header contains the HTTP response headers.
+		log.Printf("HTTP Code: %d, Message: %s", e.Code, e.Message)
+
+		if e.HTTPCode() == 409 {
+			// ... handle conflict ...
+		}
 	}
```
BrennaEpp, line 297: Can you move this code to be underneath the other block, since this is only specific to HTTP. Also add a comment at the top of this block to indicate that is allows the user to get more details on a json/xml error.

storage/doc.go
frq-dev: In a couple places within storage tests. There was this comment in one of them 
`// errorIsStatusCode returns true if err is a:
//   - googleapi.Error with httpStatusCode, or
//   - apierror.APIError with grpcStatusCode, or
//   - grpc/status.Status error with grpcStatusCode.`

storage/doc.go
```diff
@@ -274,15 +274,44 @@ To generate the signature, you must have:
...
+		log.Printf("HTTP Code: %d, Message: %s", e.Code, e.Message)
+
+		if e.Code == 409 {
+			// ... handle conflict ...
+		}
 	}
 
+This library may also return other errors that are not wrapped as [apierror.APIError]. For
+example, errors with authentication may return [cloud.google.com/go/auth.Error]  or
+[golang.org/x/oauth2/google.AuthenticationError].
```
BrennaEpp: I think we can remove `golang.org/x/oauth2/google.AuthenticationError` since we've switched over pretty much all of auth. What do you think?

storage/doc.go
```diff
@@ -274,15 +274,44 @@ To generate the signature, you must have:
 
 # Errors
 
-Errors returned by this client are often of the type [googleapi.Error].
-These errors can be introspected for more information by using [errors.As]
-with the richer [googleapi.Error] type. For example:
+Errors returned by this client are often of the type [https://pkg.go.dev/github.com/googleapis/gax-go/v2/apierror].
```
BrennaEpp: I think here just linking the package (like `golang.org/x/oauth2/google.AuthenticationError`) should work instead of the full link.

storage/doc.go
```diff
@@ -274,15 +274,44 @@ To generate the signature, you must have:
...
+You might also encounter [googleapi.Error] directly from HTTP operations.
+These types of errors can be inspected for more information by using [errors.As]
+to access the specific underlying error types and retrieve detailed information,
+including HTTP or gRPC status codes. For example:
+
+	// APIErrors often wrap a googleapi.Error (for JSON and XML calls) or a status.Status (for gRPC calls)
+	var ae *apierror.APIError
+	if ok := errors.As(err, &ae); ok {
+		// ae.HTTPCode() is the HTTP status code.
+		// ae.GRPCStatus().Code() is the gRPC status code
```
BrennaEpp, line 289: Why are these commented out? We need them in the log below

BrennaEpp (commented):
Thank you! Left a couple nits only.

BrennaEpp (approved):

PR #12302, opened 2025-05-20
test(storage): add max attempts to common functions

BrennaEpp:
This will let us get a better error message (instead of a hang) in the case of permanent "transient" errors. 

The number of attempts is not important and was not chosen for any particular reason other than probably being "enough". We can adjust if ever needed.

### DIFF
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -7071,6 +7071,7 @@ type testHelper struct {
 
 func (h testHelper) mustCreate(b *BucketHandle, projID string, attrs *BucketAttrs) {
 	h.t.Helper()
+	b = b.Retryer(WithMaxAttempts(10))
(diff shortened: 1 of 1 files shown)


storage/integration_test.go
```diff
@@ -7111,6 +7112,7 @@ func (h testHelper) mustDeleteBucket(b *BucketHandle) {
```
tritone, line 7112: Looks like you applied the option to some of these helper methods but not others? why not do it across the board?

tritone (approved):

storage/integration_test.go
BrennaEpp: I didn't want to take the time to do a thorough review right now (I am not sure if all of these functions even retry). We should move these functions out into a testutil and consolidate some of them.

PR #12319, opened 2025-05-21
refactoring

12ya:
pubsub/pstest/fake_test.go: replace deprecated rand.Seed
pubsub/, pubsublite/, spanner/ cleanup


quartzmo (changes requested):
This PR should be closed and replaced with separate PRs for each top-level module: pubsub, pubsublite, spanner, storage, vertexai.

PR #12324, opened 2025-05-22
feat(pubsub): alias pubsub v1 stubs to v2

hongalex:
Currently there is a namespace conflict when trying to import both the v1 GAPIC Pub/Sub API and the v2:

```
panic: proto: file "google/pubsub/v1/schema.proto" is already registered
        previously from: "cloud.google.com/go/pubsub/apiv1/pubsubpb"
        currently from:  "cloud.google.com/go/pubsub/v2/apiv1/pubsubpb"
```

This uses the [aliasgen](https://github.com/googleapis/google-cloud-go/tree/main/internal/aliasgen) tool used for the go-genproto alias project to create aliases in the current v1 directory.


pubsub/apiv1/pubsubpb/alias.go
```diff
@@ -0,0 +1,801 @@
...
+	Schema_Type_value                                      = src.Schema_Type_value
+	Subscription_State_name                                = src.Subscription_State_name
+	Subscription_State_value                               = src.Subscription_State_value
+	Topic_State_name                                       = src.Topic_State_name
+	Topic_State_value                                      = src.Topic_State_value
+)
+
+// Request for the Acknowledge method.
+//
+// Deprecated: Please use types in: cloud.google.com/go/pubsub/v2/apiv1/pubsubpb
```
codyoss, line 132: Note: We wanted to strongly push users to our new libs when I wrote this tool. If you are fine with the deprecation warnings on everything that is fine, but just wanted to call out that this will affect godoc for these things. Check out a genproto package as an example.

codyoss (commented):
I think before you make this change you will want to create a beta release of the v2 module you can depend on.

pubsub/apiv1/pubsubpb/alias.go
hongalex: Yeah this is fine. We are trying to move everyone to the new GAPIC and v2 types.

hongalex:
This change will be made in #12218

PR #12372, opened 2025-05-30
chore(storage): replace deprecated io/ioutil

12ya:
refs #12370


BrennaEpp (approved):
Thanks for the cleanup!

PR #12401, opened 2025-06-04
docs(impersonate): Address TODO in impersonate/example_test.go

ivanruski:
Stumbled upon this TODO while I was exploring the `storage` package and how to write to GCS Bucket by impersonating a specific GSA. 

I am not 100% whether the author of the TODO meant what I did. Nevertheless I opened the PR.

Also I am aware that I didn't open an issue first. I wasn't sure whether it is needed for a small change like this one.

### DIFF
--- a/auth/credentials/impersonate/example_test.go
+++ b/auth/credentials/impersonate/example_test.go
@@ -33,9 +33,9 @@ func ExampleNewCredentials_serviceAccount() {
 		log.Fatal(err)
 	}
 
-	// TODO(codyoss): link to option once it exists.
-
-	// Use this Credentials with a client library
+	// Use this Credentials with a client library like
+	// cloud.google.com/go/storage and WithAuthCredentials ClientOption
+	// from the google.golang.org/api/option package
 	_ = creds
 }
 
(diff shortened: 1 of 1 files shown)


auth/credentials/impersonate/example_test.go
```diff
@@ -33,10 +36,8 @@ func ExampleNewCredentials_serviceAccount() {
 		log.Fatal(err)
 	}
 
-	// TODO(codyoss): link to option once it exists.
-
-	// Use this Credentials with a client library
-	_ = creds
+	// Use this Credentials with a client library, e.g. storage
+	_, _ = storage.NewClient(context.TODO(), option.WithAuthCredentials(creds))
```
codyoss: Yeah this todo could be resolved now. I think the important part about solving it is likely describing what API to use, like in your example, in a comment. We don't want to have this module take on those two dependencies because of an example.

auth/credentials/impersonate/example_test.go
```diff
@@ -33,9 +33,9 @@ func ExampleNewCredentials_serviceAccount() {
 		log.Fatal(err)
 	}
 
-	// TODO(codyoss): link to option once it exists.
-
-	// Use this Credentials with a client library
+	// Use this Credentials with a client library like
+	// cloud.google.com/go/storage and WithAuthCredentials ClientOption
+	// from the google.golang.org/api/option package
```
codyoss:
```suggestion
	// from the google.golang.org/api/option package.
```

PR #12404, opened 2025-06-04
chore(spanner): enable spanner gRPC client to use DirectPath bound token if possible

rockspore:
This enables spanner gRPC client to use DirectPath bound token if possible.

Same as #11674. The Cloud Spanner backend is now able to process the bound token if sent.


rockspore:
cc @ko3a4ok @HailongWen

rahul2393 (approved):

rockspore:
Thank you @rahul2393!

PR #12405, opened 2025-06-04
chore(bigquery/v2): carveout v2 module

shollyman:
This PR:

* carves out a new v2 module for bigquery
* populates the root bigquery/v2 package with a skeleton doc.go indicating there's nothing here yet
* configures release-please to treat the new bigquery/v2 as an independent releasable unit of alpha quality

Open questions:
* Do we need to update `internal/postprocessor` before we're pushing gapics into this new module?
* Same with the regexps in `.github/OwlBot.yaml`?

Once this PR is submitted, we can validate that the workflows for releasing work as expected, absent any pre-release vetting we can do.

internal: b/422461961


bigquery/v2/go.mod
```diff
@@ -0,0 +1,3 @@
+module cloud.google.com/go/bigquery/v2
```
codyoss, line 1: Might need to drop something in the go.work as well for this.

shollyman: added the entry, and tidying didn't appear to touch the checksums (expected since the v2 doesn't do anything yet).

codyoss (approved):

PR #12419, opened 2025-06-06
fix(storage): Avoid integration test segfaults.

cjc25:
If the JWT private key file is not specified in an environment variable, then skip the tests which would otherwise require it. This matches what other tests do. Without this change, `jwt.Email` results in a nil dereference.

### DIFF
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -4048,6 +4048,10 @@ func TestIntegration_RequesterPaysOwner(t *testing.T) {
 		if err != nil {
 			t.Fatalf("testutil.JWTConfig: %v", err)
 		}
+		if jwt == nil {
+			t.Skip("JSON key file is not present")
+		}
+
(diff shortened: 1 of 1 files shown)


BrennaEpp:
`GCLOUD_TESTS_GOLANG_KEY` is technically required to run the integration tests (see https://github.com/googleapis/google-cloud-go/blob/main/CONTRIBUTING.md#local-setup), so I think we should fail here instead of skipping.

I think we are inconsistent with the creds we use in the tests - a while back I started documenting this but never quite finished and we've expanded since then.  If most other tests work without the env var set, maybe we can instead extract the info we need from the creds set on the client?

cjc25:
Thanks for the integration test instruction pointer! I don't have a project set up like that but I can take a look. I made these all (and the other places where we currently skip when jwt == nil) into test failures.

PR #12422, opened 2025-06-07
refactor(storage): Pipeline gRPC writes.

cjc25:
Modify the gRPC writer to send additional data while waiting for the current chunk to flush. This is a substantial refactor.

Per the Go io.Writer interface contract, we must never modify or retain the slice that the caller provides to Write. However, that doesn't mean we have to copy every byte into a writer-controlled buffer: we can refer to the byte slice in place. Therefore, if callers call Write() with more bytes than the chunk size, we can send them to the service immediately as long as we don't return from Write() until we no longer need the caller's slice.

By sending data as soon as callers provide it, we get a substantial single-stream throughput increase for large objects. This is especially evident when callers provide large buffers to Write() calls.

There are two followup investigations made possible by this refactor. The first is to flush less frequently when the caller provides write slices much larger than the chunk size. This may provide an even larger throughput improvement when Write() is called with a large buffer, and is straightforward to implement.

The second is to flush more frequently when the caller provides write slices much smaller than the chunk size. (E.g. split a 16MiB chunk into 2x8MiB sub-chunks, and flush each when they're full.) This can avoid pipeline stalls in more scenarios, by increasing the likelihood that part of the chunk is available to buffer data without waiting for a flush acknowledgement.


storage/grpc_writer.go
```diff
@@ -42,311 +44,691 @@ const (
...
+	flushSupported        bool
+	sendCRC32C            bool
+	forceOneShot          bool
+	forceEmptyContentType bool
+	append                bool
+	appendGen             int64
+	finalizeOnClose       bool
+
+	buf              []byte
+	writeQuantum     int
```
BrennaEpp, line 274: Can we add a written description of what a `writeQuantum` means so that the code is easier to understand?

storage/grpc_writer.go
```diff
@@ -42,311 +44,691 @@ const (
...
+		w.handleCompletion(c)
+	}
+	return w.streamSender.err()
 }
 
-func newGRPCWriter(c *grpcStorageClient, s *settings, params *openWriterParams, r io.Reader, pr *io.PipeReader, pw *io.PipeWriter) (*gRPCWriter, error) {
-	if params.attrs.Retention != nil {
-		// TO-DO: remove once ObjectRetention is available - see b/308194853
-		return nil, status.Errorf(codes.Unimplemented, "storage: object retention is not supported in gRPC")
+type gRPCCommandHandleChans struct {
```
BrennaEpp: Is there a more descriptive name for this? Scanning over the code it is not obvious what this is for/

storage/grpc_writer.go
cjc25: Added a sentence. I can change this name too if you don't like the use of the word "quantum" here. It's to generalize the (default) case where the service can receive messages in units of 2MiB but the buffer on the client side is larger (16MiB). For very small buffers, `writeQuantum == cap(buf)`.

"Quantum" is not a perfect analogy regardless (Requests are "really" quantized to 1 byte, and even 0-byte messages are allowed). Is "singleWriteMaxSize" better, and maybe then I can drop the comment?

storage/grpc_writer.go
cjc25: Basically, gRPCWriterCommand `handle()` implementations have a writable request channel and a readable completion channel. The other side is the buffer sender, which on `connect` gets a readable request channel and a writable completion channel. The write loop constructs these channels before connecting the buffer sender. Separating the buffer sender completions from the user-generated requests is in particular what allows us to get pipelining.

I changed the name here but I don't think it's exactly what you had in mind. Other thoughts that came to me were to call these two structs (`gRPCWriterCommandHandleChans` and `gRPCBufSenderChans` respectively) the "outer" and "inner" channel pair, or the "caller" and "implementer" channel pair, but I don't like those either since they're too generic.

AI suggested these (I'd disambiguate with prefixes as in the rest of this file):

- WriteClient and WriteHandler (too generic?)
- CommandEndpoint and SenderEndpoint (I think this is fine although I tend to think of an endpoint as an outbound concept, so SenderEndpoint feels slightly off)
- WriteRequester and WriteExecutor (also pretty generic, but maybe better than the first suggestion)

storage/grpc_writer.go
```diff
@@ -42,311 +44,700 @@ const (
...
-			return true
+
+	done := make(chan struct{})
+	cmd := &gRPCWriterCommandWrite{p: p, done: done}
+	if w.streamSender == nil {
+		w.currentCommand = cmd
+		w.initializeSender()
+	} else {
+		select {
+		case <-w.donec:
```
tritone, line 67: Is this case intended for if you send on a writer that's already closed? if so what is supposed to happen?

storage/grpc_writer.go
```diff
@@ -42,311 +44,700 @@ const (
...
+		copy(w.buf[origLen:], p)
+		return len(p), nil
 	}
-	// If retry settings allow retries normally, fall back to that behavior.
-	newr.shouldRetry = func(err error) bool {
-		if errors.Is(err, bidiWriteObjectRedirectionError{}) {
-			return true
+
+	done := make(chan struct{})
+	cmd := &gRPCWriterCommandWrite{p: p, done: done}
```
tritone, line 61: Did you profile this with many small-buffer writes? Just wanted to make sure that the extra overhead of the channel for each cmd is not too much of an issue.

storage/grpc_writer.go
```diff
@@ -42,311 +44,700 @@ const (
...
+	donec              chan struct{}
+}
+
+func (w *gRPCWriter) initializeSender() {
+	if w.attrs.ContentType == "" && !w.forceEmptyContentType {
+		w.spec.Resource.ContentType = w.detectContentType()
+	}
+	w.streamSender = w.pickBufferSender()
+
+	runWriteLoop := func(ctx context.Context) error {
```
tritone: nit: This additional wrapper seems unneeded?

storage/grpc_writer.go
```diff
@@ -939,90 +1266,84 @@ func (s *gRPCAppendBidiWriteBufferSender) sendOnConnectedStream(buf []byte, offs
...
-		if s.stream == nil {
-			sendFirstMessage = true
-			if err = s.connect(ctx); err != nil {
-				return
-			}
-		}
+type gRPCWriteRequestParams struct {
+	appendable   bool
+	bucket       string
+	routingToken *string
```
tritone, line 1329: Is having an empty string routing token valid?

storage/grpc_writer.go
tritone: Discussed this offline; sounds like it could be worth it to just preserve this outer channel between Write() calls rather than closing/recreating each time. But profiling will tell us what the actual overhead is.

storage/grpc_writer.go
cjc25: Either the writer is already closed, or the writer gets closed due to a permanent error before it can process the provided command. Basically this is how we handle asynchronous failures which the calling code may not have detected yet.

The underlying writer promises to set `w.streamResult` before closing `w.donec`

storage/grpc_writer.go
cjc25: `w.lastErr` is just to persist the prior failure across multiple `run()` retries, and it's convenient to get that from the `writeLoop` return instead of remembering it each place we break out of that loop.

But you're right it doesn't have to be named: moved these two lines into the `run()` call directly as an anonymous function.

storage/grpc_writer.go
cjc25: No it is not. Do you suggest switching this from `*string` to `string` and copying from `e.RoutingToken` in `maybeHandleRedirectionError`?

storage/grpc_writer.go
cjc25: See http://b/422440765#comment4 and comment 5 for more discussion. I'm inclined to merge like this and improve afterwards if necessary, if that's alright.

To summarize for any reader without Google internal bug access: there is a minor (few % on a very low base) CPU increase and therefore latency increase and IOPS decrease for 1-byte operations. Once op size goes above a de minimis value, this effect disappears and the throughput increase from pipelining results in slightly lower mean and tail latencies and higher IOPS. There is still a small % CPU increase, but since the write is now large enough to block on IO for more of the time, the CPU base is even lower.

We could trade the simplicity of a write completion channel per op for a somewhat less forgiving interface with a write completion channel per `storage.Writer`. This is consistent with the existing API contract based on `io.PipeWriter`, and gets us ~1/3 of the new CPU cost back. I think it's reasonable to do that in a focused followup PR.

storage/grpc_writer.go
```diff
@@ -405,9 +405,11 @@ func (w *gRPCWriter) writeLoop(ctx context.Context) error {
 		return fmt.Errorf("storage: retry deadline of %s reached after %v attempts; last error: %w", w.chunkRetryDeadline, w.attempts, w.lastErr)
 	}
 	requests := make(chan gRPCBidiWriteRequest, w.sendableUnits)
+	// only one request will be outstanding at a time.
+	requestAcks := make(chan struct{}, 1)
```
tritone, line 404: nice fix!

storage/grpc_writer.go
tritone: Sounds good, thanks for the summary here.

cjc25:
@tritone IIUC this is ready to merge now

tritone (approved):

PR #12428, opened 2025-06-10
test(storage): add retry backoff test

cojenco:


storage/client_test.go
```diff
@@ -2303,6 +2303,40 @@ func TestRetryDeadlineExceededEmulated(t *testing.T) {
...
+			config := &retryConfig{backoff: &gax.Backoff{Initial: backoff}}
+			start := time.Now()
+			if _, err := client.GetBucket(ctx, bucket, nil, idempotent(true), withRetryConfig(config)); err != nil {
+				t.Fatalf("GetBucket with backoff %v: got unexpected error %v, want nil", backoff, err)
+			}
+			durations[i] = time.Since(start)
+		}
+		// Test that extended backoff is applied and results in longer duration.
+		if durations[1] <= durations[0] {
+			t.Fatalf("GetRetryBackoff: Expected longer durations[1] %v with backoffs[1] %v than durations[0] %v with backoffs[0] %v", durations[1], backoffs[1], durations[0], backoffs[0])
```
BrennaEpp, line 2335: Unfortunately I believe this can be flaky for Go, due to the way gax.Backoff is implemented; it waits randomly between 1ns and `currRetryLimit`. `currRetryLimit` would in these cases be the time.Duration in `backoffs` for the first retry (double for the second), so on average this will pass, but there could be a case where the first random number ends up at 10ms and the second at 9ms for example.

This may be less flaky if you have only a single retry (or set Multiplier:1) and the first "Initial" is set to 1ns. Then the backoff jitter for the second loop will be <= to the first loop's backoff jitter. It could still be the case however that a very small duration is chosen for the second one; small enough that some other execution step dominates the time and makes the test flaky.

cojenco: Oooh interesting! Good to know that's how gax.Backoff works. I was also thinking to mock gax.Backoff and maybe Backoff.Pause(). I couldn't find a good mock test example. Would that be feasible?

I can change this test to a single retry.

BrennaEpp: I did a test a while back mocking gax.Backoff. I couldn't find a better way than creating an interface for backoffs and replacing gax.Backoff with a version that implemented that interface - it's a lot of changes but feasible. I don't know that we necessarily want to do that.

This old PR has the code which has since been eliminated https://github.com/googleapis/google-cloud-go/pull/11200

BrennaEpp (approved):
Maybe we can consider mocking if this test ends up being flaky.

PR #12433, opened 2025-06-10
test(storage): add universe domain signedURL DefaultSignBytes integration test

frq-dev:


storage/integration_test.go
```diff
@@ -6963,6 +6963,91 @@ func TestIntegration_UniverseDomains(t *testing.T) {
...
+		t.Skipf("%s must be set. See CONTRIBUTING.md for details", testUniverseCreds)
+	}
+	project := os.Getenv(testUniverseProject)
+	if project == "" {
+		t.Fatalf("%s must be set. See CONTRIBUTING.md for details", testUniverseProject)
+	}
+	location := os.Getenv(testUniverseLocation)
+	if location == "" {
+		t.Fatalf("%s must be set. See CONTRIBUTING.md for details", testUniverseLocation)
+	}
```
BrennaEpp: Can we extract this into a helper function since we are using this in several tests?

storage/integration_test.go
```diff
@@ -6963,6 +6963,91 @@ func TestIntegration_UniverseDomains(t *testing.T) {
...
+
+	// Get new Auth creds
+	newAuthCreds, err := credentials.DetectDefault(&credentials.DetectOptions{
+		Scopes:           scopes,
+		CredentialsFile:  credFile,
+		UniverseDomain:   universeDomain,
+		UseSelfSignedJWT: true,
+	})
+	if err != nil {
+		t.Fatalf("Cannot get Auth Credentials to create client")
```
BrennaEpp: We should log the `err` here as well.

storage/integration_test.go
```diff
@@ -6963,6 +6963,91 @@ func TestIntegration_UniverseDomains(t *testing.T) {
...
+		url, err := bucket.SignedURL(obj, &opts)
+		if err != nil {
+			t.Fatalf("unable to create signed URL: %v", err)
+		}
+
+		if err := verifySignedURL(url, nil, contents); err != nil {
+			t.Fatalf("problem with the signed URL: %v", err)
+		}
+	}, option.WithAuthCredentials(newAuthCreds), option.WithUniverseDomain(universeDomain))
+
```
BrennaEpp, line 7011: nit: remove line

storage/integration_test.go
```diff
@@ -6963,6 +6963,91 @@ func TestIntegration_UniverseDomains(t *testing.T) {
...
+		if err := writeObject(ctx, bucket.Object(obj), "text/plain", contents); err != nil {
+			t.Fatalf("writing: %v", err)
+		}
+
+		defer h.mustDeleteObject(bucket.Object(obj))
+
+		opts := SignedURLOptions{
+			Method:         "GET",
+			Expires:        time.Now().Add(30 * time.Second),
+			GoogleAccessID: jwt.Email,
```
BrennaEpp: I forget; what happens if we don't include this line with `GoogleAccessID`? Does the detection fail?

storage/integration_test.go
```diff
@@ -6963,6 +6963,91 @@ func TestIntegration_UniverseDomains(t *testing.T) {
...
+	// Get JSON info
+	jsonKey, err := os.ReadFile(credFile)
+	if err != nil {
+		t.Fatalf("cannot read the JSON key file, err: %v", err)
+	}
+
+	jwt, err := google.JWTConfigFromJSON(jsonKey, scopes...)
+	if err != nil {
+		t.Fatalf("google.JWTConfigFromJSON: %v", err)
+	}
```
BrennaEpp: This can be a separate function as well to find the email.

storage/integration_test.go
frq-dev: It still passes for http but must go through detectDefaultGoogleAccessID. I only set it so that it would guaranteed reach and test DefaultSignBytes.

storage/integration_test.go
```diff
@@ -6960,7 +6947,103 @@ func TestIntegration_UniverseDomains(t *testing.T) {
...
+type universeDomainTestVars struct {
+	universeDomain string
+	credFile       string
+	project        string
+	location       string
+}
+
+// Gets Universe Domain environment variables for Universe Domain tests
+// Returns universeDomainTestVars pointer for easy access
+func getUniverseDomainTestVars() (*universeDomainTestVars, error) {
```
BrennaEpp: nit: Generally in Go we don't use "get" in function names. https://go.dev/doc/effective_go#Getters

internal/testutil/context.go
```diff
@@ -133,6 +133,13 @@ func JWTConfig() (*jwt.Config, error) {
 	return jwtConfigFromFile(os.Getenv(envPrivateKey), nil)
 }
 
+// JWTConfigExplicit reads an explicit credentials file
+// that is passed in as a parameter. It ignores scopes.
+// If the environment variable is empty, it returns (nil, nil).
+func JWTConfigExplicit(credFile string) (*jwt.Config, error) {
```
BrennaEpp: nit: I'd just export the original `jwtConfigFromFile` or create a function in Storage instead.

internal/testutil/context.go
```diff
@@ -133,6 +133,13 @@ func JWTConfig() (*jwt.Config, error) {
 	return jwtConfigFromFile(os.Getenv(envPrivateKey), nil)
 }
 
+// JWTConfigExplicit reads an explicit credentials file
+// that is passed in as a parameter. It ignores scopes.
+// If the environment variable is empty, it returns (nil, nil).
```
BrennaEpp: > `If the environment variable is empty`

Which environment variable?

storage/integration_test.go
```diff
@@ -6960,7 +6947,103 @@ func TestIntegration_UniverseDomains(t *testing.T) {
...
+
+	// Direct connectivity is not supported yet for this feature.
+	const disableDP = "GOOGLE_CLOUD_DISABLE_DIRECT_PATH"
+	t.Setenv(disableDP, "true")
+
+	ctx := skipExtraReadAPIs(skipGRPC("not yet available in gRPC - b/308194853"), "no reads in test")
+
+	udTestVars, err := getUniverseDomainTestVars()
+	if err != nil {
+		t.Fatalf("Cannot get Universe Domain vars, err: %v", err)
```
BrennaEpp: I think we can just log the error we get here, this doesn't add much more information.

BrennaEpp:
```suggestion
		t.Fatal(err)
```

I think we can just log the error we get here, this doesn't add much more information.

storage/integration_test.go
```diff
@@ -6960,7 +6947,103 @@ func TestIntegration_UniverseDomains(t *testing.T) {
...
+	}
+
+	multiTransportTest(ctx, t, func(t *testing.T, ctx context.Context, _, prefix string, client *Client) {
+		h := testHelper{t}
+		bucket := client.Bucket(prefix + uidSpace.New())
+
+		h.mustCreate(bucket, udTestVars.project, &BucketAttrs{Location: udTestVars.location})
+		defer h.mustDeleteBucket(bucket)
+
+		obj := "testBucketSignedURL"
```
BrennaEpp: Can we change the object name per test? `uidSpaceObjects.New()`

storage/integration_test.go
BrennaEpp: AFAIK whether we have to detect the access ID or not  should not affect `defaultSignBytesFunc`?

storage/integration_test.go
```diff
@@ -6960,7 +6947,103 @@ func TestIntegration_UniverseDomains(t *testing.T) {
...
+
+	// Get new Auth creds
+	newAuthCreds, err := credentials.DetectDefault(&credentials.DetectOptions{
+		Scopes:           scopes,
+		CredentialsFile:  udTestVars.credFile,
+		UniverseDomain:   udTestVars.universeDomain,
+		UseSelfSignedJWT: true,
+	})
+	if err != nil {
+		t.Fatalf("Cannot get Auth Credentials to create client, err: %v", err)
```
BrennaEpp:
```suggestion
		t.Fatalf("cannot get Auth Credentials to create client, err: %v", err)
```

storage/integration_test.go
```diff
@@ -7026,7 +7019,7 @@ type universeDomainTestVars struct {
 
 // Gets Universe Domain environment variables for Universe Domain tests
 // Returns universeDomainTestVars pointer for easy access
-func getUniverseDomainTestVars() (*universeDomainTestVars, error) {
+func UniverseDomainTestVars() (*UniverseDomainVars, error) {
```
BrennaEpp: nit
```suggestion
func universeDomainTestVars() (*UniverseDomainVars, error) {
```

PR #12434, opened 2025-06-10
refactor(storage): syncronizing http and grpc clients

frq-dev:


storage/http_client.go
```diff
@@ -90,6 +92,11 @@ func newHTTPStorageClient(ctx context.Context, opts ...storageOption) (storageCl
 	} else {
 		var hostURL *url.URL
 
+		c, err := internaloption.AuthCreds(ctx, o)
+		if err == nil {
+			creds = c
+		}
+
```
BrennaEpp: Let's keep this in a separate change as it changes the behaviour of the client.

storage/http_client.go
```diff
@@ -106,10 +113,12 @@ func newHTTPStorageClient(ctx context.Context, opts ...storageOption) (storageCl
 		endpoint := hostURL.String()
 
 		// Append the emulator host as default endpoint for the user
-		o = append([]option.ClientOption{option.WithoutAuthentication()}, o...)
-
-		o = append(o, internaloption.WithDefaultEndpointTemplate(endpoint))
-		o = append(o, internaloption.WithDefaultMTLSEndpoint(endpoint))
+		o = append([]option.ClientOption{
+			option.WithoutAuthentication(),
+			internaloption.SkipDialSettingsValidation(),
```
BrennaEpp: This line as well

storage/storage.go
```diff
@@ -246,13 +176,21 @@ func NewClient(ctx context.Context, opts ...option.ClientOption) (*Client, error
 // You may configure the client by passing in options from the [google.golang.org/api/option]
 // package.
 func NewGRPCClient(ctx context.Context, opts ...option.ClientOption) (*Client, error) {
+	var creds *auth.Credentials
+
+	c, err := internaloption.AuthCreds(ctx, opts)
+	if err == nil {
+		creds = c
+	}
+
```
BrennaEpp: Can we extract this to a separate function used for both http and grpc? Also, I think we need to add default options before doing this.

storage/http_client.go
```diff
@@ -59,7 +59,7 @@ type httpStorageClient struct {
 
 // newHTTPStorageClient initializes a new storageClient that uses the HTTP-JSON
 // Storage API.
-func newHTTPStorageClient(ctx context.Context, opts ...storageOption) (storageClient, error) {
+func newHTTPStorageClient(ctx context.Context, opts ...storageOption) (*httpStorageClient, error) {
 	s := initSettings(opts...)
 	o := s.clientOption
 	config := newStorageConfig(o...)
```
BrennaEpp: This seems to be repeated for both clients. Can we push it to the higher level?

I was thinking maybe:

NewClient() and NewGRPCClient()
1. Add default options (these are different for http and grpc, both check respective STORAGE_EMULATOR_HOST)
2. Get credentials if env var not set (we can change this later)
3. Resolve settings + storageConfig (we should probably merge these two? seem to be serving the same purpose. but we can leave that for later)
4. Call respective newXXXStorageClient, which now should have none of the above steps.


That way all similar steps (adding default opts, resolving opts, auth...) are at the higher level and specific things (creating gapic client, creating http client, starting metrics) are at the lower level client creation function.

storage/grpc_client.go
```diff
@@ -85,10 +86,17 @@ func defaultGRPCOptions() []option.ClientOption {
 	if host := os.Getenv("STORAGE_EMULATOR_HOST_GRPC"); host != "" {
 		// Strip the scheme from the emulator host. WithEndpoint does not take a
 		// scheme for gRPC.
-		host = stripScheme(host)
+		var hostURL *url.URL
+		var err error
+
+		hostURL, err = parseURL(host)
```
BrennaEpp: Why not 
```suggestion
		hostURL, err := parseURL(host)
```

storage/storage.go
```diff
@@ -441,6 +377,22 @@ func stripScheme(host string) string {
...
+	var err error
+	if strings.Contains(host, "://") {
+		h, err = url.Parse(host)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// Add scheme for user if not supplied in STORAGE_EMULATOR_HOST
+		// URL is only parsed correctly if it has a scheme, so we build it ourselves
+		h = &url.URL{Scheme: "http", Host: host}
```
BrennaEpp: Can we rename this function to indicate that it should only be used when an emulator host is set and/or add a comment at the top of the function indicating this behaviour?

My concern here is that we don't want to be adding an http scheme outside of emulator environments, and right now the function sounds generic enough.

storage/storage.go
```diff
@@ -441,6 +377,22 @@ func stripScheme(host string) string {
...
+	}
+	return h, nil
```
BrennaEpp: This function can be simplified I believe to something like this
```suggestion
	if strings.Contains(host, "://") {
		return url.Parse(host)
	} 
	// Add scheme for user if not supplied. A URL is only parsed correctly if it has a scheme,
	// so we build it ourselves. 
	return &url.URL{Scheme: "http", Host: host}, nil
```

storage/storage.go
```diff
@@ -317,8 +255,6 @@ func CheckDirectConnectivitySupported(ctx context.Context, bucket string, opts .
 // Close need not be called at program exit.
 func (c *Client) Close() error {
 	// Set fields to nil so that subsequent uses will panic.
-	c.hc = nil
-	c.raw = nil
```
BrennaEpp, line 321: Why was this removed?

storage/bucket.go
```diff
@@ -319,7 +319,7 @@ func (b *BucketHandle) defaultSignBytesFunc(email string) func([]byte) ([]byte,
 	return func(in []byte) ([]byte, error) {
 		ctx := context.Background()
 
-		opts := []option.ClientOption{option.WithHTTPClient(b.c.hc)}
+		opts := []option.ClientOption{option.WithHTTPClient(b.c.tc.(*httpStorageClient).hc)}
```
BrennaEpp: I think we should do a type check here before grabbing the hc.

Something like `if _, ok := client.(*httpStorageClient); ok {`
 
Then if the grpc client is used, we can fail here (because it is failing anyways, right? Please verify). With a message such as: "not supported yet". Once we add endpoint/creds capture logic to GRPC we can remove that message and do the correct thing.

storage/grpc_client.go
BrennaEpp: also, why was this changed from the stripScheme function?

storage/http_client.go
```diff
@@ -62,56 +60,12 @@ type httpStorageClient struct {
...
-		endpoint := hostURL.String()
-
-		// Append the emulator host as default endpoint for the user
-		o = append([]option.ClientOption{option.WithoutAuthentication()}, o...)
-
-		o = append(o, internaloption.WithDefaultEndpointTemplate(endpoint))
-		o = append(o, internaloption.WithDefaultMTLSEndpoint(endpoint))
+	defaults, err := defaultHTTPOptions(ctx, o...)
+	if err != nil {
+		return nil, err
```
BrennaEpp: We need to add these options before trying to detect/determine the creds. Same goes for GRPC. Can you move this to the higher level function?

BrennaEpp (commented):
Some initial comments

storage/bucket.go
frq-dev: Agreed, I was thinking of a switch statement that separates the creation of the IAM client as well in the future as mentioned here: https://b.corp.google.com/issues/422801044

storage/grpc_client.go
frq-dev: It has additional error handling for validating the URL. It also creates more uniformity between grpc and http clients.

storage/storage.go
frq-dev: I agree, I'll rename it

storage/storage.go
frq-dev: These fields no longer exist

storage/bucket.go
frq-dev: Also it doesn't fail for grpc signedURL. It only fails when universeDomain is non-default. Instead of failing, I will just skip the httpClient option for grpc as the grpcclient hc was nil anyways before

storage/bucket.go
```diff
@@ -318,8 +318,13 @@ func (b *BucketHandle) detectDefaultGoogleAccessID() (string, error) {
...
 	return func(in []byte) ([]byte, error) {
 		ctx := context.Background()
+		var opts []option.ClientOption
 
-		opts := []option.ClientOption{option.WithHTTPClient(b.c.hc)}
+		if _, ok := b.c.tc.(*httpStorageClient); ok {
+			opts = []option.ClientOption{option.WithHTTPClient(b.c.tc.(*httpStorageClient).hc)}
+		} else {
+			opts = []option.ClientOption{}
+		}
```
BrennaEpp, line 327:
```suggestion
opts := []option.ClientOption{}

		if _, ok := b.c.tc.(*httpStorageClient); ok {
			opts = append()[]option.ClientOption{option.WithHTTPClient(b.c.tc.(*httpStorageClient).hc)}
```

storage/storage.go
BrennaEpp: We still should close them; we should be able to access them by casting as done in the defaultSignBytes func.

cpriti-os:
@frq-dev do you have a bug id for this task? We can add it to the backlog and prioritize this if required.

frq-dev:
@cpriti-os yeah, its b/419882831 and b/419883126

PR #12435, opened 2025-06-10
feat(bigquery/v2): bump to 2.x alpha

shollyman:
another minor commit to validate release behavior

Release-As: 2.0.0-alpha1

### DIFF
--- a/bigquery/v2/CHANGES.md
+++ b/bigquery/v2/CHANGES.md
@@ -1,1 +1,1 @@
-# Changes
+# Changelog


codyoss (approved):
nit: it may be useful to call it .*alpha1 or something incase you need more of these.

shollyman:
> nit: it may be useful to call it .*alpha1 or something incase you need more of these.

Good callout.  We'll probably need a number of these, so I've adjusted the target to `2.0.0-alpha1`

alvarowolfx (approved):

PR #12436, opened 2025-06-11
feat(spanner): add option for how to call BeginTransaction

olavloite:
Adds an option to the Spanner client for how BeginTransaction should be called.

The default for read/write transactions that are executed in a transaction runner, is to use inlined-begin. That is; the BeginTransaction is inlined with the first statement in the transaction. This reduces the number of round-trips to Spanner by 1.

The default for statement-based read/write transactions and read-only transactions is to use an explicit BeginTransaction RPC invocation. This requires one additional round-trip to Spanner, but can be more efficient if the application wants to execute multiple queries in parallel as the first statements in the transaction.

This change introduces an explicit option that can be used to change the above defaults for a single transaction. This is especially useful for statement-based read/write transactions, as these are used by the database/sql driver, and currently requires an additional round-trip for every transaction.


rahul2393 (approved):

spanner/client.go
```diff
@@ -1114,13 +1114,35 @@ func (c *Client) rwTransaction(ctx context.Context, f func(context.Context, *Rea
...
+			t.txReadOnly.qo = c.qo
+			t.txReadOnly.ro = c.ro
+			t.txReadOnly.disableRouteToLeader = c.disableRouteToLeader
+			t.wb = []*Mutation{}
+			t.txOpts = c.txo.merge(options)
+			t.ct = c.ct
+			t.otConfig = c.otConfig
+		}
+		if t.shouldExplicitBegin(attempt, options) {
+			if t == nil {
```
rahul2393, line 1129: if t is nil then I think t.shouldExplicitBegin would had already paniced?

olavloite: No, a nil receiver is allowed, as long as it is not actually being used in the function. And in this case it isn't. In that sense, defining the function with a `ReadWriteTransaction` receiver is unnecessary in this case, but as it already was like that, I didn't change it.

spanner/transaction_test.go
```diff
@@ -941,54 +941,226 @@ func TestReadWriteStmtBasedTransaction_CommitNonAbortedErrorReturned(t *testing.
...
+	for _, beginTransactionOption := range []BeginTransactionOption{DefaultBeginTransaction, ExplicitBeginTransaction, InlinedBeginTransaction} {
+		tx := client.ReadOnlyTransaction().WithBeginTransactionOption(beginTransactionOption)
+		wg := sync.WaitGroup{}
+		errs := make([]error, 0, numQueries)
+		for range numQueries {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				iter := tx.Query(ctx, Statement{SQL: SelectFooFromBar})
+				if err := consumeIterator(iter); err != nil {
```
rahul2393, line 1013: nit: appending to shared err slice without lock here.

spanner/client.go
```diff
@@ -1114,13 +1114,35 @@ func (c *Client) rwTransaction(ctx context.Context, f func(context.Context, *Rea
 			// Some operations (for ex BatchUpdate) can be long-running. For such operations set the isLongRunningTransaction flag to be true
 			t.setSessionEligibilityForLongRunning(sh)
 		}
-		if t.shouldExplicitBegin(attempt) {
+		initTx := func(t *ReadWriteTransaction) {
+			t.txReadOnly.sp = c.idleSessions
+			t.txReadOnly.sp = c.idleSessions
```
sakthivelmanii: we are assigning it twice.

sakthivelmanii (approved):

spanner/transaction_test.go
olavloite: Thanks for spotting that, fixed.

spanner/client.go
olavloite: Thanks for spotting that, fixed.

sakthivelmanii (approved):

PR #12438, opened 2025-06-11
feat(bigtable): connectivity_error_count client side metric

bhshkh:
Record [connectivity_error_count](https://cloud.google.com/bigtable/docs/client-side-metrics-descriptions#connectivity-error-count) metric

Java implementation:

https://github.com/googleapis/java-bigtable/blob/6be14faf2fd67001474f2df774029f8f7ed578cc/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracer.java#L487-L492

### DIFF
--- a/bigtable/bigtable.go
+++ b/bigtable/bigtable.go
@@ -2317,9 +2317,13 @@ func recordAttemptCompletion(mt *builtinMetricsTracer) {
 	attemptLatAttrs, _ := mt.toOtelMetricAttrs(metricNameAttemptLatencies)
 	mt.instrumentAttemptLatencies.Record(mt.ctx, elapsedTime, metric.WithAttributes(attemptLatAttrs...))
 
-	// Record server_latencies
+	// Record server_latencies and connectivity_error_count
+	connErrCountAttrs, _ := mt.toOtelMetricAttrs(metricNameConnErrCount)
 	serverLatAttrs, _ := mt.toOtelMetricAttrs(metricNameServerLatencies)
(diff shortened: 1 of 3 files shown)


bigtable/bigtable.go
```diff
@@ -2317,9 +2317,13 @@ func recordAttemptCompletion(mt *builtinMetricsTracer) {
...
 
-	// Record server_latencies
+	// Record server_latencies and connectivity_error_count
+	connErrCountAttrs, _ := mt.toOtelMetricAttrs(metricNameConnErrCount)
 	serverLatAttrs, _ := mt.toOtelMetricAttrs(metricNameServerLatencies)
 	if mt.currOp.currAttempt.serverLatencyErr == nil {
 		mt.instrumentServerLatencies.Record(mt.ctx, mt.currOp.currAttempt.serverLatency, metric.WithAttributes(serverLatAttrs...))
+		mt.instrumentConnErrCount.Add(mt.ctx, 0, metric.WithAttributes(connErrCountAttrs...))
+	} else {
+		mt.instrumentConnErrCount.Add(mt.ctx, 1, metric.WithAttributes(connErrCountAttrs...))
```
mutianf: for direct path we're not gonna get server-latency because it doesn't go through GFE / CFE. So another way to check if we've reached the server is to see if we get a cluster id / zone id in the trailer. I think we should only increment the counter when server latency is empty and cluster id / zone ids are also empty.

bhshkh: Got it. Updated the code.

PR #12443, opened 2025-06-13
feat(auth): add optionadapt module

quartzmo:


auth/optionadapt/go.mod
```diff
@@ -0,0 +1,35 @@
+module cloud.google.com/go/auth/optionadapt
```
codyoss, line 1: I wonder if this could just be an internal package in cloud.google.com/go module? That way we don't have to expose this at all and all of the gapics can still get at it.

quartzmo: Yes, great idea!

PR #12464, opened 2025-06-17
feat(bigquery/v2): add interim BigQuery aggregated service client

shollyman:
This PR provides a handwritten aggregated service client for working with the new BQ generated surface, and the instantiation mechanisms for REST and gRPC transports.

In the future we'll generate this directly via either augmentation of gapic-generator-go or via AST processing of apiv2 clients, and expose the rest of the functionality, and expose features like the default call options per RPC-service.


bigquery/v2/apiv2_monoclient/monoclient.go
```diff
@@ -0,0 +1,153 @@
...
+	err = mc.rapClient.Close()
+	if err != nil {
+		errs = append(errs, fmt.Errorf("RowAccessPolicyClient: %w", err))
+	}
+	err = mc.tblClient.Close()
+	if err != nil {
+		errs = append(errs, fmt.Errorf("TableClient: %w", err))
+	}
+	if len(errs) > 0 {
+		return fmt.Errorf("closing subclients yielded %d errors, first error: %w", len(errs), errs[0])
```
alvarowolfx: might worth adding a `groupErr` function to combine the errors into a single one, and then check for `nil`

```
// [...]
err = groupErr(errs)
if err != nil {
   return fmt.Errorf("err closing subclients: %w", err)
}
// [...]

func groupErr(errs []error) error {
	if len(errs) == 0 {
		return nil
	}
	var errMsgs []string
	for _, err := range errs {
		errMsgs = append(errMsgs, err.Error())
	}
	return fmt.Errorf("found %d errors: %v", len(errs), strings.Join(errMsgs, ", "))
}
```

bigquery/v2/apiv2_monoclient/monoclient.go
```diff
@@ -0,0 +1,153 @@
...
+	mc.rapClient, err = bigquery.NewRowAccessPolicyRESTClient(ctx, opts...)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("NewRowAccessPolicyRESTClient: %w", err))
+	}
+	mc.tblClient, err = bigquery.NewTableRESTClient(ctx, opts...)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("NewTableRESTClient: %w", err))
+	}
+	if len(errs) > 0 {
+		return nil, fmt.Errorf("NewClient: %d errors during instantiation, first error: %w", len(errs), errs[0])
```
alvarowolfx: reuse the `groupErr` method

alvarowolfx (commented):
still not happy with the name, but other than that looks good. My top suggestions would be `UnifiedClient` with `apiv2_unified` or just `Client` on `apiv2_client` pkg.

I think worth showing all the errors, maybe the error when creating the client is related to permissions, so might be good to show all errors. Tho permissions errors are only gonna appear when invoking methods. In any case, food for thought.

leahecole:
> still not happy with the name, but other than that looks good. My top suggestions would be `UnifiedClient` with `apiv2_unified` or just `Client` on `apiv2_client` pkg.
> 
> I think worth showing all the errors, maybe the error when creating the client is related to permissions, so might be good to show all errors. Tho permissions errors are only gonna appear when invoking methods. In any case, food for thought.

BigQueryClient for consistency with Node? (not that I care super hard about cross language consistency)

bigquery/v2/apiv2_monoclient/monoclient.go
shollyman: Simplified this further, it's now just an errors.Join() operation.

shollyman:
> BigQueryClient for consistency with Node? (not that I care super hard about cross language consistency)

done.

alvarowolfx:
> 
> 
> > BigQueryClient for consistency with Node? (not that I care super hard about cross language consistency)
> 
> 
> 
> done.

I think BigQueryClient is not idiomatic to go, calling it just Client is better as it is already under a bigquery package.

bigquery/v2/apiv2_monoclient/monoclient.go
alvarowolfx: oh, totally forgot about `errors.Join`, good call

shollyman:
> I think BigQueryClient is not idiomatic to go, calling it just Client is better as it is already under a bigquery package.

Done.

PR #12470, opened 2025-06-18
feat(bigquery): load job and external table opts for custom time format, null markers and source column match

alvarowolfx:
Towards internal b/374142085

### DIFF
--- a/bigquery/external.go
+++ b/bigquery/external.go
@@ -133,6 +133,27 @@ type ExternalDataConfig struct {
 	// Metadata Cache Mode for the table. Set this to
 	// enable caching of metadata from external data source.
 	MetadataCacheMode MetadataCacheMode
+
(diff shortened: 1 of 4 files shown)


alvarowolfx:
`apidiff`  is complaining that by adding the new field `NullMarkers` to `CSVOptions`, which is an array, this makes it not comparable anymore. But I don't think that is considered a breaking change for us.  Need to check if we can ignore it on a given PR.

```
Ignoring internal package cloud.google.com/go/bigquery/internal
Ignoring internal package cloud.google.com/go/bigquery/internal/query
Ignoring internal package cloud.google.com/go/bigquery/internal
Ignoring internal package cloud.google.com/go/bigquery/internal/query
CSVOptions: old is comparable, new is not
```

PR #12486, opened 2025-06-23
docs(vertexai): add deprecation note to README

qiaodev:


vertexai/README.md
```diff
@@ -2,11 +2,13 @@
 
 [![Go Reference](https://pkg.go.dev/badge/cloud.google.com/go/vertexai.svg)](https://pkg.go.dev/cloud.google.com/go/vertexai)
 
+> [!NOTE] Starting on June 24, 2025, the cloud.google.com/go/vertexai/genai package is deprecated and will be removed on June 24, 2026, except cloud.google.com/go/vertexai/genai/tokenizer.
```
quartzmo, line 5: Is there a good reason to not deprecate the entire cloud.google.com/go/vertexai module instead? Currently, the only other package is `internal`. Is there a plan to add new exported packages to this module in the future?

quartzmo (changes requested):

vertexai/README.md
qiaodev: This is approved by MSA so I don't think we can change it now.

qiaodev: Discussed in internal chats

PR #12489, opened 2025-06-24
wip(storage): remove object length limit for unfinalized reads

BrennaEpp:


storage/grpc_client.go
```diff
@@ -1654,9 +1654,20 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
...
 	size := obj.GetSize()
 
+	finalized := obj.GetFinalizeTime() != nil
+	negativeOffset := params.offset < 0
+	if !finalized && negativeOffset {
+		// Fix the offset and length of a negative-offset read at time of first
+		// response to ensure data integrity.
+		params.offset = obj.Size + params.offset
+		params.length = obj.Size - params.offset
+	}
```
abhishek10004, line 1669: Why are we updating this and why only for unfinalized objects?

storage/grpc_client.go
```diff
@@ -1882,7 +1897,7 @@ func (r *gRPCReader) runCRCCheck() error {
 func (r *gRPCReader) Read(p []byte) (int, error) {
 	// The entire object has been read by this reader, check the checksum if
 	// necessary and return EOF.
-	if r.size == r.seen || r.zeroRange {
+	if (r.finalized || r.negativeOffset) && r.size == r.seen || r.zeroRange {
```
abhishek10004, line 1906: Please use parentheses to club the last 2 conditions as well otherwise it would evaluate to "( (A || B) && C) || D", i.e. please use "(r.finalized || r.negativeOffset) && (r.size == r.seen || r.zeroRange)"

storage/grpc_client.go
```diff
@@ -1937,7 +1952,7 @@ func (r *gRPCReader) Read(p []byte) (int, error) {
 func (r *gRPCReader) WriteTo(w io.Writer) (int64, error) {
 	// The entire object has been read by this reader, check the checksum if
 	// necessary and return nil.
-	if r.size == r.seen || r.zeroRange {
+	if (r.finalized || r.negativeOffset) && r.size == r.seen || r.zeroRange {
```
abhishek10004, line 1955: Why do we need negativeOffset check for writing? Also, please use parentheses to club the last two conditions together as well.

storage/client_test.go
```diff
@@ -1408,6 +1413,181 @@ func TestWriterSmallFlushEmulated(t *testing.T) {
...
+	// Download content and validate.
+	buf := bytes.NewBuffer([]byte{})
+	r, err := o.NewRangeReader(ctx, 0, -1)
+	if err != nil {
+		t.Errorf("NewRangeReader: %v", err)
+	}
+
+	// Copy exactly the same amount of bytes as object size. This would
+	// trigger a seen == size check for the next copy.
+	if _, err := io.CopyN(buf, r, receivedObjectSize); err != nil {
```
abhishek10004, line 1522: Why do we need both these here?

BrennaEpp: Both the io.Copy? It triggers the `seen == size` case on the second copy.

That is:
- object size: 100 
- obj size from metadata: 10 (this is out of date)
- First copy: copy 0 - 10, as requested
- Second copy: `seen == size`, so we stop copying; this is an error

The `seen == size` check is only done at the start of the copy, so I'm not sure if there is another way to trigger it.

storage/grpc_client.go
BrennaEpp: Here we:
(a) convert a negative offset to its positive counterpart (ex: objsize: 10, offset: -2 would convert to offset 7, asking for bytes 8-9)
(b) explicitly set the length expected

These have no effect for regular objects with the correct object size in the metadata. For (a), the offset conversion would request the same thing. For (b), asking for length -1 is equal to asking for object size and asking for object size - offset.

We need to set this as an unfinalized object may in theory be written to while downloading. Basically, we fix the offset and length of a negative-offset read for data integrity purposes.

We can do this at the time of first response, rather than time of first request, because no responses yet == nothing written to io.Writer and nothing read from io.Reader.

storage/grpc_client.go
BrennaEpp: I think the first is what we want. The second may be fine too, but we can always return early if 0 bytes were requested, regardless of finalized status.

storage/grpc_client.go
abhishek10004: But we handle that currently also while calculating startOffset & length, remain variables (line no. 1683 to 1696) which are what eventually gets used by the reader. 

So, why do we need to udpate param.offset & param.length separately for unfinalized objects.

storage/grpc_client.go
abhishek10004: makes sense to return early if reader was created with length 0. Thanks for clarifying.

storage/grpc_client.go
BrennaEpp: You are correct, but I think the key difference here is that I am setting the fields on the `params` struct. This should change the behaviour of `reopen` in case of a retry. Maybe I can merge this with the code below, I didn't realize that it does pretty much the same thing.

BrennaEpp: Note: I should handle the edge case where the offset is still negative after the conversion.

abhishek10004: but then the current implementation (for finalized objects) also has a bug.

Let's say params.offset is -1000 and object size is 800. We're read 600 bytes, and then stream reopen happens. In this case, the offset used according to current code would be params.offset + seen, which would be -400. But we should have ideally reopened stream from -200 offset.

Could you please check this as well?

storage/client_test.go
abhishek10004: Got it. Thanks.

storage/grpc_client.go
BrennaEpp: Yes, that is my note about handling when the offset is negative after doing params.offset + objSize - really here we just need to make the offset 0 since a negative larger than object size is essentially requesting the whole object.

tritone: Discussed this offline with @abhishek10004 and wrote an emulator test; the test indicates that this issue exists even with the current http storage client. Fortunately GCSFuse does not use negative offsets larger than the object size so I'm not too concerned about downstream effects related to this PR.

I'm going to open an internal issue for a subsequent fix here.

tritone (approved):

storage/grpc_client.go
tritone: I actually was able to fix up the simple case for this pretty easily; with a larger negative offset we should just start at zero and still read to the end of the file.

cojenco (approved):

PR #12496, opened 2025-06-25
fix(spanner): pointer type custom struct decoder

### LINKED ISSUE #12495: packagename: short description of bug
## Client

Spanner

## Environment

macOS (Apple Silicon)
Go 1.24.0
Using Row.ToStruct() with custom types

## Code and Dependencies

```go
package main

import (
	"context"
	"fmt"
	"log"

	"cloud.google.com/go/spanner"
)

type A struct {
	B *B
}

type B struct {
	Val string
}

func (b *B) DecodeSpanner(input interface{}) error {
	log.Println("DecodeSpanner called") // ← never called
	if s, ok := input.(string); ok {
		b.Val = "decoded:" + s
		return nil
	}
	return fmt.Errorf("unexpected type: %T", input)
}

func main() {
	row := spanner.RowFromSlice([]any{"Val"}, []any{"foo"})

	var a A
	if err := row.ToStruct(&a); err != nil {
		log.Fatal(err)
	}

	fmt.Printf("%+v\n", a)
}
```

<details>
  <summary>go.mod</summary>

```text
module modname

go 1.24.0

require (
   // ...
)
```

</details>

## Expected behavior

When using Row.ToStruct(), and the destination struct has a field of type *T where *T implements spanner.Decoder, DecodeSpanner() should be invoked and the field should be populated accordingly.

## Actual behavior

The custom DecodeSpanner() method is not called when the destination field is a pointer.
This appears to be because a **T value is passed down during decoding, which fails the spanner.Decoder interface check.
As a result, the pointer remains nil.

https://github.com/googleapis/google-cloud-go/blob/2d66d4f5b26a488b015be82733b242ce611c0fe3/spanner/value.go#L4436

## Screenshots

n/a

## Additional context

Related internal code:
https://github.com/googleapis/google-cloud-go/blob/2d66d4f5b26a488b015be82733b242ce611c0fe3/spanner/value.go#L4436

Workarounds considered:

Changing *B to B → loses nullability

Manual decoding with GenericColumnValue → verbose and breaks struct mapping

Please let me know if there's a recommended idiomatic pattern for this case.

0daryo:
Decode double pointer custom struct.
FIX #12495

### DIFF
--- a/spanner/row_test.go
+++ b/spanner/row_test.go
@@ -2279,6 +2279,7 @@ func TestSelectAll(t *testing.T) {
 		Col1 int64
 		COL2 float64
 		Col3 CustomType[string]
+		Col4 *CustomType[string]
 	}
 
 	type testStructWithTag struct {
@@ -2451,7 +2452,7 @@ func TestSelectAll(t *testing.T) {
 			},
 		},
 		{
-			name: "success: using destination with custom type with custom decoder with some null columns",
+			name: "success: using destination with custom type with custom decoder with some null columns and pointer decoder",
 			args: args{
 				destination: &[]*testStructWithCustom{},
 				mock: newMockIterator(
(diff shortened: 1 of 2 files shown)


rahul2393 (dismissed):

rahul2393:
@0daryo Can you please rebase your branch to latest main?

rahul2393:
@0daryo Can you please rebase your changes on latest main?

0daryo:
@rahul2393
> Can you please rebase your changes on latest main?

Thanks! I rebased it.

olavloite (approved):

PR #12506, opened 2025-06-26
refactor(storage): remove unnecesary field from internal MRD struct

BrennaEpp:
The MRD doesn't need to know the full object size; we only use it for returning attrs to the user. 

This removes the extra/unnecessary use of a mutex.

### DIFF
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -1160,8 +1160,6 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params 
 	// object metadata.
 	msg := resp.response
 	obj := msg.GetMetadata()
-	// This is the size of the entire object, even if only a range was requested.
-	size := obj.GetSize()
 
 	mrd := &gRPCBidiReader{
 		stream:           resp.stream,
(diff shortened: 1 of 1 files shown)


raj-prince (approved):

storage/grpc_client.go
```diff
@@ -1359,16 +1357,12 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
 		mrd.mu.Unlock()
 	}
 
-	mrd.mu.Lock()
```
danielduhh, line 1362: IIUC, this is checking for a deadlock. Why don’t we need it anymore?

BrennaEpp: It was added as a precaution on any uses of any field in the `gRPCBidiReader` struct, because that struct is accessed/modified concurrently. 

However, the field this mutex is protecting here, `mrd.objectSize`, is only ever used to create the `Attrs` struct we return to the user. So, I removed the `objectSize` field from the `gRPCBidiReader` struct, and I moved `obj.GetSize()` to where it is used. Similar to the other field of `Attrs`, it does not need concurrency protection as we return it immediately and never use it again.

PR #12509, opened 2025-06-26
chore(pubsub/v2): carve out v2 module

hongalex:


shollyman (approved):
Generally LGTM.  You'll want to release the postprocessor after the change to releaseplease.go, as well as setting up the release please configs themselves for individually-released modules.

PR #12512, opened 2025-06-27
feat(bigquery/v2): query client based on bigquery/v2 package

alvarowolfx:
Draft on new query experience using the new bigquery/v2 client. Will break it down into smaller PR down the line, just testing the interface across languages to settle on a good interface.


bigquery/v2/query/client.go
```diff
@@ -0,0 +1,113 @@
...
+	"cloud.google.com/go/bigquery/storage/apiv1/storagepb"
+	"cloud.google.com/go/bigquery/v2/apiv2/bigquerypb"
+	"cloud.google.com/go/bigquery/v2/apiv2_client"
+	"github.com/googleapis/gax-go/v2"
+	"google.golang.org/api/option"
+)
+
+// Client is a client for running queries in BigQuery.
+type Client struct {
+	c                      *apiv2_client.Client
```
shollyman, line 31: nit: just JobClient so you can choose a single client or the monoclient.

bigquery/v2/query/client.go
```diff
@@ -0,0 +1,113 @@
...
+	rc                     *storagepb.BigQueryReadClient
+	projectID              string
+	billingProjectID       string
+	defaultJobCreationMode bigquerypb.QueryRequest_JobCreationMode
```
shollyman, line 35: To consider: defaultPostRequest rather than a single field, and use proto merge semantics to pick up defaults.

bigquery/v2/query/query.go
```diff
@@ -0,0 +1,183 @@
...
+func newQueryJobFromJobReference(c *Client, schema *bigquerypb.TableSchema, jobRef *bigquerypb.JobReference) (*Query, error) {
+	res := &bigquerypb.QueryResponse{
+		Schema:       schema,
+		JobReference: jobRef,
+	}
+	return newQueryJobFromQueryResponse(c, res)
+}
+
+// state is one of a sequence of states that a Job progresses through as it is processed.
+type state = string
```
shollyman: I really wish we had a v2 enum to use for this.  Since we don't, do we need it at all?

bigquery/v2/query/query.go
```diff
@@ -0,0 +1,183 @@
...
+const (
+	// Pending is a state that describes that the job is pending.
+	Pending state = "PENDING"
+	// Running is a state that describes that the job is running.
+	Running state = "RUNNING"
+	// Done is a state that describes that the job is done.
+	Done state = "DONE"
+)
+
+func (q *Query) checkStatus(ctx context.Context) error {
```
shollyman: this fn name confuses me a bit.  The implementation is a bit wonky since it explicitly asks for no rows but appears to have some row cache logic embedded via consumeQueryResponse.

bigquery/v2/query/query.go
```diff
@@ -0,0 +1,183 @@
...
+		case <-time.After(1 * time.Second): // TODO: exponetial backoff
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	}
+	return nil
+}
+
+// JobReference returns the job reference.
+func (q *Query) JobReference() *bigquerypb.JobReference {
```
shollyman, line 210: Why isn't JobReference just a member of the object?  What about stateless queries?

bigquery/v2/query/reader.go
```diff
@@ -0,0 +1,81 @@
...
+package query
+
+import (
+	"context"
+
+	"cloud.google.com/go/bigquery/storage/apiv1/storagepb"
+	"cloud.google.com/go/bigquery/v2/apiv2/bigquerypb"
+)
+
+// Reader is used to read the results of a query.
```
shollyman: we have a Query, a Reader, and a RowIterator?  This feels like an overlap at a high level, but we can dig in when we talk next.

bigquery/v2/query/client.go
alvarowolfx: when using the Storage integration, the table service is needed. It's used for fetching table schema, which in some cases is not cached, like if the user uses the `AttachJob` method

bigquery/v2/query/client.go
alvarowolfx: good point, still need to work on this one.

bigquery/v2/query/reader.go
alvarowolfx: just Query and RowIterators are going to be public, pushed some changes to outline that. `Reader` is an interface for reading data using `jobs.query` or using Storage Read API.

bigquery/v2/query/query.go
alvarowolfx: good point, I can make it a member, but what happens if uses fiddle with it and set it to `nil` somehow ? And for stateless queries, the job reference is build using the `QueryID`. I need to double check if is working as intended in that part, like trying to call `AttachJob` to a stateless query.

bigquery/v2/query/query.go
alvarowolfx: I removed that enum, is not needed

bigquery/v2/query/query.go
alvarowolfx: I was going to reuse the `consumeQueryResponse` method on the `newQueryJobFromQueryResponse` method.

shollyman:
putting this one to bed for the time being.

PR #12516, opened 2025-06-30
feat(bigtable): Adding support to schema bundles

ron-gal:


bigtable/proto_schema_bundle.pb
```diff
@@ -0,0 +1,6 @@
+
+q
```
bhshkh: Move this file to testdata directory.

bigtable/proto_schema_bundle.proto
```diff
@@ -0,0 +1,22 @@
+/*
```
bhshkh: Move this file to testdata directory.

bigtable/updated_proto_schema_bundle.pb
```diff
@@ -0,0 +1,7 @@
+
```
bhshkh: Move this file to testdata directory.

bigtable/updated_proto_schema_bundle.proto
```diff
@@ -0,0 +1,23 @@
+/*
```
bhshkh: Move this file to testdata directory.

bigtable/admin.go
```diff
@@ -3421,3 +3426,158 @@ func (iac *InstanceAdminClient) DeleteMaterializedView(ctx context.Context, inst
 	_, err := iac.iClient.DeleteMaterializedView(ctx, req)
 	return err
 }
+
+// SchemaBundleConf contains information about a schema bundle.
+type SchemaBundleConf struct {
+	TableID        string
+	SchemaBundleID string
+	ProtoFile      string
+}
```
bhshkh:
```suggestion
// SchemaBundleConf contains the information necessary to create or update a schema bundle.
type SchemaBundleConf struct {
	TableID        string
	ProtoSchema *ProtoSchemaInfo
	
	// Etag is used for optimistic concurrency control during updates.
	// Ignored during creation.
	Etag string
}

// ProtoSchemaInfo represents a protobuf schema.
type ProtoSchemaInfo struct {
	// Contains a protobuf-serialized
	// [google.protobuf.FileDescriptorSet](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto),
	// which could include multiple proto files.
	ProtoDescriptors []byte
}

```

1. The application may not have the necessary permissions to read the file or the file may be too large to be read into memory using os.ReadFile. So, use []byte slice instead and leave the worries of obtaining the byte slice upto the user.
2. Currently, only 1 schema bundle type is supported. In the future, more types could be introduced. The conf should support that.
3. SchemaBundle_ProtoSchema has only 1 field now. More fields could be added to the proto. So, it should be separate struct.

bigtable/admin.go
```diff
@@ -3421,3 +3426,158 @@ func (iac *InstanceAdminClient) DeleteMaterializedView(ctx context.Context, inst
...
+
+// SchemaBundles
+
+// CreateSchemaBundle creates a new schema bundle in a table.
+func (ac *AdminClient) CreateSchemaBundle(ctx context.Context, conf *SchemaBundleConf) error {
+	if conf.TableID == "" || conf.SchemaBundleID == "" {
+		return errors.New("both SchemaBundleID and TableID are required")
```
bhshkh:
```suggestion
		return errors.New("both SchemaBundleID and TableID are required in SchemaBundleConf")
```

nit

bigtable/admin.go
```diff
@@ -3421,3 +3426,158 @@ func (iac *InstanceAdminClient) DeleteMaterializedView(ctx context.Context, inst
...
+		Name: ac.schemaBundlePath(tableID, schemaBundleID),
+	}
+	var res *btapb.SchemaBundle
+
+	err := gax.Invoke(ctx, func(ctx context.Context, _ gax.CallSettings) error {
+		var err error
+		res, err = ac.tClient.GetSchemaBundle(ctx, req)
+		return err
+	}, adminRetryOptions...)
+
```
bhshkh: nit: Remove empty line 3488

bigtable/admin.go
```diff
@@ -3421,3 +3426,158 @@ func (iac *InstanceAdminClient) DeleteMaterializedView(ctx context.Context, inst
...
+	return err
+}
+
+// SchemaBundleInfo contains schema bundle metadata. This struct is read-only.
+type SchemaBundleInfo struct {
+	TableID        string
+	SchemaBundleID string
+
+	SchemaBundle []byte
+}
```
bhshkh: Add `Etag string`


Change `// SchemaBundleInfo contains schema bundle metadata. This struct is read-only.` to 

```
// SchemaBundleInfo represents information about a schema bundle. Schema bundle is a named collection of related schemas.
// This struct is read-only.
```

bigtable/admin.go
```diff
@@ -3421,3 +3426,158 @@ func (iac *InstanceAdminClient) DeleteMaterializedView(ctx context.Context, inst
...
+	}
+	if len(res.GetProtoSchema().GetProtoDescriptors()) > 0 {
+		sb.SchemaBundle = res.GetProtoSchema().GetProtoDescriptors()
+	}
+
+	return sb, nil
+}
+
+// SchemaBundles returns a list of the schema bundles in the table.
+func (ac *AdminClient) SchemaBundles(ctx context.Context, tableID string) ([]string, error) {
```
bhshkh: Why do we want to return only the names and not an infoiterator?
Noticed the same in AuthorizedViews and Tables as well. 

Maybe it makes sense in case of Tables since there are a lot of fields and nested structs. Do you anticipate the same happening here? In the current schema bundle proto, there are very few fields similar to AppProfile.

slice of names should be preferred:
- If you anticipate that tables will only ever have a few schema bundles:
   In that case, the memory overhead of a slice is negligible, and the directness of a slice might be more convenient than managing an iterator. 
(Since the underlying response has a pagination token, I'm thinking that the amount may be large.)
- If the typical use case is to get all schema bundles and then operate on the complete list, an iterator adds a small amount of boilerplate (the loop to drain it).

bigtable/admin.go
```diff
@@ -3421,3 +3426,158 @@ func (iac *InstanceAdminClient) DeleteMaterializedView(ctx context.Context, inst
...
+	}
+
+	ctx = mergeOutgoingMetadata(ctx, ac.md)
+	req := &btapb.CreateSchemaBundleRequest{
+		Parent:         fmt.Sprintf("%s/tables/%s", ac.instancePrefix(), conf.TableID),
+		SchemaBundleId: conf.SchemaBundleID,
+		SchemaBundle:   schemaBundle,
+	}
+	_, err := ac.tClient.CreateSchemaBundle(ctx, req)
+	return err
```
bhshkh: Should we wait here for operation to complete ? Similar to update ?

bigtable/admin.go
```diff
@@ -3421,3 +3426,158 @@ func (iac *InstanceAdminClient) DeleteMaterializedView(ctx context.Context, inst
...
+}
+
+// UpdateSchemaBundleConf contains all the information necessary to update or partial update a schema bundle.
+type UpdateSchemaBundleConf struct {
+	SchemaBundleConf SchemaBundleConf
+	IgnoreWarnings   bool
+}
+
+// UpdateSchemaBundle updates a schema bundle in a table according to the given configuration.
+func (ac *AdminClient) UpdateSchemaBundle(ctx context.Context, conf UpdateSchemaBundleConf) error {
```
bhshkh, line 3556: What happens when there are multiple fields that can be updated? How will UpdateSchemaBundleFromConf know which fields from SchemaBundleConf need to be included in update request? How will it differentiate between null and unset ?

bhshkh (changes requested):

bigtable/admin.go
ron-gal: Done

bigtable/admin.go
ron-gal: Done

bigtable/admin.go
ron-gal: Done

bigtable/proto_schema_bundle.pb
ron-gal: Done

bigtable/proto_schema_bundle.proto
ron-gal: Done

bigtable/updated_proto_schema_bundle.pb
ron-gal: Done

bigtable/updated_proto_schema_bundle.proto
ron-gal: Done

bigtable/admin.go
ron-gal: Done

bigtable/admin.go
ron-gal: If file\slice is empty, we don't need to update it (currently it's kinda weird, as it's the only field that can change). In the future if we have more we will need to make sense of each field - what is the best way to know it should not be updated. If push comes to shove, we can allow the future user to set the update field mask, or maybe make distinct functions per field (like we have for table: UpdateTableWithChangeStream, UpdateTableWithDeletionProtection, UpdateTableDisableAutomatedBackupPolicy...)

bigtable/admin.go
ron-gal: Done

PR #12534, opened 2025-07-08
chore(internal/postprocessor): try adding PATH explicitly to all cmds

codyoss:
We used to have to do this in the gapicgen tool for git commands. Let's just set these for all commands to see if it makes a difference. Example: https://github.com/googleapis/google-cloud-go/blob/7a3d016b4fb7cd7e855ccb61aca65973cd5ac808/internal/gapicgen/git/github.go#L104-L105

b/428197918


internal/postprocessor/execv/command.go
```diff
@@ -58,21 +61,3 @@ func (c *CmdWrapper) Output() ([]byte, error) {
 	}
 	return b, err
 }
-
-// ForEachMod runs the given function with the directory of
-// every non-internal module.
-func ForEachMod(rootDir string, fn func(dir string) error) error {
```
quartzmo, line 64: Unused code I assume.

codyoss: yup!

PR #12542, opened 2025-07-09
perf(storage): zero copy for MultiRangeDownloader

tritone:
Use the zero-copy codec to parse response messages in the MultiRangeDownloader (in addition to Reader).


storage/grpc_client.go
```diff
@@ -1135,19 +1140,27 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
 			if err != nil {
 				return err
 			}
-			resp, err = stream.Recv()
+			// Use RecvMsg to get the raw buffer slice instead of Recv().
+			databufs := mem.BufferSlice{}
+			err = stream.RecvMsg(&databufs)
 			if err != nil {
```
cojenco, line 1143: Do we want to handle `codes.NotFound` and wrap `formatObjectErr` here, similar to [NewRangeReader](https://github.com/googleapis/google-cloud-go/blob/c29ceaa10e7954851ba220b89df2b594ba102198/storage/grpc_client.go#L1607-L1611)?

I'm reviewing this alongside with the zero copy implementation in NewRangeReader. There are comments around handling these types of errors that show up on the RecvMsg call

storage/grpc_client.go
```diff
@@ -1135,19 +1140,27 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
...
+			decoder = &readResponseDecoder{
+				databufs: databufs,
+			}
+			err = decoder.readFullObjectResponse()
+			return err
 		}, s.retry, s.idempotent)
 		if err != nil {
 			// Close the stream context we just created to ensure we don't leak
 			// resources.
 			cancel()
```
cojenco, line 1157: Do we need to free the buffers here? Or is it that we only want to free the buffers in the receiver loop until fully processed? I do see that the buffers are freed on L1345

storage/grpc_client.go
```diff
@@ -1259,37 +1271,54 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
...
-				if resp.GetReadHandle().GetHandle() != nil {
-					mrd.readHandle = resp.GetReadHandle().GetHandle()
-				}
+				databufs := mem.BufferSlice{}
+				err = mrd.stream.RecvMsg(&databufs)
+				// TODO: update read handle?
 				if err == io.EOF {
 					err = nil
 				}
+				// TODO: handle not found error?
```
cojenco: Ah I see your TODO noted here. Will these be addressed in a future PR?

storage/grpc_client.go
```diff
@@ -2152,69 +2202,99 @@ func (d *readResponseDecoder) advanceOffset(n uint64) error {
 // This copies object data from the message into the buffer and returns the number of
 // bytes copied. The data offsets are incremented in the message. The updateCRC
 // function is called on the copied bytes.
-func (d *readResponseDecoder) readAndUpdateCRC(p []byte, updateCRC func([]byte)) int {
-	// For a completely empty message, just return 0
+func (d *readResponseDecoder) readAndUpdateCRC(p []byte, readID int64, updateCRC func([]byte)) (n int, found bool) {
+	// For a completely empty message, just return 0.
 	if len(d.databufs) == 0 {
-		return 0
+		return 0, true
```
cojenco: nit: I find it a bit weird for an empty message to return found as true. I guess this is a case we want to return when being called in Read? Maybe add a comment to clarify?

storage/grpc_client.go
```diff
@@ -2152,69 +2202,99 @@ func (d *readResponseDecoder) advanceOffset(n uint64) error {
...
+	}
+	offsets.currOff += uint64(n)
 
-	// We've read all the data from this message. Free the underlying buffers.
-	if d.dataOffsets.currBuf == d.dataOffsets.endBuf && d.dataOffsets.currOff == d.dataOffsets.endOff {
-		d.done = true
-		d.databufs.Free()
+	// We've read all the data for this specific range from this message.
+	if offsets.currBuf == offsets.endBuf && offsets.currOff == offsets.endOff {
+		d.done = true // Mark as done for this read, though the overall message might have more data.
```
cojenco, line 2253: curious if this will cause any issues since d.dataOffsets now is a map of multiple ranges

storage/grpc_client.go
tritone: Because of zero-copy, we have to wait until all the data is written out from the message to the user application before freeing.

storage/grpc_client.go
```diff
@@ -1253,20 +1264,32 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
 			default:
 				// This function reads the data sent for a particular range request and has a callback
 				// to indicate that output buffer is filled.
-				resp, err = mrd.stream.Recv()
-				if resp.GetReadHandle().GetHandle() != nil {
-					mrd.readHandle = resp.GetReadHandle().GetHandle()
-				}
+				databufs := mem.BufferSlice{}
+				err = mrd.stream.RecvMsg(&databufs)
+				// TODO: update read handle?
```
BrennaEpp: This TODO seems to be done below?

storage/grpc_client.go
```diff
@@ -1286,9 +1308,15 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
...
+
+							// The decoder holds the object content. writeToAndUpdateCRC writes
+							// it to the user's buffer without an intermediate copy.
+							written, _, err := decoder.writeToAndUpdateCRC(currRange.writer, id, func(b []byte) {
+								// crc update logic can be added here if needed
+							})
+
 							if err != nil {
-								currRange.callback(currRange.offset, currRange.totalBytesWritten, err)
+								mrd.activeRanges[id].callback(currRange.offset, currRange.totalBytesWritten, err)
```
BrennaEpp: Why not leave this as `currRange`?

storage/grpc_client.go
```diff
@@ -1253,20 +1264,32 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
...
 					err = nil
 				}
+				// TODO: handle not found error?
 				if err != nil {
 					// cancel stream and reopen the stream again.
 					// Incase again an error is thrown close the streamManager goroutine.
 					mrd.retrier(err, "receiver")
 				}
 
 				if err == nil {
```
BrennaEpp, line 1282: Can we remove this condition or make it an else if?

storage/grpc_client.go
BrennaEpp: I think it's okay to handle this in a future PR - it was also not converted to the custom error before this.

storage/grpc_client.go
```diff
@@ -1442,7 +1472,41 @@ func (mrd *gRPCBidiReader) reopenStream(failSpec []mrdRange) error {
...
+		id := val.GetReadRange().GetReadId()
+		mrd.mu.Lock()
+		activeRange, ok := mrd.activeRanges[id]
+		if !ok {
+			mrd.mu.Unlock()
+			continue
+		}
+
+		// Use the decoder's zero-copy write method.
+		written, _, writeErr := res.decoder.writeToAndUpdateCRC(activeRange.writer, id, nil)
```
BrennaEpp, line 1494: Should we free the databufs below after using them here?

storage/grpc_client.go
```diff
@@ -2397,7 +2473,15 @@ func (d *readResponseDecoder) readFullObjectResponse() error {
 		case fieldNum == objectRangeDataField && fieldType == protowire.BytesType:
 			// The object data field was found. Initialize the data ranges assuming
 			// exactly one range in the message.
-			msg.ObjectDataRanges = []*storagepb.ObjectRangeData{{ChecksummedData: &storagepb.ChecksummedData{}, ReadRange: &storagepb.ReadRange{}}}
+			// Create a new ObjectRangeData for each instance of this repeated field.
+			newRangeData := &storagepb.ObjectRangeData{ChecksummedData: &storagepb.ChecksummedData{}, ReadRange: &storagepb.ReadRange{}}
+			msg.ObjectDataRanges = append(msg.ObjectDataRanges, newRangeData)
+			// Get a reference to the newly added element to populate it.
+			currentRange := msg.ObjectDataRanges[len(msg.ObjectDataRanges)-1]
```
BrennaEpp: nit: Doesn't `newRangeData` already have a pointer to the current range?

storage/grpc_client.go
```diff
@@ -1975,47 +2030,38 @@ func (r *gRPCReader) WriteTo(w io.Writer) (int64, error) {
...
-			r.updateCRC(b)
-		})
-		r.seen += int64(written)
-		r.currMsg = nil
-		if err != nil {
-			return r.seen - alreadySeen, err
+	for {
+		// Write any data from the current message buffer.
+		if r.currMsg != nil && !r.currMsg.done {
+			written, _, err := r.currMsg.writeToAndUpdateCRC(w, 1, func(b []byte) {
```
BrennaEpp: Maybe we should mention in a comment that we set readID to 1 always in the case of NewRangeReader. (AI reviewer was concerned about this so I double-checked).

storage/grpc_client.go
```diff
@@ -1975,47 +2030,38 @@ func (r *gRPCReader) WriteTo(w io.Writer) (int64, error) {
...
 		// Attempt to receive the next message on the stream.
-		// Will terminate with io.EOF once data has all come through.
-		// recv() handles stream reopening and retry logic so no need for retries here.
 		err := r.recv()
 		if err != nil {
 			if err == io.EOF {
-				// We are done; check the checksum if necessary and return.
+				// We are done; check the checksum if necessary and break the loop.
 				err = r.runCRCCheck()
+				break
```
BrennaEpp: Won't the r.runCRCCheck run twice in this case now (here and L2061)?

storage/grpc_client.go
```diff
@@ -2171,69 +2220,94 @@ func (d *readResponseDecoder) advanceOffset(n uint64) error {
...
+			updateCRC(dataSlice[:n])
+		}
+		if err != nil {
+			// Return immediately on a write error.
+			return totalWritten, true, err
 		}
 	}
-	return written, nil
+
+	return totalWritten, true, nil
```
BrennaEpp, line 2312: I think we still need to set `d.done = true`?

storage/grpc_client.go
```diff
@@ -2084,9 +2131,11 @@ type readResponseDecoder struct {
 	currBuf int    // index of the current buffer being processed
 	currOff uint64 // offset in the current buffer
 	// Processed data
-	msg         *storagepb.BidiReadObjectResponse // processed response message with all fields other than object data populated
-	dataOffsets bufferSliceOffsets                // offsets of the object data in the message.
-	done        bool                              // true if the data has been completely read.
+	msg *storagepb.BidiReadObjectResponse // processed response message with all fields other than object data populated
+	// The original dataOffsets field is replaced with a map to handle multiple ranges.
+	// dataOffsets bufferSliceOffsets
+	dataOffsets map[int64]bufferSliceOffsets // New field: map ReadId to its data offsets
```
BrennaEpp: This is useful for the PR but perhaps confusing for posterity - suggest to remove references to the new vs old field

BrennaEpp (commented):
Thank you! Left some comments, mostly nits

storage/grpc_client.go
tritone: I switched to using a const for this for clarity.

storage/grpc_client.go
tritone: good catch, removed this.

storage/grpc_client.go
tritone: Done.

storage/grpc_client.go
tritone: True, fixed this up.

storage/grpc_client.go
tritone: Done

storage/grpc_client.go
tritone: I think I just missed this when fixing the merge conflict with another PR. fixed.

storage/grpc_client.go
tritone: Removed this.

storage/grpc_client.go
tritone: Good catch, added a free here as well.

storage/grpc_client.go
tritone: I'll file a separate issue for this.

storage/grpc_client.go
tritone: I think actually it would be fine to receive the next message in this case... fixed.

storage/grpc_client.go
tritone: fixed.

storage/grpc_client.go
tritone: I think it is fine but will add a test for this separately.

BrennaEpp (approved):

PR #12548, opened 2025-07-10
fix(storage): fix MultiRangeDownloader deadlock

tritone:
This fixes some potential deadlocks in MultiRangeDownloader which were caused by not releasing a lock on early return, or holding it when it wasn't needed (when sending on a chan).

### DIFF
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -1284,8 +1284,9 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params 
 					arr := resp.GetObjectDataRanges()
 					for _, val := range arr {
 						id := val.GetReadRange().GetReadId()
-						mrd.mu.Lock()
 						func() {
+							mrd.mu.Lock()
+							defer mrd.mu.Unlock()
 							currRange, ok := mrd.activeRanges[id]
 							if !ok {
 								// it's ok to ignore responses for read_id not in map as user would have been notified by callback.
@@ -1314,8 +1315,6 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params 
 								delete(mrd.activeRanges, id)
 							}
 						}()
-
-						mrd.mu.Unlock()
 					}
 				}
 			}
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -1277,6 +1277,7 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params 
 					if len(mrd.activeRanges) == 0 && mrd.numActiveRanges == 0 {
 						mrd.closeReceiver <- true
 						mrd.closeSender <- true
+						mrd.mu.Unlock()
 						return
 					}
 					mrd.mu.Unlock()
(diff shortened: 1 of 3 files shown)


storage/grpc_client.go
```diff
@@ -1287,6 +1288,7 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
 						_, ok := mrd.activeRanges[id]
 						if !ok {
 							// it's ok to ignore responses for read_id not in map as user would have been notified by callback.
+							mrd.mu.Unlock()
```
cjc25: This particular one is pretty complex - it took me a minute to find the other unlock.

Should we move the locked body into an anonymous function so we can use `defer mrd.mu.Unlock()`?

abhishek10004 (approved):
LGTM

storage/grpc_client.go
```diff
@@ -1287,6 +1288,7 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
 						_, ok := mrd.activeRanges[id]
```
cjc25: (I know it was already like this)

We sure do make a lot of calls to `mrd.activeRanges[id]`.  Since we're under a lock I think `&mrd.activeRanges[id]` is stable, which maybe will make this easier to read?

storage/grpc_client.go
```diff
@@ -1277,6 +1277,7 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
 					if len(mrd.activeRanges) == 0 && mrd.numActiveRanges == 0 {
 						mrd.closeReceiver <- true
 						mrd.closeSender <- true
+						mrd.mu.Unlock()
```
cjc25: Why do these channel writes have to be under the lock? Can we add a comment?

cjc25 (approved):

storage/grpc_client.go
tritone: good point, fixed.

storage/grpc_client.go
tritone: Actually looks like they don't have to from the other use in close(). I will release the lock first.

storage/grpc_client.go
tritone: Good idea, done.

ddelgrosso1 (approved):

storage/grpc_client.go
```diff
@@ -1284,32 +1285,35 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params
...
-						} else {
-							mrd.activeRanges[id] = mrdRange{
-								readID:              mrd.activeRanges[id].readID,
-								writer:              mrd.activeRanges[id].writer,
-								offset:              mrd.activeRanges[id].offset,
-								limit:               mrd.activeRanges[id].limit,
-								currentBytesWritten: mrd.activeRanges[id].currentBytesWritten + int64(len(val.GetChecksummedData().GetContent())),
-								totalBytesWritten:   mrd.activeRanges[id].totalBytesWritten + int64(len(val.GetChecksummedData().GetContent())),
-								callback:            mrd.activeRanges[id].callback,
+						func() {
```
abhishek10004, line 1287: Since you've refactored it to an anonymous function, can we take the lock inside it and put a defer statement to unlock it right after that (making it easier to see the unlock)?

Eg:
func() {
  lock()
  defer unlock()
  ...


}

ddelgrosso1 (approved):

PR #12510, opened 2025-06-27
feat(storage): Add AllowAllServiceAgentAccess to IpFilter

madisonhall38:
fixes #12472

PR #12549, opened 2025-07-10
revert(storage): add IpFilter to bucketAttrs

danielduhh:
Reverts googleapis/google-cloud-go#12020

Context: there's additional work that needs to happen (https://github.com/googleapis/google-cloud-go/pull/12510) to finish the implementation (e.g. grpc). We will revert this now to unblock existing releases and someone can pick this back up in the future when we have bandwidth


tritone (approved):
FYI @codyoss @quartzmo : this is a breaking change, but the removed exported types have not yet been released yet in cloud.google.com/go/storage.

PR #12560, opened 2025-07-14
chore: add pubsub/v2 to release please json

hongalex:
Need this to stop release please from generating this PR: https://github.com/googleapis/google-cloud-go/pull/12559

Following Seth's suggestion from BigQuery https://github.com/googleapis/google-cloud-go/pull/12465/files#diff-c0a64cf5a57a6a28a748210792be3a1704092023edf4016e4a95f14aa7849b40 and then fixing the release please version in different commit

### DIFF
--- a/.release-please-manifest-individual.json
+++ b/.release-please-manifest-individual.json
@@ -10,7 +10,7 @@
   "logging": "1.13.0",
   "profiler": "0.4.3",
   "pubsub": "1.49.0",
-  "pubsub/v2": "2.0.0",
+  "pubsub/v2": "0.0.0",
   "pubsublite": "1.8.2",
   "spanner": "1.83.0",
   "storage": "1.55.0",


shollyman (approved):

.release-please-manifest-individual.json
```diff
@@ -10,6 +10,7 @@
   "logging": "1.13.0",
   "profiler": "0.4.3",
   "pubsub": "1.49.0",
+  "pubsub/v2": "0.0.0",
```
codyoss, line 13: Do you want to set a value here, or you doing that in a future pr?

codyoss (approved):

.release-please-manifest-individual.json
hongalex: Yeah we're setting to a null value initially to use release-please tagging to fix a specific version later.

PR #12567, opened 2025-07-15
feat(bigtable): Support universe domain

bhshkh:
Fixes: b/432032746


bigtable/bigtable.go
```diff
@@ -49,7 +49,7 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-const prodAddr = "bigtable.googleapis.com:443"
+const prodAddr = "bigtable.UNIVERSE_DOMAIN:443"
```
daniel-sanche, line 53: how does this work? I assume `WithDefaultUniverseDomain` treats this as a template. (Is `WithDefaultUniverseDomain` part of gax?)

Maybe a comment here would be helpful

bhshkh: WithDefaultUniverseDomain isn't part of gax. Its here: https://github.com/googleapis/google-api-go-client/blob/a269dca39e5315e9bfc77cc05b3dbd64af9baa25/option/internaloption/internaloption.go#L156-L164


While creating client, user can do :
```go
universeDomain := "my_domain"
client, err := bigtable.NewClient(ctx, project, instance, option.WithUniverseDomain(universeDomain)
```

or 

```
export GOOGLE_CLOUD_UNIVERSE_DOMAIN="my_domain"
```
```go
client, err := bigtable.NewClient(ctx, project, instance,)
```
This is where the replace happens: 

https://github.com/googleapis/google-api-go-client/blob/a269dca39e5315e9bfc77cc05b3dbd64af9baa25/internal/cba.go#L193-L201

https://github.com/googleapis/google-api-go-client/blob/a269dca39e5315e9bfc77cc05b3dbd64af9baa25/internal/settings.go#L193-L207

bigtable/admin.go
```diff
@@ -1166,7 +1167,8 @@ func (ac *AdminClient) AuthorizedViewIAM(table, authorizedView string) *iam.Hand
 	return iam.InternalNewHandleGRPCClient(ac.tClient, ac.authorizedViewPath(table, authorizedView))
 }
 
-const instanceAdminAddr = "bigtableadmin.googleapis.com:443"
+// UNIVERSE_DOMAIN placeholder is replaced by the UniverseDomain from DialSettings while creating GRPC connection/dial pool.
+const instanceAdminAddr = "bigtableadmin.UNIVERSE_DOMAIN:443"
```
hongalex, line 1171: Does this need to be a different variable than `adminAddr`?

Issue #12409, opened 2025-06-05
firestore: support REST api


erlichmen:
We need to invoke Firestore through the REST API. Currently, REST support is buried inside the internal client (via `NewRESTClient()`) and is not properly exposed externally. Even though the internal client API is technically accessible, we can't use it effectively because it lacks proper `projectId` handling and exposes a different interface compared to the one provided by `NewClient()`.

I propose introducing a new public `NewRESTClient()` API that wraps the internal client’s `NewRESTClient()` implementation. Alternatively, we could introduce an option in `NewClient()` to control whether it uses REST or gRPC.

At the moment, we're forced to clone the Firestore client code and implement `NewRESTClient()` ourselves—but this is not a sustainable long-term solution.

Other client libraries (e.g., Node.js, Python) already support REST access, so achieving parity here is reasonable and beneficial.

PR #12575, opened 2025-07-16
feat(firestore): allow creating REST clients

lauraseidler:
This PR adds a new public function `NewRESTClient` to allow creating a client that uses REST. Since the available `opts` apply to the client but cannot change the client in the current implementation, this requires a new function rather than a new option.

To avoid `NewClient` and `NewRESTClient` drifting, a new internal function `newClient` is used to create the client based on a `createClient` function parameter, rather than the previously hardcoded `vkit.NewClient`. 

`NewRESTClient` is currently configured explicitly without support for the emulator, as the emulator is configured to use a GRPC connection, which clashes with using a REST client. If `FIRESTORE_EMULATOR_HOST` is set when using `NewRESTClient`, the call will return an error. 

fixes https://github.com/googleapis/google-cloud-go/issues/12409


erlichmen:
bump

hongalex (approved):

PR #12590, opened 2025-07-18
fix(pubsub/v2): acquire bounded limits in flow controller

### LINKED ISSUE #12447: pubsub: outstanding bytes and outstanding messages metrics are 0 when FlowControlIgnore is set
## Client

PubSub

## Environment

N/A

## Code and Dependencies

N/A

## Expected behavior

I expect under the default settings that outstanding messages and outstanding bytes are non-zero if the maximums are set even if `FlowControlIgnore` is set.

## Actual behavior

The flow controller has [return nil](https://github.com/googleapis/google-cloud-go/blob/770daf7b859c2efc307a38ecb680c2e8d901d342/pubsub/flow_controller.go#L129) instead of an empty block which skips the later `recordOutstandingMessages` and `recordOutstandingBytes` blocks.

## Screenshots

e.g. A chart showing how messages are slow. Delete if not necessary.

## Additional context

It would be useful for us to keep the default ignore behavior and monitor the outstanding messages and bytes before enabling erroring. We'd like to know what reasonable maximums we should set.

hongalex:
This change makes flow controller continue to track and record outstanding messages/bytes even when flow control mechanisms are disabled.

Fixes #12447

### DIFF
--- a/pubsub/v2/flow_controller.go
+++ b/pubsub/v2/flow_controller.go
@@ -147,7 +147,7 @@ func (f *flowController) acquire(ctx context.Context, size int) error {
 		}
 		if f.semSize != nil {
 			// Try to acquire the full size of the message here.
-			if !f.semSize.TryAcquire(int64(size)) {
+			if !f.semSize.TryAcquire(f.bound(size)) {
 				if f.semCount != nil {
 					f.semCount.Release(1)
 				}
--- a/pubsub/v2/flow_controller.go
+++ b/pubsub/v2/flow_controller.go
@@ -212,6 +212,7 @@ func (f *flowController) recordOutstandingMessages(ctx context.Context, n int64)
 		return
 	}
 
+	// Otherwise record this as subscriber outstanding messages.
 	recordStat(ctx, OutstandingMessages, n)
 }
 
@@ -221,5 +222,6 @@ func (f *flowController) recordOutstandingBytes(ctx context.Context, n int64) {
 		return
 	}
 
+	// Otherwise record this as subscriber outstanding bytes.
 	recordStat(ctx, OutstandingBytes, n)
 }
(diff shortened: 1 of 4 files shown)


bhshkh (approved):

pubsub/v2/flow_controller.go
```diff
@@ -172,19 +171,19 @@ func (f *flowController) acquire(ctx context.Context, size int) error {
...
+		if f.limitBehavior != FlowControlIgnore {
+			f.semCount.Release(1)
+		}
 	}
 	if f.semSize != nil {
 		outstandingBytes := atomic.AddInt64(&f.bytesRemaining, -1*f.bound(size))
 		f.recordOutstandingBytes(ctx, outstandingBytes)
-		f.semSize.Release(f.bound(size))
+		if f.limitBehavior != FlowControlIgnore {
+			f.semSize.Release(f.bound(size))
```
bhshkh, line 185: A semaphore requires that we release the exact same number of permits that we acquired.

The acquire function has different logic for different LimitExceededBehavior values:
       - In FlowControlBlock mode, it acquires f.bound(size) from the semaphore.
       - In FlowControlSignalError mode, it attempts to acquire the raw int64(size).

The release function always releases f.bound(size). It incorrectly assumes that the amount acquired was also f.bound(size). Could this manifest into a bug?

hongalex: Good catch, this has actually been out for a while. Most likely, it's because the bounding behavior doesn't trigger often (unless users set flow control bytes to very low).

bhshkh (dismissed):

pubsub/v2/flow_controller.go
hongalex: So actually I remember why this was intentional. 

In FlowControlSignalError mode, the `acquire` never goes through since the flow controller throws an error if there is not enough space for the message. In such a case, we bubble up the error `ErrFlowControllerMaxOutstandingBytes`. We shouldn't bound the message size here, since we consider "too large messages" an error.

bhshkh (approved):

PR #12601, opened 2025-07-23
docs(pubsub): add docs comment to MaxOutstandingBytes

### LINKED ISSUE #11568: packagename: short description of bug
## Client
PubSub

## Environment

Google Cloud Run 1 CPU and 512 mb memory using Cloud SQL Postgres with 2 CPU and 4.5 GB ram. 

## Code and Dependencies
```go
      github.com/enadhq/pubsub-contracts v0.36.15
```
MaxOustandingMessages is set to 50 & using StreamingPull
## Expected behavior

To be able to handle a big amount of messages in a pubsub queue

## Actual behavior

During high loads do the cloud run pubsub worker stop processing messages after a short while. As we speak do I have 600k messages in the queue and I think the worker have some issues when there is a lot of messages in the queue. If I restart the cloud run do it continue for a short while.

When I had 100 messages in the queue do it handle it perfectly but it's when it get  a lot of messages in the queue it stops processing.

hongalex:
Fixes #11568

### DIFF
--- a/pubsub/subscription.go
+++ b/pubsub/subscription.go
@@ -926,6 +926,8 @@ type ReceiveSettings struct {
 	// be treated as if it were DefaultReceiveSettings.MaxOutstandingBytes. If
 	// the value is negative, then there will be no limit on the number of bytes
 	// for unprocessed messages.
+	// This defaults to 1e9 or 1 GB. For machines that has less memory available,
+	// it is recommended to decrease this values so as to not run into OOM issues.
 	MaxOutstandingBytes int
 
 	// UseLegacyFlowControl disables enforcing flow control settings at the Cloud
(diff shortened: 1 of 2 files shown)


pubsub/subscription.go
```diff
@@ -926,6 +926,8 @@ type ReceiveSettings struct {
 	// be treated as if it were DefaultReceiveSettings.MaxOutstandingBytes. If
 	// the value is negative, then there will be no limit on the number of bytes
 	// for unprocessed messages.
+	// This defaults to 1e9 or 1 GB. For machines that has less memory available,
```
shollyman: There's odd grammar in these new comments related to plural vs singular terms.  maybe s/has/have/ and s/values/value/ here and below.

shollyman (dismissed):
see feedback re: grammar

shollyman (approved):

Issue #12600, opened 2025-07-23
spanner: data race on `mt.currOp.currAttempt`


egonelbre:
## Client

Spanner v1.83.0 together with github.com/googleapis/go-sql-spanner@v1.11.1-0.20250214171559, although the latter is unlikely to be the cuplrit.

## Environment, Code and Dependencies

It's using `golang:1.24.3` image for running all tests in https://github.com/storj/storj against production Spanner.

<details> <summary>The relevant data race warning, with application code omitted:</summary>

```
WARNING: DATA RACE
Write at 0x00c008a8ddf8 by goroutine 32446:
  cloud.google.com/go/spanner.(*RowIterator).Next.func1()
      /go/pkg/mod/cloud.google.com/go/spanner@v1.83.0/read.go:178 +0x1cb
  runtime.deferreturn()
      /usr/local/go/src/runtime/panic.go:610 +0x5d
  cloud.google.com/go/spanner.(*resumableStreamDecoder).next()
      /go/pkg/mod/cloud.google.com/go/spanner@v1.83.0/read.go:578 +0x472
  cloud.google.com/go/spanner.(*RowIterator).Next()
      /go/pkg/mod/cloud.google.com/go/spanner@v1.83.0/read.go:182 +0x4e4
  github.com/googleapis/go-sql-spanner.(*readOnlyRowIterator).Next()
      /go/pkg/mod/github.com/googleapis/go-sql-spanner@v1.11.1-0.20250214171559-1bccea5dfec5/transaction.go:66 +0x35
  github.com/googleapis/go-sql-spanner.(*rows).Columns.(*rows).getColumns.func1()
      /go/pkg/mod/github.com/googleapis/go-sql-spanner@v1.11.1-0.20250214171559-1bccea5dfec5/rows.go:66 +0x4d
  sync.(*Once).doSlow()
      /usr/local/go/src/sync/once.go:78 +0xe1
  sync.(*Once).Do()
      /usr/local/go/src/sync/once.go:69 +0x44
  github.com/googleapis/go-sql-spanner.(*rows).getColumns()
      /go/pkg/mod/github.com/googleapis/go-sql-spanner@v1.11.1-0.20250214171559-1bccea5dfec5/rows.go:65 +0x6e
  github.com/googleapis/go-sql-spanner.(*rows).Columns()
[...]

PR #12622, opened 2025-07-30
fix(spanner): remove stream wrapper for direct path check

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/12600

This change fixes a race condition that could occur during streaming operations in the Spanner client. The race condition was caused by concurrent access to the metrics tracer (mt) between the main execution goroutine and a background goroutine responsible for consuming stream trailers.

The fix addresses this issue by removing the wrappedStream implementation and moving the logic for updating metrics directly into the grpcSpannerClient methods (ExecuteStreamingSql, StreamingRead, and BatchWrite).

By moving the metrics-related logic out of the stream interceptor and into the gRPC client methods, the race condition is eliminated, as the metrics are now updated in the same goroutine that initiated the RPC.

### DIFF
--- a/spanner/client.go
+++ b/spanner/client.go
@@ -717,16 +717,7 @@ func metricsInterceptor() grpc.UnaryClientInterceptor {
 
 		statusCode, _ := status.FromError(err)
 		mt.currOp.currAttempt.setStatus(statusCode.Code().String())
-
-		isDirectPathUsed := false
-		if peerInfo.Addr != nil {
-			remoteIP := peerInfo.Addr.String()
-			if strings.HasPrefix(remoteIP, directPathIPV4Prefix) || strings.HasPrefix(remoteIP, directPathIPV6Prefix) {
-				isDirectPathUsed = true
-			}
-		}
-
-		mt.currOp.currAttempt.setDirectPathUsed(isDirectPathUsed)
+		mt.currOp.currAttempt.setDirectPathUsed(peer.NewContext(ctx, peerInfo))
 		metrics := parseServerTimingHeader(md)
 		mt.currOp.currAttempt.setServerTimingMetrics(metrics)
 		recordAttemptCompletion(mt)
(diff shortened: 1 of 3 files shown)


spanner/client.go
```diff
@@ -795,7 +748,16 @@ func metricsStreamInterceptor() grpc.StreamClientInterceptor {
 		if err != nil {
 			return nil, err
 		}
-		return newWrappedStream(s, method, cc.Target()), nil
+		mt, ok := ctx.Value(metricsTracerKey).(*builtinMetricsTracer)
+		if ok {
+			if mt != nil {
```
olavloite: nit: could we combine these into `if ok && mt != nil` to reduce nesting

spanner/grpc_client.go
```diff
@@ -192,6 +192,15 @@ func (g *grpcSpannerClient) ExecuteStreamingSql(ctx context.Context, req *spanne
 	if mt != nil && client != nil && mt.currOp.currAttempt != nil {
 		md, _ := client.Header()
 		mt.currOp.currAttempt.setServerTimingMetrics(parseServerTimingHeader(md))
+		peerInfo, ok := peer.FromContext(client.Context())
+		if ok {
+			if peerInfo.Addr != nil {
```
olavloite: nit: combine these into one if statement

spanner/grpc_client.go
```diff
@@ -192,6 +192,15 @@ func (g *grpcSpannerClient) ExecuteStreamingSql(ctx context.Context, req *spanne
 	if mt != nil && client != nil && mt.currOp.currAttempt != nil {
 		md, _ := client.Header()
 		mt.currOp.currAttempt.setServerTimingMetrics(parseServerTimingHeader(md))
+		peerInfo, ok := peer.FromContext(client.Context())
```
olavloite: Can we move this code block into a small util method and call it from the three setreaming RPCs that need it, instead of repeating it for each RPC?

olavloite (dismissed):

PR #12627, opened 2025-07-31
feat(spanner): default enable multiplex session for all operations unless explicitly set to false

rahul2393:
Read-Only, Read-Write and Partition Ops transactions now use a single multiplexed session, instead of a session from the session pool. Multiplexed sessions can handle any number of queries and read-only, read-write and partition ops transactions transactions at the same time.

The use of multiplexed sessions can be disabled for read-only, read-write and partition-ops transactions by setting the environment variables.

GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS=false
GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS_FOR_RW=false.
GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS_PARTITIONED_OPS=false.

See https://cloud.google.com/spanner/docs/sessions#multiplexed_sessions for more background information about multiplexed sessions.

### DIFF
--- a/spanner/client.go
+++ b/spanner/client.go
@@ -551,21 +551,23 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
 	if isAFEBuiltInMetricEnabled {
 		md.Append(afeMetricHeader, "true")
 	}
-
-	if isMultiplexed := strings.ToLower(os.Getenv("GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS")); isMultiplexed != "" && !config.enableMultiplexSession {
+	config.enableMultiplexSession = true
+	if isMultiplexed := strings.ToLower(os.Getenv("GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS")); isMultiplexed != "" {
 		config.enableMultiplexSession, err = strconv.ParseBool(isMultiplexed)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


spanner/client.go
```diff
@@ -551,21 +551,23 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
 	if isAFEBuiltInMetricEnabled {
 		md.Append(afeMetricHeader, "true")
 	}
-
-	if isMultiplexed := strings.ToLower(os.Getenv("GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS")); isMultiplexed != "" && !config.enableMultiplexSession {
```
sakthivelmanii, line 555: use LookupEnv to handle this which will be cleaner approach.

```
if isMultiplexed, found := os.LookupEnv("GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS"); found {
		config.enableMultiplexSession, err = strconv.ParseBool(strings.ToLower(isMultiplexed))
		if err != nil {
			return nil, spannerErrorf(codes.InvalidArgument, "GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS must be either true or false")
		}
	} else {
		config.enableMultiplexSession = true
	}
```

This will ensure that if env is present, it will use the value, if it doesn't, then it fallback to default behaviour of enabling by default.

rahul2393:
Assigning to @sakthivelmanii to take it forward.

PR #12635, opened 2025-08-01
fix(spanner): context cancel in traces in case of skipping trailers

rahul2393:


spanner/read.go
```diff
@@ -677,6 +680,10 @@ func (d *resumableStreamDecoder) tryRecv(mt *builtinMetricsTracer, retryer gax.R
 	if d.err == nil {
 		d.q.push(res)
 		if res.GetLast() {
+			if span := otrace.SpanFromContext(d.stream.Context()); span != nil && span.IsRecording() {
+				span.SetStatus(otcodes.Ok, "Stream finished successfully")
+				span.End()
+			}
 			go func(s streamingReceiver) {
```
olavloite: Would this change then also make it possible to move the `d.cancel()` call from the goroutine into the main thread (and remove the extra s.Recv() call)? Or would that then still cause either the trace or the call server-side to show up as cancelled?

spanner/read.go
```diff
@@ -677,6 +680,10 @@ func (d *resumableStreamDecoder) tryRecv(mt *builtinMetricsTracer, retryer gax.R
 	if d.err == nil {
 		d.q.push(res)
 		if res.GetLast() {
+			if span := otrace.SpanFromContext(d.stream.Context()); span != nil && span.IsRecording() {
+				span.SetStatus(otcodes.Ok, "Stream finished successfully")
+				span.End()
```
olavloite, line 685: Does this cause `span.End()` to be called twice? Once here, and once in gax? And if so; does that matter, or does OpenTelemetry accept that without issues (IIRC, OpenCensus in Java had a problem with that)

spanner/read.go
rahul2393: Removed the extra go routine

rahul2393: iter.Stop will cancel the routine now anyways

spanner/read.go
rahul2393: https://github.com/census-instrumentation/opencensus-go/blob/01e6da5fc01c42aca1e0ce315f41744876e9fcbb/trace/trace.go#L287

In opencensus lib we are making sure End only happens once.

rahul2393: For opentelemetry while span End we return early if endTime is already set

https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/span.go#L461-L464

```
func (s *recordingSpan) isRecording() bool {
	if s == nil {
		return false
	}
	return s.endTime.IsZero()
}
```

olavloite (approved):

yordis:
@rahul2393 do you have any ETA for the latest changes release?

rahul2393:
@yordis We already released it in 1.84.0

PR #12662, opened 2025-08-06
fix: receive trailers to prevent buildup of pending goroutines

olavloite:
Start a goroutine to get the pending trailers to make sure that the stream ends successfully. Remove the cancel function from the stream, to prevent iterator.Stop from potentially calling the cancel function before the trailers have been received.

### DIFF
--- a/spanner/read.go
+++ b/spanner/read.go
@@ -684,6 +684,15 @@ func (d *resumableStreamDecoder) tryRecv(mt *builtinMetricsTracer, retryer gax.R
 				span.SetStatus(otcodes.Ok, "Stream finished successfully")
 				span.End()
 			}
+			if d.cancel != nil {
+				// Remove the cancel function to prevent iter.Stop from also calling it.
+				cancel := d.cancel
+				d.cancel = nil
+				go func() {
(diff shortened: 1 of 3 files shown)


rahul2393 (approved):

spanner/read.go
```diff
@@ -684,6 +684,15 @@ func (d *resumableStreamDecoder) tryRecv(mt *builtinMetricsTracer, retryer gax.R
 				span.SetStatus(otcodes.Ok, "Stream finished successfully")
 				span.End()
 			}
+			if d.cancel != nil {
+				// Remove the cancel function to prevent iter.Stop from also calling it.
+				cancel := d.cancel
+				d.cancel = nil
```
rahul2393, line 690: nit: iter.Stop will have d.cancel nil because https://github.com/googleapis/google-cloud-go/blob/1a8820900f20e038291c4bb2c5284a449196e81f/spanner/read.go#L240

olavloite: Yes, that is intentional to prevent a possible race condition where both this goroutine and iter.Stop will try to call cancel. This makes sure that __only__ this goroutine will call cancel.

PR #12666, opened 2025-08-06
fix(auth): add UseDefaultClient: true to metadata.Options

quartzmo:
* Enable use of custom logger combined with default internal shared http.Client
  for compute token requests. The restores the previous behavior of sharing a
  single TCP connection pool as was done previously with metadata.GetWithContext.
* Add TestDefaultCredentials_OnGCE.
* Bump cloud.google.com/go/compute/metadata dependency to v0.8.0

refs: #11078
refs: #12657


codyoss (approved):
Thanks

PR #12668, opened 2025-08-06
feat(storage): WithMeterProvider allows custom meter provider configuration

cojenco:


storage/grpc_metrics.go
```diff
@@ -172,7 +173,12 @@ func newGRPCMetricContext(ctx context.Context, cfg metricsConfig) (*metricsConte
 		meterOpts = append(meterOpts, metric.WithReader(
 			metric.NewPeriodicReader(&exporterLogSuppressor{Exporter: exporter}, metric.WithInterval(interval))))
 	}
-	provider := metric.NewMeterProvider(meterOpts...)
+	var provider *metric.MeterProvider
+	if cfg.meterProvider != nil {
+		provider = cfg.meterProvider
+	} else {
+		provider = metric.NewMeterProvider(meterOpts...)
+	}
```
BrennaEpp:
```suggestion
	provider := cfg.meterProvider
	if provider == nil {
		provider = metric.NewMeterProvider(meterOpts...)
	}
```

storage/internal/experimental.go
```diff
@@ -26,6 +26,10 @@ var (
 	// Set an alternate client-side metric Exporter to emit metrics through.
 	WithMetricExporter any // func (*metric.Exporter) option.ClientOption
 
+	// WithMeterProvider is a function which is implemented by storage package.
+	// Set an alternate client-side meter provider to emit metrics through.
+	WithMeterProvider any // func (*metric.MeterProvider) option.ClientOption
```
BrennaEpp, line 31: How is this different than the metric Exporter?

BrennaEpp (commented):
Should we also test how the provider interacts with other configs?

storage/internal/experimental.go
cojenco: Unlike the `WithMetricExporter` option, which only allows a single exporter, `WithMeterProvider` supports multiple exporters.

This is essential for GCSFuse, which allows multiple exporters, such as both Prometheus and Google Cloud Monitoring. The MeterProvider also supports the interfaces and functionality required by the OpenTelemetry SDK, which a simple exporter alone may not have.

cojenco:
> Should we also test how the provider interacts with other configs?

I could add more tests using multiple client options. There are more integration tests added in GCSFuse as well https://github.com/GoogleCloudPlatform/gcsfuse/pull/3729

BrennaEpp (approved):

PR #12673, opened 2025-08-07
feat(auth): improve error message for missing case, update unit tests

### LINKED ISSUE #12672: packagename: short description of feature request
In case of a valid but empty json credential, the error message is not really easy to read. For instance, if I use an option like this 

>option.WithCredentialsJSON([]byte(`{}`)

until `cloud.google.com/go/storage@v1.54.0` it returns an error message like this:

> storage.NewClient(ctx): dialing: missing 'type' field in credentials

however, newer versions return an error message like this:

> storage.NewClient(ctx): dialing: credentials: unsupported filetype '\x00'

what is this byte 00 ? unclear 

A better error message

peczenyj:
Today if I try to create an instance of storage or other service, if I provide an empty json body as credential, the error message is not clear

`storage.NewClient(ctx): dialing: credentials: unsupported filetype '\x00'`

what is this hexadecimal byte 0? looks like some magic number, it is not easy to understand.

but until some versions ago, the error was clear about the missing type.

I trace the root cause and I note that 0 was the value for credsfile.UnknownCredType enumeration type. I try to enrich the error message by highlighting the meaning of this status. I decided to keep the default in the switch case to minimize impact

fixes #12672


auth/credentials/filetypes.go
```diff
@@ -36,6 +36,8 @@ func fileCredentials(b []byte, opts *DetectOptions) (*auth.Credentials, error) {
 	var projectID, universeDomain string
 	var tp auth.TokenProvider
 	switch fileType {
+	case credsfile.UnknownCredType:
+		return nil, errors.New("credentials: unsupported unidentified file type")
```
codyoss, line 40: I think this is already covered by the default case, so the error message you are referencing in the bug has its origin elsewhere, right?

peczenyj: Sorry, this is not a **bug**, just the error message is not obvious. the current tests only checks if there is an error, it is not about the message.

IMHO `credentials: unsupported filetype '\x00'` is hard to read. what is filetype 0?

perhaps making `CredentialType` a `Stringer` and return some readable string representation could also help

quartzmo: This change LGTM. We preserve the default case error message which may provide someone debugging with valuable information (the typecode), yet this new case intercepts the unknown type (default in `parseCredentialType`) and provides a better message for that case. The default case in this switch statement  now serves only to handle programmer error, for example if a new typecode is added to `parseCredentialType` but is not yet supported here.

quartzmo (approved):

quartzmo:
@peczenyj Thank you for this contribution!

PR #12676, opened 2025-08-08
fix(internal/librariangen): get README title from service config yaml

quartzmo:
* Update all testdata usages of ID value from generate-request.json
    * Add testdata workflows_v1.yaml
    * Replace incorrectly suppressed errors with returned errors
    * Remove unneeded gofmt from integration test scripts
    * Move goimports call to after version file creation in postprocess
    * Get module import path from first API, not last
    * Remove unused code from Generate
    * Add support for GOOGLE_SDK_GO_LOGGING_LEVEL=debug
    * Lower most log statements to debug
    * Add librariangen prefix to all log statements
    * Add librariangen prefix to all error messages


internal/librariangen/generate/generator.go
```diff
@@ -29,6 +29,7 @@ import (
 	"cloud.google.com/go/internal/postprocessor/librarian/librariangen/postprocessor"
 	"cloud.google.com/go/internal/postprocessor/librarian/librariangen/protoc"
 	"cloud.google.com/go/internal/postprocessor/librarian/librariangen/request"
+	"github.com/ghodss/yaml"
```
codyoss: For now could we use the same lib as librarian? gopkg.in/yaml.v3

It is not perfect either but the above library has not had an updated in 4 years.

codyoss (commented):
just one nit

internal/librariangen/generate/generator.go
quartzmo: Sure, no problem. I'll switch it.

codyoss (approved):

PR #12684, opened 2025-08-11
feat(bigtable): expose directaccess via a env variable

sushanb:


bigtable/bigtable.go
```diff
@@ -204,6 +210,7 @@ var (
 		Max:        2 * time.Second,
 		Multiplier: 1.2,
 	}
+	enableDirectAccess                = false
```
bhshkh: remove this. it is not needed

bigtable/bigtable.go
```diff
@@ -132,6 +133,11 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
 		option.WithGRPCDialOption(grpc.WithDefaultCallOptions(grpc.MaxCallSendMsgSize(1<<28), grpc.MaxCallRecvMsgSize(1<<28))),
 	)
 
+	if useDirectAccess, _ := strconv.ParseBool(os.Getenv("CBT_ENABLE_DIRECTPATH")); useDirectAccess {
+		enableDirectAccess = true
+		o = append(o, internaloption.EnableDirectPath(true), internaloption.EnableDirectPathXds())
+	}
```
bhshkh:
```suggestion

	enableDirectAccess, _ := strconv.ParseBool(os.Getenv("BIGTABLE_ENABLE_DIRECTPATH"))
	if enableDirectAccess {
		o = append(o, internaloption.EnableDirectPath(true), internaloption.EnableDirectPathXds())
	}
```

and then use enableDirectAccess after line 185
i.e.

return &Client{
 ....
enableDirectAccess: enableDirectAccess
}, nil

mutianf (approved):

PR #12685, opened 2025-08-11
feat(bigtable): add client epoch for each call

sushanb:


bhshkh:
Have you tested this ? Do the headers contain correct value?

mutianf:
Is there existing test that checks the request metdata? If so, can we add a test to check that the epoch time metadata is in the header?

sushanb:
> Have you tested this ? Do the headers contain correct value?

I will add a unit test but can confirm that the value is being populated via inspecting server loggs.

Issue #12448, opened 2025-06-15
firestore: WithReadOptions causes panic and ignores read settings in certain scenarios


sanbiv:
## Client

Firestore

## Environment

- Go version: go1.24+
- Platform: macOS
- Package: cloud.google.com/go/firestore

## Code and Dependencies

```go
package main

import (
    "context"
    "time"
    
    "cloud.google.com/go/firestore"
    "google.golang.org/api/option"
)

func main() {
    ctx := context.Background()
    client, err := firestore.NewClient(ctx, "project-id", option.WithCredentialsFile("path/to/credentials.json"))
    if err != nil {
        panic(err)
    }
    defer client.Close()

    tm := time.Date(2025, time.June, 14, 0, 0, 0, 0, time.UTC)

    // Case 1: CollectionGroup with WithReadOptions - PANIC
    colGroupRef := client.CollectionGroup("myCollection")
    query1 := colGroupRef.WithReadOptions(firestore.ReadTime(tm))
    // This will panic when trying to execute the query

    // Case 2: Collection Query with WithReadOptions - PANIC  
    colRef := client.Collection("myCollection")
    query2 := colRef.Query.WithReadOptions(firestore.ReadTime(tm))
    // This will also panic when trying to execute the query

    // Case 3: Collection WithReadOptions - IGNORES READ SETTINGS
    colRef2 := client.Collection("myCollection").WithReadOptions(firestore.ReadTime(tm))
    query3 := colRef2.Query // readSettings are ignored when accessing Query field
    query3 = query3.Limit(1)
    docs, err := query3.Documents(ctx).GetAll()
    if err != nil {
        panic(err)
    }
    // readSettings are not applied, uses default read behavior
}
```

<details>
  <summary>go.mod</summary>

```text
module firestore-bug-test

go 1.24

require (
    cloud.google.com/go/firestore v1.14.0
    google.golang.org/api v0.149.0
)
```

</details>

## Expected behavior

1. **Case 1 & 2**: `WithReadOptions` should work correctly with CollectionGroup and Collection Query without causing panics
2. **Case 3**: When calling `WithReadOptions` on a Collection, the read settings should be preserved and applied when accessing the `Query` field

## Actual behavior

1. **Case 1**: CollectionGroup with `WithReadOptions` causes a panic:
   ```
   panic: runtime error: invalid memory address or nil pointer dereference
   [signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x101710e78]
[...]

PR #12689, opened 2025-08-11
Init default value for Query's readSettings to avoid panic

dongnguyen-geocomply:


sanbiv:
Hi @dongnguyen-geocomply, thanks for the PR - much appreciated! However, it’s incomplete.

As I mentioned here: https://github.com/googleapis/google-cloud-go/issues/12448, the fix requires setting `readSettings` in three different places. Please check the “Suggested Solutions” section in that comment and look for the lines marked with `// add this`

I didn’t open a PR because I had asked a question, but nobody replied.

PR #12698, opened 2025-08-12
feat(bigtable): Record client_blocking_latencies

bhshkh:
This PR adds client_blocking_latencies metrics. More about the metric can be found at https://cloud.google.com/bigtable/docs/client-side-metrics-descriptions#client-blocking-latencies

### DIFF
--- a/bigtable/metrics.go
+++ b/bigtable/metrics.go
@@ -616,6 +616,15 @@ func (mt *builtinMetricsTracer) recordAttemptCompletion(attemptHeaderMD, attempT
 	attemptLatAttrs, _ := mt.toOtelMetricAttrs(metricNameAttemptLatencies)
 	mt.instrumentAttemptLatencies.Record(mt.ctx, elapsedTime, metric.WithAttributeSet(attemptLatAttrs))
 
+	// Record client_blocking_latencies
+	var clientBlockingLatencyMs float64
+	if mt.currOp.currAttempt.blockingLatencyTracker != nil {
+		messageSentNanos := mt.currOp.currAttempt.blockingLatencyTracker.getMessageSentNanos()
+		clientBlockingLatencyMs = convertToMs(time.Unix(0, int64(messageSentNanos)).Sub(mt.currOp.currAttempt.startTime))
+	}
+	clientBlockingLatAttrs, _ := mt.toOtelMetricAttrs(metricNameClientBlockingLatencies)
+	mt.instrumentClientBlockingLatencies.Record(mt.ctx, clientBlockingLatencyMs, metric.WithAttributeSet(clientBlockingLatAttrs))
+
 	// Record server_latencies
 	serverLatAttrs, _ := mt.toOtelMetricAttrs(metricNameServerLatencies)
 	if mt.currOp.currAttempt.serverLatencyErr == nil {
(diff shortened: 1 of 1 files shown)


danowar2:
Think I need a new available for Data

bigtable/metrics.go
```diff
@@ -486,3 +519,61 @@ func (mt *builtinMetricsTracer) toOtelMetricAttrs(metricName string) (attribute.
...
+	tracker, ok := ctx.Value(statsContextKey).(*blockingLatencyTracker)
+	if !ok {
+		return
+	}
+
+	switch st := s.(type) {
+	case *stats.Begin:
+		tracker.setBegin(st.BeginTime)
+	case *stats.OutPayload:
+		tracker.recordLatency(time.Now())
```
sushanb: I think you can use the SentTime time.Time
https://github.com/grpc/grpc-go/blob/19c720f666f3d405dec68363733f19e9f80ea7bf/stats/stats.go#L174

and probably use the delta between attemptTimer - sentTime as blocking latency

bigtable/metrics.go
```diff
@@ -289,6 +303,17 @@ func (tf *builtinMetricsTracerFactory) createInstruments(meter metric.Meter) err
 		return err
 	}
 
+	// Create client_blocking_latencies
+	tf.clientBlockingLatencies, err = meter.Float64Histogram(
+		metricNameClientBlockingLatencies,
+		metric.WithDescription("The artificial latency introduced by the client to limit the number of outstanding requests. The publishing of the measurement will be delayed until the attempt trailers have been received."),
```
mutianf: I think the description should be: Latencies of requests queued on gRPC channels

bigtable/metrics.go
```diff
@@ -495,3 +528,61 @@ func (mt *builtinMetricsTracer) toOtelMetricAttrs(metricName string) (attribute.
...
+
+func (h *latencyStatsHandler) HandleRPC(ctx context.Context, s stats.RPCStats) {
+	tracker, ok := ctx.Value(statsContextKey).(*blockingLatencyTracker)
+	if !ok {
+		return
+	}
+
+	switch st := s.(type) {
+	case *stats.Begin:
+		tracker.setBegin(st.BeginTime)
```
mutianf: I think the begin time should be set when the operation starts. So where we record the start time of operation latency.

bigtable/metrics.go
```diff
@@ -495,3 +528,61 @@ func (mt *builtinMetricsTracer) toOtelMetricAttrs(metricName string) (attribute.
...
+}
+
+func (t *blockingLatencyTracker) setBegin(begin time.Time) {
+	t.Lock()
+	defer t.Unlock()
+	t.begin = begin
+}
+
+func (t *blockingLatencyTracker) recordLatency(end time.Time) {
+	t.Lock()
```
mutianf: Why do we need the lock? what threads are accessing the tracker at the same time?

bhshkh: **what threads are accessing the tracker?** 
1. Application goroutine that calls the Bigtable client methods (e.g. ReadRows, ApplyBulk).

2. gRPC internal threads:
gRPC library executes the HandleRPC method of the latencyStatsHandler in its own internal goroutines. There is no guarantee that the goroutine handling stats.Begin is the same one handling stats.OutPayload for the same RPC. gRPC can use multiple workers.

**why do we need the lock?**
To avoid race conditions such as:
1. A read of t.begin in recordLatency before setBegin has finished writing it.
2. A read of t.latency in getLatency could happen while recordLatency is in the middle of writing it
3. Multiple stats.OutPayload events for the same stream could trigger tracker.recordLatency() simultaneously in different gRPC goroutines. These could interleave their read and write on t.latency which could result in an incorrect value.


I have removed the lock and used atomics

bigtable/metrics.go
bhshkh: Udpated the description to match https://cloud.google.com/bigtable/docs/client-side-metrics-descriptions#client-blocking-latencies

bigtable/metrics.go
bhshkh: Right. Updated

bigtable/metrics.go
mutianf: Yeah but in go we only implemented the second half of the description which is " this metric also includes the latencies of requests queued on gRPC channels."

sushanb (dismissed):

bigtable/metrics.go

bigtable/metrics.go

mutianf (approved):

PR #12708, opened 2025-08-14
fix(bigtable): Ignore errors while creating monitoring client and disable metrics 

bhshkh:
Any errors during creating monitoring client should not disrupt the normal functioning of the client. swallow all such errors

Context: b/432819321


bigtable/metrics.go
```diff
@@ -175,26 +180,22 @@ type builtinMetricsTracerFactory struct {
 	debugTags            metric.Int64Counter
 }
 
+// Returns error only if metricsProvider is of unknown type. Rest all errors are swallowed
```
hongalex, line 183: Trying to grok this, maybe meant to say "All other errors are swallowed with metrics disabled"

Issue #12680, opened 2025-08-08
spanner: large number of unnecessary allocations constructing mutations in BatchWrite


pbotros:
## Client

Spanner

## Environment

All, on latest master

## Code and Dependencies

In [spanner/mutation.go](https://github.com/googleapis/google-cloud-go/tree/spanner/v1.84.1/spanner/mutation.go), `mutationsProto` unnecessary allocates slices and `rand` objects:

```go
func mutationsProto(ms []*Mutation) ([]*sppb.Mutation, *sppb.Mutation, error) {
	var selectedMutation *Mutation
	var nonInsertMutations []*Mutation
...
	if len(nonInsertMutations) > 0 {
		selectedMutation = nonInsertMutations[rand.New(rand.NewSource(time.Now().UnixNano())).Intn(len(nonInsertMutations))]
	}
...
}
```

All production callers of this function ignore the second return value. The only functionality of tracking `selectedMutation, nonInsertMutations` seem to be for test.

## Expected behavior
There are fewer unnecessary allocations.

## Actual behavior
Allocations due to `rand` and slices only used for test cause massive amounts of allocations, increasing CPU and memory usage. See screenshot for a heap profile taken on a pod that performs a large number of BatchWrite calls.

## Screenshots

Heap profile viewing "alloc - space" showing large amounts of allocated space:

<img width="607" height="958" alt="Image" src="https://github.com/user-attachments/assets/30ea61b0-cc56-4317-96f4-cb92013fe21f" />

rahul2393:
Second return value(mutation key) is needed in case of multiplexed sessions enable(which will be default in upcoming release) begin txn requests.

rahul2393:
@pbotros Can you please help review the changes?

pbotros:
Thanks for addressing so quickly!

PR #12740, opened 2025-08-20
perf(spanner): improve mutationProto allocations and performance

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/12680

### DIFF
--- a/spanner/mutation.go
+++ b/spanner/mutation.go
@@ -470,10 +470,10 @@ func (m Mutation) proto() (*sppb.Mutation, error) {
 // it is convenient for sending batch mutations to Cloud Spanner.
 func mutationsProto(ms []*Mutation) ([]*sppb.Mutation, *sppb.Mutation, error) {
 	n := len(ms)
-	if n == 0 {
-		return nil, nil, nil
-	}
 	out := make([]*sppb.Mutation, 0, n)
+	if n == 0 {
+		return out, nil, nil
+	}
 	maxInsertIdx := -1
 	maxInsertVals := -1
 	nonInsertCount := 0
--- a/spanner/mutation.go
+++ b/spanner/mutation.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"math/rand"
 	"reflect"
-	"time"
 
 	sppb "cloud.google.com/go/spanner/apiv1/spannerpb"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


rahul2393:
The benchmark results indicate a significant performance improvement. The mutationsProto function is now approximately 81% faster, uses 57% less memory, and makes 16% fewer memory allocations on average.

```
goos: darwin
goarch: arm64
cpu: Apple M1 Pro
                                            │  before.txt  │              after.txt              │
                                            │    sec/op    │   sec/op     vs base                │
MutationsProto/small_number_of_mutations-10   13.778µ ± 1%   1.548µ ± 3%  -88.77% (p=0.000 n=20)
MutationsProto/large_number_of_mutations-10    43.83µ ± 1%   30.51µ ± 1%  -30.38% (p=0.000 n=20)
MutationsProto/mixed_type_of_mutations-10     13.289µ ± 1%   1.140µ ± 1%  -91.42% (p=0.000 n=20)
geomean                                        20.02µ        3.776µ       -81.14%

                                            │  before.txt  │              after.txt               │
                                            │     B/op     │     B/op      vs base                │
MutationsProto/small_number_of_mutations-10   8.844Ki ± 0%   2.961Ki ± 0%  -66.52% (p=0.000 n=20)
MutationsProto/large_number_of_mutations-10   67.10Ki ± 0%   59.16Ki ± 0%  -11.84% (p=0.000 n=20)
MutationsProto/mixed_type_of_mutations-10     7.979Ki ± 0%   2.164Ki ± 0%  -72.88% (p=0.000 n=20)
geomean                                       16.79Ki        7.237Ki       -56.90%

                                            │ before.txt  │              after.txt              │
                                            │  allocs/op  │  allocs/op   vs base                │
MutationsProto/small_number_of_mutations-10    84.00 ± 0%    67.00 ± 0%  -20.24% (p=0.000 n=20)
MutationsProto/large_number_of_mutations-10   1.343k ± 0%   1.321k ± 0%   -1.64% (p=0.000 n=20)
MutationsProto/mixed_type_of_mutations-10      64.00 ± 0%    49.00 ± 0%  -23.44% (p=0.000 n=20)
geomean                                        193.3         163.1       -15.63%
```

rahul2393:
@codyoss quick review here is really appreciated.

spanner/mutation.go
```diff
@@ -470,41 +469,39 @@ func (m Mutation) proto() (*sppb.Mutation, error) {
...
+		if m.op == opInsert {
+			if v := len(m.values); v >= maxInsertVals {
+				maxInsertVals, maxInsertIdx = v, i
+			}
+			continue
+		}
+		nonInsertCount++
+		if rand.Intn(nonInsertCount) == 0 {
+			selectedNonInsertIdx = i
+		}
```
pbotros, line 496: I like how we're avoiding allocs now, but one improvement here is to avoid doing the `rand.Intn()` in a tightloop. Would you consider doing an extra loop through the mutations? I'd imagine that may actually result in _better_ performance since `rand` methods are relatively slow.

```suggestion
```

and then later:

```
if nonInsertCount > 0 {
  selectedNonInsertIdx := rand.Intn(nonInsertCount)
  var selectedMutation *Mutation
	for i, outM := range out {
         inM := ms[i]
         if inM.op == opInsert {
			if selectedNonInsertIdx == 0 {
				selectedMutation = outM
                 break
			}
			selectedNonInsertIdx--
		}
      return out, selectedMutation, nil
	}
}
```

pbotros: Actually, tried out the diff on my local branch and it didn't seem to be any faster. So nevermind here.

pbotros, line 496: nit: at first glance it's a bit surprising that this ends up being a uniform distribution, so perhaps a comment can help here:
```suggestion
         // Pick a non-insert mutation uniformly at random.
		if rand.Intn(nonInsertCount) == 0 {
			selectedNonInsertIdx = i
		}
```

pbotros (approved):

spanner/mutation.go
```diff
@@ -470,41 +469,39 @@ func (m Mutation) proto() (*sppb.Mutation, error) {
...
-		if err != nil {
-			return nil, nil, err
-		}
-		return l, m, nil
+	if maxInsertIdx >= 0 {
+		return out, out[maxInsertIdx], nil
 	}
-
-	return l, nil, nil
+	return out, nil, nil
```
olavloite, line 504: This point will never be reached, right? Because we always have:
1. An empty input slice of mutations, which causes an empty return
2. At least one non-insert mutation.
3. Or one insert mutation.

rahul2393: Yes, this is here because Golang expects a return at the end of method.

PR #12754, opened 2025-08-22
feat: use old protoc plugin for go_proto_library

jskeet:
This mimics the behavior of Bazel for go_proto_library targets which contain a compilers attribute of `["@io_bazel_rules_go//proto:go_grpc"]`. In that case, we need to use the old version of protoc-gen-go, passing in "plugins=grpc" as an option.

We have to clone and build that explicitly within the Dockerfile, as there's a redirect from the old plugin to the new one. Furthermore, the old plugin uses the internal code from the new plugin, so we need to set the imported new plugin version to "the same one we use for everything else" directly via "go mod edit".

This is all very hacky, but appears to generate identical code to Bazel (tested with secretmanager v1beta2). We will be able to get rid of it all once we've migrated all packages to go_grpc.


codyoss (approved):
Thank you!

codyoss:
I think we can use git history later to unwind.

PR #12756, opened 2025-08-22
feat: fix snippets and client versions for multi-path APIs

jskeet:
This is one new commit on top of #12775.

With this in place, the Spanner diff is only README.md and a "generated by" comment in internal/version.go (as well as go.mod/go.sum).

Note that the release command currently has its own snippet metadata handling code, which is slightly different (basically finding files recursively, instead of predicting the presence of specific files). We'll want to refactor all of the snippet metadata handling code to remove duplication - we should figure out which approach to use.


jskeet:
Closing in favor of #12794.

PR #12757, opened 2025-08-22
test(storage): enable more ZB integration tests

tritone:
Shifts BidiRead tests to use a zonal bucket rather than a regional bucket.

Required skipping some tests which are for features not supported by ZB currently.

I'll be adding more write tests in a separate PR.


storage/integration_test.go
```diff
@@ -6939,43 +6955,6 @@ func TestIntegration_OTelTracing(t *testing.T) {
 	})
 }
 
-// Simple integration test for a zonal bucket read via storage.Reader.
-// Will test out-of-region redirect flow if it's run from outside of us-west4.
-func TestIntegration_ZonalRead(t *testing.T) {
```
tritone, line 6944: This test was no longer necessary since downloads are covered in many other tests.

ddelgrosso1 (approved):

BrennaEpp (commented):
LGTM

PR #12774, opened 2025-08-26
tests: fix tests that use secretmanager files

jskeet:
These were effectively left over from two PRs which crossed.

(I've also modified the service config so it's more sensible for Secret Manager...)


jskeet:
Closing in favor of #12794.

PR #12776, opened 2025-08-26
tools: first pass at a state file generator for google-cloud-go

jskeet:
There are no tests at the moment - the tests would be significantly larger than the code itself (which is expected to be temporary anyway). Let me know if you'd like me to work on tests anyway...

### DIFF
--- a/internal/stategen/main.go
+++ b/internal/stategen/main.go
@@ -57,8 +57,7 @@ func run(args []string) error {
 			slog.Info("skipping existing module", "module", moduleName)
 			continue
 		}
-		err = addModule(repoRoot, state, moduleName)
-		if err != nil {
+		if err := addModule(repoRoot, state, moduleName); err != nil {
 			return err
 		}
 	}


internal/stategen/main.go
```diff
@@ -0,0 +1,197 @@
...
+	state, err := parseLibrarianState(stateFilePath)
+	if err != nil {
+		return err
+	}
+	for _, moduleName := range args[1:] {
+		if stateContainsModule(state, moduleName) {
+			slog.Info("skipping existing module", "module", moduleName)
+			continue
+		}
+		err = addModule(repoRoot, state, moduleName)
```
codyoss: nit: I would usually write this as
```suggestion
		if err := addModule(repoRoot, state, moduleName); err != nil {
```

internal/stategen/main.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	return false
+}
+
+func addModule(repoRoot string, state *LibrarianState, moduleName string) error {
+	slog.Info("adding module", "module", moduleName)
+	moduleRoot := filepath.Join(repoRoot, moduleName)
+
+	// Start off with the basics which need
+	library := &LibraryState{
```
codyoss, line 81: I think this is fine for an initial pass, but I wonder if we can make this smarter by reading some of the existing configuration. There are two good sources of information in google-cloud-go.
1. The owlbot config: https://github.com/googleapis/google-cloud-go/blob/main/.github/.OwlBot.yaml
2. Our side configuration: https://github.com/googleapis/google-cloud-go/blob/main/internal/postprocessor/config.yaml

internal/stategen/main.go
```diff
@@ -0,0 +1,197 @@
...
+		ID: moduleName,
+		SourceRoots: []string{
+			moduleName,
+			"internal/generated/snippets/" + moduleName,
+		},
+		RemoveRegex: []string{
```
codyoss, line 87: I wonder if we need anything special here at all. I would kinda hope the we can simply wipe the source_roots ( not specify anything here ) and just build out the preserve-regex if needed. Thoughts?

internal/stategen/main.go
jskeet: Thanks, will fix.

internal/stategen/main.go
jskeet: I'd much rather delete everything that we expect to generate - I originally got Spanner working by preserving all the code I knew to be handwritten, but that ends up being really brittle... if someone adds a new handwritten file, it would be very easy to fail to add it to the state file.

If all the handwritten code were in one conventional place that was easily distinguishable (just by filename) from the generated code, that would be a different matter.

If there's another option which handles handwritten code, I'm all for it.

internal/stategen/main.go
jskeet: I was aware of the OwlBot config, but hadn't looked at the side configuration.

I was generally reckoning that using this as a starting point and then manually tweaking the state file for "unusual" APIs would be easier than making the tooling smart enough to handle everything. Looking at OwlBot, I figured it would be relatively few APIs that needed tweaking that way.

The side config list of input directories may be a better option than using the snippet metadata to figure out the API paths though - is that one of the things you were thinking about?

codyoss (approved):

PR #12777, opened 2025-08-26
tools: use the post-processor config to determine API paths

jskeet:
This is more reliable (and simpler) than looking for generated snippets.


jskeet:
Closing in favor of #12793

PR #12784, opened 2025-08-27
test(storage): update tests for Writer.ProgressFunc

tritone:
Improve  emulator tests for Writer.ProgressFunc and re-enable this check for the append takeover integration test.

Clarify that progressFunc should be called on all of the following: 1. Takeover 2. Flush (explicit or due to chunk being full) 3. Close. It should not be called on a 1-shot upload.

### DIFF
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -3414,7 +3414,7 @@ func TestIntegration_WriterAppendTakeover(t *testing.T) {
 				opts: &AppendableWriterOpts{
 					ChunkSize: 4 * MiB,
 				},
-				checkProgressOffsets: []int64{7 * MiB, 8 * MiB},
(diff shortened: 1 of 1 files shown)


storage/grpc_writer.go
```diff
@@ -340,7 +340,11 @@ func (w *gRPCWriter) handleCompletion(c gRPCBidiWriteCompletion) {
 		w.buf = w.buf[:0]
 	}
 	w.setSize(c.flushOffset)
```
cojenco: qq: should setSize only be called upon a flush as well?

tritone: No, setting the size on the writer should happen for takeover and close as well. We're just preserving legacy behavior for progressFunc here.

cojenco (approved):

PR #12794, opened 2025-09-01
tools(librariangen): minimize generated differences

jskeet:
- Don't invoke the GAPIC generator for non-GAPIC APIs
- Don't use the legacy gRPC generator for non-gRPC libraries
- Stop writing version, README.md and CHANGES.md files during generate
  (these will be covered by configure)
- If a whole module has a non-default version, move generated code
- Introduce a .librarian/generator-input/repo-config.yaml file for
  overrides:
  - If we don't want to generate the GAPIC clients (e.g. Datastore)
  - If the protobuf package doesn't match the directory structure
    (e.g. maps)
  - Module paths with versions
  - Client directory paths with versions

With these changes, almost all modules can be regenerated with either
no differences, or ones which are expected and can be accommodated.


.github/workflows/librariangen.yml
```diff
@@ -0,0 +1,31 @@
+---
```
codyoss: This will be covered by the regular kokro CI already as we run tests for all modules. If you think this needs it own action I would purpose we use path filtering: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-paths

internal/librariangen/Dockerfile
```diff
@@ -21,7 +21,7 @@
 FROM marketplace.gcr.io/google/debian12:latest AS builder
 
 # Set environment variables for tool versions for easy updates.
-ENV GO_VERSION=1.23.0
+ENV GO_VERSION=1.24.6
```
codyoss: nit: use 1.24.0. This makes our CI easier if we always just use the base versions of things.

.github/workflows/librariangen.yml
jskeet: I'm okay with that - but I'm slightly surprised that we aren't seeing Kokoro failures already, given the missing files that are added (well, moved) in this PR. Happy for us to look into that separately though - I may be looking in the wrong place.

Removed this file in an extra comment.

internal/librariangen/Dockerfile
jskeet: Fixed in an extra commit.

PR #12817, opened 2025-09-05
feat(librariangen): implement the build command

jskeet:
This assumes:

- All modules have at least one test file (may not be true for things like orgpolicy which are just generated from protos, no GAPIC)
- It's okay to just run `go test ./...` without worrying about integration tests

Additionally, it *may* be worth just running `go test` without the build part first; it's unclear to me whether that could lead to confusion though. Thoughts welcome.

### DIFF
--- a/internal/librariangen/build/builder.go
+++ b/internal/librariangen/build/builder.go
@@ -83,7 +83,7 @@ func goBuild(ctx context.Context, dir, module string) error {
 // goTest builds all the code under the specified directory
 func goTest(ctx context.Context, dir, module string) error {
 	slog.Info("librariangen: testing", "module", module)
-	args := []string{"go", "test", "./..."}
+	args := []string{"go", "test", "./...", "-short"}
 	return execvRun(ctx, args, dir)
 }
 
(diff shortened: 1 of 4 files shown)


internal/librariangen/build/builder.go
```diff
@@ -0,0 +1,103 @@
...
+func goBuild(ctx context.Context, dir, module string) error {
+	slog.Info("librariangen: building", "module", module)
+	args := []string{"go", "build", "./..."}
+	return execvRun(ctx, args, dir)
+}
+
+// goTest builds all the code under the specified directory
+func goTest(ctx context.Context, dir, module string) error {
+	slog.Info("librariangen: testing", "module", module)
+	args := []string{"go", "test", "./..."}
```
codyoss: nit: add `-short` flag to this command. This will exclude integration tests that expect creds.

jskeet: Aha, great! That'll require tweaking the tests a bit... I'll do it first thing on Monday morning.

jskeet: Done in a new commit. (Also fixed some test issues - and I've raised https://github.com/googleapis/librarian/issues/1960 to suggest using ContinueOnError for flag parsing errors, so we can test them.)

PR #12829, opened 2025-09-08
feat(pubsub/v2): add subscriber shutdown options

hongalex:
This introduces the ability to tell the client library how you want your messages to behave when shutdown is initiated. You can configure the behavior as `ShutdownOptions.Behavior` as `ShutdownBehaviorWaitForProcessing` or `ShutdownBehaviorNackImmediately`.

In addition, you can specify `ShutdownOptions.Timeout` to configure how long you want to wait for messages to be processed, or provide a timeout to the nack calls before returning.

### DIFF
--- a/pubsub/v2/shutdown.go
+++ b/pubsub/v2/shutdown.go
@@ -14,6 +14,24 @@
 
 package pubsub
 
+import "time"
+
+// ShutdownOptions configures the shutdown behavior of the subscriber.
+type ShutdownOptions struct {
+	// Timeout specifies the time the subscriber should wait
+	// to shutdown before killing the process.
+	// In nack mode, this specifies how long to wait
+	// for messages to be nacked after shutdown is initiated.
+	//
+	// Set to zero to immediately shutdown.
+	// Set to a negative value to disable timeout.
+	Timeout time.Duration
+
+	// Behavior defines the strategy the subscriber should use when
+	// shutting down (wait or nack messages).
+	Behavior ShutdownBehavior
+}
+
 // ShutdownBehavior defines the strategy the subscriber should take when
 // shutting down. Current options are graceful shutdown vs nacking messages.
 type ShutdownBehavior int
--- a/pubsub/v2/shutdown.go
+++ b/pubsub/v2/shutdown.go
@@ -17,16 +17,18 @@ package pubsub
 import "time"
 
 // ShutdownOptions configures the shutdown behavior of the subscriber.
-// If not specified, the behavior will default to indefinite processing,
-// that is graceful shutdown with no timeout.
+// If not specified, the default behavior is indefinite processing,
+// aka graceful shutdown with no timeout.
 type ShutdownOptions struct {
 	// Timeout specifies the time the subscriber should wait
-	// to shutdown before killing the process.
-	// In nack mode, this specifies how long to wait
-	// for messages to be nacked after shutdown is initiated.
+	// before forcefully shutting down..
+	// In ShutdownBehaviorNackImmediately mode, this configures the timeout
+	// for message nacks before shutting down.
 	//
-	// Set to zero to immediately shutdown.
+	// Set to zero to immediately shutdown (either modes)
 	// Set to a negative value to disable timeout.
+	// When ShutdownOptions is not set, the client library will
+	// assume disabled/infinite timeout, which matches the current behavior.
 	Timeout time.Duration
 
 	// Behavior defines the strategy the subscriber should use when
@@ -39,11 +41,11 @@ type ShutdownOptions struct {
 type ShutdownBehavior int
 
 const (
-	// ShutdownBehaviorWaitForProcessing means the subscriber will wait for
-	// outstanding messages to be processed before nacking messages finally.
+	// ShutdownBehaviorWaitForProcessing means the subscriber client will wait for
+	// outstanding messages to be processed.
 	ShutdownBehaviorWaitForProcessing = iota
 
-	// ShutdownBehaviorNackImmediately means the subscriber will nack all outstanding
-	// messages before closing.
+	// ShutdownBehaviorNackImmediately means the subscriber client will nack all
+	// outstanding messages before closing.
 	ShutdownBehaviorNackImmediately
 )
--- a/pubsub/v2/shutdown.go
+++ b/pubsub/v2/shutdown.go
@@ -17,6 +17,8 @@ package pubsub
 import "time"
 
 // ShutdownOptions configures the shutdown behavior of the subscriber.
+// If not specified, the behavior will default to indefinite processing,
+// that is graceful shutdown with no timeout.
 type ShutdownOptions struct {
 	// Timeout specifies the time the subscriber should wait
 	// to shutdown before killing the process.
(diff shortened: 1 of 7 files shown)


pubsub/v2/shutdown.go
```diff
@@ -0,0 +1,49 @@
...
+// that is graceful shutdown with no timeout.
+type ShutdownOptions struct {
+	// Timeout specifies the time the subscriber should wait
+	// to shutdown before killing the process.
+	// In nack mode, this specifies how long to wait
+	// for messages to be nacked after shutdown is initiated.
+	//
+	// Set to zero to immediately shutdown.
+	// Set to a negative value to disable timeout.
+	Timeout time.Duration
```
shollyman, line 35: Do we need to disambiguate the default value here?  This isn't a proto where you have field presence that can be checked.  Consider the "wait indefinitely" and "don't wait die immediately" user choices.

pubsub/v2/shutdown.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// ShutdownBehavior defines the strategy the subscriber should take when
+// shutting down. Current options are graceful shutdown vs nacking messages.
+type ShutdownBehavior int
+
+const (
+	// ShutdownBehaviorWaitForProcessing means the subscriber will wait for
+	// outstanding messages to be processed before nacking messages finally.
+	ShutdownBehaviorWaitForProcessing = iota
```
shollyman, line 52: I think either the name here needs work, or the comments do.  The comments describe "graceful shutdown" but that's not part of the name or description of ShutdownBehaviorWaitForProcessing strategy here.

pubsub/v2/shutdown.go
shollyman: Mostly the inversion of timeout behavior from the unspecified ShutdownOptions option to the empty option is what make me itchy here.  In the first we wait indefinitely, in the second we die immediately.

pubsub/v2/shutdown.go
hongalex: Hm, this nacking behavior described was actually part of the design but got dropped during iterations. Will fix this.

pubsub/v2/shutdown.go
hongalex: The "default" value for Timeout is interesting, because it only is relevant when user sets a ShutdownOption.Behavior but not Timeout.

I added a comment warning that if you don't specify both fields when setting `ShutdownOptions`, you will get the forceful shutdown.

There wasn't really a clean choice here, other than maybe to use another optional/pointer value here. In order to get the default/zero value as "wait indefinitely", we have to change the semantics of Timeout=0 to be wait indefinitely, where Timeout=-1 means shutdown immediately. To me, this is more confusing (especially in Pub/Sub where `modack(0)` equals nack / immediately redeliver) and why we decided to implement this way.

shollyman: Another option is to play with the strategies a bit more.  Adding a "legacy" strategy as the default that ignores timeout would give you the same behavior for default values across the board, and then you could have a "graceful with timeout" that doesn't need special logic for negative timeouts.

hongalex: Is the concern / suggestion to not use a pointer for `ShutdownOptions`, so the default values could match the existing behavior?

hongalex: If the above is true, I'm not very happy with the `Behavior = Legacy` and `Timeout = 0` combination though, since that means `Timeout` isn't being respected.

Expanding on this some more, I'm hoping one option would be for users who want immediate timeout to specify a zero timeout without behavior:

```
subscriber.ReceiveSettings.ShutdownOptions = &pubsub.ShutdownOptions{Timeout: 0}
```

Adding a legacy mode would make this not work.

Maybe we can throw an error which prevents the combination of ```ShutdownOptions{Timeout:0, Behavior:Nack}```?

feywind: I had to refresh my memory... basically the former behaviour in all the languages varied between these:

* Just quit immediately and do nothing
* Wait indefinitely, possibly hanging the process
* Wait some length of time and just bail

What we wanted to do is rein in these possibilities and make them consistent:

* Try to wait for things to empty out, or try to quit ASAP?
* How long to wait before giving up?
* Stop receiving new messages right away and nack any we can't process

I think "wait indefinitely with no ability to set a timeout" is sort of broken behaviour - Node used to do it that way. So the assumptions on the new design are like this:

* The user wants to quit soon
* We want to give them to ability to tell us how long to try shutting down the nice way
* If the nice way fails (or they don't want it), then we give ourselves enough time to just nack everything and bail

This gives everyone improvements in concrete behaviour. One might argue that the behaviour shouldn't change unexpectedly, but I'd argue that the existing behaviour was broken, and this is a bug fix of sorts. Either way, if we do want to try to do something different (like a legacy option), we should run that by @kamalaboulhosn at least. (Maybe to adopt the new interface, we would provide some kind of "indefinite" timeout.)

hongalex: I chatted with @shollyman and agreed that while it's unfortunate the nil and zero values of `ShutdownOptions` do different things (and is against protobuf best practices), if it's sufficiently documented, we can proceed without a "Legacy" mode. As mentioned earlier, the existence of a "Legacy" behavior would mean that users won't be able to do "quit immediately" by just setting

```&pubsub.ShutdownOptions{Timeout: 0}```

They would have to specify one of the other non-legacy options (that they're not using)

```
&pubsub.ShutdownOptions{
    Timeout: 0,
    Behavior: WaitForProcessing // this has no semantic meaning here
}
```

hongalex:
> The use of new context.Background() in iterator looks a bit odd, but LGTM given our discussions.

To capture our offline discussions, a detached context.Background() always existed in the iterator (so in flight messages are not cancellable by the user by default). This matches the default behavior. I needed to introduce a context finally because `nackInventory` requires us to be able to cancel that request after `ShutdownOptions.Timeout` duration has passed.

kurochan:
Hi @hongalex ,
In subscriber.go, I noticed that when Timeout > 0, the code always calls iter.nackInventory, regardless of the value of ShutdownOptions.Behavior.
This seems to mean that even if ShutdownBehaviorWaitForProcessing is set, messages will still be nacked immediately.
Was the intention that Behavior should only matter when Timeout is negative (infinite wait), or should the shutdown logic branch on Behavior as well? Could you clarify the intended interaction between Timeout and Behavior?

hongalex:
Approved the above PR but responding here for completeness.

This was indeed a bug, the Behavior should have been checked here (thanks for the PR).

The interaction between Behavior/Timeout is this:

1. Timeout = 0 trumps everything. This tells the library to drop everything on the ground, and shutdown immediately. You don't need to specify a Behavior here.
2. Timeout <=0 means we disabled the timeout, and are waiting indefinitely for either Behavior to complete. In `WaitForProcessing`, wait for messages to be acked or nacked and the callback returns (current library behavior). In `NackImmediately` mode, we `Nack` all messages at shutdown.
3. Timeout >= 0 means we just follow the ShutdownBehavior (wait or nack) up to the specified timeout.

kurochan:
@hongalex Thank you for your kind explanation. I understand.

PR #12839, opened 2025-09-09
fix(storage): free buffers in Bidi Reader

tritone:
Fixes a bug where in the gRPC Bidi Reader code path, allocated buffers were not being freed back to the buffer pool after the messages were read in some cases.

Also adds an emulator test to check for leaks in the codec code for Reader and MultiRangeDownloader

### DIFF
--- a/storage/client_test.go
+++ b/storage/client_test.go
@@ -29,6 +29,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -43,6 +44,7 @@ import (
 	"google.golang.org/api/option"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	expgrpc "google.golang.org/grpc/experimental"
 	"google.golang.org/grpc/mem"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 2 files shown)


storage/client_test.go
```diff
@@ -3004,6 +3006,165 @@ func TestWriterChunkRetryDeadlineEmulated(t *testing.T) {
...
+			gotContents, err := tc.downloadFunc(obj)
+			if err != nil {
+				t.Fatalf("downloading content: %v", err)
+			}
+			if !bytes.Equal(gotContents, contents) {
+				t.Errorf("downloaded bytes did not match; got %v bytes, want %v", len(gotContents), len(contents))
+			}
+			allocs, frees := bp.getAllocsAndFrees()
+			if allocs != frees {
+				t.Errorf("download: alloc'd bytes %v not equal to freed bytes %v", allocs, frees)
```
cojenco, line 3162: Nice tests! Were you able to run pprof on this change?

tritone: Yes, I put together pprof profiles on a parallel download workload and verified the fix. Will add details to the internal bug.

storage/client_test.go
```diff
@@ -3004,6 +3006,165 @@ func TestWriterChunkRetryDeadlineEmulated(t *testing.T) {
...
+
+func (bp *testBufferPool) getAllocsAndFrees() (int64, int64) {
+	bp.Lock()
+	defer bp.Unlock()
+	return bp.allocs, bp.frees
+}
+
+// Test that successful downloads using Reader and MultiRangeDownloader free
+// all of their allocated buffers.
+func TestReadCodecLeaksEmulated(t *testing.T) {
```
BrennaEpp, line 3041: Nice. It would be good to test retries as well here.

tritone: Good idea; I'll do this in a separate PR.

PR #12849, opened 2025-09-10
fix(internal/librariangen): fix CHANGES.md headers

quartzmo:
* Change ### to ## and #### to ###
* Add link to GitHub Release

fixes: googleapis/librarian#2045

### DIFF
--- a/internal/librariangen/release/release.go
+++ b/internal/librariangen/release/release.go
@@ -101,7 +101,7 @@ func updateChangelog(cfg *Config, lib *Library, t time.Time) error {
 		return fmt.Errorf("librariangen: reading changelog: %w", err)
 	}
 
-	versionString := fmt.Sprintf("### %s", lib.Version)
+	versionString := fmt.Sprintf("## %s", lib.Version)
 	if bytes.Contains(oldContent, []byte(versionString)) {
 		slog.Info("librariangen: changelog already up-to-date", "path", relativeChangelogPath, "version", lib.Version)
 		return nil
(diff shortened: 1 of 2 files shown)


internal/librariangen/release/release.go
```diff
@@ -101,16 +101,18 @@ func updateChangelog(cfg *Config, lib *Library, t time.Time) error {
...
 	}
 
 	var newEntry bytes.Buffer
 
-	date := t.Format("2006-01-02")
-	fmt.Fprintf(&newEntry, "%s (%s)\n\n", versionString, date)
+	tag := fmt.Sprintf("%s/v%s", lib.ID, lib.Version)
+	encodedTag := strings.ReplaceAll(tag, "/", "%2F")
+	releaseURL := "https://github.com/googleapis/google-cloud-go/releases/tag/" + encodedTag
+	fmt.Fprintf(&newEntry, "## %s(%s)\n\n", lib.Version, releaseURL)
```
jskeet: I think this should be a Markdown link, so with the version part in square brackets.

I also think the date is useful: I'd expect something like:

```text
## [version](link) (date)
```

e.g.

```text
## [1.2.3](https://github.com/googleapis/google-cloud-go/releases/somemodule%2Fv1.2.3) (2025-09-11)
```

I think that matches the current format other than the content of the link.

jskeet: (We'll also need to make sure that the check for "do we already have this version" is in sync.)

quartzmo: Thank, done!

PR #12859, opened 2025-09-11
chore(librarian): run yamlfmt on state.yaml

parthea:
Ran the following to fix the formatting in `.librarian/state.yaml`

```
go install github.com/google/yamlfmt/cmd/yamlfmt@latest
export PATH="/usr/local/google/home/partheniou/go/bin:$PATH"
yamlfmt .librarian/state.yaml
```

We can run `yamlfmt` any time we make changes to `.librarian/state.yaml` to reduce the diff due to formatting differences as librarian uses `yamlfmt`


codyoss:
This is now outdated. I suggest we fix this upstream in librarian instead and it will be updated on a subsequent update.

PR #12875, opened 2025-09-15
chore: refactor request/response handling and metadata updates

jskeet:
The build and generate commands have requests which are simply a single library in each case, whereas release-init has a dedicated request type containing an array of libraries. The configure command will be similar to release-init.

By sharing the same definition of "a library" across commands, the code to update snippet metadata becomes portable between generate and release-init.

I've left the "request" package name as-is, but we could rename it later if it becomes confusing.

One change in behavior compared with the previous code: we now throw an error if the snippet metadata file doesn't exist, but doesn't contain either a version placeholder or a version number. That would indicate a problem with the snippet metadata file.


jskeet:
@quartzmo The new tests in module_test.go were AI-generated (and then modified by hand), so please review those particularly carefully!

jskeet:
This fixes https://github.com/googleapis/librarian/issues/2023

jskeet:
@quartzmo: I'm happy to rebase this and force-push, or I can merge the current changes and try to resolve them that way.

Issue #12877, opened 2025-09-15
bigquery/v2: query client


alvarowolfx:
Tracking issue for work on the new query client for BigQuery using the new v2 package.

- [ ] QueryClient
- [ ] Query Handler and execution flow
- [ ] Client Options for Billing Project 
- [ ] Query Request helpers
- [ ] Query Parameter helpers
- [ ] Row parsing and RowIterator
- [ ] `adapt` package and row decoding into user defined struct
- [ ] Job retry based on error reason codes
- [ ] Storage Read API acceleration and related Client Options
- [ ] `database/sql` experimental driver

Draft implementation of query client: https://github.com/googleapis/google-cloud-go/pull/12512
Towards internal b/443991686

PR #12878, opened 2025-09-15
feat(bigquery/v2): query client and basic query execution flow

alvarowolfx:
This PR introduces the new query client with basic query execution flow (no retries yet) covering the 3 use cases:
* Running a query using `jobs.query` and obtaining a `Query` handler
* Running a query using `jobs.insert` and obtaining a `Query` handler
* Reattaching to a `Query` handler from a `JobReference` 

Implements long pooling of job status in the background to mark query as complete.

Towards #12877


alvarowolfx:
@shollyman I still have some open questions around `context.Context` propagation, which I see we have a couple of options:
1. Should we re use the `ctx` passed to the `StartQuery` call ? 
2. Should we create a new context based on that one ? 
3. Should we create a new and internal one ? 
4. Should `Wait` accept a `ctx` ? 

The implementation here uses just option 1. In this case when context is cancelled, users have to call `AttachJob` to recreated a `Query` handler. 

If we go with option 2 and/or 3, I think we need to do 4 to allow cancelling `Wait` call.

bigquery/v2/query/helper.go
```diff
@@ -0,0 +1,99 @@
...
+	"cloud.google.com/go/bigquery/v2/apiv2/bigquerypb"
+	"cloud.google.com/go/bigquery/v2/apiv2_client"
+	"cloud.google.com/go/internal/uid"
+	"github.com/googleapis/gax-go/v2"
+	"google.golang.org/api/option"
+)
+
+// Client is a client for running queries in BigQuery.
+type Client struct {
+	c         *apiv2_client.Client
```
shollyman, line 33: Clearly I need to relocate the aggregate client back into apiv2 sooner rather than later.

bigquery/v2/query/client.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+// Client is a client for running queries in BigQuery.
+type Client struct {
+	c         *apiv2_client.Client
+	projectID string
+}
+
+// NewClient creates a new query client.
+func NewClient(ctx context.Context, projectID string, opts ...option.ClientOption) (*Client, error) {
```
shollyman: I had it in my head this wouldn't be constructed using ClientOptions, but users would bring their own instantiated clients.  I feel like there's an opportunity to simplify this with the aggregated client, by either embedding or making a more convenient path to surface a query client.  But we can revisit that later.

bigquery/v2/query/client.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to run query: %w", err)
+	}
+
+	return newQueryJobFromQueryResponse(ctx, c, res)
+}
+
+// StartQueryJob from a bigquerypb.Job definition. Should have job.Configuration.Query filled out.
+func (c *Client) StartQueryJob(ctx context.Context, job *bigquerypb.Job, opts ...gax.CallOption) (*Query, error) {
+	qconfig := job.Configuration.Query
```
shollyman: need to descend in here safely: what if job.Configuration is nil?

bigquery/v2/query/client.go
```diff
@@ -0,0 +1,99 @@
...
+	if qconfig == nil {
+		return nil, fmt.Errorf("job is not a query")
+	}
+
+	if job.Id == "" {
```
shollyman: Do you mean to construct the JobReference instead?  Id is output-only

bigquery/v2/query/client.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// AttachJob set up a query job to be read from an existing one.
+func (c *Client) AttachJob(ctx context.Context, jobRef *bigquerypb.JobReference, opts ...gax.CallOption) (*Query, error) {
+	return newQueryJobFromJobReference(ctx, c, jobRef, opts...)
+}
+
+// Close closes the connection to the API service. The user should invoke this when
+// the client is no longer required.
+func (c *Client) Close() error {
```
shollyman: This gets back to the embedded vs instantiated aspect of client construction.  If the user embeds a client should we close them transitively like this?

bigquery/v2/query/query.go
```diff
@@ -0,0 +1,191 @@
...
+	"context"
+	"sync"
+	"time"
+
+	"cloud.google.com/go/bigquery/v2/apiv2/bigquerypb"
+	"github.com/googleapis/gax-go/v2"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+)
+
+// Query represents a query job.
```
shollyman: Naming consideration:  Should we avoid using existing terms for custom abstractions? Does "Query" mean more or less to BQ users?

bigquery/v2/query/query.go
```diff
@@ -0,0 +1,191 @@
...
+	return newQueryJobFromQueryResponse(ctx, c, res, opts...)
+}
+
+// Read returns a RowIterator for the query results.
+func (q *Query) Read(ctx context.Context, opts ...ReadOption) (*RowIterator, error) {
+	// TODO: proper setup iterator
+	return &RowIterator{}, nil
+}
+
+// Wait waits for the query to complete.
```
shollyman: Per our earlier discussion, the question here is whether to simply surface a done channel idom?

bigquery/v2/query/client.go
```diff
@@ -0,0 +1,99 @@
...
+	if job.Id == "" {
+		job.Id = uid.NewSpace("job", nil).New()
+	}
+
+	job, err := c.c.InsertJob(ctx, &bigquerypb.InsertJobRequest{
+		ProjectId: c.projectID,
+		Job:       job,
+	}, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to insert query: %w", err)
```
shollyman: Should we provide an option to silently reattach to a query if it exists?

bigquery/v2/query/client.go
```diff
@@ -0,0 +1,99 @@
...
+	return qc, nil
+}
+
+// StartQuery runs a query and returns a QueryJob handle.
+func (c *Client) StartQuery(ctx context.Context, req *bigquerypb.PostQueryRequest, opts ...gax.CallOption) (*Query, error) {
+	if req.QueryRequest.RequestId == "" {
+		req.QueryRequest.RequestId = uid.NewSpace("request", nil).New()
+	}
+
+	res, err := c.c.Query(ctx, req, opts...)
```
shollyman: So the construction here means we wait until the first RPC returns before we can return the query, which means Start can block for an extended time.  Should we return more quickly even it means things like Query ID or job ID can't immediately be resolved?  It also means you can probably just omit the returned error here, and move responsibility for the user to check the error on the query itself.

bigquery/v2/query/client.go
alvarowolfx: duh, wrong field 🤦

bigquery/v2/query/client.go
shollyman: Similarly, use the opaque style for proto accesses, e.g. GetConfiguration()

bigquery/v2/query/query.go
leahecole: +1 if this is a query job, imho it should be named QueryJob - the query is the actual DDL syntax parsed by BQ

bigquery/v2/query/query.go
```diff
@@ -0,0 +1,191 @@
...
+		if jobRef.Location != nil {
+			q.location = jobRef.Location.GetValue()
+		}
+	}
+
+	if res.TotalRows != nil {
+		q.cachedTotalRows = res.TotalRows.Value
+	}
+
+	// TODO: save schema, page token, total rows and parse rows
```
leahecole, line 257: still needed?

leahecole (commented):
I'm less useful here, but just left one comment +1-ing the naming thing, and am wanting to double check the TODOs are intentional and will be addressed in followup PRs

bigquery/v2/query/client.go
alvarowolfx: I think we need to accept ClientOptions here for extensibility, for example, I already added a ClientOption `WithClient` that users can pass their AggregatedClient and we can add more to accept the other sub clients.

bigquery/v2/query/client.go
alvarowolfx: good point, I'll explore how that would look

bigquery/v2/query/client.go
alvarowolfx: good call, maybe keep track of who owns the client here.

bigquery/v2/query/query.go
alvarowolfx: you both have a point, but does Job more or less to BQ users ? I would say that for new users, seeing a Job here might feel weird, but can be less of a problem as it is going to be called a `QueryJob`.

bigquery/v2/query/query.go
alvarowolfx: yeah, there are more fields to be filled here, but they are only gonna be done when the row parsing and iteration is added

bigquery/v2/query/client.go
shollyman: Another thing we could do here is something akin to getting a query client from an instantiated aggregate client.  The advantage there is we never have the query client responsible for closing clients because it is never the owning entity.

shollyman: I also wonder if we should call it something other than Client if we're clearer about the relationships between this and the underlying rpc clients.

bigquery/v2/query/helper.go
```diff
@@ -62,9 +62,13 @@ func (c *Client) StartQuery(ctx context.Context, req *bigquerypb.PostQueryReques
 
 // StartQueryJob from a bigquerypb.Job definition. Should have job.Configuration.Query filled out.
 func (c *Client) StartQueryJob(ctx context.Context, job *bigquerypb.Job, opts ...gax.CallOption) (*Query, error) {
-	qconfig := job.Configuration.Query
+	config := job.GetConfiguration()
+	if config == nil {
+		return nil, fmt.Errorf("bigquery: job is missing configuration")
+	}
+	qconfig := config.Query
```
shollyman, line 73: config.GetQuery()

bigquery/v2/query/client.go
alvarowolfx: I called it so in Go world that would be a `query.Client`, so it's namespaced on the packged. But open to other naming ideas. We can also change later since this is in preview

bigquery/v2/query/helper.go
alvarowolfx: done

bigquery/v2/query/query.go
```diff
@@ -236,7 +224,15 @@ func (q *Query) waitForQuery(ctx context.Context, opts []gax.CallOption) error {
...
-func (q *Query) consumeQueryResponse(res *bigquerypb.GetQueryResultsResponse) {
+// Common fields from jobs.query and jobs.getQueryResults
+// Needs to be updated as new fields are consumed
+type queryResponse interface {
+	GetJobComplete() *wrapperspb.BoolValue
+	GetJobReference() *bigquerypb.JobReference
+	GetTotalRows() *wrapperspb.UInt64Value
+}
+
+func (q *Query) consumeQueryResponse(res queryResponse) {
```
shollyman, line 236: I wonder if we should try to be clever with descriptors to solve this problem rather than using a bunch of internal interfaces. You could potentially just pass in a proto message and have it parse based on some combo of (field name and output type) matching or similar?  Tag ID probably isn't going to get us much, as they're not in sync across messages to my recollection.

bigquery/v2/query/client.go
shollyman: Yeah, I think this gets back to whether we call it a client.  If its not a client, we don't necessarily need a Close().  Do we think we'll ever be hanging more state off the QueryClient other than basic defaults?  Will it keep track of instantiated queries, etc?

bigquery/v2/query/query.go
alvarowolfx: I think it's an optimization that we can do later, to make it more generic and evolve as new fields are added, but even in the "final" implementation from PR https://github.com/googleapis/google-cloud-go/pull/12512/files#diff-f3d9e961ad56f6293819b28db908d8b71fa774e0312339b86eeeb8843fe29805, we only consume 8 common fields, so I think it's fine to add to the internal interface. I'm not seeing a big reason to make it more generic now, unless I'm missing something.

bigquery/v2/query/client.go
alvarowolfx: The plan for now it's to keep just settings and defaults like you have pointed out. The problem is that without a Close method, users can't close the internal `apiv2_client.Client` and free resources. This also makes passing a `apiv2_client.Client` a required argument to create this `query.Client`

bigquery/v2/query/query.go
alvarowolfx: This approach provided by @shollyman is a good approach for making the query stats available to customer later. We can do this later when making those stats public

bigquery/v2/query/query.go
```diff
@@ -0,0 +1,294 @@
...
+	"time"
+
+	"cloud.google.com/go/bigquery/v2/apiv2/bigquerypb"
+	"github.com/googleapis/gax-go/v2"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+)
+
+// Query represents a handle to a query job. Its methods can be used to wait for
+// the job to complete and to iterate over the results.
+type Query struct {
```
alvarowolfx, line 29: maybe have an internal `job` package that can be reused for LOAD/EXTRACT/COPY jobs and later can be made public.

bigquery/v2/query/helper.go
```diff
@@ -0,0 +1,101 @@
...
+// Helper for running queries in BigQuery. It is a lightweight wrapper
+// around the auto-generated BigQuery v2 client, focused on query operations.
+type Helper struct {
+	c         *apiv2_client.Client
+	projectID string
+}
+
+// NewHelper creates a new query helper. This helper should be reused instead of
+// created per-request.
+func NewHelper(c *apiv2_client.Client, projectID string, opts ...option.ClientOption) (*Helper, error) {
```
shollyman, line 39: something for followup in subsequent PRs:  should it accept ClientOption if it's not instantiating clients?

bigquery/v2/query/helper.go
```diff
@@ -0,0 +1,101 @@
...
+
+// StartQuery executes a query using the stateless jobs.query RPC. It returns a
+// handle to the running query. The returned Query object can be used to wait for
+// completion and retrieve results.
+func (h *Helper) StartQuery(ctx context.Context, req *bigquerypb.PostQueryRequest, opts ...gax.CallOption) (*Query, error) {
+	qr := req.GetQueryRequest()
+	if qr == nil {
+		return nil, fmt.Errorf("bigquery: request is missing QueryRequest")
+	}
+	if qr.GetRequestId() == "" {
```
shollyman, line 59: If you're modifying the request, perhaps we should clone then modify.  Same feedback for the other methods below where we're monkeying with requests.

bigquery/v2/query/query.go
```diff
@@ -0,0 +1,294 @@
...
+//			// Query is still running.
+//		}
+func (q *Query) Done(opts ...gax.CallOption) <-chan struct{} {
+	return q.ready
+}
+
+// Err returns the final error state of the query. It is only valid to call Err
+// after the channel returned by Done has been closed. If the query completed
+// successfully, Err returns nil.
+func (q *Query) Err() error {
```
shollyman, line 131: Should this block on done if it's only valid after?

bigquery/v2/query/query.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func (q *Query) waitForQuery(ctx context.Context, opts []gax.CallOption) error {
+	res, err := q.h.c.GetQueryResults(ctx, &bigquerypb.GetQueryResultsRequest{
+		ProjectId:  q.projectID,
+		JobId:      q.jobID,
+		Location:   q.location,
+		MaxResults: wrapperspb.UInt32(0),
+		FormatOptions: &bigquerypb.DataFormatOptions{
+			UseInt64Timestamp: true,
```
shollyman, line 217: We should migrate to the new field.  I'll refresh the protos to make sure its present.

shollyman (approved):
Let's get this in and continue iterating, I think its in much better shape now.

shollyman (approved):

shollyman (approved):

PR #12879, opened 2025-09-15
chore(storage): Instrument existing media ops with storage trace package

cojenco:
Replace existing media ops tracing with storage trace package

As aligned in previous discussions, this supports an overarching span on Reader and Writer until closed. This replaces the remaining legacy tracing in copier and composer as well, completely transitioning to using the storage trace package

### DIFF
--- a/storage/copy.go
+++ b/storage/copy.go
@@ -18,8 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-
-	"cloud.google.com/go/internal/trace"
 )
 
 // CopierFrom creates a Copier that can copy src to dst.


storage/reader.go
```diff
@@ -195,8 +193,6 @@ func (o *ObjectHandle) NewMultiRangeDownloader(ctx context.Context) (mrd *MultiR
 	// span now if there is an error.
 	if err == nil {
 		r.ctx = ctx
-	} else {
-		trace.EndSpan(ctx, err)
 	}
```
BrennaEpp, line 196: Why are we removing this?

cojenco: This reduces the risk of orphaned spans by adding the deferred pattern above. I tested with larger object sizes and through context propagation, the parent span still serves as an overarching span for the reader

PR #12881, opened 2025-09-16
fix(internal/librariangen): add link to source commit in release notes

quartzmo:
fixes: googleapis/librarian#2105

New output:

```
## [1.16.0](https://github.com/googleapis/google-cloud-go/releases/tag/secretmanager%2Fv1.16.0) (2025-09-11)

### Features

* add new GetSecret API ([abcdef1](https://github.com/googleapis/google-cloud-go/commit/abcdef123456))
* another feature ([zxcvbn0](https://github.com/googleapis/google-cloud-go/commit/zxcvbn098765))

### Bug Fixes

* correct typo in documentation ([123456a](https://github.com/googleapis/google-cloud-go/commit/123456abcdef))
```

### DIFF
--- a/internal/librariangen/release/release.go
+++ b/internal/librariangen/release/release.go
@@ -148,10 +148,10 @@ func updateChangelog(cfg *Config, lib *Library, t time.Time) error {
 					shortHash = shortHash[:7]
 				}
 				commitURL := fmt.Sprintf("https://github.com/googleapis/google-cloud-go/commit/%s", change.SourceCommitHash)
-				commitLink = fmt.Sprintf(" ([%s](%s))", shortHash, commitURL)
+				commitLink = fmt.Sprintf("([%s](%s))", shortHash, commitURL)
 			}
 
-			fmt.Fprintf(&newEntry, "* **%s:** %s%s\n", lib.ID, change.Subject, commitLink)
+			fmt.Fprintf(&newEntry, "* %s %s\n", change.Subject, commitLink)
 
 		}
 		newEntry.WriteString("\n")
--- a/internal/librariangen/release/release.go
+++ b/internal/librariangen/release/release.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/librariangen/release/release.go
```diff
@@ -117,22 +118,41 @@ func updateChangelog(cfg *Config, lib *Library, t time.Time) error {
...
 
-	changesByType := make(map[string]map[string]bool)
+	changesByType := make(map[string]map[string]*Change)
 	for _, change := range lib.Changes {
 		if changesByType[change.Type] == nil {
-			changesByType[change.Type] = make(map[string]bool)
+			changesByType[change.Type] = make(map[string]*Change)
 		}
-		changesByType[change.Type][change.Subject] = true
+		changesByType[change.Type][change.Subject] = change
```
jskeet, line 126: Just to check - if we have multiple changes with the same title for some reason, we'll only include one of them, right? I think that's probably fine. (And that was the case before, I expect.)

internal/librariangen/release/release.go
```diff
@@ -117,22 +118,41 @@ func updateChangelog(cfg *Config, lib *Library, t time.Time) error {
...
+			if change.SourceCommitHash != "" {
+				shortHash := change.SourceCommitHash
+				if len(shortHash) > 7 {
+					shortHash = shortHash[:7]
+				}
+				commitURL := fmt.Sprintf("https://github.com/googleapis/google-cloud-go/commit/%s", change.SourceCommitHash)
+				commitLink = fmt.Sprintf(" ([%s](%s))", shortHash, commitURL)
+			}
+
+			fmt.Fprintf(&newEntry, "* **%s:** %s%s\n", lib.ID, change.Subject, commitLink)
```
jskeet: Do we need the library ID in here? Isn't it going to be the same for all entries? (We're creating a change log *for that library*.)

internal/librariangen/release/release_test.go
```diff
@@ -153,11 +151,11 @@ func TestInit(t *testing.T) {
 			requestJSON: `{ "libraries": [ { "id": "secretmanager", "version": "1.16.0", "release_triggered": true, "apis": [{"path": "google/cloud/secretmanager/v1"}], "changes": [{"type": "feat", "subject": "add new GetSecret API"}] } ] }`,
 			initialRepoContent: map[string]string{
 				"secretmanager/CHANGES.md": "# Changes\n\n## [1.16.0](https://github.com/googleapis/google-cloud-go/releases/tag/secretmanager%2Fv1.16.0)\n- Already there.",
-				"internal/generated/snippets/secretmanager/apiv1/snippet_metadata.google.cloud.secretmanager.v1.json": `{"version": "1.15.0"}`,
+				"internal/generated/snippets/secretmanager/apiv1/snippet_metadata.google.cloud.secretmanager.v1.json": `{\"version\": \"1.15.0\"}`,
```
jskeet: I'm confused about this part - we wouldn't actually expect a backslash in the JSON file, would we? (It's possible that I don't understand Go string literals well enough - I'd expected that something in backticks didn't need escaping.)

internal/librariangen/release/release.go
quartzmo: I was just trying to match the existing format. But I agree it can be removed. Will do so.

internal/librariangen/release/release.go
quartzmo: I agree, we should keep this implicit de-duping.

internal/librariangen/release/release_test.go
quartzmo: Good catch! Fixed.

jskeet:
Ah - looks like you'll need Cody to review as well. I'm not in yoshi-go-admins...

jskeet:
(Closure was accidental - the buttons are really close together!)

internal/librariangen/release/release_test.go
```diff
@@ -136,12 +137,9 @@ func TestInit(t *testing.T) {
 				"secretmanager/internal/version.go": `package internal; const Version = "1.15.0"`,
 				"internal/generated/snippets/secretmanager/apiv1/snippet_metadata.google.cloud.secretmanager.v1.json": `{"version": "1.15.0"}`,
 			},
-			// This wantChangelogSubstr expectation intentionally includes two newline literals.
-			wantChangelogSubstr: `## [1.16.0](https://github.com/googleapis/google-cloud-go/releases/tag/secretmanager%2Fv1.16.0) (2025-09-11)
-
-`,
-			wantVersion:        "1.16.0",
-			wantSnippetVersion: `"version": "1.16.0"`,
+			wantChangelogSubstr: "## [1.16.0](https://github.com/googleapis/google-cloud-go/releases/tag/secretmanager%2Fv1.16.0) (2025-09-11)\n\n### Features\n\n* add new GetSecret API ([abcdef1](https://github.com/googleapis/google-cloud-go/commit/abcdef123456))\n* another feature ([zxcvbn0](https://github.com/googleapis/google-cloud-go/commit/zxcvbn098765))\n\n### Bug Fixes\n\n* correct typo in documentation ([123456a](https://github.com/googleapis/google-cloud-go/commit/123456abcdef))\n\n",
```
codyoss, line 140: nit: this is hard to read I would suggest using backticks and or a dedicated testdata file for blobs of text.

codyoss (approved):

PR #12893, opened 2025-09-17
fix(firestore): Correct ReadWrite transaction retries

bhshkh:
Retry logic has been changed as per [go/transaction-retry-matrix](http://go/transaction-retry-matrix)

### DIFF
--- a/firestore/transaction.go
+++ b/firestore/transaction.go
@@ -102,6 +102,28 @@ var (
 
 type transactionInProgressKey struct{}
 
+// isTransactionRetryable checks if an error from a transaction operation
+// indicates that the entire transaction should be retried.
+func isTransactionRetryable(err error) bool {
+	s, ok := status.FromError(err)
+	if !ok {
+		return false // Not a gRPC error
+	}
+	switch s.Code() {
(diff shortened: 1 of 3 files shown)


firestore/transaction.go
```diff
@@ -102,6 +102,28 @@ var (
...
 
+// isTransactionRetryable checks if an error from a transaction operation
+// indicates that the entire transaction should be retried.
+func isTransactionRetryable(err error) bool {
+	s, ok := status.FromError(err)
+	if !ok {
+		return false // Not a gRPC error
+	}
+	switch s.Code() {
+	case codes.Aborted,
```
JDrit: In general we can only safely retry the transaction's commit when we are guaranteed to know that the transaction did not complete under the hood. For example if the `Firestore.commit` errors returns a deadline or unavailable error you don't know if the transaction actually succeed under the hood.

To that end, I would just rename this function `isAborted` and only return `true` for `codes.Aborted`.

bhshkh: Renamed

JDrit (approved):

hongalex (approved):

Issue #8933, opened 2023-10-30
spanner: support generated column with `spanner#Row.ToStruct`, `spanner.(Insert|Update|InsertOrUpdate)Struct` 


vvakame:
**Is your feature request related to a problem? Please describe.**
I use spanner with struct. I want to try generated column, but spanner library’s struct operation can’t do that.
generated column can’t be wrote value, I use `spanner:”-“` to struct fields. but It can’t read generated column to struct field.

**Describe the solution you'd like**
introduce `spanner:",generated"` tag to struct field. likes `encoding/json` 's `json:",omitempty"`.
I want it to be ignored for spanner.InsertStruct, spanner.UpdateStruct, and spanner.InsertOrUpdateStruct, and be the target of loading for spanner#Row.ToStruct.

**Describe alternatives you've considered**
use `spanner.Insert` (not Struct) instead. likes [yo](https://github.com/cloudspannerecosystem/yo/blob/v0.5.4/templates/type.go.tpl#L121-L126) did.
However, I would like a simple solution like the proposed method, if possible, because it would require the introduction and development of new tools.

**Additional context**
none.

PR #12895, opened 2025-09-17
feat(spanner): support "generated" column tag parsing for Go struct operations

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/8933


spanner/mutation.go
```diff
@@ -198,6 +198,11 @@ func structToMutationParams(in interface{}) ([]string, []interface{}, error) {
 	var cols []string
 	var vals []interface{}
 	for _, f := range fields {
+		if f.ParsedTag != nil {
+			if tag, ok := f.ParsedTag.(spannerTag); ok && tag.Generated {
```
olavloite: nit: I would prefer the more generic `ReadOnly` instead of `Generated`. That makes it clearer what it is for.

spanner/mutation_test.go
```diff
@@ -430,6 +432,139 @@ func TestStructToMutationParams(t *testing.T) {
 	}
 }
 
+func TestStructToMutationParams_Generated(t *testing.T) {
+	type Generated struct {
+		ID   int64
+		Name string `spanner:",generated"`
```
olavloite: Why is the preceding comma needed here? Why can't it just be `spanner:"generated"`?

spanner/mutation_test.go
```diff
@@ -430,6 +432,139 @@ func TestStructToMutationParams(t *testing.T) {
 	}
 }
 
+func TestStructToMutationParams_Generated(t *testing.T) {
+	type Generated struct {
```
olavloite: nit: mark test as parallel

spanner/mutation_test.go
```diff
@@ -430,6 +432,139 @@ func TestStructToMutationParams(t *testing.T) {
...
+	wantCols := []string{"Id", "FirstName", "LastName"}
+	if !reflect.DeepEqual(write.Columns, wantCols) {
+		t.Fatalf("write.Columns mismatch\ngot  %v\nwant %v", write.Columns, wantCols)
+	}
+	if len(write.Values) != 1 {
+		t.Fatalf("write.Values length mismatch: got %d, want 1", len(write.Values))
+	}
+	if got := len(write.Values[0].Values); got != len(wantCols) {
+		t.Fatalf("write.Values[0] length mismatch: got %d, want %d", got, len(wantCols))
+	}
```
olavloite: nit: In general, these types of checks can best be written like this:

```
if g, w := len(write.Values[0].Values), len(wantCols); g != w {
  t.Fatalf("write.Values[0] length mismatch\n Got: %v\nWant: %v", g, w)
}
```

This has a couple of advantages:
1. It is less likely that you make a typo in either the if or the t.Fatalf call (e.g. in this case `len(wantCols)` is repeated in both, but a small error in a copy-paste action could easily mess that up)
2. By using a newline and aligning the ` Got` and `Want`, it is easier to spot the actual difference between the two values, especially if they are a bit long strings, and the only difference is for example a whitespace character.

spanner/value.go
```diff
@@ -5592,12 +5592,26 @@ func encodeProtoEnumArray(len int, at func(int) reflect.Value) (*proto3.Value, e
...
 	if s := t.Get("spanner"); s != "" {
 		if s == "-" {
 			return "", false, nil, nil
 		}
-		return s, true, nil, nil
+		parts := strings.Split(s, ",")
+		name = parts[0]
+		tag := spannerTag{}
+		for _, part := range parts[1:] {
+			if part == "generated" {
```
olavloite: As mentioned above; Should we consider using `readOnly` instead for this? Or is there prior art / precedence from other databases or frameworks that indicate that `generated` is a more common term?

spanner/value.go
```diff
@@ -5592,12 +5592,26 @@ func encodeProtoEnumArray(len int, at func(int) reflect.Value) (*proto3.Value, e
...
 func spannerTagParser(t reflect.StructTag) (name string, keep bool, other interface{}, err error) {
 	if s := t.Get("spanner"); s != "" {
 		if s == "-" {
 			return "", false, nil, nil
 		}
-		return s, true, nil, nil
+		parts := strings.Split(s, ",")
+		name = parts[0]
+		tag := spannerTag{}
+		for _, part := range parts[1:] {
```
olavloite, line 5613: Why do we skip the first result here? Is it because it should always be parsed as the column name?

Can we add some documentation (code comments) that explains what can go into this tag, and how it should be used (including this new tag)?

spanner/mutation_test.go
rahul2393: I am keeping the first value before comma to be parsed as alternate name which is backward compatible,  and comma will be separator for future spanner tag values similar to how GORM has
https://gorm.io/docs/models.html#Fields-Tags

> Why can't it just be spanner:"generated"?

It will be a breaking change since right now we treat the value as alternate name to be looked up in spanner rows

spanner/value.go
rahul2393: >Why do we skip the first result here? Is it because it should always be parsed as the column name?

Yes, and if we don't skip it, it will be a breaking change for existing users.

spanner/value.go
rahul2393: If we talk about GORM they support something like
```
  Name string `gorm:"->"`        // readonly (disable write permission unless it configured)
```

spanner/mutation.go
rahul2393: Done

spanner/value.go
rahul2393: Added doc.go, now instead of comma optional tags will be separate by a semicolan (;) and `->` is supported similar to GORM https://gorm.io/docs/models.html#Fields-Tags

spanner/value.go
rahul2393: Updated PR to use `readOnly` as well as how GORM parses this `->`

PR #12910, opened 2025-09-17
fix(compute/metadata): disable Client timeouts for subscription client

codyoss:
Related: #12885
Internal Bug: 427030684


codyoss:
@quartzmo It should not, no. In fact, it should help users if they are using the Subscribe method. Subscribe requests all long polling, having timeouts set would improperly kill the connection potentially before data is ready to be returned. At least that is my understanding.

quartzmo (approved):

PR #12919, opened 2025-09-19
fix(storage): copy metadata when using Copier with grpc

BrennaEpp:
We need to not set `destination` on the request unless the user specifies attrs. Since the struct is not a pointer, we have to check if it is empty to determine if it has not been set.


tritone (commented):
Looks like a good approach but Copier retry tests seem to be failing now in the CI. Can you look into this?

storage/integration_test.go
```diff
@@ -2521,6 +2529,16 @@ func TestIntegration_Copy(t *testing.T) {
 					}
 				}
 
+				// Check that metadata is copied when no destination attributes are provided.
+				if test.copierAttrs == nil {
```
tritone: nit: should just be an `else` on the above? I would expect one or the other.

tritone (dismissed):
Nice approach, thanks for the changes.

PR #12944, opened 2025-09-24
feat(librariangen): first part of implementing configure

jskeet:
This is effectively the skeleton around the real work, which is yet to be implemented.

Towards https://github.com/googleapis/librarian/issues/981

### DIFF
--- a/internal/librariangen/configure/response.go
+++ b/internal/librariangen/configure/response.go
@@ -16,7 +16,6 @@ package configure
 
 import (
 	"encoding/json"
-	"log/slog"
 	"os"
 
 	"cloud.google.com/go/internal/postprocessor/librarian/librariangen/request"
@@ -26,11 +25,7 @@ import (
 func saveResponse(response *request.Library, path string) error {
 	b, err := json.MarshalIndent(response, "", "  ")
 	if err != nil {
-		slog.Error("failed to marshal response", "error", err)
 		return err
 	}
-	if err := os.WriteFile(path, b, 0644); err != nil {
-		slog.Error("failed to write response", "error", err)
-	}
-	return err
+	return os.WriteFile(path, b, 0644)
 }


jskeet:
Will fix the vet issues when I get a chance.

jskeet:
Okay, vetting is fixed, ready for review. Note that there are no tests for request deserialization/response serialization at the moment. We could add them now, or wait until we're actually implementing the command fully.

internal/librariangen/configure/response.go
```diff
@@ -0,0 +1,36 @@
...
+func saveResponse(response *request.Library, path string) error {
+	b, err := json.MarshalIndent(response, "", "  ")
+	if err != nil {
+		slog.Error("failed to marshal response", "error", err)
+		return err
+	}
+	if err := os.WriteFile(path, b, 0644); err != nil {
+		slog.Error("failed to write response", "error", err)
+	}
+	return err
```
quartzmo: This `return err` should be moved into the if block above. The way it's written here, the `err` var initialized on line 32 will not be returned, since it is scoped to the block.

```go
	if err := os.WriteFile(path, b, 0644); err != nil {
		slog.Error("failed to write response", "error", err)
		return err
	}
	return nil
```

internal/librariangen/configure/response.go
```diff
@@ -0,0 +1,36 @@
...
+	"os"
+
+	"cloud.google.com/go/internal/postprocessor/librarian/librariangen/request"
+)
+
+// saveResponse marshals a Library struct, and writes it to configure-response.json file.
+func saveResponse(response *request.Library, path string) error {
+	b, err := json.MarshalIndent(response, "", "  ")
+	if err != nil {
+		slog.Error("failed to marshal response", "error", err)
```
quartzmo: Why log this here when it is also returned (propagated) up the call chain?

jskeet: I strongly suspect both of these were due to basically copying the code used to write out an *error* response - which has different error semantics, as we're already in an error situation. Fixed in an extra commit.

internal/librariangen/configure/response.go
jskeet: (Also fixed in the new commit.)

PR #12945, opened 2025-09-24
fix(storage): skip download of file outside of target dir

krishnamd-jkp:
This PR introduces logic in transfermanager download directories flow where the objects are skipped if the download path is outside of target directory. The changes directly add the results of such objects using addResult method bypassing the addition of such objects to workers channel


storage/transfermanager/downloader.go
```diff
@@ -948,3 +959,23 @@ func checksumObject(got, want uint32) error {
 	}
 	return nil
 }
+
+func isSubPath(localDirectory, filePath string) (bool, error) {
+	absLocalDirectory, err := filepath.Abs(localDirectory)
+	if err != nil {
+		return false, fmt.Errorf("cannot convert local directory to absolute path: %v", err)
```
cpriti-os: Consider using %w instead of %v, here and below to wrap the original error. Can help with debugging in the future.

storage/transfermanager/downloader.go
```diff
@@ -948,3 +959,23 @@ func checksumObject(got, want uint32) error {
...
+	if err != nil {
+		return false, fmt.Errorf("cannot convert file path to absolute path: %v", err)
+	}
+	rel, err := filepath.Rel(absLocalDirectory, absFilePath)
+	if err != nil {
+		return false, err
+	}
+
+	// Check if rel starts with ".."
+	isUnder := !strings.HasPrefix(rel, ".."+string(filepath.Separator)) && rel != ".."
```
cpriti-os: Since the logic in this line is a little complex, consider breaking it down for better readability or at least adding a appropriate comment to help readers.

cpriti-os: Nit: Space before and after operator.

cpriti-os (changes requested):

storage/transfermanager/downloader_test.go
```diff
@@ -232,6 +233,97 @@ func TestNumShards(t *testing.T) {
...
+
+	// Create a temporary directory to work with relative paths.
+	tempDir := t.TempDir()
+
+	testCases := []struct {
+		name           string
+		localDirectory string
+		filePath       string
+		wantIsSub      bool
+		wantErr        bool
```
tritone: This seems to be unused? It would make sense to have at least one test case that does return an error probably.

storage/transfermanager/downloader.go
```diff
@@ -191,8 +192,18 @@ func (d *Downloader) DownloadDirectory(ctx context.Context, input *DownloadDirec
...
 		filePath := filepath.Join(input.LocalDirectory, objectWithoutPrefix)
 
+		// Prevent directory traversal attacks.
+		isUnder, err := isSubPath(input.LocalDirectory, filePath)
+		if err != nil {
+			cleanFiles(inputs)
+			return fmt.Errorf("transfermanager: DownloadDirectory failed to verify path: %w", err)
+		}
+		if !isUnder {
+			log.Printf("transfermanager: skipping object with unsafe path after stripping prefix %q", objectWithoutPrefix)
```
tritone: Are you sure this is sufficient? When using a client library it's unlikely that a human would necessarily be checking the log output.

I would expect the rest of the download job to continue, but that the Downloader returns an error to indicate which file names have been skipped. That way the calling application can programmatically understand that some objects were missed in the download.

storage/transfermanager/downloader.go
```diff
@@ -948,3 +959,27 @@ func checksumObject(got, want uint32) error {
 	}
 	return nil
 }
+
+func isSubPath(localDirectory, filePath string) (bool, error) {
+	// validate paths
+	absLocalDirectory, err := filepath.Abs(localDirectory)
+	if err != nil {
+		return false, fmt.Errorf("cannot convert local directory to absolute path: %w", err)
```
cpriti-os, line 987: Add test cases for this and other errors below

storage/transfermanager/downloader.go
```diff
@@ -948,3 +959,27 @@ func checksumObject(got, want uint32) error {
 	}
 	return nil
 }
+
+func isSubPath(localDirectory, filePath string) (bool, error) {
+	// validate paths
```
cpriti-os: This comment could be more descriptive.

storage/transfermanager/downloader.go
```diff
@@ -948,3 +959,27 @@ func checksumObject(got, want uint32) error {
...
+	absLocalDirectory, err := filepath.Abs(localDirectory)
+	if err != nil {
+		return false, fmt.Errorf("cannot convert local directory to absolute path: %w", err)
+	}
+	absFilePath, err := filepath.Abs(filePath)
+	if err != nil {
+		return false, fmt.Errorf("cannot convert file path to absolute path: %w", err)
+	}
+
+	// absLocalDirectory + rel = absFilePath
```
cpriti-os: This comment is an equation, maybe explaining this in a sentence would be more helpful as a comment.

storage/transfermanager/downloader_test.go
cpriti-os: To make the tests more robust, check for expected error messages when wantErr is true if feasible.

krishnamd-jkp: Added a new test case for error

storage/transfermanager/downloader.go
krishnamd-jkp: Added the errors in download result. This will now be populated in downloader.errors

storage/transfermanager/downloader.go
krishnamd-jkp: Done

storage/transfermanager/downloader.go
krishnamd-jkp: Changed

storage/transfermanager/downloader.go
krishnamd-jkp: Provided an example

storage/transfermanager/downloader_test.go
krishnamd-jkp: Currently I'm throwing only one error. Will add wantErrMsg if the function involved more error paths

storage/transfermanager/downloader.go
krishnamd-jkp: Also, I've run the integration tests of transfer manager.

tritone: Sounds good. Since you are returning errors, let's remove the log.Printf statement. Generally we should not be doing this in the SDK (except if the user asks for debug-level logging).

storage/transfermanager/downloader_test.go
```diff
@@ -232,6 +234,113 @@ func TestNumShards(t *testing.T) {
...
+			})
+			// induce filepath.Abs() error
+			if tc.wantErr {
+				dir, _ := os.MkdirTemp("", "")
+				os.Chdir(dir)
+				os.RemoveAll(dir)
+			}
+			isSub, err := isSubPath(tc.localDirectory, tc.filePath)
+
+			if (err != nil) != tc.wantErr {
```
tritone: Instead of this check, maybe check that the actual error returned is what you expect it to be?

storage/transfermanager/downloader.go
```diff
@@ -218,9 +239,21 @@ func (d *Downloader) DownloadDirectory(ctx context.Context, input *DownloadDirec
...
+	for _, file := range illegalPathObjects {
+		// the waitgroup is further decremented in addResult method
+		d.downloadsInProgress.Add(1)
+		d.addResult(&file, &DownloadOutput{
+			Bucket:  file.Bucket,
+			Object:  file.Object,
+			Err:     fmt.Errorf("skipping download of object with unsafe path %q", file.Object),
+			skipped: true,
+		})
+	}
```
tritone, line 254: Would it be possible to add an integration test case for this? Just want to make sure that we are capturing the end-to-end on this and not e.g. inadvertently introducing a deadlock.

krishnamd-jkp: Added integration tests

storage/transfermanager/downloader_test.go
krishnamd-jkp: Done

storage/transfermanager/downloader_test.go
```diff
@@ -232,6 +234,118 @@ func TestNumShards(t *testing.T) {
...
+
+	// Create a temporary directory to work with relative paths.
+	tempDir := t.TempDir()
+
+	testCases := []struct {
+		name             string
+		localDirectory   string
+		filePath         string
+		wantIsSub        bool
+		wantErr          bool
```
cpriti-os: do we need a bool or can we use `nil` as wantErr=false?

storage/transfermanager/downloader.go
```diff
@@ -948,3 +979,28 @@ func checksumObject(got, want uint32) error {
 	}
 	return nil
 }
+
+func isSubPath(localDirectory, filePath string) (bool, error) {
+	// validate if paths can be converted to absolute paths
```
cpriti-os: Nit: go/go-style/decisions#comment-sentences

cpriti-os (changes requested):

storage/transfermanager/integration_test.go
```diff
@@ -456,6 +457,288 @@ func TestIntegration_DownloadDirectoryError(t *testing.T) {
...
+					t.Errorf("os.Open(%q): %v", path, err)
+				}
+				defer f.Close()
+
+				b := bytes.NewBuffer(make([]byte, 0, got.Attrs.Size))
+				if _, err := io.Copy(b, f); err != nil {
+					t.Errorf("io.Copy: %v", err)
+				}
+
+				if wantCRC, gotCRC := tb.contentHashes[got.Object], crc32c(b.Bytes()); gotCRC != wantCRC {
```
tritone: I think this check shouldn't be necessary? The storage library code should check the CRC implicitly. I think you can just skip all of L514-L534.

storage/transfermanager/integration_test.go
```diff
@@ -456,6 +457,288 @@ func TestIntegration_DownloadDirectoryError(t *testing.T) {
...
+					t.Errorf("Expected error but got nil for object %v: %v", got.Object, got.Err)
+				}
+				// verify that file path has not been created
+				if _, err := os.Stat(filepath.Join(localDir, got.Object)); err == nil {
+					t.Errorf("Expected error but got nil for object %v: %v", got.Object, err)
+				}
+				continue
+			}
+
+			if got, want := got.Attrs.Size, tb.objectSizes[got.Object]; want != got {
```
tritone: Again, checks here and below can be skipped for this test I think. It would be more useful to just validate the filenames in our local dir are what we expect.

storage/transfermanager/integration_test.go
```diff
@@ -456,6 +457,288 @@ func TestIntegration_DownloadDirectoryError(t *testing.T) {
...
+				if _, err := io.Copy(b, f); err != nil {
+					t.Errorf("io.Copy: %v", err)
+				}
+
+				if wantCRC, gotCRC := tb.contentHashes[got.Object], crc32c(b.Bytes()); gotCRC != wantCRC {
+					t.Errorf("object(%q) at filepath(%q): content crc32c does not match; got: %v, expected: %v", got.Object, path, gotCRC, wantCRC)
+				}
+				got.Object = "modifying this shouldn't be a problem"
+			},
+			Callback: func(outs []DownloadOutput) {
```
tritone, line 619: For the async case maybe let's just test that the callback shows the errors we expect for the illegal files and that the other objects got downloaded successfully? IMO this test is duplicating work done elsewhere, let's pare it back somewhat.

tritone (commented):
Couple comments on the tests you added, otherwise looks good.

storage/transfermanager/downloader_test.go
krishnamd-jkp: Done

storage/transfermanager/downloader.go
krishnamd-jkp: Done

storage/transfermanager/integration_test.go
krishnamd-jkp: Thought I'd check the integrity of the downloaded files. Yeah but I think it might not be the right test for that.

storage/transfermanager/integration_test.go
krishnamd-jkp: Done

PR #12950, opened 2025-09-24
chore(internal/stategen): add module prep and OwlBot/release-please cleanup

quartzmo:


internal/stategen/cleanup.go
```diff
@@ -0,0 +1,225 @@
...
+	for _, f := range jsonFiles {
+		if err := cleanupJSONFile(filepath.Join(repoRoot, f), moduleName); err != nil {
+			return fmt.Errorf("cleaning up %s: %w", f, err)
+		}
+	}
+	return nil
+}
+
+// cleanupJSONFile removes all module entries from a given JSON file.
+func cleanupJSONFile(path, moduleName string) error {
```
jskeet, line 51: Is this definitely easier than parsing it as some sort of JSON DOM? Looks like it works, and shouldn't need to be expanded in the future, so I'm okay keeping it like this - but treating JSON as plain text always makes me nervous :)

internal/stategen/cleanup.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	return os.WriteFile(path, []byte(output), 0644)
+}
+
+// cleanupPostProcessorConfig removes the module and service-config entries for
+// a given module from the internal/postprocessor/config.yaml file.
+// NOTE: This function does not remove modules from manual-clients.
+func cleanupPostProcessorConfig(repoRoot, moduleName string) error {
+	configPath := filepath.Join(repoRoot, "internal", "postprocessor", "config.yaml")
```
jskeet, line 110: Again, would this be easier with YAML parsing? I guess that wouldn't preserve comments though...

internal/stategen/cleanup.go
```diff
@@ -0,0 +1,225 @@
...
+			inServiceConfigs = false
+		}
+
+		if !inServiceConfigs {
+			newLines = append(newLines, line)
+			continue
+		}
+
+		// In service-configs section
+		if strings.HasPrefix(line, "  - input-directory:") {
```
jskeet, line 162: Are we certain that all entries start with `input-directory`? (The order shouldn't matter in general YAML)

internal/stategen/cleanup.go
```diff
@@ -0,0 +1,225 @@
...
+		return err
+	}
+
+	lines := strings.Split(string(fileBytes), "\n")
+	var newLines []string
+	for i := 0; i < len(lines); i++ {
+		line := lines[i]
+		if strings.Contains(line, "source:") {
+			// It's a source line, check it for the module name.
+			if strings.Contains(line, moduleName) {
```
jskeet: I think we should add a slash here - so that the "ai" module doesn't end up stripping "aiplatform" entries, and ditto "pubsub" and "pubsublite".

internal/stategen/README.md
```diff
@@ -16,3 +16,15 @@ $ go run ./internal/stategen . spanner apphub
 It is expected that the state file (`.librarian/state.yaml`) already
 exists. Any libraries that already exist within the state file are
 ignored, even if they're listed in the command line.
+
+**NOTE:** This is a one-time migration tool to assist in moving modules from the legacy OwlBot/release-please workflow to the new Librarian workflow.
+
+In addition to adding new modules to the `.librarian/state.yaml` file, this tool will also **remove the legacy configuration** for each migrated module from the following files:
```
jskeet: Add in the part about go imports etc?

jskeet (commented):
Looking good - most comments are notes rather than requests, but I do think we need to fix the leading-substring-overlap issue, and add the "prepare" aspects to the README.

internal/stategen/cleanup.go
quartzmo: The first commit contains my attempt at JSON and YAML parsing and manipulation. It would have worked in the end, but was fiddly and taking too much development time. The schemas of these files are all simple, and the edits are all full-line deletions. Just as with BUILD.bazel, treating them as raw text is a less deft but more pragmatic way to perform the surgery.

internal/stategen/cleanup.go
quartzmo: The first commit contains my attempt at JSON and YAML parsing and manipulation. It would have worked in the end, but was fiddly and taking too much development time. The schemas of these files are all simple, and the edits are all full-line deletions. Just as with BUILD.bazel, treating them as raw text is a less deft but more pragmatic way to perform the surgery.

internal/stategen/cleanup.go
quartzmo: Yes, they do.

internal/stategen/cleanup.go
quartzmo: Good catch! I fixed this and tested it manually against `ai`. The resulting diffs LGTM.

internal/stategen/README.md
quartzmo: Done, thanks!

internal/stategen/cleanup.go
```diff
@@ -0,0 +1,226 @@
...
+		if err := cleanupJSONFile(filepath.Join(repoRoot, f), moduleName); err != nil {
+			return fmt.Errorf("cleaning up %s: %w", f, err)
+		}
+	}
+	return nil
+}
+
+// cleanupJSONFile removes all module entries from a given JSON file.
+func cleanupJSONFile(path, moduleName string) error {
+	fileBytes, err := os.ReadFile(path)
```
codyoss, line 52: Any reason to not just parse the JSON, update it, and marshal it back out instead of doing all of this manual string parsing?

codyoss: I think I have this general concern for all of these cleanup files.

internal/stategen/prepare.go
```diff
@@ -0,0 +1,119 @@
...
+// do
+//
+//	if grep -q '//go:build ignore' $file
+//	then
+//	  echo "Deleting $file"
+//	  rm $file
+//	fi
+//
+// done
+func removeIgnoredGrpcFiles(moduleRoot string) error {
```
codyoss, line 99: Where/what are these? When searching google-cloud-go I don't see any hits for this

jskeet: Example of the kind of file it's meant to delete: https://github.com/googleapis/google-cloud-go/blob/main/dataproc/apiv1/dataprocpb/operations_grpc.pb.go

But the "//go:build ignore" should possibly just be "// +build ignore".

jskeet: (The "//go:build ignore" is what's in my migration doc, so I don't know whether these have changed recently...)

internal/stategen/cleanup.go
quartzmo: This solution is working. I can add unit tests if desired, but since this is a throwaway migration tool, I don't think it's justified.

internal/stategen/prepare.go
quartzmo: Good catch Cody! I updated the target string to `// +build ignore`. Thanks!

quartzmo:
Adding `do not merge` while I add one more step to cleanup.go: The addition of onboarded modules to `skip-module-scan-paths` in `internal/postprocessor/config.yaml`.

internal/stategen/cleanup.go
codyoss: I think a couple tests would help in that case then. Even if it is throwaway code I think we need to validate its correctness.

jskeet (dismissed):
LGTM

jskeet (approved):

codyoss (approved):
Thanks for the tests

Issue #12658, opened 2025-08-06
bigtable: support routing cookie


mutianf:
## Is your feature request related to a problem? Please describe.

Bigtable server reads `x-goog-cbt-cookie-*` from the request metadata to redirect requests. Bigtable server will also send info in the response metadata for directions for retries.

## Additional context

Implementation in C++ https://github.com/googleapis/google-cloud-cpp/issues/13447

Implementation in Java https://github.com/googleapis/java-bigtable/pull/1888

PR #12964, opened 2025-09-27
feat(client): Enable routing cookie and attempt headers for enhanced retries

bhshkh:
Implements client-side logic to support the Cloud Bigtable routing cookie protocol (see go/cbt-support-routing-cookie). This feature enhances server-initiated retries and improves resilience.

Specific modifications:

*   **Feature Flag:** The client now signals support for this feature by setting `RoutingCookie: true` in the `bigtable-features` header.
*   **Cookie Propagation:** Any headers prefixed with `x-goog-cbt-cookie-` received in response metadata are collected and sent back on subsequent requests for the same operation.
*   **Attempt Header:**
    *   The `x-goog-cbt-attempt` header is added to each outgoing RPC attempt.
    *   This integer value starts at 0 and is incremented before each call.
    *   Crucially, the counter is reset to 0 if the client receives *any* metadata (headers or trailers) from the server, indicating the request reached the backend. This helps differentiate between network issues and server-side retries.
*   **Internal Tracking:** The `opTracer` in `metrics.go` now stores the current cookies and the `routingAttempt` number.

These changes enable more sophisticated load balancing and retry strategies, by giving the service and routing layers visibility into the client's retry state and a mechanism to influence client behavior.

Fixes: #12658

### DIFF
--- a/bigtable/bigtable.go
+++ b/bigtable/bigtable.go
@@ -55,6 +55,10 @@ const (
 	queryExpiredViolationType        = "PREPARED_QUERY_EXPIRED"
 	preparedQueryExpireEarlyDuration = time.Second
 	methodNameReadRows               = "ReadRows"
+
+	// For routing cookie
+	cookiePrefix  = "x-goog-cbt-cookie-"
+	attemptHeader = "x-goog-cbt-attempt"
 )
 
 var errNegativeRowLimit = errors.New("bigtable: row limit cannot be negative")
@@ -398,6 +402,7 @@ type Table struct {
 // and enabled on the client
 func (c *Client) newFeatureFlags() metadata.MD {
 	ff := btpb.FeatureFlags{
+		RoutingCookie:            true,
 		ReverseScans:             true,
 		LastScannedRowResponses:  true,
 		ClientSideMetricsEnabled: c.metricsTracerFactory.enabled,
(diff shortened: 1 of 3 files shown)


bigtable/bigtable.go
```diff
@@ -55,6 +55,10 @@ const (
 	queryExpiredViolationType        = "PREPARED_QUERY_EXPIRED"
 	preparedQueryExpireEarlyDuration = time.Second
 	methodNameReadRows               = "ReadRows"
+
+	// For routing cookie
+	cookiePrefix  = "x-goog-cbt-cookie-"
+	attemptHeader = "x-goog-cbt-attempt"
```
mutianf: Sorry that the design doc was unclear about this, the attempt part is not implemented on the GFE and we may not need it, so we can remove this header.

however, we do keep stats on the number of attempt of a request (regardless of routing cookie) and the header for that is `bigtable-attempt`. But this is a separate feature and we can do that in a different PR in the future.

bigtable/bigtable.go
```diff
@@ -2281,15 +2286,43 @@ func gaxInvokeWithRecorder(ctx context.Context, mt *builtinMetricsTracer, method
...
+		// Inject cookie and attempt information
+		md := metadata.New(nil)
+		for k, v := range op.cookies {
+			md.Append(k, v)
+		}
+		md.Set(attemptHeader, strconv.Itoa(op.routingAttempt))
+		op.routingAttempt++
+
+		existingMD, _ := metadata.FromOutgoingContext(ctx)
+		finalMD := metadata.Join(existingMD, md)
```
mutianf, line 2296: do we need to check if md is nil when we do the Join and what happens if md is nil?

bigtable/bigtable.go
```diff
@@ -2281,15 +2286,43 @@ func gaxInvokeWithRecorder(ctx context.Context, mt *builtinMetricsTracer, method
 	mt.setMethod(method)
 
 	callWrapper := func(ctx context.Context, callSettings gax.CallSettings) error {
+		op := &mt.currOp
```
mutianf, line 2288: why do we need to get the op from metrics tracer? I assume this is just how we access headers and trailers and it doesn't have anything to do with metrics?

bigtable/metrics.go
```diff
@@ -524,14 +528,21 @@ func (a *attemptTracer) setServerLatencyErr(err error) {
 }
 
 func (tf *builtinMetricsTracerFactory) createBuiltinMetricsTracer(ctx context.Context, tableName string, isStreaming bool) builtinMetricsTracer {
-	if !tf.enabled {
-		return builtinMetricsTracer{builtInEnabled: false}
-	}
 	// Operation has started but not the attempt.
 	// So, create only operation tracer and not attempt tracer
-	currOpTracer := opTracer{}
+	currOpTracer := opTracer{
```
mutianf, line 532: I don't think metrics is the right place for this feature (like I mentioned in the other comment) because this is retry related and doesn't really have anything to do with metrics?

bigtable/bigtable.go
bhshkh: 1. Do you need to check if `md` is `nil`?
      - No, you do not need to check if md is nil before passing it to metadata.Join.

   2. What happens if `md` is `nil`?
        - The metadata.Join function is designed to handle nil arguments gracefully. If md is nil, it is simply ignored during the merge process. The function will join any other non-nil metadata objects that are passed to it. If all arguments to Join are nil, it will return a new, empty metadata.MD object.


Sample code: https://go.dev/play/p/s2T74BODEMt


```go
package main

import (
	"fmt"

	"google.golang.org/grpc/metadata"
)

func main() {
	fmt.Println("Hello, 世界")
	existingMD := metadata.New(map[string]string{"foo": "bar"})
	finalMD := metadata.Join(existingMD, nil)
	fmt.Printf("finalMD len: %v\n", finalMD.Len())
	fmt.Println("Bye, 世界")
}

```

On run:

```go
Hello, 世界
finalMD len: 1
Bye, 世界

Program exited.
```

bigtable/bigtable.go
bhshkh: Yes, it does not have anything to do with metrics.

bigtable/metrics.go
bhshkh: I did not want to add another tracer. 

I can create a follow up PR to this one. Renaming metricsTracer to Tracer and file to tracer.go

PR #12971, opened 2025-09-29
refactor(firestore): refactor tests

bhshkh:


hongalex (approved):
It looks like these are just wrapping tests inside t.Run and adding an additional empty check, which looks good to me.

PR #12976, opened 2025-09-29
feat(internal/librariangen): support for "whole repo" libraries

jskeet:
If a library has "." configured as a source path, expect to find CHANGES.md in the root directory, and version.go under the root internal directory. (In every other case, we use the library ID as a "relative root directory" for the module.)

### DIFF
--- a/internal/librariangen/release/release.go
+++ b/internal/librariangen/release/release.go
@@ -23,6 +23,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strings"
 	"time"
@@ -66,7 +67,12 @@ func Init(ctx context.Context, cfg *Config) error {
 		}
 		moduleConfig := repoConfig.GetModuleConfig(lib.ID)
 
-		moduleDir := filepath.Join(cfg.OutputDir, lib.ID)
+		var moduleDir string
+		if wholeRepoLibrary(lib) {
+			moduleDir = cfg.OutputDir
+		} else {
+			moduleDir = filepath.Join(cfg.OutputDir, lib.ID)
+		}
 		slog.Info("librariangen: processing library for release", "id", lib.ID, "version", lib.Version)
 		if err := updateChangelog(cfg, lib, now().UTC()); err != nil {
 			return writeErrorResponse(cfg.LibrarianDir, fmt.Errorf("librariangen: failed to update changelog for %s: %w", lib.ID, err))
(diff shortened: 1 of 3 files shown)


internal/librariangen/release/release_test.go
```diff
@@ -148,15 +152,38 @@ func TestInit(t *testing.T) {
...
 			changelogAlreadyUpToDate: true,
 			wantVersion:              "1.16.0",
 			wantSnippetVersion:       `"version": "1.16.0"`,
 		},
+		{
+			name: "whole repo library",
+			requestJSON: `{
+				"libraries": [{
+					"id": "wholerepo", "version": "1.16.0", "release_triggered": true,
+					"source_roots": ["."],
```
quartzmo, line 170: Can you expand this example (or add another) that involves multiple source_roots entries, as I believe google-cloud-go will have? That would also be a good way to lock down the `slices.Contains` logic in `wholeRepoLibrary`.

internal/librariangen/request/request.go
```diff
@@ -36,6 +36,11 @@ type Library struct {
 	RemoveRegex []string `json:"remove_regex,omitempty"`
 	// Changes are the changes being released (in a release request)
 	Changes []*Change `json:"changes,omitempty"`
+	// Specifying a tag format allows librarian to honor this format when creating
```
quartzmo, line 39: In Go doc style, the doc should always start with the name of the code object: `TagFormat specifies a format for Librarian to honor when...`

internal/librariangen/release/release.go
```diff
@@ -219,6 +237,10 @@ func writeErrorResponse(dir string, err error) error {
 	return err
 }
 
+func wholeRepoLibrary(lib *request.Library) bool {
```
quartzmo, line 258: Maybe add a comment here providing some background on the possible configurations of `SourcePaths` for a real-world whole repo library?

quartzmo (changes requested):

internal/librariangen/release/release_test.go
jskeet: I'm not expecting to have any repos that have "." *and* something else. For the google-cloud-go "main" module, I'm expecting to have source roots of "civil", "rpcreplay", "httpreplay" and some "internal" entries. I haven't figured out how we still write to the root CHANGES.md file in that situation though. (I don't think we'd want "." as so many other libraries would need to be excluded from the releases...)

internal/librariangen/request/request.go
jskeet: Yup - this was copied directly from state.go in the librarian repo though... I figured consistency with that was more important than fine-tuning.

internal/librariangen/release/release.go
jskeet: Yup, happy to do that.

quartzmo (approved):

PR #12980, opened 2025-09-29
refactor(bigquery/v2): job polling logic

alvarowolfx:
Split job polling logic, so later we can have a generic job vs query package to help execute and wait for jobs.

Towards https://github.com/googleapis/google-cloud-go/issues/12877

### DIFF
--- a/bigquery/v2/internal/job/handler.go
+++ b/bigquery/v2/internal/job/handler.go
@@ -1,0 +1,254 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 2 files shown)


bigquery/v2/internal/job/handler.go
```diff
@@ -0,0 +1,254 @@
...
+	// context for background pooling
+	ctx      context.Context
+	mu       sync.RWMutex
+	complete bool
+	ready    chan struct{}
+	err      error
+}
+
+// CreateFunc is a function that start/insert a new job in BigQuery. Usually jobs.insert or jobs.query.
+type CreateFunc = func(ctx context.Context, opts []gax.CallOption) (protoreflect.Message, error)
```
shollyman, line 46: CreateFunc and WaitFunc are the same type signature.  Should you just use a more general type and name (e.g. RPCFunc or something)?

bigquery/v2/query/query.go
```diff
@@ -122,134 +109,81 @@ func (q *Query) Wait(ctx context.Context) error {
...
-
-	if err != nil {
-		q.markDone(err)
-		return
-	}
-
-	q.consumeQueryResponse(&bigquerypb.GetQueryResultsResponse{
-		JobReference: res.GetJobReference(),
-	})
+func (q *Query) insertQueryFunc(job *bigquerypb.Job, projectID string, opts []gax.CallOption) internaljob.CreateFunc {
```
shollyman, line 122: It feels odd to have this method hanging off the outer message and the inner field referencing it as part of its constructor.  I wonder if we're missing a chance for a better signature for the new handler function.

Issue #11885, opened 2025-03-21
firestore: add support for omitzero tag


nathanperkins:
## Client

Firestore

## Environment

golang 1.24

## Code and Dependencies

```
type Timing struct {
  Created time.Time `firestore:"created,omitzero"`
}
```

## Expected behavior

Similar to the recent `omitzero` tag added for `marshal/json` in golang 1.24, it should support struct fields by omitting the field when the struct is the zero value.

## Actual behavior

Error:

```
  "error": "rpc error: code = Internal desc = add job \"3f31bd47-ee3a-4a2e-bb53-d8d11cae8546\" to database: create doc \"...\" in Firestore: firestore: unknown tag option: \"omitzero\"",
```

## Additional context

More info on 1.24 json tags: https://www.bytesizego.com/blog/go-124-omitzero

bhshkh:
Created https://github.com/googleapis/google-cloud-go/pull/12992

PR #12992, opened 2025-09-30
feat(firestore): Add omitzero struct tag option

bhshkh:
Adds the omitzero struct tag option for marshaling Firestore documents, mirroring the behavior of omitzero in encoding/json https://cs.opensource.google/go/go/+/refs/tags/go1.25.0:src/encoding/json/v2_encode.go.

When a struct field is tagged with omitzero (e.g., Score int `"firestore:",omitzero"`), it will be omitted from the resulting Firestore document if its value is zero.

A value is considered zero if:

- Its type T or pointer type *T implements an IsZero() bool method, and that method returns true.
- Otherwise, if the value is its type's zero value (0, false, "", nil, etc.).

This is consistent with encoding/json and allows users to skip writing fields that contain zero values, which is useful for PATCH semantics or when zero values should be treated as "unset".

Example:

```
type Update struct {
    Name string `firestore:",omitzero"`
    Age  int    `firestore:",omitzero"`
}

// doc := Update{Name: "Sam", Age: 0}
// When written to Firestore, only the 'Name' field will be present
// in the document because 'Age' is 0 and tagged with 'omitzero'.
```

Fixes: #11885


firestore/to_value.go
```diff
@@ -314,3 +325,29 @@ func isEmptyValue(v reflect.Value) bool {
...
+	}
+	if v.Type().Implements(isZeroerType) {
+		if v.Kind() == reflect.Interface || v.Kind() == reflect.Pointer {
+			if v.IsNil() {
+				return true
+			}
+			if v.Kind() == reflect.Interface && v.Elem().Kind() == reflect.Pointer && v.Elem().IsNil() {
+				return true
+			}
+		}
```
alvarowolfx, line 344: Took a bit to understand the code here, because I felt like the first `if` statement + `v.IsNil` was similar to the second statement checking for the value being an interface + pointer, but the key different is the `v.Elem()` call. 

I would rewrite like this, but it's just a nit:
```go
isInterface := v.Kind() == reflect.Interface
isPointer := v.Kind() == reflect.Pointer
if isInterface || isPointer {
	if v.IsNil() {
		return true
	}
	if isInterface && isPointer && v.Elem().IsNil() {
		return true
	}
}
```

bhshkh: I have rewritten the code.
Added integration test too.

PR #13008, opened 2025-10-01
chore: migrate pubsub and pubsub/v2 to Librarian

jskeet:
Note that the tag format for pubsub/v2 is hard-coded to use "pubsub/" as the prefix instead of "pubsub/v2".

Towards https://github.com/googleapis/librarian/issues/2419


jskeet:
https://github.com/jskeet/google-cloud-go/pull/23 is now a working generation PR, but requires https://github.com/googleapis/google-cloud-go/pull/13013 to be merged (and built) first.

Interestingly, the PR found a bug in the current snippet metadata!

jskeet:
Sample release PR: https://github.com/jskeet/google-cloud-go/pull/24

jskeet:
@hongalex I've been waiting until @shollyman gives the go-ahead before merging this. Still checking with him. Once he's happy, I'll make sure that we don't have any other pending unreleased changes from OwlBot before merging, as well. Once it *is* merged, I suggest that you pair with either @quartzmo or myself next time you want to do a release... is that okay?

jskeet:
@quartzmo: I've merged main, and bumped the version number in state.yaml. Regenerating didn't produce any changes other than the snippet metadata version number (which is out of date at the moment). I believe we can now merge this (once it's green, of course).

.github/.OwlBot.yaml
```diff
@@ -163,10 +161,6 @@ deep-preserve-regex:
   - /kms/apiv1/iam_example_test.go
   - /logging/apiv2/WriteLogEntries_smoke_test.go
   # TODO(#7336): Remove these networkconnectivity lines once breaking change is handled.
-  # Preserve pubsub apiv1 until deprecation date in a year even though it's not generated.
```
quartzmo, line 166: This comment ("Preserve pubsub apiv1 until deprecation date in a year even though it's not generated.") is being removed from the OwlBot config without being copied to `state.yaml`, which nevertheless does contain an entry for `pubsub/apiv1` including paths for `pubsub/apiv1` in `remove_regex`. Is this statement still true for the Librarian support for pubsub (apiv1) and is the new configuration consistent with this statement ("not being generated")?

jskeet: Yes - because the state file doesn't have *any* API paths to generate. We could actually remove the remove_regex entirely if we wanted to - basically generate should never be called on `pubsub` anyway.

Would you feel less concerned if I removed those remove_regex entries?

hongalex: As long as we preserve the existing `pubsub/apiv1` GAPIC, I'm not concerned about this comment sticking around.

jskeet: The whole file is going to go away pretty soon, hopefully :) But yes, we're definitely preserving pubsub/apiv1

PR #13009, opened 2025-10-01
chore: migrate compute/metadata to Librarian

jskeet:
Towards https://github.com/googleapis/librarian/issues/2417

Sample release PR: https://github.com/jskeet/google-cloud-go/pull/19


jskeet:
(vet and apidiff fail for that release PR, but I don't *think* it's due to the releasing itself...)

PR #13013, opened 2025-10-01
fix(internal/librariangen): handle multi-versioned modules

jskeet:
If a library ID includes the version (e.g. "pubsub/v2") then we assume that's the module root, and we don't move generated code into just "pubsub".

Related to https://github.com/googleapis/librarian/issues/2305

(Once we've figured out how to handle all these situations, we definitely need to document "how to create a new major version" etc.)

### DIFF
--- a/internal/librariangen/generate/generator.go
+++ b/internal/librariangen/generate/generator.go
@@ -118,7 +118,7 @@ func Generate(ctx context.Context, cfg *Config) error {
 		return fmt.Errorf("librariangen: failed to flatten output: %w", err)
 	}
 
-	if err := applyModuleVersion(cfg.OutputDir, moduleConfig.GetModulePath()); err != nil {
+	if err := applyModuleVersion(cfg.OutputDir, generateReq.ID, moduleConfig.GetModulePath()); err != nil {
 		return fmt.Errorf("librariangen: failed to apply module version to output directories: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/librariangen/generate/generator.go
```diff
@@ -217,8 +217,10 @@ func flattenOutput(outputDir string) error {
 // appropriately for versioned modules. For a module path of the form cloud.google.com/go/{id}/{version},
 // we expect to find /output/{id}/{version} and /output/internal/generated/snippets/{id}/{version}.
 // The content of these directories are be moved into /output/{id} and
```
quartzmo: There is a pre-existing typo here ("are be moved").

internal/librariangen/generate/generator.go
```diff
@@ -217,8 +217,10 @@ func flattenOutput(outputDir string) error {
...
-// /output/internal/generated/snippets/{id}
-func applyModuleVersion(outputDir, modulePath string) error {
+// /output/internal/generated/snippets/{id}. If the library ID already includes the version,
+// then we assume it's because we're handling multiple versions, and the code is already at
+// {module}/{version}.
```
quartzmo: Can you take a shot at rewriting the **two** sentences above (including the pre-existing text with the typo) for greater clarity? The behavior here is pretty clear from the unit test below, but these sentences are a bit hard to follow.

internal/librariangen/release/release.go
```diff
@@ -244,12 +247,12 @@ func writeErrorResponse(dir string, err error) error {
...
-// The "main" module of google-cloud-go might be expected to use "."
-// as a source path, but that would end up getting changes from *all*
-// nested modules (which will expand over time). That module is likely
-// to have a snowflake-like configuration.
-func wholeRepoLibrary(lib *request.Library) bool {
-	return slices.Contains(lib.SourcePaths, ".")
+// The use of a special ID of "root-module" for the module of
+// google-cloud-go containing "civil", "rpcreplay" etc is slightly
+// hacky, but avoids creating special-purpose configuration which
+// is realistically only ever going to be used by a single module.
```
quartzmo, line 253: I think maybe we need one more sentence explaining why this hack is the best choice vs the "special-purpose configuration". As it stands, I'm left wondering what that would look like and if it might be the better choice.

internal/librariangen/generate/generator.go
jskeet: Fixed in a new commit.

internal/librariangen/generate/generator.go
jskeet: I've had a go in a new commit.

internal/librariangen/release/release.go
jskeet: I've erred on the side of verbosity in the comment...

quartzmo (approved):

PR #13016, opened 2025-10-01
chore: migrate apigeeregistry apikeys appengine apphub eventarc to Librarian

quartzmo:
refs: googleapis/librarian#888


internal/postprocessor/config.yaml
```diff
@@ -1292,6 +1269,11 @@ skip-module-scan-paths:
   # ignore the root module
   - .
   # Librarian released modules
+  - eventarc
```
shollyman, line 1272: Do we want to keep this list alpha sorted?  It may make it easier to review the list of migrated APIs.

shollyman (approved):

internal/postprocessor/config.yaml
quartzmo: Yes, good call, we should. But even more importantly, we need to sort the list in `.librarian/state.yaml`, which is also not sorted. That will be the durable configuration while this one will eventually be deleted and just needs to serve its temporary purpose.

PR #13021, opened 2025-10-02
fix(pubsub/v2): respect ShutdownBehavior when handling timeout

kurochan:
Previously, `ShutdownOptions.Behavior` was never respected — when a positive `Timeout` was set, messages were always nacked. This PR fixes the shutdown logic so that messages are only nacked if `Behavior` is explicitly set to `ShutdownBehaviorNackImmediately`. This makes the `Behavior` field effective as intended.

related to https://github.com/googleapis/google-cloud-go/pull/12829


hongalex:
Thanks for the PR. This was originally part of the implementation, but I had dropped it at some point when debugging another issue. Looks good to me.

hongalex (approved):

PR #13025, opened 2025-10-02
chore: migrate root module to Librarian

jskeet:
This requires the changes in https://github.com/googleapis/google-cloud-go/pull/13013


jskeet:
Converted back to a draft, as I hadn't been aware of the shared code in `internal`.

## Short term

- If there's a quick way of moving the "currently in the root of internal" code into a subdirectory without breaking users, we could have specific source roots for "just the code we care about"
- Otherwise, we can add `internal` as a source root (in which case we can remove the global file entry for `internal/version.go`) and have a long list of `release_exclude_paths`.
  - The downside of this is that whenever anyone adds an internal subdirectory which *isn't* meant to be used, they'll have to add it to the state file (which is *intended* to be machine-edited only...) We may want to have an allow-list somewhere and a GitHub action to validate "only these directories should be omitted from the `release_exclude_paths` for the root module".

## Long term

In the longer term:

- We should consider how much of this code really needs to be internal. Could it be moved to GAX? Is there a better way of sharing it even if it *does* need to stay internal?
- We should consider separating civil, httpreplay and rpcreplay - those are inherently separate concerns IMO. (I really wouldn't expect the same dependency to handle date/time calculations and replaying RPCs...)

jskeet:
@quartzmo: This is now up-to-date with main.

.release-please-manifest.json
```diff
@@ -1,3 +1,2 @@
 {
-  ".": "0.123.0"
-}
+}
```
quartzmo, line 2: Missing newline at end of file.

jskeet: As this file *isn't* changed in other migrations, I'm going to merge as-is then create a separate PR just to add the newline. That way it shouldn't end up with merge conflicts.

PR #13032, opened 2025-10-02
chore(internal/stategen): sort libraries in state.yaml before saving

quartzmo:
* Refactor main.go functions to state.go.
* Add state_test.go.


internal/stategen/state.go
```diff
@@ -95,9 +103,176 @@ type Change struct {
...
+			api := &API{
+				Path:          serviceConfig.InputDirectory,
+				ServiceConfig: serviceConfig.ServiceConfig,
+			}
+			library.APIs = append(library.APIs, api)
+		}
+	}
+}
+
+// addApiPaths walk the module source directory to find the files to remove.
```
jskeet, line 212: This is probably my fault - the doc comment doesn't match the function name. Will create a tiny PR to fix.

PR #13039, opened 2025-10-03
test(internal/postprocessor): remove bad expectations from config test

shollyman:
This test made assertions about specific API configs present in the config file, and those assertions are now invalid now that librarian migrations have started removing configs from the postprocessor.

This test removes those bad expectations

fixes: https://github.com/googleapis/google-cloud-go/issues/13034


quartzmo (approved):
Nothing beats testing against production configs!

PR #13045, opened 2025-10-06
chore(internal/stategen): fix cleanup.go for special-case cloudtasks module

quartzmo:
Discover correct `../tasks/..` Go importpath from existing configuration
in a generalized solution for all modules.


hongalex (approved):

internal/stategen/cleanup.go
```diff
@@ -238,21 +238,48 @@ func cleanupOwlBotYaml(repoRoot, moduleName string) error {
...
+		return fmt.Errorf("loading postprocessor config: %w", err)
+	}
+	cleanupName := moduleName // Default to the module name.
+	importPrefix := "cloud.google.com/go/" + moduleName + "/"
+	for _, sc := range ppc.ServiceConfigs {
+		if strings.HasPrefix(sc.ImportPath, importPrefix) {
+			// The InputDirectory is the path to the API protos, relative to the
+			// googleapis root. It has a structure like:
+			// "google/cloud/service/v1" or "google/identity/accesscontextmanager/v1".
+			// The third component of this path is the service name as it appears
```
jskeet, line 255: I wouldn't take the third component here - that doesn't work for things like "google/spanner/v1".

But I think the full path *is* always present in the source line, e.g. `source: /google/identity/accesscontextmanager/v1/cloud.google.com/go` (before that was recently deleted). So I'd just use the whole of `sc.InputDirectory` rather than splitting it.

PR #13050, opened 2025-10-06
feat(spanner): Add transaction type to span attribute

yordis:


yordis:
Hey @rahul2393 we are running into a situation where we would like to know the transaction type; I wonder if this is a good starting point to get the information into the span

PR #13106, opened 2025-10-13
fix(internal/stategen): be more conservative about removing OwlBot lines

jskeet:
Previously, migrating module "iam" would remove all lines containing "/iam/" - even if they weren't in the iam module.


internal/stategen/cleanup.go
```diff
@@ -246,7 +246,8 @@ func cleanupOwlBotYaml(repoRoot, moduleName string) error {
 		return fmt.Errorf("loading postprocessor config: %w", err)
 	}
 	importPrefix := "cloud.google.com/go/" + moduleName + "/"
-	modulePathFragment := "/" + moduleName + "/"
+	modulePathFragment := " /" + moduleName + "/"
```
quartzmo, line 268: Can you provide an inline comment here with an example of what this pattern successfully matches and also what it avoids matching.

Issue #13113, opened 2025-10-13
pubsub: Receive may not terminate after v2.1.0 when the context is canceled


landcelita:
## Client

PubSub

## Environment

go 1.24.4

## Code and Dependencies

```go
package main

import (
	"context"
	"fmt"
	"math/rand"
	"time"

	"cloud.google.com/go/pubsub/v2"
)

func main() {
	ctx, cancel := context.WithCancel(context.Background())

	projectID := "example-project"
	client, err := pubsub.NewClient(ctx, projectID)
	if err != nil {
		panic(err)
	}
	defer client.Close()
	cctx, cancel := context.WithTimeout(ctx, 5*time.Second)
	defer cancel()

	done := make(chan struct{})
	go func() {
		defer close(done)
		receiveMessage(cctx, client)
	}()

	go func() {
		<-cctx.Done()
		fmt.Printf("cctx timeout!\n")
	}()

	select {
	case <-done:
		fmt.Printf("receiveMessage() finished; main exit\n")
	}
}

func receiveMessage(cctx context.Context, client *pubsub.Client) {
	subID := "timeout-test-sub"
	sub := client.Subscriber(subID)
	sub.ReceiveSettings.MaxOutstandingMessages = 3

	err := sub.Receive(cctx, func(ctx context.Context, msg *pubsub.Message) {
		fmt.Printf("sub.Receive(): %s\n", string(msg.Data))

		// mock the work
		time.Sleep(time.Duration(rand.Int63n(1000)) * time.Millisecond)

		msg.Ack()
		return
	})
	if err != nil {
		panic(err)
	}
}
```

<details>
  <summary>go.mod</summary>

```text
module pubsub-timeout-test

go 1.24.4

require cloud.google.com/go/pubsub/v2 v2.2.0

require (
	cloud.google.com/go v0.123.0 // indirect
	cloud.google.com/go/auth v0.17.0 // indirect
	cloud.google.com/go/auth/oauth2adapt v0.2.8 // indirect
	cloud.google.com/go/compute/metadata v0.9.0 // indirect
	cloud.google.com/go/iam v1.5.3 // indirect
	github.com/felixge/httpsnoop v1.0.4 // indirect
	github.com/go-logr/logr v1.4.3 // indirect
	github.com/go-logr/stdr v1.2.2 // indirect
	github.com/google/s2a-go v0.1.9 // indirect
	github.com/google/uuid v1.6.0 // indirect
[...]

PR #13114, opened 2025-10-13
fix(pubsub/v2): avoid Receive hang on context cancellation

landcelita:
Related to https://github.com/googleapis/google-cloud-go/issues/13113

## Cause of the Issue

In v2.0, when the context was canceled, [all messages were Nacked](https://github.com/googleapis/google-cloud-go/blob/494c545e468b5e1b2aa371811c07b4208ab4a87f/pubsub/v2/subscriber.go#L305-L312).

However, starting from v2.1, when the context is canceled, even if a message has already been received via [iter.receive()](https://github.com/landcelita/google-cloud-go/blob/0135d9305581444e1ddcdd8f4fe63e4c588b575f/pubsub/v2/subscriber.go#L323)
, the implementation [returns without calling Nack()](https://github.com/landcelita/google-cloud-go/blob/0135d9305581444e1ddcdd8f4fe63e4c588b575f/pubsub/v2/subscriber.go#L340).

As a result, since v2.1, [iterator.done()](https://github.com/landcelita/google-cloud-go/blob/14c3887819c7bfdf3de661ec807fa82b6bb3183e/pubsub/v2/iterator.go#L219) may not be executed for some keys, which causes the condition [len(it.keepAliveDeadlines) == 0](https://github.com/landcelita/google-cloud-go/blob/14c3887819c7bfdf3de661ec807fa82b6bb3183e/pubsub/v2/iterator.go#L201) to never be satisfied.

Consequently, the program remains blocked at the section where [it.wg.Wait()](https://github.com/landcelita/google-cloud-go/blob/14c3887819c7bfdf3de661ec807fa82b6bb3183e/pubsub/v2/iterator.go#L183) is called, which in turn causes [Receive’s group.Wait()](https://github.com/landcelita/google-cloud-go/blob/76ddb3498f2c986b156abaeaf7efeab1c9490725/pubsub/subscription.go#L1533) to remain blocked as well, resulting in a hang.

## Summary of the Fix

Modified the behavior so that no further messages are received once the context is canceled.

### DIFF
--- a/pubsub/v2/subscriber.go
+++ b/pubsub/v2/subscriber.go
@@ -320,7 +320,7 @@ func (s *Subscriber) Receive(ctx context.Context, f func(context.Context, *Messa
 				msgChan := make(chan []*Message)
 				errChan := make(chan error)
 				go func() {
-					// Make message pulling dependent on iterator for context cancellation
+					// Make message pulling dependent on iterator for context cancellation.
 					// If the context is cancelled while pulling messages, stop reading from stream early.
 					select {
 					case <-ctx.Done():


pubsub/v2/subscriber.go
```diff
@@ -320,6 +320,15 @@ func (s *Subscriber) Receive(ctx context.Context, f func(context.Context, *Messa
 				msgChan := make(chan []*Message)
 				errChan := make(chan error)
 				go func() {
+					// Make message pulling dependent on iterator for context cancellation
```
hongalex:
```suggestion
					// Make message pulling dependent on iterator for context cancellation.
```

Because of weirdness with how our automation handles 1 commit PRs, could I ask you to push a commit where you just make a trivial one character change like this? I edited the title to include v2, but this won't be reflected unless there are two commits.

PR #13128, opened 2025-10-13
fix(storage): persist resource in append retry

tritone:
In one retry case for appendable writes, the client receives both an object resource w/ write handle as well as an error from the server. After reconnect the new stream would not keep the original object resource. This change ensures that the object is not dropped.

Also explicitly guarantee the contract that either w.Attrs() will be non-nil, or w.Close() will return a non-nil error.

I need to make a change to the emulator as well so that the testbench behavior more closely mocks the service.


storage/grpc_writer.go
```diff
@@ -957,7 +962,8 @@ func (s *gRPCAppendBidiWriteBufferSender) sendOnConnectedStream(buf []byte, offs
 			}
 		}
 		if s.recvErr != io.EOF {
-			return nil, s.recvErr
+			err = s.recvErr
```
BrennaEpp, line 970: Can you add a comment to this method indicating that both object and error may be populated?

tritone: Added this and cleaned up the comment on sendBuffer to be clearer.

PR #13129, opened 2025-10-13
chore(internal/postprocessor): remove .repo-metadata-full.json processing from postprocessor main

quartzmo:
Add new manifest command to postprocessor for updating .repo-metadata-full.json

### DIFF
--- a/internal/postprocessor/main.go
+++ b/internal/postprocessor/main.go
@@ -268,6 +268,9 @@ func (p *postProcessor) InitializeNewModules(manifest map[string]ManifestEntry) 
 			}
 			// serviceImportPath here should be a valid ImportPath from a MicrogenGapicConfigs
 			apiName := manifest[serviceImportPath].Description
+			if apiName == "" {
+				return fmt.Errorf("no ManifestEntry.Description found for serviceImportPath %s. Cannot generate min required files", serviceImportPath)
+			}
 			if err := p.generateMinReqFilesNewMod(moduleName, modulePath, importPath, apiName); err != nil {
 				return err
 			}


internal/postprocessor/manifest.go
```diff
@@ -51,15 +51,150 @@ const (
...
 	if err != nil {
 		return nil, err
 	}
+	if err := json.Unmarshal(b, &entries); err != nil {
+		return nil, err
+	}
+	return entries, nil
+}
+
+// UpdateManifest updates the manifest file with info about all of the confs.
```
jskeet: I'd give more detail about the parameter here - especially given that:
- An empty string has a special meaning
- It's a plural name with a singular type (because it's comma-separated)

internal/postprocessor/manifest.go
```diff
@@ -51,15 +51,150 @@ const (
...
+	f, err := os.Create(manifestPath)
+	if err != nil {
+		return err
+	}
 	defer f.Close()
+
+	buf := &bytes.Buffer{}
+	buf.WriteString("{\n")
+
+	for i, key := range keys {
```
jskeet, line 173: Is there a reason this is writing manually rather than creating an in-memory data structure that can just be serialized by the JSON library?

internal/postprocessor/main.go
```diff
@@ -78,13 +81,19 @@ var (
 	internalVersionTmpl string
 )
 
+func init() {
+	manifestCmd = flag.NewFlagSet("manifest", flag.ExitOnError)
+	manifestCmd.StringVar(&modulesFlag, "modules", "", "Comma-separated list of module names to regenerate.")
```
jskeet: Maybe say that this is optional, and what happens if it's not specified?

internal/postprocessor/manifest.go
quartzmo: Just doing this is the simplest way.

internal/postprocessor/manifest.go
quartzmo: Done, thanks!

quartzmo:
> I think I'm missing the context of how this is intended to be used. What are the use cases for running it without a list of modules, and what are the use cases for running it with a list of modules?

For onboarding new modules while we still rely on OwlBot for doing that, and for refreshing the entries for OwlBot-managed modules that are promoted to GA.

PR #13135, opened 2025-10-14
chore: migrate 38 modules (recommendationengine - workstations) to Librarian

jskeet:
Modules migrated:

- recommendationengine
- recommender
- redis
- resourcemanager
- retail
- scheduler
- secretmanager
- securesourcemanager
- security
- securitycenter
- securitycentermanagement
- securityposture
- servicedirectory
- servicehealth
- shell
- shopping
- speech
- storagebatchoperations
- storageinsights
- storagetransfer
- streetview
- support
- talent
- telcoautomation
- texttospeech
- tpu
- trace
- translate
- video
- videointelligence
- vision
- visionai
- vmwareengine
- vpcaccess
- webrisk
- websecurityscanner
- workflows
- workstations

### DIFF
--- a/release-please-config-yoshi-submodules.json
+++ b/release-please-config-yoshi-submodules.json
@@ -77,7 +77,7 @@
         },
         "vmmigration": {
             "component": "vmmigration"
-        },
+        }
     },
     "plugins": [
         "sentence-case"


jskeet:
We should wait until generation has run on the current set of modules before merging this.

jskeet:
Will look into the config schema error.

jskeet:
Running `librarian generate` I see a difference in `visionai/apiv1/doc.go`. Will talk about that with @quartzmo when we meet.

jskeet:
We understand the difference in `visionai/apiv1/doc.go` - we believe this is because the first service encountered is used to construct the sample client; when using Bazel this depends on the order of the protos in the BUILD.bazel file, whereas now the protos are just alphabetical.

jskeet:
@quartzmo Just a reminder to review this if you have a chance so I can merge in my morning if the generation of today's modules works.

PR #13142, opened 2025-10-15
feat(storage): make crc32c default and pass checksum in all requests

krishnamd-jkp:
feat(storage): make crc32c default and pass checksum in trailing and per-chunk request

### DIFF
--- a/storage/grpc_writer.go
+++ b/storage/grpc_writer.go
@@ -171,6 +171,7 @@ func (c *grpcStorageClient) OpenWriter(params *openWriterParams, opts ...storage
 
 		flushSupported:        params.append,
 		sendCRC32C:            params.sendCRC32C,
+		disableCRC32C:         params.disableCRC32C,
 		forceOneShot:          params.chunkSize <= 0,
 		forceEmptyContentType: params.forceEmptyContentType,
 		append:                params.append,
@@ -244,6 +245,7 @@ type gRPCWriter struct {
 
 	flushSupported        bool
 	sendCRC32C            bool
+	disableCRC32C         bool
 	forceOneShot          bool
 	forceEmptyContentType bool
 	append                bool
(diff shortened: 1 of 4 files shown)


storage/grpc_writer.go
cpriti-os: Nit: small comment explaining why nil would be helpful

storage/grpc_writer_test.go
cpriti-os, line 54: It would be more robust if we can have a case when the oldChecksum is non-zero.

storage/grpc_writer.go
BrennaEpp: Can you add a comment explaining this function and its contract?

BrennaEpp: I don't think we need this function - it's kind of confusing to have the full attrs returned here.

BrennaEpp, line 809: Are we ensuring this is only set on the last call or on the first? Since we are doing that elsewhere can you add a comment at the top of this function indicating when `ObjectChecksums` can/cannot be filled.

BrennaEpp, line 244:
```suggestion
	fullObjectChecksum uint32
```

krishnamd-jkp: Will implement checksums for appendable writes in a different PR. Created a TODO bug

storage/writer.go
BrennaEpp, line 65: Please make an explicit note that this does not disable the CRC32C sent when SendCRC32C is true.

storage/grpc_writer.go
BrennaEpp, line 882: Do these need to be functions and at both levels of the writer?

BrennaEpp, line 814: This function seems to be doing a lot. Can we break this up or at least comment with what this does and its contract?

BrennaEpp, line 1072: Will this work in the case of a retry?

PR #13146, opened 2025-10-15
chore: migrate 9 modules (aiplatform - vmmigration) to Librarian

jskeet:
Modules migrated in this PR, relative to #13135:

- aiplatform
- cloudbuild
- compute
- datacatalog
- datastream
- dialogflow
- monitoring
- run
- vmmigration

Once #13135 is merged, this PR will be *rebased* to main and should be re-reviewed at that point. We'll also want to make sure any OwlBot PRs are merged, any releases performed etc - which may make it worth recreating this.

### DIFF
--- a/release-please-config-yoshi-submodules.json
+++ b/release-please-config-yoshi-submodules.json
@@ -50,7 +50,7 @@
         },
         "spanner/benchmarks": {
             "component": "spanner/benchmarks"
-        },
+        }
     },
     "plugins": [
         "sentence-case"


jskeet:
Also note that aiplatform changes *often*, so we should double-check for pending changes immediately before merging.

jskeet:
And indeed there are pending aiplatform changes. We may want to remove that from this PR, so we can do that once in a simple PR when we can find a day that AI Platform *hasn't* been updated...

jskeet:
Closing in favor of #13149

PR #13149, opened 2025-10-15
chore: migrate 8 modules (cloudbuild - vmmigration) to librarian

jskeet:
- cloudbuild
- datacatalog
- datastream
- dialogflow
- iam
- monitoring
- run
- vmmigration


jskeet:
Running `librarian generate` shows only differences in state.yaml and `visionai/apiv1/doc.go` (which is expected).

.github/.OwlBot.yaml
```diff
@@ -172,11 +130,6 @@ deep-preserve-regex:
   - /storage/internal/apiv2/metadata.go
   - /containeranalysis/apiv1beta1/grafeas/.*
   # Must preserve these even though they aren't generated anymore
```
quartzmo, line 132: In migrating the 5 modules below to Librarian, are we breaking in any way the intent in the comment above to "preserve these even though they aren't generated anymore"?

jskeet: No - they're explicitly preserved in state.yaml. And the comment itself is still valid for compute/apiv1/smoke_test.go. (But we should prune orphan comments at some point - I've already seen one.)

quartzmo (approved):

PR #13153, opened 2025-10-15
feat(bigtable): Enable ALTS hard bound token in Bigtable w/ direct access

rockspore:
The same thing has been introduced to Cloud Spanner and GCS.


bhshkh (approved):

rockspore:
@sushanb added the same env var here as well

PR #13168, opened 2025-10-16
test(storage): add nil attrs retry tests

tritone:
Add these emulator tests which had been applied to the 1.56.3 branch previously

### DIFF
--- a/storage/retry_conformance_test.go
+++ b/storage/retry_conformance_test.go
@@ -656,6 +656,10 @@ var methods = map[string][]retryFunc{
 				return fmt.Errorf("Writer.Close: %v", err)
 			}
 
+			if objW.Attrs() == nil {
+				return fmt.Errorf("Writer.Attrs: expected attrs for written object, got nil")
+			}
+
 			// Don't reuse obj, in case preconditions were set on the write request.
 			r, err := b.Object(obj.ObjectName()).NewReader(ctx)
 			defer r.Close()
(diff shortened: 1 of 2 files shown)


storage/retry_conformance_test.go
```diff
@@ -703,10 +707,19 @@ var methods = map[string][]retryFunc{
 
 			}
 
+			// TODO: Remove this exception and enable the attrs check below once we
```
danielduhh, line 710: you tracking this somewhere?

storage/retry_conformance_test.go
```diff
@@ -703,10 +707,19 @@ var methods = map[string][]retryFunc{
...
+			// figure out how to handle the redirect w/ write handle case.
+			// See b/451594633
 			if err := objW.Close(); err != nil {
-				return fmt.Errorf("Writer.Close: %w", err)
+				if !strings.Contains(err.Error(), "no object attributes returned") {
+					return fmt.Errorf("Writer.Close: %w", err)
+				}
 			}
 
+			// if objW.Attrs() == nil {
```
danielduhh, line 719: delete this? Look like it got moved up

tritone: That's a separate test case for appendable uploads (the one above has finalization)

storage/retry_conformance_test.go
tritone: The linked bug has the root cause but I will file a separate one to make the emulator change to match prod.

tritone: b/454024785

PR #13179, opened 2025-10-17
chore: migrate licensemanager to Librarian

jskeet:


jskeet:
This was another new one that I missed somehow when mopping up yesterday.

Issue #13176, opened 2025-10-17
containeranalysis: v1 API should be autogenerated


jskeet:
Currently we don't regenerate the [google/devtools/containeranalysis/v1](https://github.com/googleapis/googleapis/tree/master/google/devtools/containeranalysis/v1) API.

It was last updated at #10533, where @codyoss made some manual tweaks.

We should figure out how to automate those tweaks so that the library can be kept up-to-date with API changes.

PR #13180, opened 2025-10-17
chore: migrate containeranalysis to Librarian

jskeet:
See #13176 for outstanding work regarding the v1 path and grafeas proto.

refs: #13176


jskeet:
Notes:

- This doesn't include apiv1 - see #13176
- We can't merge this until #13174 is merged, as it requires the latest image

jskeet:
Hmm... I've now noticed the grafeas/grafeaspb directory which we're not currently generating. Working on it...

jskeet:
It's worse than I thought, because grafeas.proto has:

```text
option go_package = "google.golang.org/genproto/googleapis/devtools/containeranalysis/v1beta1/grafeas;grafeas";
```

... so it's not being generated where we expect. There's also grafeas_grpc.pb.go which *would* be generated, but we don't currently have.

However, grafeas.proto itself last changed in 2020. `grafeas.pb.go` was last changed in google-cloud-go two years ago (#7482) even though in googleapis-gen just a couple of days ago - it's clearly not being copied by OwlBot.

Will add this detail to #13176 and proceed with "we're migrating the status quo".

jskeet:
(Removed "do not merge" label as #13174 has been merged)

Issue #13188, opened 2025-10-20
genbot build broken by too many commits in generated PR description


quartzmo:
The genbot build for go-genproto failed with `POST https://api.github.com/repos/googleapis/go-genproto/pulls: 422 Validation Failed [{Resource:Issue Field:body Code:custom Message:body is too long (maximum is 65536 characters)}]exit status 1`

https://fusion2.corp.google.com/invocations?q=cloud-devrel%2Fclient-libraries%2Fgo%2Fgapicgen%2Fgenbot

## What happened

1. genbot doesn't track every file in the googleapis repository. It only cares about a specific, hardcoded list of proto files that are used to generate the go-genproto library.
2. For the last 5 months (since a commit on May 9th), none of the hundreds of commits to googleapis happened to change any of the proto files on that specific watchlist. This is why the bot has been succeeding every hour by correctly reporting "there are not files to regenerate." The 10/16 8am build was the last of these "quiet" runs.
3. Between the 8am and 9am builds, a new commit must have landed in googleapis that, for the first time in months, modified one of the watched proto files.
4. This single change triggered the bot to finally perform a regeneration. When it did, its logic dictated that it should create a pull request summarizing all the changes in googleapis since the last successful regeneration in May. It gathered 649 commits to create the description.
5. The combined text of those 649 commit messages was too large for GitHub's API, causing the "body is too long" error and failing the build.

PR #13189, opened 2025-10-20
feat(internal/gapicgen): fix genbot PR descriptions

quartzmo:
When a `go-genproto` regeneration is triggered after a long period of inactivity, genbot would attempt to include all intervening commits in the pull request description. This could result a PR description being rejected by GitHub.

genbot now identifies only the commits that are relevant to the go-genproto regeneration by checking which commits affected the watched proto files.

The new PR description now includes:
- A list of only the relevant commits.
- A list of the specific packages affected by each commit.

fixes: #13188

### DIFF
--- a/internal/gapicgen/git/git.go
+++ b/internal/gapicgen/git/git.go
@@ -191,7 +191,7 @@ func DeepClone(repo, dir string) error {
 	return err
 }
 
-// FilesChanged returns a list of files changed in a commit for the provdied
+// FilesChanged returns a list of files changed in a commit for the provided
 // hash in the given gitDir.
 func FilesChanged(gitDir, hash string) ([]string, error) {
 	c := execv.Command("git", "show", "--pretty=format:", "--name-only", hash)


internal/gapicgen/generator/generator.go
```diff
@@ -69,7 +70,38 @@ func gatherChanges(googleapisDir, genprotoDir string) ([]*git.ChangeInfo, error)
...
+				}
+			}
+			if onWatchlist {
+				if _, ok := affectedPkgs[commit]; !ok {
+					relevantCommits = append(relevantCommits, commit)
+				}
+				affectedPkgs[commit] = append(affectedPkgs[commit], pkg)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
quartzmo: Fixed, thanks!

internal/gapicgen/generator/generator.go
```diff
@@ -69,7 +71,43 @@ func gatherChanges(googleapisDir, genprotoDir string) ([]*git.ChangeInfo, error)
 	if err != nil {
 		return nil, err
 	}
-	changes, err := git.ParseChangeInfo(googleapisDir, commits)
+	affectedProtos := make(map[string][]string)
```
shollyman, line 74: My question out the gate is whether we care about other artifacts like service config yaml files.  e.g. if a service changes its retry policy, do we need to propagate that?

bhshkh (dismissed):

internal/gapicgen/git/git.go
```diff
@@ -182,9 +191,9 @@ func DeepClone(repo, dir string) error {
 	return err
 }
 
-// filesChanged returns a list of files changed in a commit for the provdied
+// FilesChanged returns a list of files changed in a commit for the provdied
```
shollyman: nit: since you're already touching the line, maybe fix up the s/provdied/provided/ typo

shollyman (dismissed):
I have silly questions about the new filtration, but LGTMing because I assume the error is in my understanding.

internal/gapicgen/git/git.go
quartzmo: Fixed!

internal/gapicgen/generator/generator.go
quartzmo: Here's the list of what's in scope for `genbot` currently. Its scope has been much reduced since the days when go-genproto contained all the GAPIC protos. I don't think there are any GAPIC-related files like service config yamls to worry about anymore: https://github.com/googleapis/google-cloud-go/blob/main/internal/gapicgen/generator/genproto.go#L56

PR #13190, opened 2025-10-21
doc(storage): add an example for MRD in storage example_test.go

cpriti-os:
This PR adds an example for using the `MultiRangeDownloader` in the `storage` package.

Fixes b/389735108


storage/example_test.go
```diff
@@ -1034,3 +1035,47 @@ func ExampleGenerateSignedPostPolicyV4() {
...
+	callback := func(offset, length int64, err error) {
+		if err != nil {
+			rangeErr = err
+		}
+	}
+
+	// User creates an io.Writer (e.g. a buffer) and adds it to the reader
+	// with a particular range.
+	b := new(bytes.Buffer)
+	mrd.Add(b, 0, 100, callback)
```
BrennaEpp: We should probably show an example adding multiple ranges.

storage/example_test.go
```diff
@@ -1034,3 +1035,47 @@ func ExampleGenerateSignedPostPolicyV4() {
...
+
+	// Wait for downloads to complete before closing the stream.
```
BrennaEpp: I would avoid mentioning "the stream" here as that is an internal detail.

storage/example_test.go
```diff
@@ -1034,3 +1035,47 @@ func ExampleGenerateSignedPostPolicyV4() {
...
+	mrd.Wait()
+	if err := mrd.Close(); err != nil {
+		// TODO: handle stream close error if any
```
BrennaEpp: There may be other errors here as well so I would avoid mentioning this here as well.

storage/example_test.go
```diff
@@ -1034,3 +1035,47 @@ func ExampleGenerateSignedPostPolicyV4() {
...
+	var rangeErr error
+	// Callback registered by user to be called upon completion of a range.
+	callback := func(offset, length int64, err error) {
+		if err != nil {
+			rangeErr = err
+		}
+	}
+
+	// User creates an io.Writer (e.g. a buffer) and adds it to the reader
+	// with a particular range.
```
BrennaEpp: I think we should add a comment on what the intention here is.

storage/example_test.go
```diff
@@ -1034,3 +1035,47 @@ func ExampleGenerateSignedPostPolicyV4() {
...
+		// TODO: handle error.
+	}
+
+	var rangeErr error
+	// Callback registered by user to be called upon completion of a range.
+	callback := func(offset, length int64, err error) {
+		if err != nil {
+			rangeErr = err
+		}
+	}
```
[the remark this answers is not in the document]
BrennaEpp: I think gemini has a point - while I don't like its solution very much. This can cause potentially unexpected behaviour if more than one range uses this callback, including failing `-race` tests for users. I would suggest either recording every error (maybe using errors.Join?), or explicitly stating in a comment that the error gets overwritten and that the logic should probably be changed if using the same callback for multiple ranges.

BrennaEpp (commented):
Nice sample! Some quick comments

storage/example_test.go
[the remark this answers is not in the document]
cpriti-os: Done

storage/example_test.go
cpriti-os: Done

storage/example_test.go
cpriti-os: Done

storage/example_test.go
cpriti-os: Added another.

storage/example_test.go
cpriti-os: Added.

storage/example_test.go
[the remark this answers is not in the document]
tritone: Using a mutex here is not really something we want to encourage; it could lead to deadlocks in MRD when it calls the callback!

I would suggest either 1. sending the errors back to the caller on a channel and collecting them that way or 2. have specific err1, err2, etc variables for errors from specific ranges, or using a fixed-size array to hold them.

tritone (commented):
looks good other than removing the mutex as noted in my comment reply.

storage/example_test.go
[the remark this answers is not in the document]
cpriti-os: Done.

storage/example_test.go
```diff
@@ -1034,3 +1035,62 @@ func ExampleGenerateSignedPostPolicyV4() {
...
+	// Callback registered by user to be called upon completion of a range.
+	callback := func(offset, length int64, err error) {
+		if err != nil {
+			errChan <- err
+		}
+	}
+
+	// Goroutine to collect errors from the channel.
+	var rangeErrs []error
+	errDone := make(chan struct{})
```
tritone: This channel looks unused?

cpriti-os: Was using it to signal done, used waitgroup now which is idiomatic for this scenario.

storage/example_test.go
```diff
@@ -1034,3 +1036,72 @@ func ExampleGenerateSignedPostPolicyV4() {
...
+		if err != nil {
+			errChan <- err
+		}
+	}
+
+	// User creates an io.Writer (e.g. a buffer) and adds it to the
+	// MultiRangeDownloader with a particular range. Data will be downloaded
+	// into the buffer.
+	b1 := new(bytes.Buffer)
+	mrd.Add(b1, 0, 100, callback)
```
krishnamd-jkp, line 1084: nit: because this is a sample, could make the numbers more meaningful using variable names instead of passing numbers directly to understand the parameters better.

tritone (approved):

PR #13205, opened 2025-10-22
feat(storage): calculate crc32c by default and pass checksum in trailing and per-chunk request

krishnamd-jkp:
calculate crc32c by default and pass checksum in trailing and per-chunk request

### DIFF
--- a/storage/writer.go
+++ b/storage/writer.go
@@ -41,28 +41,33 @@ type Writer struct {
 	// attributes are ignored.
 	ObjectAttrs
 
-	// SendCRC32C specifies whether to transmit a CRC32C field. It should be set
-	// to true in addition to setting the Writer's CRC32C field, because zero
-	// is a valid CRC and normally a zero would not be transmitted.
-	// If a CRC32C is sent, and the data written does not match the checksum,
-	// the write will be rejected.
+	// SendCRC32C specifies whether to transmit a CRC32C checksum. When this is
+	// true and the Writer's CRC32C field is set, that checksum is sent to GCS.
+	// If the data written does not match the checksum, the write is rejected.
+	// It is necessary to set this field to true in addition to setting the
+	// Writer's CRC32C field because zero is a valid CRC.
 	//
-	// Note: SendCRC32C must be set to true BEFORE the first call to
-	// Writer.Write() in order to send the checksum. If it is set after that
-	// point, the checksum will be ignored.
+	// When using gRPC, the client automatically calculates and sends checksums
+	// per-chunk and for the full object. However, A user-provided checksum takes
+	// precedence over the auto-calculated checksum for full object.
+	// To disable auto checksum behavior, see DisableAutoChecksum.
+	//
+	// Note: SendCRC32C must be set before the first call to Writer.Write().
 	SendCRC32C bool
 
-	// DisableAutoChecksum disables the automatic CRC32C checksum calculation and
-	// validation (both chunk-wise and full object checksum) that is performed
-	// by the writer.
+	// DisableAutoChecksum disables automatic CRC32C checksum calculation and
+	// validation in gRPC Writer. By default when using gRPC, the Writer
+	// automatically performs checksum validation for both individual chunks and
+	// the entire object. Setting this to true disables this behavior. This flag
+	// is ignored when not using gRPC.
 	//
-	// However, This does not prevent a user-provided full-object checksum from being sent.
-	// If SendCRC32C is true and the Writer's CRC32C field is populated, that
-	// checksum will be sent to GCS for validation by the gRPC writer on final write.
+	// Disabling automatic checksumming does not prevent a user-provided checksum
+	// from being sent. If SendCRC32C is true and the Writer's CRC32C field is
+	// populated, that checksum will still be sent to GCS for validation.
 	//
-	// Note: DisableAutoChecksum must be set to true BEFORE the first call to
-	// Writer.Write(). This flag Works only with gRPC writer.
-	// Does not work with appendable writes with unfinalized objects
+	// Note: DisableAutoChecksum must be set before the first call to
+	// Writer.Write(). Automatic checksumming is not enabled for writes
+	// using the HTTP client or for unfinalized writes to appendable objects in gRPC.
 	DisableAutoChecksum bool
 
 	// ChunkSize controls the maximum number of bytes of the object that the


storage/grpc_writer.go
```diff
@@ -888,7 +935,15 @@ func (s *gRPCOneshotBidiWriteBufferSender) connect(ctx context.Context, cs gRPCB
...
 
-			req := bidiWriteObjectRequest(r.buf, r.offset, r.flush, r.finishWrite)
+			sendCrc32C, disableCrc32c, attrs := s.checksumSettings()
+			var checksumOfBuf *uint32
+			if !disableCrc32c {
+				checksumOfBuf = proto.Uint32(crc32.Checksum(r.buf, crc32cTable))
+			}
+
+			objectChecksums := getFinalChecksums(s.fullObjectChecksum(), r.finishWrite, sendCrc32C, disableCrc32c, attrs)
+			req := bidiWriteObjectRequest(r.buf, r.offset, r.flush, r.finishWrite, checksumOfBuf, objectChecksums)
```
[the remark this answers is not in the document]
cpriti-os: +1

storage/grpc_writer_test.go
cpriti-os, line 1: Could you check if we can add an integration test to validate end-to-end functionality?

storage/grpc_writer.go
```diff
@@ -51,6 +53,12 @@ func (w *gRPCWriter) Write(p []byte) (n int, err error) {
 	case <-w.donec:
 		return 0, w.streamResult
 	case w.writesChan <- cmd:
+		// update fullObjectChecksum on every write and send it on finalWrite
+		if !w.disableCRC32C {
+			w.checksumMutex.Lock()
+			w.fullObjectChecksum = crc32.Update(w.fullObjectChecksum, crc32cTable, p)
+			w.checksumMutex.Unlock()
```
cpriti-os: Use defer to unlock: its safer and more idiomatic go pattern

storage/grpc_writer.go
```diff
@@ -785,25 +798,54 @@ func completion(r *storagepb.BidiWriteObjectResponse) *gRPCBidiWriteCompletion {
 	}
 }
 
-func bidiWriteObjectRequest(buf []byte, offset int64, flush, finishWrite bool) *storagepb.BidiWriteObjectRequest {
+// Server contract expects full object checksum to be sent only on first or last write.
+// Checksums of full object are already being sent on first Write during initialization of sender.
+// Send objectChecksums only on final request and nil in other cases.
+func bidiWriteObjectRequest(buf []byte, offset int64, flush, finishWrite bool, checksumOfBuf *uint32, objectChecksums *storagepb.ObjectChecksums) *storagepb.BidiWriteObjectRequest {
```
cpriti-os: [Best practice for function argument list](https://g3doc.corp.google.com/go/g3doc/style/best-practices.md?content_ref=where+a+function+requires+many+inputs+consider+introducing+an+option+struct+for+some+of+the+arguments+or+employing+the+more+advanced+variadic+options+technique&cl=head#funcargs)

cpriti-os: Nit: bufChecksum would be more consistent

storage/grpc_writer.go
```diff
@@ -785,25 +798,54 @@ func completion(r *storagepb.BidiWriteObjectResponse) *gRPCBidiWriteCompletion {
...
+		WriteOffset:     offset,
+		FinishWrite:     finishWrite,
+		Flush:           flush,
+		StateLookup:     flush,
+		ObjectChecksums: objectChecksums,
 	}
 	return req
 }
 
+// getObjectChecksums determines what checksum information to include in the final
```
cpriti-os: Nit: Provide shorter and direct comments
```
// getObjectChecksums returns the checksums for the final write request. It
// returns nil if finishWrite is false or disableCRC32C is set. User-provided
// checksums are sent if available, otherwise, the computed checksum is returned .
```

storage/grpc_writer_test.go
```diff
@@ -0,0 +1,86 @@
...
+				Crc32C: proto.Uint32(456),
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := getObjectChecksums(tt.fullObjectChecksum, tt.finishWrite, tt.sendCRC32C, tt.disableCRC32C, tt.attrs)
+			if !proto.Equal(got, tt.want) {
+				t.Errorf("getFinalChecksums() = %v, want %v", got, tt.want)
```
cpriti-os: do you mean getObjectChecksums()?

storage/writer.go
```diff
@@ -52,6 +52,20 @@ type Writer struct {
 	// point, the checksum will be ignored.
 	SendCRC32C bool
 
+	// DisableCRC32C disables the automatic CRC32C checksum calculation and
+	// validation that is performed by default.
```
cpriti-os: Nit: This comment is slightly redundant in parts, could you make it concise while keeping the necessary details

storage/grpc_writer_test.go
```diff
@@ -0,0 +1,86 @@
...
+package storage
+
+import (
+	"testing"
+
+	"cloud.google.com/go/storage/internal/apiv2/storagepb"
+	"google.golang.org/protobuf/proto"
+)
+
+func TestGetObjectChecksums(t *testing.T) {
```
cpriti-os, line 24: For better coverage, add unit tests for per-chunk checksum and test the scenario for full object checksum accumulation.

cpriti-os (changes requested):

storage/grpc_writer.go
krishnamd-jkp: Yeah thought of adding a defer statement but didn't make sense for a 1 liner. Also, the critical section doesn't throw any error.

storage/grpc_writer_test.go
krishnamd-jkp: The current integration tests already cover these changes. Any wrong calculation in checksum would lead to failure of tests like "TestIntegration_WriterChunksize", "TestIntegration_WriterContentType".

storage/grpc_writer.go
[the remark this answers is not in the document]
krishnamd-jkp: The repetitive logic is not big enough to create a function out of it and not logically cohesive and would reduce readability in my opinion. We can create a function if the logic grows in the future. What do you think?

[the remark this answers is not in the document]
krishnamd-jkp: simplified

storage/grpc_writer.go
krishnamd-jkp: Reused gRPCBidiRequest to reduce the arguments

krishnamd-jkp:
### Adhoc benchmarking results:
#### Metrics for 10 files  64MB each with 10 workers running parallel with test execution of 5 minutes

**Without warmup**
*release 1.57.0*
Total throughput extrapolated to 100MB: 0.22/GiB/s
Median throughput extrapolated to 100MB:   0.26 GiB/s
Median upload time: 3.7661s
P10 upload time: 3.1413s
P90 upload time: 6.1835s

*PR*
Total throughput extrapolated to 100MB: 0.27/GiB/s
Median throughput extrapolated to 100MB:   0.24 GiB/s
Median upload time: 3.633s
P10 upload time: 2.894s
P90 upload time: 6.295s

**With 5 min warmup**
*release 1.57.0*
Total throughput extrapolated to 100MB: 0.29 GiB/s
Median throughput extrapolated to 100MB:   0.31 GiB/s
Median upload time: 3.1737s
P10 upload time: 2.8136s
P90 upload time: 4.3056s

*PR*
Total throughput extrapolated to 100MB: 0.25 GiB/s
Median throughput extrapolated to 100MB:   0.30 GiB/s
Median upload time: 3.2925s
P10 upload time: 2.8041s
P90 upload time: 5.9242s

storage/grpc_writer.go
```diff
@@ -239,8 +248,12 @@ type gRPCWriter struct {
 	setSize           func(int64)
 	setTakeoverOffset func(int64)
 
+	fullObjectChecksum uint32
+	checksumMutex      sync.RWMutex
```
tritone: Given that this code was just refactored to use channels to send commands, I am dubious about adding a mutex here. Is there a way we can send the checksum result on a channel to process rather than creating shared state that needs to be managed via mutex? See https://go.dev/wiki/MutexOrChannel

storage/writer.go
```diff
@@ -52,6 +52,17 @@ type Writer struct {
 	// point, the checksum will be ignored.
 	SendCRC32C bool
 
+	// DisableCRC32C disables the automatic CRC32C checksum calculation and
```
tritone: Note that this only works for gRPC, and does not work for unfinalized writes to appendable objects. I would also rename it something like DisableAutoChecksum maybe?

Also, I think the docs for both this and SendCRC32C are a little unclear how they work together. I would want the user to understand:
1. If they provide a checksum via SendCRC32C, we will not do checksum calculations and send their checksum on the first request to GCS.
2. If they do not provide a checksum via SendCRC32C, do not set DisableCRC32C, and this is a finalized write via gRPC, we will automatically calculate the checksum and send it on the last message.
3. If they set DisableCRC32C we will never calculate or send a checksum

Does that sound right? And does it match the actual behavior?

storage/grpc_writer.go
```diff
@@ -785,25 +798,54 @@ func completion(r *storagepb.BidiWriteObjectResponse) *gRPCBidiWriteCompletion {
...
 }
 
+// getObjectChecksums determines what checksum information to include in the final
+// gRPC request
+//
+// function returns a populated ObjectChecksums only when finishWrite is true
+// If CRC32C is disabled, it returns the user-provided checksum if available.
+// If CRC32C is enabled, it returns the user-provided checksum if available,
+// or the computed checksum of the entire object.
+func getObjectChecksums(fullObjectChecksum func() uint32, finishWrite, sendCRC32C, disableCRC32C bool, attrs *ObjectAttrs) *storagepb.ObjectChecksums {
```
cpriti-os: Nit: This function takes in 4 boolean flags which can make it difficult to read and use, consider using a struct.

storage/writer.go
```diff
@@ -52,6 +52,17 @@ type Writer struct {
...
+	// validation that is performed by the writer.
+	//
+	// This does not prevent a user-provided checksum from being sent. If
+	// SendCRC32C is true and the Writer's CRC32C field is populated, that
+	// checksum will be sent to GCS for validation
+	//
+	// Note: DisableCRC32C must be set to true BEFORE the first call to
+	// Writer.Write().
+	DisableCRC32C bool
```
cpriti-os: Adding this as a struct field will allow users to write some data with auto checksum enabled and some with checksum disabled by setting this field after writing some data. Its worth thinking if this should be set/unset only when the Writer is created or is this flexibility required/intended?

cpriti-os (changes requested):

storage/writer.go
krishnamd-jkp: Yeah I think DisableAutoChecksum name works better.
The behavior is - 
1. If DisableAutoChecksum is set, checksum calculation in the writer is disabled i.e., both chunk-wise checksum will be disabled and full object checksum calculation is also disabled in the writer. However, if user configures their checksum, it will be sent on both first and last write
2. If DisableAutoChecksum is not set, chunk-wise calculation is sent to GCS by the writer. On the final write, the grpc writer prioritizes user's checksum over auto calculated checksum. So on the last write, if user's checksum is provided, writer sends user's checksum to GCS. If user doesn't specify any checksum, auto calculated checksum will be sent to GCS.

krishnamd-jkp:
@tritone - 
> Couple comments... I would also suggest the following:
> 
> * Take a look at #12477; do we want to support user-provided trailing checksums as well?

This PR handles trailing checksums as well. Please check "getObjectChecksums" method.
`

	if !finishWrite {
    		return nil
	}

	// send user's checksum on last write op if available
	if sendCRC32C {
		return toProtoChecksums(sendCRC32C, attrs)
	}
`

storage/grpc_writer.go
krishnamd-jkp: I think because this fullObjectChecksum is more state-like and less communication-like, I feel adding a mutex is the simplest approach here. Also because this is incremental calculation of checksum, the writer should stream incomplete full checksums which is of no use. There should be another go routine listening to these incomplete checksums and update the state anyways.

Also, lock is just like a preventive measure here because when the sender(resumable or one-shot) reads the full object checksum on final write, we expect no more write calls so ideally there are no data race conditions here but I added the lock just in case. Do you still think we should go with channels here?

krishnamd-jkp:
CPU profile benchmarking - 

The profile was captured over a duration of 300.11 seconds, with a total of 137.21 seconds of CPU time sampled. 

Top 5 CPU consuming functions - 

   * System Calls (`internal/runtime/syscall.Syscall6`):  66.58s (48.52%) 
   * Encryption (`crypto/.../gcm.gcmAesEnc`): 15.00s (10.93%)
   * Memory Operations (`runtime.memmove`): 14.66s (10.68%) of the CPU time
   * Concurrency (`runtime.futex`): 10.08s (7.35%).
   * **Checksumming (`hash/crc32.castagnoliSSE42Triple`):  8.38s (6.11%) of the CPU time.**

storage/grpc_writer.go
tritone: IMO if we're just doing this preventatively there should be no mutex at all right? Adding a mutex makes the code more complicated to read and reason about, and it does have some CPU over head (which can get worse in the future if we add additional locking elsewhere in the writer). The Writer is not expected to be threadsafe to multiple write/close calls in parallel.

On the other hand if removing it triggers the race detector, then we probably do need to synchronize. If the user of the partial checksum has access to the gRPCWriterCommandWrite then we should probably just add it to that data type and send it along with that?

storage/writer.go
tritone: Ah, then maybe we should disambiguate between per-message checksums and whole-object checksums? If this is already implemented in other clients, do they offer separate options for each of these, or just one?

tritone:
> CPU profile benchmarking -
> 
> The profile was captured over a duration of 300.11 seconds, with a total of 137.21 seconds of CPU time sampled.
> 
> Top 5 CPU consuming functions -
> 
> * System Calls (`internal/runtime/syscall.Syscall6`):  66.58s (48.52%)
> * Encryption (`crypto/.../gcm.gcmAesEnc`): 15.00s (10.93%)
> * Memory Operations (`runtime.memmove`): 14.66s (10.68%) of the CPU time
> * Concurrency (`runtime.futex`): 10.08s (7.35%).
> * **Checksumming (`hash/crc32.castagnoliSSE42Triple`):  8.38s (6.11%) of the CPU time.**

Cool, I would say this is around what I would have expected. We probably should note in the godoc that SDK auto checksumming has some amount of increased CPU overhead.

storage/writer.go
krishnamd-jkp: Can you check if this clarifies the behavior?

storage/grpc_writer.go
krishnamd-jkp: Removed the unnecessary lock. validated by running conformance tests which previously showed the race conditions.

storage/grpc_writer_test.go
krishnamd-jkp: This would be a integration test and not a unit test. I'll add it in another PR because currently the existing integration tests cover the checksum calculation and would fail if the checksum is wrongly calculated.

storage/writer.go
```diff
@@ -52,6 +52,19 @@ type Writer struct {
...
+	// DisableAutoChecksum disables the automatic CRC32C checksum calculation and
+	// validation (both chunk-wise and full object checksum) that is performed
+	// by the writer.
+	//
+	// However, This does not prevent a user-provided full-object checksum from being sent.
+	// If SendCRC32C is true and the Writer's CRC32C field is populated, that
+	// checksum will be sent to GCS for validation by the gRPC writer on final write.
+	//
+	// Note: DisableAutoChecksum must be set to true BEFORE the first call to
+	// Writer.Write(). This flag Works only with gRPC writer.
```
tritone: IMO this is still not super clear. We should communicate that automatic checksumming only works with gRPC, not specifically this flag. The godoc for SendCRC32C should probably be updated as well.

Where did we land on allowing the user to control chunk vs whole object checksums independently only giving one flag? If we want to allow fine-grained control, we could make this field something like `*AutoChecksumConfig` with separate bools for per-message and per-object.

Also, remove random capitalized words in the godoc (`Works`, `This`).

storage/writer.go
```diff
@@ -52,6 +52,19 @@ type Writer struct {
...
+	// Does not work with appendable writes with unfinalized objects
```
tritone: This is not really accurate, we still do per-message for these right?

storage/grpc_writer.go
```diff
@@ -51,6 +52,10 @@ func (w *gRPCWriter) Write(p []byte) (n int, err error) {
 	case <-w.donec:
 		return 0, w.streamResult
 	case w.writesChan <- cmd:
+		// update fullObjectChecksum on every write and send it on finalWrite
+		if !w.disableAutoChecksum {
+			w.fullObjectChecksum = crc32.Update(w.fullObjectChecksum, crc32cTable, p)
```
tritone, line 57: Just wanted to note that you are doing the work twice here for each set of bytes between here and L945. In theory it should be possible to calculate the per-message checksum once per buffer and then use those sums to update the full object checksum as well. It doesn't look like there is an easy interface to do this with in Go, but maybe worth considering if you are trying to save CPU.

storage/writer.go
krishnamd-jkp: No in this PR we have not added anything in appendable writes. Will implement checksumming for appendable writes in a different PR and remove this line.

storage/grpc_writer.go
krishnamd-jkp: In case of no retries, yes, it does seem like the same calculation is being done twice. But in case of retries, the buffer in this line and the buffer in L945 will be out of sync. And I cannot use the checksum in L945 to update the global checksum because we could be using same bytes multiple times in case of retries. So I had to separate these two computations.

storage/writer.go
krishnamd-jkp: I think giving the user an fine-grained option to disable checksum per-chunk and whole object individually would confuse the users given these are default settings. Made some changes. I think it clarifies this a bit.

storage/writer.go
```diff
@@ -41,17 +41,33 @@ type Writer struct {
...
+	// automatically performs checksum validation for both individual chunks and
+	// the entire object. Setting this to true disables this behavior. This flag
+	// is ignored when not using gRPC.
+	//
+	// Disabling automatic checksumming does not prevent a user-provided checksum
+	// from being sent. If SendCRC32C is true and the Writer's CRC32C field is
+	// populated, that checksum will be sent to GCS for validation.
+	//
+	// Note: DisableAutoChecksum must be set before the first call to
+	// Writer.Write(). It does not work with appendable writes for unfinalized
```
cpriti-os: Nit: More clear and accurate statement would be: `This feature is not yet supported in appendable writes for unfinalized objects.`

krishnamd-jkp: Noted. Will remove the comment anyways in the new PR when checksums are implemented for appendable writes.

cpriti-os (approved):

PR #13216, opened 2025-10-23
fix(spanner): Configure keepAlive time for gRPC TCP connections

sakthivelmanii:
Default keep alive time is 120 seconds. If customer wants to override they can override it by passing the keepAlive configuration in the opts.

### DIFF
--- a/spanner/client.go
+++ b/spanner/client.go
@@ -44,6 +44,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/encoding/gzip"
+	"google.golang.org/grpc/keepalive"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/peer"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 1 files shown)


surbhigarg92 (approved):

spanner/client.go
```diff
@@ -683,6 +684,7 @@ func allClientOpts(numChannels int, compression string, userOpts ...option.Clien
 		option.WithUserAgent(fmt.Sprintf("spanner-go/v%s", internal.Version)),
 		option.WithGRPCDialOption(grpc.WithChainUnaryInterceptor(addNativeMetricsInterceptor()...)),
 		option.WithGRPCDialOption(grpc.WithChainStreamInterceptor(addStreamNativeMetricsInterceptor()...)),
+		option.WithGRPCDialOption(grpc.WithKeepaliveParams(keepalive.ClientParameters{Time: 120 * time.Second})),
```
olavloite, line 687: Do we have any tests that verify that these defaults are actually used, and that the precedence that is mentioned in the comment is actually applied?

olavloite (approved):

spanner/client.go
sakthivelmanii: Discussed offline since we dont have an option

PR #13184, opened 2025-10-20
chore: migrate Spanner to Librarian

jskeet:
Fixes https://github.com/googleapis/librarian/issues/2422 - see that issue for more details.

Note: this shouldn't be merged until we've done a release with the latest changes from OwlBot (unless we're happy to manually change the next release PR).

After this PR is merged:

- Generation of API changes will still be automatic, via Librarian instead of OwlBot
- Releases will be manually initiated: @quartzmo and I will be happy to pair with the Spanner team for the first release or two

I believe the Python team will be working on a similar PR (or multiple ones) this week.

PR #13219, opened 2025-10-24
chore(.librarian): prevent spanner auto-releasing

jskeet:


jskeet:
This was intended to be in #13184.

PR #13224, opened 2025-10-24
chore(vertexai): migrate to Librarian

jskeet:
Fixes https://github.com/googleapis/librarian/issues/2427


jskeet:
@qiaodev After this is merged, releases will need to be done by Librarian. I understand this module is at least *mostly* deprecated - it's possible you won't even need another release. But if you do, it would be best to pair with myself or @quartzmo for the first release.

It looks like there *is* generated code in this module, but it's being manually updated - assuming I'm correct, that process won't change. (Librarian doesn't "know" about any APIs to run the generator for, for this module.)

qiaodev (approved):

.release-please-manifest-individual.json
```diff
@@ -4,6 +4,5 @@
   "errorreporting": "0.3.2",
   "logging": "1.13.0",
   "profiler": "0.4.3",
-  "storage": "1.57.0",
-  "vertexai": "0.15.0"
+  "storage": "1.57.0"
```
quartzmo, line 7: Was this a bad merge from main? Why restore `storage` to release-please in this PR?

jskeet: It's not actually restoring anything - it's just removing the comma from the end of the line. (We haven't migrated Storage yet.)

PR #13226, opened 2025-10-24
feat(internal): implement Bigtable specific channel pool optimizations

sushanb:
1. The hot path (selecting a connection for an RPC) is highly optimized for performance. as The list of connections is stored in an atomic.Value, and load counters are managed with atomic operations
2. The pool automatically detects and evicts the single worst-performing unhealthy connection at a regular interval.
3. If the percentage of unhealthy connections exceeds a high-water mark (PoolwideBadThreshPercent), all evictions are suspended to avoid overwhelming the system during a wider service degradation.
4. Evictions are rate-limited by a minimum interval (MinEvictionInterval) to ensure stability.
5. A ChannelHealthMonitor runs in the background, periodically probing each connection in the pool.
6. Probes are performed by sending a PingAndWarm RPC to the Bigtable backend, verifying end-to-end connectivity.
7. Connection health is evaluated based on the percentage of failed probes over a sliding time window (WindowDuration). A connection is marked unhealthy if its failure rate exceeds a configurable threshold (FailurePercentThresh).

### DIFF
--- a/bigtable/bigtable.go
+++ b/bigtable/bigtable.go
@@ -203,9 +203,9 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
 			nil,
 			btransport.WithHealthCheckConfig(btopt.DefaultHealthCheckConfig()),
 			btransport.WithDynamicChannelPool(btopt.DefaultDynamicChannelPoolConfig(defaultBigtableConnPoolSize)),
-			btransport.WithMetricsReporterConfig(btopt.DefaultMetricsReporterConfig())
+			btransport.WithMetricsReporterConfig(btopt.DefaultMetricsReporterConfig()),
 		)
-		connPool, connPoolErr = btransport.NewBigtableChannelPool(ctx, defaultBigtableConnPoolSize, btopt.BigtableLoadBalancingStrategy(), dialFunc, config.Logger, meterP)
+		connPool, connPoolErr = btransport.NewBigtableChannelPool(ctx, defaultBigtableConnPoolSize, btopt.BigtableLoadBalancingStrategy(), dialFunc, config.Logger, nil)
 	} else {
 		// use to regular ConnPool
 		connPool, connPoolErr = gtransport.DialPool(ctx, o...)


sushanb:
FYI. i have a TODO to make DynamicChannelScaleup, MetricsExporter, HealthChecker all part of BackgroundProcess interface.

bigtable/internal/transport/connpool.go
```diff
@@ -16,53 +16,600 @@ package internal
...
+	return chs.failedProbes
+}
+
+// connEntry represents a single connection in the pool.
+type connEntry struct {
+	conn          *BigtableConn
+	unaryLoad     int32           // In-flight unary requests
+	streamingLoad int32           // Active streams
+	health        connHealthState // Embedded health state
+	altsUsed      int32           // Set to 1 atomically if ALTS is used, 0 otherwise.
```
nimf: This thing is per-RPC, not per ClientConn. If the intention is to only save what's first PingAndWarm went through, this may be fine, but it gives a false impression that this conn uses ALTS for all calls.

sushanb: yeah we only cache the first rpc (prime()). We added a transport_type label for the metric so it's the best effort basis.

nimf:
Few high-level comments from the first pass:

1. Concurrent scaling down and picking a connection.
There is a chance that while `selectFunc` picks a connection to send RPC to, the connection may be closed by `removeConnections`. The RPC will be refused immediately on a closed connection. We should have some protection or workaround here.
1. Concurrent scaling and eviction.
While this doesn't crash anything it can still bring surprising edge cases. For example, there is a small chance that the pool can scale up/down while `detectAndEvictUnhealthy` is picking an index of a channel to evict. As the `detectAndEvictUnhealthy` works with a snapshot of channels prior to scaling event, the index it passes to `replaceConnection` may point to a "good" channel.
Maybe we should pass a pointer to a channel instead, or restrict running scaling and eviction at the same time.
1. It feels like `ChannelHealthMonitor` and `DynamicScaleMonitor` should live in their own files (maybe same package though).

sushanb:
> Few high-level comments from the first pass:
> 
> 1. Concurrent scaling down and picking a connection.
>    There is a chance that while `selectFunc` picks a connection to send RPC to, the connection may be closed by `removeConnections`. The RPC will be refused immediately on a closed connection. We should have some protection or workaround here.

I added  a drainingState atomic.Bool   in connEntry. 
We avoid choosing a conn in  drainState and actively wait for a period of time for the load on the chosen conn to evict to be zero and Close() it. Close() is thread safe so can be called multiple times on conn.



> 2. Concurrent scaling and eviction.
>    While this doesn't crash anything it can still bring surprising edge cases. For example, there is a small chance that the pool can scale up/down while `detectAndEvictUnhealthy` is picking an index of a channel to evict. As the `detectAndEvictUnhealthy` works with a snapshot of channels prior to scaling event, the index it passes to `replaceConnection` may point to a "good" channel.

replaceConnection takes the actual conn pointer rather than index which avoids this bug. Thanks for finding the bug.


>    Maybe we should pass a pointer to a channel instead, or restrict running scaling and eviction at the same time.
> 3. It feels like `ChannelHealthMonitor` and `DynamicScaleMonitor` should live in their own files (maybe same package though).

Will refactor.

bigtable/internal/transport/connpool_test.go
```diff
@@ -114,131 +202,285 @@ func setupTestServer(t *testing.T, service *fakeService) string {
...
+		if entry != conns[expectedIdx] {
+			t.Errorf("selectRoundRobin call %d got entry for index %d, want %d", i+1, entryIndex(conns, entry), expectedIdx)
 		}
 	}
 }
 
-func TestSelectLeastLoadedRandomOfTwo(t *testing.T) {
-	pool := &BigtableChannelPool{}
-
-	// Test empty pool
```
nimf, line 165: No action needed. But for the next time:
This test file's diff on GitHub is pretty hard to review, because it shows as if functions were replaced but they are not. I got a much saner diff locally using `--diff-algorithm=patience` but that's extra moves/efforts.
Maybe better to keep comments like this, because then the diff algorithm would have additional "anchors" to produce more readable diff.

bigtable/internal/option/option.go
```diff
@@ -201,3 +202,103 @@ func EnableBigtableConnectionPool() bool {
...
+// DynamicChannelPoolConfig holds the parameters for dynamic channel pool scaling.
+type DynamicChannelPoolConfig struct {
+	Enabled              bool          // Whether dynamic scaling is enabled.
+	MinConns             int           // Minimum number of connections in the pool.
+	MaxConns             int           // Maximum number of connections in the pool.
+	AvgLoadHighThreshold int32         // Average weighted load per connection to trigger scale-up.
+	AvgLoadLowThreshold  int32         // Average weighted load per connection to trigger scale-down.
+	MinScalingInterval   time.Duration // Minimum time between scaling operations (both up and down).
+	CheckInterval        time.Duration // How often to check if scaling is needed.
+	MaxRemoveConns       int           // Maximum number of connections to remove at once.
```
igorbernstein2: do we want these to be publicly accessible? Maybe make them internal and pull them from an env var....These bleed too much impl detail and will tie our hands if we want to use a different pool approach

bigtable/internal/transport/connpool.go
```diff
@@ -16,53 +16,344 @@ package internal
...
+
+	client := btpb.NewBigtableClient(bc.ClientConn)
+	req := &btpb.PingAndWarmRequest{
+		Name:         bc.instanceName,
+		AppProfileId: bc.appProfile,
+	}
+
+	// Use a timeout for the prime operation
+	primeCtx, cancel := context.WithTimeout(ctx, primeRPCTimeout)
+	defer cancel()
```
igorbernstein2: why do you need to cancel? the next line guarantees that the rpc is done and there is nothing to cancel

bigtable/internal/transport/connpool_test.go
```diff
@@ -361,121 +363,363 @@ func TestPoolNewStream(t *testing.T) {
 			if err := stream.RecvMsg(res); err != nil {
 				t.Fatalf("RecvMsg failed: %v", err)
 			}
-			if string(res.GetPayload().GetBody()) != "msg1" {
-				t.Errorf("RecvMsg got %q, want %q", string(res.GetPayload().GetBody()), "msg1")
```
nimf, line 365: why are we removing this check?

bigtable/internal/transport/connpool_test.go
```diff
@@ -361,121 +363,363 @@ func TestPoolNewStream(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("NewBigtableChannelPool failed: %v", err)
+		}
+		defer pool.Close()
+
+		if pool.Num() != poolSize {
+			t.Errorf("Pool size got %d, want %d", pool.Num(), poolSize)
+		}
+		// Wait for priming goroutines to likely complete
+		time.Sleep(100 * time.Millisecond)
```
nimf, line 462: no action needed, just some thoughts for future improvements. Here and everywhere in the tests where we use time.Sleep or wait for some goroutine to do its job -- such tests are usually prone to flakiness, so we should consider starting using [testing/synctest](https://go.dev/blog/synctest) package.

bigtable/internal/transport/connpool_test.go
```diff
@@ -361,121 +363,363 @@ func TestPoolNewStream(t *testing.T) {
...
+
+		if fake.getPingCount() < 1 {
```
nimf, line 464:
```suggestion
		if fake.getPingCount() < 5 {
```

bigtable/internal/transport/connpool_test_helper.go
```diff
@@ -0,0 +1,216 @@
+// Copyright 2025 Google LLC
```
nimf, line 1: If the filename is not ending with `_test.go` this will be included in build, right? I suppose we don't want that.

bigtable/internal/transport/connpool_test.go
```diff
@@ -19,262 +19,236 @@ import (
...
-	}
-	if err == nil {
-		t.Errorf("selectLeastLoadedRandomOfTwo on empty pool got nil error, want non-nil")
-	}
-	if err != errNoConnections {
-		t.Errorf("selectLeastLoadedRandomOfTwo on empty pool got error %v, want %v", err, errNoConnections)
-	}
-
-	// Test single connection pool
-	pool.conns = make([]*grpc.ClientConn, 1)
```
nimf, line 178: why removing the case of a single conn channel pool? This is a special case that is currently handled in the code, but may be removed and broken in the future, thus should be tested.

bigtable/bigtable.go
```diff
@@ -175,9 +181,31 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
...
+			defaultBigtableConnPoolSize,
+			btopt.BigtableLoadBalancingStrategy(),
+			dialFunc,
+			config.Logger,
+			nil,
+			btransport.WithHealthCheckConfig(btopt.DefaultHealthCheckConfig()),
+			btransport.WithDynamicChannelPool(btopt.DefaultDynamicChannelPoolConfig(defaultBigtableConnPoolSize)),
+			btransport.WithMetricsReporterConfig(btopt.DefaultMetricsReporterConfig()),
+		)
+		connPool, connPoolErr = btransport.NewBigtableChannelPool(ctx, defaultBigtableConnPoolSize, btopt.BigtableLoadBalancingStrategy(), dialFunc, config.Logger, nil)
```
nimf, line 208: is it duplicate? so should it be with dynamic channel pool, healthcheck, and metrics or not?

bigtable/internal/option/option.go
```diff
@@ -201,3 +202,103 @@ func EnableBigtableConnectionPool() bool {
...
+		logger.Printf(format, v...)
+	}
+}
+
+var debug = os.Getenv("CBT_ENABLE_DEBUG") == "true"
+
+// Debugf logs the given message *only if* the global Debug flag is true.
+// It reuses Logf to handle the nil logger logic and prepends "DEBUG: "
+// to the message.
+func Debugf(logger *log.Logger, format string, v ...interface{}) {
```
nimf, line 221: it is weird to see logging functionality in the `option` package. More like util or maybe just include in the connpool file? Also, should not be exported in that case.

bigtable/internal/option/option.go
```diff
@@ -201,3 +202,103 @@ func EnableBigtableConnectionPool() bool {
...
+	MaxConns             int           // Maximum number of connections in the pool.
+	AvgLoadHighThreshold int32         // Average weighted load per connection to trigger scale-up.
+	AvgLoadLowThreshold  int32         // Average weighted load per connection to trigger scale-down.
+	MinScalingInterval   time.Duration // Minimum time between scaling operations (both up and down).
+	CheckInterval        time.Duration // How often to check if scaling is needed.
+	MaxRemoveConns       int           // Maximum number of connections to remove at once.
+}
+
+// DefaultDynamicChannelPoolConfig is default settings for dynamic channel pool
+func DefaultDynamicChannelPoolConfig(initialConns int) DynamicChannelPoolConfig {
```
nimf, line 265: initialConns not used?

bigtable/internal/transport/connpool.go
```diff
@@ -75,30 +366,114 @@ func NewBigtableChannelPool(connPoolSize int, strategy btopt.LoadBalancingStrate
 		pool.selectFunc = pool.selectRoundRobin
 	}
 
+	initialConns := make([]*connEntry, connPoolSize)
 	for i := 0; i < connPoolSize; i++ {
```
nimf, line 370: what if we just `pool.addConnections(connPoolSize)`?

bigtable/internal/transport/connpool.go
```diff
@@ -108,123 +483,298 @@ func (p *BigtableChannelPool) Close() error {
...
+			probeCtx, cancel := context.WithTimeout(ctx, cfg.ProbeTimeout)
+			defer cancel()
+			// Check if context was already done before priming
+			select {
+			case <-probeCtx.Done():
+				e.health.addProbeResult(false, cfg.WindowDuration) // Count as failure if context is done
+				return
+			default:
+			}
+			// don't update conn  isAlts used entry for now.
```
nimf, line 505: I think we should update it. Channel can fallback to cloudpath, it is better to have updated info.

bigtable/internal/transport/connpool.go
```diff
@@ -108,123 +483,298 @@ func (p *BigtableChannelPool) Close() error {
...
+		idx2 := rand.Intn(numConns)
 
-	load1 := atomic.LoadInt64(&p.load[idx1])
-	load2 := atomic.LoadInt64(&p.load[idx2])
+		entry1 := conns[idx1]
+		entry2 := conns[idx2]
 
-	if load1 <= load2 {
-		return idx1, nil
+		if entry1.isDraining() || entry2.isDraining() {
```
nimf, line 715: maybe if any of them is draining, we pick the other; and only if both are draining we look for another pair?

bigtable/internal/transport/connpool.go
```diff
@@ -108,123 +483,298 @@ func (p *BigtableChannelPool) Close() error {
...
+			continue // Find another pair
+		}
+
+		if idx1 == idx2 {
```
nimf, line 719: why don't we look for another pair anymore when we pick the same index?

bigtable/internal/transport/dynamic_scale_monitor.go
```diff
@@ -0,0 +1,178 @@
...
+
+	var totalWeightedLoad int32
+	for _, entry := range conns {
+		totalWeightedLoad += entry.calculateWeightedLoad()
+	}
+	avgLoad := totalWeightedLoad / int32(numConns)
+
+	targetLoad := (dsm.config.AvgLoadLowThreshold + dsm.config.AvgLoadHighThreshold) / 2
+	if targetLoad == 0 {
+		targetLoad = 1
```
nimf, line 106: This will roughly create new channel for every concurrent RPC. Should we have a more reasonable default here? targetLoad of 0 could be only when both `AvgLoadLowThreshold` and `AvgLoadHighThreshold` are zero which seems like a misconfiguration.

bigtable/internal/transport/dynamic_scale_monitor_test.go
```diff
@@ -0,0 +1,358 @@
...
+			for _, entry := range conns {
+				totalLoad += entry.calculateWeightedLoad()
+			}
+			avgLoad := float64(totalLoad) / float64(len(conns))
+			desiredConns := int(math.Ceil(float64(totalLoad) / targetLoadFactor))
+			t.Logf("Initial size: %d, Avg load: %.2f, Total load: %d, Target desired conns: %d", tc.initialSize, avgLoad, totalLoad, desiredConns)
+
+			dynamicMonitor, ok := findMonitor[*DynamicScaleMonitor](pool)
+			if !ok {
+				t.Fatal("Could not find ChannelHealthMonitor in pool")
```
nimf, line 158:
```suggestion
				t.Fatal("Could not find DynamicScaleMonitor in pool")
```

bigtable/internal/transport/dynamic_scale_monitor_test.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+		defer pool.Close()
+
+		// Set load to trigger scale up
+		setConnLoads(pool.getConns(), 15, 0)
+
+		// 1. Simulate recent scaling
+		dynamicMonitor, ok := findMonitor[*DynamicScaleMonitor](pool)
+		if !ok {
+			t.Fatal("Could not find ChannelHealthMonitor in pool")
```
nimf, line 185:
```suggestion
			t.Fatal("Could not find DynamicScaleMonitor in pool")
```

bigtable/internal/transport/connpool.go
```diff
@@ -16,53 +16,344 @@ package internal
...
+// ChannelHealthMonitor manages the overall health checking process for a pool of connections.
+type ChannelHealthMonitor struct {
+	config           btopt.HealthCheckConfig
+	pool             *BigtableChannelPool
+	ticker           *time.Ticker
+	done             chan struct{}
+	stopOnce         sync.Once  // Add sync.Once
+	evictionMu       sync.Mutex // Guards lastEvictionTime
+	lastEvictionTime time.Time
+	evictionDone     chan struct{} // Notification for test
```
nimf, line 255: Do we still use this?

bigtable/internal/transport/connpool.go
```diff
@@ -16,53 +16,344 @@ package internal
...
+	poolCtx    context.Context    // Context for the pool's background tasks
+	poolCancel context.CancelFunc // Function to cancel the poolCtx
+
+	logger *log.Logger // logging events
+
+	// OpenTelemetry MeterProvider for custom metrics
+	meterProvider metric.MeterProvider
+	// OpenTelemetry metric instruments
+	outstandingRPCsHistogram         metric.Float64Histogram
+	perConnectionErrorCountHistogram metric.Float64Histogram
```
nimf, line 281: RPCs and errors are integers. Why do we need this histograms as floats?

nimf (commented):
Comments from the second pass.

bigtable/bigtable.go
```diff
@@ -175,9 +181,31 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
...
+		// btransport is assumed to be the package where BigtableChannelPool resides.
+		connPool, connPoolErr = btransport.NewBigtableChannelPool(
+			ctx,
+			defaultBigtableConnPoolSize,
+			btopt.BigtableLoadBalancingStrategy(),
+			dialFunc,
+			config.Logger,
+			nil,
+			btransport.WithHealthCheckConfig(btopt.DefaultHealthCheckConfig()),
+			btransport.WithDynamicChannelPool(btopt.DefaultDynamicChannelPoolConfig(defaultBigtableConnPoolSize)),
```
mutianf, line 205: nit: is it possible to pass in default channel pool size in one place?

bigtable/internal/option/option.go
```diff
@@ -201,3 +202,103 @@ func EnableBigtableConnectionPool() bool {
...
+	FailurePercentThresh int
+	// PoolwideBadThreshPercent is the "circuit breaker" threshold. If this percentage
+	// of channels in the pool are unhealthy, no evictions will occur.
+	PoolwideBadThreshPercent int
+	// MinEvictionInterval is the minimum time that must pass between eviction of unhealthy channels.
+	MinEvictionInterval time.Duration
+}
+
+// DynamicChannelPoolConfig holds the parameters for dynamic channel pool scaling.
+type DynamicChannelPoolConfig struct {
```
mutianf, line 253: Are we exposing any of these for customers to tune?

bigtable/internal/option/option.go
```diff
@@ -201,3 +202,103 @@ func EnableBigtableConnectionPool() bool {
...
+		MaxRemoveConns:       2, // Cap for removals
+	}
+}
+
+// DefaultHealthCheckConfig for HealthCheckConfig
+func DefaultHealthCheckConfig() HealthCheckConfig {
+	return HealthCheckConfig{
+		Enabled:                  true,
+		ProbeInterval:            30 * time.Second,
+		ProbeTimeout:             1 * time.Second,
```
mutianf, line 283: 1 second seems long. lets make this consistent with Java? the probe deadline is 500 ms

bigtable/internal/option/option.go
```diff
@@ -201,3 +202,103 @@ func EnableBigtableConnectionPool() bool {
...
+		WindowDuration:           5 * time.Minute,
+		MinProbesForEval:         4,
+		FailurePercentThresh:     60,
+		PoolwideBadThreshPercent: 70,
+		MinEvictionInterval:      1 * time.Minute,
```
mutianf, line 288: This seems short, maybe lets make it consistent with java as well which is 10 minutes ?

bigtable/internal/transport/connpool.go
```diff
@@ -16,53 +16,344 @@ package internal
...
+		Name:         bc.instanceName,
+		AppProfileId: bc.appProfile,
+	}
+
+	// Use a timeout for the prime operation
+	primeCtx, cancel := context.WithTimeout(ctx, primeRPCTimeout)
+	defer cancel()
+
+	var p peer.Peer
+	_, err := client.PingAndWarm(primeCtx, req, grpc.Peer(&p))
```
mutianf, line 92: just double checking, this will set the x-goog-request-params header right?

bigtable/internal/transport/connpool.go
```diff
@@ -16,53 +16,344 @@ package internal
...
+// A safety net to prevent a connection from draining indefinitely if a stream hangs.
+var maxDrainingTimeout = 2 * time.Minute
+
+// BigtableChannelPool options
+type BigtableChannelPoolOption func(*BigtableChannelPool)
+
+const (
+	primeRPCTimeout     = 10 * time.Second
+	unaryLoadFactor     = 1
+	streamingLoadFactor = 2
```
mutianf, line 53: why is streaming load factor 2 and unary load factor 1?

bigtable/internal/transport/connpool.go
```diff
@@ -75,30 +366,114 @@ func NewBigtableChannelPool(connPoolSize int, strategy btopt.LoadBalancingStrate
...
+				initialConns[j].conn.Close()
+			}
+			pool.poolCancel() // Ensure context is cancelled
 			return nil, err
 		}
-		pool.conns = append(pool.conns, conn)
-		pool.load = append(pool.load, 0)
-
+		entry := &connEntry{conn: conn}
+		initialConns[i] = entry // TODO prime the connection
```
mutianf, line 393: nit:
```suggestion
		initialConns[i] = entry 
```

bigtable/internal/transport/connpool.go
```diff
@@ -16,53 +16,344 @@ package internal
...
+
 // BigtableChannelPool implements ConnPool and routes requests to the connection
 // pool according to load balancing strategy.
-//
-// To benefit from automatic load tracking, use the Invoke and NewStream methods
-// directly on the BigtableChannelPool instance.
 type BigtableChannelPool struct {
-	conns []*grpc.ClientConn
-	load  []int64 // Tracks active requests per connection
+	conns atomic.Value // Stores []*connEntry
```
nimf, line 262: BTW, can this be `atomic.Pointer[[]*connEntry]` ? Then we won't need to typecast.

bigtable/internal/transport/connpool.go
```diff
@@ -75,30 +366,114 @@ func NewBigtableChannelPool(connPoolSize int, strategy btopt.LoadBalancingStrate
 		pool.selectFunc = pool.selectRoundRobin
 	}
 
+	initialConns := make([]*connEntry, connPoolSize)
 	for i := 0; i < connPoolSize; i++ {
+		select {
+		case <-pool.poolCtx.Done():
```
mutianf, line 372: when will poolctx be done?

bigtable/internal/transport/connpool.go
```diff
@@ -108,123 +483,298 @@ func (p *BigtableChannelPool) Close() error {
...
+		entry := conns[idx]                      // Safe, using snapshot
+		failed := entry.health.getFailedProbes() // getFailedProbes() locks internally
+		if failed > maxFailed {
+			maxFailed = failed
+			worstEntry = entry
+		}
+	}
+
+	if worstEntry != nil {
+		recordEviction() // Record eviction time *before* replacing. // Record eviction time *before* replacing.
```
mutianf, line 555:
```suggestion
		recordEviction() // Record eviction time *before* replacing.
```

bigtable/internal/transport/connpool.go
```diff
@@ -108,123 +483,298 @@ func (p *BigtableChannelPool) Close() error {
...
+			break
+		}
+	}
+
+	// If the connection isn't in the slice, it was already removed.
+	// The drain process should still be kicked off.
+	if idx == -1 {
+		btopt.Debugf(p.logger, "bigtable_connpool: Connection to replace was already removed. Draining it.")
+		// thread safe to call waitForDrainAndClose as conn.Close() can be called multiple times.
+		go p.waitForDrainAndClose(oldEntry)
```
mutianf, line 589: will this ever happen? wouldn't oldEntry.markAsDraining() means it should already be drained by smoething else?

bigtable/internal/transport/connpool.go
```diff
@@ -108,123 +483,298 @@ func (p *BigtableChannelPool) Close() error {
...
+	// Copy-on-write
+	newConns := make([]*connEntry, len(currentConns))
+	copy(newConns, currentConns)
+	newConns[idx] = newEntry
+	p.conns.Store(newConns)
+
+	btopt.Debugf(p.logger, "bigtable_connpool: Replacing connection at index %d\n", idx)
+
+	// Start the graceful shutdown process for the old connection
+	go p.waitForDrainAndClose(oldEntry)
```
mutianf, line 631: This is already done on line 589?

bigtable/internal/transport/connpool.go
```diff
@@ -262,10 +816,135 @@ func (s *refCountedStream) RecvMsg(m interface{}) error {
...
+			break
+		}
+		entry := &connEntry{conn: conn}
+		newEntries[i] = entry
+		go func(e *connEntry) {
+			primeCtx, cancel := context.WithTimeout(p.poolCtx, primeRPCTimeout)
+			defer cancel()
+			isALTS, err := e.conn.Prime(primeCtx)
+			if err != nil {
+				btopt.Debugf(p.logger, "bigtable_connpool: failed to prime new connection: %v\n", err)
```
mutianf, line 864: if the prime request failed does the connection get added to the pool?

mutianf (commented):
only looked at connpool.go and had some questions.

bigtable/otel_metrics.go
```diff
@@ -36,12 +36,41 @@ import (
...
-	unKnownAtttr                        = "unknown"
+	// Bigtable Client Metric Prefix
+	bigtableClientMetricPrefix = "bigtable.googleapis.com/internal/client/"
+	unKnownAtttr               = "unknown"
 )
 
 var latenciesBoundaries = []float64{0.0, 0.001, 0.002, 0.003, 0.004, 0.005, 0.006, 0.008, 0.01, 0.013, 0.016, 0.02, 0.025, 0.03, 0.04, 0.05, 0.065, 0.08, 0.1, 0.13, 0.16, 0.2, 0.25, 0.3, 0.4, 0.5, 0.65, 0.8, 1.0, 2.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0, 400.0, 800.0, 1600.0, 3200.0} // max is 53.3 minutes
 
+// Boundaries for the connection_pool.outstanding_rpcs histogram.
+var outstandingRPCsBoundaries = []float64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 14, 16, 20, 24, 28, 32, 40, 50, 64, 128, 256, 512, 1024}
```
mutianf, line 47: is this the same as the java metric? The buckets should be consistent.

PR #13232, opened 2025-10-27
chore(profiler): migrate module to Librarian

jskeet:


jskeet:
It's slightly unclear to me who would maintain the profiler code at this point given recent restructuring, but it's fully handwritten and the last actual code change I can find is from 2023.

cc @noahdietz who made that last code change...

quartzmo (approved):

PR #13236, opened 2025-10-27
fix(storage): add env var to allow disabling bound token

rockspore:
Inspired by https://github.com/googleapis/google-cloud-go/pull/13153.

The token issuing side is still behind so hopefully it's not too late to add this knob out of extra caution. Users will be able to use it to turn off this flow if anything goes wrong.

@tritone Sorry for not doing this sooner. Would you mind taking a look?

### DIFF
--- a/storage/grpc_client.go
+++ b/storage/grpc_client.go
@@ -103,7 +103,7 @@ func defaultGRPCOptions() []option.ClientOption {
 			internaloption.AllowNonDefaultServiceAccount(true),
 			internaloption.EnableDirectPath(true),
 			internaloption.EnableDirectPathXds())
-		if disableBoundToken, _ := strconv.ParseBool(os.Getenv("GCS_DISABLE_DIRECTPATH_BOUND_TOKEN")); !disableBoundToken {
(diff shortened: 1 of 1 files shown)


storage/grpc_client.go
```diff
@@ -101,8 +102,10 @@ func defaultGRPCOptions() []option.ClientOption {
 		defaults = append(defaults,
 			internaloption.AllowNonDefaultServiceAccount(true),
 			internaloption.EnableDirectPath(true),
-			internaloption.EnableDirectPathXds(),
-			internaloption.AllowHardBoundTokens("ALTS"))
+			internaloption.EnableDirectPathXds())
+		if disableBoundToken, _ := strconv.ParseBool(os.Getenv("GCS_DISABLE_DIRECTPATH_BOUND_TOKEN")); !disableBoundToken {
```
tritone: Maybe `STORAGE` instead of `GCS` in the env var name?

rockspore: Done. Thanks for the quick review!

PR #13239, opened 2025-10-27
fix(storage): Fix takeover response handling.

cjc25:
A takeover response can return a handle. In that case, a retry should present that handle. Otherwise it will attempt a takeover when it retries. After a new takeover, the client can't tell if a partial write after the last known flush offset was performed by itself, or by some other writer. Therefore the client can't tell if it's safe to retry from the last known flush offset, and the server will return an error if it repeats offsets.

The write handle allows the client to present enough information to the server to correctly identify replays on the same write session, avoiding an error when the client repeats bytes that are already durable.

### DIFF
--- a/storage/grpc_writer.go
+++ b/storage/grpc_writer.go
@@ -1169,8 +1169,8 @@ func (s *gRPCAppendBidiWriteBufferSender) handleStream(stream storagepb.Storage_
 
 type gRPCAppendTakeoverBidiWriteBufferSender struct {
 	gRPCAppendBidiWriteBufferSender
-	takeoverReported  bool
-	setTakeoverOffset func(int64)
+	takeoverReported         bool
+	handleTakeoverCompletion func(gRPCBidiWriteCompletion)
 }
 
 func writeObjectSpecAsAppendObjectSpec(s *storagepb.WriteObjectSpec, gen int64) *storagepb.AppendObjectSpec {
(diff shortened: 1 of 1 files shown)


cjc25:
The retry conformance tests will start to pass once the testbench supports single-threaded takeovers, in v0.56.0.

The MRD timeout is a preexisting flake, I think.

tritone (dismissed):
This looks good to me.

Does the intersection of the way the write is split up and the retry test scenarios cover a retry both on the initial takeover req and midway through the takeover write? Might be good to note that explicitly as far as how you chose how much data to write when.

cjc25:
The test found a real bug (yay!)

On rare occasions on my workstation, the takeover writer issues its first write at an earlier offset than the one returned by the server. The real service treats this as an error, but we don't model the full state machine in the testbench and it falls into retry handling, which silently drops these bytes.

This probably means that there is a race in the spot where we determine the takeover offset for future writes. I'll figure that out before merging this PR.

cjc25:
OK the bug is that we don't guarantee that the writer has handled the completion from the takeover response fully before returning control to the caller code.

Caller gets control back when `setTakeoverOffset` is called here: https://github.com/googleapis/google-cloud-go/pull/13239/files#diff-2b5c89d316046e2b33163778a826af79ec60ff23d70a93cd0d4fa1daa2a494e8R1242.

The writer has a fully up-to-date view of the correct buffer offsets when the completion put on the channel during takeover (https://github.com/googleapis/google-cloud-go/pull/13239/files#diff-2b5c89d316046e2b33163778a826af79ec60ff23d70a93cd0d4fa1daa2a494e8R1244) is processed by the writeLoop (https://github.com/googleapis/google-cloud-go/pull/13239/files#diff-2b5c89d316046e2b33163778a826af79ec60ff23d70a93cd0d4fa1daa2a494e8R447-R451)

We should guarantee that we've handled this completion before returning control to the caller. It's a small fix, I think, so I'll just include it in this PR.

PR #13241, opened 2025-10-28
test(storage): fix write emulator test

tritone:
This started breaking with the new testbench release and I'm not totally sure how it was passing before. Fix the way that bytes are read and checked in the test.

### DIFF
--- a/storage/client_test.go
+++ b/storage/client_test.go
@@ -2931,13 +2931,15 @@ func TestWriterChunkTransferTimeoutEmulated(t *testing.T) {
 						t.Fatalf("opening reading: %v", err)
 					}
 					wantLen := len(buffer)
-					got := make([]byte, wantLen)
-					n, err := r.Read(got)
-					if n != wantLen {
-						t.Fatalf("expected to read %d bytes, but got %d", wantLen, n)
+					got, err := io.ReadAll(r)
+					if err != nil {
+						t.Fatalf("reading bytes: %v", err)
 					}
-					if diff := cmp.Diff(got, buffer); diff != "" {
-						t.Fatalf("checking written content: got(-),want(+):\n%s", diff)
+					if len(got) != wantLen {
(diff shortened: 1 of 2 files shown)


storage/client_test.go
```diff
@@ -2931,13 +2931,15 @@ func TestWriterChunkTransferTimeoutEmulated(t *testing.T) {
...
+						t.Fatalf("reading bytes: %v", err)
+					}
+					if len(got) != wantLen {
+						t.Fatalf("expected to read %d bytes, but got %d", wantLen, len(got))
 					}
-					if diff := cmp.Diff(got, buffer); diff != "" {
-						t.Fatalf("checking written content: got(-),want(+):\n%s", diff)
+					if !bytes.Equal(got, buffer) {
+						t.Errorf("checking content: bytes did not match")
 					}
```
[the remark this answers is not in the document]
tritone: cmp.Diff is slow for checking large byte arrays, so we have been removing it from tests which compare object contents.

storage/client_test.go
```diff
@@ -2931,13 +2931,15 @@ func TestWriterChunkTransferTimeoutEmulated(t *testing.T) {
 						t.Fatalf("opening reading: %v", err)
 					}
 					wantLen := len(buffer)
-					got := make([]byte, wantLen)
-					n, err := r.Read(got)
-					if n != wantLen {
-						t.Fatalf("expected to read %d bytes, but got %d", wantLen, n)
+					got, err := io.ReadAll(r)
```
BrennaEpp, line 2934: Why did the previous not work? We should just have been overwritting the bytes in `got`, which should be equal to `n`.

tritone: There is no guarantee that Read fills the buffer before returning. You would need to use Readall or Copy or similar to actually ensure all the data from the reader is copied. I assume something changed in the emulator or in net/http which triggered the earlier return path.

PR #13242, opened 2025-10-28
fix(internal/librariangen): modify logging levels

quartzmo:
Add unit tests for postprocessor new library path.

### DIFF
--- a/internal/librariangen/configure/configure.go
+++ b/internal/librariangen/configure/configure.go
@@ -96,7 +96,7 @@ func Configure(ctx context.Context, cfg *Config) error {
 	if err := cfg.Validate(); err != nil {
 		return fmt.Errorf("librariangen: invalid configuration: %w", err)
 	}
-	slog.Info("librariangen: configure command started")
+	slog.Debug("librariangen: configure command started")
 	configureReq, err := readConfigureReq(cfg.LibrarianDir)
 	if err != nil {
 		return fmt.Errorf("librariangen: failed to read request: %w", err)
@@ -320,7 +320,9 @@ func goModEditReplaceInSnippets(ctx context.Context, cfg *Config, modulePath, re
 		return err
 	}
 	copyRepoFileToOutput(cfg, "internal/generated/snippets/go.mod")
-	args := []string{"go", "mod", "edit", "-replace", fmt.Sprintf("%s=%s", modulePath, relativeDir)}
+	replaceStr := fmt.Sprintf("%s=%s", modulePath, relativeDir)
+	args := []string{"go", "mod", "edit", "-replace", replaceStr}
+	slog.Info("librariangen: running go mod edit -replace", "replace", replaceStr, "directory", outputSnippetsDir)
 	return execvRun(ctx, args, outputSnippetsDir)
 }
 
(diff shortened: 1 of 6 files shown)


internal/librariangen/configure/configure.go
```diff
@@ -320,7 +320,9 @@ func goModEditReplaceInSnippets(ctx context.Context, cfg *Config, modulePath, re
 		return err
 	}
 	copyRepoFileToOutput(cfg, "internal/generated/snippets/go.mod")
-	args := []string{"go", "mod", "edit", "-replace", fmt.Sprintf("%s=%s", modulePath, relativeDir)}
+	replaceStr := fmt.Sprintf("%s=%s", modulePath, relativeDir)
+	args := []string{"go", "mod", "edit", "-replace", replaceStr}
+	slog.Info("librariangen: running go mod edit -replace %s", replaceStr, "directory", outputSnippetsDir)
```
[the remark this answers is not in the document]
quartzmo: Fixed, thanks!

internal/librariangen/postprocessor/postprocessor.go
```diff
@@ -82,7 +82,7 @@ func PostProcess(ctx context.Context, req *request.Library, outputDir, moduleDir
 // goimports runs the goimports tool on a directory to format Go files and
 // manage imports.
 func goimports(ctx context.Context, dir string) error {
-	slog.Debug("librariangen: running goimports", "directory", dir)
+	slog.Info("librariangen: running goimports", "directory", dir)
```
jskeet, line 86: Not sure *this* one needs to be info - but equally, I'm not overly bothered.

internal/librariangen/postprocessor/postprocessor_test.go
```diff
@@ -55,9 +57,40 @@ func TestPostProcess(t *testing.T) {
...
+				}
+				return nil
+			},
+			wantErr:      true,
+			singleNewAPI: true,
+		},
 	}
 
 	for _, tt := range tests {
+		tt := tt
```
jskeet, line 120: I'm not disputing whether this is needed or not, but I don't understand it - if you could explain it when we next meet, that would be great. (I'm guessing it's so that the anonymous function for `execvRun` captures that iteration's `tt` rather than the iteration variable itself?)

internal/librariangen/postprocessor/postprocessor_test.go
```diff
@@ -122,17 +173,47 @@ func TestPostProcess(t *testing.T) {
...
+				}
+			}
+
 			if tt.wantErr {
 				return
 			}
 
+			// Determine which files should have been modified based on the request.
+			wantModifiedFiles := make(map[string]bool)
+			for _, api := range req.APIs {
```
jskeet: I'm *tempted* to suggest we should just have these listed in the test case, to avoid having logic that's effectively duplicating the production code here. I can see it would be tedious though. Hmm.

internal/librariangen/postprocessor/postprocessor.go
quartzmo: keeping it

internal/librariangen/postprocessor/postprocessor_test.go
quartzmo: Fixed, thanks!

jskeet (approved):

PR #13247, opened 2025-10-28
chore(bigquery): migrate module to Librarian

jskeet:
Fixes https://github.com/googleapis/librarian/issues/2426

### DIFF
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -848,8 +848,8 @@ libraries:
       - bigtable/testdata
     tag_format: '{id}/v{version}'
   - id: bigquery
-    version: 1.71.0
-    last_generated_commit: 94ccdfe4519e0ba817bd33aa22eb9c64f88a6874
+    version: 1.72.0
+    last_generated_commit: 97763d6efb2d35474e0363c254ee6994837a08fa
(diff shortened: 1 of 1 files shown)


jskeet:
This should be reviewed carefully - I hadn't spotted the v1/v2 split before. This is expressed in (nearly) the same way as `pubsub` - there are two modules, `bigquery` and `bigquery/v2`. Unlike `pubsub` however, the `bigquery` module is the one with almost all the generated code at the moment, because `v2` is still under development.

Running `librarian generate` with the *current* container will add some imports into the v2 snippets. This is fixed in https://github.com/googleapis/google-cloud-go/pull/13248.

.librarian/state.yaml
```diff
@@ -847,6 +847,258 @@ libraries:
       - bigtable/internal/conformance
       - bigtable/testdata
     tag_format: '{id}/v{version}'
+  - id: bigquery
+    version: 1.71.0
```
quartzmo: Should this match the current value in `.release-please-manifest-individual.json`: `1.72.0`?

jskeet: Yes. Fixing...

PR #13250, opened 2025-10-28
chore(errorreporting): migrate to Librarian

jskeet:


jskeet:
Note: once this is merged, we'll get a generation commit to "catch up" to the current API definitions. The result will build and the tests still pass - at least the unit tests.

quartzmo (approved):

PR #13265, opened 2025-10-29
fix(spanner): add env var to allow disabling directpath bound token

rockspore:
Inspired by https://github.com/googleapis/google-cloud-go/pull/13153.

The token issuing side is still behind so hopefully it's not too late to add this knob out of extra caution. Users will be able to use it to turn off this flow if anything goes wrong.


rockspore:
@rahul2393 Could you PTAL? Thank you!

sakthivelmanii (approved):

PR #13269, opened 2025-10-30
fix(bigquery): make additional errors retriable: tcp timeout and http2 client connection lost

MartinSahlen:
#### Description
The [cloud.google.com/go/bigquery](http://cloud.google.com/go/bigquery) client does not automatically retry API calls that fail with a dial tcp: i/o timeout error. This type of error is a common transient network failure, especially in distributed cloud environments, and often occurs when initiating a connection.

The underlying Go error wrapper correctly identifies this as a retryable error (as seen by retryable: true in the error message), but the BigQuery client's internal retry predicate fails to catch it, immediately propagating the error to the user. This forces developers to build their own complex retry wrappers around the client, which should ideally be handled by the library's built-in resilience mechanisms.

#### Expected Behavior
When an API call (such as `jobs.insert` or `jobs.query`) fails with a `dial tcp: i/o timeout`, the client library should recognize this as a transient, retryable error and automatically retry the operation using its built-in exponential backoff strategy.

#### Actual Behavior
The API call fails immediately and returns the i/o timeout error directly to the caller. No retry is attempted by the library.
The full error message is similar to the following:
```
update table from struct: Post \"https://.../bigquery/v2/projects/.../jobs?alt=json&prettyPrint=false&uploadType=multipart\": dial tcp 34.50.146.6:443: i/o timeout (type: wrapError, retryable: true): Post \"https://.../bigquery/v2/projects/.../jobs?alt=json&prettyPrint=false&uploadType=multipart\": dial tcp 34.50.146.6:443: i/o timeout
```

#### Code Snippet
The issue can be observed with any standard API call that initiates a network request. For example, when using a Loader to start a job:
```go
package main

import (
	"context"
	"log"

)

func main() {
	ctx := context.Background()
	projectID := "your-project-id"

	client, err := bigquery.NewClient(ctx, projectID)
	if err!= nil {
		log.Fatalf("bigquery.NewClient: %v", err)
	}
	defer client.Close()

	// Assume 'gcsRef' is a *bigquery.GCSReference pointing to a file.
	// This call to Run() initiates a jobs.insert API call.
	loader := client.Dataset("my_dataset").Table("my_table").LoaderFrom(gcsRef)
	
	job, err := loader.Run(ctx)
	if err!= nil {
		// When a "dial tcp: i/o timeout" occurs, the error is returned here
		// immediately without any retry attempts from the library.
		log.Fatalf("Failed to start load job: %v", err)
	}

	//... wait for job completion
}
```

#### Additional Context & Analysis
The root cause appears to be in the library's internal `retryableError` predicate. This function does not check for errors that satisfy the `net.Error Timeout()` method.
The current implementation checks for `interface{ Temporary() bool }`:
```go
//...
	case interface{ Temporary() bool }:
		if e.Temporary() {
			return true
		}
//...
```

However, a `dial tcp: i/o timeout` is a `net.Error` where `Timeout()` returns `true`, but `Temporary()` may not. The `Temporary()` method was deprecated in Go 1.18 because its definition was ambiguous and ill-defined. Most errors that were once "temporary" are now more accurately classified as timeouts.

Because the library's predicate relies on this deprecated method and omits a check for the `Timeout()` method, it fails to identify one of the most common types of transient network errors.

The proposed fix in this PR is to update the `retryableError` predicate to also include a check for timeout errors, for example:
```go
//...
	case interface{ Timeout() bool }:
		if e.Timeout() {
			return true
		}
	case interface{ Temporary() bool }:
//...
```
Adding this case will improve the client's resilience and align its behavior with the expectation that transient network timeouts are handled automatically.

Hoping for positive feedback on this one and that we can get it merged quickly. Cheers!

### DIFF
--- a/bigquery/bigquery.go
+++ b/bigquery/bigquery.go
@@ -268,6 +268,9 @@ func retryableError(err error, allowedReasons []string) bool {
 	if err.Error() == "http2: stream closed" {
 		return true
 	}
+	if err.Error() == "http2: client connection lost" {
+		return true
+	}
 
 	switch e := err.(type) {
 	case *googleapi.Error:
--- a/bigquery/bigquery_test.go
+++ b/bigquery/bigquery_test.go
@@ -17,6 +17,7 @@ package bigquery
 import (
 	"errors"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"testing"
@@ -41,6 +42,20 @@ func TestRetryableErrors(t *testing.T) {
 			description: "http stream closed",
 			in:          errors.New("http2: stream closed"),
 			wantRetry:   true,
+		},
+		{
+			description: "http client connection lost",
+			in:          errors.New("http2: client connection lost"),
+			wantRetry:   true,
+		},
+		{
+			description: "tcp timeout error",
+			in: &net.DNSError{
+				Err: "timeout",
+
+				IsTimeout: true,
+			},
+			wantRetry: true,
 		},
 		{
 			description: "io ErrUnexpectedEOF",


alvarowolfx:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/googleapis/google-cloud-go/pull/13269/checks?check_run_id=54094289975) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

@MartinSahlen can you sign the CLA ?

MartinSahlen:
> > Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> > 
> 
> > View this [failed invocation](https://github.com/googleapis/google-cloud-go/pull/13269/checks?check_run_id=54094289975) of the CLA check for more information.
> 
> > 
> 
> > For the most up to date status, view the checks section at the bottom of the pull request.
> 
> 
> 
> @MartinSahlen can you sign the CLA ? 

Yes I did shortly after submitting the PR and also retriggered the test which passed already.

shollyman:
Thanks for the PR and the detailed writeup!

One minor request:  Could you add a testcase for this to TestRetryableErrors in bigquery_test.go?  It looks like https://pkg.go.dev/net#DNSError gives us an easy to use example error for this case.

MartinSahlen:
> Thanks for the PR and the detailed writeup!
> 
> 
> 
> One minor request:  Could you add a testcase for this to TestRetryableErrors in bigquery_test.go?  It looks like https://pkg.go.dev/net#DNSError gives us an easy to use example error for this case.

Thanks! I can give it a go. Firstly I need to understand the test structure and how to set up some mock errors/responses, will give a shout if I get stuck.

MartinSahlen:
Hi @shollyman , so I added a test. However in the meantime we have observed one more error "http2: client connection lost", which I decided to also add to this PR with a corresponding test. Perhaps the PR title should change to reflect this when squash-merging.

MartinSahlen:
hi @shollyman and @alvarowolfx, any next steps here? Or any idea of a timeline on your end?

joshk0:
By the way, the storage client handles it through an extension point that gives the user the ability to supply a custom method to determine whether a given error is retryable, in addition to the built-in logic.

https://github.com/googleapis/google-cloud-go/blob/storage/v1.57.1/storage/storage.go#L2528

I'm not sure what the overall SDK strategy is, but for bigquery, it might be nice to add some of these errors we have seen in the wild (as this change accomplishes), as well as future proofing with an extension point that lets users easily add their own retry cases.

shollyman:
Apologies, I've been OOO and playing catchup.  Taking another look now.

quartzmo (approved):

shollyman:
Thanks again for the contribution!

MartinSahlen:
> Thanks again for the contribution!

No worries! That being said, I think @joshk0's suggestion might be one to consider. We don't see these errors in the python clients, most likely because it represents 95% (or more) of users' interaction with BigQuery APIs, and such it has more robust error handling. Until we see "everything" and can make all errors retryable, having some ability for users to manually tell the library what should be retriable seems like a good stop-gap.

PR #13278, opened 2025-10-31
fix(storage): Handle StartResumableWrite errors correctly.

cjc25:
When StartResumableWrite errors, we were setting an interface to a nil with the implementation type. This uses an intermediate variable, and only assigns to the interface if there was no error.


cjc25:
This probably requires a new release branch (v1.56.4?)

tritone:
Looks good to me but I think you'll have to re-open the PR against a new branch in origin. Do you  have permission to create one? If not I can do it.

cjc25:
I'm going to abandon this PR since we now are up to 1.59.1

PR #13286, opened 2025-11-01
fix(bigtable): fix project id sent in otel

mathewhany:
Project ID sent to Google Monitoring was different than the project ID sent in the labels, which was causing an error:

```
rpc error: code = InvalidArgument desc = Field resource.labels.project_id had an invalid value of "*": if present, must be the project number or ID in the request name (projects/*).
```

This happened when you run your app on a project but connect to Bigtable from a different project. The project in the monitoring request name would be the project that you are running on, and the project in the labels would be the Bigtable project, causing the mismatch.

### DIFF
--- a/bigtable/otel_metrics.go
+++ b/bigtable/otel_metrics.go
@@ -59,7 +59,7 @@ type bigtableClientMonitoredResource struct {
 
 func (bmr *bigtableClientMonitoredResource) exporter() (metric.Exporter, error) {
 	exporter, err := mexporter.New(
-		mexporter.WithProjectID(bmr.clientProject),
+		mexporter.WithProjectID(bmr.project),
 		mexporter.WithMetricDescriptorTypeFormatter(metricFormatter),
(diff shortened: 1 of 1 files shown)


bhshkh (approved):

mathewhany:
@bhshkh Any idea for when this is going to be released? We are seeing a lot of noise in our logs because of it

sushanb:
@mathewhany  we released 1.41.0 which has the fix.

FYI,  we would like you to try Direct Access on golang. https://docs.google.com/document/d/1Auuo-cSTDExkOoQS29ac7euLCM5vK2w8lRdKpbbn6GA/edit?pli=1&tab=t.0

If you are running in GKE/GCE with no special firewall, setting an env var CBT_ENABLE_DIRECTPATH=true should do it. More details on doc. Happy to answer question if necessary.

I guess there must be low activity in December due to freezes. We can work with you in Jan to enable this feature.

PR #13298, opened 2025-11-03
chore(.github): conditionalize apidiff failures for GA APIs

quartzmo:
* Continue to label all breaking changes.
* Fail only for GA breaking changes.
* Post comment to PR for non-GA breaking changes.

closes: googleapis/librarian#2757

### DIFF
--- a/.github/workflows/apidiff.yml
+++ b/.github/workflows/apidiff.yml
@@ -107,6 +107,7 @@ jobs:
     # For non-GA versions (e.g., "apiv1beta1"), this step is skipped, and the
     # job continues.
     - name: Conditionally fail for GA APIs
+      id: ga_check # Add an ID to this step to check its outcome later.
       if: steps.detect.outputs.breaking_change_found == 'true'
       run: |
         if [[ "${{ matrix.changed }}" =~ apiv[0-9]+$ ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/apidiff.yml
```diff
@@ -78,19 +79,48 @@ jobs:
...
+    # This step runs only if a breaking change was found in the previous step.
+    # It checks if the changed module is a GA version (e.g., "apiv1", "apiv2").
+    # If it is, the `exit 1` command fails the job, blocking the PR.
+    # For non-GA versions (e.g., "apiv1beta1"), this step is skipped, and the
+    # job continues.
+    - name: Conditionally fail for GA APIs
+      id: ga_check # Add an ID to this step to check its outcome later.
+      if: steps.detect.outputs.breaking_change_found == 'true'
+      run: |
+        if [[ "${{ matrix.changed }}" =~ apiv[0-9]+$ ]]; then
```
jskeet, line 113: So this is about whether it's a GA API version (e.g. v1), rather than about whether the library is GA. These days APIs often publish beta API specifications under v1 before toggling it to GA in the service config.

Basically this might be a little more confident in something being GA than is really correct, but it's *at least* better than what we have today, so will reduce toil. In the future we could *potentially* parse the repo-metadata.json file to see if the *library* is GA.

.github/workflows/apidiff.yml
```diff
@@ -100,3 +130,39 @@ jobs:
             repo: context.repo.repo,
             labels: ['breaking change']
           })
+    # Step 4: For non-GA APIs, post the breaking change details as a comment.
+    # This step only runs if the ga_check step succeeded, meaning a breaking
+    # change was found in a non-GA API. This provides high-visibility feedback
+    # without blocking the PR, and avoids redundant comments for GA APIs where
+    # the failed presubmit check is the primary signal.
+    - name: Post breaking change details as a comment
+      if: ${{ steps.ga_check.outcome == 'success' && !github.event.pull_request.head.repo.fork }}
```
jskeet, line 139: Doesn't this need `steps.detect.outputs.breaking_change_found == 'true'` as well? Othewise I'd expect it to *always* claim that there's a breaking change, even if there isn't.

quartzmo: For `steps.ga_check.outcome` to be `success`, the `ga_check` step must have run and completed successfully, meaning it did not detect GA and did not exit with `1`, but it did run. As explained in point 2, the `ga_check` step only runs if `steps.detect.outputs.breaking_change_found` was `true`.

.github/workflows/apidiff.yml
quartzmo: Makes sense to me. Is there anything I can do within this PR (or outside of it) to address this? Add a comment or issue?

.github/workflows/apidiff.yml
jskeet: Ah right. Sure, that makes sense. I think - not a GHA expert though!

.github/workflows/apidiff.yml
jskeet: I think it would be worth filing an issue, and we can figure it out later.

quartzmo: Actually, since the consequence of a false positive is just human investigation of a change to an API with GA version (e.g. v1), I think it's good as-is.

PR #13308, opened 2025-11-04
feat: Support scopes field from impersonated credential json

sai-sunder-s:
If impersonated json contains scopes field and the user does not provide scopes through options, then use the scopes field value for the impersonated token.

Hard code the scopes for source credential to be always `cloud-platform`. Unless we do this, if user provides a scope other than cloud-platform or IAM, then the call to iamcredentials will fail. (This is an existing bug in the code)

Sample JSON:
```
{
  "delegates": [],
  "scopes": [
    "https://www.googleapis.com/auth/drive"
  ],
  "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/service-account-email@project-name.iam.gserviceaccount.com:generateAccessToken",
  "source_credentials": {
    "client_id": "oauth_client_id",
    "client_secret": "oauth_client_secret",
    "refresh_token": "user_refresh_token",
    "type": "authorized_user",
    "universe_domain": "googleapis.com"
  },
  "type": "impersonated_service_account"
}
```

### DIFF
--- a/auth/credentials/detect_test.go
+++ b/auth/credentials/detect_test.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


hongalex (approved):

quartzmo (changes requested):
LGTM, except for the changes requested by GCA. Thanks! 🥇

quartzmo (approved):

PR #13320, opened 2025-11-06
feat(storage): add support for partial success in ListBuckets

cpriti-os:
This change adds support for the returnPartialSuccess parameter in ListBuckets calls for both  JSON and gRPC client.

When this option is enabled by setting ReturnPartialSuccess = true on the BucketIterator, the GCS API will return a list of reachable buckets even if some are temporarily unavailable. The resource names of any unreachable buckets are returned in a new Unreachable field in the ListBucketsResponse 

Integration and unit tests added.


storage/integration_test.go
```diff
@@ -6305,6 +6305,52 @@ func TestIntegration_ReaderCancel(t *testing.T) {
...
+		t.Cleanup(func() { h.mustDeleteBucket(bkt) })
+
+		// Loop through the boolean cases to test both states of the new feature.
+		for _, partialSuccess := range []bool{true, false} {
+			t.Run(fmt.Sprintf("partialSuccess=%v", partialSuccess), func(t *testing.T) {
+				it := client.Buckets(ctx, projectID)
+				// Set only if true, as the default is false.
+				if partialSuccess {
+					it.ReturnPartialSuccess = partialSuccess
+				}
```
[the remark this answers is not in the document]
cpriti-os: Done

storage/http_client.go
```diff
@@ -242,6 +243,7 @@ func (c *httpStorageClient) ListBuckets(ctx context.Context, project string, opt
 			}
 			it.buckets = append(it.buckets, b)
 		}
+		it.unreachable = resp.Unreachable
```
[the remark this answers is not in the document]
cpriti-os: Expected behavior, need to propagate unreachable list from the server as is.

storage/bucket.go
```diff
@@ -2343,12 +2343,21 @@ func (c *Client) Buckets(ctx context.Context, projectID string) *BucketIterator
 type BucketIterator struct {
 	// Prefix restricts the iterator to buckets whose names begin with it.
 	Prefix string
+	// If true, the response will contain a list of unreachable buckets if the buckets are unavailable.
```
krishnamd-jkp: nit: can mention to use it.Unreachable() to get the unreachable buckets list

storage/grpc_client.go
```diff
@@ -262,6 +263,9 @@ func (c *grpcStorageClient) ListBuckets(ctx context.Context, project string, opt
 			it.buckets = append(it.buckets, b)
 		}
 
+		if resp, ok := gitr.Response.(*storagepb.ListBucketsResponse); ok {
+			it.unreachable = resp.Unreachable
```
krishnamd-jkp: the unreachable field will only contain the list of buckets in this page? If yes, we'll have to keep appending it to a global response like we do it for the buckets.

storage/http_client.go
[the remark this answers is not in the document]
cpriti-os: To add more context here: the server will keep sending the buckets that are unreachable in every iteration and can also remove the buckets that are reachable now from the unreachable list. Appending will result in a lot of duplicates and might also be inaccurate.

storage/grpc_client.go
```diff
@@ -262,6 +263,9 @@ func (c *grpcStorageClient) ListBuckets(ctx context.Context, project string, opt
 			it.buckets = append(it.buckets, b)
 		}
 
+		if resp, ok := gitr.Response.(*storagepb.ListBucketsResponse); ok {
```
krishnamd-jkp, line 266: Should we return an error if the response is not cast-able? User would not know if the response is malformed.

cpriti-os: A type assertion failure here would indicate an unexpected bug in the gapic layer of the client library, rather than a malformed response from the server. The pattern I have observed in such cases  is to not introduce error handling for these kinds of 'should-never-happen' internal mismatches, as it would add noise for a non-recoverable situation. This is more of a safeguard, not a failure check .

storage/bucket.go
cpriti-os: Done

krishnamd-jkp:
Seems like the build is failing because the underlying apiary is not updated yet with new fields. Are you planning to update it?

storage/bucket.go
```diff
@@ -2343,12 +2343,22 @@ func (c *Client) Buckets(ctx context.Context, projectID string) *BucketIterator
...
 	// Prefix restricts the iterator to buckets whose names begin with it.
 	Prefix string
-
-	ctx       context.Context
-	projectID string
-	buckets   []*BucketAttrs
-	pageInfo  *iterator.PageInfo
-	nextFunc  func() error
+	// If true, the response will contain a list of unreachable buckets if the buckets are unavailable.
+	// Use the Unreachable() method to retrieve the list.
```
tritone: What is the expected behavior if Unreachable=false and there are buckets that can't be reached?

Might be helpful to be clearer about in what circumstances someone would want to use this option (or link to the docs page if there is one)? And/or add an example for how to iterate buckets and then check for unreachable afterwards.

I would also mention Unreachable under `Client.Buckets()` docs as well (it.Prefix is also mentioned there)

storage/integration_test.go
```diff
@@ -6305,6 +6305,48 @@ func TestIntegration_ReaderCancel(t *testing.T) {
...
+
+		newBucketName := prefix + uidSpace.New()
+		bkt := client.Bucket(newBucketName)
+
+		h.mustCreate(bkt, projectID, nil)
+		t.Cleanup(func() { h.mustDeleteBucket(bkt) })
+
+		for _, partialSuccess := range []bool{true, false} {
+			t.Run(fmt.Sprintf("partialSuccess=%v", partialSuccess), func(t *testing.T) {
+				it := client.Buckets(ctx, projectID)
```
tritone, line 6329: This might take a long time to run if there are a lot of buckets in the project. Maybe use Prefix to filter the response to just the one you want?

storage/bucket.go
cpriti-os: Added comments to explain the behavior, above the field and function. Public documentation is not yet updated and regarding adding an example: isn't samples the right place to add it or is it recommended to add it in the code docs as well? Sample is on the way and will raise as soon as this change is out.

storage/integration_test.go
cpriti-os: Agreed, done.

tritone (approved):

storage/bucket.go
tritone: Thanks for the doc updates, that definitely is clearer now.

PR #13290, opened 2025-11-03
chore(internal/librariangen): add .repo-metadata.json generation

quartzmo:
closes: https://github.com/googleapis/librarian/issues/2527

PR #13322, opened 2025-11-06
chore(internal/librariangen): change release-init to release-stage

jskeet:
This makes the Go container comply with the Librarian changes in https://github.com/googleapis/librarian/commit/e6d911da1cf7bbabe476f34ab5150a0011c46e44

Fixes https://github.com/googleapis/librarian/issues/2785


jskeet:
Set to "do not merge" because we probably want to merge *and start using* #13290 quickly, but we need to coordinate using *this* change with automation.

PR #13324, opened 2025-11-06
chore(spanner): Log Spanner Client Configuration

sakthivelmanii:


spanner/client.go
```diff
@@ -614,6 +614,27 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
...
+LAR enabled: %v
+Directpath enabled: %v
+End To End Tracing enabled: %v
+Built-in metrics enabled: %v
+gRPC metrics enabled: %v
+Min Sessions: %v
+Max Sessions: %v
+Multiplex session enabled: %v
+Multiplex session enabled for RW: %v
+Multiplex session enabled for Partition Ops: %v
```
olavloite: Are these still relevant?

spanner/client.go
```diff
@@ -614,6 +614,27 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
...
+Project ID: %v
+Num of gRPC channels: %v
+LAR enabled: %v
+Directpath enabled: %v
+End To End Tracing enabled: %v
+Built-in metrics enabled: %v
+gRPC metrics enabled: %v
+Min Sessions: %v
+Max Sessions: %v
+Multiplex session enabled: %v
```
olavloite: super-nit:

```suggestion
Multiplexed session enabled: %v
```

spanner/client.go
sakthivelmanii: We can keep it till we remove the flags

spanner/client.go
sakthivelmanii: Done

spanner/integration_test.go
```diff
@@ -420,6 +420,7 @@ const (
 )
 
 func TestMain(m *testing.M) {
+	//_ = os.Setenv("GOOGLE_CLOUD_SPANNER_DISABLE_CLIENT_OPTIONS_LOGS", "true")
```
olavloite: Should this be commented out?

spanner/client.go
```diff
@@ -1504,3 +1526,13 @@ func parseServerTimingHeader(md metadata.MD) map[string]time.Duration {
...
+
+// enableClientOptionsLog returns true if env doesn't exist or false
+func enableClientOptionsLog() bool {
+	if disableLogString, found := os.LookupEnv("GOOGLE_CLOUD_SPANNER_DISABLE_CLIENT_OPTIONS_LOGS"); found {
+		if disableLog, err := strconv.ParseBool(disableLogString); err == nil {
+			return !disableLog
+		}
+	}
+	return true
+}
```
olavloite: nit, can we rename these a bit:


```suggestion
// enableLogClientOptions returns true if the environment variable for this doesn't exist or is set to false
func enableLogClientOptions() bool {
	if disableLogString, found := os.LookupEnv("GOOGLE_CLOUD_SPANNER_DISABLE_LOG_CLIENT_OPTIONS"); found {
		if disableLog, err := strconv.ParseBool(disableLogString); err == nil {
			return !disableLog
		}
	}
	return true
}
```

olavloite (approved):

spanner/integration_test.go
sakthivelmanii: fixed.

spanner/client.go
```diff
@@ -1504,3 +1526,13 @@ func parseServerTimingHeader(md metadata.MD) map[string]time.Duration {
 	}
 	return metrics
 }
+
+// enableLogClientOptions returns true if the environment variable for this doesn't exist or is set to false
+func enableLogClientOptions() bool {
```
egonelbre, line 1531: Is there a reason this was enabled by default? This information is usually provided by the user explicitly and hence they should already know it. I usually am of the mindset that libraries should be silent until something extraordinary happens.

Also, at the moment there's no way to configure this per client.

PR #13338, opened 2025-11-10
refactor(firestore): [PQ] Rename Expr to Expression

bhshkh:
Rename Expr to Expression to match other SDKs (Java, Python, etc)

Previous pull requests

- https://github.com/googleapis/google-cloud-go/pull/12217
- https://github.com/googleapis/google-cloud-go/pull/12425
- https://github.com/googleapis/google-cloud-go/pull/12538
- https://github.com/googleapis/google-cloud-go/pull/13147
- https://github.com/googleapis/google-cloud-go/pull/13199
- https://github.com/googleapis/google-cloud-go/pull/13218
- https://github.com/googleapis/google-cloud-go/pull/13194
- https://github.com/googleapis/google-cloud-go/pull/13245
- https://github.com/googleapis/google-cloud-go/pull/13270
- https://github.com/googleapis/google-cloud-go/pull/13271
- https://github.com/googleapis/google-cloud-go/pull/13279
- https://github.com/googleapis/google-cloud-go/pull/13280
- https://github.com/googleapis/google-cloud-go/pull/13281
- https://github.com/googleapis/google-cloud-go/pull/13282
- https://github.com/googleapis/gapic-generator-go/pull/1661
- https://github.com/googleapis/google-cloud-go/pull/13283
- https://github.com/googleapis/google-cloud-go/pull/13274

PR #13280, opened 2025-10-31
feat(firestore): [PQ] add raw stage

bhshkh:
Add raw stage similar to Java
https://github.com/googleapis/java-firestore/blob/742fab6583c9a6f9c47cf0496124c3c9b05fe0ee/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Pipeline.java#L997-L1021

The raw stage is an escape hatch to allow customers to consume new stages supported by the backend without having to update their SDK to a version that adds the stage.

Previous pull requests

- https://github.com/googleapis/google-cloud-go/pull/12217
- https://github.com/googleapis/google-cloud-go/pull/12425
- https://github.com/googleapis/google-cloud-go/pull/12538
- https://github.com/googleapis/google-cloud-go/pull/13147
- https://github.com/googleapis/google-cloud-go/pull/13199
- https://github.com/googleapis/google-cloud-go/pull/13218
- https://github.com/googleapis/google-cloud-go/pull/13194
- https://github.com/googleapis/google-cloud-go/pull/13245
- https://github.com/googleapis/google-cloud-go/pull/13270
- https://github.com/googleapis/google-cloud-go/pull/13271
- https://github.com/googleapis/google-cloud-go/pull/13279

PR #13218, opened 2025-10-23
feat(firestore): [PQ] add all the remaining private preview stages

bhshkh:
b/364927702

1. add all the remaining private preview stages. Merging this PR completes the implementation of all the type "Stage" subType "General" private preview features. (except literals stage which is not yet inmplemented in Java and Node. Requires additonal approvals from Firestore team and will be added to separate PR).
See "Firestore Features (Pipeline)" sheet in [go/firestore-query-tracker](http://go/firestore-query-tracker)
2. Add integration and unit tests.
3. Refactor existing stages code to remove duplicated code and rearrange in alphabetical order.
4. Modify behaviour of Data and DataTo to match existing implementation in document.go
https://github.com/googleapis/google-cloud-go/blob/9a4cb31f4d34948404d91123fb560a43aeebe83e/firestore/document.go#L64-L127

Previous pull requests

- https://github.com/googleapis/google-cloud-go/pull/12217
- https://github.com/googleapis/google-cloud-go/pull/12425
- https://github.com/googleapis/google-cloud-go/pull/12538
- https://github.com/googleapis/google-cloud-go/pull/13147
- https://github.com/googleapis/google-cloud-go/pull/13199

PR #13194, opened 2025-10-21
feat(firestore): [PQ] Add all Aggregation and Timestamp functions

bhshkh:
b/364927702

1. add all the remaining private preview aggregate functions. Merging this PR completes the implementation of all the **type: "Function" subType : "Accumulators (Aggregation)"** private preview features.
See "Firestore Features (Pipeline)" sheet in [go/firestore-query-tracker](http://go/firestore-query-tracker) for the list of features.

    Java reference:
    - https://github.com/googleapis/java-firestore/blob/ccaf9d4fac5bd87a4da3d37493ca66fdc7681bc3/google-cloud-firestore/src/main/java/com/google/cloud/firestore/pipeline/expressions/AggregateFunction.java#L43-L71
    - https://github.com/googleapis/java-firestore/blob/ccaf9d4fac5bd87a4da3d37493ca66fdc7681bc3/google-cloud-firestore/src/main/java/com/google/cloud/firestore/pipeline/expressions/AggregateFunction.java#L93-L111

2. add all the remaining private preview timestamp functions. Merging this PR completes the implementation of all the **type: "Function" subType: "Date / Timestamp"** private preview features. (except timestamp_trunc function which is not yet inmplemented in any of the SDKs. Requires additonal approvals from Firestore team and will be added to separate PR).
See "Firestore Features (Pipeline)" sheet in [go/firestore-query-tracker](http://go/firestore-query-tracker) for the list of functions.

    Java reference:
    - https://github.com/googleapis/java-firestore/blob/ccaf9d4fac5bd87a4da3d37493ca66fdc7681bc3/google-cloud-firestore/src/main/java/com/google/cloud/firestore/pipeline/expressions/Expression.java#L2262-L2517

3. Add integration tests for all functions.
4. Remove Rand function since it is not targeted for private preview.
5. Renamed numericExprOrField to numericExprOrFieldPath since field is a separate type/expression.
https://github.com/googleapis/google-cloud-go/blob/a3ee1f19068c6d3fb77ad797e29884a90d6402a2/firestore/pipeline_field.go#L21-L41

Previous pull requests

- https://github.com/googleapis/google-cloud-go/pull/12217
- https://github.com/googleapis/google-cloud-go/pull/12425
- https://github.com/googleapis/google-cloud-go/pull/12538
- https://github.com/googleapis/google-cloud-go/pull/13147
- https://github.com/googleapis/google-cloud-go/pull/13199
- https://github.com/googleapis/google-cloud-go/pull/13218

PR #13339, opened 2025-11-11
feat(firestore): [PQ] create pipeline from query

bhshkh:
Add CreateFrom() and Pipeline() similar to Java.
https://github.com/googleapis/java-firestore/blob/742fab6583c9a6f9c47cf0496124c3c9b05fe0ee/google-cloud-firestore/src/main/java/com/google/cloud/firestore/PipelineSource.java#L140-L164

Previous pull requests

- https://github.com/googleapis/google-cloud-go/pull/12217
- https://github.com/googleapis/google-cloud-go/pull/12425
- https://github.com/googleapis/google-cloud-go/pull/12538
- https://github.com/googleapis/google-cloud-go/pull/13147
- https://github.com/googleapis/google-cloud-go/pull/13199
- https://github.com/googleapis/google-cloud-go/pull/13218
- https://github.com/googleapis/google-cloud-go/pull/13194
- https://github.com/googleapis/google-cloud-go/pull/13245
- https://github.com/googleapis/google-cloud-go/pull/13270
- https://github.com/googleapis/google-cloud-go/pull/13271
- https://github.com/googleapis/google-cloud-go/pull/13279
- https://github.com/googleapis/google-cloud-go/pull/13280
- https://github.com/googleapis/google-cloud-go/pull/13281
- https://github.com/googleapis/google-cloud-go/pull/13282
- https://github.com/googleapis/gapic-generator-go/pull/1661
- https://github.com/googleapis/google-cloud-go/pull/13283
- https://github.com/googleapis/google-cloud-go/pull/13274
- https://github.com/googleapis/google-cloud-go/pull/13338

### DIFF
--- a/firestore/pipeline_integration_test.go
+++ b/firestore/pipeline_integration_test.go
@@ -2138,62 +2138,62 @@ func logicalFuncs(t *testing.T) {
 		{
 			name:     "Conditional - true",
 			pipeline: client.Pipeline().Collection(coll.ID).Select(Conditional(Equal(ConstantOf(1), ConstantOf(1)), FieldOf("a"), FieldOf("b")).As("result")),
-			want:     map[string]interface{}{"result": int64(1)},
+			want:     []map[string]interface{}{{"result": int64(1)}, {"result": int64(1)}},
 		},
 		{
 			name:     "Conditional - false",
 			pipeline: client.Pipeline().Collection(coll.ID).Select(Conditional(Equal(ConstantOf(1), ConstantOf(0)), FieldOf("a"), FieldOf("b")).As("result")),
-			want:     map[string]interface{}{"result": int64(2)},
+			want:     []map[string]interface{}{{"result": int64(2)}, {"result": int64(1)}},
 		},
 		{
 			name:     "Conditional - field true",
 			pipeline: client.Pipeline().Collection(coll.ID).Select(Conditional(Equal(FieldOf("d"), ConstantOf(true)), FieldOf("a"), FieldOf("b")).As("result")),
-			want:     map[string]interface{}{"result": int64(1)},
+			want:     []map[string]interface{}{{"result": int64(1)}, {"result": int64(1)}},
 		},
 		{
 			name:     "Conditional - field false",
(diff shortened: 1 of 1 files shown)


firestore/query.go
```diff
@@ -1768,3 +1768,378 @@ type AggregationResponse struct {
...
+			if err != nil {
+				p.err = err
+				return p
+			}
+			fields = append(fields, fp)
+		}
+		p = p.Select(fields...)
+	}
+
+	// FindNearest
```
daniel-sanche, line 1896: I don't remember seeing this in other implementations. Makes sense though. I'll have to look into it more.

How does this interact with other filters? Can we assume a query is either a findNearest query or a standard query? Or do they stack?

firestore/query.go
```diff
@@ -1768,3 +1768,378 @@ type AggregationResponse struct {
...
+			orders = append(orders, Ordering{Expr: field, Direction: direction})
+		}
+		p = p.Sort(orders...)
+	}
+	// Combine all filters
+	if len(allFilters) == 1 {
+		p = p.Where(allFilters[0])
+	} else if len(allFilters) > 1 {
+		p = p.Where(And(allFilters[0], allFilters[1:]...))
+	}
```
daniel-sanche, line 1870: Your implementation seems to differ a bit from java. 
- Java does `.where(filters).select(projections).where(orders)`
- Yours does `where(fitlers and orders).select(projections)`

Is this equivalent? (It seems like it might be, but I'm not sure one way or the other)

firestore/query.go
bhshkh: They do indeed stack.

For example, a query like coll.Where("published", ">", 1900).FindNearest(...) would first filter for books published after 1900 and then find the nearest neighbors within that subset.

firestore/query.go
bhshkh: While the Go code builds the pipeline in a slightly different order than the Java version, the resulting pipeline is functionally equivalent and will produce the correct results.

daniel-sanche (approved):
LGTM

bhshkh:
Will take Go developer review before public preview

PR #13354, opened 2025-11-13
fix(storage): Handle redirect on takeover.

cjc25:
The initial takeover response wasn't being checked for a redirect.

PR #13239 addressed successful responses. This addresses error responses.

See #12785, b/459905219


cjc25:
kokoro failure looks like a read codec flake

PR #13368, opened 2025-11-17
chore(internal/postprocessor): remove the legacy OwlBot postprocessor

jskeet:
This is now obsolete as we have moved entirely onto Librarian. The GCB trigger has already been deleted.

Fixes https://github.com/googleapis/librarian/issues/2795

### DIFF
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -4735,7 +4735,6 @@ libraries:
       - internal/kokoro
       - internal/leakcheck
       - internal/librariangen
-      - internal/postprocessor
       - internal/pretty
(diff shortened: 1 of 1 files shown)


quartzmo (changes requested):
There is a reference in `state.yaml` that should be removed: https://github.com/googleapis/google-cloud-go/blob/main/.librarian/state.yaml#L4738

jskeet:
> There is a reference in `state.yaml` that should be removed: https://github.com/googleapis/google-cloud-go/blob/main/.librarian/state.yaml#L4738

Thanks, well spotted - done in an extra commit.

PR #13379, opened 2025-11-19
feat: add default checksums for appendable writer (excludes appendable takeover writer)

krishnamd-jkp:


storage/grpc_writer.go
```diff
@@ -1142,12 +1144,14 @@ type gRPCAppendBidiWriteBufferSender struct {
...
 	routingToken *string
 
-	firstMessage *storagepb.BidiWriteObjectRequest
-
-	objectChecksums *storagepb.ObjectChecksums
+	firstMessage    *storagepb.BidiWriteObjectRequest
 	finalizeOnClose bool
 	objResource     *storagepb.Object
 
+	checksumSettings   func() (bool, bool, *ObjectAttrs)
```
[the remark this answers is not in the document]
cpriti-os: This suggestion looks valid, please check

storage/doc.go
```diff
@@ -407,6 +407,9 @@ roles which must be enabled in order to do the export successfully. To
 disable this export, you can use the [WithDisabledClientMetrics] client
 option.
 
+The gRPC client's default auto checksumming can introduce increased CPU overhead during writes
```
cpriti-os: This doesn't look related to this PR? Should you raise a different small PR: results in quicker review since its a harmless documentation change.

krishnamd-jkp: This is applicable for the whole auto checksum feature which this PR is partly (for appendable write) about, right?

storage/grpc_writer.go
```diff
@@ -1409,12 +1424,20 @@ func (s *gRPCAppendBidiWriteBufferSender) send(stream storagepb.Storage_BidiWrit
...
+		bufChecksum = proto.Uint32(crc32.Checksum(r.buf, crc32cTable))
+	}
+	objectChecksums := getObjectChecksums(&getObjectChecksumsParams{
+		fullObjectChecksum:  s.fullObjectChecksum,
+		finishWrite:         finalizeObject,
+		sendCRC32C:          sendCrc32C,
+		disableAutoChecksum: disableAutoChecksum,
+		attrs:               attrs,
+		takeoverWriter:      s.takeoverWriter,
+	})
```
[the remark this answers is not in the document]
krishnamd-jkp: Done

storage/grpc_writer.go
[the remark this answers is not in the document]
krishnamd-jkp: Done

storage/doc.go
cpriti-os: Can we provide an approximate CPU overhead in percentage to help users get an idea of the impact?

storage/grpc_writer.go
```diff
@@ -818,12 +818,19 @@ func bidiWriteObjectRequest(r gRPCBidiWriteRequest, bufChecksum *uint32, objectC
 	return req
 }
 
-type getObjectChecksumsParams struct {
-	fullObjectChecksum  func() uint32
-	finishWrite         bool
+// checksumAttrs contains all parameters required from writer in senders.
+// helps avoid piping in writer to the individual senders.
```
cpriti-os: Nit: Either capitalize Helps or replace . with, after senders.

storage/integration_test.go
```diff
@@ -1844,6 +1844,146 @@ func TestIntegration_MultiChunkWrite(t *testing.T) {
...
+				content:   bytes.Repeat([]byte("a"), 1024*1024),
+				chunkSize: 256 * 1024,
+			},
+		}
+
+		for _, tc := range testCases {
+			t.Run(tc.name, func(t *testing.T) {
+				correctCRC32C := crc32.Checksum(tc.content, crc32cTable)
+
+				// Test case 1: Upload with correct checksum should succeed.
```
cpriti-os: Too many test cases with repeated code, can we refactor it more using table driven tests approach?

storage/integration_test.go
```diff
@@ -1844,6 +1844,146 @@ func TestIntegration_MultiChunkWrite(t *testing.T) {
...
+
+					if _, err := w.Write(tc.content); err != nil {
+						t.Fatalf("Writer.Write: %v", err)
+					}
+					// Flush the buffer to finish the upload.
+					err := w.Close()
+					if err == nil {
+						t.Fatal("expected an error for incorrect checksum, but got nil")
+					}
+					if !errorIsInvalidArgument(err) {
```
cpriti-os: this is a little generic, can we check for error string as well ?

storage/grpc_writer.go
```diff
@@ -818,12 +818,19 @@ func bidiWriteObjectRequest(r gRPCBidiWriteRequest, bufChecksum *uint32, objectC
...
+type checksumAttrs struct {
```
cpriti-os: thought: If we are adding a significant amount of code related to checksums, it might make sense to move it to a different file now, rather than later when it would become even more complex.

cpriti-os (changes requested):

storage/grpc_writer.go
krishnamd-jkp: I think with the refactor, the checksum related stuff boils down to 2 structs and a function. I feel it's okay to keep this in the same file

storage/integration_test.go
krishnamd-jkp: Changed.

storage/doc.go
krishnamd-jkp: I think it could be misleading to the users if we mention the percentage as it could not be the case for them for their workloads and environments. I'll let it just be (increased overhead now). For the record, we have observed 6% increase in the CPU overhead due to checksumming.

cpriti-os:
Can you change the title to explicitly call out the limitation: add default checksums for appendable writer except takeover writer. Feel free to rephrase, just want to ensure if this is released, the release notes should not mislead users.

storage/integration_test.go
```diff
@@ -1844,6 +1844,111 @@ func TestIntegration_MultiChunkWrite(t *testing.T) {
...
+				chunkSize: 256 * 1024,
+			},
+			{
+				name:                "resumable with disabled auto checksum",
+				content:             bytes.Repeat([]byte("a"), 1024*1024),
+				chunkSize:           256 * 1024,
+				disableAutoChecksum: true,
+			},
+			{
+				name:              "resumable with incorrect checksum",
```
cpriti-os, line 1895: Include a test that verifies that incorrect checksum is handled correctly for appendable writes.

storage/integration_test.go
```diff
@@ -1844,6 +1844,111 @@ func TestIntegration_MultiChunkWrite(t *testing.T) {
...
+			},
+			{
+				name:                "oneshot with disabled auto checksum",
+				content:             []byte("small content for oneshot upload"),
+				chunkSize:           0,
+				disableAutoChecksum: true,
+			},
+			{
+				name:       "resumable with user-sent CRC32C",
+				content:    bytes.Repeat([]byte("a"), 1024*1024),
```
cpriti-os: Use constants like 1MiB for better readability, Here and everywhere else.

cpriti-os:
Have we done benchmarking and profiling for appendable writes checksum as well @krishnamd-jkp ?

krishnamd-jkp:
> Have we done benchmarking and profiling for appendable writes checksum as well @krishnamd-jkp ?

@cpriti-os Because this PR introduces checksums only for appendable writer only, the code path is similar to a resumable or one shot writers so there won't be any additional latency. The CPU overhead will also be the same because computation of checksum is done by the same function.

I'd assume things would be a little different for takeover writer as it has a slightly different logic where writer needs to accept the crc32c from server and then calculate checksums for further writes

storage/doc.go
tritone: I would rephrase because we should consider this a positive feature of gRPC. How about:

`The gRPC client will automatically compute and send CRC32C checksums for uploads using [Writer], which provides an additional layer of data integrity validation vs the HTTP client. This can optionally be disabled by using [Writer.DisableAutoChecksum].`

The DisableAutoChecksum docs are sufficient to discuss the CPU overhead I think.

storage/doc.go
```diff
@@ -407,6 +407,9 @@ roles which must be enabled in order to do the export successfully. To
...
+because of checksum computation. Users can disable automatic checksumming for gRPC writer using [Writer.DisableAutoChecksum].
+
```
tritone, line 413: Do the writer.go docs need to be updated as well since we're changing the scope of what gets auto checksummed?

storage/grpc_writer.go
```diff
@@ -818,12 +818,19 @@ func bidiWriteObjectRequest(r gRPCBidiWriteRequest, bufChecksum *uint32, objectC
...
 	sendCRC32C          bool
 	disableAutoChecksum bool
-	attrs               *ObjectAttrs
+	objectAttrs         *ObjectAttrs
+	fullObjectChecksum  func() uint32
+}
+
+type getObjectChecksumsParams struct {
+	checksumAttrs
```
tritone: Nested structs here seems confusing, why separate this out?

storage/integration_test.go
```diff
@@ -7806,3 +7954,17 @@ func setUpRequesterPaysBucket(ctx context.Context, t *testing.T, bucket, object
 func crc32c(b []byte) uint32 {
 	return crc32.Checksum(b, crc32.MakeTable(crc32.Castagnoli))
 }
+
+func incorrectChecksumError(err error) bool {
```
tritone: Adding these test helper funcs is discouraged; see https://go.dev/wiki/TestComments#assert-libraries

storage/integration_test.go
```diff
@@ -1844,6 +1844,111 @@ func TestIntegration_MultiChunkWrite(t *testing.T) {
...
+			wantErr             bool
+		}{
+			{
+				name:       "oneshot with user-sent CRC32C",
+				content:    []byte("small content for oneshot upload"),
+				chunkSize:  0,
+				sendCrc32c: true,
+			},
+			{
+				name:      "oneshot with default CRC32",
```
tritone, line 1867: IIUC these cases without sendCrc32C don't actually validate whether auto checksumming was applied? Just trying to understand what they provide vs the existing writer tests.

Maybe it would make more sense to write a test with a gRPC stream interceptor to validate that the checksum was sent in the messages where it was expected?

storage/integration_test.go
```diff
@@ -1844,6 +1844,111 @@ func TestIntegration_MultiChunkWrite(t *testing.T) {
...
 
+func TestIntegration_WriterCRC32CValidation(t *testing.T) {
+	ctx := skipZonalBucket(context.Background(), "Test for resumable and oneshot writers")
+	h := testHelper{t}
+	multiTransportTest(ctx, t, func(t *testing.T, ctx context.Context, bucket string, _ string, client *Client) {
+		testCases := []struct {
+			name                string
+			content             []byte
+			chunkSize           int
+			sendCrc32c          bool
```
tritone: nit: this would be sendCRC32C https://go.dev/wiki/CodeReviewComments#initialisms

storage/grpc_writer.go
```diff
@@ -1409,12 +1428,17 @@ func (s *gRPCAppendBidiWriteBufferSender) send(stream storagepb.Storage_BidiWrit
...
+	var bufChecksum *uint32
+	if !s.checksumAttrs.disableAutoChecksum {
+		bufChecksum = proto.Uint32(crc32.Checksum(r.buf, crc32cTable))
 	}
+	objectChecksums := getObjectChecksums(&getObjectChecksumsParams{
+		checksumAttrs:  s.checksumAttrs,
+		finishWrite:    finalizeObject,
+		takeoverWriter: s.takeoverWriter,
+	})
+	req := bidiWriteObjectRequest(r, bufChecksum, objectChecksums)
```
tritone, line 1448: It's not 100% clear to me; is this sending per-buffer checksums only, or also full object checksums for finalized objects?

storage/doc.go
krishnamd-jkp: This PR introduces checksum only for appendable writer and not the takeover one. Checksum is sent only on finalize in appendable writer . "Automatic checksumming is not enabled for writes using the HTTP client or for unfinalized writes to appendable objects in gRPC." in writer.go still seems okay.

storage/grpc_writer.go
krishnamd-jkp: It is sending both. bufChecksum (checksum of buffer) and objectChecksums (which is full object checksum) for finalized write.

storage/integration_test.go
krishnamd-jkp: sendCrc32c is used when user wants to send their own checksum. default checksumming has nothing to do with this variable. We clarified this in the last PR with the comments. 

Sure interceptor tests sounds appropriate. I'll take them up in a new PR?

storage/doc.go
tritone: We would still expect message checksums even for unfinalized, right? Maybe just a small tweak.

storage/grpc_writer.go
tritone: Gotcha, sounds good.

storage/integration_test.go
krishnamd-jkp: But its not doing any assert in the helper function but just extracting out the error (based on a grpc error or a http one) and returning a bool. I thought it would not clutter the test. wdyt?

storage/grpc_writer.go
krishnamd-jkp: Since getObjectChecksumsParams is being used at multiple places and we can pass the property of senders (checksumAttrs) directly in without opening up the checksumAttrs struct and assigning each field individually at all the callers? As these fields are needed in the helper function as is

storage/doc.go
krishnamd-jkp: Added a small detail

storage/doc.go
krishnamd-jkp: Rephrased it

storage/integration_test.go
tritone: Sure, I'm just wondering about the value add of this test vs. existing writer tests which cover checksumming already. It doesn't seem to add much new coverage as is. Interceptor test can come separately if you prefer.

storage/integration_test.go
tritone: It's not exactly an assert but it does similarly add additional layers of indirection to the test. Maybe just check the error code is as expected, rather than string matching? The error strings are not guaranteed to remain constant over time so could lead to flakes.

storage/grpc_writer.go
tritone: I was mostly wondering why not just have the fields of checksumAttrs as fields of getObjectChecksumParams directly rather than having multiple structs

krishnamd-jkp: There are two structs because we get some properties from the writer like sendCRC32C, disableAutoChecksum, objectAttrs (like attrs of checksum -> checksumAttrs). And then finishWrite and takeoverWriter which are required by getObjectChecksums function but we get them during writes. getObjectChecksumsParams is mostly to avoid the function parameters from getting too long

storage/integration_test.go
krishnamd-jkp: I'll remove the unnecessary ones in the next pr which implements interceptor test

storage/grpc_writer.go
tritone: Having getObjectChecksumParams as a type seems fine. It seems odd to use struct embedding here -- I have generally only seen it when e.g. we are implementing an interface or doing a union type. I also think checksumSettings might make more sense as a name than checksumAttrs? I see that you deleted the func that used to be called that.

Ultimately I don't think it's a huge deal, just found this a little confusing to read and follow.

tritone (dismissed):
I left a comment re: checksumAttrs struct; will approve based on other changes but maybe sync with Priti and get her thoughts before merging.

cpriti-os (dismissed):

storage/grpc_writer.go
cpriti-os: There is no real need for nested structs here, discussed offline, Dhanush will update with a flattened struct

storage/integration_test.go
```diff
@@ -3409,8 +3465,25 @@ func TestIntegration_WriterAppend(t *testing.T) {
 				w.Append = true
 				w.FinalizeOnClose = tc.finalize
 				w.ChunkSize = tc.chunkSize
+				w.SendCRC32C = tc.sendCRC
+				w.DisableAutoChecksum = tc.disableAutoChecksum
+				w.CRC32C = crc32.Checksum(tc.content, crc32cTable)
 				content := tc.content
 
+				// If incorrectChecksum is true, write data and close writer
+				// immediately to validate if writer returns error
```
cpriti-os, line 3474: Nit: Statements should end with a period in Go

PR #13390, opened 2025-11-20
feat(storage): add object contexts in Go GCS SDK

krishnamd-jkp:


storage/grpc_client.go
```diff
@@ -2662,3 +2675,19 @@ func (r *gRPCReader) reopenStream() error {
...
 	return nil
 }
+
+func toStringCustomContext(cc *CustomContext) string {
+	if cc == nil || cc.Key == "" {
+		return ""
+	}
+	var filter string
+	if cc.Value != "" {
+		filter = fmt.Sprintf(`contexts."%s"="%s"`, cc.Key, cc.Value)
```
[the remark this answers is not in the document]
krishnamd-jkp: The value should not have any quotes in it.

storage/contexts.go
```diff
@@ -0,0 +1,117 @@
...
+	return &ObjectContexts{
+		Custom: customContexts,
+	}
+}
+
+// toRawObjectContexts converts the object contexts to the raw library's ObjectContexts type.
+func toRawObjectContexts(c *ObjectContexts) *raw.ObjectContexts {
+	if c == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
krishnamd-jkp: Done

storage/contexts_test.go
```diff
@@ -0,0 +1,252 @@
...
+					if gotPayload, ok := got.Custom[k]; ok {
+						if len(wantPayload.ForceSendFields) > 0 && len(gotPayload.ForceSendFields) == 0 {
+							t.Errorf("toRawObjectContexts() mismatch for key %s: ForceSendFields missing", k)
+						}
+						if len(wantPayload.NullFields) > 0 && len(gotPayload.NullFields) == 0 {
+							t.Errorf("toRawObjectContexts() mismatch for key %s: NullFields missing", k)
+						}
+					}
+				}
+			}
```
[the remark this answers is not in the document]
krishnamd-jkp: Done

storage/grpc_client.go
```diff
@@ -2662,3 +2675,19 @@ func (r *gRPCReader) reopenStream() error {
...
+	} else {
+		filter = fmt.Sprintf(`contexts."%s":*`, cc.Key)
+	}
+	if cc.Absence {
+		filter = fmt.Sprintf(`-%s`, filter)
+	}
+	return filter
+}
```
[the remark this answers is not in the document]
krishnamd-jkp: Done

storage/storage.go
```diff
@@ -1119,6 +1119,12 @@ type ObjectAttrsToUpdate struct {
 	// extending the RetainUntil time on the object retention must be done
 	// on an ObjectHandle with OverrideUnlockedRetention set to true.
 	Retention *ObjectRetention
+
+	// Contexts allows adding, modifying, or deleting individual object contexts.
+	// To add or modify a context, set the value field in ObjectCustomContextPayload.
+	// To delete a context, set the Delete field in ObjectCustomContextPayload to true.
+	// to remove al contexts, pass &ObjectContexts{} in Contexts making sure Custom is empty.
```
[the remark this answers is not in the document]
krishnamd-jkp: Done

storage/integration_test.go
```diff
@@ -7032,6 +7032,687 @@ func (te *openTelemetryTestExporter) Unregister(ctx context.Context) {
...
+			}
+			wantContexts := &ObjectContexts{
+				Custom: map[string]ObjectContextValue{
+					"source1Key1": {Value: "source1Value1"},
+					"source2Key1": {Value: "source2Value1"},
+				},
+			}
+			if diff := cmp.Diff(dstAttrs.Contexts.Custom, wantContexts.Custom, cmpopts.IgnoreFields(ObjectContextValue{}, "UpdateTime", "CreateTime")); diff != "" {
+				t.Errorf("Inherited CustomContext mismatch (-got +want):\n%s", diff)
+			}
```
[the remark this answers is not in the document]
krishnamd-jkp: It is working as expected. Expectation is to merge the contexts on the destination object

storage/integration_test.go
```diff
@@ -7032,6 +7032,687 @@ func (te *openTelemetryTestExporter) Unregister(ctx context.Context) {
...
+		h.mustWrite(wc2, []byte("content2"))
+		defer h.mustDeleteObject(obj2)
+
+		object3Name := prefix + uidSpaceObjects.New() + "-no-ctx"
+		obj3 := client.Bucket(bucket).Object(object3Name)
+		h.mustWrite(obj3.NewWriter(ctx), []byte("content3"))
+		defer h.mustDeleteObject(obj3)
+
+		// 1. Get an existing object's metadata: Contexts included in the response.
+		t.Run("GetMetadataWithContexts", func(t *testing.T) {
```
tritone: Many of these would be more concise and easier to read as table-driven tests. https://go.dev/wiki/TableDrivenTests

In general this feels like a lot of integration tests for 1 metadata feature. How long do they take to run? can we leverage existing tests for Copy, List etc?

storage/integration_test.go
```diff
@@ -7032,6 +7032,687 @@ func (te *openTelemetryTestExporter) Unregister(ctx context.Context) {
 	te.tp.Shutdown(ctx)
 }
 
+func TestIntegration_Object_Create_CustomContexts(t *testing.T) {
```
tritone: Should stick to camelcase for the test name, e.g. TestIntegration_ObjectCreateCustomContexts

storage/contexts.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func toProtoObjectContexts(c *ObjectContexts) *storagepb.ObjectContexts {
+	if c == nil {
+		return nil
+	}
+	customContexts := make(map[string]*storagepb.ObjectCustomContextPayload)
+	for k, v := range c.Custom {
+		if !v.Delete {
```
cpriti-os, line 106: Why are we not doing anything when Delete is set to true?

storage/storage.go
```diff
@@ -1119,6 +1119,12 @@ type ObjectAttrsToUpdate struct {
 	// extending the RetainUntil time on the object retention must be done
 	// on an ObjectHandle with OverrideUnlockedRetention set to true.
 	Retention *ObjectRetention
+
+	// Contexts allows adding, modifying, or deleting individual object contexts.
+	// To add or modify a context, set the value field in ObjectCustomContextPayload.
+	// To delete a context, set the Delete field in ObjectCustomContextPayload to true.
+	// to remove all contexts, pass &ObjectContexts{} in Contexts making sure Custom is empty.
```
cpriti-os: Can there be a better way to remove all contexts, this does not look very intuitive. Mayve a separate attribute to update?

storage/http_client.go
```diff
@@ -518,6 +525,18 @@ func (c *httpStorageClient) UpdateObject(ctx context.Context, params *updateObje
 			forceSendFields = append(forceSendFields, "Retention")
 		}
 	}
+
+	if uattrs.Contexts != nil {
+		if len(uattrs.Contexts.Custom) == 0 {
+			// Sending the empty map is a no-op. Pass it in null fields
+			nullFields = append(nullFields, "Contexts")
+		} else {
```
cpriti-os, line 529: I think this nested if else block can be flattened, can you check and simplify this.

krishnamd-jkp: This follows similar structure of the code. Flattening it will make it less readable. What do you think?

storage/storage.go
krishnamd-jkp: Added a new field

storage/contexts.go
krishnamd-jkp: Added a comment.

storage/contexts.go
```diff
@@ -0,0 +1,139 @@
...
+				Value: v.Value,
+			}
+		}
+	}
+	return &storagepb.ObjectContexts{
+		Custom: customContexts,
+	}
+}
+
+func toStringCustomContext(cc *CustomContext) string {
```
tritone: This has a bit of a smell to me. Is it a requirement to make a CustomContext type in each language which does this sort of cleverness? Or would it be easier to just expose Filter directly as a field on Query and allow the user to construct the appropriate filter string themself?

storage/contexts.go
```diff
@@ -0,0 +1,139 @@
...
+
+// ObjectContexts is a container for object contexts.
+type ObjectContexts struct {
+	Custom map[string]ObjectContextValue
+	// Custom field is ignored when ClearCustomContexts is set to true.
+	ClearCustomContexts bool
+}
+
+// ObjectContextValue holds the value of a user-defined object context.
+type ObjectContextValue struct {
```
tritone: Maybe this should just be called ObjectContext, or ObjectCustomContext (to mirror the proto msg)? Having Value in the name and also as a field is confusing.

storage/contexts.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"cloud.google.com/go/storage/internal/apiv2/storagepb"
+	raw "google.golang.org/api/storage/v1"
+)
+
+// ObjectContexts is a container for object contexts.
+type ObjectContexts struct {
+	Custom map[string]ObjectContextValue
+	// Custom field is ignored when ClearCustomContexts is set to true.
+	ClearCustomContexts bool
```
tritone: This field seems unnecessary since it's used only for metadata updates.

Can you use passing an empty struct to ObjectAttrsToUpdate to signify this? (vs nil meaning don't update at all).

tritone: Actually I see that @cpriti-os said the opposite here. FWIW there are multiple fields in ObjectAttrsToUpdate with the semantics where passing an empty object is the way to clear the field.

storage/storage.go
```diff
@@ -1670,6 +1683,12 @@ type ObjectAttrs struct {
 	// ObjectHandle.Attrs will return ErrObjectNotExist if the object is soft-deleted.
 	// This field is read-only.
 	HardDeleteTime time.Time
+
+	// Contexts store custom key-value metadata that the user could
+	// annotate object with. These key-value pairs can be used to filter objects
+	// during list calls. See https://cloud.google.com/storage/docs/object-contexts
+	// for more details.
+	Contexts *ObjectContexts `json:"contexts,omitempty"`
```
BrennaEpp: Why do we have this json annotation here?

storage/contexts.go
cpriti-os: If this is the pattern followed everywhere else, we can stick to empty object for clearing.

storage/http_client.go
cpriti-os: Works fine then.

storage/contexts.go
cpriti-os, line 1: Just a thought: should we name it "object_contexts" itself? There is a slight possibility that we have/will have other resources which can have contexts?

storage/contexts_test.go
```diff
@@ -0,0 +1,287 @@
...
+			opts := cmpopts.SortSlices(func(a, b string) bool { return a < b })
+			if diff := cmp.Diff(tc.want, got, opts); diff != "" {
+				t.Errorf("toRawObjectContexts() mismatch (-want +got): %s", diff)
+			}
+		})
+	}
+}
+
+func TestToObjectContextsFromProto(t *testing.T) {
+	now := time.Now().UTC().Truncate(time.Millisecond)
```
cpriti-os: Can we use fixed time here and anywhere else? Unit tests should be as simple and deterministic as possible.

storage/grpc_client.go
```diff
@@ -630,6 +631,18 @@ func (c *grpcStorageClient) UpdateObject(ctx context.Context, params *updateObje
 		}
 	}
 
+	if uattrs.Contexts != nil {
```
cpriti-os: Would be better to use helper function here and in http_client.go to keep UpdateObject method clean and modular?

storage/contexts.go
krishnamd-jkp: We can add them in the same file when we have contexts  for other resources like buckets

storage/grpc_client.go
krishnamd-jkp: The update logic is different for gRPC and HTTP. I think it's okay as it again follows other metadata fields

storage/integration_test.go
krishnamd-jkp: Cleared redundant tests and used existing compose and copy tests.

storage/integration_test.go
krishnamd-jkp: Done

storage/contexts.go
krishnamd-jkp: Made it ObjectCustomContextPayload inline with other SDKs.

storage/contexts.go
krishnamd-jkp: Reverted

storage/storage.go
krishnamd-jkp: Removed it

storage/contexts_test.go
krishnamd-jkp: Done

storage/contexts.go
krishnamd-jkp: I introduced this to just make it user friendly. It wouldn't be user friendly to expect users to give the string themselves, would it? Given it has unintuitive format like providing "-" for negation, "key=value", "key:*" etc.

tritone: I think the place to introduce this kind of a structured format would generally be in the proto message definition rather than in the SDKs. Can you check how this is implemented in other languages and/or check in with the team doing the API launch to see what they think?

tritone: The Java API looks like it's just using filter as a text field directly so IMO it's also fine to do that in Go https://github.com/googleapis/java-storage/pull/3329/files#diff-20bcaa643fb27f1cb3b69cd788ade6590ecec05456217cc7103513ca3a89ba94

cpriti-os: +1, this should not be part of SDK and it adds unnecessary maintenance burden.

storage/http_client.go
```diff
@@ -518,6 +520,19 @@ func (c *httpStorageClient) UpdateObject(ctx context.Context, params *updateObje
 			forceSendFields = append(forceSendFields, "Retention")
 		}
 	}
+
+	if uattrs.Contexts != nil && uattrs.Contexts.Custom != nil {
+
+		if len(uattrs.Contexts.Custom) == 0 {
+			// Sending the empty map is a no-op. Pass it in null fields
```
cpriti-os: Nit: I suggest rephrasing it for clarity: "To delete all contexts, "Contexts" must be added to nullFields.
An empty Custom map in the request body is a no-op without this."

storage/contexts.go
cpriti-os, line 1: Should we add some documentation to doc.go for this new feature?

storage/storage.go
```diff
@@ -1933,6 +1954,20 @@ type Query struct {
...
+	// Filters objects based on object attributes like custom contexts.
+	// See https://docs.cloud.google.com/storage/docs/listing-objects#filter-by-object-contexts
+	// for more details.
+	Filter string
+}
+
+// CustomContext is used to filter results to objects that have a matching
+// context key and value. Absence can also be passed which will result in objects
+// not matching key and value
+type CustomContext struct {
```
cpriti-os: Is this struct used anywhere?

storage/contexts.go
krishnamd-jkp: We'll have samples to let users know about this feature: https://cloud.google.com/storage/docs/use-object-contexts#attach-modify-contexts. Also, new features go in CHANGES.md anyways.

Issue #10378, opened 2024-06-13
spanner: sessions are not being closed when client is closed


egonelbre:
I was debugging https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/159 problem and found out that using `spanner.Client` leaves about 100 sessions open per client. This in test environments can lead to a situation where creating 1000 clients ands up creating 80k sessions.

This eventually leads to a situation where you can have millions of sessions, which slows down trivial queries to 1s+.

It seems that `client.Close` does not close the batch created sessions.

**Code**

<details>
    <summary>Reproducer</summary>

```go
package main

import (
	"context"
	"errors"
	"fmt"
	"os"
	"os/signal"
	"time"

	"cloud.google.com/go/spanner"
	database "cloud.google.com/go/spanner/admin/database/apiv1"
	"cloud.google.com/go/spanner/admin/database/apiv1/databasepb"
	instance "cloud.google.com/go/spanner/admin/instance/apiv1"
	"cloud.google.com/go/spanner/admin/instance/apiv1/instancepb"
	spannerdirect "cloud.google.com/go/spanner/apiv1"
	"cloud.google.com/go/spanner/apiv1/spannerpb"
	"golang.org/x/sync/errgroup"
	"google.golang.org/api/iterator"
	"google.golang.org/api/option"
	"google.golang.org/api/option/internaloption"
	"google.golang.org/grpc"
	"google.golang.org/grpc/codes"
)

func main() {
	err := run()
	if err != nil {
		fmt.Fprint(os.Stderr, err)
		os.Exit(1)
	}
}

func run() error {
	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
	defer cancel()

	const (
		ProjectID, InstanceID, DatabaseID = "proj", "inst", "db"

		Database = "projects/" + ProjectID + "/instances/" + InstanceID + "/databases/" + DatabaseID
	)

	{ // create a new instance
		admin, err := instance.NewInstanceAdminClient(ctx)
		if err != nil {
			return fmt.Errorf("failed to create instance admin: %w", err)
		}
		op, err := admin.CreateInstance(ctx, &instancepb.CreateInstanceRequest{
			Parent:     "projects/" + ProjectID,
			InstanceId: InstanceID,
		})
		if err != nil {
			if spanner.ErrCode(err) == codes.AlreadyExists {
				goto instanceReady
			}
			return fmt.Errorf("create instance failed: %w", err)
		}
[...]

PR #13395, opened 2025-11-21
fix(spanner): avoid double decrement in session counting

egonelbre:


rahul2393:
@egonelbre Thanks for opening this PR, we are planning to remove all session pool related code in next release since by default client lib uses multiplexed sessions, are you still using regular session in your application? if yes, can you please tell us any specific reason for using it?

egonelbre:
@rahul2393 sure.

So we've delayed upgrading spanner library for some time because they ended up stalling out our CI pipeline and were unsure why. But, we don't have a nice small reproducer. Yesterday and today I finally had time to bisect the issue to https://github.com/googleapis/google-cloud-go/commit/0491ba6f258600cc518a5a2c274caa0fb6105c8a commit. After updating to latest spanner library and "GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS=false" the CI pipeline does seem to work again.

The high level scenario is following, the CI pipeline sets up 7 spanner emulators (to avoid emulator internal locks), each tests sets up and tears down a new database/tables. Tests run in a highly parallel fashion (I would guess about 8000 tests, each setting up 2-3 databases per test and 2-6 connection pools per test). After some time the spanner emulator stops accepting connections. This suggests that something isn't being properly cleaned up. I'm guessing that either connections or sessions are left open, which eventually exhaust the spanner emulator. Or that some goroutines are not being shut down. Or that some retry mechanism is causing problems. But, it could also be a spanner emulator bug with regards to multiplexed sessions that break it down.

I'm still figuring out what the next best step is in debugging multiplexed sessions. Probably will either try to add some monitoring against spanner emulators. Any suggestions for debugging would be very welcome.

This exact fix came about when I ran with that specific environment flag disabled, and had one test failure with the specific message. Directed gemini to diagnose the issue for me. And, if the single sessions are being removed I'm fine closing the PR.

egonelbre:
It looks like one of the issues is https://github.com/googleapis/google-cloud-go/pull/13396; still testing whether it resolves the full problem.

EDIT: seems that it's not sufficient, but it does avoid goroutine buildup.

egonelbre:
@rahul2393 as far as I can able to debug. It seems it's partially related to https://github.com/googleapis/google-cloud-go/issues/10378 issue; and partially that the spanner emulator falters around having ~10000 sessions open (https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/issues/159). I'm not yet sure why multiplexed connections make it worse.

rahul2393:
@olavloite / @sakthivelmanii need help in one more approval to merge this one, please help

olavloite (approved):

PR #13396, opened 2025-11-21
fix(spanner): fix createMultiplexedSession goroutine leak

egonelbre:


spanner/session.go
```diff
@@ -1067,6 +1067,7 @@ func (p *sessionPool) close(ctx context.Context) {
 	}
 	p.mu.Unlock()
 	p.hc.close()
+	close(p.multiplexedSessionReq)
```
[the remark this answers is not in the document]
egonelbre: In sessionPool.close `p.valid = false` is set to false and after that the multiplexedSessionReq is closed.

rahul2393 (approved):

rahul2393:
@olavloite / @sakthivelmanii need help in one more approval to merge this one, please help

spanner/session.go
olavloite, line 1070: This change would get lost if we remove the entire `sessionPool` implementation (which is the intention). So it might be better to put this line of code in the `Client#Close()` function instead to make sure it is not unintentionally deleted. AFAICT, there are no tests that verify that this is actually closed, meaning that it would also not be caught by any tests.

egonelbre: I think the property that should be enforced is that all started goroutines are stopped. If the channels is left open and no one is using it, then it probably doesn't matter.

Using synctest seems a good fit for this. However, it does require Go 1.25 or 1.24 with GOEXPERIMENT=synctest. I wouldn't want to bump minimum language version requirement just for that... however the test could be put behind build tag.

egonelbre: I created a draft for that test, however it fails at the moment https://github.com/googleapis/google-cloud-go/pull/13406. Need to diagnose the issue, might require some additional fixes.

olavloite: SG. I didn't really realize that the entire multiplexed session implementation lives in the session pool, so my comment that this call should be moved to outside the pool does not really make any sense.

olavloite (approved):

spanner/session.go
egonelbre: I managed to fix https://github.com/googleapis/google-cloud-go/pull/13406 and it succeeds when this PR is merged. However, the mock server still leaks, which I excluded from the check.

anowardear062-svg (approved):

anowardear062-svg (approved):

PR #13397, opened 2025-11-21
feat(auth): deprecate unsafe credentials JSON loading options

quartzmo:
* Add deprecation comments to credentials.DetectOptions.CredentialsFile
  and credentials.DetectOptions.CredentialsJSON.
* Update documentation to recommend new safe load methods.
* Add credentials.CredentialsType.
* Add credentials.NewCredentialsFromJSON and credentials.NewCredentialsFromFile.
* Add deprecation comments to idtoken.Options.CredentialsFile
  and idtoken.Options.CredentialsJSON.
* Add idtoken.NewCredentialsFromJSON and idtoken.NewCredentialsFromFile.
* Replace internal integer-based `CredentialsType` (iota) with
  public string-based constants for specifying credential file
  types. This change:
  * Eliminates the zero-value problem
  * Improves readability and debugging
  * Ensures consistency with JSON data

### DIFF
--- a/auth/credentials/detect.go
+++ b/auth/credentials/detect.go
@@ -321,12 +321,12 @@ type DetectOptions struct {
 // Google APIs can compromise the security of your systems and data. For
 // more information, refer to [Validate credential configurations from
 // external sources](https://cloud.google.com/docs/authentication/external/externally-sourced-credentials).
-func NewCredentialsFromFile(ctx context.Context, credType CredType, filename string, opts *DetectOptions) (*auth.Credentials, error) {
+func NewCredentialsFromFile(credType CredType, filename string, opts *DetectOptions) (*auth.Credentials, error) {
 	b, err := os.ReadFile(filename)
 	if err != nil {
 		return nil, err
 	}
-	return NewCredentialsFromJSON(ctx, credType, b, opts)
+	return NewCredentialsFromJSON(credType, b, opts)
 }
 
 // NewCredentialsFromJSON creates a [cloud.google.com/go/auth.Credentials] from
@@ -341,7 +341,7 @@ func NewCredentialsFromFile(ctx context.Context, credType CredType, filename str
 // Google APIs can compromise the security of your systems and data. For
 // more information, refer to [Validate credential configurations from
 // external sources](https://cloud.google.com/docs/authentication/external/externally-sourced-credentials).
-func NewCredentialsFromJSON(ctx context.Context, credType CredType, b []byte, opts *DetectOptions) (*auth.Credentials, error) {
+func NewCredentialsFromJSON(credType CredType, b []byte, opts *DetectOptions) (*auth.Credentials, error) {
 	if err := checkCredentialType(b, credType); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


auth/credentials/detect.go
```diff
@@ -52,6 +52,44 @@ var (
 	allowOnGCECheck = true
 )
 
+// CredType specifies the type of JSON credentials being provided
+// to a loading function such as [NewCredentialsFromFile] or
+// [NewCredentialsFromJSON].
+type CredType string
```
sai-sunder-s, line 58: why string and not an enum?

quartzmo: * Replace internal integer-based CredentialsType (iota) with public string-based constants for specifying credential file types. This change:
  * Eliminates the zero-value problem.
  * Improves readability and debugging.
  * Ensures consistency with JSON data.

quartzmo: I'm glad you asked about this. I initially went with an `iota` enum, the standard approach. But the string typecodes were already very much baked into the library internally in the `parseCredentialType` function. And for error messages I would need a reverse conversion back to the iota enum, which I added in the first commit in this PR: `ParseCredentialTypeString`. Given that the string typecodes were used in every operation concerning the iota enum already, and would even be completely round-tripped for an error, I decided it made sense to elevate them to the exposed types. And thus in [ad56c84](https://github.com/googleapis/google-cloud-go/pull/13397/commits/ad56c84328d0e1b3c46dcf7c86e3defd59a2af32) I was able to delete the functions below. 

The GCA review above was very favorable: `Credential Type Standardization: Introduced a new public CredentialsType string enum to replace an internal integer-based CredentialType, improving readability, debugging, and consistency across credential handling.`

```
// parseCredentialType returns the associated filetype based on the parsed
// typeString provided.
func parseCredentialType(typeString string) CredentialType {
	switch typeString {
	case "service_account":
		return ServiceAccountKey
	case "authorized_user":
		return UserCredentialsKey
	case "impersonated_service_account":
		return ImpersonatedServiceAccountKey
	case "external_account":
		return ExternalAccountKey
	case "external_account_authorized_user":
		return ExternalAccountAuthorizedUserKey
	case "gdch_service_account":
		return GDCHServiceAccountKey
	default:
		return UnknownCredType
	}
}

// ParseCredentialTypeString returns the associated filetype string based
// on the parsed type code int provided.
func ParseCredentialTypeString(credType CredentialType) string {
	switch credType {
	case ServiceAccountKey:
		return "service_account"
	case UserCredentialsKey:
		return "authorized_user"
	case ImpersonatedServiceAccountKey:
		return "impersonated_service_account"
	case ExternalAccountKey:
		return "external_account"
	case ExternalAccountAuthorizedUserKey:
		return "external_account_authorized_user"
	case GDCHServiceAccountKey:
		return "gdch_service_account"
	default:
		return "unknown"
		return "", err
	}
	return f.Type, nil
}
```

auth/internal/credsfile/parse.go
```diff
@@ -89,10 +89,11 @@ type fileTypeChecker struct {
 }
 
 // ParseFileType determines the [CredentialType] based on bytes provided.
```
hongalex, line 91: nit: remove linking to `CredentialType`

quartzmo: I'll leave this as-is since it's internal. Just to save getting it re-reviewed.

PR #13417, opened 2025-12-01
feat(auth): Implement updated design for regional access boundary

nbayati:
Bring RAB (formerly known as trust boundaries) up to date with the design changes:
- Decouple RAB fetching from token refresh and use api calls as a trigger to fetch RAB.
- Make fetching async, non blocking, and fail open
- remove no-op signal
- Implement a 6 hour TTL with a soft expiry and proactive refresh

### DIFF
--- a/auth/credentials/detect.go
+++ b/auth/credentials/detect.go
@@ -27,7 +27,7 @@ import (
 	"cloud.google.com/go/auth"
 	"cloud.google.com/go/auth/internal"
 	"cloud.google.com/go/auth/internal/credsfile"
-	"cloud.google.com/go/auth/internal/trustboundary"
+	"cloud.google.com/go/auth/internal/regionalaccessboundary"
 	"cloud.google.com/go/compute/metadata"
 	"github.com/googleapis/gax-go/v2/internallog"
 )
@@ -142,7 +142,7 @@ func DetectDefault(opts *DetectOptions) (*auth.Credentials, error) {
 	if err := opts.validate(); err != nil {
 		return nil, err
 	}
-	trustBoundaryEnabled, err := trustboundary.IsEnabled()
+	trustBoundaryEnabled, err := regionalaccessboundary.IsEnabled()
 	if err != nil {
 		return nil, err
 	}
@@ -176,11 +176,11 @@ func DetectDefault(opts *DetectOptions) (*auth.Credentials, error) {
 
 		tp := computeTokenProvider(opts, metadataClient)
 		if trustBoundaryEnabled {
-			gceConfigProvider := trustboundary.NewGCEConfigProvider(gceUniverseDomainProvider)
+			gceConfigProvider := regionalaccessboundary.NewGCEConfigProvider(gceUniverseDomainProvider)
 			var err error
-			tp, err = trustboundary.NewProvider(opts.client(), gceConfigProvider, opts.logger(), tp)
+			tp, err = regionalaccessboundary.NewProvider(opts.client(), gceConfigProvider, opts.logger(), tp)
 			if err != nil {
-				return nil, fmt.Errorf("credentials: failed to initialize GCE trust boundary provider: %w", err)
+				return nil, fmt.Errorf("credentials: failed to initialize GCE Regional Access Boundary provider: %w", err)
 			}
 
 		}
--- a/auth/credentials/detect.go
+++ b/auth/credentials/detect.go
@@ -142,7 +142,7 @@ func DetectDefault(opts *DetectOptions) (*auth.Credentials, error) {
 	if err := opts.validate(); err != nil {
 		return nil, err
 	}
-	trustBoundaryEnabled, err := regionalaccessboundary.IsEnabled()
+	regionalAccessBoundaryEnabled, err := regionalaccessboundary.IsEnabled()
 	if err != nil {
 		return nil, err
 	}
@@ -175,7 +175,7 @@ func DetectDefault(opts *DetectOptions) (*auth.Credentials, error) {
 		}
 
 		tp := computeTokenProvider(opts, metadataClient)
-		if trustBoundaryEnabled {
+		if regionalAccessBoundaryEnabled {
 			gceConfigProvider := regionalaccessboundary.NewGCEConfigProvider(gceUniverseDomainProvider)
 			var err error
 			tp, err = regionalaccessboundary.NewProvider(opts.client(), gceConfigProvider, opts.logger(), tp)
--- a/auth/credentials/filetypes.go
+++ b/auth/credentials/filetypes.go
@@ -25,7 +25,7 @@ import (
 	"cloud.google.com/go/auth/credentials/internal/impersonate"
 	internalauth "cloud.google.com/go/auth/internal"
 	"cloud.google.com/go/auth/internal/credsfile"
-	"cloud.google.com/go/auth/internal/trustboundary"
+	"cloud.google.com/go/auth/internal/regionalaccessboundary"
 )
 
 const cloudPlatformScope = "https://www.googleapis.com/auth/cloud-platform"
@@ -159,15 +159,15 @@ func handleServiceAccount(f *credsfile.ServiceAccountFile, opts *DetectOptions) 
 		return nil, err
 	}
 
-	trustBoundaryEnabled, err := trustboundary.IsEnabled()
+	regionalAccessBoundaryEnabled, err := regionalaccessboundary.IsEnabled()
 	if err != nil {
 		return nil, err
 	}
-	if !trustBoundaryEnabled {
+	if !regionalAccessBoundaryEnabled {
 		return tp, nil
 	}
-	saConfig := trustboundary.NewServiceAccountConfigProvider(opts2LO.Email, opts2LO.UniverseDomain)
-	return trustboundary.NewProvider(opts.client(), saConfig, opts.logger(), tp)
+	saConfig := regionalaccessboundary.NewServiceAccountConfigProvider(opts2LO.Email, opts2LO.UniverseDomain)
+	return regionalaccessboundary.NewProvider(opts.client(), saConfig, opts.logger(), tp)
 }
 
 func handleUserCredential(f *credsfile.UserCredentialsFile, opts *DetectOptions) (auth.TokenProvider, error) {
@@ -210,35 +210,35 @@ func handleExternalAccount(f *credsfile.ExternalAccountFile, opts *DetectOptions
 	if err != nil {
 		return nil, err
 	}
-	trustBoundaryEnabled, err := trustboundary.IsEnabled()
+	regionalAccessBoundaryEnabled, err := regionalaccessboundary.IsEnabled()
 	if err != nil {
 		return nil, err
 	}
-	if !trustBoundaryEnabled {
+	if !regionalAccessBoundaryEnabled {
 		return tp, nil
 	}
 
 	ud := resolveUniverseDomain(opts.UniverseDomain, f.UniverseDomain)
-	var configProvider trustboundary.ConfigProvider
+	var configProvider regionalaccessboundary.ConfigProvider
 
 	if f.ServiceAccountImpersonationURL == "" {
 		// No impersonation, this is a direct external account credential.
-		// The trust boundary is based on the workload/workforce pool.
+		// The Regional Access Boundary is based on the workload/workforce pool.
 		var err error
-		configProvider, err = trustboundary.NewExternalAccountConfigProvider(f.Audience, ud)
+		configProvider, err = regionalaccessboundary.NewExternalAccountConfigProvider(f.Audience, ud)
 		if err != nil {
 			return nil, err
 		}
 	} else {
-		// Impersonation is used. The trust boundary is based on the target service account.
+		// Impersonation is used. The Regional Access Boundary is based on the target service account.
 		targetSAEmail, err := impersonate.ExtractServiceAccountEmail(f.ServiceAccountImpersonationURL)
 		if err != nil {
-			return nil, fmt.Errorf("credentials: could not extract target service account email for trust boundary: %w", err)
+			return nil, fmt.Errorf("credentials: could not extract target service account email for Regional Access Boundary: %w", err)
 		}
-		configProvider = trustboundary.NewServiceAccountConfigProvider(targetSAEmail, ud)
+		configProvider = regionalaccessboundary.NewServiceAccountConfigProvider(targetSAEmail, ud)
 	}
 
-	return trustboundary.NewProvider(opts.client(), configProvider, opts.logger(), tp)
+	return regionalaccessboundary.NewProvider(opts.client(), configProvider, opts.logger(), tp)
 }
 
 func handleExternalAccountAuthorizedUser(f *credsfile.ExternalAccountAuthorizedUserFile, opts *DetectOptions) (auth.TokenProvider, error) {
@@ -257,20 +257,20 @@ func handleExternalAccountAuthorizedUser(f *credsfile.ExternalAccountAuthorizedU
 	if err != nil {
 		return nil, err
 	}
-	trustBoundaryEnabled, err := trustboundary.IsEnabled()
+	regionalAccessBoundaryEnabled, err := regionalaccessboundary.IsEnabled()
 	if err != nil {
 		return nil, err
 	}
-	if !trustBoundaryEnabled {
+	if !regionalAccessBoundaryEnabled {
 		return tp, nil
 	}
 
 	ud := resolveUniverseDomain(opts.UniverseDomain, f.UniverseDomain)
-	configProvider, err := trustboundary.NewExternalAccountConfigProvider(f.Audience, ud)
+	configProvider, err := regionalaccessboundary.NewExternalAccountConfigProvider(f.Audience, ud)
 	if err != nil {
 		return nil, err
 	}
-	return trustboundary.NewProvider(opts.client(), configProvider, opts.logger(), tp)
+	return regionalaccessboundary.NewProvider(opts.client(), configProvider, opts.logger(), tp)
 }
 
 func handleImpersonatedServiceAccount(f *credsfile.ImpersonatedServiceAccountFile, opts *DetectOptions) (auth.TokenProvider, error) {
@@ -306,19 +306,19 @@ func handleImpersonatedServiceAccount(f *credsfile.ImpersonatedServiceAccountFil
 	if err != nil {
 		return nil, err
 	}
-	trustBoundaryEnabled, err := trustboundary.IsEnabled()
+	regionalAccessBoundaryEnabled, err := regionalaccessboundary.IsEnabled()
 	if err != nil {
 		return nil, err
 	}
-	if !trustBoundaryEnabled {
+	if !regionalAccessBoundaryEnabled {
 		return tp, nil
 	}
 	targetSAEmail, err := impersonate.ExtractServiceAccountEmail(f.ServiceAccountImpersonationURL)
 	if err != nil {
-		return nil, fmt.Errorf("credentials: could not extract target service account email for trust boundary: %w", err)
+		return nil, fmt.Errorf("credentials: could not extract target service account email for Regional Access Boundary: %w", err)
 	}
-	targetSAConfig := trustboundary.NewServiceAccountConfigProvider(targetSAEmail, ud)
-	return trustboundary.NewProvider(opts.client(), targetSAConfig, opts.logger(), tp)
+	targetSAConfig := regionalaccessboundary.NewServiceAccountConfigProvider(targetSAEmail, ud)
+	return regionalaccessboundary.NewProvider(opts.client(), targetSAConfig, opts.logger(), tp)
 }
 func handleGDCHServiceAccount(f *credsfile.GDCHServiceAccountFile, opts *DetectOptions) (auth.TokenProvider, error) {
 	return gdch.NewTokenProvider(f, &gdch.Options{
--- a/auth/grpctransport/grpctransport.go
+++ b/auth/grpctransport/grpctransport.go
@@ -415,7 +415,11 @@ func (c *grpcCredentialsProvider) GetRequestMetadata(ctx context.Context, uri ..
 		}
 	}
 	metadata := make(map[string]string, len(c.metadata)+1)
-	headers.SetAuthMetadata(token, metadata)
+	var reqURL string
+	if len(uri) > 0 {
+		reqURL = uri[0]
+	}
+	headers.SetAuthMetadata(token, reqURL, metadata)
 	for k, v := range c.metadata {
 		metadata[k] = v
 	}
--- a/auth/grpctransport/grpctransport.go
+++ b/auth/grpctransport/grpctransport.go
@@ -417,7 +417,7 @@ func (c *grpcCredentialsProvider) GetRequestMetadata(ctx context.Context, uri ..
 		}
 	}
 	metadata := make(map[string]string, len(c.metadata)+1)
-	headers.SetAuthMetadata(token, c.endpoint, metadata)
+	headers.SetAuthMetadata(ctx, token, c.endpoint, metadata)
 	for k, v := range c.metadata {
 		metadata[k] = v
 	}
--- a/auth/grpctransport/grpctransport.go
+++ b/auth/grpctransport/grpctransport.go
@@ -326,6 +326,7 @@ func dial(ctx context.Context, secure bool, opts *Options) (*grpc.ClientConn, er
 				creds:                creds,
 				metadata:             metadata,
 				clientUniverseDomain: opts.UniverseDomain,
+				endpoint:             transportCreds.Endpoint,
 			}),
 		)
 		// Attempt Direct Path
@@ -373,6 +374,7 @@ type grpcCredentialsProvider struct {
 	// Additional metadata attached as headers.
 	metadata             map[string]string
 	clientUniverseDomain string
+	endpoint             string
 }
 
 // getClientUniverseDomain returns the default service domain for a given Cloud
@@ -415,11 +417,7 @@ func (c *grpcCredentialsProvider) GetRequestMetadata(ctx context.Context, uri ..
 		}
 	}
 	metadata := make(map[string]string, len(c.metadata)+1)
-	var reqURL string
-	if len(uri) > 0 {
-		reqURL = uri[0]
-	}
-	headers.SetAuthMetadata(token, reqURL, metadata)
+	headers.SetAuthMetadata(token, c.endpoint, metadata)
 	for k, v := range c.metadata {
 		metadata[k] = v
 	}
--- a/auth/internal/regionalaccessboundary/external_accounts_config_providers.go
+++ b/auth/internal/regionalaccessboundary/external_accounts_config_providers.go
@@ -21,8 +21,8 @@ import (
 )
 
 const (
-	workloadAllowedLocationsEndpoint  = "https://iamcredentials.%s/v1/projects/%s/locations/global/workloadIdentityPools/%s/allowedLocations"
-	workforceAllowedLocationsEndpoint = "https://iamcredentials.%s/v1/locations/global/workforcePools/%s/allowedLocations"
+	workloadAllowedLocationsEndpoint  = "https://iamcredentials.googleapis.com/v1/projects/%s/locations/global/workloadIdentityPools/%s/allowedLocations"
+	workforceAllowedLocationsEndpoint = "https://iamcredentials.googleapis.com/v1/locations/global/workforcePools/%s/allowedLocations"
 )
 
 var (
@@ -78,7 +78,7 @@ type workforcePoolConfigProvider struct {
 }
 
 func (p *workforcePoolConfigProvider) GetRegionalAccessBoundaryEndpoint(ctx context.Context) (string, error) {
-	return fmt.Sprintf(workforceAllowedLocationsEndpoint, p.universeDomain, p.poolID), nil
+	return fmt.Sprintf(workforceAllowedLocationsEndpoint, p.poolID), nil
 }
 
 func (p *workforcePoolConfigProvider) GetUniverseDomain(ctx context.Context) (string, error) {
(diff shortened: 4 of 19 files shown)


auth/internal/regionalaccessboundary/regional_access_boundary.go
```diff
@@ -0,0 +1,458 @@
...
+
+// handleFetchFailure triggers the cooldown period using exponential backoff.
+func (p *DataProvider) handleFetchFailure(ctx context.Context) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	p.cooldownExpiry = time.Now().Add(p.cooldownDuration)
+
+	// Exponential backoff for cooldown, up to cacheTTL max (6 hours)
+	nextCooldown := p.cooldownDuration * 2
```
quartzmo, line 374: The simple doubling (`cooldownDuration * 2`) here is missing a random jitter. This could cause cyclic thundering herds on expiry if many clients fail synchronously. 

`auth/internal/retry/retry.go` uses `math/rand` with `1 + rand.Int63n(int64(b.cur))` for its jitter mechanism. We could adopt a similar approach for the cooldown expiry.

auth/credentials/detect.go
```diff
@@ -142,7 +142,7 @@ func DetectDefault(opts *DetectOptions) (*auth.Credentials, error) {
 	if err := opts.validate(); err != nil {
 		return nil, err
 	}
-	trustBoundaryEnabled, err := trustboundary.IsEnabled()
+	trustBoundaryEnabled, err := regionalaccessboundary.IsEnabled()
```
quartzmo: The local var `trustBoundaryEnabled` should also be renamed.

auth/internal/regionalaccessboundary/regional_access_boundary.go
```diff
@@ -0,0 +1,458 @@
...
+// is enabled via the GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED environment variable.
+//
+// If the environment variable is not set or empty, it is considered false.
+//
+// The environment variable is interpreted as a boolean with the following
+// (case-insensitive) rules:
+//   - "true", "1" are considered true.
+//   - All other values (including "false", "0", or invalid strings) are considered false.
+func isRegionalAccessBoundaryEnabled() (bool, error) {
+	val := strings.ToLower(os.Getenv("GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED"))
```
quartzmo, line 85: Can the `GOOGLE_AUTH_TRUST_BOUNDARY_ENABLED` env var be renamed? 

If not, can it be replaced in documentation with a new env var that uses the Regional Access Boundary name, and can that new env var be **added** to this detection helper?

auth/internal/regionalaccessboundary/external_accounts_config_providers.go
quartzmo: This code is not changed in this PR, and appears to have been adapted from `externalaccount.go`, but it can be improved:

These regex definitions should use strict `^` and `$` anchors to avoid parsing malformed strings unintentionally.

auth/internal/regionalaccessboundary/external_accounts_config_providers.go
```diff
@@ -21,8 +21,8 @@ import (
 )
 
 const (
-	workloadAllowedLocationsEndpoint  = "https://iamcredentials.%s/v1/projects/%s/locations/global/workloadIdentityPools/%s/allowedLocations"
-	workforceAllowedLocationsEndpoint = "https://iamcredentials.%s/v1/locations/global/workforcePools/%s/allowedLocations"
+	workloadAllowedLocationsEndpoint  = "https://iamcredentials.googleapis.com/v1/projects/%s/locations/global/workloadIdentityPools/%s/allowedLocations"
+	workforceAllowedLocationsEndpoint = "https://iamcredentials.googleapis.com/v1/locations/global/workforcePools/%s/allowedLocations"
```
quartzmo, line 25: Just confirming that it is OK to remove universe domain support from these endpoint templates?

auth/internal/regionalaccessboundary/regional_access_boundary.go
```diff
@@ -0,0 +1,458 @@
...
+
+	token.Metadata[ProviderKey] = p
+	return token, nil
+}
+
+// GetHeaderValue immediately returns a valid header if it's cached, or kicks off a background fetch
+// if it is unpopulated or expired.
+func (p *DataProvider) GetHeaderValue(ctx context.Context, reqURL string, accessToken *auth.Token) string {
+	// Skip lookup for regional endpoints.
+	if strings.Contains(reqURL, "rep.googleapis.com") || strings.Contains(reqURL, "rep.sandbox.googleapis.com") {
```
quartzmo: Using `strings.Contains(reqURL, "rep.googleapis.com")` over a full URL is vulnerable to query parameter injections (`https://pubsub.googleapis.com/v1/projects/...?tracking=rep.googleapis.com`).

The consequences with RAB might not be that serious, but just for good engineering practice you can instead parse `reqURL` using standard `url.Parse()` and check the `url.Host` explicitly to prevent query parameter injection.

quartzmo: Not sure if this is intentional, but I believe this `strings.Contains` check will always fail to skip for gRPC traffic. In standard gRPC transport, the `reqURL` being passed into `SetAuthMetadata` is typically just the method path ( like `/google.pubsub.v1.Publisher/CreateTopic`) rather than a full URL with hostname.

Fixing `GetHeaderValue` would require passing it the actual connection endpoint via context or some other mechanism and may be too difficult, so this is just an FYI.

auth/internal/regionalaccessboundary/external_accounts_config_providers.go
quartzmo: I think I found the answer to this: "The library will verify if the credential is using the GDU universe (i.e., googleapis.com). If the credential is for a non-GDU universe, the lookup API call will be skipped."

auth/internal/regionalaccessboundary/regional_access_boundary.go
nbayati: Thanks for the suggestion! I updated the logic to select a random number between half of the base and the full base when assigning cooldownExpiry. 
`cooldownDuration` would still be a simple doubling, but when it's time to assign the next cooldownExpiry, we will pick a new random jitter between the bounds of cooldownDuration/2 and cooldownDuration.

auth/internal/regionalaccessboundary/regional_access_boundary.go
nbayati: We are planning on removing the env var altogether when the feature is ready to be launched, as RAB lookup is supposed to be enabled by default without an opt-out option. The current ETA of the endpoint readiness is mid April, which means we can launch by end of April. With this, do you still recommend renaming the env var, or can we keep it around for now and remove it before launch so we don't have to introduce yet another env var?

auth/internal/regionalaccessboundary/external_accounts_config_providers.go
nbayati: Correct. The RAB feature is only applicable to GDU so the endpoint will always be `googleapis.com`

auth/internal/regionalaccessboundary/regional_access_boundary.go
nbayati: Thanks for flagging this! I've updated the code to pass the service URL.

quartzmo (changes requested):
```
chrisdsmith@chrisdsmith:~/projects/rab-review/repos/google-cloud-go/auth$ go test --race -run TestDataProvider_GetHeaderValue ./internal/regionalaccessboundary/...
==================
WARNING: DATA RACE
Write at 0x00c0000145a0 by goroutine 15:
  cloud.google.com/go/auth/internal/regionalaccessboundary.(*mockConfigProvider).GetRegionalAccessBoundaryEndpoint()
      /usr/local/google/home/chrisdsmith/projects/rab-review/repos/google-cloud-go/auth/internal/regionalaccessboundary/regional_access_boundary_test.go:593 +0x44
  cloud.google.com/go/auth/internal/regionalaccessboundary.(*DataProvider).fetchAsync()
      /usr/local/google/home/chrisdsmith/projects/rab-review/repos/google-cloud-go/auth/internal/regionalaccessboundary/regional_access_boundary.go:320 +0xef
  cloud.google.com/go/auth/internal/regionalaccessboundary.(*DataProvider).GetHeaderValue.gowrap3()
      /usr/local/google/home/chrisdsmith/projects/rab-review/repos/google-cloud-go/auth/internal/regionalaccessboundary/regional_access_boundary.go:307 +0x5d

Previous read at 0x00c0000145a0 by goroutine 14:
  cloud.google.com/go/auth/internal/regionalaccessboundary.TestDataProvider_GetHeaderValue.func1.1()
      /usr/local/google/home/chrisdsmith/projects/rab-review/repos/google-cloud-go/auth/internal/regionalaccessboundary/regional_access_boundary_test.go:667 +0x371
  testing.tRunner()
      /usr/lib/go-1.24/src/testing/testing.go:1792 +0x225
  testing.(*T).Run.gowrap1()
      /usr/lib/go-1.24/src/testing/testing.go:1851 +0x44

Goroutine 15 (running) created at:
  cloud.google.com/go/auth/internal/regionalaccessboundary.(*DataProvider).GetHeaderValue()
      /usr/local/google/home/chrisdsmith/projects/rab-review/repos/google-cloud-go/auth/internal/regionalaccessboundary/regional_access_boundary.go:307 +0x8ee
  cloud.google.com/go/auth/internal/regionalaccessboundary.TestDataProvider_GetHeaderValue.func1.1()
      /usr/local/google/home/chrisdsmith/projects/rab-review/repos/google-cloud-go/auth/internal/regionalaccessboundary/regional_access_boundary_test.go:661 +0x248
  testing.tRunner()
      /usr/lib/go-1.24/src/testing/testing.go:1792 +0x225
  testing.(*T).Run.gowrap1()
      /usr/lib/go-1.24/src/testing/testing.go:1851 +0x44

Goroutine 14 (running) created at:
  testing.(*T).Run()
      /usr/lib/go-1.24/src/testing/testing.go:1851 +0x8f2
  cloud.google.com/go/auth/internal/regionalaccessboundary.TestDataProvider_GetHeaderValue.func1()
      /usr/local/google/home/chrisdsmith/projects/rab-review/repos/google-cloud-go/auth/internal/regionalaccessboundary/regional_access_boundary_test.go:656 +0x244
  testing.tRunner()
      /usr/lib/go-1.24/src/testing/testing.go:1792 +0x225
  testing.(*T).Run.gowrap1()
      /usr/lib/go-1.24/src/testing/testing.go:1851 +0x44
==================
--- FAIL: TestDataProvider_GetHeaderValue (0.19s)
    --- FAIL: TestDataProvider_GetHeaderValue/Regional_endpoint_skip (0.16s)
        --- FAIL: TestDataProvider_GetHeaderValue/Regional_endpoint_skip/Injected_Query_Param (0.01s)
            testing.go:1490: race detected during execution of test
FAIL
FAIL	cloud.google.com/go/auth/internal/regionalaccessboundary	0.210s
FAIL
```

The good news is that this is a test-only race, not a production bug.

The race detector is complaining about the mockConfigProvider. The background goroutine increments a counter with m.endpointCallCount++, while the test thread is simultaneously reading it in a polling loop to check if the fetch was triggered.

I think you need to protect the call counts in the mock with a sync.Mutex and Lock/Unlock calls in the mockConfigProvider methods.

quartzmo (approved):
Great job!!!

auth/internal/regionalaccessboundary/regional_access_boundary.go
```diff
@@ -0,0 +1,473 @@
...
+	// Skip lookup for regional endpoints.
+	if !strings.Contains(reqURL, "://") {
+		reqURL = "https://" + reqURL
+	}
+	if u, err := url.Parse(reqURL); err == nil {
+		host := u.Host
+		if h, _, err := net.SplitHostPort(host); err == nil {
+			host = h
+		}
+		if host == "rep.googleapis.com" || strings.HasSuffix(host, ".rep.googleapis.com") ||
```
nbayati, line 260: I decided not to add checks against iam and sts endpoints here, mainly because:
1. the calls to iam and sts don't necessarily go through this flow. For example [here](https://github.com/googleapis/google-cloud-go/blob/b7c7bf874b7c12c128ca31bff5c00cf7d837267e/auth/credentials/internal/externalaccount/externalaccount.go#L317), the STS token exchange logic uses creates its own set of headers and uses the raw Client: tp.client to execute the request, which means it won't go through the GetHeaderValue flow.  
2. the backend will ignore the header if it's presented for an iam or sts request. Our design says we don't have to go out of our way to ensure the header is attached to all the iam and sts calls, but if it is attached, it's not an issue. 
3. it's different from the regional endpoint check, since we want to optimize by not fetching the header for regional endpoints, but if a credential is calling iam or sts, they'll be calling other endpoints next that might be regional or global. Adding a check here won't save performance, and it might make the logic brittle and hard to maintain if we have to extract the target and check against iam and sts URLs.

auth/internal/regionalaccessboundary/regional_access_boundary.go
```diff
@@ -0,0 +1,473 @@
...
+	token, err := p.base.Token(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if token.Metadata == nil {
+		token.Metadata = make(map[string]interface{})
+	}
+
+	token.Metadata[ProviderKey] = p
```
lsirac: I think we should avoid mutating the token returned by the wrapped provider here. The TokenProvider contract treats returned tokens as read-only, and some providers return cached/shared token pointers.

The GCE path looks especially risky because computeTokenProvider returns a cached provider, and compute tokens use the package-level computeTokenMetadata map. So adding ProviderKey in place can mutate shared metadata, race between concurrent callers, or leak this RAB provider pointer across otherwise separate compute credential instances.

Could we return a shallow copy of the token and a cloned Metadata map, then add ProviderKey to the cloned metadata before returning?

auth/internal/regionalaccessboundary/regional_access_boundary.go
```diff
@@ -0,0 +1,473 @@
...
+	if ud == "" {
+		ud = internal.DefaultUniverseDomain
+	}
+	g.ud = ud
+}
+
+// GetRegionalAccessBoundaryEndpoint constructs the Regional Access Boundary lookup URL for a GCE environment.
+// It uses cached service account email after the first call.
+func (g *GCEConfigProvider) GetRegionalAccessBoundaryEndpoint(ctx context.Context) (string, error) {
+	g.saOnce.Do(func() { g.fetchSA(ctx) })
```
lsirac: fetchAsync now applies cooldown when endpoint construction fails, but here a first-call EmailWithContext failure sets saEmailErr permanently. Later cooldown retries will keep returning the same cached error.

Is this intentional? Could we avoid caching failed service-account-email lookups, or only mark the email initialized after a successful lookup? Otherwise this path defeats the retry/cooldown behavior the PR adds.

auth/grpctransport/grpctransport.go
```diff
@@ -326,6 +326,7 @@ func dial(ctx context.Context, secure bool, opts *Options) (*grpc.ClientConn, er
 				creds:                creds,
 				metadata:             metadata,
 				clientUniverseDomain: opts.UniverseDomain,
+				endpoint:             transportCreds.Endpoint,
```
lsirac, line 329: I think DirectPath will still lose the endpoint RAB needs for the regional endpoint skip. 

After this, `configureDirectPath` can replace the existing options with new `grpcCredentialsProvider` instances in `directpath.go`. Those providers currently only set `creds`, so `c.endpoint` is empty when `GetRequestMetadata` calls `headers.SetAuthMetadata(token, c.endpoint, metadata)`. That means `GetHeaderValue` can't recognize `*.rep.googleapis.com` and skip `x-allowed-locations` for regional endpoints.

Can the DirectPath providers also carry the original service endpoint, ideally before adding resolver prefixes?

auth/internal/transport/headers/headers.go
```diff
@@ -31,31 +37,26 @@ func SetAuthHeader(token *auth.Token, req *http.Request) {
...
-		m["x-allowed-locations"] = headerVal
-	}
-}
-
-func getTrustBoundaryHeader(token *auth.Token) (val string, present bool) {
-	if data, ok := token.Metadata[internal.TrustBoundaryDataKey]; ok {
-		if tbd, ok := data.(internal.TrustBoundaryData); ok {
-			return tbd.TrustBoundaryHeader()
+	if provider, ok := token.Metadata[regionalaccessboundary.ProviderKey].(regionalAccessBoundaryProvider); ok {
+		if headerVal := provider.GetHeaderValue(context.Background(), reqURL, token); headerVal != "" {
```
lsirac: I think this should preserve the caller’s context instead of using context.Background().

In the gRPC path, GetRequestMetadata receives the per-RPC context, but SetAuthMetadata drops it here. GetHeaderValue synchronously calls GetUniverseDomain before it can return, and on GCE that can hit metadata under sync.Once.

So a per-RPC auth metadata call can ignore the RPC deadline or cancellation and block even though the RAB fetch path is intended to be async/fail-open.

Could SetAuthMetadata take a context and pass the gRPC request context through to GetHeaderValue?

auth/internal/regionalaccessboundary/regional_access_boundary.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+		if host == "rep.googleapis.com" || strings.HasSuffix(host, ".rep.googleapis.com") ||
+			host == "rep.sandbox.googleapis.com" || strings.HasSuffix(host, ".rep.sandbox.googleapis.com") {
+			return ""
+		}
+	}
+
+	// Skip lookup for non-default universe domains.
+	uniDomain, err := p.configProvider.GetUniverseDomain(ctx)
+	if err == nil && uniDomain != "" && uniDomain != internal.DefaultUniverseDomain {
```
lsirac: Should universe-domain lookup errors fail open by skipping RAB?

auth/internal/regionalaccessboundary/external_accounts_config_providers.go
```diff
@@ -21,13 +21,13 @@ import (
...
 )
 
 var (
-	workforceAudiencePattern = regexp.MustCompile(`//iam\.([^/]+)/locations/global/workforcePools/([^/]+)`)
-	workloadAudiencePattern  = regexp.MustCompile(`//iam\.([^/]+)/projects/([^/]+)/locations/global/workloadIdentityPools/([^/]+)`)
+	workforceAudiencePattern = regexp.MustCompile(`^//iam\.([^/]+)/locations/global/workforcePools/([^/]+)$`)
```
lsirac: This needs to include the provider portion. Otherwise it only matches `//iam.googleapis.com/projects/123/locations/global/workloadIdentityPools/my-pool`
but not `//iam.googleapis.com/projects/123/locations/global/workloadIdentityPools/my-pool/providers/my-provider`, which won't work.

Should be anchoring the full expected shape: ^//iam\.([^/]+)/projects/([^/]+)/locations/([^/]+)/workloadIdentityPools/([^/]+)/providers/([^/]+)$

Also, could we add/reuse a shared external-account audience parser instead of having a RAB-specific regex here?

auth/internal/regionalaccessboundary/external_accounts_config_providers_test.go
lsirac, line 1: These RAB parser tests are using shortened pool resource names, but external-account credential `audience` values include the provider segment, e.g. `.../workloadIdentityPools/my-pool/providers/my-provider` and `.../workforcePools/my-pool/providers/my-provider`.

We need to update this to use the full provider-resource audience strings. Otherwise the tests incorrectly pass for a shape that `filetypes.go` will not normally pass into `NewExternalAccountConfigProvider`.

auth/internal/regionalaccessboundary/regional_access_boundary_test.go
```diff
@@ -649,290 +628,151 @@ func (m *mockTokenProvider) Token(ctx context.Context) (*auth.Token, error) {
 
 func TestDataProvider_Token(t *testing.T) {
 	ctx := context.Background()
+	baseToken := &auth.Token{Value: "base-token"}
```
lsirac: Could we extend this test to assert that Token() does not mutate the token returned by the wrapped provider?

Right now this only checks that the returned token has ProviderKey, so the test would still pass if baseToken.Metadata is mutated in place.

Let's pre-populate baseToken.Metadata with a sentinel value, call provider.Token(ctx), then assert:
- the original baseToken.Metadata does not contain ProviderKey
- the returned token preserves the existing metadata entries
- the returned metadata map is not the same map as baseToken.Metadata

auth/internal/regionalaccessboundary/regional_access_boundary_test.go
```diff
@@ -649,290 +628,151 @@ func (m *mockTokenProvider) Token(ctx context.Context) (*auth.Token, error) {
...
-					t.Errorf("second run GetTrustBoundaryEndpoint call count = %d, want %d", tt.mockConfig.endpointCallCount, tt.secondRun.wantEndpointCallCount)
-				}
+		if val2 == "" {
+			t.Errorf("Second call did not return cached header within timeout")
+		} else if val2 != "0xABC" {
+			t.Errorf("Second call returned %q, expected %q", val2, "0xABC")
+		}
+	})
+
+	t.Run("Endpoint_fetch_failure_sets_cooldown", func(t *testing.T) {
```
lsirac, line 858: Let's add a GCE-specific transient-failure regression test here

auth/internal/transport/headers/headers_test.go
```diff
@@ -82,22 +79,22 @@ func TestSetAuth(t *testing.T) {
...
-			if got, ok := req.Header["X-Allowed-Locations"]; ok != tt.wantTBHeader || (ok && got[0] != tt.wantTBHeaderVal) {
-				t.Errorf("X-Allowed-Locations header: got %q, want %q (present: %v)", got, tt.wantTBHeaderVal, tt.wantTBHeader)
+			if got, ok := req.Header["X-Allowed-Locations"]; ok != tt.wantRABHeader || (ok && got[0] != tt.wantRABHeaderVal) {
+				t.Errorf("X-Allowed-Locations header: got %q, want %q (present: %v)", got, tt.wantRABHeaderVal, tt.wantRABHeader)
 			}
 
 			// Test gRPC
 			gotMeta := make(map[string]string)
-			SetAuthMetadata(token, gotMeta)
+			SetAuthMetadata(token, "https://example.com/v1", gotMeta)
```
lsirac: Could we add coverage that the caller context reaches the RAB provider in the gRPC metadata path?

The current mock provider only returns a static value, so this test would still pass if SetAuthMetadata uses context.Background().

After changing SetAuthMetadata to accept a context, this test could pass a context with a sentinel value or cancellation/deadline and assert the mock provider saw that same context. Ideally there would also be a small grpcCredentialsProvider.GetRequestMetadata test proving the per-RPC context is passed through.

auth/internal/retry/retry.go
```diff
@@ -22,10 +22,6 @@ import (
 	"time"
 )
 
-const (
-	maxRetryAttempts = 5
-)
-
 var (
 	syscallRetryable = func(error) bool { return false }
```
lsirac, line 26: This is hardcoded to always return false, so the check on line 149 in this file and transient socket errors from the iamcredentials client fall through as non-retryable on every platform.

The compute/metadata package has the same stub in compute/metadata/retry.go but pairs it with compute/metadata/retry_linux.go, which swaps in a real implementation under a build tag. 

Could we add a matching auth/internal/retry/retry_linux.go, or give the default implementation in auth/internal/retry/retry.go something real to check so the branch on line 149 is actually reachable?

auth/credentials/filetypes.go
```diff
@@ -306,19 +306,19 @@ func handleImpersonatedServiceAccount(f *credsfile.ImpersonatedServiceAccountFil
...
 	}
 	targetSAEmail, err := impersonate.ExtractServiceAccountEmail(f.ServiceAccountImpersonationURL)
 	if err != nil {
-		return nil, fmt.Errorf("credentials: could not extract target service account email for trust boundary: %w", err)
+		return nil, fmt.Errorf("credentials: could not extract target service account email for Regional Access Boundary: %w", err)
 	}
-	targetSAConfig := trustboundary.NewServiceAccountConfigProvider(targetSAEmail, ud)
-	return trustboundary.NewProvider(opts.client(), targetSAConfig, opts.logger(), tp)
+	targetSAConfig := regionalaccessboundary.NewServiceAccountConfigProvider(targetSAEmail, ud)
+	return regionalaccessboundary.NewProvider(opts.client(), targetSAConfig, opts.logger(), tp)
```
lsirac, line 321: I think this target-credential RAB wrapping can interact badly with the source credential built earlier in this function.

Above, sourceTP is built through fileCredentials(f.CredSource, &sourceOpts). If RAB is enabled, that source credential may already be wrapped with its own RAB provider. Then the impersonation provider uses the source token for the internal generateAccessToken call, and SetAuthHeader can see the source token’s ProviderKey, start a source allowedLocations lookup, and potentially attach the source x-allowed-locations header to the IAMCredentials impersonation call.

But the boundary that should matter for user resource calls is the final impersonated service account, which is what this outer wrapping handles. The source-boundary lookup/header seems like extra lookup traffic and log noise at best, and it relies on IAMCredentials ignoring an unrelated source-boundary header.

Could we suppress RAB when building the impersonation source credential, or explicitly skip RAB header injection for IAMCredentials / STS internal auth plumbing calls?

auth/internal/regionalaccessboundary/regional_access_boundary.go
nbayati: Great catch, thanks for flagging this!

auth/internal/regionalaccessboundary/regional_access_boundary.go
nbayati: Good catch, I think the idea of only trying once was from when the call was blocking. I've updated the code to make it cache the successful result, but not the error. This way it can try to get the SA again after the cooldown.

auth/grpctransport/grpctransport.go
nbayati: Done!

auth/internal/transport/headers/headers.go
nbayati: Good call, done!

auth/internal/regionalaccessboundary/regional_access_boundary.go
nbayati: Yes, fixed!

auth/internal/regionalaccessboundary/external_accounts_config_providers.go
nbayati: Ah thanks for flagging this. A recent commit had broken this flow and I hadn't realized it. 
I decided against creating a shared parser for now, because it'd require creating a new package under auth that both RAB and other external account logics could use. I'm going to have a follow up PR to enable RAB, and I'll see if I can cleanly refactor it then.

auth/internal/regionalaccessboundary/external_accounts_config_providers_test.go
nbayati: Done!

auth/internal/regionalaccessboundary/regional_access_boundary_test.go
nbayati: Done!

auth/internal/transport/headers/headers_test.go
nbayati: I have addressed your suggestions regarding context propagation in the gRPC path:

I updated SetAuthMetadata in `headers.go` to accept a context.Context and passed it through to GetHeaderValue instead of using context.Background().

I extended TestSetAuth in `headers_test.go` by passing a context with a custom key/value to SetAuthMetadata and asserting that the mock provider receives that same context.

I also added a new test, TestGrpcCredentialsProvider_GetRequestMetadata_Context, in `grpctransport_test.go`. This test creates a mock state with a custom context and verifies that GetRequestMetadata passes that per-RPC context all the way down to the RAB provider's GetHeaderValue method.

auth/internal/retry/retry.go
nbayati: Great catch, thank you!

auth/credentials/filetypes.go
nbayati: Added a iam/sts check to the `GetHeaderValue` method to skip RAB lookup.

auth/internal/regionalaccessboundary/regional_access_boundary.go
```diff
@@ -0,0 +1,500 @@
...
+	g.saMu.Lock()
+	g.saEmail = email
+	g.saMu.Unlock()
+
+	return fmt.Sprintf(serviceAccountAllowedLocationsEndpoint, email), nil
+}
+
+// GetUniverseDomain retrieves the universe domain from the GCE metadata server.
+// It uses a cached value after the first call.
+func (g *GCEConfigProvider) GetUniverseDomain(ctx context.Context) (string, error) {
```
lsirac, line 497: I think this will permanently cache transient universe-domain lookup failures via `udOnce` + `udErr`.

That means if the first metadata universe-domain lookup fails transiently, every later `GetHeaderValue` call will keep failing open forever for this provider, even after metadata recovers. This is the same as the earlier `saOnce`/`saEmailErr` issue that was fixed for service-account email lookup.

auth/internal/regionalaccessboundary/regional_access_boundary.go
```diff
@@ -0,0 +1,500 @@
...
+	return &newToken, nil
+}
+
+// GetHeaderValue immediately returns a valid header if it's cached, or kicks off a background fetch
+// if it is unpopulated or expired.
+func (p *DataProvider) GetHeaderValue(ctx context.Context, reqURL string, accessToken *auth.Token) string {
+	if !strings.Contains(reqURL, "://") {
+		reqURL = "https://" + reqURL
+	}
+	if u, err := url.Parse(reqURL); err == nil {
```
lsirac, line 258: Could we add a small regression test for resolver-scheme endpoints, since url.Parse leaves Host empty there?

e.g. see endpoints in https://pkg.go.dev/google.golang.org/grpc#NewClient

lsirac (approved):
Great work!

auth/internal/regionalaccessboundary/regional_access_boundary.go
nbayati: You're right. I think we have two options, either change the underlying method that fetches the UD not to permanently cache failures, or write our own code to fetch it. I need to see why it was written that way in the internal method, and will address it in my follow up PR.

auth/internal/regionalaccessboundary/regional_access_boundary.go
nbayati: Done!

PR #13419, opened 2025-12-01
feat(bigtable): introduce prime() method and load counting based on unary,streaming and ability to remove connection

sushanb:
no-op from https://github.com/sushanb/google-cloud-go/pull/1

### DIFF
--- a/bigtable/internal/transport/connpool_test.go
+++ b/bigtable/internal/transport/connpool_test.go
@@ -635,7 +635,7 @@ func TestSelectLeastLoadedRandomOfTwo(t *testing.T) {
 }
 
 func TestCachingStreamDecrement(t *testing.T) {
-  t.Skip("skipped as impacting generation https://github.com/googleapis/google-cloud-go/issues/13383")
+	t.Skip("skipped as impacting generation https://github.com/googleapis/google-cloud-go/issues/13383")
 	ctx := context.Background()
 	poolSize := 1
 	fake := &fakeService{}


bigtable/internal/transport/connpool.go
```diff
@@ -16,53 +16,244 @@ package internal
...
+	primeRPCTimeout = 10 * time.Second
+)
+
 var errNoConnections = fmt.Errorf("bigtable_connpool: no connections available in the pool")
 var _ gtransport.ConnPool = &BigtableChannelPool{}
 
+// BigtableConn wraps grpc.ClientConn to add Bigtable specific methods.
+type BigtableConn struct {
+	*grpc.ClientConn
+	isALTSConn atomic.Bool
```
igorbernstein2, line 97: I think you want to tristate this - you are currently handling 2 cases:
- prime was successful over alts
- prime was successful over tls/cloudpath

But you are missing the case that the channel was dead and was not able to send the prime rpc. Which will result in failed channels coming up as cloud path failing in metrics. I think capturing that the priming response was missing AuthInfo is useful to be able to set the transport_type to unknown

bigtable/internal/transport/connpool.go
```diff
@@ -16,53 +16,244 @@ package internal
...
+	streamingLoad atomic.Int32 // In-flight streaming requests
+	errorCount    atomic.Int64 // Errors since the last metric report
+	drainingState atomic.Bool  // True if the connection is being gracefully drained.
+
+}
+
+// isALTSUsed reports whether the connection is using ALTS aka Direct Access.
+// best effort basis
+func (e *connEntry) isALTSUsed() bool {
+	if e.conn == nil {
```
igorbernstein2, line 152: please add a comment noting when e.conn is nil

igorbernstein2 (approved):

bigtable/internal/transport/connpool.go
sushanb: I will address this in another PR

bigtable/internal/transport/connpool.go
sushanb: I will address this in another PR

PR #13457, opened 2025-12-09
feat(pubsub/v2): add keep alive support

hongalex:
This change introduces a new communication protocol during streaming pull, where the client library will send periodic pings to the server (already happening) and the client library will check for periodic pings from the server.

This necessitates a second background goroutine that periodically monitors the server for pings. The previous client library behavior of sending pings to the server is refactored into this goroutine (originally in the `it.sender` function that handles message leases). 

If both server and client pings are unsuccessful after a period of time, we will assume the stream is broken and we will close and reopen them.

### DIFF
--- a/pubsub/v2/iterator_test.go
+++ b/pubsub/v2/iterator_test.go
@@ -802,6 +802,12 @@ func TestExactlyOnceProcessRequests(t *testing.T) {
 // When the server doesn't respond with keep alive pings,
 // test that the client reopens the stream.
 func TestStreamingPullKeepAlive_ReopenStream(t *testing.T) {
+	// save old variables
+	oldClient := clientPingInterval
+	oldServerPing := serverPingTimeoutDuration
+	oldServerMonitor := serverMonitorInterval
+	oldProtocol := protocolVersion
+
 	clientPingInterval = 1 * time.Second
 	// any ping check should result in stream closure
 	serverPingTimeoutDuration = 0 * time.Second
@@ -814,6 +820,13 @@ func TestStreamingPullKeepAlive_ReopenStream(t *testing.T) {
 	// since it is assumed that newer versions of the client library
 	// will always be communicating with the latest protocol version.
 	protocolVersion = 0
+
+	t.Cleanup(func() {
+		clientPingInterval = oldClient
+		serverPingTimeoutDuration = oldServerPing
+		serverMonitorInterval = oldServerMonitor
+		protocolVersion = oldProtocol
+	})
 
 	c, srv := newFake(t)
 	ctx, cancel := context.WithCancel(t.Context())


pubsub/v2/iterator.go
```diff
@@ -61,6 +61,12 @@ var (
 
 	// The total amount of time to retry acks/modacks with exactly once delivery enabled subscriptions.
 	exactlyOnceDeliveryRetryDeadline = 600 * time.Second
+
+	// specifies the protocol to communicate with the Pub/Sub servers for streaming pull.
+	// this value cannot be set by users, and should be monotonically increasing.
+	clientPingInterval        = time.Duration(30 * time.Second)
+	serverMonitorInterval     = time.Duration(10 * time.Second)
+	serverPingTimeoutDuration = time.Duration(15 * time.Second)
```
shollyman: These defaults look a little odd to me, but I don't have much experience with pingback health monitoring under load.  Is the timeout being 1.5x the interval too aggressive?

Also, will these be user tunable down the line?

pubsub/v2/pullstream.go
```diff
@@ -20,13 +20,16 @@ import (
...
 	"sync"
+	"sync/atomic"
 	"time"
 
 	pb "cloud.google.com/go/pubsub/v2/apiv1/pubsubpb"
 	gax "github.com/googleapis/gax-go/v2"
 	"google.golang.org/grpc"
 )
 
+const protocolVersion int64 = 1
```
shollyman: There's several references to this new protocol version but I'm not clear what it is.  Is this IBV or something else?  The comments here don't communicate it well.

go.work.sum
shollyman, line 1: Get this workspace stuff outta heeeere! :)

shollyman (approved):
Mostly minor feedback, otherwise LGTM

pubsub/v2/iterator.go
hongalex: The client ping default matches what we had before, so this isn't new information.

The server monitoring values match the [Java PR](https://github.com/googleapis/java-pubsub/blob/4e59581dfd648faf00141cd7f6cc766ce313d4c7/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java#L137). The idea is that we try to recreate the stream if we haven't received a response from the server in a while. Receiving messages from Pub/Sub resets this timer, and receiving the periodic 30s pings from the server also reset this.

None of these are user adjustable.

go.work.sum
hongalex: so much nicer 👍

pubsub/v2/pullstream.go
hongalex: It is IBV, but it isn't exposed to clients. When upgrading to a higher version of the client library, the client library is able to negotiate with the server which streaming pull protocol to use (in this case, the expectation of sending and receiving pings).

hongalex: Added comment to this to clarify this isn't settable, and is used for the client library / server to negotiate which behavior to use.

SamuelSchaub: It should be settable by the user based on the [specification](https://goto.google.com/streamingpull-keepalive)

hongalex: The original spec was ambiguous, but since Java also removed the ability to set protocol version, I'm inclined to leave it out.

I believe that end-users shouldn't need to care about protocol version. We are adding this so that the library can communicate with a newer streaming pull protocol in a backwards compatible way (the server knows we are using client library version >1.x based on this value. Users shouldn't need to disable this keep alive feature. If we introduce a new feature in the future that users may want control over, we can consider exposing protocol versions.

pubsub/v2/iterator.go
```diff
@@ -139,42 +152,45 @@ func newMessageIterator(subc *vkit.SubscriptionAdminClient, subName string, po *
...
+		ps:                  ps,
+		po:                  po,
+		subc:                subc,
+		projectID:           projectID,
+		subID:               subID,
+		subName:             subName,
+		kaTick:              time.After(keepAlivePeriod),
+		ackTicker:           ackTicker,
+		nackTicker:          nackTicker,
+		pingTicker:          time.NewTicker(clientPingInterval),
```
shollyman: minor nit:  It looks slightly odd that some tickers are constructed here and some are constructed above outside of the iterator.

I also wonder if ticker consolidation might be warranted:  client ping is currently defined as 3x server ping, so you could just handle client ping every N server pings?  You'd likely still need the same number of goroutines so I don't know a refactor would get you much more than getting rid of a channel.

pubsub/v2/iterator.go
```diff
@@ -861,12 +874,64 @@ func (it *messageIterator) retryModAcks(m map[string]*AckResult, deadlineSec int
 	}
 }
 
+// streamKeepAliveHandler sends to and handles responses from the stream
+// to maintain stream aliveness. We send pings to the server on a timer
+// and monitor for server pings on a timer. When both time out, then
+// we will close the stream and attempt to reopen.
+// This is unrelated to iterator.keepAliveDeadlines which handles
+// message leases keep alives.
+func (it *messageIterator) streamKeepAliverHandler() {
```
shollyman: nit:  Aliver might be a typo

pubsub/v2/iterator.go
hongalex: Moved ticker construction to same place.

I think for simplicity, the pings should be separate, even if a bit verbose. It's easier for someone to read that Client pings have happen every 30 seconds versus trying to read the logic of the server ping.

PR #13461, opened 2025-12-10
chore(storage): add configurations and defaults for PCU feature

cpriti-os:


storage/pcu.go
```diff
@@ -0,0 +1,199 @@
...
+// DefaultNamingStrategy provides a default implementation for naming temporary parts.
+type DefaultNamingStrategy struct{}
+
+var randSrc = rand.New(rand.NewSource(time.Now().UnixNano()))
+
+// NewPartName creates a unique name for a temporary part to avoid hotspotting.
+func (d *DefaultNamingStrategy) NewPartName(bucket, prefix, finalName string, partNumber int) string {
+	rnd := randSrc.Uint64()
+	return path.Join(prefix, fmt.Sprintf("%s-part-%d-%x", finalName, partNumber, rnd))
+}
```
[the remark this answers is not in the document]
cpriti-os: used thread safe generator

storage/pcu.go
```diff
@@ -0,0 +1,210 @@
...
+	"fmt"
+	"math/rand"
+	"path"
+	"runtime"
+	"sync"
+	"time"
+)
+
+// ParallelUploadConfig holds configuration for Parallel Composite Uploads.
+// Setting this config on a Writer enables PCU.
```
krishnamd-jkp: Can we have a flag in Writer instead similar to Append because accidentally setting this config to empty config could also trigger a PCU even if it is not intended?

storage/pcu.go
```diff
@@ -0,0 +1,210 @@
...
+	// NamingStrategy provides a strategy for naming temporary part objects.
+	// Defaults to a strategy that includes a random element to avoid hotspotting.
+	NamingStrategy PartNamingStrategy
+
+	// MetadataDecorator allows adding custom metadata to temporary part objects.
+	MetadataDecorator PartMetadataDecorator
+}
+
+// PartRetryPolicy defines the retry behavior for uploading parts.
+type PartRetryPolicy struct {
```
krishnamd-jkp: Add a RetryPolicy as well like this https://github.com/googleapis/google-cloud-go/blob/main/storage/storage.go#L2506?

storage/pcu.go
```diff
@@ -0,0 +1,210 @@
...
+func (s PartCleanupStrategy) String() string {
+	switch s {
+	case CleanupAlways:
+		return "always"
+	case CleanupOnSuccess:
+		return "on_success"
+	case CleanupNever:
+		return "never"
+	default:
+		return fmt.Sprintf("PartCleanupStrategy(%d)", s)
```
krishnamd-jkp, line 94: Is this supposed to be an error?

storage/pcu.go
```diff
@@ -0,0 +1,210 @@
...
+type ParallelUploadConfig struct {
+	// MinSize is the minimum size of an object in bytes to use PCU.
+	// If an object's size is less than this value, a simple upload is performed.
+	// If this is not set, a default of 64 MiB will be used.
+	// To enable PCU for all uploads regardless of size, set this to 0.
+	MinSize *int64
+
+	// PartSize is the size of each part to be uploaded in parallel.
+	// Defaults to 16MiB. Must be a multiple of 256KiB.
+	PartSize int64
```
krishnamd-jkp, line 43: Why not make this (and others) as pointer as well similar to MinSize? Would be easier to check for default

storage/pcu.go
```diff
@@ -0,0 +1,210 @@
...
+	baseWorkers            = 4
+	maxWorkers             = 16
+	defaultTmpObjectPrefix = "gcs-go-sdk-pcu-tmp/"
+	maxComposeComponents   = 32
+)
+
+func (c *ParallelUploadConfig) defaults() {
+	if c.MinSize == nil {
+		c.MinSize = new(int64)
+		*c.MinSize = int64(64 * 1024 * 1024) // 64 MiB
```
krishnamd-jkp: Add these constants also to above constants group

storage/pcu.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	if c.BufferPoolSize == 0 {
+		c.BufferPoolSize = c.NumWorkers + 1
+	}
+	if c.TmpObjectPrefix == "" {
+		c.TmpObjectPrefix = defaultTmpObjectPrefix
+	}
+	if c.RetryPolicy == nil {
+		c.RetryPolicy = &PartRetryPolicy{
+			MaxRetries: 3,
```
krishnamd-jkp: Same here. Make these part of the above constants group

storage/pcu_test.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		if cfg.CleanupStrategy != CleanupAlways {
+			t.Errorf("CleanupStrategy default is incorrect. got %v, want %v", cfg.CleanupStrategy, CleanupAlways)
+		}
+		if _, ok := cfg.NamingStrategy.(*DefaultNamingStrategy); !ok {
+			t.Errorf("NamingStrategy default is incorrect. got %T, want %T", cfg.NamingStrategy, &DefaultNamingStrategy{})
+		}
+	})
+
+	t.Run("no overrides", func(t *testing.T) {
```
krishnamd-jkp: Isn't this test overriding the defaults?

storage/pcu_test.go
```diff
@@ -0,0 +1,171 @@
...
+	if parsedPartNum != partNumber {
+		t.Errorf("NewPartName() did not include the correct part number. Got %d, want %d", parsedPartNum, partNumber)
+	}
+
+	if randSuffix == 0 {
+		t.Errorf("NewPartName() did not include a non-zero random hex part. Got %x", randSuffix)
+	}
+}
+
+func TestParallelUploadConfig_defaults(t *testing.T) {
```
krishnamd-jkp, line 82: Can we change this to Table driven test since the conditions are all duplicated?

storage/pcu.go
cpriti-os: Sounds good, will add the flag when making changes to Writer

storage/pcu.go
cpriti-os: This is intentional. Returning a formatted string with the integer value is a common pattern in Go for String() methods on enums. It makes the function robust against unknown values and provides useful information for debugging without causing a panic

storage/pcu.go
cpriti-os: The reason MinSize is a pointer is because its the only case where 0 is a valid choice and cannot be used to override with default. 
The others however should not have zero so will be overriden by the default value. Keeping others as values makes the API simpler for our users and aligns with general Go principles of preffering value over pointers whenever possible.

storage/pcu_test.go
cpriti-os: Done

storage/pcu_test.go
cpriti-os: it meant should not override user configuration, used a better name

storage/pcu.go
cpriti-os: done

storage/pcu.go
cpriti-os: done

storage/pcu.go
cpriti-os: done

PR #13481, opened 2025-12-15
fix(storage): optimize gRPC writer with zero-copy and lazy allocation

cpriti-os:
This change eliminates unnecessary double-buffering in the gRPC writer,
reducing memory usage by up to 50% for Parallel Composite Uploads (PCU)
and large single-shot uploads.

Key changes:
- Implements a Zero-Copy path in `gRPCWriter`: if in one-shot mode, the writer now
  passes the user's memory directly to the gRPC transport layer instead
  of copying it into an internal buffer.
- Implements Lazy Allocation: The internal buffer (`w.buf`) is now
  initialized to `nil` and is only allocated if the "Slow Path" (buffering)
  is triggered. This allows purely zero-copy workloads to run with
  near-zero overhead.
- Adds `chunkSize` to `gRPCWriter` to support logic when `w.buf` is nil.
- Adds comprehensive tests verifying memory addresses to ensure the
  optimization is active and strictly correct during retries.

### DIFF
--- a/storage/grpc_writer_test.go
+++ b/storage/grpc_writer_test.go
@@ -192,11 +192,10 @@ func TestGRPCWriter_MemoryAllocationPaths(t *testing.T) {
 			}
 
 			// Verify memory address logic:
-			// The last byte of the request buffer should match the last byte of the input data for zero-copy.
+			// The last byte of the last request buffer should match the last byte of the input data for zero-copy.
+			// For buffering/copying, the pointers must differ.
 			idx := len(reqs) - 1
 			bufIdx := len(reqs[idx].buf)
-			// For zero-copy, the underlying array pointer of the request buffer must match the input data.
-			// For buffering/copying, the pointers must differ.
 			isZeroCopy := &reqs[idx].buf[bufIdx-1] == &data[tt.dataSize-1]
 			if isZeroCopy != tt.wantZeroCopy {
 				if tt.wantZeroCopy && tt.forceOneShot {


storage/grpc_writer_test.go
```diff
@@ -99,3 +102,160 @@ func TestGetObjectChecksums(t *testing.T) {
...
+		if &reqs[0].buf[0] != &data[0] {
+			t.Errorf("One-shot upload bypassed zero-copy path; data was copied")
+		}
+	})
+
+	t.Run("Resumable_Buffering", func(t *testing.T) {
+		chunkSize := 256 * 1024
+		mockSender := &mockZeroCopySender{}
+		w := &gRPCWriter{
+			buf:           make([]byte, chunkSize), // Buffered path requires pre-allocated buffer
```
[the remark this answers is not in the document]
cpriti-os: Looking into this

storage/grpc_writer.go
```diff
@@ -373,21 +375,26 @@ func (w *gRPCWriter) gatherFirstBuffer() error {
...
-				// which fit in one chunk, even though we will cut the request into
-				// w.writeQuantum units when we do start sending.
-				origLen := len(w.buf)
-				w.buf = w.buf[:origLen+len(v.p)]
-				copy(w.buf[origLen:], v.p)
-				close(v.done)
-			} else {
-				// Too large. Handle it in writeLoop.
+			// If zero-copy one-shot is requested, OR the payload is larger than the buffer,
+			// bypass staging entirely and hand off to the writeLoop immediately.
```
krishnamd-jkp: nit: can we use "buffering" instead of "staging" for better clarity?

storage/grpc_writer.go
```diff
@@ -568,6 +577,23 @@ func (c *gRPCWriterCommandWrite) handle(w *gRPCWriter, cs gRPCWriterCommandHandl
 		return nil
 	}
 
+	// Zero-Copy send.
+	if w.forceOneShot {
+		done, err := c.zeroCopyWrite(w, cs)
```
krishnamd-jkp: nit: If this is used only for single shot uploads, can we change the function name to reflect it?

storage/grpc_writer.go
```diff
@@ -691,6 +717,76 @@ func (c *gRPCWriterCommandWrite) handle(w *gRPCWriter, cs gRPCWriterCommandHandl
...
+			return false, w.preRunCtx.Err()
+		}
+	}
+
+	c.p = nil
+	return c.markDone(), nil
+}
+
+// Helper to ensure we don't close done twice and keep the main logic clean.
+func (c *gRPCWriterCommandWrite) markDone() bool {
```
krishnamd-jkp: Can we just add a bool to `gRPCWriterCommandWrite` to check if it is done instead of checking by receiving on a closed channel?

storage/grpc_writer.go
```diff
@@ -691,6 +717,76 @@ func (c *gRPCWriterCommandWrite) handle(w *gRPCWriter, cs gRPCWriterCommandHandl
...
+	// If we've already sent everything, we're done.
+	if skip >= int64(len(c.p)) {
+		return c.markDone(), nil
+	}
+
+	pending := c.p[skip:]
+
+	// Pre-emptively get the context channel to avoid closure overhead in the loop.
+	var ctxDone <-chan struct{}
+	if w.preRunCtx != nil {
```
krishnamd-jkp: Is this check necessary?

storage/grpc_writer.go
```diff
@@ -691,6 +717,76 @@ func (c *gRPCWriterCommandWrite) handle(w *gRPCWriter, cs gRPCWriterCommandHandl
...
+	if skip < 0 {
+		skip = 0
+	}
+
+	// If we've already sent everything, we're done.
+	if skip >= int64(len(c.p)) {
+		return c.markDone(), nil
+	}
+
+	pending := c.p[skip:]
```
krishnamd-jkp: Why can't we override c.p here directly instead of a new variable?

storage/grpc_writer.go
```diff
@@ -691,6 +717,76 @@ func (c *gRPCWriterCommandWrite) handle(w *gRPCWriter, cs gRPCWriterCommandHandl
...
+		case completion, ok := <-cs.completions:
+			if !ok {
+				return false, w.streamSender.err()
+			}
+			w.handleCompletion(completion)
+		case <-cs.requestAcks:
+			ackOutstanding = false
+		case <-ctxDone:
+			return false, w.preRunCtx.Err()
+		}
```
krishnamd-jkp, line 756: Doesn't the sender ensure this? Is this redundant?

cpriti-os: No, the sender doesn't, without the write will return while the data is still being sent and the user might overwrite the buffer.

storage/grpc_writer.go
cpriti-os: Done

storage/grpc_writer.go
cpriti-os: I think usage should not be reflected in a function name, the function name correctly indicates what the function does and not when its called.

storage/grpc_writer.go
cpriti-os: We can, done

storage/grpc_writer.go
cpriti-os: Yes, not required, removed

storage/grpc_writer.go
cpriti-os: Bool would have to protected using a mutex which adds complexity. Simplified the code by using sync.Once

krishnamd-jkp: Aren't writes synchronous?

storage/grpc_writer_test.go
```diff
@@ -99,3 +102,162 @@ func TestGetObjectChecksums(t *testing.T) {
 		})
 	}
 }
+func TestGRPCWriter_MemoryAllocationPaths(t *testing.T) {
+	dataSize := 1024 * 1024 // 1 MB payload
+	data := make([]byte, dataSize)
+	data[0] = 1
+	data[dataSize-1] = 2
+
+	t.Run("OneShot_ZeroCopy", func(t *testing.T) {
```
krishnamd-jkp: Any chance of making this a table driven format?

cpriti-os: Done

storage/grpc_writer_test.go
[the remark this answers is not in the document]
cpriti-os: Done

krishnamd-jkp:
Approving based on the changes. Just check if the synchronous nature of the single shot write does not qualify for a breaking change.

storage/grpc_writer.go
```diff
@@ -565,17 +573,33 @@ type gRPCWriterCommand interface {
 }
 
 type gRPCWriterCommandWrite struct {
-	p    []byte
-	done chan struct{}
+	p             []byte
+	done          chan struct{}
+	initialOffset int64
+	hasStarted    bool
+	once          sync.Once
```
tritone: Perhaps give this a more descriptive name (`closeOnce` maybe?)

storage/grpc_writer.go
```diff
@@ -695,10 +719,73 @@ func (c *gRPCWriterCommandWrite) handle(w *gRPCWriter, cs gRPCWriterCommandHandl
...
+}
+
+func (c *gRPCWriterCommandWrite) zeroCopyWrite(w *gRPCWriter, cs gRPCWriterCommandHandleChans) error {
+	if !c.hasStarted {
+		c.initialOffset = w.bufBaseOffset
+		c.hasStarted = true
+	}
+
+	// Calculate the offset delta. If w.bufBaseOffset > c.initialOffset,
+	// the server persisted data from a previous attempt; we must skip those bytes.
```
tritone: How is this relevant with ChunkSize=0, which doesn't have any retries?

storage/grpc_writer_test.go
```diff
@@ -99,3 +102,149 @@ func TestGetObjectChecksums(t *testing.T) {
 		})
 	}
 }
+func TestGRPCWriter_MemoryAllocationPaths(t *testing.T) {
+	dataSize := 1024 * 1024 // 1 MB payload
```
tritone: It might be good to test with some larger payloads (larger than the msg size of 2 MiB).

storage/grpc_writer.go
cpriti-os: Agreed, this is not necessary anymore, earlier this was supposed to support data > chunkSize in resumable writes. Removed.

storage/grpc_writer.go
cpriti-os: Done

storage/grpc_writer_test.go
cpriti-os: Agreed, but there's a catch, even for resumable writes where dataSize > chunkSize, we use the user buffer directly to send the data (fullBufs) and use the buffer only for the tail IIUC. So if I increase the the data size I cannot test the buffering accurately.

tritone: Perhaps you can have tests for multiple sizes then?

tritone (approved):

storage/grpc_writer_test.go
cpriti-os: Thanks, added.

PR #13483, opened 2025-12-15
test(bigquery): decouple legacy sql view testing from public dataset

shollyman:
Fixes: https://github.com/googleapis/google-cloud-go/issues/13474

### DIFF
--- a/bigquery/table_integration_test.go
+++ b/bigquery/table_integration_test.go
@@ -617,7 +617,7 @@ func TestIntegration_TableUseLegacySQL(t *testing.T) {
 		if err := view.Create(ctx, tm); err != nil {
 			t.Errorf("view creation for %q failed with legacy %t", view.FullyQualifiedName(), isLegacy)
 		}
-		view.Delete(ctx)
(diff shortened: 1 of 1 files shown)


bigquery/table_integration_test.go
```diff
@@ -608,21 +608,14 @@ func TestIntegration_TableUseLegacySQL(t *testing.T) {
 		t.Skip("Integration tests skipped")
 	}
 	ctx := context.Background()
-	table := newTable(t, schema)
-	defer table.Delete(ctx)
-	for i, test := range useLegacySQLTests {
```
bhshkh, line 613: Should TestIntegration_QueryUseLegacySQL be updated as well? 
That testcase uses useLegacySQLTests too.

shollyman: That test runs queries in the correct region with the colocated sample datasets.  I'm choosing to leave that in for the time being, but we'll clean that up down the line.

bhshkh (approved):

PR #13489, opened 2025-12-16
chore(storage): add initial framework for PCU core logic

cpriti-os:
This initial implementation includes:
- The `pcuState` struct and configuration for managing the upload process.
- The `initPCU` function to initialize the worker pool and buffers.
- The core `worker` logic to process upload tasks from a channel.
- The `uploadPart` function, which uploads a single data part to a temporary GCS object.

### DIFF
--- a/storage/pcu.go
+++ b/storage/pcu.go
@@ -304,7 +304,6 @@ func (s *pcuState) uploadPart(task uploadTask) (*ObjectHandle, *ObjectAttrs, err
 	pw.ChunkSize = 0 // Force single-shot upload for parts.
 	// Clear fields not applicable to parts or that are set by compose.
 	pw.ObjectAttrs.CRC32C = 0
-	pw.SendCRC32C = false
 	pw.ObjectAttrs.MD5 = nil
 	setPartMetadata(pw, s, task)
 
@@ -331,7 +330,7 @@ func setPartMetadata(pw *Writer, s *pcuState, task uploadTask) {
 	}
 	pw.ObjectAttrs.Metadata = md
 	pw.ObjectAttrs.Metadata[pcuPartNumberMetadataKey] = partNumberStr
-	pw.ObjectAttrs.Metadata[pcuFinalObjectMetadataKey] = pw.o.object
+	pw.ObjectAttrs.Metadata[pcuFinalObjectMetadataKey] = s.w.o.object
 	if s.config.metadataDecorator != nil {
 		s.config.metadataDecorator.Decorate(&pw.ObjectAttrs)
 	}
--- a/storage/pcu_test.go
+++ b/storage/pcu_test.go
@@ -523,10 +523,9 @@ func TestSetPartMetadata(t *testing.T) {
 				ObjectAttrs: ObjectAttrs{
 					Metadata: tc.initialMetadata,
 				},
+				o: &ObjectHandle{bucket: "my-bucket", object: tc.finalObjectName},
 			}
-			partWriter := &Writer{
-				o: &ObjectHandle{object: tc.finalObjectName},
-			}
+			partWriter := &Writer{}
 			state := &pcuState{
 				w: sourceWriter,
 				config: &parallelUploadConfig{


storage/pcu.go
```diff
@@ -201,3 +208,129 @@ type uploadResult struct {
...
+			s.resultCh <- uploadResult{partNumber: task.partNumber, err: s.ctx.Err()}
+			s.bufferCh <- task.buffer
+			return
+		default:
+			handle, attrs, err := s.uploadPartFn(s, task)
+			s.resultCh <- uploadResult{partNumber: task.partNumber, obj: attrs, handle: handle, err: err}
+			s.bufferCh <- task.buffer
+		}
+	}
+}
```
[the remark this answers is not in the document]
cpriti-os: Done

storage/pcu.go
```diff
@@ -201,3 +208,129 @@ type uploadResult struct {
...
+	// Clear fields not applicable to parts or that are set by compose
+	pw.ObjectAttrs.CRC32C = 0
+	pw.ObjectAttrs.MD5 = nil
+
+	// Apply decorators.
+	pw.ObjectAttrs.Metadata[xGoogMetaGcsPCUPartNumber] = strconv.Itoa(task.partNumber)
+	pw.ObjectAttrs.Metadata[xGoogMetaGcsPCUFinalObject] = s.w.o.object
+	if s.config.MetadataDecorator != nil {
+		s.config.MetadataDecorator.Decorate(&pw.ObjectAttrs)
+	}
```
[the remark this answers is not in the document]
cpriti-os: Done

storage/pcu.go
```diff
@@ -115,15 +117,18 @@ type PartMetadataDecorator interface {
...
+	defaultMinSize             = 64 * 1024 * 1024 // 64 MiB
+	baseWorkers                = 4
+	maxWorkers                 = 16
+	defaultTmpObjectPrefix     = "gcs-go-sdk-pcu-tmp/"
+	maxComposeComponents       = 32
+	defaultMaxRetries          = 3
+	defaultBaseDelay           = 100 * time.Millisecond
+	defaultMaxDelay            = 5 * time.Second
+	chunkSizeMultiple          = 256 * 1024 // 256 KiB
+	xGoogMetaGcsPCUPartNumber  = "x-goog-meta-gcs-pcu-part-number"
```
krishnamd-jkp: Can we change the variable name to it's functionality rather than the header name itself?

storage/pcu.go
```diff
@@ -201,3 +209,164 @@ type uploadResult struct {
...
 }
+
+func (w *Writer) initPCU(ctx context.Context) error {
+	// TODO: Check if PCU is enabled on the Writer
+
+	// TODO: Get the config from the Writer
+	cfg := &ParallelUploadConfig{}
+	cfg.defaults()
+
+	if cfg.PartSize%(chunkSizeMultiple) != 0 {
```
krishnamd-jkp: Can we instead fall back to the nearest multiple of 256KiB like we do for chunkSize in gRPC writer? We can have the documentation inform the same.

storage/pcu.go
```diff
@@ -201,3 +209,164 @@ type uploadResult struct {
...
+	}
+	// TODO: Assign the state to the Writer
+
+	for i := 0; i < cfg.BufferPoolSize; i++ {
+		state.bufferCh <- make([]byte, cfg.PartSize)
+	}
+
+	state.workerWG.Add(cfg.NumWorkers)
+	for i := 0; i < cfg.NumWorkers; i++ {
+		go state.worker()
```
krishnamd-jkp, line 246: nit: can you add 1 to the wg inside the loop instead? More readable this way?

storage/pcu.go
```diff
@@ -201,3 +209,164 @@ type uploadResult struct {
...
+	}
+
+	state.collectorWG.Add(1)
+	go state.resultCollector()
+
+	// Handle to get the first buffer
```
krishnamd-jkp: nit: Missing period.

storage/pcu.go
```diff
@@ -201,3 +209,164 @@ type uploadResult struct {
...
+				s.resultCh <- uploadResult{partNumber: t.partNumber, obj: attrs, handle: handle, err: err}
+			}(task)
+		}
+	}
+}
+
+// TODO: add retry logic
+func (s *pcuState) uploadPart(task uploadTask) (*ObjectHandle, *ObjectAttrs, error) {
+	partName := s.config.NamingStrategy.NewPartName(s.w.o.bucket, s.config.TmpObjectPrefix, s.w.o.object, task.partNumber)
+	partHandle := s.w.o.c.Bucket(s.w.o.bucket).Object(partName)
```
krishnamd-jkp, line 298: Can we have the writer called as "writer" instead of "w"? This is not readable.

storage/pcu.go
```diff
@@ -201,3 +209,164 @@ type uploadResult struct {
...
+
+	pw := partHandle.NewWriter(s.ctx)
+	pw.ObjectAttrs.Name = partName
+	pw.ObjectAttrs.Size = task.size
+	pw.SendCRC32C = s.w.SendCRC32C
+	pw.ChunkSize = 0 // Force single-shot upload for parts
+	// Clear fields not applicable to parts or that are set by compose
+	pw.ObjectAttrs.CRC32C = 0
```
krishnamd-jkp, line 306: If CRC32C of the part is set to 0, isn't it also necessary not to sendCRC32C? like 
```pw.SendCRC32C = false```?

File: storage/pcu.go [hunk 9]
```diff
@@ -201,3 +209,164 @@ type uploadResult struct {
...
+	pw.ObjectAttrs.MD5 = nil
+	setPartMetadata(pw, s, task)
+
+	n, err := pw.Write(task.buffer[:task.size])
+	if err != nil {
+		pw.CloseWithError(err)
+		return nil, nil, fmt.Errorf("failed to write part %d: %w", task.partNumber, err)
+	}
+	if int64(n) != task.size {
```
krishnamd-jkp: I think this is redundant. If writer's n is not task.size, err will definitely be populated. If not, there's some issue with the underlying writer itself.

storage/pcu.go
```diff
@@ -201,3 +209,164 @@ type uploadResult struct {
...
+	case <-state.ctx.Done():
+		return state.ctx.Err()
+	case state.currentBuffer = <-state.bufferCh:
+		state.bytesBuffered = 0
+	}
+	state.started = true
+	return nil
+}
+
+func (s *pcuState) worker() {
```
krishnamd-jkp, line 265: Can you add a small description for this function explaining what it is doing? Since this is one part of the core logic?

storage/pcu.go
```diff
@@ -201,3 +209,164 @@ type uploadResult struct {
...
+				case <-s.ctx.Done():
+					s.resultCh <- uploadResult{partNumber: t.partNumber, err: s.ctx.Err()}
+					return
+				default:
+				}
+
+				handle, attrs, err := s.uploadPartFn(s, t)
+
+				// After the upload, check if the context has been canceled.
+				// If it has, the context error is more important than the upload error.
```
krishnamd-jkp: We might lose information here. If there are 2 different errors, we might always return only the first error and overwrite the second error. Can we instead report a series of errors which happened during the uploads?

krishnamd-jkp: We can maybe have both in the state. firstErr and a list of errors.

storage/pcu_test.go
```diff
@@ -161,9 +163,379 @@ func TestParallelUploadConfig_defaults(t *testing.T) {
...
+		ctx:      pCtx,
+		cancel:   cancel,
+		resultCh: make(chan uploadResult, 2),
+		partMap:  make(map[int]*ObjectHandle),
+	}
+
+	state.collectorWG.Add(1)
+	go state.resultCollector()
+
+	// Successful result
```
krishnamd-jkp: nit: Missing period. Here and other places too.

storage/pcu_test.go
```diff
@@ -161,9 +163,379 @@ func TestParallelUploadConfig_defaults(t *testing.T) {
...
+
+	select {
+	case result := <-state.resultCh:
+		if result.err != nil {
+			t.Errorf("worker unexpected error: %v", result.err)
+		}
+		if result.partNumber != 1 || result.handle == nil || result.handle.object != "mockPart" {
+			t.Errorf("worker result mismatch: got %v", result)
+		}
+	case <-time.After(1 * time.Second):
```
krishnamd-jkp, line 290: nit: explain the wait statement here.

storage/pcu_test.go
```diff
@@ -161,9 +163,379 @@ func TestParallelUploadConfig_defaults(t *testing.T) {
...
+		t.Errorf("worker timeout waiting for buffer return")
+	}
+
+	close(state.uploadCh)
+	state.workerWG.Wait()
+}
+
+func TestPCUWorker_FailedTask(t *testing.T) {
```
krishnamd-jkp, line 308: Can you add comments after each stage? "Check for error", "Check if context has been cancelled", "Check if buffer is returned"?

storage/pcu_test.go
```diff
@@ -161,9 +163,379 @@ func TestParallelUploadConfig_defaults(t *testing.T) {
...
+		if result.err == nil || result.err.Error() != uploadErr.Error() {
+			t.Errorf("worker error mismatch: got %v, want %v", result.err, uploadErr)
+		}
+		if result.partNumber != 1 {
+			t.Errorf("worker partNumber mismatch: got %v", result.partNumber)
+		}
+	case <-time.After(1 * time.Second):
+		t.Errorf("worker timeout waiting for result")
+	}
+
```
krishnamd-jkp, line 343: Can you check if context has been cancelled as well?

storage/pcu_test.go
```diff
@@ -161,9 +163,379 @@ func TestParallelUploadConfig_defaults(t *testing.T) {
...
+	// Verify the buffer was correctly returned to the channel.
+	select {
+	case <-state.bufferCh:
+	case <-time.After(1 * time.Second):
+		t.Errorf("worker timeout waiting for buffer return after cancel")
+	}
+}
+
+func TestPCUWorker_ChannelClose(t *testing.T) {
```
krishnamd-jkp: Can be better with the naming here. Doesn't come out clear which channel is being closed and which error should be checked.

storage/pcu.go
cpriti-os: Not recommended by go style guide, w is preferred over writer.

storage/pcu.go
cpriti-os: This should me more efficient, since we are doing a single atomic operation using one Add over N operations, perf over readability.

storage/pcu_test.go
cpriti-os: Done

storage/pcu_test.go
cpriti-os: Done

storage/pcu_test.go
cpriti-os: Done

storage/pcu.go
cpriti-os: Done

storage/pcu.go
cpriti-os: Done

storage/pcu.go
cpriti-os: Done

storage/pcu.go
```diff
@@ -211,14 +211,18 @@ type uploadResult struct {
...
 
-	// TODO: Get the config from the Writer
+	// TODO: Get the config from the Writer.
 	cfg := &ParallelUploadConfig{}
 	cfg.defaults()
 
-	if cfg.PartSize%(chunkSizeMultiple) != 0 {
-		return fmt.Errorf("PartSize must be a multiple of 256KiB")
+	// Fall back to the nearest multiple of 256KiB.
+	if cfg.PartSize%chunkSizeMultiple != 0 {
```
krishnamd-jkp: Can you do something like: https://github.com/googleapis/google-cloud-go/blob/c6773ceb279f2146eb13a6795d79a0415f71ea41/storage/grpc_writer.go#L1469?

storage/pcu.go
```diff
@@ -201,3 +209,168 @@ type uploadResult struct {
...
+			return
+		case task, ok := <-s.uploadCh:
+			if !ok {
+				return
+			}
+			func(t uploadTask) {
+				// Ensure the buffer is returned to the pool.
+				defer func() { s.bufferCh <- t.buffer }()
+				defer func() {
+					if r := recover(); r != nil {
```
krishnamd-jkp: Can you add a test case for panic and check if buffer is being returned?

storage/pcu_test.go
cpriti-os: The error statement below explains the intent of wait, will add a comment once to make the intent clear. Adding comment for this everywhere might not be too useful

storage/pcu_test.go
cpriti-os: In this specific test (TestPCUWorker_FailedTask), I am only unit-testing the worker's ability to report an error and return its buffer. The propagation of that error to a global context cancellation is verified separately in TestPCUState_ResultCollector which ensures that any error received from a worker triggers setError and cancels the context

storage/pcu.go
cpriti-os: Reused it.

storage/pcu.go
cpriti-os: Done

storage/pcu.go
cpriti-os: Done

storage/pcu.go
cpriti-os: Done

storage/pcu.go
cpriti-os: Removed

storage/pcu.go
cpriti-os: Added.

storage/pcu_test.go
krishnamd-jkp: Yeah I get it that we're waiting for a result from a channel and do not want to wait indefinitely, adding a comment will explain the intent better. Yeah adding it at one place should do.

storage/pcu.go
krishnamd-jkp: The error is still being overwritten here. We'll always get the first error, won't we?

storage/pcu.go
```diff
@@ -40,7 +43,7 @@ type ParallelUploadConfig struct {
 
 	// PartSize is the size of each part to be uploaded in parallel.
 	// Defaults to 16MiB. Must be a multiple of 256KiB.
-	PartSize int64
+	PartSize int
```
tritone: I missed the previous PR on this, but -- can we revert any exported fields and types related to PCU until we have something ready to be used? If the previous code in pcu.go had been released, changes like this line would be considered breaking.

You can just change everything to non-exported until we have something ready for customers.

storage/pcu_test.go
```diff
@@ -161,9 +163,444 @@ func TestParallelUploadConfig_defaults(t *testing.T) {
...
+
+	buffer := make([]byte, 10)
+	state := &pcuState{
+		ctx:      ctx,
+		cancel:   cancel,
+		bufferCh: make(chan []byte, 1),
+		uploadCh: make(chan uploadTask, 1),
+		resultCh: make(chan uploadResult, 1),
+		uploadPartFn: func(s *pcuState, task uploadTask) (*ObjectHandle, *ObjectAttrs, error) {
+			panic("simulated panic")
```
tritone: Under what circumstances would we have a panic to recover from?

In general we don't want to use panics for error handling.

storage/pcu.go
```diff
@@ -201,3 +210,168 @@ type uploadResult struct {
...
+				}
+
+				// Always send a result to the collector.
+				s.resultCh <- uploadResult{partNumber: t.partNumber, obj: attrs, handle: handle, err: err}
+			}(task)
+		}
+	}
+}
+
+// TODO: add retry logic.
```
tritone, line 295: Retry logic should be fully encapsulated in the individual part's Writer for the most part?

storage/pcu.go
cpriti-os: This is redundant now, returning nil is better for successful uploads, removed it.

storage/pcu.go
cpriti-os: Writer would not retry these uploads if I understand correctly because I am using single shot uploads (chunkSize = 0). This should be solved by my optimization PR for gRPC write but till then I am planning to keep the TODO.

storage/pcu_test.go
cpriti-os: The recover() block is intended as a defensive safeguard at the goroutine boundary to ensure that any unexpected  failure doesn't crash the entire client while guaranteeing that the data buffer is returned to the pool so other workers can proceed. However I agree, we do not expect any panics so should not be handling it. Removed.

storage/pcu.go
cpriti-os: Done

PR #13516, opened 2025-12-29
feat: bigquery driver can now limit maxRetries of retryAble errors. When that happens, it returns RetryExhaustedError with all N errors listed.

nishant-rudderstack:
bigquery driver can now limit maxRetries of retryAble errors. When that happens, it returns RetryExhaustedError with all N errors listed.


nishantsharma:
@shollyman @hongalex  PTAL. I am unable to make an internal production release because this PR is not merged upstream. And, without this fix, we have seen infinite retries holding up BQ projects in some recent cases.

PR #13520, opened 2025-12-29
feat(bigtable): add ip protocol for bigtable connection

sushanb:
we will need this for populating direct_access/compatible as it has an `ip_preference` metric field.


bigtable/internal/transport/connpool.go
```diff
@@ -137,6 +169,17 @@ func (bc *BigtableConn) Prime(ctx context.Context, fullInstanceName, appProfileI
 		return err
 	}
 
+	// ip protocol will be -1 if it addr is nil/default, 0 is ipv4 and 1 if ipv6.
+	if p.Addr != nil {
+		if tcpAddr, ok := p.Addr.(*net.TCPAddr); ok {
+			if tcpAddr.IP.To4() != nil {
```
nimf: Theoretically `IP` field of `net.TCPAddr` can be nil. Should we protect from this here?

sushanb: Added.

PR #13521, opened 2025-12-29
feat(bigtable): add client startup time metrics. 

sushanb:
TODO: on adding the transport type as we need to make Prime() rpc blocking.

### DIFF
--- a/bigtable/internal/transport/connpool_test.go
+++ b/bigtable/internal/transport/connpool_test.go
@@ -26,12 +26,8 @@ import (
 	"sync"
 	"sync/atomic"
 	"testing"
-	"testing/synctest"
 	"time"
 
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/sdk/metric"
-	"go.opentelemetry.io/otel/sdk/metric/metricdata"
 	"google.golang.org/grpc/codes"
 	testpb "google.golang.org/grpc/interop/grpc_testing"
 	"google.golang.org/grpc/metadata"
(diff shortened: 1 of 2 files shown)


bigtable/internal/transport/connpool.go
```diff
@@ -384,9 +385,37 @@ func NewBigtableChannelPool(ctx context.Context, connPoolSize int, strategy btop
...
+	// record the client startup time
+	// TODO: currently Prime() is non-blocking, we will make Prime() blocking and infer the transport type here.
+	transportType := "unknown"
+	pool.recordClientStartUp(startupTimer, transportType)
+
 	return pool, nil
 }
 
+func (p *BigtableChannelPool) recordClientStartUp(elapsedTimer time.Time, transportType string) {
+	if p.meterProvider != nil {
```
nimf: nit: prefer
```go
if p.meterProvider == nil {
  return
}

...
clientStartupTime, err := meter.Float64Histogram(...
if err != nil {
  return
}

elapsed := float64(time.Since(elapsedTimer).Milliseconds())
```
This decreases levels of indentation and makes it easier to grasp normal code path.

go/go-style/decisions#indent-error-flow

bigtable/internal/transport/connpool_test.go
```diff
@@ -1473,6 +1476,72 @@ func TestConnPoolStatisticsVisitor(t *testing.T) {
...
+
+	pool := &BigtableChannelPool{
+		poolCtx:       ctx,
+		meterProvider: provider,
+	}
+
+	startTime := time.Now().Add(-500 * time.Millisecond)
+	transportType := "directpath"
+
+	pool.recordClientStartUp(startTime, transportType)
```
nimf: is it hard to test this via `NewBigtableChannelPool`?

bigtable/internal/transport/connpool.go
```diff
@@ -299,7 +300,7 @@ func (p *BigtableChannelPool) getConns() []*connEntry {
 // NewBigtableChannelPool creates a pool of connPoolSize and takes the dial func()
 // NewBigtableChannelPool primes the new connection in a non-blocking goroutine to warm it up.
 // We keep it consistent with the current channelpool behavior which is lazily initialized.
-func NewBigtableChannelPool(ctx context.Context, connPoolSize int, strategy btopt.LoadBalancingStrategy, dial func() (*BigtableConn, error), opts ...BigtableChannelPoolOption) (*BigtableChannelPool, error) {
+func NewBigtableChannelPool(ctx context.Context, connPoolSize int, strategy btopt.LoadBalancingStrategy, dial func() (*BigtableConn, error), startupTimer time.Time, opts ...BigtableChannelPoolOption) (*BigtableChannelPool, error) {
```
nimf: is adding this time argument only for testing purposes?
If yes, then we can use the [synctest](https://pkg.go.dev/testing/synctest) package nowadays for time-dependent tests.

Also, it seems like the passed value is not checked in the test, it just tests for a positivity.

bigtable/internal/transport/connpool.go
```diff
@@ -384,9 +385,37 @@ func NewBigtableChannelPool(ctx context.Context, connPoolSize int, strategy btop
 		btopt.Debugf(pool.logger, "bigtable_connpool: failed to create metrics reporter: %v\n", err)
 	}
 	pool.startMonitors()
+
+	// record the client startup time
+	// TODO: currently Prime() is non-blocking, we will make Prime() blocking and infer the transport type here.
```
nimf, line 390: Will this be released before we make Prime blocking?

bigtable/internal/transport/connpool.go
sushanb: Done.

bigtable/internal/transport/connpool.go
sushanb: No, we will release after everything is supported and this new channel pool is guarded by a flag `BIGTABLE_CONN_POOL` which is set as false too.

i will send the cls for making prime blocking shortly.

bigtable/internal/transport/connpool.go
sushanb: synctest requires Go 1.25 or 1.24 with GOEXPERIMENT=synctest. we are still using 1.24.

nimf: I think we can use `toolchain go1.25.5` directive in `go.mod` while still having `go 1.24` there. This should allow library users to use the library with go 1.24+ but our tests will run with 1.25 and synctest.
But it's up to you. As this is internal API, we can revert that in the future.

bigtable/bigtable.go
```diff
@@ -119,6 +119,7 @@ func NewClient(ctx context.Context, project, instance string, opts ...option.Cli
 
 // NewClientWithConfig creates a new client with the given config.
 func NewClientWithConfig(ctx context.Context, project, instance string, config ClientConfig, opts ...option.ClientOption) (*Client, error) {
+	elaspedTimer := time.Now()
```
nimf: typo here. Also, this is a bit confusing that this is called a timer while this is just a point in time -- same applies to `startupTimer` argument of `NewBigtableChannelPool`

bigtable/internal/transport/connpool.go
sushanb: . it worked!! thanks

bigtable/bigtable.go
sushanb: Fixed.

nimf (approved):
The only thing I don't like in this PR is adding the new `clientCreationTimestamp` argument to the `NewBigtableChannelPool` function. It already has 5 arguments. We don't really need this argument because we can get the time instant inside the `NewBigtableChannelPool`. But we "need" this to test the client startup metric, so I was thinking how can we do this in another way.

We provide the `dial` function as one of the arguments. What if for that test we add `time.Sleep` inside the `dialFunc` we create in the test? As our test runs with synctest the test won't acutally sleep but rather fast-forward time.

On the other hand this is internal API, so we can leave it as is and make this change later. Up to you.

sushanb:
> The only thing I don't like in this PR is adding the new `clientCreationTimestamp` argument to the `NewBigtableChannelPool` function. It already has 5 arguments. We don't really need this argument because we can get the time instant inside the `NewBigtableChannelPool`. But we "need" this to test the client startup metric, so I was thinking how can we do this in another way.
> 
> We provide the `dial` function as one of the arguments. What if for that test we add `time.Sleep` inside the `dialFunc` we create in the test? As our test runs with synctest the test won't acutally sleep but rather fast-forward time.
> 
> On the other hand this is internal API, so we can leave it as is and make this change later. Up to you.

i agree as the most time intensive part is in NewBigtableChannelPool. Semantically, we want to measure the time when customer intiialize NewClientWithConfig to when client ready so that;s why we add the clientCreationTimestamp

bhshkh:
@sushanb please do not merge PRs without first updating the branch.

https://github.com/googleapis/google-cloud-go/issues/13552

sushanb:
oops sorry i forgot to sync with upstream merge there. thanks for fixing!!!

PR #13524, opened 2025-12-30
refactor(storage): refactor MultiRangeDownloader

tritone:
Complete rewrite of storage.MultiRangeDownloader. The new design should be more resilient to concurrency issues, deadlocks, retries, etc.

### DIFF
--- a/storage/emulator_test.sh
+++ b/storage/emulator_test.sh
@@ -31,6 +31,7 @@ if [ "$minor_ver" -lt "$min_minor_ver" ]; then
     exit 0
 fi
 
+export DOCKER_API_VERSION=1.39
 export STORAGE_EMULATOR_HOST="http://localhost:9000"
 export STORAGE_EMULATOR_HOST_GRPC="localhost:8888"
 
@@ -66,6 +67,7 @@ function cleanup() {
     docker stop $CONTAINER_NAME
     unset STORAGE_EMULATOR_HOST;
     unset STORAGE_EMULATOR_HOST_GRPC;
+    unset DOCKER_API_VERSION
 }
 trap cleanup EXIT
 


storage/reader.go
```diff
@@ -431,27 +434,27 @@ func (mrd *MultiRangeDownloader) Add(output io.Writer, offset, length int64, cal
...
-	mrd.reader.wait()
+	mrd.impl.wait()
 }
 
 // GetHandle returns the read handle. This can be used to further speed up the
 // follow up read if the same object is read through a different stream.
 func (mrd *MultiRangeDownloader) GetHandle() []byte {
-	return mrd.reader.getHandle()
+	return mrd.impl.getHandle(context.TODO()) // TODO: Consider plumbing context from caller
 }
```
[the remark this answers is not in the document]
tritone: This would be a breaking change; instead I'm just going to make this wait on the manager's ctx only.

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+	req := &rangeRequest{
+		output:     cmd.output,
+		offset:     cmd.offset,
+		length:     cmd.length,
+		origOffset: cmd.offset,
+		origLength: cmd.length,
+		callback:   cmd.callback,
+		readID:     m.readIDCounter,
+	}
+	m.readIDCounter++
```
krishnamd-jkp, line 488: Is this thread safe? Can we make use of readIDGenerator struct at the start of the file instead, if not?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+	}
+	req.offset = start
+	req.length = objSize - start
+}
+
+func (m *multiRangeDownloaderManager) handleCloseCmd(ctx context.Context, cmd *mrdCloseCmd) {
+	if m.permanentErr == nil {
+		m.permanentErr = cmd.err
+		if m.permanentErr == nil {
+			m.permanentErr = errClosed
```
krishnamd-jkp: This makes it seem like the MRD operation has failed. Can we have a different parameter indicating that MRD has been closed?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+	}
+
+	mCtx, cancel := context.WithCancel(ctx)
+
+	// Create the manager
+	manager := newMultiRangeDownloaderManager(mCtx, c, s, params, readSpec, cancel, ctx)
+
+	mrd := &MultiRangeDownloader{
+		impl:    manager,
+		spanCtx: ctx,
```
BrennaEpp: Does this mean that the span context for `Object.NewMultiRangeDownloader` is used throughout the life of the MRD?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+	}
+	if length < 0 {
+		m.runCallback(offset, length, fmt.Errorf("storage: MultiRangeDownloader.Add limit cannot be negative"), callback)
+		return
+	}
+
+	cmd := &mrdAddCmd{output: output, offset: offset, length: length, callback: callback}
+	select {
+	case m.cmdC <- cmd:
+	case <-m.ctx.Done():
```
BrennaEpp, line 258: Can we eliminate lines 256-259 as we have this here?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+		return mrd, nil
+	case <-ctx.Done():
+		cancel()
+		manager.wg.Wait()
+		return nil, ctx.Err()
+	}
+}
+
+// --- mrdCommand Interface and Implementations ---
+type mrdCommand interface {
```
BrennaEpp, line 128: Can you add a brief description of how this interface works?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+
+// --- multiRangeDownloaderManager ---
+
+type multiRangeDownloaderManager struct {
+	ctx          context.Context
+	cancel       context.CancelFunc
+	client       *grpcStorageClient
+	settings     *settings
+	params       *newMultiRangeDownloaderParams
+	wg           sync.WaitGroup
```
BrennaEpp: Can you rename this to be more specific to what this group syncs?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+	cmdC         chan mrdCommand
```
BrennaEpp: I generally prefer for channels to use the plural of the contents, unless already taken by a different variable:
```suggestion
	cmds         chan mrdCommand
```

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+			if m.permanentErr == nil {
+				m.permanentErr = err
+			}
+			m.attrsOnce.Do(func() { close(m.attrsReady) })
+		}
+	}
+	return err
+}
+
+func (m *multiRangeDownloaderManager) handleAddCmd(ctx context.Context, cmd *mrdAddCmd) {
```
BrennaEpp, line 473: Are calls to `handleAddCmd` guaranteed to run in sequence?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+			}
+		}
+		if len(rangesToResend) > 0 {
+			m.currentSession.SendRequest(&storagepb.BidiReadObjectRequest{ReadRanges: rangesToResend})
+		}
+		return nil
+	}, m.settings.retry, true)
+}
+
+var errBidiReadRedirect = errors.New("bidi read object redirected")
```
BrennaEpp, line 671: Should this error have more or different information when surfaced to the user? This doesn't seem actionable

storage/reader.go
```diff
@@ -387,19 +386,23 @@ func (r *Reader) ReadHandle() ReadHandle {
...
 
+// type multiRangeDownloader interface {
+// 	add(output io.Writer, offset, limit int64, callback func(int64, int64, error))
+// 	wait()
+// 	close() error
+// 	getHandle() []byte
+// 	error() error
+// }
+
```
BrennaEpp: Remove

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+	cmdC         chan mrdCommand
+	sessionRespC chan mrdSessionResult
+	publicMRD    *MultiRangeDownloader
+
+	// State
+	currentSession *bidiReadStreamSession
+	readIDCounter  int64
+	pendingRanges  map[int64]*rangeRequest
+	permanentErr   error
+	waiters        []chan struct{}
```
BrennaEpp, line 217: Why is this a range?

BrennaEpp (commented):
Some initial comments/questions

storage/reader.go
```diff
@@ -387,19 +386,23 @@ func (r *Reader) ReadHandle() ReadHandle {
 //
 // This API is currently in preview and is not yet available for general use.
 type MultiRangeDownloader struct {
-	Attrs  ReaderObjectAttrs
-	reader multiRangeDownloader
-	ctx    context.Context
-}
+	// Attrs is populated when NewMultiRangeDownloader returns.
+	Attrs ReaderObjectAttrs
+	impl  internalMultiRangeDownloader
```
krishnamd-jkp, line 393: Can we call it manager instead?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+}
+type mrdAddCmd struct {
+	output   io.Writer
+	offset   int64
+	length   int64
+	callback func(int64, int64, error)
+}
+
+func (c *mrdAddCmd) apply(ctx context.Context, m *multiRangeDownloaderManager) {
+	m.handleAddCmd(ctx, c)
```
krishnamd-jkp, line 139: Why can't we have the logic inside this exact method instead of a different function again?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+			}
+			return err
+		}
+		m.currentSession = session
+
+		// Wait for the first message to populate attributes
+		select {
+		case firstResult := <-m.sessionRespC:
+			if firstResult.err != nil {
+				m.currentSession.Shutdown()
```
krishnamd-jkp: This is done when manager exits the loop anyway, right?  line 351

storage/reader.go
tritone: Done

storage/grpc_reader_multi_range.go
tritone: Yes, there is a parent span that lasts the lifetime of the MRD. This was the case for the original impl as well.

storage/grpc_reader_multi_range.go
tritone: Done.

storage/grpc_reader_multi_range.go
tritone: I thought it was clearer to have these commands trigger methods on the manager since they use the manager internals heavily; I guess there are various ways to do this though.

storage/grpc_reader_multi_range.go
tritone: I added a comment.

storage/grpc_reader_multi_range.go
tritone: Done

storage/grpc_reader_multi_range.go
tritone: I think theoretically MRD.Wait() could be called from multiple threads. In this case we would need to track and unblock all of these when we are done.

storage/grpc_reader_multi_range.go
tritone: It feels clearer to me to not even attempt to add the range if the ctx is already done when add is called. Does that make sense? It's probably fine either way.

storage/grpc_reader_multi_range.go
tritone: I think you're right, removed.

storage/grpc_reader_multi_range.go
tritone: Yes because cmds are run one at a time in the event loop.

storage/grpc_reader_multi_range.go
tritone: It's no longer an issue with this design because cmds are run one at a time by the event loop, so you can't have two add commands being applied simultaneously. I think the code at the top to generate ids can be removed.

storage/grpc_reader_multi_range.go
tritone: Via tests I think close returns a nil error unless there was a perm err before this. I will confirm.

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,860 @@ func (g *readIDGenerator) Next() int64 {
...
+	for _, dataRange := range resp.GetObjectDataRanges() {
+		readID := dataRange.GetReadRange().GetReadId()
+		req, exists := m.pendingRanges[readID]
+		if !exists || req.completed {
+			continue
+		}
+
+		content := dataRange.GetChecksummedData().GetContent()
+		req.bytesWritten += int64(len(content))
+		_, err := req.output.Write(content)
```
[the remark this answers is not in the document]
tritone: Discussed this offline, it is the caller's responsibility to ensure a non-blocking writer. We can consider this as an option later.

storage/grpc_reader_multi_range.go
tritone: This is never surfaced to the user, it's just a sentinel error to trigger the redirect loop.

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+			m.currentSession.SendRequest(&storagepb.BidiReadObjectRequest{ReadRanges: rangesToResend})
+		}
+		return nil
+	}, m.settings.retry, true)
+}
+
+var errBidiReadRedirect = errors.New("bidi read object redirected")
+
+func (m *multiRangeDownloaderManager) handleStreamEnd(result mrdSessionResult) {
+	m.currentSession = nil
```
krishnamd-jkp: The streams context should be cancelled (s.shutDown()) before making this as null?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+	start := objSize + req.offset
+	if start < 0 {
+		start = 0
+	}
+	req.offset = start
+	req.length = objSize - start
+}
+
+func (m *multiRangeDownloaderManager) handleCloseCmd(ctx context.Context, cmd *mrdCloseCmd) {
+	if m.permanentErr == nil {
```
krishnamd-jkp: This can be simplified. Also this is repeated at many places - 
```go
if m.permanentErr == nil {
	m.permanentErr = errClosed
}
```

Can we have something like m.setPermanentError?

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+		m.permanentErr = cmd.err
+		if m.permanentErr == nil {
+			m.permanentErr = errClosed
+		}
+	}
+	m.attrsOnce.Do(func() { close(m.attrsReady) })
```
krishnamd-jkp: This is done when event loop exits anyway, can we remove it? And other instances of this logic as well

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+		readIDCounter: 1,
+		readSpec:      readSpec,
+		attrsReady:    make(chan struct{}),
+		spanCtx:       spanCtx,
+	}
+}
+
+// Methods implementing internalMultiRangeDownloader
+func (m *multiRangeDownloaderManager) add(output io.Writer, offset, length int64, callback func(int64, int64, error)) {
+	if err := m.getPermanentError(); err != nil {
```
krishnamd-jkp: This will be a data race I think. We can alternatively check for permanent error in handleAddCmd which will be thread-safe.

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+		origOffset: cmd.offset,
+		origLength: cmd.length,
+		callback:   cmd.callback,
+		readID:     m.readIDCounter,
+	}
+	m.readIDCounter++
+
+	// Attributes should be ready if we are processing Add commands
+	if req.offset < 0 {
+		m.convertToPositiveOffset(req)
```
krishnamd-jkp: The errors inside m.convertToPositiveOffset will fail the range but the range is still added to the stream. Can we skip the range when the range is failed? maybe get an error from ```convertToPositiveOffset```  and skip the range when err is not nil.

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+		offset:     cmd.offset,
+		length:     cmd.length,
+		origOffset: cmd.offset,
+		origLength: cmd.length,
+		callback:   cmd.callback,
+		readID:     m.readIDCounter,
+	}
+	m.readIDCounter++
+
+	// Attributes should be ready if we are processing Add commands
```
krishnamd-jkp, line 490: We should also check a case where if offset is greater than the object size, the range should be failed and not added to the stream. Otherwise, a permanent error will be set on the MRD when server gives a Out of Range error.

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,834 @@ func (g *readIDGenerator) Next() int64 {
...
+	if req.offset < 0 {
+		m.convertToPositiveOffset(req)
+	}
+	m.pendingRanges[req.readID] = req
```
krishnamd-jkp: This can be moved below the if condition below if the range will be failed if session is not available

storage/grpc_reader_multi_range.go
```diff
@@ -38,3 +51,860 @@ func (g *readIDGenerator) Next() int64 {
...
+			if m.permanentErr == nil {
+				m.permanentErr = err
+			}
+		} else if m.permanentErr == nil {
+			m.permanentErr = errClosed
+		}
+		m.failAllPending(m.permanentErr)
+		m.attrsOnce.Do(func() { close(m.attrsReady) })
+	}
+}
```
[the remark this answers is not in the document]
krishnamd-jkp: I agree. I think there is a bit of code duplication here. Can be simplified a bit I think.

storage/grpc_reader_multi_range.go
```diff
@@ -14,27 +14,846 @@
...
+	wait()
+	getHandle() []byte
+	getPermanentError() error
+	getSpanCtx() context.Context
+}
+
+// --- grpcStorageClient method ---
+
+func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params *newMultiRangeDownloaderParams, opts ...storageOption) (*MultiRangeDownloader, error) {
+	s := callSettings(c.settings, opts...)
```
krishnamd-jkp, line 59: Should we make sure if WithGRPCBidiReads option is set? 
Like - 
```go
if !c.config.grpcBidiReads {
	return nil, errors.New("storage: MultiRangeDownloader requires the experimental.WithGRPCBidiReads option")
}
```

storage/integration_test.go
```diff
@@ -425,8 +425,8 @@ func TestIntegration_MultiRangeDownloader(t *testing.T) {
 				want = content[k.offset : k.offset+k.limit]
 			}
 			if !bytes.Equal(k.buf.Bytes(), want) {
-				t.Errorf("Error in read range offset %v, limit %v, got: %v; want: %v",
-					k.offset, k.limit, k.buf.Bytes(), want)
+				t.Errorf("Error in read range offset %v, limit %v, got: %v bytes; want: %v bytes",
+					k.offset, k.limit, len(k.buf.Bytes()), len(want))
```
cpriti-os, line 429: Why are we removing data matching check here?

storage/grpc_reader_multi_range.go
cpriti-os: This fix is not visible

storage/grpc_reader_multi_range.go
cpriti-os: +1 , semantically this might imply a failure when none occurred, will be good to confirm.

storage/grpc_reader_multi_range.go
```diff
@@ -14,27 +14,846 @@
...
+
+	// Create the manager
+	manager := &multiRangeDownloaderManager{
+		ctx:           mCtx,
+		cancel:        cancel,
+		client:        c,
+		settings:      s,
+		params:        params,
+		cmds:          make(chan mrdCommand, 1),
+		sessionResps:  make(chan mrdSessionResult, 100),
```
cpriti-os: would it be better  to define these as constants

storage/grpc_reader_multi_range.go
```diff
@@ -14,27 +14,846 @@
...
+
+func (c *mrdErrorCmd) apply(ctx context.Context, m *multiRangeDownloaderManager) {
+	select {
+	case c.respC <- m.permanentErr:
+	case <-ctx.Done():
+		close(c.respC)
+	}
+}
+
+// --- mrdSessionResult ---
```
cpriti-os, line 187: Nit: Add descriptive comment.

storage/grpc_reader_multi_range.go
```diff
@@ -14,27 +14,846 @@
...
+		pendingRanges: make(map[int64]*rangeRequest),
+		readIDCounter: 1,
+		readSpec:      readSpec,
+		attrsReady:    make(chan struct{}),
+		spanCtx:       ctx,
+	}
+
+	mrd := &MultiRangeDownloader{
+		impl:    manager,
+		spanCtx: ctx,
```
cpriti-os: do we still need this here given impl is managing it now?

storage/grpc_reader_multi_range.go
```diff
@@ -14,27 +14,846 @@
...
+	}
+	manager.publicMRD = mrd
```
krishnamd-jkp: Seems unnecessary

storage/grpc_reader_multi_range.go
```diff
@@ -14,27 +14,846 @@
...
+		m.readSpec.ReadHandle = result.redirect.ReadHandle
+		if ensureErr := m.ensureSession(m.ctx); ensureErr != nil {
+			if !m.isRetryable(ensureErr) {
+				m.permanentErr = ensureErr
+				m.attrsOnce.Do(func() { close(m.attrsReady) })
+				m.failAllPending(m.permanentErr)
+			}
+		}
+	} else if m.isRetryable(err) {
+		if len(m.pendingRanges) > 0 {
```
krishnamd-jkp: Why this condition? We should re-establish the connection irrespective of the pending ranges, right?

storage/integration_test.go
BrennaEpp: This doesn't remove the check, it only removes all the bytes from the output (probably to avoid long and not all that useful logs). It prints the lengths when the bytes don't match up. We do the same in several other tests.

storage/grpc_reader_multi_range.go
BrennaEpp: I think we should keep the old name. Having "NewMultiRangeDownloader" as an overarching span is confusing.

storage/grpc_reader_multi_range.go
BrennaEpp: Makes sense

storage/grpc_reader_multi_range.go
BrennaEpp: Let's add a comment at the top here indicating that, so we know everything inside here doesn't need to be thread-safe (ex. the readIdCounter).

storage/grpc_reader_multi_range.go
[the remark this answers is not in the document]
BrennaEpp: Even a non-blocking writer will take some time to write though - if this is blocking Add() calls in the meantime it may be noticeable performance degradation for certain workloads.

storage/grpc_reader_multi_range.go
tritone: Done.

storage/integration_test.go
tritone: +1 to what Brenna said.

storage/grpc_reader_multi_range.go
tritone: Should be in line 33 and following in the latest commit.

storage/grpc_reader_multi_range.go
tritone: Good point, fixed.

storage/grpc_reader_multi_range.go
tritone: Added a top level comment to the mrdCommand type.

storage/reader.go
tritone: Manager feels like a very overloaded word in this context since this is not the MRDManager type... feel free to rename it to something else in a subsequent PR.

storage/grpc_reader_multi_range.go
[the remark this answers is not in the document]
tritone: Definitely worth investigating but if we are just writing messages to in-memory buffers it will be trivial. For file io I agree it will be a bigger deal perhaps.

storage/grpc_reader_multi_range.go
tritone: Confirmed this.

storage/grpc_reader_multi_range.go
tritone: good point, done.

storage/grpc_reader_multi_range.go
tritone: This is a bit of a tricky case because it is possible if there is another concurrent writer to the object, the Size will be out-of-date and these calls will in fact succeed. I don't think we validate this in the existing MRD code and it's on the caller to decide how to handle this.

storage/grpc_reader_multi_range.go
krishnamd-jkp, line 1: I think zero copy codec changes are not implemented. Any plans of adding them?

storage/grpc_reader_multi_range.go
tritone: added this

storage/grpc_reader_multi_range.go
cpriti-os: Yes but the MRD works with the single version of the object which is what we should stick to ? Without this validation we would set permanent error if there is any one invalid range provided. (not getting data for any valid range provided after this)

storage/grpc_reader_multi_range.go
tritone: Added these now.

storage/grpc_reader_multi_range.go
tritone: Thanks, yeah I think this was referenced in an earlier version but no longer needed. removed.

storage/grpc_reader_multi_range.go
tritone: Good point, we could need to re-establish even if there are no pending ranges currently.

storage/grpc_reader_multi_range.go
tritone: The MRD is supposed to support an object that grows; see the tailing reads example: https://github.com/GoogleCloudPlatform/golang-samples/blob/main/storage/rapid/read_appendable_object_tail.go

Can you check with the GCSFuse team on the expected behavior here? I know they have logic in their code to recover from these types of permanent errors.

storage/grpc_reader_multi_range.go
tritone: Done

storage/grpc_reader_multi_range.go
tritone: Done

storage/grpc_reader_multi_range.go
tritone: Not needed anymore, correct :)

storage/grpc_reader_multi_range.go
tritone: You're right, this was only necessary when exiting the event loop (or when the attrs are actually successfully processed). Removed other instances.

storage/grpc_reader_multi_range.go
tritone: It is though probably harmless because it just short-circuits in this case. In any case looks like we do check in handleAddCmd as well so I can remove.

storage/grpc_reader_multi_range.go
tritone: Done

storage/grpc_reader_multi_range.go
tritone: Done.

cpriti-os (dismissed):

storage/grpc_reader_multi_range.go
cpriti-os: Sounds good, will confirm with them, if required can be fixed in a subsequent PR

PR #13534, opened 2026-01-04
[WIP] feat: add checksums for small and single shot json uploads

krishnamd-jkp:


storage/http_client.go
```diff
@@ -983,11 +985,145 @@ func (c *httpStorageClient) newRangeReaderJSON(ctx context.Context, params *newR
...
+	multiChunkChan chan bool
+
+	state writerState
+
+	// closeOnce ensures that the channels are closed at most once.
+	closeOnce sync.Once
+	// communicateOnce ensures that we only signal the uploader goroutine once,
+	// either with a checksum or to transition to multi-chunk mode.
+	communicateOnce sync.Once
+}
```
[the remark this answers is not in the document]
krishnamd-jkp: Changed it to context error which should be thread safe.

storage/integration_test.go
```diff
@@ -1882,6 +1882,11 @@ func TestIntegration_WriterCRC32CValidation(t *testing.T) {
 				chunkSize:  256 * 1024,
 				sendCRC32C: true,
 			},
```
cpriti-os, line 1884: Integration tests to cover the new auto checksum calculation scenarios would be good to add.

storage/http_client.go
```diff
@@ -983,11 +985,155 @@ func (c *httpStorageClient) newRangeReaderJSON(ctx context.Context, params *newR
...
+
+func (hiw *httpInternalWriter) closeChannels() {
+	hiw.closeOnce.Do(func() {
+		close(hiw.multiChunkChan)
+		close(hiw.checksumChan)
+	})
 }
 
-func (hiw httpInternalWriter) Flush() (int64, error) {
+func (hiw *httpInternalWriter) doOnce(f func()) {
```
cpriti-os: This name is very generic, use a bit descriptive name,

storage/http_client.go
```diff
@@ -983,11 +985,155 @@ func (c *httpStorageClient) newRangeReaderJSON(ctx context.Context, params *newR
...
+	serverChecksumChan chan uint32
+	// unblocks the uploader goroutine when the upload transitions to multi-chunk mode.
+	multiChunkChan chan bool
+
+	state writerState
+
+	// closeOnce ensures that the channels are closed at most once.
+	closeOnce sync.Once
+	// communicateOnce ensures that we only signal the uploader goroutine once,
+	// either with a checksum or to transition to multi-chunk mode.
```
cpriti-os: Nit: ""Ensures that either the computed checksum for single-chunk uploads or the multi-chunk transition signal is sent to the uploader goroutine exactly once"

storage/http_client.go
```diff
@@ -983,11 +985,155 @@ func (c *httpStorageClient) newRangeReaderJSON(ctx context.Context, params *newR
...
+	if hiw.ctx.Err() == nil {
+		hiw.communicateOnce.Do(f)
+	}
+}
+
+// validateChecksum validates the computed checksum against the server-provided checksum.
+func (hiw *httpInternalWriter) validateChecksum() error {
+	var err error
+	if hiw.ctx.Err() == nil {
+		hiw.communicateOnce.Do(func() {
```
cpriti-os: why not use doOnce here?

cpriti-os (changes requested):
Initial review

storage/integration_test.go
krishnamd-jkp: The existing tests - TestIntegration_WriterCRC32CValidation already cover these scenarios.

storage/http_client.go
```diff
@@ -983,11 +985,155 @@ func (c *httpStorageClient) newRangeReaderJSON(ctx context.Context, params *newR
...
+	})
+	return err
+}
+
+func (hiw *httpInternalWriter) Write(data []byte) (n int, err error) {
+	if hiw.checksumDisabled {
+		return hiw.PipeWriter.Write(data)
+	}
+
+	if hiw.ctx.Err() != nil {
```
tritone, line 1075: Is this additional check necessary? Won't the PipeWriter return the ctx err anyway?

storage/http_client.go
```diff
@@ -983,11 +985,155 @@ func (c *httpStorageClient) newRangeReaderJSON(ctx context.Context, params *newR
...
+		return 0, hiw.ctx.Err()
+	}
+
+	if hiw.state == stateSingleShot {
+		hiw.fullObjectChecksum = crc32.Update(hiw.fullObjectChecksum, crc32cTable, data)
+		return hiw.PipeWriter.Write(data)
+	}
+	hiw.mu.Lock()
+	defer hiw.mu.Unlock()
+	if hiw.state == stateSingleChunk {
```
tritone, line 1085: I'm confused by this... shouldn't the singleChunk vs. resumable logic be handled entirely inside the apiary layer? I would expect that we need special logic only when we have ChunkSize=0.

storage/http_client.go
```diff
@@ -983,11 +985,155 @@ func (c *httpStorageClient) newRangeReaderJSON(ctx context.Context, params *newR
...
+	bufferedWriter *bufio.Writer
+	ctx            context.Context
+	cancel         context.CancelFunc
+
+	checksumDisabled   bool
+	fullObjectChecksum uint32
+
+	// In single-chunk mode, the computed checksum is sent on this channel to
+	// the uploader goroutine just before the writer is closed.
+	checksumChan chan uint32
```
tritone, line 1025: I'm surprised that this degree of synchronization (3 channels + 1 mutex + 2 sync.Once) is required, especially since we don't expect the writer to be thread-safe to simultaneous Write/Close calls.

Is there any way to simplify this?

IIUC we just need to communicate to the goroutine doing the writing that the chunkSize was zero when writing started (can be sent on a channel). Then, we need to block returning from close until the checksum has been validated if need be (which can wait on donec). What am I missing here?

storage/http_client.go
krishnamd-jkp: To prevent any further checksum calculation and trying to send signals over channels (although we don't because the context error is checked before sending over a channel)

storage/http_client.go
krishnamd-jkp: Erm I think this is what is decided on the design finally. We buffer till chunksize is reached, calculate the checksum, set it on the object and then carry on with the apiary call.

PR #13539, opened 2026-01-06
test(storage): add and improve tests for MultiRangeDownloader

cpriti-os:
- The previously skipped `TestIntegration_ReadSameFileConcurrentlyUsingMultiRangeDownloader` test has been unskipped and refactored. It now uses `sync.WaitGroup` for managing concurrency, increases the test object size to 10MB, and randomizes offset/limit values for `Add` calls to better simulate varied request patterns and stress the internal event loop.
- **New MultiRangeDownloader Retry Conformance Tests**: Three new conformance tests have been added for the `MultiRangeDownloader`. These tests verify retry behavior for multiple concurrent requests on a single stream, ensure correct error handling for invalid download ranges (returning `codes.OutOfRange`), and confirm mid-stream resumption capabilities for large object downloads.

### DIFF
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -90,8 +90,8 @@ const (
 	testUniverseLocation   = "TEST_UNIVERSE_LOCATION"
 	testUniverseCreds      = "TEST_UNIVERSE_DOMAIN_CREDENTIAL"
 	// Location and Zone for zonal buckets tests
-	testZonalLocation = "us-central1"
-	testZonalZone     = "us-central1-a"
+	testZonalLocation = "us-west4"
+	testZonalZone     = "us-west4-a"
 )
 
 var (
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -41,6 +41,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -65,7 +66,6 @@ import (
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/sdk/trace/tracetest"
 	"golang.org/x/oauth2/google"
-	"golang.org/x/sync/errgroup"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 	itesting "google.golang.org/api/iterator/testing"
@@ -529,11 +529,11 @@ func TestIntegration_MRDCallbackReturnsDataLength(t *testing.T) {
 // TestIntegration_ReadSameFileConcurrentlyUsingMultiRangeDownloader tests for potential deadlocks
 // or race conditions when multiple goroutines call Add() concurrently on the same MRD multiple times.
 func TestIntegration_ReadSameFileConcurrentlyUsingMultiRangeDownloader(t *testing.T) {
-	t.Skip("flaky - https://github.com/googleapis/google-cloud-go/issues/12655")
 	multiTransportTest(skipAllButZonal(context.Background(), "Bidi Read API test"), t, func(t *testing.T, ctx context.Context, bucket string, _ string, client *Client) {
-		content := make([]byte, 5<<20)
+		// Use a 10MB object to allow for many non-overlapping range requests.
+		content := make([]byte, 10<<20)
 		rand.New(rand.NewSource(0)).Read(content)
-		objName := "MultiRangeDownloader"
+		objName := "MultiRangeDownloaderConcurrentReads"
 
 		// Upload test data.
 		obj := client.Bucket(bucket).Object(objName)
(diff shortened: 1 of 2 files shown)


storage/retry_conformance_test.go
```diff
@@ -316,6 +318,133 @@ var methods = map[string][]retryFunc{
...
+
+			if int64(buf.Len()) != 3*MiB {
+				return fmt.Errorf("resumption data length mismatch: got %d, want %d", buf.Len(), 3*MiB)
+			}
+			if !bytes.Equal(buf.Bytes(), randomBytes3MiB) {
+				return fmt.Errorf("resumption data mismatch: content does not match randomBytes3MiB")
+			}
+
+			return mrd.Close()
+		},
```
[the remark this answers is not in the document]
cpriti-os: they are inside storage.objects.get which is intentional

storage/integration_test.go
```diff
@@ -550,63 +550,59 @@ func TestIntegration_ReadSameFileConcurrentlyUsingMultiRangeDownloader(t *testin
...
-							res[taskID].offset = x
-							res[taskID].limit = y
-							res[taskID].err = err
-							// Send each callback result to subGroup.
-							done <- err
-						})
-						return <-done
+					taskID := gID*perGoRoutineAddCount + j
+					// Randomize offset/limit slightly to ensure varied request patterns.
+					offset := int64((taskID * 2048) % (len(content) - 4096))
```
krishnamd-jkp: Explain or add them as constants to understand these figures better

storage/integration_test.go
```diff
@@ -550,63 +550,59 @@ func TestIntegration_ReadSameFileConcurrentlyUsingMultiRangeDownloader(t *testin
...
+					limit := int64(1024 + rand.Intn(2048))
+
+					r := &rangeRes{offset: offset, limit: limit}
+					results[taskID] = r
+
+					reader.Add(&r.buf, offset, limit, func(o, l int64, err error) {
+						r.err = err
+						wg.Done()
```
krishnamd-jkp, line 597: Move wg.Done() outside the for loop?

storage/integration_test.go
```diff
@@ -550,63 +550,59 @@ func TestIntegration_ReadSameFileConcurrentlyUsingMultiRangeDownloader(t *testin
...
+					taskID := gID*perGoRoutineAddCount + j
+					// Randomize offset/limit slightly to ensure varied request patterns.
+					offset := int64((taskID * 2048) % (len(content) - 4096))
+					limit := int64(1024 + rand.Intn(2048))
+
+					r := &rangeRes{offset: offset, limit: limit}
+					results[taskID] = r
+
+					reader.Add(&r.buf, offset, limit, func(o, l int64, err error) {
+						r.err = err
```
krishnamd-jkp, line 594: Can you set the offset and limit as well here? And maybe compare them with the original ones just to be sure.

storage/retry_conformance_test.go
```diff
@@ -316,6 +318,133 @@ var methods = map[string][]retryFunc{
...
+			// Valid range must succeed
+			if errValid != nil {
+				return fmt.Errorf("valid range failed unexpectedly: %v", errValid)
+			}
+			if !bytes.Equal(bufValid.Bytes(), randomBytesToWrite[:3]) {
+				return fmt.Errorf("valid range data mismatch")
+			}
+
+			// Invalid range must report OutOfRange
+			if status.Code(errInvalid) != codes.OutOfRange {
```
krishnamd-jkp: After the fix, the error won't be a gRPC error but a custom client thrown error. Can we just check for non-nil here for now?

storage/integration_test.go
tritone: wg.Done needs to be called for each iteration so this seems correct as is.

storage/integration_test.go
```diff
@@ -550,63 +550,59 @@ func TestIntegration_ReadSameFileConcurrentlyUsingMultiRangeDownloader(t *testin
...
-			})
-		}
-
-		// Wait for all top-level goroutines to complete
-		if err := eG.Wait(); err != nil {
-			t.Errorf("Possible deadlock or race condition detected during concurrent Add calls: %v\n", err)
+			}(i)
 		}
 
+		// Wait for all range-specific callbacks to complete.
```
tritone: This actually just waits for each range to be added, not for the callbacks to be called. reader.Wait() waits until all work is complete.

storage/retry_conformance_test.go
tritone: I think the out-of-range behavior needs to change in order for the test to pass consistently; otherwise it's a permanent error.

tritone (dismissed):
Thanks for adding these, couple comments but generally looks good.

storage/integration_test.go
krishnamd-jkp: Oh right!

storage/integration_test.go
cpriti-os: Explained

storage/integration_test.go
cpriti-os: Done

storage/integration_test.go
cpriti-os: Agreed, clarified.

storage/retry_conformance_test.go
cpriti-os: Kept the test to only check for invalid range.

krishnamd-jkp (approved):

tritone (approved):

PR #13555, opened 2026-01-08
chore: remove redundant isRetryable function in MRD

krishnamd-jkp:


storage/internal/test/conformance/retry_tests.json
```diff
@@ -40,324 +43,6 @@
       ],
       "preconditionProvided": false,
       "expectSuccess": true
-    },
-    {
```
cpriti-os, line 44: revert these?

PR #13576, opened 2026-01-12
bigtable/benchmark_test: use idiomatic range in loop

12ya:


12ya:
put all in 1 PR: #13577

PR #13585, opened 2026-01-12
fix: remove mandatory attrs response in MRD

krishnamd-jkp:
This removes mandatory check on metadata in the response from the server. Creation of multiple MRDs will not work if mandatory check is present. If read handle is provided, the server might not return metadata, so removing the mandatory check

### DIFF
--- a/storage/grpc_reader_multi_range.go
+++ b/storage/grpc_reader_multi_range.go
@@ -115,7 +115,9 @@ func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params 
 			manager.wg.Wait()
 			return nil, manager.permanentErr
 		}
-		mrd.Attrs = *manager.attrs
+		if manager.attrs != nil {
+			mrd.Attrs = *manager.attrs
+		}
 		return mrd, nil
 	case <-ctx.Done():
 		cancel()
@@ -487,7 +489,7 @@ func (m *multiRangeDownloaderManager) handleAddCmd(ctx context.Context, cmd *mrd
 	}
 	m.readIDCounter++
 
-	// Attributes should be ready if we are processing Add commands
+	// Convert to positive offset only if attributes are available.
 	if m.attrs != nil && req.offset < 0 {
 		err := m.convertToPositiveOffset(req)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


tritone (commented):
Thanks for your work on this, a couple thoughts...
1. We should add an integration test for this case (creating MRD using a read handle).
2. We should get persisted size in the first response even with the read handle correct? In that case I would expect us to be able to convert to positive offset using that.
3. Are there deadlocks as a result of waiting on the attrs in NewMultiRangeDownloader?

Can discuss all this via chat and/or when we meet.

storage/grpc_reader_multi_range.go
```diff
@@ -488,7 +488,7 @@ func (m *multiRangeDownloaderManager) handleAddCmd(ctx context.Context, cmd *mrd
 	m.readIDCounter++
 
 	// Attributes should be ready if we are processing Add commands
```
cpriti-os: Should we make this comment clearer? We will convert only if attributes is available?

storage/grpc_reader_multi_range.go
```diff
@@ -519,21 +519,12 @@ func (m *multiRangeDownloaderManager) convertToPositiveOffset(req *rangeRequest)
...
-		m.failRange(req, err)
-		return err
-	}
-	start := objSize + req.offset
-	if start < 0 {
-		start = 0
-	}
+	start := max(objSize+req.offset, 0)
 	req.offset = start
-	req.length = objSize - start
```
cpriti-os, line 536: Isn't this required if length specified is 0

storage/integration_test.go
```diff
@@ -573,14 +574,23 @@ func TestIntegration_MRDWithReadHandle(t *testing.T) {
 			res2.limit = y
 			res2.err = err
 		})
+		mrd2.Add(&res3.buf, negativeOffset, limit, func(x, y int64, err error) {
```
cpriti-os, line 577: add another case with limit 0? To maintain what was done before we should read till EOF in that case?

storage/grpc_reader_multi_range.go
krishnamd-jkp: Fixed

storage/integration_test.go
krishnamd-jkp: Done

cpriti-os (approved):

PR #13592, opened 2026-01-13
chore(spanner): remove session pool and only use multiplexed sessions

rahul2393:
**Remove Session Pool**
  Removes the session pool from the Go Spanner client. The client now only uses a single multiplexed session for all operations. Multiplexed sessions have been the default since v1.73.0, but after this change, it is no longer possible to fall
  back to regular sessions.

 **Observability: Fewer Sessions on Spanner**
  After this change, the client no longer creates and initializes a session pool (default: 100 sessions) when a Client is created. The number of sessions observed on Spanner will be significantly lower, as each Client now only creates 1
  multiplexed session instead of a pool of 100 regular sessions.

   **What this means for users**:
  - Spanner session count metrics will drop significantly (from ~100 per client to 1).
  - SessionPoolConfig options like MinOpened, MaxOpened, MaxIdle, MaxBurst, WriteSessions, HealthCheckWorkers, and HealthCheckInterval are now deprecated and ignored. Only MultiplexSessionCheckInterval remains active.
  - The DefaultSessionPoolConfig no longer sets pool-related defaults.
  - InactiveTransactionRemovalOptions is deprecated and ignored.

   **No Public API Breaking Changes**
  All removed code is internal. The SessionPoolConfig struct and its fields are retained (but deprecated) for backward compatibility. Existing code that sets SessionPoolConfig values will continue to compile but the values will have no effect.

  **Code Changes**
  This removes ~7,300 lines of code and simplifies large parts of the codebase:
  - Removed the sessionPool struct, session health checking, session recycling, and batch session creation logic.
  - Renamed internal sessionPool to sessionManager to better reflect its new purpose.
  - Removed isMultiplexed checks and fallback paths throughout the transaction layer, as all sessions are now multiplexed.
  - Removed session-pool-specific OpenCensus/OpenTelemetry metrics (maxAllowedSessionsCount, maxInUseSessionsCount).
  - Simplified sessionHandle (removed valid field, changed to RWMutex for read-heavy access pattern).
  - Updated or removed tests that were specific to session pool behavior.


spanner/internal/testutil/inmem_spanner_server.go
```diff
@@ -775,48 +754,6 @@ func (s *inMemSpannerServer) CreateSession(ctx context.Context, req *spannerpb.C
 	return session, nil
 }
 
-func (s *inMemSpannerServer) BatchCreateSessions(ctx context.Context, req *spannerpb.BatchCreateSessionsRequest) (*spannerpb.BatchCreateSessionsResponse, error) {
```
olavloite, line 778: We should not remove this function. `BatchCreateSessions` is still a valid server RPC for Spanner.

spanner/client.go
```diff
@@ -156,13 +156,6 @@ func createGCPMultiEndpoint(cfg *grpcgcp.GCPMultiEndpointOptions, config ClientC
 					AffinityKey: "name",
 				},
 			},
-			{
```
olavloite, line 159: I don't think we should remove this. The fact that the client does not call this RPC anymore, does not mean that customer applications are guaranteed not to call it (e.g. by unwrapping the underlying gRPC client).

spanner/client.go
```diff
@@ -524,16 +517,11 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
...
+	// These are no longer used as multiplexed sessions are now the only option.
 	if config.MaxOpened == 0 {
 		config.MaxOpened = uint64(pool.Num() * 100)
 	}
-	if config.MaxBurst == 0 {
-		config.MaxBurst = DefaultSessionPoolConfig.MaxBurst
-	}
-	if config.incStep == 0 {
-		config.incStep = DefaultSessionPoolConfig.incStep
-	}
```
olavloite, line 536: I think that we should be consistent and either remove all default configuration options that are being set here, or none of them. (And I think that keeping them as-is is the safest, so that means undoing this change)

spanner/client.go
```diff
@@ -554,47 +542,10 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
...
-		config.enableMultiplexedSessionForPartitionedOps = true
-		config.SessionPoolConfig.MinOpened = experimentalHostMinSessions
-	}
-	// Do not initialize the session pool with any regular sessions if multiplexed sessions have been enabled for all
-	// operations, and the application has not configured a custom number of min sessions.
-	if config.enableMultiplexSession && config.enableMultiplexedSessionForRW && config.enableMultiplexedSessionForPartitionedOps && config.MinOpened == DefaultSessionPoolConfig.MinOpened {
-		config.SessionPoolConfig.MinOpened = 0
-	}
+	// MinOpened is no longer used as regular sessions have been removed.
+	config.SessionPoolConfig.MinOpened = 0
```
olavloite: I don't think that we should always set this to zero. If the application has set this value to for example 50, then it should remain 50, even though it is not used.

spanner/client.go
```diff
@@ -630,16 +580,10 @@ Directpath enabled: %v
 End To End Tracing enabled: %v
 Built-in metrics enabled: %v
 gRPC metrics enabled: %v
-Min Sessions: %v
-Max Sessions: %v
-Multiplexed session enabled: %v
-Multiplexed session enabled for RW: %v
-Multiplexed session enabled for Partition Ops: %v
+Multiplexed session enabled: true (always - session pool removed)
```
olavloite: I would suggest that we keep this simple like this:

```suggestion
Multiplexed session enabled: true
```

The reason is that any new customers would not have any context on any session pool, so this would not have much meaning to them.

spanner/client.go
```diff
@@ -612,7 +563,6 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
 	sc.mu.Unlock()
 
 	// Create a session pool.
-	config.SessionPoolConfig.sessionLabels = sessionLabels
 	sp, err := newSessionPool(sc, config.SessionPoolConfig)
```
olavloite: Maybe we should rename this function?

spanner/client.go
olavloite: Also update this

olavloite: Also update this

olavloite: Are these still used? Or can they be removed?

olavloite: Change to explicitly take a multiplexed session

olavloite: (here and elsewhere): The `replaceSession` functions can be removed. They are only relevant for regular sessions.

olavloite, line 1368: This should also be removed (both here and elsewhere). Replacing a session based on `Session not found` is only relevant for regular sessions.

spanner/client.go
rahul2393: incStep was private field so it was safe to remove it, others are publicly expose fields so I kept it

spanner/client.go
rahul2393: - openSessionCount - Reports if multiplexed session exists (0 or 1)
- getSessionTimeoutsCount - Timeout waiting for multiplexed session
- acquiredSessionsCount - Multiplexed session acquired count
- releasedSessionsCount - Multiplexed session released count
- gfeLatency - GFE latency histogram
- gfeHeaderMissingCount - GFE header missing count

rahul2393: sh.recycle() still reports `releasedSessionsCount` metric for Multiplexed session released count

spanner/client.go
rahul2393: I added them back.
But how will the customer unwrap it? since both grpcSpannerClient and spannerClient interfaces are private, 
I removed them to make sure nothing in the client library accidentally calls it.

spanner/client.go
rahul2393: I removed the rest

spanner/client_test.go
```diff
@@ -728,149 +727,32 @@ func TestClient_Single_InvalidArgument(t *testing.T) {
...
-		}
-		if err != nil {
-			t.Fatal(err)
-		}
-		rowCount++
-	}
-	if rowCount != SelectSingerIDAlbumIDAlbumTitleFromAlbumsRowCount {
-		t.Fatalf("row count mismatch\nGot: %v\nWant: %v", rowCount, SelectSingerIDAlbumIDAlbumTitleFromAlbumsRowCount)
-	}
+	t.Skip("Session pool has been removed. Session-not-found retry behavior is no longer applicable with multiplexed sessions.")
```
olavloite: As far as I'm concerned, we can just remove these tests directly. (Or do that in a follow-up PR, but we should eventually remove them, instead of skipping them)

spanner/client_test.go
```diff
@@ -4466,73 +3768,7 @@ func TestClient_WriteStructWithCustomTypes(t *testing.T) {
 }
 
 func TestReadWriteTransaction_ContextTimeoutDuringCommit(t *testing.T) {
-	t.Parallel()
```
olavloite, line 4469: I would prefer that we keep this test (potentially with some changes). The main thing that is being tested here is what happens when a DEADLINE_EXCEEDED error happens during a Commit. This is a bit of a special case, because it means that we do not know whether the transaction actually succeeded or not, as we do not know whether the Commit reached the server or not.

spanner/grpc_client.go
```diff
@@ -49,7 +49,6 @@ type spannerClient interface {
 	Close() error
 	Connection() *grpc.ClientConn
 	CreateSession(context.Context, *spannerpb.CreateSessionRequest, ...gax.CallOption) (*spannerpb.Session, error)
-	BatchCreateSessions(context.Context, *spannerpb.BatchCreateSessionsRequest, ...gax.CallOption) (*spannerpb.BatchCreateSessionsResponse, error)
```
olavloite, line 52: I don't think that we should remove the RPC implementation, even though the client does not call the RPC anymore. This could in theory be used by customer applications.

(E.g. ListSessions is also not used by the client library, but is still present in this file)

spanner/regression_test.go
```diff
@@ -72,8 +72,7 @@ func directPathEnabled() bool {
 // destructive context augmentation call.
 // Please see https://github.com/googleapis/google-cloud-go/issues/11656
 func TestAllHeadersForwardedAppropriately(t *testing.T) {
-	// 0. Turn off session multiplexing per #11308.
-	t.Setenv("GOOGLE_CLOUD_SPANNER_MULTIPLEXED_SESSIONS", "0")
+	t.Skip("session pool has been removed - this test validates session pool behavior with BatchCreateSessions")
```
olavloite: Can we modify this test to check for a `CreateSession` call instead? The main thing that it is testing is that the correct headers are being sent, and not the session pool behavior.

spanner/request_id_header_test.go
```diff
@@ -1749,86 +1731,7 @@ func TestRequestIDHeader_Commit_ContextDeadlineExceeded(t *testing.T) {
 }
 
 func TestRequestIDHeader_VerifyChannelNumber(t *testing.T) {
-	t.Parallel()
```
olavloite, line 1752: I don't think that we should remove this test. It seems to work with multiplexed sessions as well, and the main thing it is checking is that we are actually using multiple gRPC channels, and that this is correctly set in the request IDs.

spanner/session.go
```diff
@@ -34,135 +30,114 @@ import (
...
+	// session is a pointer to a session object.
 	session *session
-	// client is the RPC channel to Cloud Spanner. It is set only once during session acquisition.
+	// client is the RPC channel to Cloud Spanner.
 	client spannerClient
-	// checkoutTime is the time the session was checked out of the pool.
+	// checkoutTime is the time the session was checked out.
 	checkoutTime time.Time
-	// lastUseTime is the time the session was last used after checked out of the pool.
+	// lastUseTime is the time the session was last used.
```
olavloite: I would expect that we can get rid of this as well (please feel free to leave this in place if that complicates things, but I would not expect it to be needed anymore, as mux sessions don't expire based on when they were last used)

spanner/session.go
```diff
@@ -34,135 +30,114 @@ import (
...
-	stack []byte
 	// eligibleForLongRunning tells if the inner session is eligible to be long-running.
 	eligibleForLongRunning bool
-	// if the inner session object is long-running then the stack gets logged once.
-	isSessionLeakLogged bool
 }
 
-// recycle gives the inner session object back to its home session pool. It is
-// safe to call recycle multiple times but only the first one would take effect.
+// recycle returns the session handle back to the pool.
```
olavloite: Maybe something like this:

```suggestion
// recycle marks the session handle as no longer in use
```

spanner/session.go
```diff
@@ -34,135 +30,114 @@ import (
...
 	lastUseTime time.Time
-	// trackedSessionHandle is the linked list node which links the session to
-	// the list of tracked session handles. trackedSessionHandle is only set if
-	// TrackSessionHandles has been enabled in the session pool configuration.
-	trackedSessionHandle *list.Element
-	// stack is the call stack of the goroutine that checked out the session
-	// from the pool. This can be used to track down session leak problems.
-	stack []byte
 	// eligibleForLongRunning tells if the inner session is eligible to be long-running.
 	eligibleForLongRunning bool
```
olavloite: This is also no longer needed (and if there are references to this somewhere in the code, then that is an indication that there is more that can be cleaned up)

spanner/session.go
```diff
@@ -182,34 +157,23 @@ func (sh *sessionHandle) getTransactionID() transactionID {
 	return sh.session.tx
 }
 
-// destroy destroys the inner session object. It is safe to call destroy
-// multiple times and only the first call would attempt to
-// destroy the inner session object.
+// destroy destroys the inner session object.
 func (sh *sessionHandle) destroy() {
```
olavloite: I would expect that it should be possible to remove this as well. When would you need to call this?

spanner/session.go
```diff
@@ -223,47 +187,24 @@ func (sh *sessionHandle) updateLastUseTime() {
...
-	nextCheck time.Time
-	// checkingHelath is true if currently this session is being processed by
-	// health checker. Must be modified under health checker lock.
-	checkingHealth bool
 	// md is the Metadata to be sent with each request.
 	md metadata.MD
-	// tx contains the transaction id if the session has been prepared for
-	// write.
+	// tx contains the transaction id if the session has been prepared for write.
 	tx transactionID
```
olavloite: I think that this can be removed (it could probably have been removed even before this cleanup)

spanner/session.go
```diff
@@ -350,438 +238,207 @@ func (s *session) setTransactionID(tx transactionID) {
...
-// getNextCheck returns the timestamp for next healthcheck on the session.
-func (s *session) getNextCheck() time.Time {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	return s.nextCheck
-}
-
 // recycle turns the session back to its home session pool.
 func (s *session) recycle() {
 	s.setTransactionID(nil)
```
olavloite: This can probably be removed. I think that it is a left-over from when we prepared transactions in the pool

spanner/session.go
```diff
@@ -350,438 +238,207 @@ func (s *session) setTransactionID(tx transactionID) {
...
-	}
-	s.pool.decNumInUseLocked(context.Background())
-	s.pool.mu.Unlock()
+	// Multiplexed sessions don't need to be recycled to a pool
 }
 
-// destroy removes the session from its home session pool, healthcheck queue
-// and Cloud Spanner service.
+// destroy removes the session from Cloud Spanner service.
 func (s *session) destroy(isExpire, wasInUse bool) bool {
```
olavloite: When would this be used? Is it maybe an indication that there is more code that could be cleaned up?

spanner/session.go
```diff
@@ -350,438 +238,207 @@ func (s *session) setTransactionID(tx transactionID) {
...
+	if s.isMultiplexed {
+		return
+	}
 	err := s.client.DeleteSession(contextWithOutgoingMetadata(ctx, s.md, true), &sppb.DeleteSessionRequest{Name: s.getID()})
-	// Do not log DeadlineExceeded errors when deleting sessions, as these do
-	// not indicate anything the user can or should act upon.
 	if err != nil && ErrCode(err) != codes.DeadlineExceeded {
 		logf(s.logger, "Failed to delete session %v. Error: %v", s.getID(), err)
 	}
 }
```
olavloite: Same as above: Can we remove this as well?

spanner/session.go
```diff
@@ -1067,57 +573,29 @@ func (p *sessionPool) close(ctx context.Context) {
...
 
-// newSessionHandle creates a new session handle for the given session for this
-// session pool. The session handle will also hold a copy of the current call
-// stack if the session pool has been configured to track the call stacks of
-// sessions being checked out of the pool.
-func (p *sessionPool) newSessionHandle(s *session) (sh *sessionHandle) {
+// newSessionHandle creates a new session handle for the given session.
+func (p *sessionManager) newSessionHandle(s *session) (sh *sessionHandle) {
 	sh = &sessionHandle{session: s, checkoutTime: time.Now(), lastUseTime: time.Now()}
 	if s.isMultiplexed {
```
olavloite: Remove this? Or maybe refactor this a bit so that an error is returned if it is called with a non-multiplexed session?

spanner/session.go
```diff
@@ -1138,150 +615,27 @@ func (p *sessionPool) getRoundRobinClient() spannerClient {
...
-				return nil, err
-			}
-			p.mu.Unlock()
-		}
-	}
+// take returns a multiplexed session.
+//
+// Deprecated: Regular session pool has been removed. This method now returns
+// a multiplexed session for backward compatibility.
+func (p *sessionManager) take(ctx context.Context) (*sessionHandle, error) {
```
olavloite: It should be possible to remove this as well. That would again force a cleanup of all references to this function (which is unexported, so no external usage)

spanner/session.go
```diff
@@ -1315,656 +663,76 @@ func (p *sessionPool) takeMultiplexed(ctx context.Context) (*sessionHandle, erro
...
-	}
-	return a
-}
-
 // sessionResourceType is the type name of Spanner sessions.
 const sessionResourceType = "type.googleapis.com/google.spanner.v1.Session"
 
-// isSessionNotFoundError returns true if the given error is a
-// `Session not found` error.
+// isSessionNotFoundError returns true if the given error is a `Session not found` error.
```
olavloite: This is probably no longer needed

olavloite (approved):

olavloite (approved):

PR #13596, opened 2026-01-14
chore(spanner): add experimental location-aware foundation classes

rahul2393:
- Adds experimental location routing base classes (recipe parsing, key-to-target encoding,
    range cache, channel finder, router).
  - Integrates routing into request lifecycle for reads/queries/transactions under
    GOOGLE_SPANNER_EXPERIMENTAL_LOCATION_API.
  - Adds golden coverage using textprotos:
      - finder_test.textproto
      - range_cache_test.textproto
      - recipe_test.textproto
  - final RPC integration with RPC calls will be tracked as follow-up.

### DIFF
--- a/spanner/key_recipe.go
+++ b/spanner/key_recipe.go
@@ -412,9 +412,14 @@ func (r *keyRecipe) encodeKeyInternal(finder valueFinder, keyType keyType) *targ
 			}
 			value, lookupStatus := r.resolvePartValue(part, finder, lookupIndex)
 			switch lookupStatus {
+			case valueLookupFound:
+				// Continue with encoding for the found value.
 			case valueLookupMissing:
 				state = encodeStateEndOfKeys
 			case valueLookupFailed:
+				state = encodeStateFailed
+			default:
+				// Defensive fallback if valueLookupStatus grows in the future.
 				state = encodeStateFailed
 			}
 			if lookupStatus != valueLookupFound {
@@ -432,10 +437,10 @@ func (r *keyRecipe) encodeKeyInternal(finder valueFinder, keyType keyType) *targ
 	}
 
 	if partIndex == len(r.parts) || (keyType != keyTypeFull && state == encodeStateEndOfKeys) {
-		switch keyType {
-		case keyTypePrefixSuccessor:
+		if keyType == keyTypePrefixSuccessor {
 			target.start = makePrefixSuccessor(target.start)
-		case keyTypeIndex:
+		}
+		if keyType == keyTypeIndex {
 			target.limit = makePrefixSuccessor(target.start)
 		}
 		return target
--- a/spanner/key_recipe_cache.go
+++ b/spanner/key_recipe_cache.go
@@ -103,8 +103,12 @@ func fingerprintExecuteSQLRequest(req *sppb.ExecuteSqlRequest) uint64 {
 	for _, name := range paramNames {
 		hashString(h, name)
 		if typ, ok := req.GetParamTypes()[name]; ok {
-			bytes, _ := proto.MarshalOptions{Deterministic: true}.Marshal(typ)
-			hashBytes(h, bytes)
+			typeBytes, err := proto.MarshalOptions{Deterministic: true}.Marshal(typ)
+			if err != nil {
+				hashBytes(h, nil)
+				continue
+			}
+			hashBytes(h, typeBytes)
 		} else {
 			hashUint64(h, uint64(valueKindCase(req.GetParams().GetFields()[name])))
 		}
@@ -391,6 +395,8 @@ func valueKindCase(value *structpb.Value) int32 {
 	}
 }
 
+// lruCache is a non-thread-safe fixed-size LRU map.
+// Callers must provide external synchronization for all operations.
 type lruCache[K comparable, V any] struct {
 	maxSize int
 	items   map[K]*list.Element
--- a/spanner/batch.go
+++ b/spanner/batch.go
@@ -298,7 +298,7 @@ func (t *BatchReadOnlyTransaction) Execute(ctx context.Context, p *Partition) *R
 	// Read or query partition.
 	if p.rreq != nil {
 		rpc = func(ctx context.Context, resumeToken []byte, opts ...gax.CallOption) (streamingReceiver, error) {
-			client, err := client.StreamingRead(ctx, &sppb.ReadRequest{
+			req := &sppb.ReadRequest{
 				Session:             p.rreq.Session,
 				Transaction:         p.rreq.Transaction,
 				Table:               p.rreq.Table,
@@ -310,7 +310,11 @@ func (t *BatchReadOnlyTransaction) Execute(ctx context.Context, p *Partition) *R
 				ResumeToken:         resumeToken,
 				DataBoostEnabled:    p.rreq.DataBoostEnabled,
 				DirectedReadOptions: p.rreq.DirectedReadOptions,
-			}, opts...)
+			}
+			if t.locationRouter != nil {
+				t.locationRouter.prepareReadRequest(req)
+			}
+			client, err := client.StreamingRead(ctx, req, opts...)
 			if err != nil {
 				return client, err
 			}
@@ -327,7 +331,7 @@ func (t *BatchReadOnlyTransaction) Execute(ctx context.Context, p *Partition) *R
 		}
 	} else {
 		rpc = func(ctx context.Context, resumeToken []byte, opts ...gax.CallOption) (streamingReceiver, error) {
-			client, err := client.ExecuteStreamingSql(ctx, &sppb.ExecuteSqlRequest{
+			req := &sppb.ExecuteSqlRequest{
 				Session:             p.qreq.Session,
 				Transaction:         p.qreq.Transaction,
 				Sql:                 p.qreq.Sql,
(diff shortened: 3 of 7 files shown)


spanner/ssformat.go
```diff
@@ -0,0 +1,398 @@
...
+// newTargetRange creates a new targetRange with the given boundaries.
+func newTargetRange(start, limit []byte, approximate bool) *targetRange {
+	return &targetRange{
+		start:       start,
+		limit:       limit,
+		approximate: approximate,
+	}
+}
+
+// isPoint returns true if this range represents a single key (point lookup).
```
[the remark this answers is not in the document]
rahul2393: The semantics are:
  
  | Representation    | Meaning |
| -------- | ------- |
| start: K, limit: empty  | Point lookup for exactly key K   |
| start: K1, limit: K2 | Range [K1, K2)        |
  
  When merging point K into range [start, limit):
  - If K >= limit, extend to include K via limit = successor(K)
  - Result: range [min(start, K), successor(K)) which correctly includes point K

spanner/ssformat.go
```diff
@@ -0,0 +1,398 @@
...
+	}
+	if other.isPoint() {
+		if bytes.Compare(other.start, r.limit) >= 0 {
+			r.limit = makePrefixSuccessor(other.start)
+		}
+	} else if bytes.Compare(other.limit, r.limit) > 0 {
+		r.limit = other.limit
+	}
+	r.approximate = r.approximate || other.approximate
+}
```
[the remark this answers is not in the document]
rahul2393: - isPoint() means "single key lookup", NOT "unbounded range".
- An empty limit indicates a point query for exactly the key in start.

Regarding your example (merging two points P and Q where P < Q):
  
r = {start: P, limit: empty}
other = {start: Q, limit: empty}
  
The result is actually {start: P, limit: successor(Q)}, which correctly covers both points.

spanner/key_recipe_cache.go
```diff
@@ -0,0 +1,416 @@
...
+
+func trimStringKeyMap[V any](m map[string]V, maxSize int) {
+	if len(m) <= maxSize {
+		return
+	}
+	for key := range m {
+		delete(m, key)
+		break
+	}
+}
```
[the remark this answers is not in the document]
rahul2393: Updated the use LRU cache

spanner/key_recipe.go
```diff
@@ -0,0 +1,656 @@
...
+			value, lookupStatus := r.resolvePartValue(part, finder, lookupIndex)
+			switch lookupStatus {
+			case valueLookupMissing:
+				state = encodeStateEndOfKeys
+			case valueLookupFailed:
+				state = encodeStateFailed
+			}
+			if lookupStatus != valueLookupFound {
+				break
+			}
```
olavloite, line 427: nit: could we maybe change this block to this for readability (it makes it a bit clearer that we are entering a state where we stop the encoding):

```suggestion
			if lookupStatus != valueLookupFound {
				switch lookupStatus {
				case valueLookupMissing:
					state = encodeStateEndOfKeys
				case valueLookupFailed:
					state = encodeStateFailed
				default:
					// Should not happen unless an additional value is added to the enum.
					state = encodeStateFailed
				}
				// Stop
				break
			}
```

spanner/key_recipe.go
```diff
@@ -0,0 +1,656 @@
...
+		return value, valueLookupFound
+	}
+	current := value
+	for _, structIndex := range part.structIdentifiers {
+		listValue := current.GetListValue()
+		if listValue == nil || int(structIndex) < 0 || int(structIndex) >= len(listValue.GetValues()) {
+			return nil, valueLookupFailed
+		}
+		current = listValue.GetValues()[structIndex]
+	}
```
olavloite, line 386: This seems not to be covered by any tests (at least, if I run all tests and enable coverage reporting, then this shows up as not covered). Could that be right?

spanner/key_recipe.go
```diff
@@ -0,0 +1,656 @@
...
+			lookupIndex := valueIndex
+			if part.shouldConsumeValueIndex() {
+				valueIndex++
+			}
+			value, lookupStatus := r.resolvePartValue(part, finder, lookupIndex)
+			switch lookupStatus {
+			case valueLookupMissing:
+				state = encodeStateEndOfKeys
+			case valueLookupFailed:
+				state = encodeStateFailed
```
olavloite, line 420: We don't seem to do anything special with the `encodeStateFailed`. If we get into that state, we still return something that seems to be a 'valid' result. There also does not seem to be any tests that cover this. Is it basically impossible to get into that state, and is this part of the code only there for 'safety'? Or could it actually be that we get into this state? If so, could we have a test for it (potentially in a follow-up PR)?

spanner/key_recipe.go
```diff
@@ -0,0 +1,656 @@
...
+		case keyRecipePartInvalid:
+			state = encodeStateFailed
+		}
+		if state != encodeStateOK {
+			break
+		}
+	}
+
+	if partIndex == len(r.parts) || (keyType != keyTypeFull && state == encodeStateEndOfKeys) {
+		switch keyType {
```
olavloite: nit: should we consider using a `if` statement here? `switch` is normally used when you want to cover all (or almost all) possible values. This only covers 2 of the 4 possible enum values. (The `if` statement above only partly covers the `keyTypeFull`, as that part is only evaluated if `partIndex != len(r.parts)`)

spanner/key_recipe_cache.go
```diff
@@ -0,0 +1,449 @@
...
+
+	paramNames := make([]string, 0, len(req.GetParams().GetFields()))
+	for name := range req.GetParams().GetFields() {
+		paramNames = append(paramNames, name)
+	}
+	sort.Strings(paramNames)
+	for _, name := range paramNames {
+		hashString(h, name)
+		if typ, ok := req.GetParamTypes()[name]; ok {
+			bytes, _ := proto.MarshalOptions{Deterministic: true}.Marshal(typ)
```
olavloite: nit: should we at least check that `bytes` is not nil to prevent a panic if marshalling for some weird reason fails?

spanner/key_recipe_cache.go
```diff
@@ -0,0 +1,449 @@
...
+		return nil
+	}
+	c.mu.Lock()
+	recipe, _ := c.schemaRecipes.Get(tableName)
+	c.mu.Unlock()
+	if recipe == nil {
+		return nil
+	}
+	return recipe.mutationToTargetRange(mutation)
+}
```
olavloite, line 254: This function does not appear to be covered by any of the test cases.

spanner/key_recipe_cache.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+}
+
+func valueKindCase(value *structpb.Value) int32 {
+	if value == nil || value.GetKind() == nil {
+		return 0
+	}
+	switch value.GetKind().(type) {
+	case *structpb.Value_NullValue:
+		return 1
```
olavloite, line 382: nit: this condition does not seem to be covered by any of the tests. Does none of the tests use a `null` value for a parameter / key? If not, should we add that?

spanner/key_recipe_cache.go
```diff
@@ -0,0 +1,449 @@
...
+	case *structpb.Value_StructValue:
+		return 5
+	case *structpb.Value_ListValue:
+		return 6
+	default:
+		return 0
+	}
+}
+
+type lruCache[K comparable, V any] struct {
```
olavloite, line 400: nit: can we add a clarifying comment to this struct that it is not thread safe? It is currently only used by `keyRecipeCache`, and inside that struct only accessed while holding the mutex of `keyRecipeCache`, so the current usage is safe. But it could in the future be used for other purposes within the same package.

spanner/location_router.go
```diff
@@ -0,0 +1,74 @@
...
+type locationRouter struct {
+	finder *channelFinder
+}
+
+func isExperimentalLocationAPIEnabled() bool {
+	enabled, _ := strconv.ParseBool(os.Getenv(experimentalLocationAPIEnvVar))
+	return enabled
+}
+
+func newLocationRouter() *locationRouter {
```
olavloite, line 37: `locationRouter` does not seem to be covered by any of the tests. (Or are they covered by the integration tests?)

spanner/read.go
```diff
@@ -74,6 +74,7 @@ func stream(
 		setTimestamp,
 		release,
 		gsc,
+		nil,
```
olavloite, line 77: Why is this `nil` here? AFAICT, the only place where `stream(..)` is called directly is in `batch.go`. Does this mean that we never get a cache update during a (batch) read-only transaction?

spanner/key_recipe.go
rahul2393: encodeStateFailed is intentional and reachable (e.g. invalid struct_identifiers path or invalid typed value
such as malformed UUID). In that state we intentionally fall back to an approximate routing range (not an exact key), so behavior remains safe. I added focused tests for these failed paths.

spanner/key_recipe.go
rahul2393: The switch is intentionally partial because only PrefixSuccessor/Index mutate output here, but I agree if-
  statements are clearer

spanner/key_recipe_cache.go
rahul2393: There wasn’t a panic path here in practice, but I made the marshal handling explicit now: on unexpected
  marshal failure we hash a deterministic empty payload and continue.

spanner/key_recipe_cache.go
rahul2393: Added explicit null-param coverage for both fingerprinting and prepared-query shape matching (untyped NULL kind path), so this branch is now tested.

spanner/key_recipe_cache.go
rahul2393: Good nit; this makes future reuse safer.

spanner/location_router.go
rahul2393: Added focused unit tests for location_router (flag parsing, nil safety, and prepare/observe flows via ResultSet/PartialResultSet cache updates), so this is now directly covered and not only via integration paths.

spanner/read.go
rahul2393: Thanks, added the support

PR #13600, opened 2026-01-14
chore(storage): add chunking, composing and cleanup logic for PCU

cpriti-os:
#### Asynchronous Chunking and Uploads
* Data written by the user is buffered into fixed-size parts. Once a part buffer is full, or during Close(), it's dispatched for asynchronous upload via a dedicated pool of worker goroutines.
* The pcuState.write() method handles buffering incoming data, dispatching full buffers, and acquiring new buffers from a pool (bufferCh).
* pcuState.flushCurrentBuffer() encapsulates the logic for taking a filled buffer, incrementing the part number (s.partNum++), creating an uploadTask, and sending it to the uploadCh for processing by a worker.

#### Recursive Composition for Large Objects
* To handle objects larger than the maximum allowed composite components (currently 32), a multi-level composition strategy is implemented
* pcuState.composeParts() manages the recursive composition process, creating intermediate objects (up to 32 components per compose operation) until a final object is formed.
* Each level of intermediate composition can occur in parallel using goroutines, with errors aggregated and handled.
* Temporary object names for parts and intermediates are generated using a default naming strategy, including random bytes (generateRandomBytes) to prevent hot-spotting

#### Robust Cleanup
* Temporary part and intermediate objects are actively managed and cleaned up upon successful completion or error, based on a configurable cleanupStrategy (e.g., cleanupAlways, cleanupOnSuccess).
* pcuState.doCleanup() orchestrates the deletion of all temporary parts and intermediate composed objects, performing these deletions concurrently using a semaphore to limit  goroutines 
* Deletion errors are recorded in failedDeletes and can be retrieved in future.

### DIFF
--- a/storage/pcu.go
+++ b/storage/pcu.go
@@ -373,6 +373,10 @@ func (s *pcuState) resultCollector() {
 			s.mu.Lock()
 			s.partMap[result.partNumber] = result.handle
 			s.mu.Unlock()
+		} else {
+			// Both are nil: this is an impossible state that indicates a logical error.
+			// Setting an error to prevent silent data corruption.
+			s.setError(fmt.Errorf("upload result missing both error and handle for part %d", result.partNumber))
 		}
 	}
 }
@@ -448,13 +452,14 @@ func (s *pcuState) flushCurrentBuffer() error {
 		return s.firstErr
 	}
 	s.partNum++
+	pNum := s.partNum
+	s.mu.Unlock()
+
 	task := uploadTask{
-		partNumber: s.partNum,
+		partNumber: pNum,
 		buffer:     s.currentBuffer,
 		size:       s.bytesBuffered,
 	}
-	s.mu.Unlock()
-
 	// Clear current state so the next Write call picks up a fresh buffer.
 	s.currentBuffer = nil
 	s.bytesBuffered = 0


storage/pcu.go
```diff
@@ -196,6 +198,11 @@ type pcuState struct {
 
 	// Function to upload a part; can be overridden for testing.
 	uploadPartFn func(s *pcuState, task uploadTask) (*ObjectHandle, *ObjectAttrs, error)
+	// Function to delete an object; can be overridden for testing.
+	deleteFn       func(ctx context.Context, h *ObjectHandle) error
+	doCleanupFn    func(s *pcuState)
```
krishnamd-jkp: Add comments similar to above functions.

storage/pcu.go
krishnamd-jkp: If this is true, we'll also have to send CRC32C along with this. Is this being handled?

storage/pcu.go
```diff
@@ -362,3 +376,260 @@ func (s *pcuState) setError(err error) {
...
+	if err != nil {
+		return 0, err
+	}
+
+	total := len(p)
+	for len(p) > 0 {
+		// Acquire a buffer from the pool if we don't have one.
+		if s.currentBuffer == nil {
+			// Fail-fast check before taking a new buffer.
+			s.mu.Lock()
```
krishnamd-jkp, line 414: Can you move this logic to a function. This is repeated at multiple places.

storage/pcu.go
```diff
@@ -362,3 +376,260 @@ func (s *pcuState) setError(err error) {
...
+	case <-s.ctx.Done():
+		// Return buffer to pool if we couldn't dispatch.
+		s.bufferCh <- task.buffer
+		return s.ctx.Err()
+	case s.uploadCh <- task:
+		return nil
+	}
+}
+
+func (s *pcuState) close() error {
```
krishnamd-jkp, line 483: When is close called? When writer.Close() is called? If so, make sure that this function is thread safe as close can be called parallelly. Also make sure it is idempotent

storage/pcu.go
```diff
@@ -362,3 +376,260 @@ func (s *pcuState) setError(err error) {
...
+			return levelErr
+		}
+		finalComps = nextLevel
+	}
+
+	// Final Compose
+	composer := s.w.o.ComposerFrom(finalComps...)
+	composer.ObjectAttrs = s.w.ObjectAttrs
+	composer.KMSKeyName = s.w.ObjectAttrs.KMSKeyName
+	composer.SendCRC32C = s.w.SendCRC32C
```
krishnamd-jkp, line 612: Same here. If SendCRC32C is true, checksum also has to be sent.

storage/pcu.go
```diff
@@ -362,3 +376,260 @@ func (s *pcuState) setError(err error) {
...
+				start := idx * maxComposeComponents
+				end := min(start+maxComposeComponents, len(finalComps))
+
+				// Level-based naming with hash to prevent exceeding 1024 bytes.
+				h := hex.EncodeToString(generateRandomBytes(4))
+				compName := path.Join(s.config.tmpObjectPrefix, fmt.Sprintf("int-%s-lv%d-%d", h, level, idx))
+
+				interHandle := s.w.o.c.Bucket(s.w.o.bucket).Object(compName)
+				composer := interHandle.ComposerFrom(finalComps[start:end]...)
+				// composer.SendCRC32C = s.w.SendCRC32C
```
krishnamd-jkp: Remove this?

storage/pcu.go
```diff
@@ -362,3 +376,260 @@ func (s *pcuState) setError(err error) {
...
+		wg.Wait()
+		close(failedCh)
+	}()
+
+	for h := range failedCh {
+		s.failedDeletes = append(s.failedDeletes, h)
+	}
+
+	if n := len(s.failedDeletes); n > 0 {
+		fmt.Printf("Warning: failed to delete %d temporary objects. Use FailedCleanupHandles() for manual retry.\n", n)
```
krishnamd-jkp: Logs can be ignored. Anyway we can still notify the user programmatically?

storage/pcu_test.go
```diff
@@ -544,6 +546,621 @@ func TestSetPartMetadata(t *testing.T) {
...
+			ctx := context.Background()
+			state := &pcuState{
+				ctx:      ctx,
+				started:  true,
+				bufferCh: make(chan []byte, 5),
+				uploadCh: make(chan uploadTask, 5),
+				config:   &parallelUploadConfig{partSize: partSize},
+			}
+
+			// Pre-fill buffer pool
```
krishnamd-jkp: Missing period. Here and other places too.

storage/pcu_test.go
```diff
@@ -544,6 +546,621 @@ func TestSetPartMetadata(t *testing.T) {
...
+}
+
+func TestPCUState_WriteStopsOnError(t *testing.T) {
+	state := &pcuState{
+		started:  true,
+		firstErr: fmt.Errorf("background failure"),
+	}
+
+	n, err := state.write([]byte("payload"))
+	if n != 0 || err == nil {
```
krishnamd-jkp, line 632: Check the returned error as well

storage/pcu_test.go
```diff
@@ -544,6 +546,621 @@ func TestSetPartMetadata(t *testing.T) {
...
+		{
+			name: "Map with items in reverse order",
+			partMap: map[int]*ObjectHandle{
+				3: handle3,
+				2: handle2,
+				1: handle1,
+			},
+			expectedOrder: []*ObjectHandle{handle1, handle2, handle3},
+		},
+		{
```
krishnamd-jkp, line 762: nit: can do away with some of the test cases here

krishnamd-jkp:
initial review

storage/pcu.go
cpriti-os: Done

storage/pcu.go
cpriti-os: This is redundant now since cleanup errors is not relevant for MPU like APIs

storage/pcu.go
cpriti-os: Done

storage/pcu_test.go
cpriti-os: Done

storage/pcu_test.go
cpriti-os: Done

storage/pcu.go
cpriti-os: There are just three lines that can be moved to a function but it can only possibly replace three instances, I think adding a function for so small a logic will not add to readability.

storage/pcu_test.go
cpriti-os: Done

storage/pcu.go
cpriti-os: Done

storage/pcu.go
```diff
@@ -362,3 +378,255 @@ func (s *pcuState) setError(err error) {
...
+		// fall back to a standard empty object creation.
+		if len(s.partMap) == 0 {
+			ow := s.w.o.NewWriter(s.w.ctx)
+			if ow == nil {
+				err = fmt.Errorf("failed to create writer for empty object")
+				return
+			}
+			ow.ObjectAttrs = s.w.ObjectAttrs
+			err = ow.Close()
+			return
```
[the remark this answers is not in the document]
cpriti-os: Done

storage/pcu_test.go
```diff
@@ -544,6 +546,598 @@ func TestSetPartMetadata(t *testing.T) {
...
+			name: "should send task when buffer has data",
+			setup: func() *pcuState {
+				return &pcuState{
+					ctx:           context.Background(),
+					partNum:       0,
+					bytesBuffered: 50,
+					currentBuffer: baseBuffer,
+					uploadCh:      make(chan uploadTask, 1),
+					bufferCh:      make(chan []byte, 1),
+					mu:            sync.Mutex{},
```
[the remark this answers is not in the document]
cpriti-os: This does not exist, resolving

storage/pcu.go
```diff
@@ -362,3 +378,255 @@ func (s *pcuState) setError(err error) {
...
+		go runDelete(h)
+	}
+	wg.Wait()
+}
+
+// Generates size random bytes.
+func generateRandomBytes(n int) []byte {
+	b := make([]byte, n)
+	_, _ = rand.Read(b)
+	return b
```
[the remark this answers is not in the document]
krishnamd-jkp: Since this is required for the flow to work, handle the error from rand.Read as well.

storage/pcu.go
```diff
@@ -179,7 +181,6 @@ type pcuState struct {
 	partMap map[int]*ObjectHandle
 	// Handles to intermediate composite objects, keyed by their object name.
 	intermediateMap map[string]*ObjectHandle
-	failedDeletes   []*ObjectHandle
```
krishnamd-jkp, line 182: Are we handling this elsewhere?

storage/pcu.go
krishnamd-jkp: Check s.ctx.Err() before sending it on the channel and also add a select statement selecting on ctx.Done() also to make sure this is not blocking.

krishnamd-jkp: Same here. Listen on ctx.Done() as well to make sure there is a listener or that the channel is not closed.

krishnamd-jkp, line 329: Are we always sure none of these pointers are nil. Just trying to avoid nil errors. Please check.

storage/pcu.go
```diff
@@ -300,16 +318,14 @@ func (s *pcuState) uploadPart(task uploadTask) (*ObjectHandle, *ObjectAttrs, err
 	pw := partHandle.NewWriter(s.ctx)
 	pw.ObjectAttrs.Name = partName
 	pw.ObjectAttrs.Size = task.size
-	pw.SendCRC32C = s.w.SendCRC32C
+	pw.DisableAutoChecksum = s.w.DisableAutoChecksum
 	pw.ChunkSize = 0 // Force single-shot upload for parts.
 	// Clear fields not applicable to parts or that are set by compose.
-	pw.ObjectAttrs.CRC32C = 0
 	pw.ObjectAttrs.MD5 = nil
```
krishnamd-jkp: Nit: Need not do it explicitly. Check if this can be removed.

storage/pcu.go
```diff
@@ -300,16 +318,14 @@ func (s *pcuState) uploadPart(task uploadTask) (*ObjectHandle, *ObjectAttrs, err
...
 	setPartMetadata(pw, s, task)
 
-	_, err := pw.Write(task.buffer[:task.size])
-	if err != nil {
-		pw.CloseWithError(err)
+	if _, err := pw.Write(task.buffer[:task.size]); err != nil {
+		_ = pw.CloseWithError(err)
```
krishnamd-jkp: This need not be done explicitly, right? I think writer sets the internal error automatically.

storage/pcu.go
krishnamd-jkp, line 372: Do we have to check a case where err is not nil and handle is nil?

storage/pcu.go
```diff
@@ -362,3 +378,259 @@ func (s *pcuState) setError(err error) {
...
+		return nil
+	}
+
+	// Capture state for the task while under lock, then release immediately.
+	s.mu.Lock()
+	if s.firstErr != nil {
+		s.mu.Unlock()
+		return s.firstErr
+	}
+	s.partNum++
```
krishnamd-jkp, line 454: Should this be mutex protected? Write is sequential anyways, right?

storage/pcu.go
```diff
@@ -362,3 +378,259 @@ func (s *pcuState) setError(err error) {
...
+				interHandle := s.w.o.c.Bucket(s.w.o.bucket).Object(compName)
+				composer := interHandle.ComposerFrom(finalComps[start:end]...)
+
+				attrs, err := s.composeFn(s.ctx, composer)
+				if err != nil {
+					errOnce.Do(func() { levelErr = err })
+					return
+				}
+
+				s.mu.Lock()
```
krishnamd-jkp: I think we can use a different mutex here and not the mutex used in other operations to decrease the contention for lock. Or maybe consider using a channel here.

storage/pcu.go
```diff
@@ -362,3 +378,259 @@ func (s *pcuState) setError(err error) {
...
+
+	runDelete := func(h *ObjectHandle) {
+		defer wg.Done()
+		sem <- struct{}{}
+		defer func() { <-sem }()
+
+		// Use WithoutCancel to ensure cleanup isn't killed by parent context cancellation
+		// Ignore cleanup errors here since its best effort and will rely on bucket
+		// lifecycle policies if cleanup fails.
+		_ = s.deleteFn(context.WithoutCancel(s.ctx), h)
```
krishnamd-jkp, line 643: Are we not capturing delete errors atleast if not failing the operations?

storage/pcu.go
cpriti-os: We cannot handle this anymore, go/go-pcu-dd for design change. This will keep the behavior consistent even if we swap the implementation

storage/pcu.go
cpriti-os: Adreed, should be nil by default. Removed

storage/pcu.go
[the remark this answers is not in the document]
cpriti-os: Not required as per documentation https://godoc.corp.google.com/pkg/crypto/rand?content_ref=note+that+no+error+handling+is+necessary+as+read+always+succeeds#Read 
Note: This function was just moved here from the test and not added.

storage/pcu.go
cpriti-os: Done.

cpriti-os: Done

storage/pcu.go
cpriti-os: Keeping the partNum increment inside the  lock ensures the uploadTask state is captured atomically. We don't want to increment or create a new task if there is an error.

storage/pcu.go
cpriti-os: Even if we do, its not valid since err is not nil?

storage/pcu.go
cpriti-os: No, we can't provide much value with the captured errors. The idea would be to set up lifecycle policies and cleanup would be best effort. We don't want to change the user experience if we swap the implementation of PCU.

storage/pcu.go
cpriti-os: Done

storage/pcu.go
cpriti-os: In this case, nil checks will unnecessarily bloat the code. The pointers like s.w.o with Object and Bucket cannot be unset on the writer (writer is created using an object handle) associated with pcuState and other PCU configurations would be set by default and cannot be nil. (In the next PR, I would also remove some configurable options for PCU and therefore would always be set by default). Still added a check in initPCU() for added safety.

storage/pcu.go
cpriti-os: Removed lock from go routine and did batch updates to reduce contention for lock

storage/pcu.go
krishnamd-jkp: Sorry I meant when both are nil

storage/pcu.go
krishnamd-jkp: But uploadTask also can be created without a mutex I believe, because writes are sequential anyways?

storage/pcu.go
cpriti-os: Should be impossible to get both nil but agreed, don't see any harm in handling it. Added the branch.

storage/pcu.go
cpriti-os: Moved task creating outside the lock. I still think a better practice is to do increments of such variables like partNum inside the mutex as increment should ideally only happen when there are no errors. This is safer and more reliable and incrementing a value should take negligible CPU time given we are already acquiring the lock for checking errors.

krishnamd-jkp: Sure okay. I was also seeing this part logically as it may confuse readers if creating task and incrementing partNum is not thread safe (which I think it is). Resolving it.

krishnamd-jkp (approved):

PR #13642, opened 2026-01-17
feat(pubsub/v2): introduce per stream flow control

hongalex:
This PR introduces the ability to switch the client side flow controller to be applied per stream.

Consider the following environment where NumStreams = 10, MaxMessages=100. 
The server will deliver 100 messages per stream, for a maximum of 1000 messages on the client.
The client's flow controller only understands that MaxMessages = 100, and will only process 100 messages at once.
The other 900 messages are pulled but stuck in the client before being delivered to users, possibly resulting in messages
timing out.

We recommend most users to enable this where they wish to open multiple streams (ReceiveSettings.NumGoroutines)

### DIFF
--- a/pubsub/v2/subscriber_test.go
+++ b/pubsub/v2/subscriber_test.go
@@ -403,7 +403,9 @@ func TestPerStreamFlowControl(t *testing.T) {
 	// This test verifies that flow control can be applied per-stream (per-goroutine)
 	// instead of per-subscriber.
 
-	// Use a fake server.
+	// Use a fake server. The fake doesn't currently implement server-side flow
+	// control, so it will deliver messages as soon as client requests them.
+	// This is perfect for testing both forms of flow control.
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	client, srv := newFake(t)


pubsub/v2/subscriber.go
```diff
@@ -294,6 +308,14 @@ func (s *Subscriber) Receive(ctx context.Context, f func(context.Context, *Messa
 		iter := newMessageIterator(s.c.SubscriptionAdminClient, s.name, po)
 		iter.enableTracing = s.enableTracing
 
+		if s.ReceiveSettings.EnablePerStreamFlowControl {
```
alvarowolfx, line 318: why this needs to be checked again here ? it looks like is in the same function context. Something can change between L273 and L311 that this value is changed ?

pubsub/v2/subscriber.go
```diff
@@ -258,11 +266,17 @@ func (s *Subscriber) Receive(ctx context.Context, f func(context.Context, *Messa
...
-		MaxOutstandingMessages: maxCount,
-		MaxOutstandingBytes:    maxBytes,
-		LimitExceededBehavior:  FlowControlBlock,
-	})
+
+	var fc flowController
+	// When using the legacy per-client flow control, limits
+	// are applied per-client rather than per stream/iterator.
+	if !s.ReceiveSettings.EnablePerStreamFlowControl {
+		fc = newSubscriberFlowController(FlowControlSettings{
```
alvarowolfx: is this Subscribe Flow Controller already released for customers ? Because in the `EnablePerStreamFlowControl` you mention that the default value is going to be `false` to keep backward compatibility, but looking at the code, it seems like it was actually enabled by default.

alvarowolfx: oh I see, this is a NOT `EnablePerStreamFlowControl`.

pubsub/v2/subscriber.go
alvarowolfx: I noticed later that they are checking against the same `s.ReceiveSettings.EnablePerStreamFlowControl` but one checks if is NOT enabled and them enabled it, this one checks if is enabled and them enabled it too. I'm confused.

shollyman (commented):
I don't think I realized pubsub was flow control at the client, I had assumed it was per stream flow control.

Are there customers who may want both a shared client flow controller and per stream flow controller limits simultaneously?  I'm mostly just poking the abstraction to see if this should be the boolean toggle as articulated or something that supports additional flow control modes.

Should there be a test as well?  Something that asserts that connections have different fc instances or something?

pubsub/v2/subscriber.go
hongalex: Yeah so the declaration needs to happen at different places. Outside of the for loop, there's a single flow controller that limits the total resources for each iterator/stream/for-loop. This is the current default per stream flow controller.

Looking at this more, I realized that I had the correct implementation before but made changes based on a Gemini review that makes the shadowing logic incorrect now. Will be reverting that commit.

pubsub/v2/subscriber.go
```diff
@@ -308,14 +308,17 @@ func (s *Subscriber) Receive(ctx context.Context, f func(context.Context, *Messa
...
 
+		var fc flowController
 		if s.ReceiveSettings.EnablePerStreamFlowControl {
 			fc = newSubscriberFlowController(FlowControlSettings{
 				MaxOutstandingMessages: maxCount,
 				MaxOutstandingBytes:    maxBytes,
 				LimitExceededBehavior:  FlowControlBlock,
 			})
+		} else {
+			fc = clientFlowController
```
alvarowolfx, line 325: code is much clearer now. I think the problem to me know is that this `fc = clientFlowController` is creating a copy of the flowController, so it has per stream flow control anyway here, just creating a copy of the client level flow controller. 

I think we need to make `fc` a pointer and if we want to re use the same controller from the client level for all streams.

pubsub/v2/subscriber.go
alvarowolfx: code is much clear now to me, but now I raised another point on the controller being shared or not on the client level

hongalex:
@shollyman a test has been added (courtesy of gemini) to verify both modes.

It's a bit more complicated than asserting different `flowController` units, but validates the behavior through callback checking.

shollyman (approved):
Thanks for the improvements here, and the testing to validate it!

PR #13647, opened 2026-01-19
fix: clean up bigtable references to context.Background()

tbpg:


bigtable/otel_metrics.go
```diff
@@ -185,7 +185,7 @@ func newOtelMetricsContext(ctx context.Context, cfg metricsConfig) (*otelMetrics
 	}
 	if !cfg.disableExporter {
 		meterOpts = append(meterOpts, metric.WithReader(
-			metric.NewPeriodicReader(&exporterLogSuppressor{Exporter: exporter}, metric.WithInterval(interval))))
+			metric.NewPeriodicReaderWithContext(ctx, &exporterLogSuppressor{Exporter: exporter}, metric.WithInterval(interval))))
```
hongalex: I don't see this method in the docs, only the original [NewPeriodicReader](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#NewPeriodicReader)

tbpg: Strange -- I removed this, but something got weird with how the force push was displayed. I pushed a new commit and this file went away. Please take another look.

PR #13652, opened 2026-01-21
fix(storage): deadlock in event loop while coordinating channels

cpriti-os:


storage/grpc_reader_multi_range.go
```diff
@@ -374,19 +394,42 @@ func (m *multiRangeDownloaderManager) eventLoop() {
 	}
 
 	for {
+		var nextReq *storagepb.BidiReadObjectRequest
+		var targetChan chan<- *storagepb.BidiReadObjectRequest
+
+		// Only try to send if we have queued requests
+		if m.unsentRequests.Len() > 0 && m.currentSession != nil {
+			nextReq = m.unsentRequests.Front().Value.(*storagepb.BidiReadObjectRequest)
```
krishnamd-jkp: Good to have some checks here

storage/grpc_reader_multi_range.go
```diff
@@ -374,19 +394,42 @@ func (m *multiRangeDownloaderManager) eventLoop() {
...
+			targetChan = m.currentSession.reqC
+		}
 		select {
 		case <-m.ctx.Done():
 			return
+		// This path only triggers if space is available in the channel.
+		// It never blocks the eventLoop.
+		case targetChan <- nextReq:
+			m.unsentRequests.Remove(m.unsentRequests.Front())
```
krishnamd-jkp: Check for nil "m.unsentRequests.Front()" before removing. the element cannot be nil

storage/grpc_reader_multi_range.go
```diff
@@ -655,7 +676,8 @@ func (m *multiRangeDownloaderManager) ensureSession(ctx context.Context) error {
 			}
 		}
 		if len(rangesToResend) > 0 {
-			m.currentSession.SendRequest(&storagepb.BidiReadObjectRequest{ReadRanges: rangesToResend})
+			retryReq := &storagepb.BidiReadObjectRequest{ReadRanges: rangesToResend}
+			m.unsentRequests.PushFront(retryReq)
```
krishnamd-jkp: We can similarly push wait and close commands at the front to prioritize them.

storage/grpc_reader_multi_range.go
```diff
@@ -655,7 +689,8 @@ func (m *multiRangeDownloaderManager) ensureSession(ctx context.Context) error {
...
```
krishnamd-jkp, line 685: We can send this directly to the stream as stream is just being created?

cpriti-os: Yes we can potentially do that and mostly wont face any issue in practice, this is more consistent with the overall deadlock fix. This adheres to the decoupling that we want to achieve while preserving the priority of the request by pushing it in front.

storage/grpc_reader_multi_range.go
```diff
@@ -374,10 +381,37 @@ func (m *multiRangeDownloaderManager) eventLoop() {
...
+		var nextReq *storagepb.BidiReadObjectRequest
+		var targetChan chan<- *storagepb.BidiReadObjectRequest
+
+		// Only try to send if we have queued requests
+		if m.unsentRequests.Len() > 0 && m.currentSession != nil {
+			nextElem = m.unsentRequests.Front()
+			if nextElem != nil {
+				var ok bool
+				// Safely assert the type and check for nil value.
+				nextReq, ok = nextElem.Value.(*storagepb.BidiReadObjectRequest)
```
tritone: Having to do a cast here seems weird. Maybe use a wrapper around the container/list type to force that elements are of this specific type?

storage/grpc_reader_multi_range.go
tritone: +1

storage/grpc_reader_multi_range.go
```diff
@@ -374,10 +381,37 @@ func (m *multiRangeDownloaderManager) eventLoop() {
...
 		case <-m.ctx.Done():
 			return
-		case cmd := <-m.cmds:
+		// This path only triggers if space is available in the channel.
+		// It never blocks the eventLoop.
+		case targetChan <- nextReq:
+			if nextElem != nil {
+				m.unsentRequests.Remove(nextElem)
+			}
+		case cmd := <-cmdsChan:
```
krishnamd-jkp, line 406: Maybe we can do this later but if close is sent without a wait, it might potentially close the MRD before all the add commands in the list are processed. We might need to change the handling of wait and close command a bit. But this can come later I think.

storage/grpc_reader_multi_range.go
```diff
@@ -935,3 +927,28 @@ func readerAttrsFromObject(o *ObjectAttrs) ReaderObjectAttrs {
...
+
+func newRequestQueue() *requestQueue {
+	return &requestQueue{l: list.New()}
+}
+
+func (q *requestQueue) PushBack(r *storagepb.BidiReadObjectRequest)  { q.l.PushBack(r) }
+func (q *requestQueue) PushFront(r *storagepb.BidiReadObjectRequest) { q.l.PushFront(r) }
+func (q *requestQueue) Len() int                                     { return q.l.Len() }
+
+func (q *requestQueue) Front() *storagepb.BidiReadObjectRequest {
```
krishnamd-jkp, line 943: nit: add unit tests. can go in later

krishnamd-jkp (approved):

storage/grpc_reader_multi_range.go
cpriti-os: will take it up in a future PR

storage/grpc_reader_multi_range.go
cpriti-os: Done.

storage/grpc_reader_multi_range.go
cpriti-os: Agreed, close can now be more "abrupt" , but that will only be in a case where user does not follow the best practice of calling wait before close. This looks sound logic for now, can rethink later if needed.

PR #13720, opened 2026-01-23
fix(bigquery): revert timestamp_output_format to use_int64_timestamp

nishantsharma:
This breaks ALL BigQuery operations including schema creation, queries, and result fetching.

## Solution

Revert to the older `use_int64_timestamp` option which:
- Has identical functionality (timestamps returned as int64)
- Is fully supported by BigQuery API
- Was working before the upstream change

## Changes

| File | Change |
|------|--------|
| `bigquery/iterator.go` | Remove `defaultTimestampWireFormat` const, revert 2 API calls |
| `bigquery/job.go` | Revert 1 API call |
| `bigquery/query.go` | Revert struct field |
| `bigquery/query_test.go` | Update test expectations |

## Testing

- This restores the exact behavior from before upstream commit `59efe32388`
- CI should pass with this change (was failing due to the unsupported API option)

## Related

- Upstream commit: googleapis/google-cloud-go@59efe32388
- Upstream issue: internal b/447623020 (Google's tracking for transitioning to new format)

### DIFF
--- a/bigquery/iterator.go
+++ b/bigquery/iterator.go
@@ -25,10 +25,6 @@ import (
 	"google.golang.org/api/iterator"
 )
 
-const (
-	defaultTimestampWireFormat string = "INT64"
-)
-
 // Construct a RowIterator.
 func newRowIterator(ctx context.Context, src *rowSource, pf pageFetcher) *RowIterator {
 	it := &RowIterator{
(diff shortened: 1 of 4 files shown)


nishantsharma:
Was only intended to be used in my fork only. Got opened upstream by mistake.

PR #13721, opened 2026-01-23
test(spanner): run integration test for uuid against prod DB

gagangupt16:


gagangupt16:
Gemini summary analysis is incorrect. The PR is not removing coverage. It is removing duplicate tests which were present as we were not testing UUID against production database.

PR #13740, opened 2026-01-28
doc(bigtable): Correct ReadRow size

bhshkh:
fixes: b/333570192


shollyman (commented):
Presumably this changed at some point?  Is this documented in the cloudsite somewhere that we can link to if we think it might be updated in the future?  Or its so rare that it doesn't matter.

bhshkh:
> Presumably this changed at some point? Is this documented in the cloudsite somewhere that we can link to if we think it might be updated in the future? Or its so rare that it doesn't matter.

Thanks for the review!
Confirmed the link that can be used with the CBT product team in b/333570192#comment4 and updated the PR.

shollyman (approved):

PR #13743, opened 2026-01-29
chore(storage): add force_direct_connectivity gRPC header

cpriti-os:


storage/experimental/experimental.go
```diff
@@ -107,3 +107,10 @@ func WithGRPCBidiReads() option.ClientOption {
 func WithZonalBucketAPIs() option.ClientOption {
 	return internal.WithZonalBucketAPIs.(func() option.ClientOption)()
 }
+
+// WithDirectConnectivityEnforced is a function which is implemented by the storage package.
```
krishnamd-jkp: nit: Can you follow the same comment pattern other options are using? "option-name provides an [option.ClientOption] ..."

storage/grpc_client.go
```diff
@@ -160,16 +166,62 @@ func newGRPCStorageClient(ctx context.Context, opts ...storageOption) (*grpcStor
...
+		if !ok {
+			md = metadata.MD{}
+		}
+
+		// Merge with existing x-goog-request-params if present.
+		if vals := md.Get(requestParamsHeaderKey); len(vals) > 0 {
+			md.Set(requestParamsHeaderKey, vals[0]+"&"+param)
+		} else {
+			md.Set(requestParamsHeaderKey, param)
+		}
```
krishnamd-jkp, line 194: Can define a local function which can set this as it is repeated twice?

krishnamd-jkp:
Also, when fallback option is used and server returns error, changing it to opted out will come in phase 2?

storage/grpc_client.go
cpriti-os: done

storage/experimental/experimental.go
cpriti-os: Done

PR #13747, opened 2026-01-31
chore: replace old go teams with cloud-sdk-go-team

suztomo:
chore: replace old go teams with cloud-sdk-go-team

b/479547457


suztomo:
In CODEOWNERS file, I see other teams that I may rename later. I'll revisit this pull request.

PR #13749, opened 2026-02-02
feat(retry): adding support for max-retry-duration for all api other than resumable-upload

raj-prince:
**Description:**
- Added support for [WithMaxRetryDuration](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) retry option to allow users to configure a maximum duration for retry attempts, similar to the existing [WithMaxAttempts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) option.
- Option was used internally for write-flow which has default 32 second chunk level max-retry deadline. Configured max-retry-timeout will be used if user set this value.

**Testing:**
- Validated unit test - `go test -short`
- Validated emulator based test - `bash emulator_test.sh`

### DIFF
--- a/storage/client_test.go
+++ b/storage/client_test.go
@@ -2727,7 +2727,7 @@ func TestRetryMaxDurationEmulated(t *testing.T) {
 			t.Fatalf("creating bucket: %v", err)
 		}
 		// Set up many errors to ensure we don't run out during the duration
-		instructions := map[string][]string{"storage.buckets.get": {"return-503", "return-503", "return-503", "return-503", "return-503", "return-503", "return-503", "return-503", "return-503", "return-503"}}
+		instructions := map[string][]string{"storage.buckets.get": slices.Repeat([]string{"return-503"}, 10)}
 		testID := createRetryTest(t, client, instructions)
 		ctx = callctx.SetHeaders(ctx, "x-retry-test-id", testID)
 		// Set a short max retry duration with a very small backoff to ensure we hit the duration limit


storage/client_test.go
```diff
@@ -2719,6 +2719,39 @@ func TestRetryMaxAttemptsEmulated(t *testing.T) {
...
+				t.Errorf("GetBucket: got unexpected error %v; want 503", err)
+			}
+		}
+		// Error should be wrapped so it indicates that retry deadline has been reached.
+		if got, want := err.Error(), "retry deadline of"; !strings.Contains(got, want) {
+			t.Errorf("got error: %q, want to contain: %q", got, want)
+		}
+		if got, want := err.Error(), "200ms"; !strings.Contains(got, want) {
+			t.Errorf("got error: %q, want to contain duration: %q", got, want)
+		}
```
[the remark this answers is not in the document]
raj-prince: Don't want to add compound condition in test.

storage/storage.go
```diff
@@ -2497,6 +2497,28 @@ func (wb *withMaxAttempts) apply(config *retryConfig) {
 	config.maxAttempts = &wb.maxAttempts
 }
 
+// WithMaxRetryDuration configures the maximum duration for which requests can be retried.
+// Once this deadline is reached, no further retry attempts will be made, and the last
+// error will be returned.
+// For example, if you set WithMaxRetryDuration(10*time.Second), retries will stop after
+// 10 seconds even if the maximum number of attempts hasn't been reached.
+// Without this setting, operations will continue retrying until either the context is
+// canceled, a deadline is reached, or the maximum number of attempts is exhausted.
```
cpriti-os: This means context deadline right? Sentence can be reframed to make it clearer.

storage/invoke_test.go
```diff
@@ -270,6 +270,25 @@ func TestInvoke(t *testing.T) {
...
+			expectFinalErr:    false,
+			wantAttempts:      intPointer(3),
+		},
+		{
+			desc:              "maxRetryDuration reached before maxAttempts",
+			count:             20,
+			initialErr:        &googleapi.Error{Code: 500},
+			finalErr:          nil,
+			isIdempotentValue: true,
+			retry:             &retryConfig{policy: RetryAlways, maxAttempts: intPointer(100), maxRetryDuration: time.Millisecond * 100},
```
cpriti-os: To avoid any kind of flakiness, wouldn't it be better to use mock clock where we can reliably trigger max duration

storage/storage.go
```diff
@@ -2497,6 +2497,28 @@ func (wb *withMaxAttempts) apply(config *retryConfig) {
...
+// A value of 0 allows infinite retries (subject to other constraints).
```
cpriti-os, line 2508: Adding a test to ensure retries are infact infinite (or > 1) when set to 0 for completeness.

storage/storage.go
```diff
@@ -2572,7 +2594,6 @@ type retryConfig struct {
 	maxAttempts *int
 	// maxRetryDuration, if set, specifies a deadline after which the request
 	// will no longer be retried. A value of 0 allows infinite retries.
-	// maxRetryDuration is currently only set by Writer.ChunkRetryDeadline.
```
cpriti-os, line 2575: Do we need to modify logic and test for ensuring the code behaves as expected when both are set: maxRetryDuration and chunkRetryDeadline

storage/storage.go
raj-prince: Yes cancel in case of ContextWithCancel and deadline for ContextWithTimeout, changed a bit.

storage/storage.go
raj-prince: I don't see, retryConfig.maxRetryDuration is being used in any write flow.

Added as part of https://github.com/googleapis/google-cloud-go/pull/11476, and usage got deleted from this PR: https://github.com/googleapis/google-cloud-go/pull/12422.

Could you please double check?

storage/invoke_test.go
raj-prince: Yeah, make sense. 

But that would required more changes, can we take that as a separate PR, given existing similar test is implemented like that. Also, I have changed the variable to ensure no flakiness.

cpriti-os: Sounds good

storage/storage.go
cpriti-os: I think the updated PR now manages this risk.

storage/client_test.go
```diff
@@ -2719,6 +2719,39 @@ func TestRetryMaxAttemptsEmulated(t *testing.T) {
...
 
+// Test that errors are wrapped correctly if retry happens until max retry duration is reached.
+func TestRetryMaxDurationEmulated(t *testing.T) {
+	transportClientTest(context.Background(), t, func(t *testing.T, ctx context.Context, project, bucket string, client storageClient) {
+		_, err := client.CreateBucket(ctx, project, bucket, &BucketAttrs{}, nil)
+		if err != nil {
+			t.Fatalf("creating bucket: %v", err)
+		}
+		// Set up many errors to ensure we don't run out during the duration
+		instructions := map[string][]string{"storage.buckets.get": {"return-503", "return-503", "return-503", "return-503", "return-503", "return-503", "return-503", "return-503", "return-503", "return-503"}}
```
krishnamd-jkp: nit: Can we keep passing the errors consistent with the other test? Using a for loop.

PR #13750, opened 2026-02-02
feat(storage): adding more context to the custom should-retry method

raj-prince:
**Description:**
- Adding new interface for ShouldRetry method to get more context, like, attempt#, id, operation_name, object_name, bucket_name.
- Marking the previous ErrorFun as deprecated to migrate on the new interface.
- GCSFuse will use it for logging purpose - which will be helpful in connecting multiple retries for same overall request.

**Test:**
- Unit test - passed.

### DIFF
--- a/storage/example_test.go
+++ b/storage/example_test.go
@@ -1114,9 +1114,9 @@ func ExampleWithErrorFunc() {
 	}
 	defer client.Close()
 
-	// Basic retry logic using the legacy signature
+	// Basic retry logic using the legacy signature.
 	customRetry := func(err error) bool {
-		// Only retry on specific errors
+		// Only retry on specific errors.
 		return storage.ShouldRetry(err)
 	}
 
@@ -1126,7 +1126,7 @@ func ExampleWithErrorFunc() {
 		storage.WithPolicy(storage.RetryAlways),
 	)
 
-	// Operations on this object will use the custom retry logic
+	// Operations on this object will use the custom retry logic.
 	_, err = obj.Attrs(ctx)
 	if err != nil {
 		// TODO: handle error.
@@ -1141,19 +1141,19 @@ func ExampleWithErrorFuncWithContext() {
 	}
 	defer client.Close()
 
-	// Advanced retry logic with full retry context
+	// Advanced retry logic with full retry context.
 	// This provides comprehensive context including operation, bucket, and object information.
 	customRetryWithContext := func(err error, retryCtx *storage.RetryContext) bool {
 		// Log the retry attempt and operation context
 		log.Printf("Retry decision for %s (bucket=%s, object=%s), invocation=%s, attempt=%d: %v",
 			retryCtx.Operation, retryCtx.Bucket, retryCtx.Object, retryCtx.InvocationID, retryCtx.Attempt, err)
 
-		// Don't retry after 3 attempts
+		// Don't retry after 3 attempts.
 		if retryCtx.Attempt > 3 {
 			return false
 		}
 
-		// Use the default ShouldRetry logic for other cases
+		// Use the default ShouldRetry logic for other cases.
 		return storage.ShouldRetry(err)
 	}
 
@@ -1163,7 +1163,7 @@ func ExampleWithErrorFuncWithContext() {
 		storage.WithPolicy(storage.RetryAlways),
 	)
 
-	// Operations on this object will use the custom retry logic
+	// Operations on this object will use the custom retry logic.
 	_, err = obj.Attrs(ctx)
 	if err != nil {
 		// TODO: handle error.
--- a/storage/invoke.go
+++ b/storage/invoke.go
@@ -59,6 +59,7 @@ func (r *retryConfig) runShouldRetry(err error, retryCtx *RetryContext) bool {
 	if r == nil || r.shouldRetry == nil {
 		return ShouldRetry(err)
 	}
+
 	return r.shouldRetry(err, retryCtx)
 }
 
@@ -90,6 +91,9 @@ func withObject(object string) runOption {
 // run determines whether a retry is necessary based on the config and
 // idempotency information. It then calls the function with or without retries
 // as appropriate, using the configured settings.
+// TODO: consider replacing the functional option (runOption) pattern with a
+// hardcoded struct based approach if parameter related changes requires for all
+// the callers. Ref: http://shortn/_ciY2iWLh2J
 func run(ctx context.Context, call func(ctx context.Context) error, retry *retryConfig, isIdempotent bool, opts ...runOption) error {
 	options := &runOptions{}
 	for _, opt := range opts {
--- a/storage/invoke_test.go
+++ b/storage/invoke_test.go
@@ -537,7 +537,7 @@ func TestInvokeWithRetryContext(t *testing.T) {
 			var lastInvocationID string
 			callCounter := 0
 
-			// Custom shouldRetry function that captures RetryContext
+			// Custom shouldRetry function that captures RetryContext.
 			customRetry := &retryConfig{
 				policy: RetryAlways,
 				shouldRetry: func(err error, retryCtx *RetryContext) bool {
@@ -546,7 +546,7 @@ func TestInvokeWithRetryContext(t *testing.T) {
 						return false
 					}
 
-					// Capture the context for verification
+					// Capture the context for verification.
 					capturedContexts = append(capturedContexts, &RetryContext{
 						Attempt:      retryCtx.Attempt,
 						InvocationID: retryCtx.InvocationID,
@@ -555,13 +555,13 @@ func TestInvokeWithRetryContext(t *testing.T) {
 						Object:       retryCtx.Object,
 					})
 
-					// Remember the invocation ID from the first call
+					// Remember the invocation ID from the first call.
 					if len(capturedContexts) == 1 {
 						lastInvocationID = retryCtx.InvocationID
 					}
 
-					// Retry if we haven't reached the retry limit
-					// retryCtx.Attempt is the current attempt that just failed
+					// Retry if we haven't reached the retry limit.
+					// retryCtx.Attempt is the current attempt that just failed.
 					return retryCtx.Attempt <= test.retryAttempts
 				},
 				backoff: &gax.Backoff{Initial: time.Millisecond},
@@ -569,14 +569,14 @@ func TestInvokeWithRetryContext(t *testing.T) {
 
 			call := func(ctx context.Context) error {
 				callCounter++
-				// Fail for the first retryAttempts calls, succeed on call retryAttempts+1
+				// Fail for the first retryAttempts calls, succeed on call retryAttempts+1.
 				if callCounter <= test.retryAttempts {
 					return test.errorToReturn
 				}
 				return nil
 			}
 
-			// Run with the operation metadata
+			// Run with the operation metadata.
 			err := run(ctx, call, customRetry, true,
 				withOperation(test.operation),
 				withBucket(test.bucket),
@@ -586,8 +586,8 @@ func TestInvokeWithRetryContext(t *testing.T) {
 				s.Fatalf("expected nil error after retries, got: %v", err)
 			}
 
-			// Verify we got the expected number of retry contexts
-			// shouldRetry is called once per failed attempt
+			// Verify we got the expected number of retry contexts.
+			// shouldRetry is called once per failed attempt.
 			expectedCalls := test.retryAttempts
 			if len(capturedContexts) != expectedCalls {
 				s.Errorf("expected %d retry contexts, got %d", expectedCalls, len(capturedContexts))
@@ -597,12 +597,12 @@ func TestInvokeWithRetryContext(t *testing.T) {
 			for i, retryCtx := range capturedContexts {
 				expectedAttempt := i + 1
 
-				// Check attempt number
+				// Check attempt number.
 				if retryCtx.Attempt != expectedAttempt {
 					s.Errorf("attempt %d: expected Attempt=%d, got %d", i, expectedAttempt, retryCtx.Attempt)
 				}
 
-				// Check invocation ID is consistent
+				// Check invocation ID is consistent.
 				if retryCtx.InvocationID == "" {
 					s.Errorf("attempt %d: InvocationID should not be empty", i)
 				}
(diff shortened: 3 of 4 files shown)


storage/storage.go
```diff
@@ -2551,28 +2551,51 @@ func (ws *withPolicy) apply(config *retryConfig) {
...
+// The shouldRetry function can have one of two signatures for backward compatibility:
+//   - func(err error) bool  (legacy signature)
+//   - func(err error, currentAttempt int, invocationID string) bool  (new signature with retry context)
+//
+// The new signature provides additional parameters:
+//   - currentAttempt: the current attempt number (1-based, so first call is attempt 1)
+//   - invocationID: a unique identifier for the current operation invocation
+//
+// The function signature is automatically detected at runtime.
+func WithErrorFunc(shouldRetry interface{}) RetryOption {
```
tritone: Using an empty interface signature for something like this really seems like an antipattern because it means that using an incorrect type will trigger a runtime error rather than compile time.

Can you just add a additional RetryOption that takes the new function type as its argument?

storage/client_test.go
```diff
@@ -2719,6 +2719,39 @@ func TestRetryMaxAttemptsEmulated(t *testing.T) {
 	})
 }
 
+// Test that errors are wrapped correctly if retry happens until max retry duration is reached.
+func TestRetryMaxDurationEmulated(t *testing.T) {
```
cpriti-os, line 2723: This test is a part of different PR.

storage/grpc_client.go
```diff
@@ -1218,7 +1218,7 @@ func (c *grpcStorageClient) NewRangeReader(ctx context.Context, params *newRange
 			}
 			err = decoder.readFullObjectResponse()
 			return err
-		}, s.retry, s.idempotent)
+		}, s.retry, s.idempotent, "", "", "")
```
cpriti-os: Why are these empty strings?

storage/storage.go
```diff
@@ -2535,8 +2535,27 @@ func (ws *withPolicy) apply(config *retryConfig) {
 	config.policy = ws.policy
 }
 
+// RetryContext provides comprehensive context about a retry attempt.
+// It is passed to custom retry functions configured via WithErrorFuncWithContext.
+type RetryContext struct {
+	// Attempt is the current attempt number (1-based, so first call is attempt 1).
+	Attempt int
+	// InvocationID is a unique identifier for the current operation invocation.
+	InvocationID string
```
meet2mky, line 2573: Is `InvocationID` same for subsequent request if it's getting retried? We ideally want to correlate the current retried request to previous ones to understand the particular operation instance which is getting retried continuously.

raj-prince: Yes. That's the purpose of keeping invocation ID. I'll update attribute doc.

storage/grpc_client.go
raj-prince: Used function-optional parameters to reduce the impact of this change.

storage/storage.go
```diff
@@ -2562,6 +2562,23 @@ func (ws *withPolicy) apply(config *retryConfig) {
 	config.policy = ws.policy
 }
 
+// RetryContext provides comprehensive context about a retry attempt.
+// It is passed to custom retry functions configured via WithErrorFuncWithContext.
+type RetryContext struct {
+	// Attempt is the current attempt number (1-based, so first call is attempt 1).
+	Attempt int
+	// InvocationID is a unique identifier for the current operation invocation.
+	// This will be same for all attempts of the same operation, used to correlated
```
krishnamd-jkp: nit: used to correlate

storage/storage.go
```diff
@@ -2589,16 +2606,59 @@ type withErrorFunc struct {
 }
 
 func (wef *withErrorFunc) apply(config *retryConfig) {
+	// Wrap legacy signature to new signature
+	config.shouldRetry = func(err error, ctx *RetryContext) bool {
```
krishnamd-jkp: Can we change the variable name here to not confuse it with regular "ctx context.Context"? Other places as well

storage/storage.go
```diff
@@ -2589,16 +2606,59 @@ type withErrorFunc struct {
...
+//
+// - Transient network errors such as connection reset and io.ErrUnexpectedEOF.
+//
+// - Errors which are considered transient using the Temporary() interface.
+//
+// - Wrapped versions of these errors.
+//
+// This option can be used to retry on a different set of errors than the
+// default. Users can use the default ShouldRetry function inside their custom
+// function if they only want to make minor modifications to default behavior.
```
krishnamd-jkp, line 2640: Can you also suggest a usage of RetryContext in these comments?

storage/storage_test.go
```diff
@@ -1038,10 +1038,10 @@ func TestObjectRetryer(t *testing.T) {
 			name: "set ErrorFunc only",
 			call: func(o *ObjectHandle) *ObjectHandle {
 				return o.Retryer(
-					WithErrorFunc(func(err error) bool { return false }))
+					WithErrorFuncWithContext(func(err error, ctx *RetryContext) bool { return false }))
```
krishnamd-jkp: This test case is for WithErrorFunc. Can you add another test case for WithErrorFuncWithContext? Here and other places as well

storage/retry_context_test.go
```diff
@@ -0,0 +1,118 @@
...
+// limitations under the License.
+
+package storage
+
+import (
+	"errors"
+	"testing"
+)
+
+func TestRetryContextSignatures(t *testing.T) {
```
krishnamd-jkp, line 22: I think this test is redundant as the signature is tested anyways in the other two tests in the file.

storage/invoke.go
```diff
@@ -53,19 +53,59 @@ var (
 	})
 )
 
-func (r *retryConfig) runShouldRetry(err error) bool {
+// runShouldRetry calls the configured shouldRetry function.
+// The shouldRetry function has already been normalized to the new signature
```
krishnamd-jkp: nit: can we remove "new signature" as it doesn't share any context about the timeline?

storage/invoke.go
```diff
@@ -53,19 +53,59 @@ var (
...
+
+// runOptions holds optional metadata for retry contexts.
+type runOptions struct {
+	operation string
+	bucket    string
+	object    string
+}
+
+// RunOption configures optional metadata for retry contexts.
+type RunOption func(*runOptions)
```
krishnamd-jkp: Should these be exported? Given "run" itself is an unexported function.

storage/invoke_test.go
krishnamd-jkp, line 1: Can we add a test case to test the functionality of RetryContext in run function?

storage/integration_test.go
```diff
@@ -1676,7 +1676,7 @@ func TestIntegration_PublicAccessPrevention(t *testing.T) {
 		}
 
 		ctxWithTimeout, cancelCtx := context.WithTimeout(ctx, time.Second*10)
-		a = o.Retryer(WithErrorFunc(retrier), WithPolicy(RetryAlways)).ACL()
+		a = o.Retryer(WithErrorFuncWithContext(retrier), WithPolicy(RetryAlways)).ACL()
```
krishnamd-jkp: Any reason for this change?

storage/storage.go
```diff
@@ -2562,6 +2562,23 @@ func (ws *withPolicy) apply(config *retryConfig) {
 	config.policy = ws.policy
 }
 
+// RetryContext provides comprehensive context about a retry attempt.
+// It is passed to custom retry functions configured via WithErrorFuncWithContext.
+type RetryContext struct {
+	// Attempt is the current attempt number (1-based, so first call is attempt 1).
+	Attempt int
```
meet2mky, line 2569: NIT: Do we need attempt? It seems irrelevant except that it can immediately tell which # retry it is but that you can also tell based on the InvocationID count in the logs as well.

storage/invoke.go
raj-prince: Not required, making it unexported.

storage/retry_context_test.go
raj-prince: Removing this file all together given added test in the invoke_test.go.

storage/integration_test.go
raj-prince: removing the change.

storage/storage.go
raj-prince: Yeah, I agree we can get the count from InvocationID too. But this optimization will not save much in terms of memory/cpu. As in the worst case there could be O(K) parallel request will be retried in the worst case.

Also, keeping the attempt for completeness - other clients of Go SDK may implement complex retry logic based on attempt and operation or error (although rare).

krishnamd-jkp:
Can you check the failing kokoro test?

storage/storage.go
```diff
@@ -2589,16 +2606,62 @@ type withErrorFunc struct {
...
 
 type retryConfig struct {
 	backoff     *gax.Backoff
 	policy      RetryPolicy
-	shouldRetry func(err error) bool
+	shouldRetry func(error, *RetryContext) bool
 	maxAttempts *int
 	// maxRetryDuration, if set, specifies a deadline after which the request
 	// will no longer be retried. A value of 0 allows infinite retries.
+	// maxRetryDuration can be set via WithMaxRetryDuration or Writer.ChunkRetryDeadline.
```
cpriti-os: This looks unrelated to this PR, please sync this branch and remove any unwanted changes.

storage/storage_test.go
```diff
@@ -1467,6 +1468,214 @@ func TestRetryer(t *testing.T) {
...
+			}
+			defer c.Close()
+
+			c.SetRetry(tc.options...)
+
+			if c.retry == nil || c.retry.shouldRetry == nil {
+				t.Fatal("shouldRetry not set on Client")
+			}
+
+			// Verify the function can be called with RetryContext
```
cpriti-os, line 1665: Nit: Comments end with a period, here and anywhere else.

storage/storage.go
```diff
@@ -2589,16 +2606,62 @@ type withErrorFunc struct {
...
+// - Transient network errors such as connection reset and io.ErrUnexpectedEOF.
+//
+// - Errors which are considered transient using the Temporary() interface.
+//
+// - Wrapped versions of these errors.
+//
+// This option can be used to retry on a different set of errors than the
+// default. Users can use the default ShouldRetry function inside their custom
+// function if they only want to make minor modifications to default behavior.
+// RetryContext can be used to improve observability by logging InvocationID to
```
cpriti-os: should we document that the ctx can be nil? Can this lead to any panic in certain edge cases?

raj-prince: Yeah, not implemented for couple of methods, e.g., in grpc_reader.go.

So, documenting for defensive check, but later can be removed once started providing the context from the every invocation.

storage/example_test.go
```diff
@@ -1105,3 +1105,67 @@ func ExampleMultiRangeDownloader() {
...
+
+func ExampleWithErrorFuncWithContext() {
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		// TODO: handle error.
+	}
+	defer client.Close()
+
+	// Advanced retry logic with full retry context
```
cpriti-os: Nit: Period

storage/example_test.go
```diff
@@ -1105,3 +1105,67 @@ func ExampleMultiRangeDownloader() {
...
+		return storage.ShouldRetry(err)
+	}
+
+	bucket := client.Bucket("my-bucket")
+	obj := bucket.Object("my-object").Retryer(
+		storage.WithErrorFuncWithContext(customRetryWithContext),
+		storage.WithPolicy(storage.RetryAlways),
+	)
+
+	// Operations on this object will use the custom retry logic
```
cpriti-os: Nit: Period

storage/grpc_client.go
```diff
@@ -590,7 +590,7 @@ func (c *grpcStorageClient) DeleteObject(ctx context.Context, bucket, object str
 	}
 	err := run(ctx, func(ctx context.Context) error {
 		return c.raw.DeleteObject(ctx, req, s.gax...)
-	}, s.retry, s.idempotent)
+	}, s.retry, s.idempotent, withOperation("DeleteObject"), withBucket(bucket), withObject(object))
```
cpriti-os, line 593: The functional option pattern is more idiomatic for public APIs but looks like over engineering in this case for an internal run function where this SDK handles all the run calls and is not really optional. go/gotip/episodes/51?content_ref=you+could+export+all+the+fields+which+allows+callers+to+make+a+config+and+set+values+c+config+size+10+good+fit+struct+is+very+stable+and+unlikely+to+change+or+used+in+non+complex+ways#pattern-4

storage/invoke_test.go
```diff
@@ -492,3 +492,197 @@ func TestInvokeWithCookie(t *testing.T) {
...
+			call := func(ctx context.Context) error {
+				callCounter++
+				// Fail for the first retryAttempts calls, succeed on call retryAttempts+1
+				if callCounter <= test.retryAttempts {
+					return test.errorToReturn
+				}
+				return nil
+			}
+
+			// Run with the operation metadata
```
cpriti-os: Nit: Period

storage/invoke.go
```diff
@@ -53,19 +53,58 @@ var (
...
 
-func (r *retryConfig) runShouldRetry(err error) bool {
+// runShouldRetry calls the configured shouldRetry function if it exists,
+// otherwise it falls back to the default ShouldRetry function.
+func (r *retryConfig) runShouldRetry(err error, retryCtx *RetryContext) bool {
 	if r == nil || r.shouldRetry == nil {
 		return ShouldRetry(err)
 	}
-	return r.shouldRetry(err)
+	return r.shouldRetry(err, retryCtx)
```
cpriti-os, line 63: To prevent downstream panics, should we create a default empty retryCtx?

storage/grpc_client.go
raj-prince: Yeah, agree. But I don't find better solution.

Hardcoding the function parameter will lead to a lot of changes (64 places), and at many places we don't have value like, object, bucket etc. E.g., Hmac related operations doesn't have bucket or object, permission related operation doesn't have object.

storage/storage.go
```diff
@@ -2589,13 +2606,60 @@ type withErrorFunc struct {
...
+// correlate retries of the same operation together, or to implement more complex
+// retry logic such as conditional retries based on the operation type or attempt.
+// Note: the retryCtx may be nil for some unimplemented methods so always a good
```
cpriti-os, line 2644: Should we call it experimental and the interface or functionality can change in the future? Also currently, we need a follow up task to cover all APIs

storage/grpc_client.go
cpriti-os: Agreed, it should be fine for now, maybe a TODO would be better to address this when we cover all APIs.

storage/invoke.go
raj-prince: Either way fine to me, but more inclined towards sending nil.

Anyway we have documented in the public doc to add a defensive check for nil retryContext. Keeping nil will be a clear indication that a particular method is not providing retryCtx and retryCtx will be passed in the follow-up PR (grpc_reader, grpc_writer).

PR #13751, opened 2026-02-02
prototype: copy snippets into module's internal directory

jskeet:
(This is just for the sake of discussion.)


bhshkh:
This has already been done in 

https://github.com/googleapis/google-cloud-go/pull/19917

https://github.com/googleapis/google-cloud-go/pull/14705

PR #13755, opened 2026-02-03
feat(bigtable): Use connection factory to create connection and require connections to be primed.

sushanb:
lift and sift of https://github.com/sushanb/google-cloud-go/pull/5

### DIFF
--- a/bigtable/internal/transport/connpool_synctest_test.go
+++ b/bigtable/internal/transport/connpool_synctest_test.go
@@ -26,9 +26,6 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/sdk/metric"
 	"go.opentelemetry.io/otel/sdk/metric/metricdata"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/metadata"
-	"google.golang.org/grpc/status"
 )
 
 func TestRecordClientStartUp(t *testing.T) {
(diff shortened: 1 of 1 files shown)


bigtable/internal/transport/connpool.go
```diff
@@ -391,21 +405,12 @@ func NewBigtableChannelPool(ctx context.Context, connPoolSize int, strategy btop
 			break
 		}
 
-		conn, err := dial()
+		entry, err := pool.factory.newEntry(ctx)
```
nimf, line 408: As discussed, this changes priming from concurrent to sequential. We should fix it before release.

bigtable/internal/transport/connpool_synctest_test.go
```diff
@@ -102,3 +105,76 @@ func TestRecordClientStartUp(t *testing.T) {
...
+	tests := []struct {
+		name           string
+		dialFunc       func() (*BigtableConn, error)
+		primeErrors    []error
+		ctxTimeout     time.Duration
+		wantErr        bool
+		wantPrimeCalls int
+	}{
+		{
+			name:        "Timeout During Backoff",
```
nimf: why this one is in connpool_synctest_test and the others in connpool_test?

bigtable/internal/transport/connpool.go
sushanb: Yes, agreed.

bigtable/internal/transport/connpool_test.go
```diff
@@ -178,12 +178,126 @@ func TestNewBigtableChannelPoolEdgeCases(t *testing.T) {
...
+			wantPrimeCalls: 3,
+		},
+		{
+			name:           "Context Timeout",
+			dialFunc:       goodDialFunc,
+			primeErrors:    []error{someErr},
+			ctxTimeout:     -1 * time.Millisecond, // Already expired
+			wantErr:        true,
+			wantPrimeCalls: 0,
+		},
```
nimf, line 239: There is no "Timeout during backoff" case now?

nimf (approved):

bigtable/internal/transport/connpool_test.go
sushanb: i will add.

PR #13758, opened 2026-02-03
feat(storage): add multistream options to MRD

krishnamd-jkp:


storage/reader.go
```diff
@@ -154,6 +154,73 @@ func (o *ObjectHandle) NewRangeReader(ctx context.Context, offset, length int64)
 	return r, err
 }
 
+// MRDOption is an option for MultiRangeDownloader.
+type MRDOption interface {
```
tritone, line 158: Let's avoid exporting all this until the underlying code is ready.

tritone (commented):
Options pattern looks good to me

PR #13775, opened 2026-02-04
feat(spanner): add ClientContext support

aseering:
Add support for the new ClientContext field to Spanner's Go Client.

This enables support for Secure Parameters, as well as other future features.

### DIFF
--- a/spanner/client_context_test.go
+++ b/spanner/client_context_test.go
@@ -18,6 +18,7 @@ package spanner
 
 import (
 	"context"
+	"reflect"
 	"testing"
 
 	sppb "cloud.google.com/go/spanner/apiv1/spannerpb"
@@ -41,14 +42,8 @@ func makeClientContext(kv map[string]string) *sppb.RequestOptions_ClientContext 
 	return ctx
 }
 
-func TestClientContext_Query(t *testing.T) {
-	t.Parallel()
-	ctx := context.Background()
-	server, client, teardown := setupMockedTestServer(t)
-	defer teardown()
-
-	stmt := Statement{SQL: "SELECT 1"}
-	server.TestSpanner.PutStatementResult(stmt.SQL, &StatementResult{
+func addSelect1Result(server *MockedSpannerInMemTestServer) {
+	_ = server.TestSpanner.PutStatementResult("SELECT 1", &StatementResult{
 		Type: StatementResultResultSet,
 		ResultSet: &sppb.ResultSet{
 			Metadata: &sppb.ResultSetMetadata{
@@ -63,7 +58,16 @@ func TestClientContext_Query(t *testing.T) {
 			},
 		},
 	})
+}
 
+func TestClientContext_Query(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	server, client, teardown := setupMockedTestServer(t)
+	defer teardown()
+
+	stmt := Statement{SQL: "SELECT 1"}
+	addSelect1Result(server)
 	cc := makeClientContext(map[string]string{"test-key": "test-value"})
 
 	// 1. Test propagation via QueryOptions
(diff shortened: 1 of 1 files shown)


spanner/batch.go
```diff
@@ -158,7 +158,7 @@ func (t *BatchReadOnlyTransaction) PartitionReadUsingIndexWithOptions(ctx contex
 		Index:               index,
 		Columns:             columns,
 		KeySet:              kset,
-		RequestOptions:      createRequestOptions(readOptions.Priority, readOptions.RequestTag, ""),
+		RequestOptions:      createRequestOptions(readOptions.Priority, readOptions.RequestTag, "", readOptions.ClientContext),
```
olavloite: This does not appear to be using the default `ClientContext` of the client.

spanner/batch.go
```diff
@@ -225,7 +225,7 @@ func (t *BatchReadOnlyTransaction) partitionQuery(ctx context.Context, statement
 		Params:              params,
 		ParamTypes:          paramTypes,
 		QueryOptions:        qOpts.Options,
-		RequestOptions:      createRequestOptions(qOpts.Priority, qOpts.RequestTag, ""),
+		RequestOptions:      createRequestOptions(qOpts.Priority, qOpts.RequestTag, "", qOpts.ClientContext),
```
olavloite: This does not appear to be using the default ClientContext of the client.

spanner/pdml.go
```diff
@@ -72,7 +72,7 @@ func (c *Client) partitionedUpdate(ctx context.Context, statement Statement, opt
 		Params:         params,
 		ParamTypes:     paramTypes,
 		QueryOptions:   options.Options,
-		RequestOptions: createRequestOptions(options.Priority, options.RequestTag, ""),
+		RequestOptions: createRequestOptions(options.Priority, options.RequestTag, "", options.ClientContext),
```
olavloite: Here also: Should we merge the default `ClientContext` from the client?

spanner/secure_parameters_test.go
```diff
@@ -0,0 +1,595 @@
+/*
```
olavloite: nit: could we rename this file to `client_context_test.go`? Currently, `ClientContext` only contains secure parameters, but might contain more in the future, and I think it would be good to add additional tests for that into this file.

spanner/secure_parameters_test.go
```diff
@@ -0,0 +1,595 @@
...
+		ResultSet: &sppb.ResultSet{
+			Metadata: &sppb.ResultSetMetadata{
+				RowType: &sppb.StructType{
+					Fields: []*sppb.StructType_Field{
+						{Name: "Col1", Type: &sppb.Type{Code: sppb.TypeCode_INT64}},
+					},
+				},
+			},
+			Rows: []*structpb.ListValue{
+				{Values: []*structpb.Value{{Kind: &structpb.Value_NumberValue{NumberValue: 1}}}},
```
olavloite: This is actually wrong. INT64 should be encoded as a StringValue. NumberValue is only used for FLOAT32 and FLOAT64.

The reason that it (by accident) does not cause an error, is that the test does not actually read the data that is returned. If it tries to do that, then it will fail.

spanner/secure_parameters_test.go
```diff
@@ -0,0 +1,595 @@
...
+	}
+	iter.Stop()
+
+	reqs := drainRequestsFromServer(server.TestSpanner)
+	sqlReqs := []*sppb.ExecuteSqlRequest{}
+	for _, req := range reqs {
+		if sqlReq, ok := req.(*sppb.ExecuteSqlRequest); ok {
+			sqlReqs = append(sqlReqs, sqlReq)
+		}
+	}
```
olavloite: You can use `requestsOfType(..)` for this. See https://github.com/googleapis/google-cloud-go/blob/a8056809ab367b2040e7a0522dc369a3e3f15020/spanner/client_test.go#L5481 for an example.

spanner/secure_parameters_test.go
```diff
@@ -0,0 +1,595 @@
...
+	if len(sqlReqs) != 1 {
+		t.Fatalf("expected 1 ExecuteSqlRequest, got %d", len(sqlReqs))
+	}
```
olavloite: nit (here and elsewhere): The preferred way to check for this in Go is like this:

```go
if g, w := len(sqlReqs), 1; g != w {
  t.Fatalf("num ExecuteSqlRequest mismatch\n Got: %d\nWant: %d", g, w)
}
```

spanner/secure_parameters_test.go
```diff
@@ -0,0 +1,595 @@
...
+	gotReq := sqlReqs[0]
+	if !proto.Equal(gotReq.RequestOptions.ClientContext, clientContext) {
+		t.Errorf("mismatch in ClientContext:\ngot:  %v\nwant: %v", gotReq.RequestOptions.ClientContext, clientContext)
+	}
+
+	// 2. Test propagation via ClientConfig default
```
olavloite: I think that it would be cleaner to move this to its own test method, as nothing from this point on shares anything with the previous test (it even uses its own test server).

spanner/secure_parameters_test.go
```diff
@@ -0,0 +1,595 @@
...
+		ResultSet: &sppb.ResultSet{
+			Metadata: &sppb.ResultSetMetadata{
+				RowType: &sppb.StructType{
+					Fields: []*sppb.StructType_Field{
+						{Name: "Col1", Type: &sppb.Type{Code: sppb.TypeCode_INT64}},
+					},
+				},
+			},
+			Rows: []*structpb.ListValue{
+				{Values: []*structpb.Value{{Kind: &structpb.Value_NumberValue{NumberValue: 1}}}},
```
olavloite: Here also: This should be a `StringValue` with value `"1"`

spanner/secure_parameters_test.go
```diff
@@ -0,0 +1,595 @@
...
+	if len(sqlReqs) != 1 {
+		t.Fatalf("expected 1 ExecuteSqlRequest, got %d", len(sqlReqs))
+	}
+	gotReq := sqlReqs[0]
+	if !proto.Equal(gotReq.RequestOptions.ClientContext, expectedContext) {
+		t.Errorf("mismatch in Merged ClientContext:\ngot:  %v\nwant: %v", gotReq.RequestOptions.ClientContext, expectedContext)
+	}
+}
+
+func TestClientContext_PDML(t *testing.T) {
```
olavloite: I think that if you were to run this test with a default `ClientContext` on the client, that it would not be used for PDML.

spanner/secure_parameters_test.go
```diff
@@ -0,0 +1,595 @@
...
+	if len(sqlReqs) != 1 {
+		t.Fatalf("expected 1 ExecuteSqlRequest, got %d", len(sqlReqs))
+	}
+	gotReq := sqlReqs[0]
+	if !proto.Equal(gotReq.RequestOptions.ClientContext, clientContext) {
+		t.Errorf("mismatch in ClientContext:\ngot:  %v\nwant: %v", gotReq.RequestOptions.ClientContext, clientContext)
+	}
+}
+
+func TestClientContext_BatchReadOnlyTransaction(t *testing.T) {
```
olavloite: I think that if you were to run this test with a default `ClientContext` on the client, that it would not be used.

spanner/secure_parameters_test.go
```diff
@@ -0,0 +1,595 @@
...
+)
+
+func TestClientContext_Query(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	server, client, teardown := setupMockedTestServer(t)
+	defer teardown()
+
+	stmt := Statement{SQL: "SELECT 1"}
+	server.TestSpanner.PutStatementResult(stmt.SQL, &StatementResult{
```
olavloite: I think it would be good to extract setting up a result for `SELECT 1` into a separate helper function, as it is repeated quite often for the various test methods in this file.

spanner/secure_parameters_test.go
```diff
@@ -0,0 +1,595 @@
...
+	clientContext := &sppb.RequestOptions_ClientContext{
+		SecureContext: map[string]*structpb.Value{
+			"test-key": {Kind: &structpb.Value_StringValue{StringValue: "test-value"}},
+		},
+	}
+
+	_, err := client.ReadWriteTransaction(ctx, func(ctx context.Context, tx *ReadWriteTransaction) error {
+		_, err := tx.BatchUpdateWithOptions(ctx, []Statement{stmt}, QueryOptions{ClientContext: clientContext})
+		return err
+	})
```
olavloite: (Applies both for this test and for the other tests that use a read/write transaction): What is the intended outcome for a request in a read/write transaction, when:
1. There is a default `ClientContext` on the client.
2. There is a `ClientContext` set for the transaction (in the `TransactionOptions`)
3. There is a `ClientContext` in the `QueryOptions` for the request

Issue #13788, opened 2026-02-05
bigquery: timestamp_output_format causes 400 on query results


kurochan:
## Client

BigQuery

## Environment

Production on GKE
go version: go1.25.7

## Code and Dependencies

package: cloud.google.com/go/bigquery v1.73.0

## Expected behavior

Query results are returned successfully (same as v1.72.0 and earlier).

## Actual behavior

Query fails with:

```
googleapi: Error 400: timestamp_output_format is not supported yet., invalid
```

## Additional context

Started after upgrading to cloud.google.com/go/bigquery v1.73.0.  
The v1.73.0 release includes `transition format options (#13422)` (commit https://github.com/googleapis/google-cloud-go/commit/59efe3238807932eca03b21803f91898def7a6e7 ), which appears to replace `UseInt64Timestamp` with `TimestampOutputFormat=INT64`. The BigQuery REST API seems to reject `timestamp_output_format` currently, causing a regression.

Related issue: [googleapis/google-cloud-go#13786](https://github.com/googleapis/google-cloud-go/issues/13786)

jadn123:
Glad our integration tests caught this before we bumped the dependency.
Out of curiosity: do you have release/CI tests that would be expected to catch this kind of regression before publishing? If so, what’s the gap here?

shollyman:
Apologies, this was due to how the bigquery service has setup allowlisting for this feature.  Our CI tests were considered allowed and we didn't see allowlist rejection.  Cleaning this up now.

shollyman:
Working on getting the updated release out via https://github.com/googleapis/google-cloud-go/pull/13801

shollyman:
Closing this, 1.73.1 is released.  Thanks for bringing it to our attention!

PR #13789, opened 2026-02-05
fix(bigquery): revert to useInt64Timestamp for REST format options

kurochan:
Revert BigQuery REST format options to `useInt64Timestamp` to avoid the
`timestamp_output_format is not supported yet` 400 ERRORs introduced in v1.73.0.

Related: https://github.com/googleapis/google-cloud-go/issues/13788

### DIFF
--- a/bigquery/iterator.go
+++ b/bigquery/iterator.go
@@ -26,7 +26,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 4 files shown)


quartzmo (approved):

shollyman:
Thanks for this!

PR #13792, opened 2026-02-05
feat(storage): add multistream support to MRD

krishnamd-jkp:


storage/grpc_reader_multi_range.go
```diff
@@ -35,10 +35,13 @@ import (
 const (
 	mrdCommandChannelSize  = 1
 	mrdResponseChannelSize = 100
+	mrdSendChannelSize     = 100
 	// This should never be hit in practice, but is a safety valve to prevent
 	// unbounded memory usage if the user is adding ranges faster than they
 	// can be processed.
 	mrdAddInternalQueueMaxSize = 50000
+	defaultTargetPendingBytes  = 1 << 30 // 1 GiB
```
cpriti-os, line 44: Will it be better to tune this based on machine spec, a little intelligent default than a standard one?

storage/reader.go
```diff
@@ -154,6 +154,73 @@ func (o *ObjectHandle) NewRangeReader(ctx context.Context, offset, length int64)
...
+type targetPendingBytes int
+
+func (c targetPendingBytes) apply(params *newMultiRangeDownloaderParams) {
+	params.targetPendingBytes = int(c)
+}
+
+// withTargetPendingBytes returns an mrdOption that sets target pending
+// bytes on the MRD to c. If number of connections in the MRD is less than
+// maximum connections, MRD will trigger creation of a new connection when
+// outstanding bytes on each of the existing stream exceed c.
```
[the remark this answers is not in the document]
cpriti-os: Looks like a good suggestion.

storage/reader.go
```diff
@@ -164,7 +231,7 @@ func (o *ObjectHandle) NewRangeReader(ctx context.Context, offset, length int64)
 
 // NewMultiRangeDownloader creates a multi-range reader for an object.
 // Must be called on a gRPC client created using [NewGRPCClient].
-func (o *ObjectHandle) NewMultiRangeDownloader(ctx context.Context) (mrd *MultiRangeDownloader, err error) {
+func (o *ObjectHandle) NewMultiRangeDownloader(ctx context.Context, opts ...mrdOption) (mrd *MultiRangeDownloader, err error) {
```
cpriti-os: Shouldn't this change come in the next PR when we finally expose the actual mrdOptions and interface? Don't want to change the public interface for a half baked feature, could be misleading for users (I get this is harmless if nothing is sent). Let me know if I am understanding this incorrectly

storage/grpc_reader_multi_range.go
```diff
@@ -421,17 +519,43 @@ func (m *multiRangeDownloaderManager) eventLoop() {
 	}
 }
 
-func (m *multiRangeDownloaderManager) establishInitialSession() error {
+func (m *multiRangeDownloaderManager) addNewStream() {
+	m.streamCreating = true
+	m.sessionIDCounter++
+	id := int(m.sessionIDCounter)
+	go func(id int) {
+		newSession, err := m.createNewSession(id, proto.Clone(m.readSpec).(*storagepb.BidiReadObjectSpec), false)
```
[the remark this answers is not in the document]
cpriti-os: Please check this

storage/reader.go
krishnamd-jkp: yeah makes sense.  we can just have the default values till options are exposed to the user.

storage/grpc_reader_multi_range.go
```diff
@@ -421,17 +519,43 @@ func (m *multiRangeDownloaderManager) eventLoop() {
...
+		}
+		m.addStreams <- &addStreamCmd{
+			id: id,
+			stream: &mrdStream{
+				id:            id,
+				session:       newSession,
+				pendingRanges: make(map[int64]*rangeRequest),
+			},
+		}
+	}(id, proto.Clone(m.readSpec).(*storagepb.BidiReadObjectSpec))
```
cpriti-os: Should we add the updated read handle to this spec (lastReadHandle)?

storage/grpc_reader_multi_range.go
```diff
@@ -515,21 +639,29 @@ func (m *multiRangeDownloaderManager) handleAddCmd(ctx context.Context, cmd *mrd
 	}
 	m.readIDCounter++
 
+	// Attributes should be ready if we are processing Add commands
```
cpriti-os: Nit: Period after comments.

storage/grpc_reader_multi_range.go
```diff
@@ -515,21 +639,29 @@ func (m *multiRangeDownloaderManager) handleAddCmd(ctx context.Context, cmd *mrd
...
+	streamToSend := m.streamPicker.pick(m.streams)
+	if streamToSend < 0 {
+		m.failRange(nil, req, errors.New("storage: no streams available"))
+		return
+	}
+
 	// Convert to positive offset only if attributes are available.
 	if m.attrs != nil && req.offset < 0 {
-		err := m.convertToPositiveOffset(req)
+		err := m.convertToPositiveOffset(nil, req)
```
cpriti-os, line 745: Will this nil not result to totalRangeBytes not being updated and therefore stale information for the stream?

storage/grpc_reader_multi_range.go
```diff
@@ -538,10 +670,32 @@ func (m *multiRangeDownloaderManager) handleAddCmd(ctx context.Context, cmd *mrd
...
+	if m.streamCreating || len(m.streams) >= m.params.maxConnections {
+		return false
+	}
+	if len(m.streams) < m.params.minConnections {
+		return true
+	}
+
+	// Beyond minConnections, we only add if all existing active streams are at capacity.
+	allAtCapacity := true
+	for _, stream := range m.streams {
```
cpriti-os: Can we do something better than iterate to understand if streams are at capacity? What if we keep track of the number of streams at capacity whenever they meet targetPendingRanges or targetPendingBytes and take them out of capacity when the results are processed. if the number of streams at capacity = number of current streams, add one
This looks like too many iterations for each Add command.

storage/grpc_reader_multi_range.go
```diff
@@ -932,17 +1123,22 @@ type requestQueue struct {
 	l *list.List
 }
 
+type mrdQueuedRequest struct {
```
cpriti-os: Nit: Could be named better like pendingRequest? mrd prefix is redundant in this file, we should have avoided it overall but lets not add to it.

storage/grpc_reader_multi_range.go
```diff
@@ -720,23 +898,34 @@ func (m *multiRangeDownloaderManager) handleStreamEnd(result mrdSessionResult) {
...
+				req.completed = true
+				m.runCallback(req.origOffset, req.bytesWritten, err, req.callback)
+			}
 		}
+		stream.pendingRanges = make(map[int64]*rangeRequest)
+		stream.totalRanges = 0
+		stream.totalRangeBytes = 0
 	}
-	m.pendingRanges = make(map[int64]*rangeRequest)
+	m.unsentRequests = newRequestQueue()
```
cpriti-os: Why are we initializing Requests with a Queue?

storage/grpc_reader_multi_range.go
```diff
@@ -682,28 +857,31 @@ func (m *multiRangeDownloaderManager) ensureSession(ctx context.Context) error {
...
+		stream.session = nil
 	}
 	err := result.err
 	var ensureErr error
 
 	if result.redirect != nil {
 		m.readSpec.RoutingToken = result.redirect.RoutingToken
 		m.readSpec.ReadHandle = result.redirect.ReadHandle
-		ensureErr = m.ensureSession(m.ctx)
+		ensureErr = m.ensureSession(m.ctx, stream)
```
cpriti-os: Since now we have multiple streams, re establishing an existing stream should not halt other streams from processing new commands. This synchronous call could become a bottleneck

storage/grpc_reader_multi_range.go
krishnamd-jkp: m.ReadSpec stores the latest read handle (till that point) anytime there is a new readhandle sent by the server

storage/grpc_reader_multi_range.go
krishnamd-jkp: We would not have picked the stream till that point so we would not have increased the totalRangeBytes anyways.

storage/grpc_reader_multi_range.go
krishnamd-jkp: But mrd prefix is needed anyways because this is all part of storage package, right? pendingRequest doesn't indicate that it is relevant to MRD. Although it is inside the MRD file, it can very well be used in a different file where we would lose the context that it an MRD related artifact.

storage/grpc_reader_multi_range.go
krishnamd-jkp: This is to clear any pending requests left in the queue upon failure. just like we are clearing ```stream.pendingRanges``` with ```make(map[int64]*rangeRequest)``` in line 924

cpriti-os: Nit: add  a function like Clear() where we can call list.Init() to make this clearer.

storage/grpc_reader_multi_range.go
krishnamd-jkp: I've checked this on a smaller machine as well. Second stream improved throughput only after 1 GB of data is requested.

storage/grpc_reader_multi_range.go
krishnamd-jkp: Added a global counter

storage/grpc_reader_multi_range.go
krishnamd-jkp: Calling list.Init() directly. The documentation of Init() - "Init initializes or clears list l."

storage/grpc_reader_multi_range.go
krishnamd-jkp: Refactored

storage/retry_conformance_test.go
cpriti-os: Optional: mrd

storage/integration_test.go
```diff
@@ -627,6 +642,107 @@ func TestIntegration_MRDWithReadHandle(t *testing.T) {
...
+		manager.params.targetPendingBytes = 1
+
+		if err != nil {
+			t.Fatalf("NewMultiRangeDownloader: %v", err)
+		}
+
+		addCount := 100
+
+		const (
+			rangeSize  = 1024 // Minimum size of a range request.
```
cpriti-os: Nit: should we keep this low to make the integration test as light weight as possible while doing the needed functional testing. I think same applies to using 10MB object.

storage/integration_test.go
```diff
@@ -627,6 +642,107 @@ func TestIntegration_MRDWithReadHandle(t *testing.T) {
...
+
+			reader.Add(&r.buf, offset, limit, func(o, l int64, err error) {
+				r.err = err
+				r.gotOffset = o
+				r.gotLimit = l
+				wg.Done()
+			})
+			// Sleep to make sure addStreams channel is also selected
+			// and not only addCommands.
+			time.Sleep(20 * time.Millisecond)
```
cpriti-os: Sleeps should be avoided if possible in tests (can lead to flakiness). Can we do something reliable to ensure addStream commands have been sent before adding new ranges, maybe check if unsentRequests is empty or polling the number of streams with timeout ?

storage/grpc_reader_multi_range.go
```diff
@@ -55,6 +58,67 @@ type internalMultiRangeDownloader interface {
 	getSpanCtx() context.Context
 }
 
+// streamPickerStrategy is an interface which each stream picker must implement.
+type streamPickerStrategy interface {
+	pick(streams map[int]*mrdStream) int
+}
+
+// minBytesPicker picks the stream with minimum outstanding bytes.
+type minBytesPicker struct{}
```
cpriti-os, line 68: Is this the only picker we are supporting?

storage/grpc_reader_multi_range.go
```diff
@@ -55,6 +58,67 @@ type internalMultiRangeDownloader interface {
...
+	id              int
+	pendingRanges   map[int64]*rangeRequest
+	session         *bidiReadStreamSession
+	reconnecting    bool
+	atCapacity      bool
+	totalRanges     int
+	totalRangeBytes int64
+	// statsRanges and statsRangeBytes help in tests to track
+	// distribution of ranges on different streams
+	statsRanges     uint
```
cpriti-os, line 104: This is a good addition and makes tests more reliable but also adds operation consuming CPU in production with no real functionality, can we do something so that this stats tracking is disabled in actual production? On the other hand, I do believe these are good stats and we might be interested in it for advanced observability.

storage/grpc_reader_multi_range.go
krishnamd-jkp: Keeping this for now. Will add minRangesPicker later if needed.

storage/grpc_reader_multi_range.go
```diff
@@ -183,23 +280,39 @@ func (c *mrdGetHandleCmd) apply(ctx context.Context, m *multiRangeDownloaderMana
 	}
 }
 
-type mrdErrorCmd struct {
```
cpriti-os, line 186: We do not need this anymore?

storage/grpc_reader_multi_range.go
```diff
@@ -221,19 +334,25 @@ type multiRangeDownloaderManager struct {
...
+	lastReadHandle     []byte
+	pendingRangesCount int
+	attrs              *ReaderObjectAttrs
+	attrsReady         chan struct{}
+	attrsOnce          sync.Once
+	spanCtx            context.Context
+	callbackWg         sync.WaitGroup
+	streamCreating     bool
+	streamPicker       streamPickerStrategy
+	sessionIDCounter   int
```
cpriti-os: why is this session counter and not stream Counter?

storage/grpc_reader_multi_range.go
```diff
@@ -352,12 +473,34 @@ func (m *multiRangeDownloaderManager) runCallback(origOffset, numBytes int64, er
 	}()
 }
 
+func (m *multiRangeDownloaderManager) getReqAndTargetStream(req *rangeRequest) (*storagepb.BidiReadObjectRequest, *mrdStream) {
+	streamToSend := m.streamPicker.pick(m.streams)
```
cpriti-os: nit: streamID?

storage/grpc_reader_multi_range.go
```diff
@@ -374,23 +517,23 @@ func (m *multiRangeDownloaderManager) eventLoop() {
 		m.callbackWg.Wait()
 	}()
 
-	// Blocking call to establish the first session and get attributes.
-	if err := m.establishInitialSession(); err != nil {
```
cpriti-os, line 378: why did we move this to the New function from here?

cpriti-os (changes requested):

storage/grpc_reader_multi_range.go
```diff
@@ -421,17 +578,61 @@ func (m *multiRangeDownloaderManager) eventLoop() {
...
+	}(id, clonedSpec)
+}
+
+func (m *multiRangeDownloaderManager) createNewSession(id int, readSpec *storagepb.BidiReadObjectSpec, waitForResult bool) (*bidiReadStreamSession, *storagepb.BidiReadObjectSpec, error) {
 	retry := m.settings.retry
 
 	var firstResult mrdSessionResult
+	var newSession *bidiReadStreamSession
+	// Make a copy of the spec to work with internally.
+	currentSpec := proto.Clone(readSpec).(*storagepb.BidiReadObjectSpec)
```
cpriti-os: Why are we cloning it again here, are we not sending a cloned spec anyways?

storage/grpc_reader_multi_range.go
krishnamd-jkp: Yes it was not used earlier either.

storage/grpc_reader_multi_range.go
krishnamd-jkp: consolidated this into a single function which is reused for opening new streams, reconnecting streams as well.

storage/grpc_reader_multi_range.go
krishnamd-jkp: Removed it.

storage/integration_test.go
```diff
@@ -90,8 +90,8 @@ const (
 	testUniverseLocation   = "TEST_UNIVERSE_LOCATION"
 	testUniverseCreds      = "TEST_UNIVERSE_DOMAIN_CREDENTIAL"
 	// Location and Zone for zonal buckets tests
-	testZonalLocation = "us-west4"
-	testZonalZone     = "us-west4-a"
+	testZonalLocation = "us-central1"
```
cpriti-os: Revert these

storage/grpc_reader_multi_range.go
```diff
@@ -55,34 +59,88 @@ type internalMultiRangeDownloader interface {
...
+
 // --- grpcStorageClient method ---
 // Top level entry point into the MultiRangeDownloader via the storageClient interface.
 func (c *grpcStorageClient) NewMultiRangeDownloader(ctx context.Context, params *newMultiRangeDownloaderParams, opts ...storageOption) (*MultiRangeDownloader, error) {
 	if !c.config.grpcBidiReads {
 		return nil, errors.New("storage: MultiRangeDownloader requires the experimental.WithGRPCBidiReads option")
 	}
 	s := callSettings(c.settings, opts...)
+	// Force the use of the custom codec to enable zero-copy reads.
+	s.gax = append(s.gax, gax.WithGRPCOptions(
```
cpriti-os, line 131: Why did we not need this before?

storage/grpc_reader_multi_range.go
```diff
@@ -575,172 +801,254 @@ func (m *multiRangeDownloaderManager) handleCloseCmd(ctx context.Context, cmd *m
 }
 
 func (m *multiRangeDownloaderManager) handleWaitCmd(ctx context.Context, cmd *mrdWaitCmd) {
-	if len(m.pendingRanges) == 0 {
+	if m.pendingRangesCount == 0 && m.unsentRequests.Len() == 0 {
```
cpriti-os, line 807: when will pending ranges be 0 when there are unsentRequests? Add a small comment

storage/grpc_reader_multi_range.go
```diff
@@ -575,172 +801,254 @@ func (m *multiRangeDownloaderManager) handleCloseCmd(ctx context.Context, cmd *m
...
+		stream.totalRanges = 0
+		stream.totalRangeBytes = 0
+		stream.atCapacity = false
+	}
+	// fail all the ranges in unsent requests
+	for req := m.unsentRequests.l.Front(); req != nil; req = req.Next() {
+		m.failRange(nil, req.Value.(*rangeRequest), err)
 	}
-	m.pendingRanges = make(map[int64]*rangeRequest)
+	m.unsentRequests.l.Init()
```
cpriti-os: Calling this directly violates encapsulation.  It would be better for requestQueue to provide its own Clear() or Reset() method.

storage/grpc_reader_multi_range.go
krishnamd-jkp: We were doing it in newBidiStream method but now that will be called by multiple go routines which modify a global settings struct leading to race conditions. So I just moved it so it is done once.

storage/reader.go
```diff
@@ -154,6 +154,73 @@ func (o *ObjectHandle) NewRangeReader(ctx context.Context, offset, length int64)
...
+
+func (c minConnections) apply(params *newMultiRangeDownloaderParams) {
+	params.minConnections = int(c)
+}
+
+// withMinConnections returns an mrdOption which sets minimum connections
+// on the MRD to c. The call to NewMultiRangeDownloader will create one connection
+// and return with an MRD. The remaining connections will be created in the background
+// to avoid open latency.
+func withMinConnections(c int) mrdOption {
```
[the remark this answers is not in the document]
krishnamd-jkp: Will export these after this PR.

storage/integration_test.go
```diff
@@ -627,6 +642,98 @@ func TestIntegration_MRDWithReadHandle(t *testing.T) {
...
+		}
+
+		// Wait for all goroutines to finish adding their ranges.
+		wg.Wait()
+		// Wait for all reads to complete.
+		reader.Wait()
+
+		if manager.streamIDCounter != manager.params.maxConnections {
+			t.Fatalf("Manager did not scale up to maxConnections; got %d, want %d", len(manager.streams), manager.params.maxConnections)
+		}
```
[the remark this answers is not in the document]
krishnamd-jkp: IDs start from 0. Changed comparison to manager.streamIDCounter in message.

cpriti-os (approved):

Issue #10822, opened 2024-09-05
WithIgnoreFieldMismatch option works for just first record in embedded / nested struct array case.


smrfs:
**Current Behavior**
When using WithIgnoreFieldMismatch option, I cannot retrieve after second records in embedded / nested struct array case, and always need to implement Loader / Saver methods per a entity also workaround to ignore the error ErrFieldMismatch in Loader method.

**Requested Behavior**
Please enhance WithIgnoreFieldMismatch option behavior to retrieve all records even in embedded / nested struct array case.

**Reproduced Code**
Please confirm following code behavior like following 3 steps.
1. Uncomment CASE1 field and code and run the code. Note that you can retrieve 2 records.
2. Undo the step1 uncomments and run the code. Note that you can only retrieve 1 record.
3. Uncomment CASE OPT code and use known workaround. Note that you can retrieve 2 records.

```
package main

import (
	"context"
	"log"
	"os"

	"cloud.google.com/go/datastore"
)

type Tier1 struct {
	Tier1Field string
	Tier2      []*Tier2
}

type Tier2 struct {
	Tier2Field1 string

	// CASE1: First time, uncomment / enable following field to put the data and retrieve it correctly.
	// CASE2: Second time, undo / comment out following field to reproduct the issue.
	//Tier2Field2 string
}

func (entity *Tier2) Load(p []datastore.Property) error {
	return datastore.LoadStruct(entity, p)
	// CASE OPTION: We know that this workaround works regardless WithIgnoreFieldMismatch option use.
	/*
		err := datastore.LoadStruct(entity, p)
		if fmerr, ok := err.(*datastore.ErrFieldMismatch); ok && fmerr != nil {
			err = nil
		}
		return err
	*/
}

func (entity *Tier2) Save() ([]datastore.Property, error) {
	return datastore.SaveStruct(entity)
}

func main() {
	os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", "menu4today-850612-eb5fb6a47e42.json")

	ctx := context.Background()
	client, err := datastore.NewClient(ctx, "menu4today-850612", datastore.WithIgnoreFieldMismatch())
	if err != nil {
		panic(err)
	}

	key := datastore.NameKey("TempEntity", "test", nil)

	// CASE1: First time, use following code to put the data and retrieve it correctly.
	// CASE2: Second time, undo / comment out following code to reproduct the issue.
	/*
		newEntity := &Tier1{
			Tier1Field: "tier1",
[...]

PR #13799, opened 2026-02-05
datastore: fix WithIgnoreFieldMismatch for nested slices

bhshkh:
Fixes: #10822

This change ensures that propertyLoader.load continues iterating through slice elements even if individual elements encounter loading errors (like ErrFieldMismatch).

Previously, when loading a slice of structs, if one element caused an error (e.g. valid field mismatch), the loader would abort immediately. This prevented subsequent elements from being loaded, even if the user intended to ignore these mismatches using WithIgnoreFieldMismatch.

The fix captures the first error encountered but continues the loop to ensure all valid data is loaded. The error is returned at the end, preserving the existing error reporting behavior (which WithIgnoreFieldMismatch relies on).

### DIFF
--- a/datastore/load.go
+++ b/datastore/load.go
@@ -78,13 +78,16 @@ func (l *propertyLoader) load(codec fields.List, structValue reflect.Value, p Pr
 	if !ok {
 		return l.loadOneElement(codec, structValue, p, prev)
 	}
+	var firstErr string
 	for _, val := range sl {
 		p.Value = val
 		if errStr := l.loadOneElement(codec, structValue, p, prev); errStr != "" {
-			return errStr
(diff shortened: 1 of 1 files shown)


datastore/load.go
```diff
@@ -78,13 +78,16 @@ func (l *propertyLoader) load(codec fields.List, structValue reflect.Value, p Pr
 	if !ok {
 		return l.loadOneElement(codec, structValue, p, prev)
 	}
+	var firstErr string
```
shollyman, line 81: This looks odd to me because you're propagating strings rather than errors, but it adheres to the loadOneElement contract.

shollyman (approved):

PR #13815, opened 2026-02-09
fix(storage): fix panic error in transfermanager downloads

krishnamd-jkp:


storage/transfermanager/downloader_test.go
```diff
@@ -510,116 +510,111 @@ func TestCalculateRange(t *testing.T) {
...
-	if ok {
-		t.Error("context was not cancelled")
-	}
-
-	if ctxErr := context.Cause(in.ctx); !errors.Is(ctxErr, errCancelAllShards) {
-		t.Errorf("context.Cause: error should wrap %q; intead got: %v", errCancelAllShards, ctxErr)
-	}
-
-	wg.Wait()
+			// WaitAndClose will wait for downloadsInProgress, including our gatherShards goroutine.
```
cpriti-os: Nit: remove 'our'

storage/transfermanager/downloader.go
```diff
@@ -485,9 +485,11 @@ func (d *Downloader) gatherShards(in *DownloadObjectInput, out *DownloadOutput,
 			// If a shard errored, track the error and cancel the shared ctx.
 			errs = append(errs, shardOut.Err)
 			in.cancelCtx(errCancelAllShards)
+			continue
+		}
+		if shardOut.shard > 0 {
+			orderedChecksums[shardOut.shard-1] = crc32cPiece{sum: shardOut.crc32c, length: shardOut.shardLength}
```
cpriti-os, line 492: do we need a guard rail for ordered checksums as well?

storage/transfermanager/downloader_test.go
```diff
@@ -510,116 +510,111 @@ func TestCalculateRange(t *testing.T) {
...
+				{shard: 2, crc32c: crc32c([]byte(piece3)), shardLength: int64(len(piece3))},
+			},
+			checkCRC: true,
+			wantErr:  false,
+		},
+		{
+			desc:   "one shard fails",
+			shards: 3,
+			outputs: []*DownloadOutput{
+				{shard: 0, Err: errors.New("shard failure")},
```
cpriti-os, line 543: will shared 0 never arrive without error?

storage/transfermanager/downloader.go
```diff
@@ -485,9 +485,11 @@ func (d *Downloader) gatherShards(in *DownloadObjectInput, out *DownloadOutput,
 			// If a shard errored, track the error and cancel the shared ctx.
 			errs = append(errs, shardOut.Err)
 			in.cancelCtx(errCancelAllShards)
+			continue
+		}
+		if shardOut.shard > 0 {
```
cpriti-os: Nit: a comment to explain why shard 0 is skipped

cpriti-os (changes requested):

storage/transfermanager/downloader.go
krishnamd-jkp: The panic is caused by this unsafe indexing so it is important to have this check.

storage/transfermanager/downloader_test.go
krishnamd-jkp: No it won't arrive. gatherShards collects all shards except for the first one. So unless there is an error shardOut.shard will never be 0

storage/transfermanager/downloader.go
krishnamd-jkp: I think it is already clear that it is to avoid the indexing error, no?

storage/transfermanager/downloader.go
cpriti-os: yes, I meant in addition to this.

storage/transfermanager/downloader.go
cpriti-os: yes, but why we are using -1 in the next statement. A small explanation could help in the long run.

PR #13817, opened 2026-02-09
feat(datamanager): add new clients

shollyman:
PiperOrigin-RevId: 822625565

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	go.opentelemetry.io/otel/trace v1.39.0
 	golang.org/x/oauth2 v0.35.0
 	google.golang.org/api v0.265.0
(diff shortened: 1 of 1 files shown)


shollyman:
bazel rules still need updates

internal/generated/snippets/go.mod
```diff
@@ -3,7 +3,7 @@ module cloud.google.com/go/internal/generated
 go 1.24.0
 
 require (
-	cloud.google.com/go v0.123.0 // indirect
+	cloud.google.com/go v0.123.0
```
quartzmo, line 6: This is interesting, why did the snippets module dependency on the root module lose the "indirect" comment?

quartzmo (approved):

PR #13829, opened 2026-02-10
chore(preview): remove several veneers

noahdietz:
Removes some veneer'd clients that do not need to be in `preview`:
* biguqery
* bigtable
* datastore
* firestore
* pubsub
* pubsublite
* spanner
* storage

This will free up the deletion of several `internal` subpackages.

Also deletes `internal/leakcheck`.


noahdietz:
Need https://github.com/googleapis/google-cloud-go/pull/13830 to be submitted first

PR #13830, opened 2026-02-10
chore(preview): remove extraneous snippets

noahdietz:
Removes all snippets except for our initial test API compute/apiv1. Tidy up the go.mod/sum files as well.

This allows us to more easily remove the extraneous clients.


shollyman (approved):
Hold me closer, tiny dancer.

PR #13859, opened 2026-02-18
chore(internal/godocfx): support individual package .repo-metadata.json

hongalex:
The single internal/.repo-metadata-full.json file was removed in
favor of individual .repo-metadata.json files located within package
directories. `godocfx` was dependent on fetching this.

This PR updates the metadata resolution to:
1. Search upwards from the package directory to the module root for a `.repo-metadata.json` file.
2. Load the API description from the local metadata file.
3. Provide a fallback mechanism for tests and older module versions.

Also updated test goldens to reflect metadata resolution changes and
improved symbol linking.

Fixes: b/462186715

### DIFF
--- a/internal/godocfx/parse.go
+++ b/internal/godocfx/parse.go
@@ -715,13 +715,14 @@ func (d *friendlyAPINamer) friendlyAPIName(importPath string, module *packages.M
 			metaFile := filepath.Join(currDir, ".repo-metadata.json")
 			b, err := os.ReadFile(metaFile)
 			if err == nil {
-				if err := json.Unmarshal(b, &meta); err == nil {
-					d.mu.Lock()
-					d.metadata[currDir] = meta
-					d.mu.Unlock()
-					found = true
-					break
+				if err := json.Unmarshal(b, &meta); err != nil {
+					return "", fmt.Errorf("failed to decode repo metadata in %s: %v", metaFile, err)
 				}
+				d.mu.Lock()
+				d.metadata[currDir] = meta
+				d.mu.Unlock()
+				found = true
+				break
 			}
 
 			if currDir == module.Dir || currDir == "." || currDir == "/" {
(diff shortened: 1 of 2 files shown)


internal/godocfx/testdata/golden/cloud.google.com/go/storage@v1.33.0/index.yml
```diff
@@ -550,15 +550,15 @@ items:
             const (
             	// ScopeFullControl grants permissions to manage your
             	// data and permissions in Google Cloud Storage.
-            	ScopeFullControl = raw.DevstorageFullControlScope
+            	ScopeFullControl = <a href="https://pkg.go.dev/google.golang.org/api/storage/v1">raw</a>.<a href="https://pkg.go.dev/google.golang.org/api/storage/v1#DevstorageFullControlScope">DevstorageFullControlScope</a>
```
shollyman, line 553: It immediately looks a bit odd to me that html markup is injected like this in a yaml file, but I'm not super familiar here.  Are we comparing yaml content or yaml that's been rendered as markup for embedding in a webpage?

internal/godocfx/parse.go
```diff
@@ -654,67 +653,103 @@ func hasPrefix(s string, prefixes []string) bool {
...
 //
 // See https://github.com/googleapis/google-cloud-go/issues/5949.
-func (d *friendlyAPINamer) friendlyAPIName(importPath string) (string, error) {
-	var err error
-	d.getOnce.Do(func() {
-		resp, getErr := http.Get(d.metaURL)
-		if err != nil {
-			err = fmt.Errorf("error getting repo metadata: %v", getErr)
-			return
+func (d *friendlyAPINamer) friendlyAPIName(importPath string, module *packages.Module) (string, error) {
```
shollyman, line 687: This func can return an error, but it seems like this updated implementation doesn't ever return one?  Are we swallowing errors that are useful to return, like there's invalid metadata found in the traversal path?

internal/godocfx/testdata/golden/cloud.google.com/go/storage@v1.33.0/index.yml
hongalex: This wasn't a change that was introduced here. Even at head, running `go test -run=TestGoldens --update-goldens` produces this change.

internal/godocfx/parse.go
hongalex: Yeah good point. Removed the error swallowing and updated tests.

PR #13860, opened 2026-02-18
feat(bigtable): add preemptive connection recycler

sushanb:


bigtable/client.go
```diff
@@ -64,8 +65,13 @@ type ClientConfig struct {
 	// TODO: support user provided meter provider
 	MetricsProvider MetricsProvider
 
-	// If true, enable dynamic channel pool
-	EnableDynamicChannelPool bool
+	// DisableDynamicChannelPool disables the dynamic channel resizing based on load
+	// Dynamic channel resizing  is enabled by default to resize based on load and avoid queuing of requests.
+	DisableDynamicChannelPool bool
```
mutianf, line 70: This makes the code a little hard to read. Is it possible to have a constructor function that returns true for these variables by default?

bigtable/internal/transport/conn_recycler.go
```diff
@@ -0,0 +1,115 @@
...
+		done:   make(chan struct{}),
+		rng:    rand.New(rand.NewSource(time.Now().UnixNano())),
+	}
+}
+
+// Start begins the periodic monitoring.
+func (cr *ConnectionRecycler) Start(ctx context.Context) {
+	btopt.Debugf(cr.pool.logger, "bigtable_connpool: ConnectionRecyler starting...")
+
+	if !cr.config.Enabled {
```
mutianf: Is this necessary? If the client didn't enable the recycler we won't even create this object?

bigtable/internal/option/option.go
```diff
@@ -271,3 +271,26 @@ func DefaultMetricsReporterConfig() MetricsReporterConfig {
 		ReportingInterval: 1 * time.Minute,
 	}
 }
+
+// ConnectionRecycleConfig controls the behavior of the connection recycler.
+type ConnectionRecycleConfig struct {
+	// MaxAge is the base lifespan of a connection.
+	MaxAge time.Duration
+	// Jitter is the random buffer added to MaxAge which can allow for connection to be recycled.
+	Jitter time.Duration
```
mutianf: Maybe MaxJitter?

bigtable/internal/transport/conn_recycler.go
```diff
@@ -0,0 +1,115 @@
...
+		return
+	}
+	// default to 1 minute
+	freq := cr.config.RunFrequency
```
mutianf, line 54: Should we also make sure RunFrequency is < maxAge?

bigtable/internal/option/option.go
sushanb: Done.

bigtable/internal/transport/conn_recycler.go
sushanb: Done.

PR #13865, opened 2026-02-18
chore(spanner): replace `interface{}` with `any` for statement Params

Zarux:
This change is a minor quality of life change mixed with some modernisation.

Using `any` over `interface{}` is generally considered the preferred syntax in modern Go.
When creating `Params` on `Statement` gopls will suggest `map[string]interface{}{}` which will then have to be changed to `map[string]any{}` manually (or with `go fix`).

I kept the change small to minimize review time, but with the recent `go fix` changes in 1.26 i would suggest running that to catch everything.

So this PR will make that part the spanner package a smidge more enjoyable to work with as well as modernise the code a bit.

### DIFF
--- a/spanner/statement.go
+++ b/spanner/statement.go
@@ -40,12 +40,12 @@ import (
 // Spanner types.
 type Statement struct {
 	SQL    string
-	Params map[string]interface{}
+	Params map[string]any
 }
 
 // NewStatement returns a Statement with the given SQL and an empty Params map.
 func NewStatement(sql string) Statement {
-	return Statement{SQL: sql, Params: map[string]interface{}{}}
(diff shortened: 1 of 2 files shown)


rahul2393:
@Zarux Thanks for opening the PR, changes looks good to me, but why only these variables, is there any problem if we make this change for all occurances of interface in client lib?

Zarux:
It shouldn't be problem to change every occurrence to `any`, however the PR would be significantly bigger. I chose to only change here as `statement` was where i had annoyance with it not being `any` so i chose to only change there to minimise the pr size. 

I can update the PR to change every occurrence instead if you want

PR #13873, opened 2026-02-20
feat(bigtable): Use direct access when supported. Guarded by enabling…

sushanb:
Current limitations:

1. Does not use PeerInfo for checking transport
2. Does not have retry logic into Prime (only one attempt for success)
3. Guarded by enabling CBT_BIGTABLE_CONN_POOL

### DIFF
--- a/bigtable/internal/transport/connpool.go
+++ b/bigtable/internal/transport/connpool.go
@@ -345,6 +345,9 @@ type BigtableChannelPool struct {
 	directAccessFeatureFlagsMD metadata.MD
 	directAccessDialer         func() (*BigtableConn, error)
 
+	// Add the cached gauge instrument
+	daEligibleGauge metric.Int64Gauge
+
 	// background monitors
 	monitors []Monitor
 }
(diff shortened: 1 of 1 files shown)


bigtable/internal/transport/connpool.go
```diff
@@ -449,6 +491,57 @@ func NewBigtableChannelPool(ctx context.Context, connPoolSize int, strategy btop
...
+	p.reportDirectAccessMetric(false)
+	return nil, false
+}
+
+// reportDirectAccessMetric records the direct_access/compatible metric.
+func (p *BigtableChannelPool) reportDirectAccessMetric(isEligible bool) {
+	if p.meterProvider == nil {
+		return
+	}
+	meter := p.meterProvider.Meter(clientMeterName)
```
igorbernstein2: arent instruments supposed live for the lifetime of the process and not per record?

sushanb: The instrument was invoked per client lifetime once so decided to create it at fly.

Changed to lifetime .

bigtable/internal/transport/connpool.go
```diff
@@ -105,6 +105,20 @@ func WithMeterProvider(mp metric.MeterProvider) BigtableChannelPoolOption {
...
 }
 
+// WithDirectAccessDialer provides the dialer for direct access
+func WithDirectAccessDialer(directAccessDialer func() (*BigtableConn, error)) BigtableChannelPoolOption {
+	return func(p *BigtableChannelPool) {
+		p.directAccessDialer = directAccessDialer
+	}
+}
+
+// WithDirectAccessFeatureFlagsMetadata provides the feature flags required for DirectAccess
```
bhshkh, line 115: Is it documented somewhere what feature flags are ? or what metadata for feature flags is supposed to contain?

I was thinking feature flags is something internal to the library and the Bigtable service and not exposed to the user.

sushanb: yes, the client automatically creates the feature flags required for direct access and tries to use direct access. 

This is hidden to the end users of the client.

PR #13877, opened 2026-02-23
chore(spanner): integrate location aware routing with RPCs

rahul2393:
- Add `locationAwareSpannerClient` wrapper that intercepts RPCs and routes
    them to the server endpoint resolved by the existing `channelFinder`,
    falling back to the default gRPC channel when no endpoint is available
  - Add `endpointClientCache` that creates and caches per-address gRPC
    connections
  - Add transaction affinity tracking so Commit/Rollback route to the same
    server that handled the transaction, with read-only transactions routed
    independently per-request based on key ranges
  - Move request preparation (`prepareReadRequest`, etc.) and response
    observation (`observePartialResultSet`, etc.) from transaction/batch
    code into the client wrapper, keeping routing concerns in one place

### DIFF
--- a/spanner/batch.go
+++ b/spanner/batch.go
@@ -371,7 +371,6 @@ func (t *BatchReadOnlyTransaction) Execute(ctx context.Context, p *Partition) *R
 		t.setTimestamp,
 		t.release,
 		asGRPCSpannerClient(client),
-		nil,
 	)
 }
 
--- a/spanner/client.go
+++ b/spanner/client.go
@@ -491,16 +491,11 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
 
 	var pool gtransport.ConnPool
 	var endpointClientOpts []option.ClientOption
-	// When non-zero, this overrides the endpoint client pool size. This is used
-	// for GCPMultiEndpoint mode where endpoint clients should keep a single
-	// connection per endpoint.
-	var endpointClientPoolSize int
 
 	if gme != nil {
 		// Use GCPMultiEndpoint if provided.
 		pool = &gmeWrapper{gme}
 		endpointClientOpts = append(endpointClientOpts, opts...)
-		endpointClientPoolSize = 1
 	} else {
 		// Create gtransport ConnPool as usual if MultiEndpoint is not used.
 		// gRPC options.
@@ -580,7 +575,6 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
 	var locationRouter *locationRouter
 	if isExperimentalLocationAPIEnabled() {
 		sc.baseClientOpts = endpointClientOpts
-		sc.endpointClientPoolSize = endpointClientPoolSize
 		epCache := newEndpointClientCache(sc.createEndpointClient)
 		locationRouter = newLocationRouter(epCache)
 	}
--- a/spanner/endpoint_client_cache.go
+++ b/spanner/endpoint_client_cache.go
@@ -42,19 +42,31 @@ func (e *grpcChannelEndpoint) IsHealthy() bool {
 type endpointClientCache struct {
 	mu            sync.RWMutex
 	endpoints     map[string]*grpcChannelEndpoint
+	inflight      map[string]*endpointClientCreation
 	clientFactory func(ctx context.Context, address string) (spannerClient, error)
+	closed        bool
+}
+
+type endpointClientCreation struct {
+	done chan struct{}
+	ep   channelEndpoint
 }
 
 func newEndpointClientCache(clientFactory func(ctx context.Context, address string) (spannerClient, error)) *endpointClientCache {
 	return &endpointClientCache{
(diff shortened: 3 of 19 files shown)


spanner/batch.go
```diff
@@ -376,12 +370,8 @@ func (t *BatchReadOnlyTransaction) Execute(ctx context.Context, p *Partition) *R
...
 		t.setTimestamp,
 		t.release,
-		client.(*grpcSpannerClient),
-		func(prs *sppb.PartialResultSet) {
-			if t.locationRouter != nil {
-				t.locationRouter.observePartialResultSet(prs)
-			}
-		},
+		asGRPCSpannerClient(client),
+		nil,
```
olavloite: I think that we can remove the callback argument entirely from this function, as this is now handled through the client.

spanner/client.go
```diff
@@ -490,10 +490,17 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
 	}
 
 	var pool gtransport.ConnPool
+	var endpointClientOpts []option.ClientOption
+	// When non-zero, this overrides the endpoint client pool size. This is used
+	// for GCPMultiEndpoint mode where endpoint clients should keep a single
+	// connection per endpoint.
+	var endpointClientPoolSize int
```
olavloite: This feels a bit weird, as it seems to indicate that `endpointClientPoolSize` could have any value. But in reality it is only 0 or 1. Would it make more sense to make this a bool? Or in some other way refactor away the entire variable (including the same field in `sessionClient`), and instead only rely on `gme != nil`?

spanner/endpoint_client_cache.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	c.mu.RUnlock()
+
+	// Slow path: write lock + create.
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if ep, ok := c.endpoints[address]; ok {
+		return ep
+	}
+	client, err := c.clientFactory(context.Background(), address)
```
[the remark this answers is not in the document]
olavloite: @rahul2393 I think that this is a valid concern. This function creates a new gRPC channel, which could be a slow operation. The `endpointClientCache#Get(..)` method is ultimately called from functions like `ExecuteSql` and similar, and those do accept a context that could be passed in here (through multiple layers). I think it would make sense to plug that through.

Another concern here is that the gRPC channel creation happens while holding `c.mu`. This will block all readers and writers of the cache, which means that one slow gRPC channel creation could in theory slow down the entire client. Can we try to refactor the locking here in such a way that channel creation does not happen while holding an exclusive lock?

spanner/location_aware_client.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+// --- Pass-through methods ---
+
+func (c *locationAwareSpannerClient) CallOptions() *vkit.CallOptions {
+	return c.defaultClient.CallOptions()
+}
+
+func (c *locationAwareSpannerClient) Close() error {
+	return c.defaultClient.Close()
```
olavloite: Are you sure that we should close this client here? Isn't this a shared client?

spanner/location_aware_client.go
```diff
@@ -0,0 +1,346 @@
...
+// contains a transaction ID.
+type affinityTrackingStream struct {
+	grpc.ClientStream
+	router             *locationRouter
+	affinityEndpoint   channelEndpoint
+	trackReadOnlyBegin bool
+	readOnlyStrong     bool
+	trackAffinity      bool
+	once               sync.Once
+	inner              spannerpb.Spanner_ExecuteStreamingSqlClient
```
olavloite: nit: maybe add a small comment here to indicate that this is actually used for both ExecuteStreamingSql and StreamingRead. This declaration makes it look like it is only used for ExecuteStreamingSql, but because the interface for StreamingRead is exactly the same, it actually works for both.

spanner/sessionclient.go
```diff
@@ -106,6 +106,13 @@ type sessionClient struct {
 	metricsTracerFactory *builtinMetricsTracerFactory
 	channelIDMap         map[*grpc.ClientConn]uint64
 
+	// baseClientOpts holds the client options used for creating endpoint-specific
+	// gRPC connections in location-aware routing.
+	baseClientOpts []option.ClientOption
+	// endpointClientPoolSize overrides the gRPC connection pool size for
+	// endpoint-specific clients when non-zero.
+	endpointClientPoolSize int
```
olavloite: See also above: It is a bit weird that this in reality only has value 0 or 1. If we cannot easily refactor that, can we at least add a comment that explains that.

spanner/location_aware_client.go
rahul2393: Good catch. This wrapper does not own the underlying round-robin client; that client is shared and is closed by Client.Close(). I changed locationAwareSpannerClient.Close() to be a no-op and added a test to make that ownership explicit.

PR #13879, opened 2026-02-23
chore(all): updates towards go 1.26

shollyman:
This PR represents the move to include go 1.26 as the new upper bound for this repository, and performs common tasks like advancing the minimum version of go in accordance with the supported version policy for go, and update various infrastructure tools like testing scripts.


shollyman:
Will handle the cleanup of apidiff.yaml in a followup change.

PR #13881, opened 2026-02-23
chore: remove outdated team/accounts from CODEOWNERS file

suztomo:
Removing outdated cloud-native-db-dpes team from CODEOWNERS (b/486902511).

This pull request also removes individual accounts from the entries because the bigtable-team already has the account. Igor is part of bigtable-team. Jonathan responded it's ok to remove.

b/479547457

As part of this change, we are also removing admin permissions from the repository settings in .github/sync-repo-settings.yaml. Teams that previously had admin access have been removed or updated to have push access.


suztomo:
I may be able to add a change for b/479547457#comment18.

noahdietz:
Swapped to Go lead @shollyman for review.

PR #13887, opened 2026-02-24
fix(internal/godocfx): search subdirectories for metadata as well

hongalex:
#13859 was the precursor to this, but only searched from package -> root module, and thus couldn't find `.repo-metadata.json` files for packages like [Translate](https://github.com/googleapis/google-cloud-go/blob/main/translate/translate.go)

The appropriate file lives at `translate/apiv3/.repo-metadata.json`, so we need to walk through the subdirectories as well.

There's one remaining issue that the zipped version of translate (v1.27.1) doesn't actually contain this new json file, so simply running `./godocfx cloud.google.com/go/translate` will fail, but `./godocfx cloud.google.com/go/translate@main` (at head) will succeed. This should be fine since publish_docs should only be done on new releases, but if this assumption is wrong, we can update this as a fallback.

### DIFF
--- a/internal/godocfx/parse.go
+++ b/internal/godocfx/parse.go
@@ -744,29 +744,33 @@ func (d *friendlyAPINamer) friendlyAPIName(importPath string, module *packages.M
 			// This is useful when the root module directory doesn't have it, but its subdirectories (like apiv*) do.
 			// Example: For root pkg cloud.google.com/go/auditmanager, search module subdirectories
 			// for the first instance of .repo-metadata.json (e.g. in apiv1/).
-			filepath.Walk(module.Dir, func(path string, info os.FileInfo, err error) error {
-				if err != nil || found {
+			err := filepath.WalkDir(module.Dir, func(path string, de os.DirEntry, err error) error {
+				if err != nil {
 					return err
 				}
 				// Skip subdirectories that are separate modules.
 				// This is needed to avoid overlap in nested modules, like in cloud.google.com/go/pubsub/v2
-				if path != module.Dir && info.IsDir() {
+				if path != module.Dir && de.IsDir() {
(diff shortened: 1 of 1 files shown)


internal/godocfx/parse.go
```diff
@@ -734,6 +738,49 @@ func (d *friendlyAPINamer) friendlyAPIName(importPath string, module *packages.M
 			}
 			currDir = parent
 		}
+
+		if !found {
+			// Search subdirectories for .repo-metadata.json.
```
shollyman, line 743: This feels a bit itchy, since we'll just find the first metadata file we encounter and use that.  How often does this occur in practice?  For APIs where this is the case, are all the metadata files in the various subdirectories consistent with regards to name we're pulling from the metadata files?

I'm mostly trying to understand if we have missing or mislocated metadata files, and we should address that.

hongalex: There is at least one missing metadata file (pubsub/v2), but I haven't found evidence of more beyond that.

> For APIs where this is the case, are all the metadata files in the various subdirectories consistent with regards to name we're pulling from the metadata files?

The bulk of metadata files were generated into `apiv1`, and those all match the product name. There are a few exceptions which makes matching an `apiv{0-9}.*` subdirectory pattern harder, such as [longrunning](https://github.com/googleapis/google-cloud-go/tree/main/longrunning/autogen), or maybe we can make an exception for that using the fallback mechanism.

I think [Storage](https://github.com/googleapis/google-cloud-go/tree/main/storage) was a bit iffy, but transfermanager & dataflux would report GCS as the description, whereas the storage/control/apiv2 has its own [metadata](https://github.com/googleapis/google-cloud-go/blob/main/storage/control/apiv2/.repo-metadata.json) file.

PR #13888, opened 2026-02-24
chore: handle permanent errors in MRD multistream

krishnamd-jkp:


storage/grpc_reader_multi_range.go
```diff
@@ -804,7 +837,7 @@ func (m *multiRangeDownloaderManager) handleWaitCmd(ctx context.Context, cmd *mr
 	// unsentRequests could be non-empty when eventLoop is busy
 	// in select statements other than Add commands and cleared up
 	// existing pending ranges.
-	if m.pendingRangesCount == 0 && m.unsentRequests.Len() == 0 {
+	if m.pendingRangesCount <= 0 && m.unsentRequests.Len() == 0 {
```
cpriti-os: Why <, can it be -1?

storage/grpc_reader_multi_range.go
```diff
@@ -605,13 +635,17 @@ func (m *multiRangeDownloaderManager) addNewStream() {
...
+		m.streamCreating = false
+		return
+	}
 	go func(id int, readSpec *storagepb.BidiReadObjectSpec) {
 		newSession, newSpec, err := m.createNewSession(id, readSpec, false)
 		if err != nil || newSession == nil {
-			// If we can't create a stream, just reset the flag so we can try again later.
-			// We don't want to kill the whole manager for a single failed dynamic stream.
+			// If we can't create a stream, the handler checks the health of
+			// manager and then decides to kill the manager.
```
cpriti-os, line 644: Are we killing the manager if we have one failed stream?

storage/grpc_reader_multi_range.go
```diff
@@ -831,12 +864,15 @@ func (m *multiRangeDownloaderManager) handleReconnectStreamCmd(ctx context.Conte
...
-			m.setPermanentError(errClosed)
+		m.failStream(stream, cmd.err)
+		if len(m.streams) == 0 && !m.streamCreating {
+			if !errors.Is(cmd.err, context.Canceled) && !errors.Is(cmd.err, errClosed) {
+				m.setPermanentError(cmd.err)
+			} else if m.getPermanentError() == nil {
+				m.setPermanentError(errNoStreams)
+			}
+			m.failAllPending(m.getPermanentError())
 		}
```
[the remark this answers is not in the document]
cpriti-os: +1

storage/grpc_reader_multi_range.go
krishnamd-jkp: Yes if it so happens that when all the streams are dead and there is no stream getting created, the eventloop can come to a stop.

storage/grpc_reader_multi_range.go
krishnamd-jkp: This is just a safety check with pendingRangesCount being updated in batches (during failed streams) or individually (when completing a range). The event loop will not exit if pendingRangesCount doesn't reach 0 exactly.

storage/grpc_reader_multi_range.go
```diff
@@ -312,10 +322,24 @@ func (c *reconnectStreamCmd) apply(ctx context.Context, m *multiRangeDownloaderM
...
 
 func (c *mrdAddStreamErrorCmd) apply(ctx context.Context, m *multiRangeDownloaderManager) {
 	m.streamCreating = false
+	if len(m.streams) == 0 {
+		finalErr := c.err
+		if finalErr == nil {
+			finalErr = m.getPermanentError()
+		}
+		if finalErr == nil {
+			finalErr = errNoStreams
```
cpriti-os: Can we also add why the last stream creation attempt failed, no streams available does not give enough information for debugging.

storage/grpc_reader_multi_range.go
cpriti-os, line 901: In a rare case, this could cause a deadlock right if reqC is full/blocked/slow?

storage/grpc_reader_multi_range.go
```diff
@@ -557,7 +587,7 @@ func (m *multiRangeDownloaderManager) eventLoop() {
 			m.addNewStream()
 		}
 		// Notify waiters if all ranges are done.
-		if m.pendingRangesCount == 0 && m.unsentRequests.Len() == 0 {
+		if m.pendingRangesCount <= 0 && m.unsentRequests.Len() == 0 {
```
cpriti-os: <0 may mask potential bugs where ranges are not removed  properly. Doesn't look like we should handle <= 0 scenario here and anywhere else.

storage/grpc_reader_multi_range.go
```diff
@@ -1026,6 +1063,20 @@ func (m *multiRangeDownloaderManager) failRange(mrdStream *mrdStream, req *range
...
+		if !req.completed {
+			totalBytes += req.length - req.bytesWritten
+			req.completed = true
+			m.runCallback(req.origOffset, req.bytesWritten, err, req.callback)
+		}
+	}
+	mrdStream.updateCapacity(m, -len(mrdStream.pendingRanges), -totalBytes)
+	mrdStream.pendingRanges = make(map[int64]*rangeRequest)
+	delete(m.streams, mrdStream.id)
+}
```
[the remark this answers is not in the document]
cpriti-os: Verify this

storage/grpc_reader_multi_range.go
krishnamd-jkp: We use the actual error when present. We're only adding "no streams available" when the error is nil.

storage/grpc_reader_multi_range.go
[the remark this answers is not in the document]
krishnamd-jkp: Simplified it. Wouldn't need a special function again

storage/grpc_reader_multi_range.go
krishnamd-jkp: Added a more descriptive error

storage/grpc_reader_multi_range.go
krishnamd-jkp: Removed

storage/grpc_reader_multi_range.go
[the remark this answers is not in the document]
cpriti-os: This function  might still be a good addition? 
```
func (m *multiRangeDownloaderManager) checkAndFailManager(streamErr error) {
    if len(m.streams) == 0 && !m.streamCreating {
        err := fmt.Errorf("no streams available. Last observed error: %w", streamErr)
        m.setPermanentError(err)
        m.failAllPending(m.getPermanentError())
    }
}
```

storage/grpc_reader_multi_range.go
```diff
@@ -79,13 +84,18 @@ func (m *minBytesPicker) pick(streams map[int]*mrdStream) int {
 		if len(stream.session.reqC) >= cap(stream.session.reqC) {
 			continue
 		}
-		if minBytes == -1 || stream.totalRangeBytes < minBytes {
-			minBytes = stream.totalRangeBytes
+
+		// Calculate normalized score.
+		// Score = (PendingRanges / TargetRanges) + (PendingBytes / TargetBytes)
+		// Lower score is better (least loaded).
+		score := float64(stream.totalRanges)/float64(p.targetPendingRanges) +
```
cpriti-os, line 91: can these be zero here?

krishnamd-jkp: Made sure the default values for target pending ranges and target pending bytes are positive during the initialization.

storage/grpc_reader_multi_range.go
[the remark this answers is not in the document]
krishnamd-jkp: This seems to only check for one reason (there could be more reasons to fail the manager in the future)  . failManager (if added) should not check for a specific condition. That is why I didn't add a new function for 2 lines.

storage/grpc_reader_multi_range.go
krishnamd-jkp: reqC will not be full because the stream just got created after reconnection. If it is slow or blocked, the event loop will also be blocked if event loop is handling this message (yes, it could process other channels). If send on a new gRPC session is failing, it could be a bigger issue with the transport layer.

cpriti-os (approved):

PR #14021, opened 2026-03-01
test(storage): Fix a write integration test.

cjc25:
The writer integration edge cases test included an unnecessary deletion. It also was a little bit brittle to specific implementation details of when flushes are implicitly scheduled on both the client and server side.

### DIFF
--- a/storage/integration_test.go
+++ b/storage/integration_test.go
@@ -4200,8 +4200,8 @@ func TestIntegration_WriterAppendEdgeCases(t *testing.T) {
 			t.Fatalf("w2.Close: %v", err)
 		}
 
-		// If we add yet another takeover writer to finalize the object, tw should
-		// return an error on flush.
+		// If we add yet another takeover writer to finalize the object, that should
+		// succeed.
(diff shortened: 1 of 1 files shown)


cpriti-os:
Patched and tried this locally, failed with same error :'( 
```
=== RUN   TestIntegration_WriterAppendEdgeCases/zonalBucket
    integration_test.go:4219: tw.Write or tw.Flush: got error rpc error: code = NotFound desc = No such object: golang-grpc-test-20260302-36769735971304-0003/object1, want FailedPrecondition or Aborted
```

cjc25:
> Patched and tried this locally, failed with same error :'(
> 
> ```
> === RUN   TestIntegration_WriterAppendEdgeCases/zonalBucket
>     integration_test.go:4219: tw.Write or tw.Flush: got error rpc error: code = NotFound desc = No such object: golang-grpc-test-20260302-36769735971304-0003/object1, want FailedPrecondition or Aborted
> ```

Oops, I missed that `w2` overwrites `tw`'s generation. NOT_FOUND can indeed be returned in that case, since the object generation `tw` refers to was implicitly deleted by the overwrite.

cpriti-os:
Thanks for the investigation and fix! :)

cpriti-os (approved):

PR #14123, opened 2026-03-04
feat(bigtable): make downscalling less aggressive

sushanb:
will only downscale if three consecutive runs happen.


bigtable/internal/transport/dynamic_scale_monitor_test.go
```diff
@@ -116,7 +124,8 @@ func TestDynamicChannelScaling(t *testing.T) {
 				setConnLoads(conns, 10, 0) // Avg load 10, right at threshold.
 			},
 			// Total load = 20. Desired = ceil(20 / 6.5) = 4. Add 2.
-			wantSize: 4,
+			wantSize:      4,
+			evaluateCalls: 1,
```
jackdingilian, line 129: Can we add a new case that it doesn't scale down with 2 evaluate calls?

sushanb: done.

PR #14126, opened 2026-03-05
chore(storage): add DirectPath diagnostic info in gRPC request header

cpriti-os:


storage/grpc_dp_diag.go
```diff
@@ -0,0 +1,125 @@
...
+
+// directPathDiagnostic evaluates the provided options and environment to determine
+// why gRPC DirectPath (high-throughput VPC routing) is not being utilized.
+func directPathDiagnostic(ctx context.Context, opts ...option.ClientOption) string {
+	res, err := internaloption.NewUnsafeResolver(opts...)
+	if err != nil {
+		return reasonEndpointFetchError
+	}
+
+	if strings.EqualFold(os.Getenv(directPathDisableEnvVar), "true") {
```
krishnamd-jkp, line 55: Move this to the top of the function before resolving options.

storage/grpc_dp_diag.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return reasonUndetermined
+}
+
+func isDirectPathCompatible(endpoint string) bool {
+	if endpoint == "" {
+		return false
+	}
+	// DirectPath requires no scheme or the dns:/// scheme specifically.
+	if strings.Contains(endpoint, "://") && !strings.HasPrefix(endpoint, "dns:///") {
```
krishnamd-jkp, line 122: The condition and the comments do not look like the agree with each other. Maybe I'm missing something?

storage/grpc_dp_diag.go
```diff
@@ -0,0 +1,125 @@
...
+	defaultKey             = "default"
+	serviceAccountTokenKey = "instance/service-accounts/default/token"
+)
+
+// directPathDiagnostic evaluates the provided options and environment to determine
+// why gRPC DirectPath (high-throughput VPC routing) is not being utilized.
+func directPathDiagnostic(ctx context.Context, opts ...option.ClientOption) string {
+	res, err := internaloption.NewUnsafeResolver(opts...)
+	if err != nil {
+		return reasonEndpointFetchError
```
krishnamd-jkp: Is this the right error when we cannot resolve options?

storage/grpc_dp_diag.go
```diff
@@ -0,0 +1,125 @@
...
+
+	if res.ResolvedGRPCConnIsCustom() {
+		return reasonCustomGRPCConn
+	}
+
+	if res.ResolvedHTTPClientIsCustom() {
+		return reasonCustomHTTPClient
+	}
+
+	if !metadata.OnGCE() {
```
krishnamd-jkp, line 89: Do you know if this induces additional latency?

cpriti-os: It'll induce some minimal latency but this is only done while client creation and then cached.

storage/grpc_dp_diag.go
cpriti-os: Done

storage/grpc_dp_diag.go
cpriti-os: We are returning false if prefix dns:/// doesn't exist which is what the comment says. More specifically if there is a scheme (://) then it should be dns:/// for direct path compatibility.

storage/grpc_dp_diag.go
cpriti-os: threw Internal Error

PR #14147, opened 2026-03-09
feat(bigtable): derive conn pool from resolver

sushanb:


bigtable/client.go
```diff
@@ -165,6 +165,18 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
 
 	enableBigtableConnPool := btopt.EnableBigtableConnectionPool()
 	if enableBigtableConnPool {
+		uResolver, err := internaloption.NewUnsafeResolver(o...)
+		if err != nil {
+			// uResolver never returns err
+			return nil, fmt.Errorf("failed to resolve client options: %w", err)
```
shollyman: Rather than failing on the experimental API, might be better to simply use your default pool size instead.

sushanb: Done.

PR #14157, opened 2026-03-11
Revert "chore: librarian release pull request: 20260311T074420Z"

cpriti-os:
Reverts googleapis/google-cloud-go#14151

### DIFF
--- a/.librarian/state.yaml
+++ b/.librarian/state.yaml
@@ -5889,7 +5889,7 @@ libraries:
       - internal/generated/snippets/speech/
     tag_format: '{id}/v{version}'
   - id: storage
-    version: 1.61.1
+    version: 1.61.0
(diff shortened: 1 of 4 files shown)


quartzmo:
What's the current state of the `1.61.1` release? What happened?

quartzmo:
It looks like https://pkg.go.dev/cloud.google.com/go/storage@v1.61.1 was successfully released.

Placing `do not merge` on this rollback pending further investigation.

krishnamd-jkp:
We noticed a small dependency issue with release 1.61.1 release where storage's go.mod used `cloud.google.com/go/auth@v0.18.3-0.20260310051336-87cdcc9f7568` instead of a stable release. Since it is a minor change, we thought we'd rollback the release instead of creating a new release. 

(I think the auto merge went through despite adding the `do not merge` label.)

PR #14160, opened 2026-03-11
feat(bigtable): add CBT_ENABLE_DIRECTPATH=false in case user want to disable directpath

sushanb:
2. Enable sending the same feature flags

### DIFF
--- a/bigtable/internal/directaccess/loopback.go
+++ b/bigtable/internal/directaccess/loopback.go
@@ -17,10 +17,13 @@ package directaccess
 import (
 	"fmt"
 	"net"
+
+	btopt "cloud.google.com/go/bigtable/internal/option"
 )
 
 // CheckLoopbackInterfaceUp verifies that at least one loopback interface is UP.
 func CheckLoopbackInterfaceUp() error {
+	btopt.Debugf(nil, "directaccess: Checking for UP loopback interfaces")
 	ifaces, err := net.Interfaces()
 	if err != nil {
 		return fmt.Errorf("failed to list network interfaces: %w", err)
@@ -53,24 +56,42 @@ func onlyLoopback(iface net.Interface) error {
 	return nil
 }
 
+// CheckLocalIPv6Addresses searches the host's non-loopback network interfaces
+// to verify if the specified IPv6 address is currently plumbed and active.
+// It returns the interface where the address is configured, or an error if not
 func CheckLocalIPv6Addresses(ip *net.IP) (*net.Interface, error) {
+	btopt.Debugf(nil, "directaccess: Searching for local IPv6 address: %s", ip.String())
 	return findLocalAddress(func(i net.IP) bool { return i.To4() == nil && i.Equal(*ip) }, skipLoopback)
 }
 
+// CheckLocalIPv4Addresses searches the host's non-loopback network interfaces
+// to verify if the specified IPv4 address is currently plumbed and active.
+// It returns the interface where the address is configured, or an error if not found.
 func CheckLocalIPv4Addresses(ip *net.IP) (*net.Interface, error) {
+	btopt.Debugf(nil, "directaccess: Searching for local IPv4 address: %s", ip.String())
 	return findLocalAddress(func(i net.IP) bool { return i.To4() != nil && i.Equal(*ip) }, skipLoopback)
 }
 
+// CheckLocalIPv6LoopbackAddress verifies that the standard IPv6 loopback
+// address (::1) is configured and assigned to a local loopback interface.
 func CheckLocalIPv6LoopbackAddress() error {
+	btopt.Debugf(nil, "directaccess: Searching for IPv6 loopback (::1)")
 	_, err := findLocalAddress(func(i net.IP) bool { return i.Equal(net.ParseIP("::1")) }, onlyLoopback)
 	return err
 }
 
+// CheckLocalIPv4LoopbackAddress verifies that the standard IPv4 loopback
+// address (127.0.0.1) is configured and assigned to a local loopback interface.
 func CheckLocalIPv4LoopbackAddress() error {
+	btopt.Debugf(nil, "directaccess: Searching for IPv4 loopback (127.0.0.1)")
 	_, err := findLocalAddress(func(i net.IP) bool { return i.Equal(net.ParseIP("127.0.0.1")) }, onlyLoopback)
 	return err
 }
 
+// findLocalAddress is a generic helper that iterates through all local network interfaces.
+// It applies the ifaceFilter to skip unwanted interfaces (e.g., ignoring loopbacks
+// or inactive NICs) and checks if any assigned IP satisfies the ipMatches condition.
+// It returns the first matching network interface, or an error if exhausted.
 func findLocalAddress(ipMatches func(net.IP) bool, ifaceFilter func(net.Interface) error) (*net.Interface, error) {
 	ifaces, err := net.Interfaces()
 	if err != nil {
@@ -90,5 +111,6 @@ func findLocalAddress(ipMatches func(net.IP) bool, ifaceFilter func(net.Interfac
 			}
 		}
 	}
+	btopt.Debugf(nil, "directaccess: Target address not found on any valid interface")
 	return nil, fmt.Errorf("address not found on any interface")
 }


bigtable/client.go
```diff
@@ -233,6 +235,13 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
 		}
 
 	} else {
+		enableDirectAccess, _ := strconv.ParseBool(os.Getenv("CBT_ENABLE_DIRECTPATH"))
+		if enableDirectAccess {
+			o = append(o, directPathOptions...)
+			if disableBoundToken, _ := strconv.ParseBool(os.Getenv("CBT_DISABLE_DIRECTPATH_BOUND_TOKEN")); !disableBoundToken {
+				o = append(o, internaloption.AllowHardBoundTokens("ALTS"))
+			}
+		}
```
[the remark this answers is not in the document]
sushanb: Wrong finding.

bigtable/internal/transport/connpool.go
```diff
@@ -447,13 +448,24 @@ func NewBigtableChannelPool(ctx context.Context, connPoolSize int, strategy btop
 	var firstConn *BigtableConn
 
 	if pool.directAccessDialer != nil {
-		var isDirectAccess bool
-		if firstConn, isDirectAccess = pool.checkIfDirectAccessCompatible(); isDirectAccess {
+		directAccessConn, isDirectAccess := pool.checkIfDirectAccessCompatible()
+		disableDirectAccess := os.Getenv("CBT_ENABLE_DIRECTPATH") == "false"
```
nimf, line 452: If user disabled directpath we don't need to check if direct acccess compatible. I.e., check if directpath disabled first, if it is disabled, then directAccessConn is null and isDirectAccess is false.

bigtable/client.go
```diff
@@ -203,7 +206,7 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
 			// options
 			btransport.WithInstanceName(fullInstanceName),
 			btransport.WithAppProfile(config.AppProfile),
-			btransport.WithFeatureFlagsMetadata(ffMD),
+			btransport.WithFeatureFlagsMetadata(directAccessMD),
```
nimf, line 209: why do we set both featureFlagsMetadata and directAccessFeatureFlagsMetadata to directAccessMD? We don't know if we will use directpath or not yet, because we check CBT_ENABLE_DIRECTPATH in the bigtable channel pool later.

bigtable/internal/transport/connpool.go
sushanb: Even if user disabled, we still want to check and export the metrics.

bigtable/client.go
sushanb: i am making featureflags agnostic of transport. we will send the same feature flag now for both directpath and cloudpath.

sushanb:
> Sorry, I'm confused. Why if we use bigtable channel pool we enable direct access by default and check if CBT_ENABLE_DIRECTPATH is false, while when we don't use bigtable channel pool we enable direct access only when CBT_ENABLE_DIRECTPATH is true?

We are getting rid of legacy channel pool and will use the bigtable channel pool. that is the reason.

bigtable/internal/transport/connpool.go
nimf: Let's do it in a separate goroutine then, not to delay channel pool creation.

bigtable/client.go
nimf: hmmm... But this means we will send `DirectAccessRequested` as true even if the user disables directpath?

sushanb: Yeah. I kept as it for now. I agree that it is not nice to send the featureflags if user explicitly disables direct access. I will do as a followup.

bigtable/internal/transport/connpool.go
sushanb: i think your idea is great. skipped the check if If user disabled directpath

bigtable/internal/directaccess/gcp.go
```diff
@@ -0,0 +1,31 @@
...
+	"log"
+
+	"cloud.google.com/go/compute/metadata"
+)
+
+// IsRunningOnGCP checks that the code is running on GCP by checking the metadata server.
+// Basically, reads /sys/class/dmi/id/product_name
+func IsRunningOnGCP() error {
+	if metadata.OnGCE() {
+		log.Println("Detected running on GCE via metadata.OnGCE()")
```
nimf: Wait, do we want this logged? Was it just for debugging here?

bigtable/internal/directaccess/metadata_server.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode == 200 {
+		address := net.ParseIP(strings.TrimSuffix(string(body), "\n"))
```
nimf:
```suggestion
		address := net.ParseIP(strings.TrimSpace(string(body)))
```

bigtable/internal/transport/connpool.go
```diff
@@ -1102,6 +1125,77 @@ func (cf *connectionFactory) primeWithRetry(ctx context.Context, conn *BigtableC
...
+		return
+	}
+	if err := directaccess.CheckLocalIPv4LoopbackAddress(); err != nil {
+		btopt.Debugf(p.logger, "bigtable_connpool: Direct Access investigation: IPv4 loopback missing: %v", err)
+		p.reportDirectAccessFailure("loopback_misconfigured")
+		return
+	}
+	if err := directaccess.CheckLocalIPv6LoopbackAddress(); err != nil {
+		btopt.Debugf(p.logger, "bigtable_connpool: Direct Access investigation: IPv6 loopback missing: %v", err)
+		p.reportDirectAccessFailure("loopback_misconfigured")
```
nimf: Will IPv4 only machine also have an IPv6 loopback?

bigtable/internal/transport/connpool.go
```diff
@@ -1102,6 +1125,77 @@ func (cf *connectionFactory) primeWithRetry(ctx context.Context, conn *BigtableC
...
+
+	// If the metadata server assigned IPs, but the guest OS hasn't configured any of them on an interface
+	if !v4Plumbed && !v6Plumbed {
+		btopt.Debugf(p.logger, "bigtable_connpool: Direct Access investigation: No metadata IPs are plumbed to local interfaces. Original error: %v", originalErr)
+		p.reportDirectAccessFailure("nic_misconfigured")
+		return
+	}
+
+	btopt.Debugf(p.logger, "bigtable_connpool: Direct Access investigation: Running on GCP, metadata reachable, IPs assigned and plumbed, but Direct Access still failed. Original error: %v", originalErr)
+	p.reportDirectAccessFailure("")
```
nimf:
```suggestion
	p.reportDirectAccessFailure("unknown")
```

bigtable/client.go
```diff
@@ -180,9 +178,14 @@ func NewClientWithConfig(ctx context.Context, project, instance string, config C
 
 		fullInstanceName := fmt.Sprintf("projects/%s/instances/%s", project, instance)
 
+		directAccessDialerOptions := make([]option.ClientOption, len(o))
+		copy(directAccessDialerOptions, o)
+		directAccessDialerOptions = append(directAccessDialerOptions, directPathOptions...)
+		// enable hard bound tokens by default
```
nimf, line 184: why we don't need to check for `CBT_DISABLE_DIRECTPATH_BOUND_TOKEN` for BigtableConnPool but need to check it otherwise on line 242 below?

sushanb: good question. @rockspore can we now enable the direct bound token by default?

bigtable/internal/transport/connpool.go
sushanb: made it per protocol.

bigtable/internal/directaccess/gcp.go
sushanb: Oops local testing bled through. sorry

PR #14166, opened 2026-03-12
fix(storage): decouple Recv loop in gRPCOneshotBidiWriteBufferSender …

cpriti-os:
…to prevent deadlocks

Noticed a deadlock when OneShot sender was bombarded with a lot of data during a PCU benchmarking run

### DIFF
--- a/storage/grpc_writer.go
+++ b/storage/grpc_writer.go
@@ -240,7 +240,7 @@ func (c *grpcStorageClient) OpenWriter(params *openWriterParams, opts ...storage
 	return w, nil
 }
 
-// gRPCWriter is a wrapper around the the gRPC client-stream API that manages
+// gRPCWriter is a wrapper around the gRPC client-stream API that manages
 // sending chunks of data provided by the user over the stream.
 type gRPCWriter struct {
 	preRunCtx context.Context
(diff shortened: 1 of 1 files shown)


storage/grpc_writer.go
```diff
@@ -1019,59 +1019,101 @@ func (s *gRPCOneshotBidiWriteBufferSender) connect(ctx context.Context, cs gRPCB
...
+		<-recvDone
+		// Prefer recvErr since that's where RPC errors are delivered
+		if recvErr != nil {
+			s.streamErr = recvErr
+		} else if sendErr != nil {
+			s.streamErr = sendErr
+		}
+		if s.streamErr == io.EOF {
+			s.streamErr = nil
 		}
```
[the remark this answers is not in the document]
cpriti-os: Fixed

storage/grpc_writer.go
```diff
@@ -1019,59 +1019,99 @@ func (s *gRPCOneshotBidiWriteBufferSender) connect(ctx context.Context, cs gRPCB
...
+						if firstSend {
+							proto.Merge(req, s.firstMessage)
+							firstSend = false
+						}
 
-			if err := stream.Send(req); err != nil {
-				_, s.streamErr = drainInboundStream(stream)
-				if err != io.EOF {
-					s.streamErr = err
+						if err := stream.Send(req); err != nil {
```
krishnamd-jkp, line 1047: This is still a blocking call, no?

storage/grpc_writer.go
```diff
@@ -1019,59 +1019,99 @@ func (s *gRPCOneshotBidiWriteBufferSender) connect(ctx context.Context, cs gRPCB
...
-				return
-			}
+			}()
+			close(sendDone)
+		}()
 
-			if r.finishWrite {
-				stream.CloseSend()
-				// Oneshot uploads only read from the response stream on completion or
-				// failure
```
krishnamd-jkp, line 1059: Check if this is required by design.

storage/grpc_writer.go
cpriti-os: Yes, but its isolated in its own goroutine

storage/grpc_writer.go
cpriti-os: This is prone to deadlocks as observed, it was the design earlier probably because we didn't use this writer for any concurrent and heavy workloads which can cause the send to block. So, no this should not be required by design, decoupling is a standard for grpc streaming.

PR #14173, opened 2026-03-13
chore: experimental release of area120

jskeet:
<details><summary>area120: v0.10.0</summary></details>


quartzmo:
The description above has an empty `details` HTML tag, is that intentional?

jskeet:
> The description above has an empty details HTML tag, is that intentional?

Yes. That's the format that legacylibrarian expects, except there would normally be release notes - we don't have any, but I've confirmed that to be non-fatal. (At least as far as I could test.) It's one of the things we're testing here.

PR #14182, opened 2026-03-16
feat(bigquery): add http tracing telemetry data to Dataset and Model operations

westarle:
This PR introduces the foundational OpenTelemetry T4 (network) tracing instrumentation for the BigQuery Go client library, focusing initially on Dataset and Model operations. This work
   aligns the BigQuery veneer with the Client Libraries Observability (CLO) v1 Product Requirements for network spans.

   **Changes Included:**
   * **Static Identity Attributes:** Updated `NewClient` to inject `internaloption.WithTelemetryAttributes` into the client configuration.
   * **Typed Trace Helpers:** Created a centralized set of strongly-typed helper functions (e.g., `setDatasetTraceMetadata`, `setModelTraceMetadata`) in a new `trace.go` file.
   * **Dataset & Model Instrumentation:** Updated network-bound operations in `dataset.go` (e.g., `Create`, `Metadata`, `Update`, `Delete`, `list*`) and `model.go` to invoke the typed trace
   helpers prior to executing the backend REST API calls.
   * **Testing:** Added comprehensive table-driven tests (`TestTracingTelemetryAttributes` and `TestTracingTelemetryAttributes_Retry`) to verify the correct injection and presence of static
   attributes, `resource_name`, `url_template`, and `resend_count` on the emitted spans.


bigquery/bigquery.go
```diff
@@ -83,6 +84,18 @@ func NewClient(ctx context.Context, projectID string, opts ...option.ClientOptio
 		option.WithScopes(Scope),
 		option.WithUserAgent(fmt.Sprintf("%s/%s", userAgentPrefix, internal.Version)),
 	}
+
+	if gax.IsFeatureEnabled("TRACING") {
+		o = append(o, internaloption.WithTelemetryAttributes(map[string]string{
```
quartzmo, line 89: Will these attributes reach the transport wrappers in cloud.google.com/go/auth? I believe our "new" auth package is only used when explicitly enabled via `internaloption.EnableNewAuthLibrary()`. In the default/legacy path, google-api-go-client [transport logic](https://github.com/googleapis/google-api-go-client/blob/main/transport/http/dial.go#L312) will not use these attributes.

@shollyman Can you confirm that the client in `bigquery/bigquery.go` still does not use the new auth package? I don't see the `EnableNewAuthLibrary` function call. Here it is in a bigquery gapic for example: https://github.com/westarle/google-cloud-go/blob/main/bigquery/storage/apiv1/big_query_read_client.go#L54

shollyman: for the core bigquery service, EnableNewAuthLibrary is deeper, down in the generated discovery: https://github.com/googleapis/google-api-go-client/blob/e7df9fe0b92461f87b6d267a600e6825d1221e75/bigquery/v2/bigquery-gen.go#L151

westarle: Just in case, I also put some tests that make sure expected attributes are being set on spans.

PR #14190, opened 2026-03-17
feat(auth): implement actionable errors logging in grpc transport

westarle:
This PR introduces "Actionable Errors" logging to the gRPC transport wrappers (`otelHandler`), providing out-of-the-box structured logging for failed network attempts. These logs give visibility into client-side errors (like network partitions or permission drops) while allowing correlation with backend traces.

Refactoring
* Extracted the formatting and emission logic into a dedicated `logActionableError` helper function.
* Split the monolithic `HandleRPC` block into `handleRPCError` and `handleRPCSuccess` for improved readability.
* Added unit tests in `grpctransport_otel_test.go` using a memory-backed JSON `slog.Handler`

### DIFF
--- a/auth/grpctransport/grpctransport_otel_test.go
+++ b/auth/grpctransport/grpctransport_otel_test.go
@@ -621,6 +621,8 @@ func TestDial_OpenTelemetry_Disabled(t *testing.T) {
 }
 
 func TestHandleRPC_ActionableErrors(t *testing.T) {
+	// Do not add t.Parallel() to these tests. The global resetting will cause
+	// flaky tests because they will stomp over each other's feature flag state.
 	gax.TestOnlyResetIsFeatureEnabled()
 	defer gax.TestOnlyResetIsFeatureEnabled()
 	t.Setenv("GOOGLE_SDK_GO_EXPERIMENTAL_LOGGING", "true")


auth/grpctransport/grpctransport.go
```diff
@@ -515,53 +531,134 @@ func (h *otelHandler) HandleRPC(ctx context.Context, s stats.RPCStats) {
...
+		attrs := []attribute.KeyValue{
 			attribute.String("rpc.response.status_code", "OK"),
 		}
+		span.SetAttributes(attrs...)
 	}
-	span.SetAttributes(attrs...)
-	h.Handler.HandleRPC(ctx, s)
+}
+
+func logActionableError(ctx context.Context, logger *slog.Logger, st *status.Status, errorType string, rpcStatusCode string, staticAttrs []slog.Attr, err error, apiErr *apierror.APIError, isStatusOk bool) {
```
quartzmo: 9 parameters is a lot. I think this func probably only needs ctx, logger, err, and apiErr. Can you try this:

1. Bind `staticAttrs` to the logger at initialization time using `.With(...)` instead of passing it dynamically.
2. Inside `logActionableError`, pull the gRPC `status.Status` directly from the original `err `using `status.FromError(err)`.
3. Drop redundant arguments st, rpcStatusCode, isStatusOk, errorType, and staticAttrs from the signature.

quartzmo: I think we could possibly move `logActionableError` to gax-go/v2/internallog for reuse across both grpc and rest, if that fits with your plans. We would need to parameterize the transport system type (passing 'grpc' or 'http') so it's not locked into gRPC.

auth/grpctransport/grpctransport_otel_test.go
```diff
@@ -611,3 +619,376 @@ func TestDial_OpenTelemetry_Disabled(t *testing.T) {
 		})
 	}
 }
+
+func TestHandleRPC_ActionableErrors(t *testing.T) {
+	gax.TestOnlyResetIsFeatureEnabled()
```
quartzmo, line 626: Might be good to add a comment here not to add t.Parallel() to these tests in the future, as the global resetting will cause flaky tests because they will stomp over each other's feature flag state.

PR #14193, opened 2026-03-17
feat(spanner): Switch to using builtin open telemetry for EEF

kinsaurralde:
Switching to built in open telemetry since the eef metrics are internal metrics


rahul2393:
@kinsaurralde please change the first commit message

Issue #14212, opened 2026-03-19
spanner: read deadlock on session initialization


laurynaslubys:
## Client

Spanner

## Description

A service using the spanner library version v1.87.0 failed to start up. After investigating the goroutine dump we discovered that one goroutine performing a readonly transaction was stuck inside the spanner library for multiple hours at:
```
goroutine 7138 [chan send, 1058 minutes]:
cloud.google.com/go/spanner.(*sessionPool).takeMultiplexed(0x4001be6600, {0x4a77190, 0x40098a5650})
        cloud.google.com/go/spanner@v1.87.0/session.go:1309 +0x10c
cloud.google.com/go/spanner.(*ReadOnlyTransaction).acquireSingleUse(0x4001c42680, {0x4a77190, 0x40098a5650})
        cloud.google.com/go/spanner@v1.87.0/transaction.go:1010 +0x278
cloud.google.com/go/spanner.(*ReadOnlyTransaction).acquire(0x4001c42680, {0x4a77190, 0x40098a5650})
        cloud.google.com/go/spanner@v1.87.0/transaction.go:987 +0x4c
cloud.google.com/go/spanner.(*txReadOnly).prepareExecuteSQL(0x4001c42688, {0x4a77190?, 0x40098a5650?}, {{0x400af2f4a0, 0x1d8}, 0x40098a55c0}, {0x0, 0x400054f950, 0x3, {0x4002396090, ...}, ...})
        cloud.google.com/go/spanner@v1.87.0/transaction.go:750 +0x58
cloud.google.com/go/spanner.(*txReadOnly).query(0x4001c42688, {0x4a77190?, 0x40098a5590?}, {{0x400af2f4a0?, 0x2c?}, 0x40098a55c0?}, {0x0, 0x400054f950, 0x3, {0x4002396090, ...}, ...})
        cloud.google.com/go/spanner@v1.87.0/transaction.go:689 +0xf8
cloud.google.com/go/spanner.(*txReadOnly).QueryWithOptions(0x4001c42688, {0x4a77190, 0x40098a5590}, {{0x400af2f4a0?, 0x43c87483d?}, 0x40098a55c0?}, {0x0, 0x0, 0x3, {0x4002396090, ...}, ...})
        cloud.google.com/go/spanner@v1.87.0/transaction.go:645 +0x134
```

[...]

PR #14215, opened 2026-03-19
fix(spanner): replace multiplexed session request loop with shared in-flight creation

rahul2393:
Fixes: https://github.com/googleapis/google-cloud-go/issues/14212

Reworked multiplexed session initialization to use a shared in-flight creation latch instead of the old request
  goroutine and unbuffered channels. This fixes the startup deadlock triggered by a cancelled read-only transaction
  waiter and adds regression coverage for cancellation and client shutdown while session creation is in flight.
  
   sessionManager used an unbuffered multiplexedSessionReq channel and a single goroutine to coordinate multiplexed session creation. A cancelled waiter could cause that goroutine to return, leaving future callers blocked forever trying to send on the request channel during takeMultiplexed.

  Replace that request/ack flow with a shared in-flight creation record on sessionManager. When no multiplexed session exists, the manager starts one background creation with a manager-owned context and exposes a done channel for all waiters in that creation epoch. Waiters now block on done or their own context, and completion is broadcast by closing the shared channel. This removes the deadlock where one cancelled context could kill shared coordination, and it also lets Client.Close release any waiters blocked behind an in-flight initial create.

### DIFF
--- a/spanner/session.go
+++ b/spanner/session.go
@@ -335,7 +335,7 @@ func (p *sessionManager) ensureMultiplexedSessionCreationLocked(force bool) *mul
 	if !force && p.multiplexedSession != nil {
 		return nil
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	ctx, cancel := context.WithCancel(context.Background())
 	creation := &multiplexedSessionCreation{
 		done:   make(chan struct{}),
 		cancel: cancel,


spanner/session_test.go
```diff
@@ -112,6 +113,111 @@ func TestMultiplexSessionWorker(t *testing.T) {
...
+		return nil
+	})
+
+	done := make(chan error, 1)
+	go func() {
+		readCtx, readCancel := context.WithTimeout(ctx, 15*time.Second)
+		defer readCancel()
+		_, err := client.Single().ReadRow(readCtx, "Albums", Key{"foo"}, []string{"SingerId", "AlbumId", "AlbumTitle"})
+		done <- err
+	}()
```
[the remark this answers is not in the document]
rahul2393: Multiplexed session creation starts during client construction, not on the first ReadRow. We call ensureMultiplexedSessionCreationLocked(true), which sets client.sm.multiplexedSessionCreation and launches the background create.

So in TestMultiplexedSessionCreationCloseReleasesWaiters, this block is correct:
  - create client
  - waitFor until client.sm.multiplexedSessionCreation != nil
  - start the ReadRow goroutine
  - client.Close()

  The waitFor is not waiting for ReadRow to trigger creation. It is only making sure the startup creation is already in flight before the waiter goroutine is started.

  If anything, starting ReadRow first would make the test less precise, because the thing being validated is “a waiter blocked behind the startup create is released by Close(),” and startup creation is already initiated by NewClientWithConfig, not by ReadRow.

spanner/session.go
```diff
@@ -342,73 +328,72 @@ func (p *sessionManager) recordOTStat(ctx context.Context, m metric.Int64Counter
...
-			return
-		}
+func (p *sessionManager) ensureMultiplexedSessionCreationLocked(force bool) *multiplexedSessionCreation {
+	if p.multiplexedSessionCreation != nil {
+		return p.multiplexedSessionCreation
+	}
+	if !force && p.multiplexedSession != nil {
+		return nil
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
olavloite: This will effectively make 30 seconds the upper bound for a timeout for the `CreateSession` RPC. That is: If someone configures `CreateSession` to have a 60 second timeout, then that will be ignored, and this 30s value will effectively be used instead. Is that intentional?

olavloite (approved):
LGTM, with an open question regarding the upper bound for the timeout for `CreateSession`. I'm fine with leaving this as it is for now, as `CreateSession` should never take more than 30s. But I do think that we should at least do a follow-up to fix it, as it is not logical.

spanner/session.go
rahul2393: Thanks for pointing this out, this got copied from the old code which used 30s for the startup/refresh force create of mux session, the new code used 30s for every shared multiplexed-session creation epoch.

Removed it now,
 - the manager still owns cancellation for shutdown
  - CreateSession timeout policy now comes only from the configured GAPIC call options
  - default behavior remains 30s(default gax), but a user-configured longer CreateSession timeout is no longer capped at 30s by the manager layer.

PR #14232, opened 2026-03-23
feat(storage): integrate parallel composite upload (PCU) into storage writer

cpriti-os:


storage/pcu.go
```diff
@@ -141,23 +146,37 @@ type uploadResult struct {
 	obj        *ObjectAttrs
 	handle     *ObjectHandle
 	err        error
+	size       int64
 }
 
 func (w *Writer) initPCU(ctx context.Context) error {
```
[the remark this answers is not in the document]
cpriti-os: PCU is only used for internal identifiers

storage/pcu.go
```diff
@@ -405,8 +435,10 @@ func (s *pcuState) close() error {
 		close(s.resultCh)
 		s.collectorWG.Wait()
 
-		// Cleanup is always attempted.
-		defer s.doCleanupFn(s)
+		// Cleanup is always attempted. We do it in the background to not block returning.
+		defer func() {
+			go s.doCleanupFn(s)
+		}()
```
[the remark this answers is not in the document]
cpriti-os: cleanup function uses WithoutCancel which will ensure cleanup is attempted

storage/writer.go
```diff
@@ -197,31 +254,45 @@ type Writer struct {
...
-			return n, werr
+
+	if pcu == nil && !w.opened {
+		// First time initialization: freeze the configuration to either PCU or standard.
+		if w.EnableParallelUpload {
+			if !w.isGRPCClient() {
+				// PCU only supported for gRPC.
+				// Nullify the config and proceed with standard upload.
+				// Log prominent warning.
+				log.Println("storage: ParallelUploadConfig is ignored because Parallel Uploads are only supported for gRPC clients. Proceeding with standard upload.")
```
krishnamd-jkp: Why don't we throw an error instead of doing something different than the user asked?

krishnamd-jkp: Also should we postpone logging in a different initiative?

storage/pcu.go
```diff
@@ -168,7 +187,7 @@ func (w *Writer) initPCU(ctx context.Context) error {
 		config:          cfg,
 		settings:        s,
 		bufferCh:        make(chan []byte, s.bufferPoolSize),
-		uploadCh:        make(chan uploadTask, cfg.maxConcurrency), // Buffered to prevent worker starvation
+		uploadCh:        make(chan uploadTask, cfg.MaxConcurrency), // Buffered to prevent worker starvation
```
krishnamd-jkp: nit: missing period. Here and other places too.

storage/pcu.go
```diff
@@ -329,6 +346,19 @@ func (s *pcuState) write(p []byte) (int, error) {
 				return total - len(p), s.ctx.Err()
 			case s.currentBuffer = <-s.bufferCh:
 				s.bytesBuffered = 0
+			default:
+				if s.buffersAlloc < s.settings.bufferPoolSize {
+					s.currentBuffer = make([]byte, s.config.PartSize)
```
krishnamd-jkp, line 352: Just confirming. This should be safe because write is synchronous?

storage/pcu.go
```diff
@@ -521,6 +553,12 @@ func (s *pcuState) composeParts() error {
 	if err != nil {
 		return err
 	}
+
+	// Perform client-side CRC32C validation if a user-provided checksum was specified.
+	if s.w.SendCRC32C && s.w.CRC32C != 0 && attrs.CRC32C != s.w.CRC32C {
+		return fmt.Errorf("storage: object was uploaded, but its CRC32C (%d) does not match the expected CRC32C (%d)", attrs.CRC32C, s.w.CRC32C)
+	}
+
```
krishnamd-jkp, line 562: Is this still required if you're doing compose crc32c validation already?

storage/writer.go
```diff
@@ -197,31 +254,45 @@ type Writer struct {
...
+			if err := w.openWriter(); err != nil {
+				return 0, err
+			}
 		}
 	}
-	return n, err
+
+	if pcu != nil {
+		return w.wrapWriteError(pcu.write(p))
+	}
```
krishnamd-jkp: nit: You can preempt this condition to avoid `if pcu == nil && !w.opened {` in line 269 and call `pcu.write(p)` after init. Just to simplify.

storage/writer.go
```diff
@@ -271,6 +342,36 @@ func (w *Writer) Flush() (int64, error) {
...
+	closed, werr, pcu := w.closed, w.err, w.pcu
+	w.mu.Unlock()
+
+	if closed {
+		return werr
+	}
+
+	if pcu != nil || (!w.opened && w.EnableParallelUpload) {
+		if w.EnableParallelUpload && !w.isGRPCClient() && pcu == nil {
+			log.Println("storage: ParallelUploadConfig is ignored because Parallel Uploads are only supported for gRPC clients. Proceeding with standard upload.")
```
krishnamd-jkp: Is this log right here?

krishnamd-jkp (commented):
initial review

storage/writer.go
cpriti-os: One reason which I also highlighted in the design doc is it helps user use the same code for different transports.

cpriti-os: Here we need to log the warning, we can move this later to use the same package/format in the logging initiative.

storage/pcu.go
cpriti-os: apparently compose does not support user provided checksumming, found this issue while running the integration test

storage/writer.go
cpriti-os: Done

storage/writer.go
cpriti-os: removed, logging anything doesn't add any value here

storage/pcu.go
cpriti-os: Fixed

storage/pcu.go
cpriti-os: yes

storage/integration_test.go
```diff
@@ -8716,3 +8720,246 @@ func setUpRequesterPaysBucket(ctx context.Context, t *testing.T, bucket, object
...
+				// Verify object size and existence.
+				attrs, err := obj.Attrs(ctx)
+				if err != nil {
+					t.Fatalf("obj.Attrs: %v", err)
+				}
+				if attrs.Size != tc.expected {
+					t.Errorf("Object size mismatch: got %d, want %d", attrs.Size, tc.expected)
+				}
+
+				// Verify cleanup of intermediate/part objects.
```
krishnamd-jkp, line 8807: You can move this to the end of the test which gives it more time while we check the contents of the object.

storage/integration_test.go
```diff
@@ -8716,3 +8720,246 @@ func setUpRequesterPaysBucket(ctx context.Context, t *testing.T, bucket, object
...
+					} else {
+						t.Errorf("content mismatch: got %d bytes, want %d bytes", len(gotContent), len(tc.content))
+					}
+				}
+			})
+		}
+	})
+}
+
+func TestIntegration_ParallelUploadConcurrency(t *testing.T) {
```
krishnamd-jkp, line 8844: What exactly does this test? Maybe add a small description

storage/integration_test.go
```diff
@@ -8716,3 +8720,246 @@ func setUpRequesterPaysBucket(ctx context.Context, t *testing.T, bucket, object
...
+	multiTransportTest(ctx, t, func(t *testing.T, ctx context.Context, bucket string, _ string, client *Client) {
+		h := testHelper{t}
+
+		content := bytes.Repeat([]byte("z"), 12<<20) // 12 MiB
+		correctCRC := crc32.Checksum(content, crc32cTable)
+
+		testCases := []struct {
+			name        string
+			crc32c      uint32
+			expectError bool
```
krishnamd-jkp: nit: Make this string (pass expected error message) and compare it directly against this instead of hardcoding it in the test below.

storage/integration_test.go
cpriti-os: Done

storage/integration_test.go
cpriti-os: added

storage/writer.go
```diff
@@ -183,6 +183,10 @@ type Writer struct {
 	// parallel. Supported exclusively for gRPC clients. If used with a JSON
 	// client, the configuration is ignored and a standard upload is performed.
 	//
+	// Upon completion of a parallel upload, the Writer makes a best-effort attempt to clean up any temporary parts created.
+	// It is recommended to set appropriate bucket lifecycle policies to reliably clean up any leaked objects to avoid unnecessary storage costs.
```
krishnamd-jkp: nit: Leftover would be better instead of leak

cpriti-os: Done

PR #14250, opened 2026-03-25
fix(firestore): rename to match Java and expose functions on expression

bhshkh:
1. Rename TruncWithPlaces (Java's implementaion [here](https://github.com/googleapis/java-firestore/blob/f9439c92f31c347ed31e42378177630468d5ed29/google-cloud-firestore/src/main/java/com/google/cloud/firestore/pipeline/expressions/Expression.java#L3775-L3787)), LtrimWithValues, etc to match Java
2. Change signature of ArrayIndexOf and introduce ArrayLastIndexOf to match Java
3. Earlier the methods such LogicalMaximum, CountDistinct, etc couldn't be chained on expression and were only defined as static methods. E.g. users couldn't do `FieldOf("rating").LogicalMaximum(4.5)` but could do `LogicalMaximum(FieldOf("published"), 1900)`

### DIFF
--- a/firestore/pipeline_expression.go
+++ b/firestore/pipeline_expression.go
@@ -326,8 +326,6 @@ type Expression interface {
 	IsError() BooleanExpression
 	// FieldExists returns a boolean expression that checks if the field exists.
 	FieldExists() BooleanExpression
-	// Exists is an alias for FieldExists.
-	Exists() BooleanExpression
 	// IsAbsent returns a boolean expression that checks if the field is absent.
 	IsAbsent() BooleanExpression
 	// IfAbsent creates an expression that returns a default value if an expression evaluates to an absent value.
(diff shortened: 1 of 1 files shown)


firestore/pipeline_expression.go
```diff
@@ -315,6 +322,14 @@ type Expression interface {
 	//
 	// The parameter 'catchExprOrValue' is the expression or value to return if the receiver expression errors.
 	IfError(catchExprOrValue any) Expression
+	// IsError returns a boolean expression that checks if the expression evaluates to an error.
+	IsError() BooleanExpression
+	// FieldExists returns a boolean expression that checks if the field exists.
+	FieldExists() BooleanExpression
+	// Exists is an alias for FieldExists.
```
shollyman: This feels like we're inheriting and propagating a naming mistake from somewhere else?  Can we identify which name is preferred and only expose that?

bhshkh: removed exists

PR #14252, opened 2026-03-25
chore(spanner): add routing hints and cache updates for begin/commit

rahul2393:
## Summary

  This change extends key-aware routing to include mutation-based `BeginTransaction` and `Commit` request payloads, and teaches the response path to consume `CacheUpdate` from `Transaction` and `CommitResponse`.

  ## What changed

  - Populate `RoutingHint` on `BeginTransaction` when a `mutationKey` is present
  - Populate `RoutingHint` on `Commit` from the first mutation in the request
  - Keep existing transaction-affinity routing for `Commit`/`Rollback`
  - Apply `CacheUpdate` from:
    - `Transaction`
    - `CommitResponse`
  - Add focused `KeyAwareChannel` tests for the new request/response behavior

### DIFF
--- a/spanner/client.go
+++ b/spanner/client.go
@@ -643,6 +643,9 @@ func newClientWithConfig(ctx context.Context, database string, config ClientConf
 	var locationRouter *locationRouter
 	if isExperimentalLocationAPIEnabled() {
 		sc.baseClientOpts = endpointClientOpts
+		if conn := pool.Conn(); conn != nil {
+			sc.endpointAuthority = normalizeAuthorityTarget(conn.Target())
+		}
 		epCache := newEndpointClientCache(sc.createEndpointClient)
 		locationRouter = newLocationRouter(epCache)
 	}
@@ -692,6 +695,16 @@ Multiplexed session enabled: true
 		locationRouter:       locationRouter,
 	}
 	return c, nil
+}
+
+func normalizeAuthorityTarget(target string) string {
+	if idx := strings.Index(target, ":///"); idx >= 0 {
+		return strings.TrimSuffix(target[idx+4:], "/")
+	}
+	if idx := strings.Index(target, "://"); idx >= 0 {
+		return strings.TrimSuffix(target[idx+3:], "/")
+	}
+	return strings.TrimSuffix(target, "/")
 }
 
 // NewMultiEndpointClient is the same as NewMultiEndpointClientWithConfig with
--- a/spanner/endpoint_authority_test.go
+++ b/spanner/endpoint_authority_test.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2026 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package spanner
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"cloud.google.com/go/spanner/apiv1/spannerpb"
+	"cloud.google.com/go/spanner/internal/testutil"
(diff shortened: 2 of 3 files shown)


spanner/key_recipe_cache.go
```diff
@@ -253,6 +253,28 @@ func (c *keyRecipeCache) mutationToTargetRange(mutation *sppb.Mutation) *targetR
 	return recipe.mutationToTargetRange(mutation)
 }
 
+func (c *keyRecipeCache) computeMutationKeys(hint *sppb.RoutingHint, mutation *sppb.Mutation) bool {
+	if hint == nil || mutation == nil {
+		return false
+	}
+	c.mu.Lock()
+	if len(c.schemaGeneration) > 0 {
+		hint.SchemaGeneration = append([]byte(nil), c.schemaGeneration...)
```
olavloite, line 254: Here also we are updating the hint while the function might return false. Can we document exactly what this method does (or is intended to do), and when it is supposed to return true/false?

spanner/channel_finder.go
```diff
@@ -100,17 +100,24 @@ func (f *channelFinder) findServerBeginTransaction(ctx context.Context, req *spp
...
 
+func (f *channelFinder) fillCommitRoutingHint(ctx context.Context, req *sppb.CommitRequest) {
+	if req == nil || len(req.GetMutations()) == 0 {
+		return
+	}
+	hint := ensureCommitRoutingHint(req)
+	if !f.recipeCache.computeMutationKeys(hint, req.GetMutations()[0]) {
+		return
+	}
+	_ = f.fillRoutingHint(ctx, true, rangeModeCoveringSplit, &sppb.DirectedReadOptions{}, hint)
```
olavloite: Why are we ignoring the return value here? If this is intentional, can we add a comment that explains it?

spanner/key_recipe_cache.go
```diff
@@ -253,6 +253,28 @@ func (c *keyRecipeCache) mutationToTargetRange(mutation *sppb.Mutation) *targetR
 	return recipe.mutationToTargetRange(mutation)
 }
 
+func (c *keyRecipeCache) computeMutationKeys(hint *sppb.RoutingHint, mutation *sppb.Mutation) bool {
```
olavloite: Same as in Java: This is a hard-to-read function, as it both returns a value and has a potential side-effect of updating the routing hint.

spanner/channel_finder.go
```diff
@@ -100,17 +100,24 @@ func (f *channelFinder) findServerBeginTransaction(ctx context.Context, req *spp
...
-	}
 	return f.fillRoutingHint(ctx, preferLeaderFromTransactionOptions(req.GetOptions()), rangeModeCoveringSplit, &sppb.DirectedReadOptions{}, hint)
 }
 
+func (f *channelFinder) fillCommitRoutingHint(ctx context.Context, req *sppb.CommitRequest) {
+	if req == nil || len(req.GetMutations()) == 0 {
+		return
+	}
+	hint := ensureCommitRoutingHint(req)
+	if !f.recipeCache.computeMutationKeys(hint, req.GetMutations()[0]) {
```
olavloite: Why exactly are using only the first mutation here?

spanner/location_router_test.go
```diff
@@ -138,8 +138,14 @@ func TestLocationRouter_NilSafety(t *testing.T) {
 	router.prepareExecuteSQLRequest(context.Background(), &sppb.ExecuteSqlRequest{})
 	router.prepareBeginTransactionRequest(context.Background(), nil)
 	router.prepareBeginTransactionRequest(context.Background(), &sppb.BeginTransactionRequest{})
+	router.prepareCommitRequest(context.Background(), nil)
```
olavloite, line 156: Can we add a couple of test cases for:
1. A commit with a mutation that is currently not in the cache
2. A single-use read/write transaction

spanner/channel_finder.go
rahul2393: Updated to use the logic used for selecting mutaiton Key for BeginTransaction RPC

spanner/channel_finder.go
rahul2393: I updated the commit path so the mutation-routing helper now returns the selected endpoint, and the commit RPC uses that endpoint unless transaction affinity overrides it. So we no longer ignore the routing result in this path.

PR #14264, opened 2026-03-26
feat(auth): enrich invoke-level telemetry with transport detail

westarle:
This PR has auth code fill in TelemetryTransportData in order to send details from the transport back up to higher layers.  See https://github.com/googleapis/gax-go/pull/496. This in turn allows code in those higher layers to report telemetry information for the last request that is only known by the transport layer.

### Changes Made
* **Logical Hostname Extraction:** Updated extraction logic to parse logical hostnames and ports directly from `req.URL` (HTTP) and `cc.Target()` (gRPC) prior to invoking the transport
layer.
* **Implicit Port Fallback:** Added scheme-based port fallbacks (`80` for HTTP, `443` for HTTPS) when handling URLs with implicit ports.

### DIFF
--- a/auth/grpctransport/grpctransport.go
+++ b/auth/grpctransport/grpctransport.go
@@ -495,8 +495,10 @@ func addOpenTelemetryStatsHandler(dialOpts []grpc.DialOption, opts *Options, end
 func extractHostPort(target string) (string, int) {
 	if idx := strings.Index(target, "://"); idx != -1 {
 		target = target[idx+3:]
-		// Ensure any leading slashes from the scheme suffix are stripped
-		target = strings.TrimLeft(target, "/")
+		// Ensure any leading authorities (like 8.8.8.8 in dns://8.8.8.8/foo) are stripped
+		if slashIdx := strings.Index(target, "/"); slashIdx != -1 {
+			target = target[slashIdx+1:]
+		}
 	}
 	host, portStr, err := net.SplitHostPort(target)
 	if err != nil {
--- a/auth/grpctransport/grpctransport_otel_test.go
+++ b/auth/grpctransport/grpctransport_otel_test.go
@@ -1066,6 +1066,7 @@ func TestExtractHostPort(t *testing.T) {
 		{"xds:///my-service:80", "my-service", 80},
 		{"dns:///[::1]:8080", "::1", 8080},
 		{"google.com:foo", "google.com", 0},
+		{"dns://8.8.8.8/lb.example.com:443", "lb.example.com", 443},
 	}
 
 	for _, tt := range tests {


auth/httptransport/transport.go
```diff
@@ -205,28 +207,52 @@ type otelAttributeTransport struct {
...
+	}
+
+	var data *gax.TransportTelemetryData
+	if gax.IsFeatureEnabled("METRICS") {
+		data = gax.ExtractTransportTelemetry(req.Context())
+		if data != nil {
+			traceHook := &httptrace.ClientTrace{
+				GotConn: func(info httptrace.GotConnInfo) {
+					if info.Conn != nil && info.Conn.RemoteAddr() != nil {
+						host, portStr, err := net.SplitHostPort(info.Conn.RemoteAddr().String())
```
quartzmo: Can we optimize this by checking the boolean [info.Reused](https://pkg.go.dev/net/http/httptrace#GotConnInfo), and skipping the call to `net.SplitHostPort` when `true`? We should only need to parse for new connections. I'm not sure how much time is going to spent in `SplitHostPort`, but it [looks](https://cs.opensource.google/go/go/+/refs/tags/go1.26.1:src/net/ipsock.go;l=165) somewhat involved.

auth/grpctransport/grpctransport.go
```diff
@@ -475,15 +480,57 @@ func addOpenTelemetryStatsHandler(dialOpts []grpc.DialOption, opts *Options) []g
...
+	return func(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		transportData := gax.ExtractTransportTelemetry(ctx)
+		if transportData == nil {
+			return invoker(ctx, method, req, reply, cc, opts...)
+		}
+
+		err := invoker(ctx, method, req, reply, cc, opts...)
+
+		if target := cc.Target(); target != "" {
+			host, port := extractHostPort(target)
```
quartzmo: Running `extractHostPort` on every single unary call also calls the potentially expensive `net.SplitHostPort`. The [cc.Target()](https://pkg.go.dev/google.golang.org/grpc#ClientConn.Target) ought to be static for the client conn:

> Target returns the target string of the ClientConn.

Can we instead parse the host/port during the `dial`  slash `addOpenTelemetryStatsHandler` step and pass them as args from `addOpenTelemetryStatsHandler` to `openTelemetryUnaryClientInterceptor` here? I think the way to do it is by parsing the value from `transportCreds.Endpoint` (available in `dial`).

quartzmo (changes requested):

auth/grpctransport/grpctransport.go
westarle: done, this is much nicer.

auth/httptransport/transport.go
westarle: Now I just use req.URL.Hostname(), this should be faster.

auth/grpctransport/grpctransport.go
```diff
@@ -456,10 +458,14 @@ func (c *grpcCredentialsProvider) RequireTransportSecurity() bool {
...
 }
 
-func addOpenTelemetryStatsHandler(dialOpts []grpc.DialOption, opts *Options) []grpc.DialOption {
+func addOpenTelemetryStatsHandler(dialOpts []grpc.DialOption, opts *Options, endpoint string) []grpc.DialOption {
 	if opts.DisableTelemetry {
 		return dialOpts
 	}
+	if gax.IsFeatureEnabled("METRICS") {
+		host, port := extractHostPort(endpoint)
+		dialOpts = append(dialOpts, grpc.WithChainUnaryInterceptor(openTelemetryUnaryClientInterceptor(host, port)))
```
cpriti-os, line 467: @westarle Not supporting Streaming is intended?

auth/grpctransport/grpctransport.go
```diff
@@ -475,15 +481,56 @@ func addOpenTelemetryStatsHandler(dialOpts []grpc.DialOption, opts *Options) []g
...
 		logger:  scopedLogger,
 	}))
 }
 
+// Extract the host and port from a target address
+func extractHostPort(target string) (string, int) {
+	if idx := strings.Index(target, "://"); idx != -1 {
+		target = target[idx+3:]
+		// Ensure any leading slashes from the scheme suffix are stripped
+		for strings.HasPrefix(target, "/") {
```
cpriti-os: Can this loop be replaced with `strings.TrimLeft(target, "/")`

auth/grpctransport/grpctransport.go
westarle: for now, yes.

auth/grpctransport/grpctransport.go
westarle: done thank you!

auth/grpctransport/grpctransport.go
```diff
@@ -475,15 +481,54 @@ func addOpenTelemetryStatsHandler(dialOpts []grpc.DialOption, opts *Options) []g
...
 		logger:  scopedLogger,
 	}))
 }
 
+// Extract the host and port from a target address
+func extractHostPort(target string) (string, int) {
+	if idx := strings.Index(target, "://"); idx != -1 {
+		target = target[idx+3:]
+		// Ensure any leading slashes from the scheme suffix are stripped
+		target = strings.TrimLeft(target, "/")
```
quartzmo: Can we use `net/url.Parse` here to parse and separate scheme, authority, and path more reliably?

If a target is `dns://8.8.8.8/lb.example.com:443`, this manual parsing might inadvertently extract `8.8.8.8/lb.example.com` as the address, leaking internal DNS resolver details into public telemetry. It's kind of a contrived example tbh, but still...

auth/httptransport/transport.go
```diff
@@ -205,28 +206,53 @@ type otelAttributeTransport struct {
...
+				attrs = append(attrs, attribute.String("gcp.resource.destination.id", resName))
+			}
+			if resendCountStr, ok := callctx.TelemetryFromContext(req.Context(), "resend_count"); ok {
+				if count, err := strconv.Atoi(resendCountStr); err == nil {
+					attrs = append(attrs, attribute.Int("http.request.resend_count", count))
+				}
 			}
+			if urlTemplate, ok := callctx.TelemetryFromContext(req.Context(), "url_template"); ok {
+				attrs = append(attrs, attribute.String("url.template", urlTemplate))
+				span.SetName(fmt.Sprintf("%s %s", req.Method, urlTemplate))
```
quartzmo, line 225: `urlTemplate` is guaranteed to be low-cardinality here, right? (Just checking...)

auth/grpctransport/grpctransport.go
```diff
@@ -475,15 +481,54 @@ func addOpenTelemetryStatsHandler(dialOpts []grpc.DialOption, opts *Options) []g
...
+}
+
+// openTelemetryUnaryClientInterceptor returns an interceptor that populates
+// TransportTelemetryData with the server peer address.
+func openTelemetryUnaryClientInterceptor(host string, port int) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		transportData := gax.ExtractTransportTelemetry(ctx)
+		if transportData != nil {
+			if host != "" {
+				transportData.SetServerAddress(host)
```
quartzmo, line 521: The `*TransportTelemetryData` pointer usage here and in the HTTP equivalent is not completely thread-safe.

Assuming this code is called via gax.Invoke (which scopes context per call using `InjectTransportTelemetry`), this is safe. However, if a user uses the gRPC or HTTP wrapper directly for requests in parallel while sharing a base context, it could trigger a data race.

Typical practice (outside of gax.Invoke) is often to reuse a base context (e.g., context.Background()) across multiple concurrent requests, thus with this code possibly writing to the shared `data.SetServerAddress` at virtually the same time. This risks overwriting request-specific values when the hosts differ. Even worse, concurrent writes to a non-thread-safe struct field can trigger the race detector.

While I don't think that we absolutely need to fix this for usage via gax.Invoke, here's the possible solution:

```
// 1. Pull the pointer
orig := gax.ExtractTransportTelemetry(req.Context())

// 2. Clone it (Shallow copy is fine for primitive fields)
var clone gax.TransportTelemetryData
if orig != nil {
    clone = *orig 
}

// 3. Set host for THIS request only
clone.SetServerAddress(req.URL.Hostname())

// 4. Scope it to THIS request's context
req = req.WithContext(gax.InjectTransportTelemetry(req.Context(), &clone))
```

auth/httptransport/transport.go
quartzmo: Looks good! Nice solution.

auth/grpctransport/grpctransport.go
westarle: I liked the idea of using `net/url.Parse` but what I found is that it behaves badly on endpoints that aren't well formatted (like it treats bare "googleapis.com/xyz" as the scheme and explodes on eg. "[::1]:443"

It is a great catch on dns://... ; I fixed it.

auth/httptransport/transport.go
westarle: yes template needs to be low-cardinality like "/{name=projects/\*/things/\*}"

auth/grpctransport/grpctransport.go
westarle: I like the solution, but it breaks protocol with how invoke works today (today we allocate the struct in Invoke to avoid an allocation in the RPC path.)

I think you're saying that to experience the breakage a customer would have to
1. grab a raw Clients
2. inject a TransportTelemetryData into the context used for those clients
3. make parallel requests on the client

Maybe I can put a comment on TransportTelemetryData to say it's only for requests managed by Invoke?

quartzmo: > we allocate the struct in Invoke to avoid an allocation in the RPC path

Hmmm, good point!

quartzmo (approved):

PR #14265, opened 2026-03-26
feat(auth): Implement Actionable Errors logging for HTTP requests

westarle:
This PR introduces structured, OTel-compliant telemetry logging for actionable errors within the `google-cloud-go/auth` HTTP transport wrapper (`otelAttributeTransport`).

### DIFF
--- a/auth/httptransport/httptransport_otel_test.go
+++ b/auth/httptransport/httptransport_otel_test.go
@@ -23,20 +23,20 @@ import (
 	"log/slog"
 	"net/http"
 	"net/http/httptest"
-	"strings"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/googleapis/gax-go/v2"
+	"github.com/googleapis/gax-go/v2/callctx"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/sdk/trace/tracetest"
-	"github.com/googleapis/gax-go/v2/callctx"
 	oteltrace "go.opentelemetry.io/otel/trace"
 )
 
@@ -1086,19 +1086,19 @@ func TestErrorTrackingBody_EdgeCases(t *testing.T) {
 		bodySize := 10 * 1024
 		bodyData := bytes.Repeat([]byte("A"), bodySize)
 		resp := &http.Response{
-			StatusCode: 400,
-			Status:     "400 Bad Request",
-			Body:       io.NopCloser(bytes.NewReader(bodyData)),
+			StatusCode:    400,
+			Status:        "400 Bad Request",
+			Body:          io.NopCloser(bytes.NewReader(bodyData)),
 			ContentLength: -1,
 		}
-		
+
 		h, logBuf := setup(resp)
 		req, _ := http.NewRequestWithContext(context.Background(), "GET", "http://example.com", nil)
 		r, err := h.RoundTrip(req)
 		if err != nil {
 			t.Fatalf("RoundTrip error: %v", err)
 		}
-		
+
 		readBytes, err := io.ReadAll(r.Body)
 		if err != nil {
 			t.Fatalf("ReadAll error: %v", err)
@@ -1107,7 +1107,7 @@ func TestErrorTrackingBody_EdgeCases(t *testing.T) {
 			t.Errorf("read %d bytes, want %d", len(readBytes), bodySize)
 		}
 		r.Body.Close()
-		
+
 		if logBuf.Len() == 0 {
 			t.Fatal("expected log output")
 		}
@@ -1119,31 +1119,31 @@ func TestErrorTrackingBody_EdgeCases(t *testing.T) {
 		if msg != "400 Bad Request" {
 			t.Errorf("expected fallback message '400 Bad Request', got %q", msg)
 		}
-		})
+	})
 
-		t.Run("Early Close (Short Read)", func(t *testing.T) {
+	t.Run("Early Close (Short Read)", func(t *testing.T) {
 		bodyData := []byte(`{"error":{"message":"early close error"}}` + strings.Repeat(" ", 1000))
 		resp := &http.Response{
-			StatusCode: 400,
-			Status:     "400 Bad Request",
-			Body:       io.NopCloser(bytes.NewReader(bodyData)),
+			StatusCode:    400,
+			Status:        "400 Bad Request",
+			Body:          io.NopCloser(bytes.NewReader(bodyData)),
 			ContentLength: -1,
 		}
-		
+
 		h, logBuf := setup(resp)
 		req, _ := http.NewRequestWithContext(context.Background(), "GET", "http://example.com", nil)
 		r, err := h.RoundTrip(req)
 		if err != nil {
 			t.Fatalf("RoundTrip error: %v", err)
 		}
-		
+
 		buf := make([]byte, 10)
 		n, err := r.Body.Read(buf)
 		if err != nil || n != 10 {
 			t.Fatalf("expected 10 bytes, got %d, err %v", n, err)
 		}
 		r.Body.Close()
-		
+
 		if logBuf.Len() == 0 {
 			t.Fatal("expected log output on early close")
 		}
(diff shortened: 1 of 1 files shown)


auth/httptransport/transport.go
```diff
@@ -226,35 +255,201 @@ func (t *otelAttributeTransport) RoundTrip(req *http.Request) (*http.Response, e
...
+// responses are typically small and infrequent, the overhead of double-parsing is acceptable.
+// To prevent memory spikes and out-of-memory errors from unexpectedly large error responses,
+// we limit the bytes buffered in memory to 1MB. The original response will still be able to
+// read the full stream.
+func cloneResponse(resp *http.Response) (*http.Response, error) {
+	if resp == nil || resp.Body == nil || resp.Body == http.NoBody {
+		return resp, nil
+	}
+
+	const maxErrorReadBytes = int64(1 << 20) // 1MB
```
quartzmo: This seems like a lot, maybe risking memory spikes and GC pressure. Should we tighten the limit to something closer to typical api error payloads?

auth/httptransport/transport.go
```diff
@@ -226,35 +255,201 @@ func (t *otelAttributeTransport) RoundTrip(req *http.Request) (*http.Response, e
...
+		for k, v := range info.Metadata {
+			logAttrs = append(logAttrs, slog.String("gcp.errors.metadata."+k, v))
+		}
+
+		logAttrs = append(logAttrs, slog.String("error.type", info.ErrorType))
+		if info.StatusCode != "" {
+			logAttrs = append(logAttrs, slog.String("rpc.response.status_code", info.StatusCode))
+		}
+
+		logger.LogAttrs(ctx, slog.LevelDebug, msg, logAttrs...)
```
quartzmo, line 384: Any concerns about PII from the error being logged at DEBUG here? Just checking...

auth/httptransport/transport.go
```diff
@@ -180,33 +183,59 @@ func addOpenTelemetryTransport(trans http.RoundTripper, opts *Options) http.Roun
...
+			for _, attr := range attrs {
+				logAttrs = append(logAttrs, slog.String(string(attr.Key), attr.Value.AsString()))
+			}
+			scopedLogger = scopedLogger.With(logAttrs...)
+		}
 	}
+
 	otelOpts := []otelhttp.Option{
-		otelhttp.WithSpanOptions(trace.WithAttributes(staticAttrs...)),
+		otelhttp.WithSpanOptions(trace.WithAttributes(traceAttrs...)),
```
shollyman: Style questions:  Should we only set the option if tracing is enabled (e.g. in the IsFeatureEnabled block above)?  Does it warrant a length check on attrs?  I don't know the otel impl well enough under the hood to understand if an empty `trace.WithAttributes()` has special meaning or side effects.

auth/httptransport/transport.go
shollyman: consider hoisting this to top of file to make the constant more obvious/discoverable.

auth/httptransport/transport.go
```diff
@@ -226,35 +255,201 @@ func (t *otelAttributeTransport) RoundTrip(req *http.Request) (*http.Response, e
...
+
+	clone := *resp
+	clone.Body = io.NopCloser(bytes.NewReader(buf.Bytes()))
+	return &clone, err
+}
+
+// readResponseError attempts to extract an actionable error payload from the HTTP response.
+// It clones the response stream up to a safe limit, parses the JSON error, and ensures
+// a baseline fallback message is always populated to guarantee high-quality telemetry.
+func readResponseError(resp *http.Response) error {
```
shollyman: Should we simply not do response body parsing if the response is too large?  My thinking:

* Errors are typically smaller.  Even if they're huge we're just as likely to have too much info to stick in a telemetry attribute anyways.
* It avoids complexity and overhead from cloning huge responses.
* If we're only reading a subset of bytes then we're likely to end up with an unparseable partial response body.

auth/httptransport/transport.go
```diff
@@ -226,35 +255,201 @@ func (t *otelAttributeTransport) RoundTrip(req *http.Request) (*http.Response, e
...
+		return nil
+	}
+
+	errToParse := &googleapi.Error{
+		Code:    resp.StatusCode,
+		Message: resp.Status,
+	}
+
+	if resp.Body != nil {
+		clone, ioErr := cloneResponse(resp)
```
quartzmo: Go's `http.Client.Do()` is designed to return as soon as response headers are parsed. Deferring the error body peek until the client library caller executes `Read()` is critical for preserving standard Go `net/http` separation of concerns and architectural hygiene.

In standard Go HTTP transport patterns, the `RoundTripper` is low-level plumbing responsible for byte transportation, not message parsing. It should remain non-blocking and let the caller determine when and how to consume the body. 

If our `RoundTripper` eager-reads the body *inside* `RoundTrip`, the `Do()` call cannot return until the body read completes. This breaks standard `net/http` semantics by coupling header receipt with body consumption. If a server sends error headers but then stalls or hangs while sending the payload, the caller's `Do()` call hangs — robbing the caller of the ability to inspect the status code or fail fast based on headers alone. By deferring peeking via a wrapped body reader, we respect standard Go contracts and keep the transport layer fast and predictable.

A common mitigation is to defer peeking until the client calls `resp.Body.Read()` via a wrapped body reader. This is the same pattern used by the official OpenTelemetry Go library (`otelhttp`), where `otelhttp.Transport` wraps the response body and delays span completion until the stream is fully read or closed. 

This ensures that the transport layer is non-blocking and memory-efficient, letting the caller control the pace of reading. You can see this pattern in action in the `otelhttp` source code:
*   [otelhttp.Transport.RoundTrip (L182)](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/net/http/otelhttp/v0.67.0/instrumentation/net/http/otelhttp/transport.go#L182) (Where `res.Body` is wrapped)
*   [`wrappedBody.Read` (L242-L253)](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/net/http/otelhttp/v0.67.0/instrumentation/net/http/otelhttp/transport.go#L242-L253) (Where the span ends on `io.EOF`)
*   [`wrappedBody.Close` (L272-L278)](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/net/http/otelhttp/v0.67.0/instrumentation/net/http/otelhttp/transport.go#L272-L278) (Where the span ends on `Close`)

auth/httptransport/transport.go
```diff
@@ -362,94 +374,80 @@ func (t *otelAttributeTransport) onError(req *http.Request, resp *http.Response,
...
+				errToParse = &googleapi.Error{
+					Code:    b.resp.StatusCode,
+					Message: errResp.Error(),
 				}
 			}
 		}
 	}
 
-	return errToParse
+	b.t.logAndSpanError(b.req, b.resp, errToParse, nil, b.span, b.logger)
```
quartzmo, line 489: I think holding the lock while we perform this reading and logging is overkill. If reading, googleapi.CheckResponse and logging hangs for some reason, concurrent `Read` or `Close` calls on the same body will block. Instead, if the lock only protects flipping `recorded` to true, things still work: Once recorded is flipped, the buffer is effectively frozen and safe to read. If our reading/parsing/logging operation errors for some reason, then per our telemetry guidelines, we can just let the telemetry data go.

```go
func (b *errorTrackingBody) Close() error {
	b.mu.Lock()
	shouldRecord := !b.recorded
	b.recorded = true
	b.mu.Unlock()

	// We can close the network stream immediately.
	err := b.ReadCloser.Close()

	// 2. Do heavy in-memory telemetry parsing
	if shouldRecord {
		b.recordError() // If this panics here, the socket is already released
	}
	
	return err
}
```

auth/httptransport/transport.go
```diff
@@ -266,8 +269,24 @@ func (t *otelAttributeTransport) RoundTrip(req *http.Request) (*http.Response, e
 		return resp, err
 	}
 
-	if err != nil || resp.StatusCode >= 400 {
-		t.onError(req, resp, err, span, logger)
+	if err != nil {
+		t.logAndSpanError(req, resp, err, err, span, logger)
+	} else if resp.StatusCode >= 400 {
+		if resp.Body != nil && resp.Body != http.NoBody && (resp.ContentLength < 0 || resp.ContentLength <= maxErrorReadBytes) {
```
quartzmo, line 302: Just wondering if this gate on the response length might be premature: `resp.ContentLength <= maxErrorReadBytes`. If we remove this, won't the `errorTrackingBody` still just limit its reading to the first maxErrorReadBytes (8kb), and log that partial read? Is there any harm in logging the first 8kb of a longer response?

quartzmo: Hmm, on second look, I think the answer to my question is that sending truncated error JSON to googleapi.CheckResponse will probably mess things up. So this check ought to remain.

auth/httptransport/transport.go
westarle: done

auth/httptransport/transport.go
westarle: done!

auth/httptransport/transport.go
westarle: Good question, but I don't think so!

* AIP 193 says these should not contain PII
* The customer is responsible for reviewing logs go/client-libraries:o11y-controls
* We also log request/response payloads at DEBUG

quartzmo (approved):

PR #14279, opened 2026-03-27
chore: add librarian.yaml

JoeWang1127:
Add librarian.yaml, regenerate repository using google-cloud-go.

The googleapis SHA is the same as the last legacylibrarian generate ([59d5f2b](https://github.com/googleapis/googleapis/commit/59d5f2b46924714af627ac29ea6de78641a00835)).

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 language: go
-version: 0.9.0-72797a39eec2-20260327144318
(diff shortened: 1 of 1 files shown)


JoeWang1127:
> As discussed, the escaping in the snippet metadata is surprising and may impact downstream consumption, so we should investigate why there's a change here.

The json file is updated by `json.MarshalIndent`. By default, this function will convert  `&` to `\u0026` to prevent XSS attacks when JSON is embedded directly in HTML.

This conversion should have no impact on customers.

PR #14281, opened 2026-03-27
ci: add a workflow to verify regeneration

JoeWang1127:
Add a workflow to make sure generated files are not being manually edited.

The workflow runs once per day.

Fixes https://github.com/googleapis/librarian/issues/4422

### DIFF
--- a/.github/workflows/generation_check.yaml
+++ b/.github/workflows/generation_check.yaml
@@ -8,6 +8,7 @@ on:
   workflow_dispatch:
 permissions:
   contents: read
+  issues: write
 jobs:
   regeneration:
     runs-on: ubuntu-latest
@@ -24,3 +25,27 @@ jobs:
           go install golang.org/x/tools/cmd/goimports@latest
       - name: Regeneration
         run: ./.github/workflows/regenerate.sh
+      - name: Create issue on diff
+        if: failure()
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if [ -n "$(git status --porcelain)" ]; then
+            TITLE="Daily regeneration check found diff"
+            RUN_URL="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+            DIFF_STAT=$(git diff --stat)
+            BODY="The daily [regeneration check]($RUN_URL) found a diff.
+
+            Diff summary:
+            \`\`\`
+            $DIFF_STAT
+            \`\`\`"
+
+            EXISTING_ISSUE=$(gh issue list --state open --search "in:title \"$TITLE\"" --json number --jq '.[0].number')
+            if [ -z "$EXISTING_ISSUE" ]; then
+              gh issue create --title "$TITLE" --body "$BODY"
+            else
+              echo "Issue #$EXISTING_ISSUE already exists, adding a comment."
+              gh issue comment "$EXISTING_ISSUE" --body "Another failure with diff occurred: $RUN_URL"
+            fi
+          fi


shollyman (commented):
The goal of this check is to ensure there's no manual changes applied to files we consider to be automatically generated?

I think my biggest concern here is the cost of this validation since it runs on every presubmit change.  Can we move this to into our nightly testing, or run this is a prerequisite of releasing?  If there is a deviation it should be remediated at next regeneration, but perhaps I'm misunderstanding the risks here?

JoeWang1127:
> Can we move this to into our nightly testing, or run this is a prerequisite of releasing?

Nightly testing is fine, as long as we have a check to prevent manually editing of generated files. I'll make the change.

.github/workflows/regenerate.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo "==== Install protoc ===="
+curl -fsSL --retry 5 --retry-delay 15 -o /tmp/protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v25.7/protoc-25.7-linux-x86_64.zip
+sha256sum -c <(echo 877408bab02767938d1e5555f11c39dfe05e96f2a9571bc59dd2639f33d9954e /tmp/protoc.zip)
+sudo unzip -o /tmp/protoc.zip -d /usr/local
+protoc --version
+
+echo "==== Regenerate all the code ===="
+version=$(sed -n 's/^version: *//p' librarian.yaml)
+go run github.com/googleapis/librarian/cmd/librarian@${version} generate --all
```
codyoss, line 26: Where is the version value coming from here?

.github/workflows/generation_check.yaml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          go-version: '1.26.x'
+      - name: Install tools
+        run: |
+          go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.36.11 && \
+          go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3.0 && \
+          go install github.com/googleapis/gapic-generator-go/cmd/protoc-gen-go_gapic@v0.57.0 && \
+          go install golang.org/x/tools/cmd/goimports@latest
+      - name: Regeneration
+        run: ./.github/workflows/regenerate.sh
```
codyoss: nit: add newline.

.github/workflows/regenerate.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# If there is any difference between the generated code and the
+# committed code that is an error. All the inputs should be pinned,
+# including the generator version and the googleapis SHA.
+git diff --exit-code
```
codyoss: nit: add newline

.github/workflows/regenerate.sh
codyoss, line 26: What if there are changes since the last generation? For instance another commit has landed in googleapis. Would this not cause a diff?

JoeWang1127: In librarian.yaml, there's a config for librarian version:
```
version: v0.9.1-0.20260327183345-63a5ecfc3492
```

JoeWang1127: No, librarian will download the googleapis repo at a given commit SHA and the SHA is fixed in librarian.yaml.

So as long as we don't change librarian.yaml, the generated code should be the same.

.github/workflows/regenerate.sh
JoeWang1127: Done.

.github/workflows/generation_check.yaml
JoeWang1127: Done.

JoeWang1127:
Discussed with @shollyman, I'll update the workflow to create an issue if failed.

shollyman (approved):

PR #14285, opened 2026-03-29
chore: update header in internal/version.go

JoeWang1127:
Update the header in `internal/version.go` to include the same URL in librarian.

`internal/version.go` will be regenerated in legacylibrarian stage so we should keep the header the same across two tools.

### DIFF
--- a/internal/librariangen/module/_internal_version.go.txt
+++ b/internal/librariangen/module/_internal_version.go.txt
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      http://www.apache.org/licenses/LICENSE-2.0
+//      https://www.apache.org/licenses/LICENSE-2.0
 //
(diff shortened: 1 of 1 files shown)


internal/librariangen/module/_internal_version.go.txt
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      http://www.apache.org/licenses/LICENSE-2.0
+//      https://www.apache.org/licenses/LICENSE-2.0
```
codyoss, line 7: Are you sure this is correct? If I go to https://www.apache.org/licenses/LICENSE-2.0.txt it uses "http" there?

JoeWang1127: In https://apache.org/legal/src-headers.html#headers, it said https is acceptable.

If both are accepted, I think using https is better.

image

PR #14291, opened 2026-03-29
feat(bigtable): finalize all reason for direct access

sushanb:


bigtable/internal/transport/connpool.go
```diff
@@ -1155,49 +1162,179 @@ func (p *BigtableChannelPool) investigateDirectAccessFailure(originalErr error)
...
+		btopt.Debugf(p.logger, "bigtable_connpool: Direct Access investigation: xDS check failed: %v", err)
+		p.reportDirectAccessFailure(failReason)
+		return
+	}
+
+	if len(endpoints) == 0 {
+		btopt.Debugf(p.logger, "bigtable_connpool: Direct Access investigation: xDS returned 0 endpoints.")
+		p.reportDirectAccessFailure("xds_empty_endpoints")
+		return
+	}
```
nimf: This condition will never be true, because we return an error from the FetchXdsEndpoints if we got empty results. see https://github.com/googleapis/google-cloud-go/pull/14291/changes#diff-b08793c63504fefa787b7bbe411faf2dafc97b1407f5db145e571b0706c2229dR162

sushanb: Done.

sushanb:
> LGTM after vet fixes. Also, we should move all of the investigation logic (and its helpers) out of the bigtable conn pool, it has nothing to do with connection pooling.

I agree.

PR #14315, opened 2026-03-31
chore: update librariangen image SHA

JoeWang1127:
Update legacylibrarian Go image SHA, this image contains latest template of version.go.

Regenerate all libraries with:
```
# Sync library versions
go run github.com/googleapis/librarian/tool/cmd/syncversion@${V} .
# Regenerate
go run github.com/googleapis/librarian/cmd/librarian@${V} generate --all
```

### DIFF
--- a/bigtable/internal/version.go
+++ b/bigtable/internal/version.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      https://www.apache.org/licenses/LICENSE-2.0
+//     https://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -202,6 +202,7 @@ libraries:
       - path: google/cloud/beyondcorp/clientconnectorservices/v1
       - path: google/cloud/beyondcorp/clientgateways/v1
   - name: biglake
+    version: 0.0.0
     apis:
       - path: google/cloud/biglake/v1
       - path: google/cloud/biglake/hive/v1beta
@@ -247,7 +248,7 @@ libraries:
           import_path: bigquery/v2/apiv2
           path: google/cloud/bigquery/v2
   - name: bigtable
-    version: 1.44.0
+    version: 1.45.0
     apis:
       - path: google/bigtable/v2
       - path: google/bigtable/admin/v2
--- a/internal/generated/snippets/biglake/apiv1/snippet_metadata.google.cloud.biglake.v1.json
+++ b/internal/generated/snippets/biglake/apiv1/snippet_metadata.google.cloud.biglake.v1.json
@@ -8,7 +8,7 @@
     ],
     "language": "GO",
     "name": "cloud.google.com/go/biglake/apiv1",
-    "version": ""
+    "version": "0.0.0"
   },
   "snippets": [
     {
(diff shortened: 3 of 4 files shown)


bigtable/internal/version.go
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//      https://www.apache.org/licenses/LICENSE-2.0
+//     https://www.apache.org/licenses/LICENSE-2.0
```
jskeet, line 7: Are the other libraries already using https? It's odd to see only a single internal/version.go being changed here.

librarian.yaml
```diff
@@ -202,6 +202,7 @@ libraries:
       - path: google/cloud/beyondcorp/clientconnectorservices/v1
       - path: google/cloud/beyondcorp/clientgateways/v1
   - name: biglake
+    version: 0.0.0
```
jskeet, line 205: Is anything going to expect this version to exist as a tag? I note that biglake isn't in .librarian/state.yaml at all yet - is that deliberate? (And if it's going to be added, will *that* have version 0.0.0 or not? Because if it has something else, that would potentially end up being synchronized back to librarian.yaml...)

librarian.yaml
```diff
@@ -247,7 +248,7 @@ libraries:
           import_path: bigquery/v2/apiv2
           path: google/cloud/bigquery/v2
   - name: bigtable
-    version: 1.44.0
+    version: 1.45.0
```
jskeet, line 251: Why is this PR bumping the version of bigtable? (I think some more context on the PR might be useful.)

bigtable/internal/version.go
JoeWang1127: I overlooked the space in two templates (one from librarian and one from librariangen).

The difference is caused by a manual release of bigtable.

The new image will use the same template.

librarian.yaml
JoeWang1127: Go team manually release bigtable thus has this version update.

I used `syncversion` tool to update the version.

Moving forward, I've updated legacylibrarian release stage to update the version from state.yaml to librarian.yaml.

jskeet:
I now understand and agree with the bigtable changes. The use of 0.0.0 when that doesn't exist as a tag is still potentially worrying.

librarian.yaml
JoeWang1127: > Is anything going to expect this version to exist as a tag?

I tested the version, 0.0.0, using legacylibrarian. legacylibrarian doesn't expect this tag.

> I note that biglake isn't in .librarian/state.yaml at all yet - is that deliberate?

biglake is already in state.yaml, the version is 0.0.0.

The reason I added this version is b/c `librarian add` didn't assign an initial version. It did cause a problem in legacylibrarian release stage so I had to manually add a version in state.yaml.

Later, I fixed the problem in librarian (https://github.com/googleapis/librarian/pull/4951) and sync the version from state.yaml.

JoeWang1127:
> I now understand and agree with the bigtable changes. The use of 0.0.0 when that doesn't exist as a tag is still potentially worrying.

0.0.0 is used as a sign of not finding tags in the language repo: https://github.com/googleapis/librarian/blob/1144f1c1baf392135d8c621caaecb7c3f4826ba9/internal/legacylibrarian/legacylibrarian/release_stage.go#L215

librarian.yaml
jskeet: Apologies, I thought I'd looked at all of state.yaml, but I hadn't.

I've found the code in legacylibrarian now: https://github.com/googleapis/librarian/blob/main/internal/legacylibrarian/legacylibrarian/release_stage.go#L215C2-L215C32

All good.

PR #14322, opened 2026-04-01
feat(firestore): refactor pipeline API for uniform stage options

bhshkh:
Design decision [here](https://docs.google.com/document/d/1diPUhXr2kZh69KcAsuOMZMj_rfcykGsE2-phmBIt8vs/edit?usp=sharing)

**Overview**
This PR standardizes the API surface for the Firestore Pipeline builder in the Go SDK, ensuring idiomatic conformance and aligning closely with cross-platform design paradigms (Java/Node.js). It replaces complex stage-specific configuration structs with a universal functional option pattern and introduces typed slice constants for varied expressions.

**Key Features and API Changes:**

*   **Universal `RawOptions` Escape Hatch:**
    Introduced `RawOptions` to serve as a universal pass-through for backend options that might not be natively modeled by the Go SDK (e.g., `RawOptions{"limit": 5, "distance_field": "dist"}`). This deprecates intermediate parameter structs like `PipelineFindNearestOptions` across pipeline stages. 

*   **Standardized Variadic Arguments with Slices:** 
    Methods that formerly relied on raw variadic argument unpacking (e.g. `...Ordering`, `...any`) have been updated to accept typed slices. To preserve clean query ergonomics, we introduced domain-specific helper constructors (`Orders()`, `Fields()`, `Accumulators()`, `Selectables()`).
    *   `Pipeline.Sort(Orders(Ascending(FieldOf("f")), Descending(FieldOf("__name__"))))`
    *   `Pipeline.Aggregate(Accumulators(Count(DocumentID).As("total")))`
    *   `Pipeline.Select(Fields("a", "b"))`

*   **Functional Stage Options Migration:**
    Implemented standard functional options for specialized stage modifiers:
    *   `WithFindNearestLimit()`, `WithFindNearestDistanceField()`
    *   `WithAggregateGroups()`
    *   `WithUnnestIndexField()`
    *   `WithExplainMode()`

**Internal Code Simplification & Testing:**

*   **Flattened Stage Hierarchy (`baseStage` removal):**
    Removed the `baseStage` struct embedding across our internal pipeline stages. We replaced it with explicit `options map[string]any` fields within each stage and introduced a shared `stageOptionsToProto(options)` helper. This simplifies the internal struct hierarchy, removes opaque state inherited from struct embedding, and makes configuration serialization significantly more explicit and easier to trace inside each stage's `toProto()` implementation.
*   **Pipeline Stage Proto Parsing:** 
    Refactored internal `pipelineStage` structures (`newAggregateStage`, `newFindNearestStage`, `newSortStage`, etc.) to reliably map the new `RawOptions` down to standard `*pb.Value` payloads.
*   **Tests for `AlwaysUseImplicitOrderBy`:**
    Added dedicated test cases (`TestQuery_AlwaysUseImplicitOrderBy`) to explicitly verify the behavior of the existing `WithAlwaysUseImplicitOrderBy` client toggle, ensuring it appropriately serializes implicit inequality fields and `__name__` into the `OrderBy` proto requests.
*   **Integration Suite Migration:**
    Fully migrated the integration suite inside `pipeline_integration_test.go` and `pipeline_test.go` to assert against the newly stabilized `RawOptions` and slice-builder interfaces.

### DIFF
--- a/bigquery/v2/apiv2_smoketests/job_integration_test.go
+++ b/bigquery/v2/apiv2_smoketests/job_integration_test.go
@@ -52,9 +52,12 @@ func TestInitStatelessQuery(t *testing.T) {
 			// assert that it must be the case.
 			if bv := queryResp.GetJobComplete(); bv != nil && bv.Value {
 				if jobRef := queryResp.GetJobReference(); jobRef != nil {
-					// We ended up with a job.  Ensure there's a reason at least.
-					if queryResp.GetJobCreationReason() != nil {
-						t.Error("there's a job reference in the response but no reason for it")
+					// We got a job reference, at least make sure it has the expected fields populated.
+					if jobRef.GetProjectId() == "" {
+						t.Errorf("JobReference present but ProjectId not populated.")
+					}
+					if jobRef.GetJobId() == "" {
+						t.Errorf("JobReference present but JobId not populated.")
 					}
 				} else {
 					if rowcount := len(queryResp.GetRows()); rowcount != 1 {


firestore/pipeline_integration_test.go
```diff
@@ -650,7 +642,8 @@ func TestIntegration_PipelineStages(t *testing.T) {
 		})
 	})
 	t.Run("Select", func(t *testing.T) {
-		iter := client.Pipeline().Collection(coll.ID).Select("title", "author.name").Limit(1).Execute(ctx).Results()
+		t.Skip("Skipping functional test failure")
```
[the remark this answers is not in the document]
bhshkh: Done

firestore/pipeline_integration_test.go
```diff
@@ -815,10 +808,12 @@ func TestIntegration_PipelineStages(t *testing.T) {
 		}
 	})
 	t.Run("UnnestWithIndexField", func(t *testing.T) {
+		t.Skip("Skipping functional test failure")
+		t.Skip("Skipping functional test failure")
```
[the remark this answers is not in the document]
bhshkh: Done

firestore/pipeline_integration_test.go
```diff
@@ -1160,6 +1155,7 @@ func TestIntegration_Query_Pipeline(t *testing.T) {
 	})
 
 	t.Run("Select", func(t *testing.T) {
+		t.Skip("Skipping functional test failure")
```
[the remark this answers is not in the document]
bhshkh: Done

firestore/pipeline_integration_test.go
```diff
@@ -1281,44 +1277,44 @@ func objectFuncs(t *testing.T) {
...
+		// 	name:     "MapValues",
+		// 	pipeline: client.Pipeline().Collection(coll.ID).Select(Fields(MapValues("m1").As("values"))),
+		// 	want:     map[string]interface{}{"values": []interface{}{int64(1), int64(2)}},
+		// },
+		// {
+		// 	name:     "MapEntries",
+		// 	pipeline: client.Pipeline().Collection(coll.ID).Select(Fields(MapEntries("m1").As("entries"))),
+		// 	want:     map[string]interface{}{"entries": []interface{}{map[string]interface{}{"k": "a", "v": int64(1)}, map[string]interface{}{"k": "b", "v": int64(2)}}},
+		// },
 	}
```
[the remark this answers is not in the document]
bhshkh: Done

firestore/pipeline_integration_test.go
```diff
@@ -1366,107 +1362,107 @@ func arrayFuncs(t *testing.T) {
...
-		{
-			name:     "ArrayFilter",
-			pipeline: client.Pipeline().Collection(coll.ID).Select(ArrayFilter("a", "x", GreaterThan(FieldOf("x"), 1)).As("filter")),
-			want:     map[string]interface{}{"filter": []interface{}{int64(2), int64(3)}},
-		},
+		// {
+		// 	name:     "ArrayFilter",
+		// 	pipeline: client.Pipeline().Collection(coll.ID).Select(Fields(ArrayFilter("a", "x", GreaterThan(FieldOf("x"), int64(1))).As("filter"))),
+		// 	want:     map[string]interface{}{"filter": []interface{}{int64(2), int64(3)}},
+		// },
```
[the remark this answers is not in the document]
bhshkh: Done

firestore/pipeline.go
```diff
@@ -112,21 +112,115 @@ func WithExplainMode(mode ExplainMode) ExecuteOption {
...
-func WithRawExecuteOptions(options map[string]any) ExecuteOption {
-	return newFuncExecuteOption(func(eo *executeSettings) {
-		if eo.RawOptions == nil {
-			eo.RawOptions = make(map[string]any)
-		}
-		for k, v := range options {
-			eo.RawOptions[k] = v
-		}
-	})
+type RawOptions map[string]any
```
shollyman, line 129: I don't have a good sense of how well defined the keys are for the general options mechanism.  My read here is that the map is used to invalidate a previous option with the same key, as opposed to other option patterns where its just a list that allows potential multiple modifications.  Is that the driving factor here?

Do we have any problems with key conflicts across these disparate option interfaces?  If that's a property we want do we need a test that ensures all defined options are uniquely keyed?

bhshkh: 1. Is that the driving factor here ?
  Yes, exactly. The driving factor for using a map[string]any is that it directly mirrors the underlying Firestore Protocol Buffer definition. 

   In the protobuf, a pipeline stage is defined as:
    ```
        message Stage {
          string name = 1;
          repeated Value args = 2;
          map<string, Value> options = 3; // <-- The backend expects a map
        }
    ```
   Because the backend enforces unique string keys for options per stage, the Go SDK uses a map to naturally accumulate them. When a user passes multiple options (e.g.,
  pipeline.Collection("users", opt1, opt2)), the SDK iterates through them and applies them to a single map.    This gives us standard Go "last-writer-wins" semantics: if opt2 writes to the same
  key as opt1, it intentionally overrides it.

  2. Do we have any problems with key conflicts across disparate option interfaces?
  No, because the map is scoped per stage instance. There is no global options map. When a user creates a stage, a brand-new, empty map is instantiated exclusively for that specific stage:

      ```
       func (ps *PipelineSource) Collection(path string, opts ...CollectionOption) *Pipeline {
           options := make(map[string]any) // <-- Brand new map scoped only to this stage
           for _, opt := range opts {
               if opt != nil {
                   opt.applyStage(options)
               }
           }
           // ...
      ```
      Therefore, if FindNearest uses the key "limit" and a theoretical future Sort option also uses the key "limit", they will never conflict because their options are written to completely isolated maps.

bhshkh:
update and delete stages were [added](https://github.com/googleapis/google-cloud-go/pull/14331) while the current PR was in review. Refactored those to match the rest of the stages and their options.


Also, added a test for unique keys.

firestore/pipeline.go
bhshkh: added a test for unique keys per stage

firestore/pipeline.go
```diff
@@ -347,206 +509,271 @@ func (p *Pipeline) Offset(offset int) *Pipeline {
...
+}
+
 // AddFields adds new fields to outputs from previous stages.
 //
 // This stage allows you to compute values on-the-fly based on existing data from previous
 // stages or constants. You can use this to create new fields or overwrite existing ones (if there
 // is name overlaps).
 //
-// The added fields are defined using [Selectable]s
+// The added fields are defined using [Selectable]s.
```
hongalex: Right now the linking isn't working. Adding an apostrophe works, or you could just omit the s entirely.

```suggestion
// The added fields are defined using [Selectable]'s.
```

firestore/pipeline.go
```diff
@@ -347,206 +509,271 @@ func (p *Pipeline) Offset(offset int) *Pipeline {
...
+			g = []any{}
+		}
+		ao["groups"] = append(g, groups...)
+	})
 }
 
 // Aggregate performs aggregation operations on the documents from previous stages.
 // This stage allows you to calculate aggregate values over a set of documents. You define the
 // aggregations to perform using [AliasedAggregate] expressions which are typically results of
 // calling [AggregateFunction.As] on [AggregateFunction] instances.
```
hongalex: I know it wasn't part of the PR but this link doesn't exist.

```suggestion
// calling AggregateFunction.As on [AggregateFunction] instances.
```

firestore/pipeline_source.go
```diff
@@ -14,13 +14,6 @@
...
 
-import (
-	"fmt"
-	"reflect"
-
-	pb "cloud.google.com/go/firestore/apiv1/firestorepb"
-)
-
 // PipelineSource is a factory for creating Pipeline instances.
 // It is obtained by calling [Client.Pipeline()].
```
hongalex:
```suggestion
// It is obtained by calling [Client.Pipeline].
```

firestore/transaction_test.go
```diff
@@ -347,7 +347,7 @@ func TestTransactionErrors(t *testing.T) {
 			if err := tx.Delete(c.Doc("C/a")); err != nil {
 				return err
 			}
-			p := c.Pipeline().Collection("C").Select("x")
+			p := c.Pipeline().Collection("C").Select([]any{"x"})
```
hongalex: Consider using `Fields` to encourage that behavior (in future maintainers / LLMs)

bhshkh: Done

firestore/pipeline.go
shollyman: Thanks for the explanations!

PR #14324, opened 2026-04-01
fix(storage): close HTTP2 response bodies to prevent flow-control leaks

krishnamd-jkp:
Fixes an issue where HTTP2 stream connections might exhaust flow-control quota and cause INTERNAL_ERROR in long-lived connections.

- Modifies `httpReader.Read` to properly close the previous stream body before reopening it.
- Adds a defer block in `parseReadResponse` to ensure response bodies aren't leaked when an error is returned during header parsing.
- Ensures bodies are closed when stream discarding (`io.CopyN`) or `X-Goog-Generation` header parsing fails in `readerReopen`. These fixes apply equally to both the JSON and XML reader code paths.

### DIFF
--- a/storage/http_client.go
+++ b/storage/http_client.go
@@ -1563,7 +1563,7 @@ func parseReadResponse(res *http.Response, params *newRangeReaderParams, reopen 
 		// the total size.
 		size, err = strconv.ParseInt(cr[strings.LastIndex(cr, "/")+1:], 10, 64)
 		if err != nil {
-			err = fmt.Errorf("storage: invalid Content-Range %q", cr)
+			err = fmt.Errorf("storage: invalid Content-Range %q: %w", cr, err)
 			return nil, err
 		}
 


storage/http_client.go
```diff
@@ -1547,20 +1559,23 @@ func parseReadResponse(res *http.Response, params *newRangeReaderParams, reopen
...
+			err = fmt.Errorf("storage: invalid Content-Range %q", cr)
+			return nil, err
 		}
 		// Content range is formatted <first byte>-<last byte>/<total size>. We take
 		// the total size.
 		size, err = strconv.ParseInt(cr[strings.LastIndex(cr, "/")+1:], 10, 64)
 		if err != nil {
-			return nil, fmt.Errorf("storage: invalid Content-Range %q", cr)
+			err = fmt.Errorf("storage: invalid Content-Range %q", cr)
+			return nil, err
```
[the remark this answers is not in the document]
cpriti-os: Fix this?

storage/http_client.go
```diff
@@ -1515,14 +1519,18 @@ func readerReopen(ctx context.Context, header http.Header, params *newRangeReade
 			//      https://cloud.google.com/storage/docs/transcoding#range,
 			// thus we have to manually move the body forward by seen bytes.
 			if decompressiveTranscoding(res) && seen > 0 {
-				_, _ = io.CopyN(io.Discard, res.Body, seen)
+				if _, err := io.CopyN(io.Discard, res.Body, seen); err != nil {
+					res.Body.Close()
+					return err
```
cpriti-os: This is now throwing an error instead of silently failing, can there be scenarios where successful user workflows start failing because of this new error?

krishnamd-jkp: This is not harmful because if error is thrown in between stream and we do not return an error, the next read in http reader in line1403 will see the error again when reading res.Body. I'll remove it anyways as this error path is covered.

PR #14327, opened 2026-04-01
chore: add a config check

JoeWang1127:
Add a presubmit check to verify library name and version are consistent among `state.yaml` and `librarian.yaml`.

Update the version of pubsub and pubsub/v2 since these two libraries are released with an old version of legacylibrarian. Using legacylibrarian v0.9.1 and above will update version in librarian.yaml automatically.

For https://github.com/googleapis/librarian/issues/4422
Fixes https://github.com/googleapis/librarian/issues/4943

### DIFF
--- a/.github/workflows/config_check.yml
+++ b/.github/workflows/config_check.yml
@@ -2,13 +2,9 @@
 name: config check
 on:
   pull_request:
-    paths:
-      - librarian.yaml
-      - .librarian
   push:
-    paths:
-      - librarian.yaml
-      - .librarian
+    branches:
+      - main
 permissions:
   contents: read
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/config_check.yml
```diff
@@ -0,0 +1,28 @@
+---
+name: config check
+on:
+  pull_request:
+  push:
+    branches:
```
shollyman: Can we add path filtering as well so this runs only if the PR touches the relevant yaml files?

internal/generated/snippets/pubsub/v2/apiv1/snippet_metadata.google.pubsub.v1.json
```diff
@@ -8,7 +8,7 @@
     ],
     "language": "GO",
     "name": "cloud.google.com/go/pubsub/v2/apiv1",
-    "version": "2.5.0"
+    "version": "2.5.1"
```
shollyman, line 11: Unrelated, or you were trying to induce a difference for testing?

JoeWang1127: This updates is caused by https://github.com/googleapis/google-cloud-go/pull/14312

The release was done via an old version of legacylibrarian (v0.8.3) so the version in librarian.yaml is not updated. The latest version of legacylibrarian (v0.9.1 and above) will update version in librarian.yaml as well.

The check detected the version discrepancy and I updated it to be same in state.yaml.

.github/workflows/config_check.yml
JoeWang1127: Done.

PR #14332, opened 2026-04-03
feat(auth): update otelgrpc and otelhttp to v0.67.0

westarle:
These update gRPC telemetry to include semconv 1.40 attributes:
* `rpc.system.name` instead of `rpc.system`
* `rpc.method` is `"$service/$method"` and `rpc.service` is removed
* `rpc.grpc.status_code` is removed (in favor of `rpc.response.status_code`


auth/grpctransport/grpctransport_otel_test.go
```diff
@@ -245,10 +238,8 @@ func TestDial_OpenTelemetry_Enabled(t *testing.T) {
 					Code: codes.Unset,
 				},
 				Attributes: []attribute.KeyValue{
-					keyRPCStatusCode.Int64(0),
-					keyRPCMethod.String("Echo"),
-					keyRPCService.String("echo.Echoer"),
-					keyRPCSystem.String(valRPCSystemGRPC),
+					keyRPCMethod.String("echo.Echoer/Echo"),
+					attribute.String("rpc.system.name", "grpc"),
```
shollyman, line 242: Why not just use/update the key constants at the top of the file?

westarle: I'd like to move things to strings https://github.com/googleapis/google-cloud-go/issues/14347

so I don't want to introduce keyRPCSystemName

and I can't bring myself to set keyRPCSystem to the new "rpc.system.name".  I want to see the string in the test.

I'd prefer to do the cleanup in a future PR, lmk if you want it sooner!

shollyman: That's fine for now, just wanted to check since the tests were already favoring the key constants.

PR #14368, opened 2026-04-07
test(firestore): optimize integration tests and fix resource leaks

bhshkh:
Fixes: #14350

Without these optimizations: `cloud.google.com/go/firestore	1459.794s`
With these optimizations:       `cloud.google.com/go/firestore	79.084s`
The execution time is reduced by 94.58%

**1. Optimization: Pre-built Indexes**
Previously, tests that required composite or vector indexes dynamically generated new collection IDs, built the required indexes programmatically, and waited for them to complete. This added significant overhead (often several minutes) to each test suite run. 

* **Fix:** Tests requiring indexes (`TestIntegration_QueryDocuments_WhereEntity`, `TestIntegration_QueryUnary`, `TestIntegration_AggregationQueries`, `TestIntegration_AggregationQueries_WithRunOptions`, and `TestIntegration_FindNearest`) now use a static subcollection name (`"indexed_collection"`) under a uniquely generated parent document. This ensures isolation across parallel runs while allowing tests to leverage persistent, pre-built indexes.
* **Cleanup:** Removed all programmatic index management logic and API clients used strictly for index creation/deletion during tests.
* **Docs:** Updated `CONTRIBUTING.md` with the required `gcloud firestore indexes composite create` commands to build these indexes manually, aligning the Firestore testing workflow with the existing Datastore setup.

**2. Fixing Resource Leaks**
The test suite was permanently leaking documents due to incomplete cleanup logic and early process termination.
* **Fix `TestMain` Execution Order:** Modified `TestMain` so `cleanupIntegrationTest()` runs unconditionally before `os.Exit()`. Previously, if `m.Run()` failed, the process would terminate immediately and bypass the global cleanup.
* **Eradicate "Missing" Documents:** Updated the global `deleteCollection` utility to use `DocumentRefs(ctx)` instead of `Documents(ctx)`. `Documents()` only returns existing documents, skipping over empty parent documents and permanently stranding their subcollections. `DocumentRefs()` uses the underlying `ListDocuments` API to ensure all paths—including missing parents—are found and recursively cleaned.
* **Enforce Test Hygiene:** Added missing `t.Cleanup` hooks to several tests (e.g., BulkWriter and Aggregation tests) and moved existing cleanup registrations to occur immediately after `NewDoc()`. This guarantees that cleanup handlers are registered even if the subsequent backend creation times out or fails.

**3. Aligning with Firestore Native Behavior**
* **Fix:** Updated `Aggregations_with_limit` to explicitly use `.OrderBy()` for the `Limit` test to enforce deterministic result sets. 
**4. Flaky Test Fixes**
* **Fix `TestDetectProjectID`:** Explicitly clear the `GOOGLE_CLOUD_PROJECT` environment variable using
      `t.Setenv("GOOGLE_CLOUD_PROJECT", "")` before attempting to instantiate a client without credentials. This
      prevents the test from inadvertently detecting a project ID and failing in environments where
      `GOOGLE_CLOUD_PROJECT` is already set.

### DIFF
--- a/firestore/integration_test.go
+++ b/firestore/integration_test.go
@@ -167,7 +167,6 @@ func initIntegrationTest() {
 	integrationTestStruct.Ref = refDoc
 }
 
-
 // deleteCollection recursively deletes the documents in the specified collection
 func deleteCollection(ctx context.Context, coll *CollectionRef) error {
 	bulkwriter := iClient.BulkWriter(ctx)
@@ -2679,7 +2678,7 @@ func TestIntegration_BulkWriter_Create(t *testing.T) {
 			wantStatusCode: codes.AlreadyExists,
 		},
 	}
-	
+
 	// Add successfully created testcases ref to cleanup list
 	for _, tc := range testcases {
 		if tc.wantStatusCode == codes.OK {


firestore/integration_test.go
```diff
@@ -164,151 +161,23 @@ func initIntegrationTest() {
...
-			if err != nil {
-				r.Errorf("Failed to delete index \"%s\": %+v\n", indexName, err)
-			}
-		})
-	}
-}
-
 // deleteCollection recursively deletes the documents in the specified collection
 func deleteCollection(ctx context.Context, coll *CollectionRef) error {
 	bulkwriter := iClient.BulkWriter(ctx)
```
quartzmo, line 172: Is a new `BulkWriter` created for every subcollection? Can we optimize this by passing the `BulkWriter` down to `deleteCollection` (or a helper) to share a single instance across the entire tree deletion?

